PR #1, opened 2019-04-03
Add recording support with Reaper Project generation

pljones:
Includes the following changes

* Initial .gitignore
Administrative

* Fix up warning message
* Not all Windows file systems are case insensitive
Bugfixes

* (Qt5) Use QCoreApplication for headless
Possible solution to get the application to run as a headless server but it loses the nice history graph, so not ideal.

* Avoid ESC closing chat
Because ESC shouldn't close the chat window. Or the main app window.

* Add console logging support for Windows
Whilst looking for the headless support, I found this idea for Windows logging.  New improved version.  This makes far fewer changes.

----

* Add recording support with Reaper Project generation
The main feature!
    * New -r option to enable recording of PCM files and conversion to Reaper RPP with WAV files
    * New -R option to set the directory in which to create recording sessions
    You need to specify the -R option, there's no default... so I guess -r and -R could be combined.
    * New -T option to convert a session directory with PCM files into a Reaper RPP with WAV files
    You can use -T on "failed" sessions, if the -r option captures the PCMs but the RPP converter doesn't run for some reaon. (It was useful during development, maybe less so once things seem stable.)

The recorder is implemented as a new thread with queuing from the main "real time" server thread.

When a new client connects or if its audio format changes (e.g. mono to stereo), a new RIFF WAVE file is started.  Each frame of decompressed audio for each client written out as LPCM to the file.  When the client disconnects, the RIFF WAVE headers are updated to reflect the file length.

Once all clients disconnect, the session is considered ended and a Reaper RPP file is written.


pljones:
Sorry, wrong place for the pull origin.

corrados:
> You need to specify the -R option, there's no default... so I guess -r and -R could be combined.
I only use the -R option on my branch. If with -R a directory is given, the recording is started. So I removed the -r option.

> You can use -T on "failed" sessions, if the -r option captures the PCMs but the RPP converter doesn't run for some reaon. (It was useful during development, maybe less so once things seem stable.)
I will remove it on my integration branch.

pljones:
> 
> 
> > You need to specify the -R option, there's no default... so I guess -r and -R could be combined.
> 
> I only use the -R option on my branch. If with -R a directory is given, the recording is started. So I removed the -r option.

The only thing I thought of is that "-R <dir>" could mean "record the WAV files into <dir>" and "-r" could mean "create a Reaper RPP" (maybe switching -r and -R), which would allow for adding support for other project formats in future... but as it's only Reaper, there's no reason to mess things around currently.

PR #6, opened 2019-05-19
SVG history graph

pljones:
**This isn't ready to merge.**

http://jamulus.drealm.info/Jamulus/history.svg

I've got the code running on my local server and it "works" in as much as it generates viewable SVG that looks pretty much the same as the JPEG version.  I've kept the code as close as I could to the history graph code, too.

However... there's some clear problems with the maths in this version, which I don't understand...  I didn't think I'd changed anything.

Ideally, the grid calculation code could be separated out from the rendering code, so it would be only at the last minute (i.e. Update or even Save) that it decided to construct a JPEG or an SVG file.

I'll have another look tomorrow (i.e. I'll take a break from staring at the code...) but please feel free to comment.

Thanks,

-- Peter

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -158,7 +158,6 @@ HEADERS += src/audiomixerboard.h \
     src/connectdlg.h \
     src/global.h \
     src/clientdlg.h \
-    src/historygraph.h \
     src/serverdlg.h \
     src/multicolorled.h \
     src/multicolorledbar.h \
@@ -169,13 +168,15 @@ HEADERS += src/audiomixerboard.h \
     src/settings.h \
     src/socket.h \
     src/soundbase.h \
-    src/svghistorygraph.h \
     src/testbench.h \
     src/util.h \
     src/analyzerconsole.h \
     src/recorder/jamrecorder.h \
     src/recorder/creaperproject.h \
-    src/recorder/cwavestream.h
+    src/recorder/cwavestream.h \
+    src/historygraph.h \
+    src/jpeghistorygraph.h \
+    src/svghistorygraph.h
 
 HEADERS_OPUS = libs/opus/include/opus.h \
     libs/opus/include/opus_multistream.h \
@@ -268,12 +269,14 @@ SOURCES += src/audiomixerboard.cpp \
     src/settings.cpp \
     src/socket.cpp \
     src/soundbase.cpp \
-    src/svghistorygraph.cpp \
     src/util.cpp \
     src/analyzerconsole.cpp \
     src/recorder/jamrecorder.cpp \
     src/recorder/creaperproject.cpp \
-    src/recorder/cwavestream.cpp
+    src/recorder/cwavestream.cpp \
+    src/historygraph.cpp \
+    src/jpeghistorygraph.cpp \
+    src/svghistorygraph.cpp
 
 SOURCES_OPUS = libs/opus/src/opus.c \
     libs/opus/src/opus_decoder.c \
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -158,6 +158,7 @@ HEADERS += src/audiomixerboard.h \
     src/connectdlg.h \
     src/global.h \
     src/clientdlg.h \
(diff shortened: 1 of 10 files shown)


pljones:
OK, now I think it's ready.

corrados:
Hi Peter, I will merge your changes into an integration branch. There I will do some modifications to your code before I merge it to the master. That will take some time. In the meantime it would be good if you do not do any changes on the same code to avoid merge conflicts. Best Regards, Volker

pljones:
Hi @corrados 

I've spotted a couple of issues, the main one being use of `unsigned int` where the code originally used a signed value.  I've added an appropriate `static_cast` as the value held should be unsigned but it's usage is signed at certain places.

Also, I've adjusted the plotting of the points - maybe I missed these bits when converting.
```
diff --git a/src/historygraph.cpp b/src/historygraph.cpp
index ae4c703..dae378e 100644
--- a/src/historygraph.cpp
+++ b/src/historygraph.cpp
@@ -168,7 +168,7 @@ void AHistoryGraph::DrawFrame ( const int iNewNumTicksX )
     {
         int         iBottomExtraTickLen = 0;
         const int   iCurX = gridFrameX + static_cast<int> ( dayXSpace * ( i + 1 ) );
-        const QDate curXAxisDate = curDate.addDays ( i - iNumTicksX + 1 );
+        const QDate curXAxisDate = curDate.addDays ( 0 - static_cast<signed>( iNumTicksX ) + i + 1 );

         // text (print only every "iXAxisTickStep" tick)
         if ( !( i % iXAxisTickStep ) )
@@ -219,7 +219,7 @@ void AHistoryGraph::AddMarker ( const SHistoryData& curHistoryData )
         curDate.daysTo ( curHistoryData.DateTime.date() );

     // check range, if out of range, do not plot anything
-    if ( -iXAxisOffs > ( static_cast<int>(iNumTicksX) - 1 ) )
+    if ( -iXAxisOffs > ( static_cast<signed>( iNumTicksX ) - 1 ) )
     {
         return;
     }
@@ -229,9 +229,9 @@ void AHistoryGraph::AddMarker ( const SHistoryData& curHistoryData )
         static_cast<double> ( curHistoryData.DateTime.time().minute() ) / 60;

     // calculate the actual point in the graph (in pixels)
-    int curPointX = gridFrameX + static_cast<int> ( dayXSpace * ( iNumTicksX + iXAxisOffs ) );
+    int curPointX = gridFrameX + static_cast<int> ( dayXSpace * ( static_cast<signed>( iNumTicksX ) + iXAxisOffs ) );
     int curPointY = gridFrameY + static_cast<int> ( static_cast<double> (
-        gridFrameHeight ) / ( iYAxisEnd - iYAxisStart ) * dYAxisOffs );
+        gridFrameHeight ) / ( static_cast<signed>( iYAxisEnd ) - iYAxisStart ) * dYAxisOffs );
     QString curPointColour = MarkerNewColor;
     int curPointSize = iMarkerSizeNewCon;

@@ -252,5 +252,5 @@ void AHistoryGraph::AddMarker ( const SHistoryData& curHistoryData )
         break;
     }

-    point( curPointX, curPointY, curPointSize, curPointColour );
+    point( curPointX - curPointSize / 2, curPointY - curPointSize / 2, curPointSize, curPointColour );
 }
```

corrados:
Hi Peter (@pljones),
I have not spent any time on the integration yet. So the code is not modified yet. So the best thing for me would be to wait for you until things have settled and then I will start the integration work. We are not on a hurry anyway ;-). So just apply your changes and if you are really done with all the coding, simply create a new merge request to my branch "corrados:integration_history".

Two thing you may want to do yourself instead of me doing it on the integration branch. I am not a fan of having too many source files. So I would merge jpeghistorygraph and svghistorygraph into historygraph so that we only have one new file. Also, I would like to have the Open Source header in all of the source files.
Best Regards,
Volker

pljones:
I've been doing too much java at work, hence one file per class disease... ;)

OK, I'll make the updates.

PR #7, opened 2019-05-26
Debian/Ubuntu .deb building with Icon for eg. Gnome

braindef:
Hello corrados
I could do so far what you've suggested.

-Its now all in the directory *distributions*
-I added the desktop launcher (I needed to edit your the .desktop file)
-the built release for Debian 9+ and Ubuntu 18+ is in "releases"

With kind regards
Marc


distributions/build-debian-package.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+sudo apt-get install devscripts build-essential lintian dh-make
+sudo apt-get install libqt4-dev libjack-jackd2-dev
```
corrados: Hi Marc, the compile info on the Jamulus homepage is a bit outdated. Jamulus should be compiled with Qt5 instead of Qt4. I am not 100 % sure what is the aquivalent devel package for Qt5 as it was libqt4-dev for Qt4. Do you know the name of the packet?

distributions/debian/control
```diff
@@ -0,0 +1,15 @@
+Source: jamulus
+Section: sound
+Priority: optional
+Maintainer: "Marc Landolt jr" <debian@marclandolt.ch>
+Build-Depends: debhelper (>= 9), libc6-dev (>=2.24) | libc-dev (>=2.24),g++ (>= 4:8.3), make (>=4.1), dpkg-dev (>= 1.17.11), libqt4-dev, libjack-jackd2-dev
```
corrados: See my above comment about Qt4.

distributions/debian/control
```diff
@@ -0,0 +1,15 @@
...
+Standards-Version: 3.9.5
+Homepage: http://http://llcon.sourceforge.net/
+Vcs-Git: git://github.com/corrados/jamulus.git
+Vcs-Browser: https://github.com/corrados/jamulus
+ 
+Package: jamulus
+Architecture: amd64
+Depends: ${shlibs:Depends}, ${misc:Depends}, libqt4-dev, qtchooser, jackd
```
corrados: See my above comment about Qt4.

src/res/jamulus.desktop
```diff
@@ -2,8 +2,8 @@
 Name=Jamulus
 Comment=Jam Session
 GenericName=Internet Jam Session Software
-Exec=jamulus
-Icon=jamulus
+Exec=Jamulus
+Icon=Jamulus
```
corrados: Jamulus can be configured to use a lower case binary. This can be done via qmake command line argument. Since usually Linux tools are with lower letter, maybe for the Debian package you should also make use of that option.
Another thing: What about moving this file in the new distributions file? Or do you think the current location is ok?

corrados:
Thanks Marc. I have added some review comments to your changes files. Best Regards, Volker

src/res/jamulus.desktop
braindef: this is no problem. Shall I or have you already changed that?

distributions/build-debian-package.sh
braindef: ok, i will checkt that, its maybe qtdeclarative5-dev, but i must test first, i took the installed packages from https://github.com/corrados/jamulus/blob/master/INSTALL#L42 so you shold change that too. after i tested successfully i will answer here what exact library is needed and if it works with qt5

src/res/jamulus.desktop
corrados: Do you mean the lower case binary of the movement of the file? You can find the change for the lower case name here: "https://github.com/corrados/jamulus/commit/127e9ac7866d5b750a3b17568d24bc81aa2f192f#diff-ff31b6d635c626acb697d69ef0e864bd".
The file location I did not change. Feel free to move it if you like.

distributions/build-debian-package.sh
corrados: If you like you can do the change in the INSTALL file yourself. The only thing I have to do afterwards is to update the Jamulus Homepage as well.

braindef: well I am still testing, i have some problems with qt5/qjackctl in virtualBOX, still debugging it. Because otherwise I would need to re-setup computers for each test. So I prefere to do the tests on a virtualBOX image, so I can use snapshots. Still not working. i will tell you the results... btw: I like challenges -- at least those sort of challenges

src/res/jamulus.desktop
braindef: It seems that this does not have an effect on the filename. As far as I know qmake takes the name for the final binary from the Jamulus.pro file name, so you (or I) would change it to lower case then the binary would also be lower case. At least on debian and ubuntu it worked that way. So if you change that, the destination file will be lower case. Two things why I think it would be better if you'd do that: i can not say if it has any impact on the windows build. i can not say if it has any impact on the OSX build. So I could get a appropriate build-env on Windows and I don't own a Mac / X-Code

distributions/build-debian-package.sh
braindef: didn't work with virtualBOX, so tested the with new physical installation.

src/res/jamulus.desktop
braindef: i just realized that you probably want to intentionally have uppercase J for Windows / OSX. i did now use the parameter you gave me: https://github.com/braindef/jamulus/blob/master/distributions/debian/rules#L6

braindef:
I pulled your changes to my fork. I will make a new pull request so you will not get conflicts. I merged 3h Ago, so if you sleep at the moment there are good chances that you can merge without conflicts

PR #12, opened 2019-09-23
Travis-ci

braindef:
[https://marclandolt.ch/corrados/00.png](url)
[https://marclandolt.ch/corrados/01.png](url)

Hi Corrados

you need to move the two files that i downlaod with curl too... 00.png then it works...
https://travis-ci.org/braindef/jamulus/builds/588260536

on the second picture you see that i use "pwd" to see the current path, and as you see in the 4 commits, i just do trial and error. 01.png

Greetings Marc

PS: remember to change all occurances from brainef to corrados... i should have used variables but i have way much troubles in my live since i have my IT Degree...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,7 @@ matrix:
     - mkdir ASIOSDK2
     - cp -rvu ASIOSDK2.3.2/* ASIOSDK2
     - pwd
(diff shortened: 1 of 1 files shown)


corrados:
Thank you very much for your fixes :-).

PR #13, opened 2020-03-20
Enhance historygraph to allow size based on date

pljones:
Due to the increasing number of connections, the *number* of history
entries is no longer a good way to size the width of the graph.
In addition to adding a much larger FIFO, I have chosen to add
a date constraint.  This means I do not have to keep making the FIFO
bigger daily and, when numbers fall, the graph will not change
shape.


corrados:
Good idea, thank you.

PR #14, opened 2020-03-21
Feature/covid19 date limit on historygraph

pljones:
OK, so an extension of the previous change (the previous one was the "easy" bit).

This adds support for passing the history graph number of nodes and days as program arguments.

(And I tuned the point sizes a little.)

### DIFF
--- a/src/global.h
+++ b/src/global.h
@@ -92,12 +92,8 @@ LED bar:      lbr
 // file name for logging file
 #define DEFAULT_LOG_FILE_NAME           "Jamulussrvlog.txt"
 
-/* History Definitions ****************************************************************/
-// default number of history items to store
-#define NUM_ITEMS_HISTORY           6400
-// default oldest item to draw
-#define MAX_DAYS_HISTORY            60
-/* *** */
+// Default oldest item to draw in history graph (days ago)
+#define DEFAULT_DAYS_HISTORY            60
 
 // default server address
 #define DEFAULT_SERVER_ADDRESS          "jamulus.fischvolk.de"
--- a/src/global.h
+++ b/src/global.h
@@ -92,6 +92,13 @@ LED bar:      lbr
 // file name for logging file
 #define DEFAULT_LOG_FILE_NAME           "Jamulussrvlog.txt"
 
+/* History Definitions ****************************************************************/
+// default number of history items to store
+#define NUM_ITEMS_HISTORY           6400
+// default oldest item to draw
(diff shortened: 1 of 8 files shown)


corrados:
Hi Peter, I had done some code style changes to your previous merge request. Bevor you create new merge requests, please first pull again from the official repo.
In my opinion we do not need the number of nodes command line argument. The number of days is sufficient. So please remove remove the number of nodes input. Thanks and Best Regards

corrados:
Oh, I just saw in the Network graph that you have indeed used the latest commit to base your new work on. So just ignore the first part of my previous message.

pljones:
600 items was covering over 60 days up until the start of the covid 19 crisis.  10 days ago traffic started rising such that whereas I was getting around 10 lines a day, now I'm getting 200 lines a day.

So I don't know now long 4800 will last.

Are there any performance implications of storing more FIFO items for the history than is really necessary?

But really, rather than recompiling when the 4800 runs out (as it will), I'd like to just up the number of FIFO entries passed in and restart.  Then, if (when) traffic drops, I can drop the size of the FIFO, again without recompiling.

corrados:
In my opinion it will never make sense to show that much days. The graph would look ugly in that case. Just make the number of nodes even higher, e.g. 20000. I do not think we will run into performance issues if that number is a large one. But keep in mind, setting this number to, e.g., 20000 would mean that the graph would contain 20000 dots. In that case you would not see any meaningful information anyway. Even if we have 4800 dots in the graph is too much for getting meaningful information out of the graph other that the server is always in use. What do you think?

pljones:
That's what the day count does.  No matter how many dots, it then skips until the oldest is less than that many days old.  So I now have more than enough dots for my 60 days - but the graph only displays 60 days.

corrados:
That is clear. Just a bit of background: when I created the history graph, I wanted to have a simple solution for the time period shown in the graph. So I just took the number of dots as a simple mechanism. But your new implementation, specifying the actual number of days to be displayed is a better implementation. So the number of dots are less important. That is the reason I recommend just to specify the number of days for the length of the history graph at the command line and Jamulus internally takes care that we have enough space in the buffer to full the requested number of day. Of course, some limit of that buffer is also meaningful to avoid a memory problem but this has not to be set by the command line in my opinion.

pljones:
I understand.  I am still tuning the value for my server but it a large FIFO isn't an issue, I don't need to do that.

PR #18, opened 2020-03-25
Use pkg-config to find Jack

vandenoever:
The hardcoded directories for jack are not always correct.
It is more flexible to use pkg-config to find Jack.


corrados:
Thanks for your patch. I'll try it out on my laptop today.

corrados:
It worked. Thank you for your patch.

PR #26, opened 2020-03-29
Enable scrolling for mixer board

doloopuntil:
Wraps the mixer board in a QScrollArea so a horizontal scroll bar appears when more clients are connected than the window can display.


corrados:
Thank you very much for your reference code. I have now finished the implementation on the master branch, see: https://github.com/corrados/jamulus/commit/bef3506a38b65066ff77a6cf72558ca5a46c2a69
Your reference helped me a lot.

PR #36, opened 2020-04-03
Feature/channel meters

pljones:
Hi Volker,

OK, so it works now.

I've tested with a private instance of the server running the new code and a single instance of the client running the new code, so not exactly extensive.  One upside of the "opt in" is that the server doesn't send out the levels to _all_ clients, only those opted in - which means new clients, as only they can opt on.

Naturally, I'm totally, 100% certain nothing could possibly go wrong.

Let me know if you can see anything I could have done better or if I've missed any style compliance.

Thanks

-- Peter


src/protocol.cpp
```diff
@@ -628,6 +628,14 @@ if ( rand() < ( RAND_MAX / 2 ) ) return false;
         case PROTMESSID_CLM_REQ_CONN_CLIENTS_LIST:
             bRet = EvaluateCLReqConnClientsListMes ( InetAddr );
             break;
+
+        case PROTMESSID_CLM_REQ_CHANNEL_LEVEL_LIST:
```
corrados, line 632: I am missing the new protocol description in the header of this file. All other protocol messages are described there.

src/server.cpp
```diff
@@ -431,6 +431,18 @@ CServer::CServer ( const int          iNewMaxNumChan,
...
         this, SLOT ( OnCLReqConnClientsList ( CHostAddress ) ) );
 
+    QObject::connect ( &ConnLessProtocol,
+        SIGNAL ( CLReqChannelLevelList ( CHostAddress, bool ) ),
+        this, SLOT ( OnCLReqChannelLevelList ( CHostAddress, bool ) ) );
+
+    qRegisterMetaType< Int16_t >();
+    qRegisterMetaType< CVector< int16_t > >();
+    qRegisterMetaType< CVector< CVector< int16_t > > >();
+    QObject::connect ( this,
```
corrados, line 441: Hmm. You are including the entire audio buffers in the message. Do you think we can get performance issues by this?

src/server.cpp
```diff
@@ -989,6 +1001,26 @@ JitterMeas.Measure();
...
+                if ( vecChannels[ vecChanIDsCurConChan[i] ].ChannelLevelsRequired() )
+                {
+                    emit AudioFrames ( static_cast<int16_t>(iNumClients),
+                                       vecChanIDsCurConChan,
+                                       vecvecsData );
+                    break;
+                }
+            }
+        }
+        iFrameCount++;
```
corrados, line 1022: I would put this in an "else {}"

src/server.cpp
```diff
@@ -989,6 +1001,26 @@ JitterMeas.Measure();
...
+        if ( iFrameCount > CHANNEL_LEVEL_UPDATE_TIME )
+        {
+            iFrameCount = 0;
+
+            // Enable channel levels if any client has requested them
+            for ( int i = 0; i < iNumClients; i++ )
+            {
+                if ( vecChannels[ vecChanIDsCurConChan[i] ].ChannelLevelsRequired() )
+                {
+                    emit AudioFrames ( static_cast<int16_t>(iNumClients),
```
corrados, line 1015: That is not gonna work correctly. I tried your version out on my laptop and I saw that if I made a short noise, the level meter would not show any input. So what you do here is to measure the level at an arbitrary instant in time. But what you would have to do is to measure the level continuously and smooth it by some filter operation and then pick the current level at the channel_level_update_time. Of course this requires additional calculations at the server but this would be the right way of doing it.

src/server.h
```diff
@@ -112,6 +113,16 @@ class CHighPrecisionTimer : public QThread
 };
 #endif
 
+// Minimal int16_t wrapper for queuing
```
corrados, line 116: Can you explain in short words why you need an additional class here?

corrados:
I added some comments in your code (see above).

src/server.h
pljones: Because `int16_t` is a `#define` not a type.  The queuing declarations required real types, as far as I could tell.  I _think_ I could have used `short` - but that would then have been inconsistent with the datatype for the value in question.

src/server.cpp
pljones: True enough but it's not intended to be perfect, only a rough indicator that there's some degree of activity.  A short burst isn't enough to do that.  It's only tracking one in 200 frames, for instance, so any burst could be missed entirely.

src/server.cpp
pljones: I guess I could pass the buffer to the calculation method and then emit just the result.  That would save the copy but run the calculation in the main thread (although, only once every 200 frames).

src/protocol.cpp
pljones: They were in the same commit with this code, dfb3ea7:
```
diff --git a/src/protocol.h b/src/protocol.h
index 59ed83a..e0c17f4 100755
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -71,6 +71,8 @@
 #define PROTMESSID_CLM_REQ_VERSION_AND_OS     1012 // request version number and operating system
 #define PROTMESSID_CLM_CONN_CLIENTS_LIST      1013 // channel infos for connected clients
+#define PROTMESSID_CLM_REQ_CHANNEL_LEVEL_LIST 1015 // request the channel level list
+#define PROTMESSID_CLM_CHANNEL_LEVEL_LIST     1016 // channel level list

 // lengths of message as defined in protocol.cpp file
 #define MESS_HEADER_LENGTH_BYTE         7 // TAG (2), ID (2), cnt (1), length (2)
```

src/server.cpp
pljones: Each frame causes the frame count to increment.  Putting it in an else would mean the first frame of each 200 did _not_ cause the count to increment, meaning 201 frames per update.

src/server.cpp
pljones: It would also save having to define `Int16_t`, too, as that would be passed directly to the method.

src/protocol.cpp
corrados: Sorry, I was a bit unclear. I meant something like this: https://github.com/corrados/jamulus/blob/master/src/protocol.cpp#L67

src/server.cpp
corrados: Ah, ok. Thanks for the clarification.

src/server.cpp
corrados: Ok, I would have to see it how it performs if many clients are connected and we have multiple mixer faders. BTW, I saw that the new LED bar next to the fader did have more than just three LEDs. Why did you use a larger number of LEDs?

src/protocol.cpp
pljones: Aha!  I've not seen that before (not having needed it).  I'll have a go.

src/server.cpp
pljones: I've added smoothing in the latest commit.

src/server.cpp
pljones: OK, I've moved where the calculation is done.  I don't think (given it's only for opted in clients and only if at least one is opted in) it'll cause too much CPU stress and it does save the CPU and memory stress of the large copy operation, so I'd agree it's a win.

corrados: Great, thanks.

corrados:
Just for your information, I have compiled your new code on my server and I have started the "centest, Oberrad, Germany" server with your latest code on your Git branch. I have also compiled the code from your branch on my Windows PC and I connected the centest. Then I could see my mixer fader light but there was anotherone connected to the server and he was playing but I could not see his LEDs changing at all. They were all grey all the time.
You can test it yourself if you like.

src/server.cpp
pljones: Thinking about it... We're currently using 2 bits per level 0 to 3; we could get 0 to 15 by using 4 bits per level, which would be "full precision" for the level meter.  (Easier than 3 bits...)

pljones:
> ... Then I could see my mixer fader light but there was another one connected to the server and he was playing but I could not see his LEDs changing at all. They were all grey all the time.
OK, I'll take a look - possibly I'm writing the values to the wrong place (`i` rather than `iChId`) or something.  Thanks.

src/server.cpp
corrados: Ok, you can try it if you like. 4 bits should not be too bad in terms of network load.

corrados:
Hi Peter, I would like to do some coding in the server for 64 samples frame size support. That may interfere with your work. I hope that you do not get too many merge conflicts by this... Just that you know what I am currently working on.

src/server.cpp
```diff
@@ -1591,12 +1591,69 @@ void CServer::OnAudioFrames ( const Int16_t                    iNumClients,
...
+
     CVector<uint16_t> vecChannelLevels ( iNumClients );
     for ( int i = 0; i < iNumClients; i++ )
     {
         const int iChId = vecChanIDsCurConChan[i];
-        vecChannelLevels[iChId] = 3; // TODO: work out the value from vecvecsData[iChId]
+        const int iStereoVecSize = vecvecsData[iChId].Size();
+        short     sMax = 0;
+
+        for ( int i = 0; i < iStereoVecSize; i += 6 ) // 2 * 3 = 6 -> stereo
```
pljones: Um.  Um...  Line 1605 `for ( int i...`.  Line 1611 `for ( int i....`.

Um.

pljones:
I'm going to close this pull request due to the major repair.  I'll resubmit with a cleaner/clearer commit history...

PR #37, opened 2020-04-04
Store mute settings for faders.

doloopuntil:
Also increase stored fader settings capacity to 200, since up to 50 clients are now supported.

I found this feature useful when singing with other people. We usually mute ourselves as we don't need to hear our own voice (slightly late) in the mix coming back from the server. It also prevents very noisy feedback if someone forgets to plug their headphones in before connecting to the server. By saving the mute settings, it's much easier to start a session and being ready to sing. Feel free to reject the pull request if you don't think this feature is useful in general.

### DIFF
--- a/src/chatdlg.cpp
+++ b/src/chatdlg.cpp
@@ -98,7 +98,7 @@ void CChatDlg::AddChatText ( QString strChatText )
 #if QT_VERSION < QT_VERSION_CHECK(5, 0, 0)
         txvChatWindow, 0, QAccessible::ValueChanged
 #else
-        new QAccessibleEvent ( txvChatWindow, QAccessible::ValueChanged )
+        new QAccessibleValueChangeEvent ( txvChatWindow, strChatText )
 #endif
         );
 }
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -250,6 +250,11 @@ void CChannelFader::SetFaderIsSolo ( const bool bIsSolo )
     pcbSolo->setChecked ( bIsSolo );
 }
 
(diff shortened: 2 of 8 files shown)


corrados:
Thank you for the code. I did not store the Mute settings on purpose because the fader setting is already stored. So you simply can move your fader down and this is like your channel is muted.

BTW: I know that for singers it is very hard to sing to the delayed signal from the server. But if you mute that signal and only hear the local signal, you will not be in sync with the others. You might run into the problem that you get slower and slower if the musicians try to match the speed of the other singers.

src/chatdlg.cpp
```diff
@@ -98,7 +98,7 @@ void CChatDlg::AddChatText ( QString strChatText )
 #if QT_VERSION < QT_VERSION_CHECK(5, 0, 0)
         txvChatWindow, 0, QAccessible::ValueChanged
 #else
-        new QAccessibleEvent ( txvChatWindow, QAccessible::ValueChanged )
+        new QAccessibleValueChangeEvent ( txvChatWindow, strChatText )
```
corrados, line 101: I just noticed that you changed the QAccessibleEvent. Can you please explain what was wrong with the previous version? Did it not work as expected?

doloopuntil: When I built and ran the code in Qt 5.9, using the Debug configuration, this line would raise an assertion in [qaccessible.h](https://github.com/qt/qtbase/blob/dev/src/gui/accessible/qaccessible.h#L675), so I followed the instructions in the Qt code and used the recommended QAccessibleEvent 
 subclass.

corrados: Ah, ok. Thanks for the info. BTW: I merged your code. Thank you for your contribution.

PR #38, opened 2020-04-04
Update raspijamulus.sh

fredsiva:
adding libjack-jackd2-dev to the list of packages so that the compilation works on a Pi4.  
Without this, the make command command of lines 78 and 79 did not produce any result.

Tested on 3 consecutive full PI builds to make sure this was the root cause of my problems.
Happy to discuss!

### DIFF
--- a/distributions/raspijamulus.sh
+++ b/distributions/raspijamulus.sh
@@ -5,7 +5,7 @@ OPUS="opus-1.1"
 NCORES=$(nproc)
 
 # install required packages
-pkgs='alsamixergui build-essential qt5-default libasound2-dev cmake libglib2.0-dev'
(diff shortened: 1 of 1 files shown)


corrados:
In that script I compile my own version of jack, therefore I do not need the official _dev package.

fredsiva:
I see there is indeed a compilation of jack, but in my case, the raspijamulus.sh script does NOT work until I add the _dev package.

corrados:
I worked for me on a clean Raspbian. Maybe something with the compilation of jack did not work the first place. My script simply checks if the directory exists and if yes, it assumes that the compilation was correct. Can you delete the distributions/jack2 directory and try again without the _dev package installed?
As far as I remember, that official jack _dev package even interfered with my special version in some way but I cannot remember. If this _dev package is needed, something else does not work as expected.

PR #39, opened 2020-04-04
Feature/channel meters

pljones:
OK, I've reworked the commits into a nice clean series.

I've tested locally with two clients, both supporting the channel levels.  Each can see the levels, when they want to, for the client transmitting audio.

This version isn't compatible with the previous version!  (4 bits per channel now.)

I've a remaining concern over what happens on a busier server where server channels and clients are changing a lot (so you could get a new client joining on channel 0 or only channels 0 and 4 being in use) - I'm not quite sure what happens then.  I don't remember clearly whether I tested that yet - I'll give it a go tomorrow (with two clients where the earlier joining one disconnects).

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -64,6 +64,7 @@ CClient::CClient ( const quint16  iPortNumber,
     bFraSiFactDefSupported           ( false ),
     bFraSiFactSafeSupported          ( false ),
     eGUIDesign                       ( GD_ORIGINAL ),
+    bDisplayChannelLevels            ( false ),
     bJitterBufferOK                  ( true ),
     strCentralServerAddress          ( "" ),
     bUseDefaultCentralServerAddress  ( true ),
@@ -278,6 +279,9 @@ void CClient::OnNewConnection()
     // Same problem is with the jitter buffer message.
     Channel.CreateReqConnClientsList();
     CreateServerJitterBufferMessage();
+
+    // send opt-in / out for Channel Level updates
+    SetDisplayChannelLevels ( bDisplayChannelLevels );
 }
 
 void CClient::CreateServerJitterBufferMessage()
@@ -384,6 +388,14 @@ bool CClient::GetAndResetbJitterBufferOKFlag()
     // since per definition the jitter buffer status is OK if both the
     // put and get status are OK
     return bSocketJitBufOKFlag;
+}
+
+void CClient::SetDisplayChannelLevels ( const bool bNDCL )
+{
+    bDisplayChannelLevels = bNDCL;
+
+    // tell any connected server about the change
+    ConnLessProtocol.CreateCLReqChannelLevelListMes ( Channel.GetAddress(), bNDCL );
 }
 
 void CClient::SetSndCrdPrefFrameSizeFactor ( const int iNewFactor )
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -29,7 +29,8 @@
 CClient::CClient ( const quint16  iPortNumber,
                    const QString& strConnOnStartupAddress,
                    const int      iCtrlMIDIChannel,
-                   const bool     bNoAutoJackConnect ) :
+                   const bool     bNoAutoJackConnect,
+                   QTextStream&   tsNC ) :
     vstrIPAddress                    ( MAX_NUM_SERVER_ADDR_ITEMS, "" ),
     ChannelInfo                      (),
     vecStoredFaderTags               ( MAX_NUM_STORED_FADER_SETTINGS, "" ),
@@ -45,7 +46,9 @@ CClient::CClient ( const quint16  iPortNumber,
     bWindowWasShownChat              ( false ),
     bWindowWasShownProfile           ( false ),
     bWindowWasShownConnect           ( false ),
+    tsConsole                        ( tsNC ),
     Channel                          ( false ), /* we need a client channel -> "false" */
+    ConnLessProtocol                 ( tsConsole ),
     eAudioCompressionType            ( CT_OPUS ),
     iCeltNumCodedBytes               ( OPUS_NUM_BYTES_MONO_LOW_QUALITY ),
     eAudioQuality                    ( AQ_NORMAL ),
@@ -190,6 +193,10 @@ CClient::CClient ( const quint16  iPortNumber,
         SIGNAL ( CLVersionAndOSReceived ( CHostAddress, COSUtil::EOpSystemType, QString ) ),
         SIGNAL ( CLVersionAndOSReceived ( CHostAddress, COSUtil::EOpSystemType, QString ) ) );
 #endif
+
+    QObject::connect ( &ConnLessProtocol,
+        SIGNAL ( CLChannelLevelListReceived ( CHostAddress, CVector<uint16_t> ) ),
+        this, SLOT ( OnCLChannelLevelListReceived ( CHostAddress, CVector<uint16_t> ) ) );
 
     // other
     QObject::connect ( &Sound, SIGNAL ( ReinitRequest ( int ) ),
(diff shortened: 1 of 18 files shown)


corrados:
I just updated the "centest" server with your new code. Right now only a few people are on the servers. Let's wait until this evening when more people are active to see how it performs with many people connected to the server.
First tests showed that the 4 bits really seem to make a noticeable difference. I hope the protocol network traffic is not too high with this but my feeling is that it should be just fine.

src/client.cpp
```diff
@@ -29,7 +29,8 @@
 CClient::CClient ( const quint16  iPortNumber,
                    const QString& strConnOnStartupAddress,
                    const int      iCtrlMIDIChannel,
-                   const bool     bNoAutoJackConnect ) :
+                   const bool     bNoAutoJackConnect,
+                   QTextStream&   tsNC ) :
```
corrados, line 33: I assume you introduced this to debug your new protocol messages. Can you please remove this (and the other places, too)?
When I do debugging during implementation I have made good experience with the following:
qDebug() << "This is the integer I am interested in: " << 12345;
Works great for me. Did you know this command?

src/protocol.h
```diff
@@ -71,6 +71,8 @@
 #define PROTMESSID_CLM_REQ_VERSION_AND_OS     1012 // request version number and operating system
 #define PROTMESSID_CLM_CONN_CLIENTS_LIST      1013 // channel infos for connected clients
 #define PROTMESSID_CLM_REQ_CONN_CLIENTS_LIST  1014 // request the connected clients list
+#define PROTMESSID_CLM_REQ_CHANNEL_LEVEL_LIST 1015 // request the channel level list
```
corrados, line 74: Hm, I can see that you have used a connection-less message also for the enabling of the feature. I have not studied your code in great detail yet, but if this is the case, we can have issues with that. E.g. if you have temporary network issues and you switch on the LEDs at the client at that moment, it may happen that the REQ_CHANNEL_LEVEL does not make it through to the server. In that case the GUI control shows that LED is enabled but the server does not send any levels.
So you should use a normal message for enabling the feature (for normal messages we have a check for lost packets and a re-send mechanism) and, what you already did, a connection-less message for the transmit of the levels from the server to the client to get the lowest network traffic for the updates. If one connection-less packet with the levels does not make it to the client is not an issue at all since we have continuous updates of the levels.
Do you agree with this?

src/protocol.h
```diff
@@ -71,6 +71,8 @@
...
+#define PROTMESSID_CLM_CHANNEL_LEVEL_LIST     1016 // channel level list
```
corrados, line 75: As I wrote above: For this message (updating the levels) it is perfect to use a connection-less message (just to make sure for that for this message type no change is required).

corrados:
One more thing: Can you please update the manual for the new feature you added to the Settings dialog: https://github.com/corrados/jamulus/blob/master/src/res/homepage/manual.md

src/client.cpp
```diff
@@ -64,6 +67,7 @@ CClient::CClient ( const quint16  iPortNumber,
     bFraSiFactDefSupported           ( false ),
     bFraSiFactSafeSupported          ( false ),
     eGUIDesign                       ( GD_ORIGINAL ),
+    bDisplayChannelLevels            ( false ),
```
corrados, line 70: I would enable the new functionality by default, i.e. for a new user which had not started Jamulus before, the channel meters should appear. So I would set there here to "true" instead of "false".

src/client.cpp
pljones: Ah, no - I wasn't.  And yes, I'd meant to remove the debugging code.  Thanks!

src/protocol.h
pljones: Oh right, yes, that makes sense.

src/client.cpp
pljones: OK - and yes, I'll get onto the manual changes, too.

corrados: Sorry for giving you a hard time. But you should know me from previous code deliveries so you knew what to expect ;-).

src/protocol.h
pljones: Hm.  OK, looking through the code, I don't spot an example I can directly follow for the request.

For the CLM, the client "implicitly" sends its address, which the server uses to determine the channel in question.

For the standard request, clients do not appear to send their address -- but I don't see how the server then knows _which_ channel the request is for - some how, the "emit" just seems to get there.

Can you explain what happens - i.e. how I get the client opt in to the server right channel?

src/client.cpp
pljones: Oh, you're no more vicious than I am when reviewing code :).

src/protocol.h
corrados: You could take a look at, e.g., the "CreateJitBufMes". No need to set the addess since the client is in connected state and the address is already stored in the channel.

corrados: The problem is that you have to find the correct place to send the message. You sould put it in "OnNewConnection()" to update the current setting of the client in the server.

pljones: Yes, it's there.  src/client.cpp:
```
 288     CreateServerJitterBufferMessage();
 289
 290     // send opt-in / out for Channel Level updates
 291     SetDisplayChannelLevels ( bDisplayChannelLevels );
```

pljones: So...
CClient::CreateServerJitterBufferMessage()
-> Channel::CreateJitBufMes( ... ) -> Protocol::CreateJitBufMes( ... )
-> Protocol::CreateAndSendMessage (  )
...
-> CProtocol::EvaluateJitBufMes ( ... ) -> emit ChangeJittBufSize ( ... )
-> CProtocol SIGNAL ( ChangeJittBufSize ( ... ) ) -> CChannel SLOT ( OnJittBufSizeChange )

But which channel?  Magic happened.  server.cpp never sees the signal.

CChannel::OnJittBufSizeChange then leads through into
```
src/server.h:574:    void OnServerAutoSockBufSizeChangeCh0  ( int iNNumFra ) { vecChannels[0].CreateJitBufMes  ( iNNumFra ); }
src/server.h:575:    void OnServerAutoSockBufSizeChangeCh1  ( int iNNumFra ) { vecChannels[1].CreateJitBufMes  ( iNNumFra ); }
src/server.h:576:    void OnServerAutoSockBufSizeChangeCh2  ( int iNNumFra ) { vecChannels[2].CreateJitBufMes  ( iNNumFra ); }
s
```
But, again, I still don't see how the server knew which channel received the message to route the signal correctly.

corrados: > CChannel SLOT ( OnJittBufSizeChange ) [...] But which channel?

If that message would be your new message. Then in function OnJittBufSizeChange() in the "if ( bIsServer )" section you would update your new variable bChannelLevelsRequired.
But maybe I did not fully understand your question.

corrados: BTW: The "AutoSockBufSize" is something different and should not be used as a reference.

pljones: So it sounds like there is magic happening in the "connected" message handling that "just knows" where the message came from so it goes to the right channel?  OK, if it just works, it just works :).

corrados: Not really magic :-). The server has a channel for each connected client. The client just only "talks" to its own channel at the server.

pljones: Yeah, just took some mental context switching for "connected" channel over UDP to make sense...  It's "magic" in that it's hidden in the protocol layer implementation and the upper layer (message handling) doesn't need to worry.

corrados:
I just had a chance to test your new code on centest with 6 other musicians. First: it looks great and is really a benefit to see who is playing and who is not. Even the update rate is great. You did a really good job!
Unfortunately, there is still an issue in your code (which may be a bit hard to debug). I entered the server after 4 others were already connected. When then one of the clients which were already connected left the server, suddenly my own LED bar did not work anymore. It seems the assignment of levels did get confused when the number of clients change.

pljones:
Yes, that's what I was worried about.  As I'm rewriting around the client / channel code, I'll try to see where it's going wrong.  I _think_ it would break even with just two clients, where the client on channel 0 leaves.  I can test that easily enough locally.

(Note the next version will again rewrite the protocol, of course, so once it's ready, I'll close this pull request and rewrite history again...)

pljones:
OK, I seem to have fixed it.  I'll also do the manual updates before the next PR (and rebase against the latest master).

PR #42, opened 2020-04-05
Feature/channel meters

pljones:
OK, so next attempt :)

The proposed manual changes are here:
https://github.com/pljones/jamulus/blob/feature/channel-meters/src/res/homepage/manual.md


corrados:
I have just updated the centest server with your newest code. I'll test it later when more people are online. Feel free to test on this server, too, if you like.

src/global.h
```diff
@@ -195,6 +199,9 @@ LED bar:      lbr
 // list
 #define PING_UPDATE_TIME_SERVER_LIST_MS 2000 // ms
 
+// defines the interval between Channel Level updates from the server
+#define CHANNEL_LEVEL_UPDATE_INTERVAL   100  // number of frames
```
corrados, line 203: The number is not given in ms but in frames. So it is dependent on the frame size and sample rate, right?

src/protocol.cpp
```diff
@@ -1924,6 +1995,65 @@ bool CProtocol::EvaluateCLReqConnClientsListMes ( const CHostAddress& InetAddr )
...
+    CreateAndImmSendConLessMessage ( PROTMESSID_CLM_CHANNEL_LEVEL_LIST,
+                                     vecData,
+                                     InetAddr );
+}
+
+bool CProtocol::EvaluateCLChannelLevelListMes  ( const CHostAddress&     InetAddr,
+                                                 const CVector<uint8_t>& vecData )
+{
+    int       iPos     = 0; // init position pointer
+    const int iDataLen = vecData.Size();  // four bits per channel, 2 channels per byte
```
corrados, line 2026: You could introduce a plausability check here. The maximum possible number of channels at the server is given in the global.h file. You could check that the length of the data vector does not exceed that number, i.e. the number of bits times the maximum possible number of channels.

corrados:
Hmm, too bad. I just checked in some modifications in the protocol which made your commit to have merge conflicts.
I tested your new version and it seems to run just fine. Since we already had some iterations with the pull requests, I would recommend to merge your code to the master now and do fixes from there. Can you please re-base your pull request once more so that I can merge your code without merge conflicts? Thank you.

src/global.h
pljones: Yes -- will your work affect this?  If the frame frequency drops, I guess the channel level update interval should also drop?

src/protocol.cpp
pljones: Yep, I'll add that (just before rebasing).

src/global.h
corrados: I am currently working on supporting different frame sizes in the server. So I have to make sure that your update time stays the same no matter which frame size is in use by the server right now. No big deal, I just have to remember it that there is something to do :-).

pljones:
OK rebased with check for MAX_NUM_CHANNELS added - is it okay?

Testing: it compiles clean still :).

corrados:
Great, thank you for your fast response. The code looks good to me. I'll merge it now. Again, thank you very much for this contribution. In these times with croweded servers it is a really useful tool. Unfortunately, all the existing servers have to update to a new version so that it works for them, too.

pljones:
Thanks for the comments.  I've updated jamulus.drealm.info now.

Does the central server know the version of a registering server?  Be interesting to have stats on how fast each version usually gets rolled out to server.

corrados:
Yes, this is already implemented. Just define the following and re-compile: https://github.com/corrados/jamulus/blob/master/src/global.h#L78. Then open the Server List.

PR #44, opened 2020-04-06
Multichannel CoreAudio Support

emlynmac:
The CoreAudio layer to figure out what inputs are available only worked on multiple channels per stream devices.

This PR adds support for the case where _input_ devices have multiple discrete buffers and have need to pull the data out of those buffers properly.

### DIFF
--- a/mac/sound.cpp
+++ b/mac/sound.cpp
@@ -23,7 +23,7 @@
 \******************************************************************************/
 
 #include "sound.h"
-
+#include <QDebug>
 
 /* Implementation *************************************************************/
 CSound::CSound ( void       (*fpNewProcessCallback) ( CVector<short>& psData, void* arg ),
@@ -100,7 +100,7 @@ CSound::CSound ( void       (*fpNewProcessCallback) ( CVector<short>& psData, vo
         throw CGenErr ( tr ( "CoreAudio input AudioHardwareGetProperty call failed. "
                              "It seems that no sound card is available in the system." ) );
     }
-
+    
     iPropertySize               = sizeof ( AudioDeviceID );
     stPropertyAddress.mSelector = kAudioHardwarePropertyDefaultOutputDevice;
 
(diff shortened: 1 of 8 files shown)


corrados:
Thanks for your code. I cannot merge your changes directly but will integrate the important changes of the code in the master code of Jamulus as soon as I have a time spot available (very busy right now...).

corrados:
I added the essential changes from your branch to the Git master. Can you please checkout that version and verify that this code works for you?

emlynmac:
Code changes look good visually - will build it and check this evening.
Cheers!

emlynmac:
Confirmed - working with my Phonic Helix FW

corrados:
Great, thank you. I will close this pull request now.

corrados:
Do you have an idea what could go wrong here: https://sourceforge.net/p/llcon/discussion/software/thread/7c443f2ed9/#eeab
It seems your new code introduced the issue for this user.

emlynmac:
Are you still using the same QT / macOS SDK?

I’ve updated to the latest and saw that the CoreAudio callback registration fails _sometimes_. I have no idea yet what is causing it.
Maybe back out the changes I made and see if that resolves the issue for them?

I’ll be digging into why I sometimes get the CoreAudio registration failure sometimes.

corrados:
Yes, the same QT and SDK.
As a quick test I will change to order of the if conditions in callbackIO to see if that helps.
Thanks for your investigations.

emlynmac:
OK perfect.

I also see lots of commits on master :)

Does this mean that 64 samples frame is now in place there?

I’m planning to investigate from master.

And this is Volker, yes?

I’m Emlyn - in Vancouver, BC.

Cheers,

corrados:
Yes, this is Volker and yes, the 64 samples version should now work. I just tried it out with my band mates and it works great.

emlynmac:
Hi,

I’ve had a dig through and I’m consistently seeing the out callback denied by CoreAudio.

I tried separating out the callbacks from one function, which got rid of the error but didn’t get it functional. The error I was seeing was kAudioHardwareIllegalOperationError. Now, I see a different error: 
HALC_ShellDriverPlugIn::Open: Can't get a pointer to the Open routine.


I also found a path I missed - dealing with output with discrete channels. I’ve put a fix for this up at https://github.com/emlynmac/jamulus/tree/fix-out-multichannel

I’m wondering if there’s something happening with core audio not being entirely happy with how it’s being used.

I found some sample code from apple:




This is grabbing audio and then playing out, but it’s using AudioUnits in addition to the AUHal.

I’m wondering if this might be the way to go to make it work.

If I get time, I can have a look at making this work, but got 3 other projects on the go at the moment and a small child :)

E

corrados:
Wow, that took me quite long to figure out what goes wrong...
Can you please test this version: https://sourceforge.net/p/llcon/discussion/software/thread/7c443f2ed9/#b038/04be
if it still runs fine with your sound card?

emlynmac:
I will give it a shot. Is this from the master branch ?

corrados:
Yes, this is from master branch.

emlynmac:
Ah man!

THANK YOU!

I’m glad you found the issue!

This is working great for me, although very strangely, I have to build it with the QT 5.12 toolchain; if I use Xcode, it does not work.

How can I buy you a beer?

Emlyn

corrados:
How about a virtual one: 🍺
:-)

emlynmac:
Cheers! 

🍻

corrados:
I just changed the Mac audio interface to support Aggregated Devices. I had to change a lot of code. Could you try out the latest code on Git master if it still works with your sound card?

PR #48, opened 2020-04-09
Use server internal address in server list

pljones:
I changed a few casts that were giving warnings about incompatible types, as well (in the areas I was changing anyway).

Also added a random IP generator - I didn't work out how the test bench works but it compiles :).

Tests:

New to New
- local "central" server IP-1, port A
- local "slave" server IP-1, port B, registers with "central"
- local client IP-1, port C, requests server list from "central", sees "central" and "slave"
- local client IP-2, port C, requests server list from "central", sees "central" and "slave"

New to Old
- local client IP-1, port C, requests server list from default, gets normal list

Didn't do (could have done for _one_ slave to a central, I think)
- New server to Old central
- Old server to New central
- Old client to New server

### DIFF
--- a/src/protocol.cpp
+++ b/src/protocol.cpp
@@ -203,26 +203,28 @@ CONNECTION LESS MESSAGES
         ... -----------------+----------------------------------+ ...
         ...  2 bytes country | 1 byte maximum connected clients | ...
         ... -----------------+----------------------------------+ ...
-        ... ---------------------+------------------+ ...
-        ...  1 byte is permanent | 2 bytes number n | ...
-        ... ---------------------+------------------+ ...
-        ... ----------------------------------+ ...
-        ...  n bytes UTF-8 string server name | ...
-        ... ----------------------------------+ ...
-        ... ------------------+----------------------------+ ...
-        ...  2 bytes number n | n bytes UTF-8 string topic | ...
-        ... ------------------+----------------------------+ ...
+        ... ---------------------+ ...
+        ...  1 byte is permanent | ...
+        ... ---------------------+ ...
+        ... ------------------+----------------------------------+ ...
+        ...  2 bytes number n | n bytes UTF-8 string server name | ...
+        ... ------------------+----------------------------------+ ...
+        ... ------------------+---------------------------------------------+ ...
+        ...  2 bytes number n | n bytes UTF-8 string server interal address | ...
+        ... ------------------+---------------------------------------------+ ...
         ... ------------------+---------------------------+
         ...  2 bytes number n | n bytes UTF-8 string city |
         ... ------------------+---------------------------+
 
+    - "country" is according to "Common Locale Data Repository" which is used in
+      the QLocale class
     - "maximum connected clients" is the maximum number of clients which can
       be connected to the server at the same time
     - "is permanent" is a flag which indicates if the server is permanent
       online or not. If this value is any value <> 0 indicates that the server
       is permanent online.
-    - "country" is according to "Common Locale Data Repository" which is used in
-      the QLocale class
+    - "server interal address" represents the IPv4 address as a dotted quad to
+      be used by clients with the same external IP address as the server.
 
 
 - PROTMESSID_CLM_UNREGISTER_SERVER: Unregister a server
@@ -1401,14 +1403,15 @@ bool CProtocol::EvaluateCLServerFullMes()
 }
 
 void CProtocol::CreateCLRegisterServerMes ( const CHostAddress&    InetAddr,
+                                            const CHostAddress&    LInetAddr,
                                             const CServerCoreInfo& ServerInfo )
 {
     int iPos = 0; // init position pointer
 
     // convert server info strings to utf-8
-    const QByteArray strUTF8Name  = ServerInfo.strName.toUtf8();
-    const QByteArray strUTF8Topic = ServerInfo.strTopic.toUtf8();
-    const QByteArray strUTF8City  = ServerInfo.strCity.toUtf8();
+    const QByteArray strUTF8LInetAddr = LInetAddr.InetAddr.toString().toUtf8();
+    const QByteArray strUTF8Name      = ServerInfo.strName.toUtf8();
+    const QByteArray strUTF8City      = ServerInfo.strCity.toUtf8();
 
     // size of current message body
     const int iEntrLen =
@@ -1417,7 +1420,7 @@ void CProtocol::CreateCLRegisterServerMes ( const CHostAddress&    InetAddr,
         1 /* maximum number of connected clients */ +
         1 /* is permanent flag */ +
         2 /* name utf-8 string size */ + strUTF8Name.size() +
-        2 /* topic utf-8 string size */ + strUTF8Topic.size() +
+        2 /* server internal address utf-8 string size */ + strUTF8LInetAddr.size() +
         2 /* city utf-8 string size */ + strUTF8City.size();
 
     // build data vector
@@ -1442,8 +1445,8 @@ void CProtocol::CreateCLRegisterServerMes ( const CHostAddress&    InetAddr,
     // name
     PutStringUTF8OnStream ( vecData, iPos, strUTF8Name );
 
-    // topic
-    PutStringUTF8OnStream ( vecData, iPos, strUTF8Topic );
+    // server internal address (formerly unused topic)
+    PutStringUTF8OnStream ( vecData, iPos, strUTF8LInetAddr );
 
     // city
     PutStringUTF8OnStream ( vecData, iPos, strUTF8City );
@@ -1458,6 +1461,8 @@ bool CProtocol::EvaluateCLRegisterServerMes ( const CHostAddress&     InetAddr,
 {
     int             iPos     = 0; // init position pointer
     const int       iDataLen = vecData.Size();
+    QString         sLocHost; // temp string for server internal address
+    CHostAddress    LInetAddr;
     CServerCoreInfo RecServerInfo;
 
     // check size (the first 6 bytes)
@@ -1468,7 +1473,7 @@ bool CProtocol::EvaluateCLRegisterServerMes ( const CHostAddress&     InetAddr,
 
     // port number (2 bytes)
     RecServerInfo.iLocalPortNumber =
-        static_cast<int> ( GetValFromStream ( vecData, iPos, 2 ) );
+        static_cast<uint16_t> ( GetValFromStream ( vecData, iPos, 2 ) );
 
     // country (2 bytes)
     RecServerInfo.eCountry =
@@ -1491,11 +1496,21 @@ bool CProtocol::EvaluateCLRegisterServerMes ( const CHostAddress&     InetAddr,
         return true; // return error code
     }
 
-    // server topic
+    // server internal address
     if ( GetStringFromStream ( vecData,
                                iPos,
-                               MAX_LEN_SERVER_TOPIC,
-                               RecServerInfo.strTopic ) )
+                               MAX_LEN_IP_ADDRESS,
+                               sLocHost ) )
+    {
+        return true; // return error code
+    }
+
+    if ( sLocHost.isEmpty() )
+    {
+        // old server, empty "topic", register as external address
+        LInetAddr = InetAddr;
+    }
+    else if ( !LInetAddr.InetAddr.setAddress( sLocHost ) )
     {
         return true; // return error code
     }
@@ -1516,7 +1531,7 @@ bool CProtocol::EvaluateCLRegisterServerMes ( const CHostAddress&     InetAddr,
     }
 
     // invoke message action
-    emit CLRegisterServerReceived ( InetAddr, RecServerInfo );
+    emit CLRegisterServerReceived ( InetAddr, LInetAddr, RecServerInfo );
 
     return false; // no error
 }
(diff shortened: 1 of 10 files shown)


src/protocol.cpp
```diff
@@ -1618,11 +1633,11 @@ bool CProtocol::EvaluateCLServerListMes ( const CHostAddress&     InetAddr,
 
         // IP address (4 bytes)
         const quint32 iIpAddr =
-            static_cast<int> ( GetValFromStream ( vecData, iPos, 4 ) );
+            static_cast<uint32_t> ( GetValFromStream ( vecData, iPos, 4 ) );
```
corrados, line 1636: Why not cast to "quint32"?

corrados: Same question applies to other places in the code, too.

src/serverlist.cpp
```diff
@@ -321,13 +328,17 @@ void CServerListManager::CentralServerRegisterServer ( const CHostAddress&    In
...
+            // Check if server is already registered.
+            // Use external address and local port to identify a server
+            // (there could be more than one server on the external address
+            // but they would have to use separate ports).
+            // The very first list entry must not be checked since
             // this is per definition the central server (i.e., this server)
             int iSelIdx = ciInvalidIdx; // initialize with an illegal value
             for ( int iIdx = 1; iIdx < iCurServerListSize; iIdx++ )
             {
-                if ( ServerList[iIdx].HostAddr == InetAddr )
```
corrados, line 330: The operator== checks that IP number and port are the same. So I see no need to check for equal local port number. Even if we have port translation, each local port of each server gets translated to a different port number so we are fine. So I would remove the "&& ServerList[iIdx].iLocalPortNumber == ServerInfo.iLocalPortNumber". Do you agree?

src/serverlist.cpp
```diff
@@ -415,20 +425,19 @@ void CServerListManager::CentralServerQueryServerList ( const CHostAddress& Inet
             {
                 // check if the address of the client which is requesting the
                 // list is the same address as one server in the list -> in this
-                // case he has to connect to the local host address
+                // case he has to connect to the local host address and port
+                // to allow for NAT.
                 if ( vecServerInfo[iIdx].HostAddr.InetAddr == InetAddr.InetAddr )
                 {
-                    vecServerInfo[iIdx].HostAddr.InetAddr =
-                        QHostAddress ( QHostAddress::LocalHost );
```
corrados, line 422: Ok, so no "LocalHost" required anymore since if we are running on the same PC, the local IP address should also work, right?

src/util.cpp
```diff
@@ -890,6 +890,20 @@ bool NetworkUtil::ParseNetworkAddress ( QString       strAddress,
 }
 
 
+CHostAddress NetworkUtil::GetLocalAddress()
+{
+    QTcpSocket socket;
+    socket.connectToHost("8.8.8.8", 53); // google DNS, or something else reliable
```
corrados, line 896: This might be critical. The function is a blocking function. If, e.g., the laptop does not have internet access but the server shall be used for local network only, the Jamulus server would not start up.

src/util.cpp
```diff
@@ -890,6 +890,20 @@ bool NetworkUtil::ParseNetworkAddress ( QString       strAddress,
...
+    if (socket.waitForConnected()) {
```
corrados, line 897: In the Qt docu I can read: "QAbstractSocket::waitForConnected(int msecs = 30000)". So we have a default time-out of 30 seconds? If that is the case, this is too long. A connection should not take longer than a few ms. As a Jamulus server user I would expect that at least after 3 seconds the GUI appears, otherwise I would think it is crashed and would kill the process. So I would set the time-out to 3 seconds. Do you agree? Can you please test this time-out that it works correctly by disconncting your laptop/PC from the internet and start the Jamulus server?

src/util.cpp
corrados: You should not hard-code the IP/port here. Please use a #define and put it in global.h where, e.g., the jamulus.fischvolk.de address is also given.

src/util.cpp
```diff
@@ -890,6 +890,20 @@ bool NetworkUtil::ParseNetworkAddress ( QString       strAddress,
...
+        return CHostAddress( socket.localAddress(), 0 );
+    } else {
+        qWarning()
```
corrados, line 900: Where does this goes? If you are on Linux, do you see it on the command line? Is it needed also for the Windows user to get this warning? But the problem is that we have an auto-start of the Jamulus server on Windows. If we would get a warning message dialog, this might interfere with the auto start. So I guess it is better to silently set it to localhost. I assume in 99.9999 % of the cases we can successfully get the local address.

src/protocol.cpp
pljones: I'll go back over the ones I changed to check what the variable is used for and make sure the declaration and cast both use that target data type.

src/serverlist.cpp
pljones: Ah, right, of course.  I think of "addr" only in terms of "IP address" not "IP address and port number".

src/serverlist.cpp
pljones: It should, yes.  Even I can't think of when a server would not route its own IP to itself.

src/util.cpp
pljones: Ah right, OK.

src/util.cpp
pljones: Ugh, forgot about the server UI - yes, of course.  It'll be easier to test it putting a non-existent IP in.

src/util.cpp
pljones: If you're starting up the server, it's logged along with the usual start up messages.

And yes, it's highly unlikely to fail and the fallback to localhost is at least safe.

The only other thing I could think of was trying the given central server on port 80 - but if you're running your own central server, it might not have port 80 open.  So there's not really much improvement.

I hope that socket disconnects when it goes out of scope - I forgot to check that.

src/protocol.cpp
pljones: Done
91b9c04 Resolve casts to target types

src/serverlist.cpp
pljones: Done
d4636c5 CHostAddr operator== does the job

src/util.cpp
pljones: Done and tested.  I went with half a second - the target IP used really should respond that fast if the machine doing the look up has adequate internet connectivity.  Otherwise localhost is probably for the best.
02632ea Tune IP look up code

pljones:
Added test: Old client, new central server, new server - worked okay.  Again, all local.

corrados:
Ok, it looks good now. In the meantime I have changed some code on the master. The pull request still says "the branch has no conflicts". You have more experience with Github: What is the best practice? Shall I simply merge or do you recommend to re-base your pull request first?

pljones:
Rebased, tested again to make sure.  Looks okay.

corrados:
Thank you very much for your code. This issue was really confusing a lot of Jamulus users. I hope these issues are gone now. I will test the new implementation the next days on the central server.

corrados:
I just updated the Central Server with your newest code. Keep the fingers crossed that it runs table ;-). Please do some testing on that server if your implementation works fine. I'll also do this here in my local network.

corrados:
I just tested it. It works great! :-)

pljones:
Yep, tried from both old and new clients to new (default) central and new jamulus.drealm.info and the new local server shows up and connects perfectly.

corrados:
Hi Peter, can you please take a look at: https://sourceforge.net/p/llcon/discussion/server/thread/38777b088a/?limit=100#bf93
I think this behaviour is new and is related to your changes. Do you have an idea how to get back to the original behavior with the displayed name?

pljones:
Strange.  I was seeing my serverinfo name during testing - clicking on it got that name selected in the connect box, connecting used the internal IP address.

pljones:
Yes, strange.
```
Jamulus  28130     1  0 13:25 ?        00:00:00 /opt/Jamulus/bin/Jamulus -s -n -F -p 54850 -u 16 -e jamulus.fischvolk.de -m /opt/Jamulus/run/status.html -l /opt/Jamulus/log/Jamulus.log -y /opt/Jamulus/log/history.svg -L -R /opt/Jamulus/run/recording -a drealm Jamulus server 1 -o drealm Jamulus server 1;London;224 -w Welcome to drealm Jamulus server 1 - (recording enabled: CC BY-NC-SA 4.0 - no cover versions!) *NOTE* Client 3.4.2 or later required here!
```
I restarted my server, registering with different details.  "drealm Jamulus server 1" does not appear.  What _still_ appears is my previous server name, `jamulus.drealm.info`.  The IP and port wouldn't have changed, so I'd have expected the CServerList to have been updated...

BUT!!

We never try.  `iCurServerListSize < MAX_NUM_SERVERS_IN_SERVER_LIST` prevents _any_ processing of registration messages, even from existing servers.

Hm.

corrados:
Yes it does. A lot of strange things can happen if the server list is full. We simply have to avoid a full server list. I hope my last commit will help.

pljones:
I think moving the `if` is definitely a good idea, though - an existing server should always be able to update registration details.

corrados:
Agreed.

corrados:
Unfortunately your code does not work as expected. A lot of people complained in the forums and now I tried it for myself and could reproduce it. I now reverted the Central Server binary with an old one.
When I started the client and server on the same PC, the LHostAddr unfortunately did not contain my local IP number but the public IP number of my router. So the "socket.localAddress()" seems not to output the correct IP address on my PC :-(.
Do you have an idea how to solve the issue?

corrados:
It seems to work on Linux but does not work on Windows. I'll do some more testing now...

corrados:
Wait, I was using an old server on Windows. Let me do some more investigations... I'll report back here.

pljones:
The central server will report the external address for any "old" version of the server, rather than local host.  That's possibly something that should be corrected, indeed!

corrados:
We should substitute "localhost" in that case. That would be consistent with the previous setting.

pljones:
```
diff --git a/src/protocol.cpp b/src/protocol.cpp
index c64addf..e645048 100755
--- a/src/protocol.cpp
+++ b/src/protocol.cpp
@@ -1508,7 +1508,7 @@ bool CProtocol::EvaluateCLRegisterServerMes ( const CHostAddress&     InetAddr,
     if ( sLocHost.isEmpty() )
     {
         // old server, empty "topic", register as external address
-        LInetAddr = InetAddr;
+        LInetAddr = CHostAddress( QHostAddress( QHostAddress::LocalHost ), 0 );
     }
     else if ( !LInetAddr.InetAddr.setAddress( sLocHost ) )
     {
```

corrados:
Ok, I think the issue is in EvaluateCLServerListMes with:
            // old central server, empty "topic", default to server address
            LInetAddr.InetAddr.setAddress( iIpAddr );
Do you agree? Shall I fix it?

pljones:
No, I think it's the patch above.

corrados:
Thanks, I'll apply it. What does the substitution in EvaluateCLServerListMes() does? Why not set to localhost there?

pljones:
That would be needed for a message returned from an old central server to a new client.  I suppose it would do no harm to make the same change there.

pljones:
```
diff --git a/src/protocol.cpp b/src/protocol.cpp
index c64addf..da71258 100755
@@ -1674,8 +1674,8 @@ bool CProtocol::EvaluateCLServerListMes ( const CHostAddress&     InetAddr,
         CHostAddress LInetAddr;
         if ( strLHostAddr.isEmpty() )
         {
-            // old central server, empty "topic", default to server address
-            LInetAddr.InetAddr.setAddress( iIpAddr );
+            // old central server, empty "topic", default to localhost
+            LInetAddr.InetAddr.setAddress( QHostAddress::LocalHost );
         }
         else if ( !LInetAddr.InetAddr.setAddress( strLHostAddr ) )
         {
```

corrados:
Ok, I agree. I'll add both and start again testing. Thanks.

corrados:
Ok, now it works :-).

pljones:
OK, now it doesn't work :(

corrados:
Ok, get back with your test results here. With my tests (old/new versions, Windows/Linux) all worked as expected.

pljones:
Locally - with a local central server - it works with or without the patches for both Linux and Windows _clients_.  I'll try a Windows server (another hole in the firewall...).

pljones:
Just to be clear (as posted elsewhere).  Expected results are as follows:

Old client + Old server (with same external IP as client) + Old central server -> old behaviour - clients on the server _should_ see it (I don't think they do, not sure why); clients on other machines won't, as they can't get a ping response from themselves.  (`--showallservers` just shows the full list from the central server, so not really relevant here.)

New client + Old server + Old central server -> `EvaluateCLServerListMes` returning "localhost" for missing value gives the same as above; returning external IP may or may not help clients not on the server itself.

Any client + New server + Old central server -> Same as above two - the central server needs to send the address or it can't be used.

Any client + Old server + New central server -> `EvaluateCLRegisterServerMes` returning "localhost" for missing value gives the same as above; again, external IP _might_ help but probably not.

Any client + New server + New central server -> This should give the correct behaviour.

All of this, it must be remembered, is on the basis that the new server gets the right internal IP address.  That _should_ be the case in a domestic setting where they're running clients and servers on the same LAN.

corrados:
Thanks for the summary. I totally agree to you.
I made a release 3.4.7 and I also replaced the Central Server software by the newest one. I know that it is a bit risky but the 200 server limit seems to be such a big issue for the Jamulus users that I wanted the new version out as soon as possible so that the North America Central Server will be filled up with server as soon as possible. Lets see what the user feedback will be...

corrados:
I just noticed that you only use the IP address of LHostAddr and do not use the port number. Right now it is a CHostAddress type but if it is true that you only need the IP address, you should use QHostAddress instead. Do you agree?

pljones:
My understanding was that was that was being done already (EvaluateCLServerListMes).

pljones:
Oh - to avoid confusion, perhaps that's for the best, yes.

pljones:
PR raised.
https://github.com/corrados/jamulus/pull/61

Issue #35, opened 2020-04-01
Core dump with -R if directory exists


gilgongo:
Jamulus 3.4.4 on Ubuntu 18.04.4 LTS amd64 with libqt4-dev 4:4.8.7+dfsg-7ubuntu1 

**Steps to re-create**:

1. Stop Jamulus server and create dir with jamulus write permissions:

```
mkdir /home/gilgongo/JamRecordings
chown jamulus:jamulus /home/gilgongo/JamRecordings
```

2. Start server with UID jamulus GID jamulus with `-R /home/gilgongo/JamRecordings`

3. Jamulus core dumps and shuts down with error:

```
Mar 31 22:31:44 covid19 jamulus[12427]: - recording directory name: /home/gilgongo/JamRecordings/
Mar 31 22:31:44 covid19 jamulus[12427]: terminate called after throwing an instance of 'std::runtime_error'
Mar 31 22:31:44 covid19 jamulus[12427]:   what():  /home/gilgongo/JamRecordings does not exist but could not be created
Mar 31 22:31:45 covid19 systemd[1]: jamulus.service: Main process exited, code=dumped, status=6/ABRT
Mar 31 22:31:45 covid19 systemd[1]: jamulus.service: Failed with result 'core-dump'.
```

**Expected behaviour**: either keep server running but ignore -R and show warning (if Jamulus is expecting to create `JamRecordings` dir itself), or just allow Jamulus to write to the directory.

Note that letting the server create the directory works (and works across subsequent server reboots) but obviously required jamulus xrw on the parent dir, which may not be a good idea.

pljones:
Does the user Jamulus is running as have permission that location?

gilgongo:
The JamRecordings directory was owned by Jamulus, so I assume the error  "does **not** exist but could not be created" would seem incorrect, no?

But in any case it works if (for example) I give it a path of /tmp/somedir, but do not create somedir before starting the server.

So I guess the expected behaviour here (assuming Jamulus wants to create the directory itself) would be for the message to say something like "_/home/gilgongo/JamRecordings already exists, ignoring -R_" and continue working without a core dump. 

Ideally though, if the directory already exits and the permissions are OK (eg owned by Jamulus?), then Jamulus should just use it.

pljones:
The path needs to be accessible.

If `/home` or `/home/gilgongo` is not accessible (does not have read and execute access), then the content, no matter whether they exist or not, is not accessible.

gilgongo:
[...]

Issue #30, opened 2020-03-30
 Jamulus crash when history file is given


gilgongo:
_This issue has been migrated from Sourceforge. Created 28-03-2020_

The version 3.4.4 seems to crash when history file is provided:

```
Jamulus -n -s -p 22124 -l /jam/jamulus.log -y /jam/history.jpg
- no GUI mode chosen
- server mode chosen
- selected port number: 22124
- logging file name: /jam/jamulus.log
- history file name: /jam/history.jpg
- welcome message: Wellcome to Jamulus docker server.
Qt5 requires a windowing system to paint a JPEG image; image will use SVG
AHistoryGraph::Start ( /jam/history.jpg )
```

This behaviour is reproduced on Docker image: probably it missing some dependencies.

> Volker 29-03-2020: pljones has changed the history graph recently so that it can use SVG, see https://github.com/corrados/jamulus/pull/6. I hope he can give you support on this.

> Volker 29-03-2020: SVG uses QXmlStreamWriter. Does this need a special library? Do you know this?
Do you have an error call stack available to see where it crashes?

pljones:
Basically,
> Qt5 requires a windowing system to paint a JPEG image
so, if you use `-n`, it's assumed that you don't _want_ to use a windowing system, so you do not have the option to use a JPEG image.  The filename you give has SVG data written to it.

If you want to view it, use a `.svg` as the suffix - so either change your start up command or move the file afterwards.

corrados:
This is obviously a configuration issue and can be solved by using correct command line options. I will close this Issue now.

PR #51, opened 2020-04-10
Multiple frame sizes in recorder

pljones:
OK, after going off in the wrong direction, here's my second attempt :).

Nice and simple.  Tell the recorder the frame size when it's decided.

As far as I can tell, it's handling server frames, so it's fine to do that just the once.  (I'd started thinking it was handling client frames -- but they've already been pre-processed.)

I've tested with an old client and this new version run with and without -F and the recording worked okay.

### DIFF
--- a/src/recorder/creaperproject.cpp
+++ b/src/recorder/creaperproject.cpp
@@ -55,7 +55,7 @@ using namespace recorder;
  * @param trackItem the details of where the item is in the track, along with the RIFF WAVE filename
  * @param iid the sequential item id
  */
-CReaperItem::CReaperItem(const QString& name, const STrackItem& trackItem, const qint32& iid)
+CReaperItem::CReaperItem(const QString& name, const STrackItem& trackItem, const qint32& iid, int frameSize)
 {
     QString wavName = trackItem.fileName; // assume RPP in same location...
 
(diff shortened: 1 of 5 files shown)


corrados:
Great, you are extremely quick.
Since it is all your code I'll merge it without a code review.
What about these: https://github.com/corrados/jamulus/issues/30 and https://github.com/corrados/jamulus/issues/35? Are these real issues to be addressed?

pljones:
#30 - yes it core dumps.  The choice between JPEG (which will fail with `-n`) and SVG is made based on filename (`CServerLogging::EnableHistory`).  Bit of a hack... :)

#35 - I'm almost certain it's a set up problem.  If you run the server twice with -R, first time it'll create the directory if it's not there, second time it'll happily not create the directory.  If it's can't "see" the directory, due to permissions issues, it's likely not to be able to create it either.

corrados:
OK, I saw your comments in the Github Issues. Thanks for taking care.
There is another issue reported in the forum which relates to your code: https://sourceforge.net/p/llcon/discussion/server/thread/2d5a42c1fe/. I assume that shortly there will also be an issue created for it at Github.

sehe:
@pljones Thanks for this work, it improved a few things I had already drafted up (e.g. CurOpusDecoder and the many different calls to `opus_custom_decode`).

I have one question about the -F server option (to put the server in 64 byte frame mode). When is this feature handy? I'm assuming it will be 100% compatible with 128b-frame clients? So, am I right in thinking that this merely helps tune the server-side latency in more precise steps by selecting a more fine-grained jitter buffer size?

corrados:
Yes, correct.

PR #61, opened 2020-04-12
Tidy away local port in CServerCoreInfo etc

pljones:
This clears up "iLocalPort".  Tested locally.  The protocol layer isn't much changed - changed the comment regarding the port to make it clear what it's holding.

Also allowed a full central server to update registered existing servers.


corrados:
Thank you!

PR #74, opened 2020-04-15
Allow central server to confirm registration

pljones:
I'll note comments from the earlier review.


src/protocol.cpp
```diff
@@ -304,13 +305,31 @@ CONNECTION LESS MESSAGES
...
 
+- PROTMESSID_CLM_REGISTER_SERVER_RESP: Version number and operating system
+
+    +---------------+
+    | 1 byte status |
+    +---------------+
+
+    - "status":
+      0 - success
+      1 - failed due to central server list being full
```
pljones, line 323: Reference the enum.

src/protocol.cpp
```diff
@@ -2090,6 +2113,40 @@ bool CProtocol::EvaluateCLChannelLevelListMes  ( const CHostAddress&     InetAdd
     return false; // no error
 }
 
+void CProtocol::CreateCLRegisterServerResp  ( const CHostAddress& InetAddr,
+                                              const int           iStatus )
```
pljones, line 2117: Use the enum but validate that the value is in range as a byte (0 to 255).

src/protocol.cpp
```diff
@@ -2090,6 +2113,40 @@ bool CProtocol::EvaluateCLChannelLevelListMes  ( const CHostAddress&     InetAdd
...
+    const int iDataLen = vecData.Size();
+
+    if ( iDataLen != 1 )
+    {
+        return true;
+    }
+
+    // known values are 0 and 1 currently
+    // but allow 0 to 255
+    int iStatus = static_cast<int> ( GetValFromStream ( vecData, iPos, 1 ) );
```
pljones, line 2143: Retrieve as int, validate that the value is positive and in range for the enum, then cast.

src/protocol.cpp
```diff
@@ -2090,6 +2113,40 @@ bool CProtocol::EvaluateCLChannelLevelListMes  ( const CHostAddress&     InetAdd
...
+
+    // invoke message action
+    emit CLRegisterServerResp ( InetAddr, iStatus );
```
pljones, line 2146: Emit the enum.

src/protocol.h
```diff
@@ -130,6 +131,8 @@ class CProtocol : public QObject
     void CreateCLChannelLevelListMes   ( const CHostAddress&      InetAddr,
                                          const CVector<uint16_t>& vecLevelList,
                                          const int                iNumClients );
+    void CreateCLRegisterServerResp    ( const CHostAddress&      InetAddr,
+                                         int                      iStatus );
```
pljones, line 135: enum

src/protocol.h
```diff
@@ -305,4 +310,6 @@ public slots:
     void CLReqConnClientsList         ( CHostAddress           InetAddr );
     void CLChannelLevelListReceived   ( CHostAddress           InetAddr,
                                         CVector<uint16_t>      vecLevelList );
+    void CLRegisterServerResp         ( CHostAddress           InetAddr,
+                                        int                    iStatus );
```
pljones, line 314: enum

src/server.cpp
```diff
@@ -438,6 +438,10 @@ CServer::CServer ( const int          iNewMaxNumChan,
         SIGNAL ( CLReqServerList ( CHostAddress ) ),
         this, SLOT ( OnCLReqServerList ( CHostAddress ) ) );
 
+    QObject::connect ( &ConnLessProtocol,
+        SIGNAL ( CLRegisterServerResp ( CHostAddress, int ) ),
+        this, SLOT ( OnCLRegisterServerResp ( CHostAddress, int ) ) );
+
```
pljones, line 444: enum

src/server.h
```diff
@@ -373,6 +375,14 @@ public slots:
         ServerListManager.CentralServerRegisterServer ( InetAddr, LInetAddr, ServerInfo );
     }
 
+    void OnCLRegisterServerResp ( CHostAddress /* unused */,
+                                  int          iStatus )
+    {
+        ServerListManager.SetSvrRegStatus ( static_cast<ESvrRegStatus> ( iStatus ) );
```
pljones, line 381: enum (so no need to cast here).

src/serverdlg.cpp
```diff
@@ -534,6 +544,31 @@ void CServerDlg::UpdateGUIDependencies()
     // server list is enabled and not the default address is used
     edtCentralServerAddress->setEnabled (
         !bCurUseDefCentServAddr && bCurSerListEnabled );
+
+    QString strStatus;
```
pljones, line 548: Open to views on these but there needs to be a reasonable gap between the checkbox label and the status label, so the text should not be too long.

src/serverlist.cpp
```diff
@@ -297,7 +311,7 @@ void CServerListManager::OnTimerPollList()
     // server entry) and the predefined servers if they are still valid.
     // Note that we have to use "ServerList.size()" function in the for loop
     // since we may remove elements from the server list inside the for loop.
-    for ( int iIdx = 1 + iNumPredefinedServers; iIdx < ServerList.size(); iIdx++ )
+    for ( int iIdx = 1 + iNumPredefinedServers; iIdx < ServerList.size(); )
```
pljones, line 314: Extra fix...

src/serverlist.cpp
```diff
@@ -488,14 +537,25 @@ void CServerListManager::SlaveServerRegisterServer ( const bool bIsRegister )
...
         else
         {
             // unregister server
+            eSvrRegStatus = SRS_UNREGISTERED;
+            emit SvrRegStatusChanged();
+
             pConnLessProtocol->CreateCLUnregisterServerMes ( SlaveCurCentServerHostAddress );
         }
     }
+    else if ( bIsRegister )
```
pljones, line 556: May be able to remove the `if` now.  (Was having strange problems at one point...)

src/serverlist.h
```diff
@@ -171,13 +171,22 @@ class CServerListManager : public QObject
 
     QLocale::Country GetServerCountry() { return ServerList[0].eCountry; }
 
+    ESvrRegStatus GetSvrRegStatus() { return eSvrRegStatus; }
+
+    void SetSvrRegStatus ( ESvrRegStatus eStatus )
+    {
+        eSvrRegStatus = eStatus;
+        emit SvrRegStatusChanged();
```
pljones, line 179: _May_ be able to stop the time here - but that seems to kick off `timeout()`.  So long as it doesn't make things loop...

src/serverlist.h
```diff
@@ -195,10 +204,23 @@ class CServerListManager : public QObject
...
+    // count of registration retries
+    int                     iSvrRegRetries;
+
 public slots:
     void OnTimerPollList();
     void OnTimerPingServerInList();
     void OnTimerPingCentralServer();
-    void OnTimerRegistering() { SlaveServerRegisterServer ( true ); }
+    void OnTimerCLRegisterServerResp();
+    void OnTimerRegistering()
```
pljones, line 218: single line style

src/protocol.cpp
```diff
@@ -2114,11 +2115,19 @@ bool CProtocol::EvaluateCLChannelLevelListMes  ( const CHostAddress&     InetAdd
...
+                                              const ESvrRegStatus eStatus )
 {
     int              iPos = 0; // init position pointer
     CVector<uint8_t> vecData( 1 );
 
+    int iStatus = static_cast<int> ( eStatus );
+    if ( iStatus < ESvrRegStatus::SRS_MIN_RESP
+         || iStatus > ESvrRegStatus::SRS_MAX_RESP )
+    {
+        qWarning() << "CProtocol::CreateCLRegisterServerResp: eStatus" << eStatus << "invalid value";
```
corrados, line 2127: please remove the qWarning()

pljones: There doesn't appear to be any way to tell the caller of the create method that it's been called badly.  Have got any ideas what could be done in that case?

corrados: Hmm, is it not just for debugging during developement? Is it intended to give a warning for the end user of Jamulus (i.e. not a code developer)?

src/util.h
```diff
@@ -572,6 +572,20 @@ enum ECSAddType
 };
 
 
+// Registration status ---------------------------------------------------------
+enum ESvrRegStatus
+{
+    SRS_BAD_ADDRESS = -4,
```
corrados, line 578: Please use just positive numbers starting from 0. This would be consistent with the existing enums used for the protocol.

src/serverlist.cpp
corrados: Can you please give some more info on what was wrong here?

src/util.h
pljones: The negative values aren't part of the protocol, that's why I used negative.  Only the the non-negative values are exchanged at the protocol layer.

src/serverlist.cpp
pljones: `removeAt(i);` then `i++` -- it means you skip an entry in the list.

src/protocol.cpp
pljones: Actually, as this is only _called_ by the "central server" code, it should never trigger (if the range of values was increased, the MIN and MAX would increase - all this would do would highlight that).  So yes, it's actually useless.  I'll get rid of it.

src/util.h
pljones: Thinking about it more, the two separate enums would probably be more logical - they represent two different things:
- central server outcome of request
- slave server registration status (which may not even include registering)

src/serverlist.cpp
corrados: Ups... Good that you found that.

src/serverlist.cpp
corrados: Do you have fixes in the queue or are you done with this?

pljones: Still going to do these last few I commented on - they're not show stoppers, it works without them...  (It might work _better_ as is, indeed!  But some of it I changed when things were being a bit weird... so may be better removed.)

Not had much time over the last couple of days (Easter was better :) ).  So I'll be reviewing it all again, maybe trying a couple of things out.  What I'll then do is probably close off this PR, squash all the commits and raise a new one to tidy up.

pljones:
OK, I've reviewed everything again and retested.  There's a fair number of changes, so I'll close this off and open a new PR with the squashed push.

src/serverlist.cpp
corrados: No rush, take your time. You have spend a lot of time already for Jamulus coding. Thank you very much for that!

PR #85, opened 2020-04-16
Upgrade OPUS codec to v1.3.1

doloopuntil:
This change updates the OPUS codec from v1.1 to v1.3.1. The most recent version [may reduce CPU usage](https://people.xiph.org/~jm/opus/opus-1.2/) without sacrificing latency as it's been optimised for different architectures (x86, ARM). I've tested it as follows:
* Jamulus client on Mac (x86_64, Intel), builds and works.
* Jamulus client on Windows (x86_64, AMD), builds and works.
* Jamulus server on Ubuntu (x86_64, AMD), builds and works.
* Jamulus server on Debian (ARM, Raspberry Pi 4), builds and works.

Feel free to accept or reject, depending on whether there is appetite for the updated codec. The OPUS source code was downloaded from [here](https://opus-codec.org/release/stable/2019/04/12/libopus-1_3_1.html).


corrados:
Thank you very much for your contribution!
I would also be very interested in your changes in https://github.com/doloopuntil/jamulus/commit/a2f74047c65656dfd7e50fcb0532cd796089c7fd
Could you create a pull request just for these changes (i.e. get rid of the many "connect" lines in the code)?

PR #87, opened 2020-04-17
Allow central server to confirm registration

pljones:
Also display this state in the server UI.

Rebased and retested with the latest master.

### DIFF
--- a/src/protocol.cpp
+++ b/src/protocol.cpp
@@ -289,6 +289,7 @@ CONNECTION LESS MESSAGES
 
     note: does not have any data -> n = 0
 
+
 - PROTMESSID_CLM_CHANNEL_LEVEL_LIST: The channel level list
 
     +----------------------------------+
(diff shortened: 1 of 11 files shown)


src/protocol.cpp
```diff
@@ -304,13 +305,32 @@ CONNECTION LESS MESSAGES
...
     PROTMESSID_CLM_REQ_CHANNEL_LEVEL_LIST to opt in
 
 
+- PROTMESSID_CLM_REGISTER_SERVER_RESP: result of registration request
+
+    +---------------+
+    | 1 byte status |
+    +---------------+
+
+    - "status":
```
corrados, line 321: As far as I understood, you are transmitting the enum now and not a single bit, right? Then you should update the documentation here, too.

corrados: I will merge your changes now. Please fix this in another pull request later on.

corrados:
Thank you for your contribution!

corrados:
The central server code is now updated. If you like you can test it now with the official Central Server.

pljones:
Looks good.  The styling on the status messages makes them more obvious.  The only thing is it appears to have lost the "and will re-try" after a timeout - unless that's no longer happening (which could also make sense).

corrados:
Are you referring to this line which I have changed? https://github.com/corrados/jamulus/blob/master/src/serverdlg.cpp#L565

I did this because the original text was too technical in my opinion.

pljones:
Yeah, it was a bit clumsy - I'm happy they got changed.  But it lost information, too - it mentioned retries, now it doesn't.

Looking at it, this one may well be wrong:
https://github.com/corrados/jamulus/blob/master/src/serverdlg.cpp#L569
I think, because it _got_ a response - just like success and full states - it won't retry.

corrados:
Yes, but I think if it does not succeed and the error is not "server list full", something fatal does not work and the user has to do something to solve it anyway. I do not think the "retry" will give useful additional information to the user. Do you agree?

corrados:
> Looking at it, this one may well be wrong:

What would you write instead?

pljones:
What it _means_ is that the server response was successfully received but not understood.

The retries are just the "standard ping" ones, so "retrying" is misleading, really.

Basically, the slave server is too old to understand and needs to be upgraded.

corrados:
done

pljones:
Based off #91 I added this patch:
https://github.com/corrados/jamulus/pull/99

- Moves the stringifying code out
- Passes "nogui" down to server side components
- Logs connection state to central server

PR #88, opened 2020-04-17
Refactor signal to slot connections using recursive templates

doloopuntil:
This change removes code bloat and allows the maximum number of supported clients to be adjusted in a single place rather than in multiple places.
Note that this change requires Qt5's new signal/slot syntax (https://wiki.qt.io/New_Signal_Slot_Syntax). The code will no longer compile using Qt4.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -17,13 +17,13 @@ Linux
 Required packages: 
 
 - Build-Essential
-- Qt4/Qt5 (devel packages, too!)
+- Qt5 (devel packages, too!)
 - Jack (devel packages, too!)
 - qjackctl can be a good help configure jack.
 
 On the most common Linux distributions, the following command should prepare the system for compilation:
 
-sudo apt-get install build-essential libqt4-dev libjack-jackd2-dev
+sudo apt-get install build-essential qt5-default libjack-jackd2-dev
 
 On Fedora use:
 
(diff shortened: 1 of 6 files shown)


corrados:
Thank you for the pull request.
I was not aware that it brakes compatibility to Qt4. I am now unsure wether it is a good idea to brake Qt4 compatibility now. Maybe I should first ask the people in the Discussion forum for their opinion on that...

corrados:
I just created a thread in the Discussion forum: https://sourceforge.net/p/llcon/discussion/533517/thread/6d3f3feb2d/

doloopuntil:
Great thanks! The lack of backwards compatibility with Qt4 is the reason why I didn't create a pull request earlier on. I couldn't find a way around it unfortunately. On the other hand, given that Qt5 was released at the end of 2012, I would expect it to be well supported by any modern Linux distribution and Mac OS and Windows releases. As usual, the final decision is yours :)

corrados:
Maybe I put a #if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0) around your code and leave the old code in there. Of course, that would not really clean up the Jamulus code but at least we have your code in the Jamulus code base and later on it would be easy to just remove the old code. What do you think about that?

doloopuntil:
That actually does sound like the best solution at this time. You're right, it will be trivial to remove the Qt4 code later and not break backwards compatibility immediately.

corrados:
Good to hear that you have the same opinion. Since it is your code and you know it best, could you please do the "#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)" changes? If you do not have time to do it it's fine, I could do not myself. But right now I am very busy.

doloopuntil:
Will do, will send an update tomorrow

doloopuntil:
Refactored as agreed above in https://github.com/corrados/jamulus/pull/92. Closing this pull request.

PR #92, opened 2020-04-18
Refactor signal to slot connections using recursive templates

doloopuntil:
This change allows the maximum number of supported clients to be adjusted in a single place rather than in multiple places.
Note that this change requires Qt5's new signal/slot syntax (https://wiki.qt.io/New_Signal_Slot_Syntax), but is still backwards compatible with Qt4.
Code bloat will be eventually reduced by removing support for Qt4.

Supersedes https://github.com/corrados/jamulus/pull/88.


corrados:
Thank you!

PR #98, opened 2020-04-18
debian package number

braindef:
... that the debian/ubuntu/mint package manager shows the right package number (3.5.2?)

### DIFF
--- a/distributions/debian/changelog
+++ b/distributions/debian/changelog
@@ -1,4 +1,4 @@
-jamulus (3.4.7) UNRELEASED; urgency=medium
+jamulus (3.5.2) UNRELEASED; urgency=medium
 
   * Initial release.
 


trebmuh:
mmm well, that's not really the _3.5.2_, that is the _3.5.1 + git master head_. It'll be 3.5.2 when 3.5.2 will be released. Putting 3.5.2 here now, and leaving people to build from those debian rules could be confusing (when debugging for example).

braindef:
at the moment it's 3.4.7 which is some sort of confusing too, shall i change to 3.5.1?

trebmuh:
Yes, it can be confusing. What about _3.5.1+gitmasterhead_ ?

corrados:
Does it mean that the file distributions/debian/changelog has to be maintained everyt time I make a new release? This is not good. Right now I only change the ChangeLog in the Jamulus root directory and the Jamulus.pro file (see, e.g., this commit: 6615c97561eba547cb7445544394acbc0e81df8b). I do not want to maintain more files than this.
Is there a way to update this version number from the Jamulus.pro file?

trebmuh:
Let's ping @mirabilos to check what he thinks about this.

mirabilos:
Olivier Humbert dixit:

>Let's ping @mirabilos to check what he thinks about this.

I’m uploading Jamulus to Debian, so this in-tree packaging will
be unnecessary in the medium term.

For until then, a version number that is smaller than the one
used by the Debian package would be great, but you’re already
using them (current is 3.5.1 which is smaller than 3.5.1-1).

For prereleases we often use the tilde operator, so 3.5.2~git
would work well. (Tilde sorts below anything else, even end
of string.)

trebmuh:
Well, it will be unnecessary **for the Debian project purpose** but there are a lot of other case where it can be usefull.

mirabilos:
Olivier Humbert dixit:

>Well, it will be unnecessary **for the Debian project purpose** but
>there are a lot of other case where it can be useful.

I can provide backports (within Debian) and a PPA for *buntu (LTS,
ideally). I do the same for MuseScore.

If you have any other ideas in mind, do tell ☻

trebmuh:
Another usecase you mean? Well, the same but for a user using any Debian derivative outside of the things you provide (which are very great, to take me wrong, but doesn't pass the "bus factor" level).

mirabilos:
Olivier Humbert dixit:

>Another usecase you mean? Well, the same but for a user using any
>Debian derivative outside of the things you provide (which are very

OK, point.

braindef:
what about integrating the modification of jamulus/distributions/debian/changelog into makefile or qt.pro file with a sed or awk so that it gets changed automatically?

@mirabilos i've seen that you are debian developper/maintainer, would it be possible to integrate jamulus into the next release of debian? I was trying getting jamulus into debian about 1 year before, but it failed because i dont know any debian maintainers.

with kind regards
Marc jr. Landolt

mirabilos:
@braindef that’s _precisely_ what I’m currently working on ☻ but I’m wary of getting it into Debian stable+1 as long as the opus_custom situation is not resolved. I’ve requested the opus maintainers to fix it, though, and until then (or if I get an OK from the stable release managers) it can live in experimental.

PR #99, opened 2020-04-18
Log connection status when no GUI active

pljones:


src/serverlist.cpp
```diff
@@ -479,27 +481,30 @@ void CServerListManager::CentralServerQueryServerList ( const CHostAddress& Inet
 /* Slave server functionality *************************************************/
 void CServerListManager::StoreRegistrationResult ( ESvrRegResult eResult )
 {
+    if ( bConsoleLogging )
+    {
+        qInfo() << "Registration response received:" << svrRegResultToString ( eResult );
```
corrados, line 486: That is interesting. Does qInfo() does the same as your ConsoleWriterFactory?

corrados: If yes, I will clean up the code today with this regard.

pljones: It's a standard Qt global - I don't know how they implement it.

pljones: The only thing to watch out for is it adds its own spacing, formatting and newlines.  So a string variable is output with quotes -- you need to `"stuff.. " + value + " ..more stuff"` rather than `"stuff.." << value << "..more stuff"` to avoid the quotes, I think.  Not tried yet.

corrados: Ok, I'll do the changes.

src/util.h
```diff
@@ -593,6 +622,19 @@ enum ESvrRegResult
     SRR_CENTRAL_SVR_FULL = 1
 };
 
+inline QString svrRegResultToString ( ESvrRegResult eSvrRegResult )
```
corrados, line 625: Why do we have this additional function? Isn't svrRegStatusToString() enough?

src/serverlist.cpp
corrados: qInfo() does not output on the Windows CMD if you start Jamulus -s -n on the Windows CMD. So we still have to use your ConsoleWriterFactory. Can you please changes this here, too?

src/serverlist.cpp
```diff
@@ -575,15 +578,23 @@ void CServerListManager::SlaveServerRegisterServer ( const bool bIsRegister )
...
     else
     {
-        eSvrRegStatus = SRS_BAD_ADDRESS;
+        SetSvrRegStatus ( SRS_BAD_ADDRESS );
         emit SvrRegStatusChanged();
     }
 }
+void CServerListManager::SetSvrRegStatus ( ESvrRegStatus eNSvrRegStatus )
+{
+    if ( bConsoleLogging )
```
corrados, line 594: I changed my mind (sorry for that). We should output all registration infos on the console. It does not hurt and my give some users additional valuable information. So please remove all this bConsoleLogging.

src/main.cpp
```diff
@@ -542,10 +542,11 @@ int main ( int argc, char** argv )
                              bCentServPingServerInList,
                              bDisconnectAllClients,
                              bUseDoubleSystemFrameSize,
-                             eLicenceType );
+                             eLicenceType,
+                             !bUseGUI );
```
corrados, line 546: see below comment, !bUseGUI is not needed.

src/server.cpp
```diff
@@ -233,7 +233,8 @@ CServer::CServer ( const int          iNewMaxNumChan,
                    const bool         bNCentServPingServerInList,
                    const bool         bNDisconnectAllClients,
                    const bool         bNUseDoubleSystemFrameSize,
-                   const ELicenceType eNLicenceType ) :
+                   const ELicenceType eNLicenceType,
+                   const bool         bNConsoleLogging ) :
```
corrados, line 237: see below comment, bNConsoleLogging is not needed.

src/server.cpp
```diff
@@ -247,11 +248,13 @@ CServer::CServer ( const int          iNewMaxNumChan,
                                 strServerInfo,
                                 iNewMaxNumChan,
                                 bNCentServPingServerInList,
-                                &ConnLessProtocol ),
+                                &ConnLessProtocol,
+                                bNConsoleLogging ),
```
corrados, line 252: see below comment, bNConsoleLogging is not needed.

src/server.cpp
```diff
@@ -247,11 +248,13 @@ CServer::CServer ( const int          iNewMaxNumChan,
...
     bAutoRunMinimized         ( false ),
     strWelcomeMessage         ( strNewWelcomeMessage ),
     eLicenceType              ( eNLicenceType ),
-    bDisconnectAllClients     ( bNDisconnectAllClients )
+    bDisconnectAllClients     ( bNDisconnectAllClients ),
+    bConsoleLogging           ( bNConsoleLogging )
```
corrados, line 257: see below comment, bNConsoleLogging is not needed.

src/server.h
```diff
@@ -178,7 +178,8 @@ class CServer :
               const bool         bNCentServPingServerInList,
               const bool         bNDisconnectAllClients,
               const bool         bNUseDoubleSystemFrameSize,
-              const ELicenceType eNLicenceType );
+              const ELicenceType eNLicenceType,
+              const bool         bNConsoleLogging );
```
corrados, line 182: see below comment, bNConsoleLogging is not needed.

src/server.h
```diff
@@ -368,6 +369,7 @@ class CServer :
     QString                    strWelcomeMessage;
     ELicenceType               eLicenceType;
     bool                       bDisconnectAllClients;
+    bool                       bConsoleLogging;
```
corrados, line 372: see below comment, bNConsoleLogging is not needed.

src/serverlist.cpp
```diff
@@ -30,13 +30,15 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
                                          const QString& strServerInfo,
                                          const int      iNumChannels,
                                          const bool     bNCentServPingServerInList,
-                                         CProtocol*     pNConLProt )
+                                         CProtocol*     pNConLProt,
+                                         const bool     bNConsoleLogging )
```
corrados, line 34: see below comment, bNConsoleLogging is not needed.

src/serverlist.cpp
```diff
@@ -30,13 +30,15 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
...
     : iNumPredefinedServers     ( 0 ),
       eCentralServerAddressType ( AT_MANUAL ), // must be AT_MANUAL for the "no GUI" case
       bCentServPingServerInList ( bNCentServPingServerInList ),
       pConnLessProtocol         ( pNConLProt ),
       eSvrRegStatus             ( SRS_UNREGISTERED ),
-      iSvrRegRetries            ( 0 )
+      iSvrRegRetries            ( 0 ),
+      bConsoleLogging           ( bNConsoleLogging )
```
corrados, line 41: see below comment, bNConsoleLogging is not needed.

src/serverlist.h
```diff
@@ -206,6 +208,9 @@ class CServerListManager : public QObject
     // count of registration retries
     int                     iSvrRegRetries;
 
+    // Enable headless logging to console
+    bool                    bConsoleLogging;
```
corrados, line 212: see below comment, bConsoleLogging is not needed.

src/util.h
pljones: There are two enums.  It's mostly to ensure the behaviour is correct - it wasn't entirely clear from the report that raised this.  This logs when the Central Server response is received; status gets logged when the Slave server changes state - either from receiving a response, or not.

src/serverlist.cpp
pljones: No problem - saves passing it down :)

src/serverlist.cpp
pljones: Oh - hm.  OK.  I'll set up a macro I think, to make life simpler in future.

corrados: Hm, wait for your change. Do we need an output on Windows cmd at all? If not, qInfo would do the job... Whats your opinion?

pljones: For future info:
https://doc.qt.io/qt-5/qtglobal.html#qInfo
> Under Windows, the message is sent to the console, if it is a console application; otherwise, it is sent to the debugger.
Maybe it's the _if it is a console application_ - did you run with `-n` from the console?

corrados: yes, I tried it with the -n option

pljones: OK, I tried various things to bring it into line and had no luck.  I don't know C++ or Qt well enough to know why...  I'll have another look later.

corrados: No problem. Let's keep it as is.

PR #102, opened 2020-04-19
make the desktop icon square

mirabilos:
Desktop environments really want square icons, and this makes it easier to install the icon.

I’d prefer one in a power-of-two size (256×256 or 512×512) instead of 406 pixels, though, but this doesn’t seem to have any vector source, so we’ll have to live with this.


corrados:
We have a vector source for it, see: https://github.com/corrados/jamulus/blob/master/src/res/icon.xcf

mirabilos:
corrados dixit:

>We have a vector source for it, see: https://github.com/corrados/jamulus/blob/master/src/res/icon.xcf

Ah okay, then please hold off merging this, and I’ll re-export in
a more suitable size.

Thank you!

>Wow, many pull requests today ;-).

I started looking at this by a recommendation from an OSAMC member
and before I consider using it, a proper Debian package is needed,
so I’m packaging it for Debian ☺

mirabilos:
Updated by regenerating it from the source at the right size. Also, installing it into the correct directory.

corrados:
Great, Thank you.

PR #103, opened 2020-04-19
link the server installation from README.md as well

mirabilos:
so we don’t need to install `INSTALL.md` in the Debian binary package (all except this one link is only needed for the from-source installation)


corrados:
Thanks!

PR #113, opened 2020-04-19
Improve installer packaging mechanism for macOS

doloopuntil:
This change automates the creation of an installer image for macOS containing both a client and server version of Jamulus. It also creates an installer image offering the "classic" Mac installation experience, where the user drags the app bundle into their local Application folder from the installer window.


corrados:
Thank you very much! Great installer experience now under Mac. Much better than before.

corrados:
I just noticed that after I merged your code, Travis is no longer happy:
https://travis-ci.org/github/corrados/jamulus/jobs/676970093
I guess the changes in Jamulus.pro have caused the issue since the qmake fails. Do you have any idea what goes wrong there?

doloopuntil:
I think it may be because I introduced a templated `Info.plist` file that Qt + `make` can parse, but Qt + Xcode cannot. I don't know Travis but in the log I can see:
```
/Users/travis/build/corrados/jamulus/Debug/Jamulus.app/Contents/Info.plist
warning: invalid character '@' in Bundle Identifier at column 0. This string must be a uniform type identifier (UTI) that contains only alphanumeric (A-Z,a-z,0-9), hyphen (-), and period (.) characters.
```
The @ character is part of a tag in `Info.plist`. I think the tags get replaced by `make` when building the Jamulus bundle, but Xcode doesn't know what to do with those and at the end of the deployment it cannot build the bundle because `Info.plist` is corrupted. 

I guess an option could be to take the `Info.plist` file from a working Jamulus.app bundle and use it to override the `QMAKE_INFO_PLIST` variable in `qmake` when you build with Travis.

doloopuntil:
Another alternative is to use the `mac/deploy_mac.sh` build script instead of the custom Travis build script.

corrados:
I just implemented a simple solution to the problem. It is already applied to Git and Travis succeeded again.

corrados:
New idea: Maybe it is better to implement it the other way around:
Instead of
!contains(CONFIG, "istravis") {
we may introduce
contains(CONFIG, "performdeployment") {.

If we do that, the travis.yml file could be reverted back to its original code and you would have to call qmake with qmake "CONFIG+=performdeployment" [...].
What do you think?

doloopuntil:
I think there's an even simpler solution: https://github.com/doloopuntil/jamulus/commit/22c53363a5cba4d98c0b3fc2710a6b2b9cb7fea7.

I would recommend to use the existing deployment scripts `mac/deploy_mac.sh`, `windows/deploy_windows.bat` and `linux/deploy_linux.sh` to build and deploy using Travis, so there's parity between what Travis builds and what developers can build straight off the repository. Unless there's a reason to maintain different deployment modes for the same OS? I've been working on an improved Windows deployment script and I'll send a provisional PR at some point this week. I setup Travis at my end and the diff above works for the macOS build: https://travis-ci.com/github/doloopuntil/jamulus/jobs/321799524

corrados:
Ok, I am fine with using deploy_mac.sh in Travis. On the other hand I would still additionally implement the performdeployment CONFIG parameter, just for the case someone wants to compile Jamulus with xcode. It does not hurt to implement this and maybe someone benefits from it.
If you also worked on a Windows installer, there are recently requests to install the 32 bit Jamulus binary on a 64 bit Windows (since we have Jack Audio compatiblilty issues right now). Is it possible to let the user chose during installation process if he want to install the 32 bit or 64 bit version of Jamulus? Of yourse, if NSIS detects a 32 bit Windows, he will not be asked to install the 64 bit version.
What do you think?

doloopuntil:
Makes sense, I'll make the CONFIG change so we can support both Xcode and standard make. I will have to investigate if NSIS can offer the choice between 32 and 64 bit and let you know. I think I can fix the broken Windows Travis deployment too.

corrados:
Sounds great. Looking forward to your contribution.

corrados:
Regarding this:

> Is it possible to let the user chose during installation process if he want to install the 32 bit or 64 bit version of Jamulus?

There is new information in https://github.com/corrados/jamulus/issues/93. I think we do not need that anymore. I will look at this today. If it does not crash anymore with that workaround, we keep the Windows installer as is, i.e. on 64 bit Windows we only install the 64 bit Jamulus. Do you agree?

doloopuntil:
Great news! I agree. I still have a few improvements for the Windows deployment script I can propose (mainly as few steps as possible to successfully build Jamulus from a Windows command line) and I have actually fixed the unattended Qt installation process that I believe was preventing Travis to build Jamulus for Windows.

PR #115, opened 2020-04-19
Allow client to bind on a random port

stanislas-m:
I some cases, we just want to run the Jamulus client on a random available port on our computer. :)

This PR prevents the 0 value to be incremented, so we can let the system choose a port for us.

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -54,23 +54,35 @@ void CSocket::Init ( const quint16 iPortNumber )
 
     if ( bIsClient )
     {
-        // Per definition use the port number plus ten for the client to make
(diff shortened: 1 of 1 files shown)


vocobox:
Thanks. If the port is not default, will it be announced to the public server? This was mentionned in https://github.com/corrados/jamulus/issues/91 but I think the ticket has been closed without fix.

corrados:
Thanks!

pljones:
> Thanks. If the port is not default, will it be announced to the public server?
That only applies to servers registering with the central server so that they are announced to clients.  It's not about whether other clients on the same server see yours.

PR #124, opened 2020-04-20
Add config for mingw compiler for Windows

stanislas-m:
This allows Jamulus to be cross-compiled with MinGW compiler.


corrados:
Thanks!

PR #126, opened 2020-04-21
Use a sortable and more standard datetime logging format

mirabilos:
The current format (`d.M.yyyy, HH:mm:ss`) resembles the old pre-ISO-8601 German format, except it omitted the mandatory leading zeroes, worsening the already bad sortability.

Change to `yyyy-MM-dd HH:mm:ss` which is ISO but with a space instead of the letter `T` in the middle, as is common for cross-national (and even increasingly common for national) use and perfectly sortable.

I also squeezed in a fixup for commit e2014956355a578b7df0691d3614fca669c14ab7 which I noticed too late, i.e. after #102 was already merged (to my excuse, in the Debian package I’m working on the path only shows up on one line), as to not overflow you with MRs.

### DIFF
--- a/src/serverlogging.cpp
+++ b/src/serverlogging.cpp
@@ -118,24 +118,17 @@ void CServerLogging::ParseLogFile ( const QString& strFileName )
         QStringList strlistCurLine = strCurLine.split ( "," );
 
         // check number of separated strings condition
-        if ( strlistCurLine.size() == 4 )
+        if ( strlistCurLine.size() == 3 )
         {
             // first entry
-            QDate curDate =
-                QDate::fromString ( strlistCurLine.at ( 0 ).trimmed(),
-                "d.M.yyyy" );
+            QDateTime curDateTime =
+                QDateTime::fromString ( strlistCurLine.at ( 0 ).trimmed(),
+                "yyyy-MM-dd HH:mm:ss" );
 
-            // second entry
-            QTime curTime =
-                QTime::fromString ( strlistCurLine.at ( 1 ).trimmed(),
-                "hh:mm:ss" );
-
-            if ( curDate.isValid() && curTime.isValid() )
+            if ( curDateTime.isValid() )
             {
-                QDateTime curDateTime ( curDate, curTime );
-
                 // check if server stop or new client connection
-                QString strAddress = strlistCurLine.at ( 2 ).trimmed();
+                QString strAddress = strlistCurLine.at ( 1 ).trimmed();
 
                 if ( strAddress.isEmpty() )
                 {
@@ -147,8 +140,8 @@ void CServerLogging::ParseLogFile ( const QString& strFileName )
                 {
                     QHostAddress curAddress;
                     
-                    // third entry is IP address
-                    if ( curAddress.setAddress ( strlistCurLine.at ( 2 ).trimmed() ) )
+                    // second entry is IP address
+                    if ( curAddress.setAddress ( strlistCurLine.at ( 1 ).trimmed() ) )
                     {
                         // new client connection
                         JpegHistoryGraph.Add ( curDateTime, curAddress );
@@ -168,9 +161,6 @@ QString CServerLogging::CurTimeDatetoLogString()
     // time and date to string conversion
     const QDateTime curDateTime = QDateTime::currentDateTime();
 
-    // format date and time output according to "3.9.2006, 11:38:08"
-    return QString().setNum ( curDateTime.date().day() ) + "." +
-        QString().setNum ( curDateTime.date().month() ) + "." +
-        QString().setNum ( curDateTime.date().year() ) + ", " +
-        curDateTime.time().toString();
+    // format date and time output according to "2006-09-30 11:38:08"
+    return curDateTime.toString("yyyy-MM-dd HH:mm:ss");
 }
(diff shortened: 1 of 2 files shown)


src/serverlogging.cpp
```diff
@@ -168,9 +168,6 @@ QString CServerLogging::CurTimeDatetoLogString()
     // time and date to string conversion
     const QDateTime curDateTime = QDateTime::currentDateTime();
 
-    // format date and time output according to "3.9.2006, 11:38:08"
-    return QString().setNum ( curDateTime.date().day() ) + "." +
-        QString().setNum ( curDateTime.date().month() ) + "." +
-        QString().setNum ( curDateTime.date().year() ) + ", " +
-        curDateTime.time().toString();
+    // format date and time output according to "2006-09-30 11:38:08"
```
corrados, line 171: This function: https://github.com/corrados/jamulus/blob/master/src/serverlogging.cpp#L103
will not work anymore with your change.

mirabilos:
corrados dixit:

>This function: https://github.com/corrados/jamulus/blob/master/src/serverlogging.cpp#L103
>will not work anymore with your change.

Oh ouch, it parses the log as well… thanks for pointing this out.

I’ve updated it, do you think it can work?

corrados:
The problem is that you will not be able to read existing history files. E.g. @pljones operates a Jamulus server homepage where he shows the server history in a plot:
http://jamulus.drealm.info/jamulus
See "recent events". If you upgrades to your new Jamulus version, he would not be able to read the existing file and the plot would be empty.
@pljones what is your opinion to the change?

pljones:
I'd prefer the ISO format, really, myself.  I've not changed it for the "backward compatibility" issue.  Perhaps there could be (yet another) command line option to override the default format (simply `--log-iso`)?

corrados:
No, I would not introduce a new command line argument for that. If you are happy with the change (and I think you are one of the "power users" of this functionality) I am fine with the change.
Since your server uses it, could you please checkout mirabilos branch and check if the new code works fine for you?

pljones:
The code looks good - I'll update my site when I next build -- just needs to take my PHP reformatter out.

mirabilos:
corrados dixit:

>See "recent events". If you upgrades to your new Jamulus version, he
>would not be able to read the existing file and the plot would be
>empty.

I see.

If necessary, I could refactor the parsing code so that it
understands either format.

pljones:
@mirabilos - no, not necessary - I only parse to reformat into the format you're making it.

Only other thing would be to use UTC rather than server time... but that might be stretching things.

mirabilos:
Peter L Jones dixit:

>@mirabilos - no, not necessary - I only parse to reformat into the
>@format you're making it.

☻

>Only other thing would be to use UTC rather than server time... but
>that might be stretching things.

I wasn’t daring to. It’s the sort of things proper admins expect,
but GUI users can be confused. But feel free to one-up this.
QDateTime::currentDateTimeUtc() was introduced in Qt 4.7 so is
probably recent enough ☺

bye,
//mirabilos
-- 
22:59⎜<Vutral> glaub ich termkit is kompliziert | glabe nicht das man
damit schneller arbeitet | reizüberflutung │ wie windows │ alles evil
zuviel bilder │ wie ein spiel | 23:00⎜<Vutral> die meisten raffen auch
nicht mehr von windows | 23:01⎜<Vutral> bilderbücher sind ja auch nich
wirklich verbreitet als erwachsenen literatur	‣ who needs GUIs thus?

pljones:
http://victorypoint.ddns.net/
^ Al Udell is also using the logs (and, uh, my PHP code to parse).  I've explained it's about to change.

corrados:
@pljones, have you already tested the new code? If yes, did it work as expected?

pljones:
I'll be happy with it - I've not had time to look at anything the last few days.  Probably be sometime at the weekend before I do.

corrados:
OK, no hurry. I'm not planning to change anything in the server logging soon so I can merge this pull request when you are done.

corrados:
Since there is a bug fix for debian also included in this merge request, I will apply it now to the master. If we find bugs in the server history processing, we can then fix it on the master branch.

pljones:
Hi,

And just to confirm - I've upgraded my server to latest master, converted my log file to the new format and updated my PHP and JavaScript and everything's looking fine.

For those wanting to migrate the data from before this patch, here's the bash command I ran:
```
while read dt x; do dt=${dt//,/}; tm=${tm//,/}; dd=${dt%%.*};  mm=${dt#*.}; mm=${mm%.*}; yy=${dt##*.}; printf '%04d-%02d-%02d %s\n' $yy $mm $dd "$x"; done < Jamulus.log > /tmp/Jamulus.log.new
```
(And then, copy the /tmp/Jamulus.log.new over Jamulus.log ... I leave the degree of back up to the server admin's discretion...  I just went for it... forgetting I'd not actually upgraded the server at this point so shuddering when the graph went blank!!)

mirabilos:
Peter L Jones dixit:

>For those wanting to migrate the data from before this patch, here's the bash command I ran:

This doesn’t look right; for one, you don’t define $tm before using it.

How about this:

sed \
    -e 's/^\([0-9]\)\./0&/' \
    -e 's/^\(..\.\)\([0-9]\.\)/\10\2/' \
    -e 's/^\(..\).\(..\).\(....\),/\3-\2-\1/' \
    <Jamulus.log >/tmp/Jamulus.log.new

Mildly tested (on about two dozen log lines, plus some manually
created corner cases).

pljones:
`tm` isn't really used - the bit in the loop was "left over".

Issue #100, opened 2020-04-18
hardcoded text for nōn-free licence


mirabilos:
The current server (at least I *hope* it’s in the server, not in the client) hard-codes a text that requires all contributions to be licenced under a nōn-free, very restrictive, licence (CC-BY-NC-SA 4.0), which is inaccepable (e.g. the bylaws of a charity I work with require Free licences to be used).

I wish to package Jamulus for Debian and therefore make it available to a more general audience. Please provide a way, urgently, for a server administrator to set their own terms of use.

trebmuh:
Duplicate #81 .

corrados:
Since it is a duplicate, I will close this issue now.

mirabilos:
corrados dixit:

>Since it is a duplicate, I will close this issue now.

In the other issue, the problem at hand isn’t really
addressed, and discussion has dissolved into something
completely unrelated and irrelevant both.

Perhaps it’d make sense to close _that_ and keep a
focussed discussion here?

corrados:
Ok, done.

mirabilos:
corrados dixit:

>Ok, done.

Thanks.

OK, so on topic: is the message sent from the server to the client
for the latter to accept before joining already?

If so I’d propose a simple command line option (perhaps -L file),
whose OPTARG is read as UTF-8 encoded text file and then used
instead of the hardcoded text.

This plus some documentation¹ would fully satisfy my needs.

I could probably cobble this together if it has chances to be
accepted. I’m not too versed in C++/Qt but an experienced C
programmer.


① I’ll write a manpage for Jamulus(1) as part of the Debian packaging
  anyway so would put it there plus the Markdown documentation.

mirabilos:
I just started actually running/testing Jamulus and saw that the text is only shown optionally. This works for me as a first start (offering another text file as terms the client has to click through would be nice, but only as a feature request; disabling the checkbox in the server GUI by default would also be useful).

corrados:
I don't know why but I did not read your last comments (got no email or the email got lost).

> If so I’d propose a simple command line option (perhaps -L file)

As you already found out, this is exactly what was already implemented in Jamulus :-).

For futher discussion, please also see https://github.com/corrados/jamulus/pull/128 where we continue the discussion started here.

mirabilos:
corrados dixit:

>> If so I’d propose a simple command line option (perhaps -L file)
>
>As you already found out, this is exactly what was already implemented in Jamulus :-).

[...]

PR #128, opened 2020-04-21
Address part of the licence dialogue

mirabilos:
- mention _which_ Creative Commons licence is used (CC-BY-NC-SA 4.0)
 - do not default to enable the checkbox on GUI server start

Advances, but does not completely close, #100

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -560,10 +560,6 @@ int main ( int argc, char** argv )
                              eLicenceType );
             if ( bUseGUI )
             {
-                // special case for the GUI mode: we want the licence type to be
-                // creative commons per default (if not given in the settings file)
-                Server.SetLicenceType ( LT_CREATIVECOMMONS );
-
                 // load settings from init-file
                 CSettings Settings ( &Server, strIniFileName );
                 Settings.Load();
--- a/src/serverdlg.cpp
+++ b/src/serverdlg.cpp
@@ -58,8 +58,8 @@ CServerDlg::CServerDlg ( CServer*        pNServP,
 
     // CC licence dialog switch
     chbUseCCLicence->setWhatsThis ( tr ( "<b>Show Creative Commons Licence "
-        "Dialog:</b> If enabled, a Creative Commons Licence dialog is shown "
-        "each time a new user connects the server." ) );
+        "Dialog:</b> If enabled, a Creative Commons BY-NC-SA 4.0 Licence "
+        "dialog is shown each time a new user connects the server." ) );
 
     // Make My Server Public flag
     chbRegisterServer->setWhatsThis ( tr ( "<b>Make My Server Public:</b> If "
--- a/src/serverdlgbase.ui
+++ b/src/serverdlgbase.ui
@@ -56,7 +56,7 @@
    <item>
     <widget class="QCheckBox" name="chbUseCCLicence">
      <property name="text">
-      <string>Show Creative Commons Licence Dialog</string>
+      <string>Show Creative Commons BY-NC-SA 4.0 Licence Dialog</string>
      </property>
     </widget>
    </item>


src/main.cpp
```diff
@@ -560,10 +560,6 @@ int main ( int argc, char** argv )
                              eLicenceType );
             if ( bUseGUI )
             {
-                // special case for the GUI mode: we want the licence type to be
```
corrados, line 563: If the server GUI is started for the very first time (freh install), the CC licence check shall be enabled by default.

mirabilos:
corrados dixit:

>If the server GUI is started for the very first time (freh install),
>the CC licence check shall be enabled by default.

No, it should be disabled by default because it forces a nōn-free
licence on its users otherwise.

bye,
//mirabilos
-- 
> Hi, does anyone sell openbsd stickers by themselves and not packaged
> with other products?
No, the only way I've seen them sold is for $40 with a free OpenBSD CD.
	-- Haroon Khalid and Steve Shockley in gmane.os.openbsd.misc

corrados:
No, it does not force anything. It is just a default on the first start of the server GUI. If you disable it, it will be stored and your setting stays as you wish.
You many Ninjam servers do you know which do not have a CC licence agreement?

corrados:
Github does not work correctly right now :-/. It should be:
"How many Ninjam servers do you know which do not force you to agree to a CC licence?"

mirabilos:
corrados dixit:

>"How many Ninjam servers do you know which do not force you to agree to a CC licence?"

What is Ninjam?

corrados:
This is the mother of all online jamming applications :-)
https://www.cockos.com/ninjam

mirabilos:
Anyway, I’m concerned with a nōn-free default. Creative Commons are not equal Creative Commons. As I said in #100, the bylaws of one of the charities I work with won’t allow it because it’s too restrictive. Furthermore, for an online choir rehearsal, it’d be too demanding if we sing music from contemporary composers still under copyright.

corrados:
That is absolutely no problem. You start the server for the first time and uncheck the licence check. That's it.
Imagine someone who has no experience with providing streaming servers at all. Imagine he starts the Jamulus server and simply registers it in the official server list. Now someone enters his server and streams copyrighted music over his server. If other clients connect, his server will stream that copyrighted music to all the other clients. So the server operator might get in legal trouble. It is unclear if the one who has streamed to music to the server is responsible or the server operator which distributes that stream to all the other connected clients. Are you a legal expert? I am not. But I can say if every single client must accept a CC licence, he is definitely responsible on his own if he streams copyrighted music to the server.

mirabilos:
corrados dixit:

>That is absolutely no problem. You start the server for the first time
>and uncheck the licence check. That's it.

Yes, but in a Free Software ecosystem, the defaults ought to not
be nōn-free…

>Imagine someone who has no experience with providing streaming servers

Exactly this person should not force something nōn-free on their users.

>at all. Imagine he starts the Jamulus server and simply registers it in
>the official server list. Now someone enters his server and streams
>copyrighted music over his server. If other clients connect, his server

At best a service provider or disturbance liability, which
the BGH mostly denied.

>clients. Are you a legal expert? I am not. But I can say if every
>single client must accept a CC licence, he is definitely responsible on
>his own if he streams copyrighted music to the server.

No, the server provider would still be liable, but can then
take the user into recourse. The server’s redistribution would
still not be allowed. But see above.

corrados:
Ok, I think at this point we need some more opinions. E.g. @pljones, @gilgongo, @sthenos or others, can you please comment on this?

trebmuh:
I think that it is safe to have a default text saying that a user shouldn't be broadcasting music without first thinking if laws allows him/her to do so and he/she is responsible for what he/she is broadcasting.
This default text might be overwritten by the server administrator, then, if he/she do something to overwrite the default text, that indicates that he/she knows exactly what he/she is doing and take the responsibility for doing so.

gilgongo:
I could try asking someone I know who works in copyright and has a particular interest in CC and software licensing. They also know a lot people in the field, so if they can't help directly they may know somebody who can.

corrados:
Thanks @trebmuh. But I think this text would best fit in the server welcome message you can activate with the -w command line argument.
@gilgongo, the problem is that the software is used worldwide. I guess we have different law in each different country.
Ok, if that discussion is all about simply a default state of that check box, maybe the simplest solution would be to just change the default as @mirabilos is proposing and everybody is happy...?

corrados:
Changing the default setting does not mean that the function is no longer available. It simply is in the responsibility of the server operator to activate or do not activate it. I can live with that.

pljones:
I agree with @corrados on this.  The NC version is fairly extreme but I think it's necessary as a default.  I think it's probably the best protection for someone who hasn't a clue.

Unfortunately, I don't think the server admin is likely to have read the license, in my experience.  If they do, they'll probably not understand it.  They'll just find the dialogue annoying and turn it off.

Yes, NINJAM servers - nearly all, there's a few that turned it off - have a CC licence.  JamTaba optionally allows you to automatically agree to whatever the server licence is without reading it.  I'm sure most of its users see that as a benefit over ReaNINJAM, again, purely to avoid an annoying popup they want to get past.

From memory and from a good couple of decades back...  Given there is a limit on the number of client connections, I think "broadcasting" wouldn't apply.  Copying could, though.  I'm even uncertain over "publishing" (except where recording is being done and those recordings made available - I publish).

But anyone running a service that allows others to transfer information that _could_ be subject to copyright (and other) law, needs to understand what their responsibilities are.

trebmuh:
@corrados 
> Thanks @trebmuh. But I think this text would best fit in the server welcome message you can activate with the -w command line argument.

I wasn't writing it doesn't have to be on the server side.

mirabilos:
Peter L Jones dixit:

>I agree with @corrados on this.  The NC version is fairly extreme but I
>think it's necessary as a default. I think it's probably the best
>protection for someone who hasn't a clue.

Hrm, how about what I pushed then? Keep it default, but allow the
distributor, per cpp define, to change the default?

>Unfortunately, I don't think the server admin is likely to have read
>the license, in my experience. If they do, they'll probably not
>understand it. They'll just find the dialogue annoying and turn it off.

That’s worse, ofc…

>Yes, NINJAM servers - nearly all, there's a few that turned it off -
>have a CC licence.

But _which_ CC licence? About half of them are acceptable, the other
half aren’t. (But I guess none is always acceptable.)

Other server-specific terms may be needed still, depending on
legislation or to tell people the rules of the private server
they found by accident…

gilgongo:
@corrados 

> the problem is that the software is used worldwide. I guess we have different law in each different country.

I just texted him and  pretty much his first question was "Is this taking place across jurisdictional boarders?" He'll call me tomorrow and explain the details he says.

gilgongo:
@corrados @mirabilos 

As I suspected, the advice is that because of the complexities of music copyrights on the Internet internationally right now, it's best to just leave it to server operators to put up any licence, agreement, or rules of any kind they like as a condition of entering a server. Or none at all. It's up to them. 

The Jamulus project is not in a position to offer guidance on copyright licensing for users, and would identify Jamulus as being a "mere conduit" under laws that might have that concept. 

Hope that helps...

trebmuh:
Agreed for that. Plus, a defaulting protective license seems to be smart enough.

mirabilos:
Jonathan dixit:

>it's best to just leave it to server operators to put up any licence,
>agreement, or rules of any kind they like as a condition of entering a
>server. Or none at all. It's up to them.

That’s what I requested as end result, indeed.

corrados:
@mirabilos I changed my mind. Let's uncheck the licence on first start of the server GUI as you had it in your initial commit. Please change it back and I will merge it.

@gilgongo Thank you for clarifying things on the legal front a bit.

@pljones You are a power user of the Jamulus server. You use recording and publish the recordings on your homepage. Therefore you definitely need the licence agreement. And you know exactly what you configure and what consequences they have.

@trebmuh The reason why I changed my mind:
In all previous Jamulus server GUI implementations there was no check box for licence agreement at all. It was only available via command line switch. Now we have the setting in the GUI and the setting is also stored so if the user checks it, it will be recovered on the next start of the Jamulus server.
If mirabilos has a problem with just the default value (i.e. when the software is started for the very first time) and it gives him trouble to include Jamulus in his distribution he works for, then I think I should support him. After all it is just a default setting. The licence check box is prominent visible in the server GUI so every Jamulus server operator sees it.

Thank you all for your opinions on that matter.

mirabilos:
corrados dixit:

>@mirabilos I changed my mind. Let's uncheck the licence on first start
>@of the server GUI as you had it in your initial commit. Please change
>@it back and I will merge it.

OK, done.

>@gilgongo Thank you for clarifying things on the legal front a bit.

+1

>@pljones You are a power user of the Jamulus server. You use recording
>@and publish the recordings on your homepage. Therefore you definitely
>@need the licence agreement. And you know exactly what you configure
>@and what consequences they have.

I was wondering whether a short note in the documentation about these
options would be useful, but from what we heard, it’s too different
internationally, and so “The Jamulus project is not in a position to
offer guidance” is probably a good stance to have.

>After all it is just a default setting.

That was my thought as well.

>The licence check box is prominent visible in the server GUI so every
>Jamulus server operator sees it.

That being said, is there any chance for access control, and/or for
the idea that server operators should be able to configure their own
click-through texts (which we probably should discuss in #100) or, if
UDP packet size is a concern, perhaps a link…

trebmuh:
@corrados I don't think it is a good idea since it does look less protective for the end user.$

If @mirabilos really want to change this defaulting option because of a Debian needs, he can patches the Debian package (it's very easy to do using dquilt). Because right now, the default behavior is changing for every one, just because of a Debian package need.

But hey, you're the boss on here, so that's your call!

_(and I can understand that having jamulus distributed within Debian can be good for the project)_

corrados:
> less protective for the end user [...] jamulus distributed within Debian

Yes, it is a trade-off. I decided in favor of mirabilos since I want to support his work. And as I mentioned above, the situation has already improved for the end user since in previous versions of Jamulus there was no GUI control for it at all. It is just about the default on the very first startup of the server.

mirabilos:
corrados dixit:

>> less protective for the end user [...] jamulus distributed within Debian
>
>Yes, it is a trade-off. I decided in favor of mirabilos since I want to

If there’s still doubt, I’d be just as happy with the ifdef solution.
It’s just that what we forward upstream at first is what we’ve built
ourselves.

gilgongo:
Thread is getting a bit hard to follow, and I missed @pljones making a comment on the type of licence that would offer the best "protection" and another statement made about what might constitute broadcasting and copying. 

I know these comments are made in good faith, but for the avoidance of doubt as this is a public forum:  

**The Jamulus project is not in a position to make any statement about legal qualification, use of any licence, agreement, or other form of contract when it comes to people using Jamulus to play music together over the public Internet.**

PR #136, opened 2020-04-24
48x48 -> 512x512 (Minor bug)

braindef:
Auto Building the Debian Package was broken... i cloned 15 mins ago, so it should be up to date

Greetings
Marc jr.


corrados:
Thanks Marc. The bug is already fixed in another pull request: https://github.com/corrados/jamulus/pull/126

Regarding the version number, I just merged: https://github.com/corrados/jamulus/pull/98

PR #145, opened 2020-04-25
Implemented panning for channels.

tarmoj:
No work with style for the panning slider,  also it could be hidden in mono mode.

Please have a look to the changes, I suggest to do it in a separate branch, I used "panning".

Seemed to work for me. Please comment and let know! It was no easy job to find out all the connections and logic of the code...


pljones:
Might be a good idea to fix up all the spacing problems and add the protocol documentation.

corrados:
Thank you for your code. Unfortunately, there are a lot of things I would like to have changed before I can merge it to the Jamulus master branch. The question is if it is faster if I just used your code as a template and code it on the master branch myself or I spend the time to write the codereview and let you make the changes. Maybe the first option is the better one.

tarmoj:
Hi!

Either way that is better for you - it is nice if the functionality arrives to the software in some way, you decide, what makes more sense. But I am a bit curious - I wrote the functionality based very closely on your code how gain value is passed to the server and I tried to follow your style as much I could. So I am a bit stunned that it is so much different from what you would think :) Anyway - I follow with great interest how it develops and thanks a lot for the software!
tarmo

corrados:
It is not so bad as my initial post implies :-). I will do a quick code review today to let you know what issues I have with your code.

src/audiomixerboard.cpp
```diff
@@ -40,6 +40,8 @@ CChannelFader::CChannelFader ( QWidget*     pNW,
     plbrChannelLevel            = new CMultiColorLEDBar ( pLevelsBox );
     pFader                      = new QSlider           ( Qt::Vertical, pLevelsBox );
 
+    pPan						= new QSlider			(Qt::Horizontal, pLevelsBox);
```
corrados, line 43: spaces and tabs should be consistent with the existing code

src/audiomixerboard.cpp
```diff
@@ -64,6 +66,11 @@ CChannelFader::CChannelFader ( QWidget*     pNW,
     pFader->setRange        ( 0, AUD_MIX_FADER_MAX );
     pFader->setTickInterval ( AUD_MIX_FADER_MAX / 9 );
 
+    // setup panning slider
+    pPan->setRange( 0, AUD_MIX_FADER_MAX);
```
corrados, line 70: The problem is that if the user changed the pan, he will never be able to set it back exactly in the middle again. We would need some sort of "snap to the middle" functionality.
Also, have you checked how the new slider looks in "fancy" GUI style? Does it show up nicely?

src/audiomixerboard.cpp
```diff
@@ -243,8 +259,9 @@ void CChannelFader::SetupFaderTag ( const ESkillLevel eSkillLevel )
 
 void CChannelFader::Reset()
 {
-    // init gain value -> maximum value as definition according to server
+    // init gain and pan value -> maximum value as definition according to server
     pFader->setValue ( AUD_MIX_FADER_MAX );
+    pPan->setValue ( AUD_MIX_FADER_MAX/2 );
```
corrados, line 264: Ok, you can re-use AUD_MIX_FADER_MAX but we may create a new one for pan. But this is low priority.

src/audiomixerboard.cpp
```diff
@@ -309,6 +338,20 @@ void CChannelFader::SendFaderLevelToServer ( const int iLevel )
     }
 }
 
+void CChannelFader::SendPanValueToServer ( const int iPan )
+{
+
+    // if mute flag is set or other channel is on solo, do not apply the pan
```
corrados, line 344: Why is this condition? If you do not send the pan to the server, the following could happen: The Mute is enabled. Then the user changes the pan which is not sent to the server. Then he unmutes. I assume the pan is not correct at the server then.

src/audiomixerboard.cpp
```diff
@@ -309,6 +338,20 @@ void CChannelFader::SendFaderLevelToServer ( const int iLevel )
...
+    if ( ( pcbMute->checkState() == Qt::Unchecked ) &&
+         ( !bOtherChannelIsSolo || IsSolo() ) )
+    {
+        // emit signal for new pan value
+        double dPan =  static_cast<double> ( iPan ) / AUD_MIX_FADER_MAX;
```
corrados, line 349: Why to use "double" type here?

src/audiomixerboard.cpp
```diff
@@ -562,6 +605,13 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent, Qt::WindowFlags ) :
 
 
     // Connections -------------------------------------------------------------
+
+    for (int i=0; i< static_cast<int>( vecpChanFader.size() ); i++) {
```
corrados, line 609: Does this work with Qt4?

src/audiomixerboard.cpp
```diff
@@ -908,6 +958,20 @@ void CAudioMixerBoard::UpdateGainValue ( const int    iChannelIdx,
     emit ChangeChanGain ( iChannelIdx, dValue );
 }
 
+void CAudioMixerBoard::OnPanValueChanged ( const double dValue )
+{
+
+    // get channel index from sender according t, this SHOULD coincide with the chanID.
+    CChannelFader * channelFader = qobject_cast<CChannelFader * >(sender());
```
corrados, line 965: I want to have consistency in the code. The pan can be treated exactly like the fader value of that channel. I do not want to have a different type of implementation for the pan.

src/server.cpp
```diff
@@ -1352,6 +1358,10 @@ void CServer::ProcessData ( const CVector<CVector<int16_t> >& vecvecsData,
             // get a reference to the audio data and gain of the current client
             const CVector<int16_t>& vecsData = vecvecsData[j];
             const double            dGain    = vecdGains[j];
+            const double			dPan = vecdPannings[j];
+            const double			dPanCoefL = sqrt(1-dPan); // faster: (1-dPan)
+            const double			dPanCoefR = sqrt(dPan); // (dPan)
+            // see http://write.flossmanuals.net/csound/b-panning-and-spatialization/ for better formula, if needed
 
             // if channel gain is 1, avoid multiplication for speed optimization
```
corrados, line 1366: I check for channel gain == 1 for speed optimiziation but your Pan coefficient is always multiplied. I am not sure how much speed we gain from this "gain == 1" check. Maybe we could do the multiplications also with the dGain always. But I think the less processing we have per block, the lower the jitter will be.

src/server.cpp
```diff
@@ -1352,6 +1358,10 @@ void CServer::ProcessData ( const CVector<CVector<int16_t> >& vecvecsData,
             // get a reference to the audio data and gain of the current client
             const CVector<int16_t>& vecsData = vecvecsData[j];
             const double            dGain    = vecdGains[j];
+            const double			dPan = vecdPannings[j];
+            const double			dPanCoefL = sqrt(1-dPan); // faster: (1-dPan)
```
corrados, line 1362: If the dPan is 0.5, then we have sqrt(0.5) = 0.7. So we would get an attenuation. Is that wanted?

src/server.cpp
```diff
@@ -1389,20 +1400,22 @@ void CServer::ProcessData ( const CVector<CVector<int16_t> >& vecvecsData,
                     {
                         // left channel
                         vecsOutData[k] = Double2Short (
-                            vecsOutData[k] + vecsData[i] * dGain );
+                                    vecsOutData[k] + vecsData[i] * dGain * dPanCoefL);
```
corrados, line 1403: Maybe instead of having a dGain and a dPan we should combine both and just multiply each sample with that single gain value, this would safe us some multiplications.

src/audiomixerboard.cpp
tarmoj: Thanks a lot for the reviews. I comment below and please let me know what you prefer -  if I make the corrections as much I can and make a new pull request (most likely you need to put your hands on it on some points anyway) or you prefer to handel the code yourself.

Tabs and spacing -  sure. Sorry, I had tabs for identation first and then after change corrected as many places I found but several probably slipped through. Will correct that, of course.

src/audiomixerboard.cpp
tarmoj: Yes, this is not implemented yet. A user friendly way would be maybe  to have doubleclick on the slider handle to put it to default position (center)  but this requires subclassing QSlider and reimplementing its mouseDoubleClickEvent  https://doc.qt.io/qt-5/qwidget.html#mouseDoubleClickEvent -  or adding a small button to center it. 

In practice I don't think it is very crucial, if the pan is set to the middle by hand more or less, it sounds in center enough.

I did not do any work yet on the styleSheet or look and feel, this is now just to test the functionality.

src/audiomixerboard.cpp
tarmoj: Yes, I agree, for better readability and logic of the code.

tarmoj: Did that. I noticed also that the pan states are not stored and restored as volume fader values yet. But maybe it is ok, when new client connects, the pan is always in the centre.

src/audiomixerboard.cpp
tarmoj: Why not? This is not Qt specific, rather C++.  So it might depend on the compiler more than Qt version, but mist likely there is no problem. I did not try a build with Qt4 though.

src/audiomixerboard.cpp
tarmoj: I understand. The reason to try out this solution was since creating 50 more signals and slots (like fader) seemed not so elegant - as it was when I started to work on this code. Now you have created a conditional  way to reduce it for Qt5 (using connectFaderSignalsToMixerBoardSlots and the slotID) but maybe there is still a way to keep it simpler for Qt4, too? 

I am not sure, is it so or can it be used so that the channelID of a fader/pan can be the same as index of the according CChannelFader object?? Then perhaps just we could add a field iChanID to CChannelFader, set it on creation and make in the same time the connections? Something like this in CAudioMixerBoard::CAudioMixerBoard:

    {
        vecpChanFader[i] = new CChannelFader ( this, pMainLayout );
        vecpChanFader[i]->Hide();
       vecpChanFader[i]->chanID = i;
       QObject::connect ( vecpChanFader[i],  SIGNAL ( panValueChanged ( double ) ),
                           this, SLOT ( OnPanValueChanged ( double ) ) ); 
      QObject::connect ( vecpChanFader[i],  SIGNAL ( gainValueChanged ( double ) ),
                           this, SLOT ( OnGainValueChanged ( double ) ) );
    }

and the slot gets the iChanID as:

CChannelFader * channelFader = qobject_cast<CChannelFader * >(sender());
int iChanID = channelFade->iChanID;
emit ChangeChanPan ( iChanID, dValue );

(sorry this code is just as an example, not correct).

src/server.cpp
tarmoj: I believe that for nowadays computers a multiplication is really very little load. If to think about audio processing and how complex computations are done in audio convolution plugins, for example, I don't think this is really matters much. And if to use panning (if not in mono mode), it needs multiplication anyway.

src/server.cpp
tarmoj: Yes, this sounds better. If we use linear calculation and in case pan=0.5, it sounds (seems) weaker when sound is in the center than on the edges. This is a bit unnatrural. The sqrt balances it a bit, so on pan=0.5 both channels get 0.7 of the power and that seems more real. Even better would be to use a calulation like:
gainL =  sin((dPan + 0.5) * $M_PI*2) * dGain
gainR = cos((dPan + 0.5) * $M_PI*2) * dGain
(then the maximum power is a bit before the hard left/right)
but i think it is not so important here and we can save some computation power.

src/server.cpp
tarmoj: Yes, to have dGainL and dGainR computer once by chaneel (in case of stereo) and not by every sample is better.

src/server.cpp
pljones: Don't know if this is of any help:

If pan is centre (0.5) and gain is unattenuated (1), then the value to multiply by should be 1 for both channels - and the optimiser should skip doing anything.

Unattentuated hard left and right should sum to -6dB for mono - i.e. multiply by 0.5 on one side, 0 on the other.  (Stereo pan should be mixing further to mono and ending up about -4.5dB on either channel, if I remember theory correctly..., that's the 0.7.)

src/audiomixerboard.cpp
corrados: Sorry for the delay. Now I again have time for your pull request :-).
I do not find a double-click very intuitive. Let's keep the issue with the centering open for now.

src/audiomixerboard.cpp
corrados: The Jamulus users will not like it, I guarentee you ;-). Storing of the pan will later be a must have but for a first start it is ok to not have it.

src/audiomixerboard.cpp
corrados: Do you agree with my concern?

src/audiomixerboard.cpp
corrados: I know that it is not nice to have 50 lines of code for each connection. All this code will be cleaned up as soon as I drop the support for Qt4. My plan is to drop support of Qt4 as soon as Qt6 is available for the Open Source community (which will most likely be in about a year). Until then I want to keep the code consistent. I do not want to have the Pan differently implemented than all the others. So please do it exactly like the faders are implemented.

src/server.cpp
corrados: Ok, speed optimization can be done later. That has not to be done in the first place. So it is ok to have additional multiplications for now.

src/server.cpp
corrados: Ok, but my concern is that after adding the pan feature the signal amplitude is lower than it was before that change. Usually the people will have the pan in the middle. Then they may complain that the level has changed by the new version of Jamulus.

corrados:
Ok, so now it would be good if you could implement all the changes and update the pull request. Then I can do a second code review.
Thanks.

src/server.cpp
tarmoj: I understand it now that it influences especially situation when there are two different signals on different channels (like vocals in one and guitar in other). I think the solution would be to have an enable/disable button close to channel pan slider so user can switch it out easily (by channel). If panning is out, then both channels would be multiplied with 1, ie no change. Otherwise if to use panning, the situation is different from previous experience anyway, so I would not worry as long the panning effect sound fine.

tarmoj: I will get to the changes and hope to submit new pull request next week. Thanks!

corrados: I would not add another button. The faders should have as little GUI controls as possible.
As an initial implementation we accept the level drop (actually it is not too bad, maybe it is not really noticeable).

src/audiomixerboard.cpp
```diff
@@ -661,6 +662,58 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent, Qt::WindowFlags ) :
     QObject::connect ( vecpChanFader[48], SIGNAL ( gainValueChanged ( double ) ), this, SLOT ( OnGainValueChangedCh48 ( double ) ) );
     QObject::connect ( vecpChanFader[49], SIGNAL ( gainValueChanged ( double ) ), this, SLOT ( OnGainValueChangedCh49 ( double ) ) );
 
+//Pans: ----
```
corrados, line 665: Ups... Didn't you follow the latest code on master? I decided to remove Qt4 compatibility recently and removed all this code. So the actual code which is used is the Qt5 code which is much cleaner (no code repetitions).

corrados: I just saw that you merged from master but this was an old version. Please merge again with the latest master code. (Note that the source code of Jamulus is heavily modified currently. So, unfortunately, it will require some merging to do on private branches...)

tarmoj:
Hi,

I corrected the issues as much I could from the previous discussion and implemented storing panning values. The latter needs more testing with more clients connected.

I have note worked with the appearance of the slider. I imagine it could use perhaps L and R letters to its sides but I leave it up to you to design it as you find better.

I hope that merging from went well with the current state went well and I did not mess anything up too much. For any case my forked repo is:
https://github.com/tarmoj/jamulus/tree/panning 

tarmo

src/audiomixerboard.h
```diff
@@ -94,27 +98,38 @@ class CChannelFader : public QObject
...
+    void OnPanValueChanged ( int value )  {SendPanValueToServer ( value ); }
     void OnMuteStateChanged ( int value );
 
 signals:
     void gainValueChanged ( double value );
+    void panValueChanged  (double value  );
     void soloStateChanged ( int value );
 };
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
```
corrados, line 110: This seems to be a merge issue. This check and the #else branch should be deleted.

src/audiomixerboard.h
```diff
@@ -139,33 +154,164 @@ class CAudioMixerBoard :
...
+                              const double dValue );
+
     CVector<CChannelFader*> vecpChanFader;
     QGroupBox*              pGroupBox;
     QHBoxLayout*            pMainLayout;
     bool                    bDisplayChannelLevels;
     bool                    bNoFaderVisible;
     QString                 strServerName;
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
```
corrados, line 182: Same as above, this is a merge error.

src/audiomixerboard.h
```diff
@@ -139,33 +154,164 @@ class CAudioMixerBoard :
...
+    void UpdatePanValue ( const int    iChannelIdx,
+                           const double dValue );
+
+
+#endif
+
+public slots:
+    void OnPanValueChanged ( const double dValue );
+
+#if QT_VERSION < 0x50000  // MOC does not expand macros in Qt 4, so we cannot use QT_VERSION_CHECK(5, 0, 0)
```
corrados, line 204: All of this is gone on the master.

src/channel.cpp
```diff
@@ -28,6 +28,7 @@
 // CChannel implementation *****************************************************
 CChannel::CChannel ( const bool bNIsServer ) :
     vecdGains              ( MAX_NUM_CHANNELS, 1.0 ),
+	vecdPannings		   ( MAX_NUM_CHANNELS, 0.5),
```
corrados, line 31: You seem to work with tabs. In the Jamulus source code, I replace tabs with spaces everywhere.

src/protocol.h
```diff
@@ -96,7 +97,8 @@ class CProtocol : public QObject
     void CreateJitBufMes ( const int iJitBufSize );
     void CreateReqJitBufMes();
     void CreateChanGainMes ( const int iChanID, const double dGain );
-    void CreateConClientListMes ( const CVector<CChannelInfo>& vecChanInfo );
+    void CreateChanPanMes ( const int iChanID, const double dPan );
+	void CreateConClientListMes ( const CVector<CChannelInfo>& vecChanInfo );
```
corrados, line 101: Same as above, it is a tab but should be spaces.

src/audiomixerboard.cpp
```diff
@@ -119,6 +128,10 @@ CChannelFader::CChannelFader ( QWidget*     pNW,
     pFader->setAccessibleName ( tr ( "Local mix level setting of the current audio "
         "channel at the server" ) );
 
+    pPan->setWhatsThis ( tr ( "<b>Panning:</b> Sets the panning position from Left to Right of the channel. "
```
corrados, line 131: I have changed this recently. I do not want to have the HTML codes in the translation (see, e.g, the "Mute" help text as a reference).

src/audiomixerboard.cpp
```diff
@@ -560,7 +596,169 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent, Qt::WindowFlags ) :
 
 
     // Connections -------------------------------------------------------------
+
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
     connectFaderSignalsToMixerBoardSlots<MAX_NUM_CHANNELS>();
+
+#else
```
corrados, line 603: Merge issue

corrados:
> I corrected the issues as much I could from the previous discussion and implemented storing panning values. The latter needs more testing with more clients connected.

Thanks for your fixes. I just put some more comments in the code.

> I have note worked with the appearance of the slider. I imagine it could use perhaps L and R letters to its sides but I leave it up to you to design it as you find better.

Thats fine. I am a fan of "keep it simple". Let's do the fine-tuning later.

> I hope that merging from went well with the current state went well and I did not mess anything up too much.

I found some merge issues and marked it in your code. Can you please try to fix these?

One open issue is the following: Your new Pan features needs support from the servers. There are a lot of servers out there which have an old Jamulus version and will not update to a new version. So there will be a lot of servers where we show pan sliders in the Jamulus client but they will have no effect. Users will complain about this, I'm 100 % sure.

With the level meters we had a similar problem. But for these it is a bit easier since we get feedback from the server. I implemented a fix in the Jamulus client that the meters are not shown if the server does not support them.

tarmoj:
Hi,

Yes, I had a confusion with the merge -  last time you were quite specific that the Qt4 support must be there so I did not believe it was removed and added the Qt4 specific sections back :)
Now fixed.

Sorry for the tabulation -  I started with different QtCreator settings and did not notice to correct some of them.

Working with older servers -  very good point, there should be a way to know if server supports panning or not. I am sure as that you can implement better than I would (and it is quite complex code base to get a clear grasp of it... ) so I better will not dwelve into it. What I thought -  to add a messagse to protocol like PROTMESSID_REQ_SERVER_VERSION and  PROTMESSID_SERVER_VERSION - so the client asks for server version in the connection. If sufficient version is not returned, disable panning. The same should happen when the mode is MONO-MONO (ie no panning shown).

I am afraid I cannot put much more time into this. I truly hope that this code can be of any use and you can continue with it at a convenient moment.

tarmo

corrados:
> I am sure as that you can implement better than I would [...] to add a messagse to protocol like PROTMESSID_REQ_SERVER_VERSION and PROTMESSID_SERVER_VERSION - so the client asks for server version in the connection

That is a good idea. I will implement that.

> I am afraid I cannot put much more time into this. I truly hope that this code can be of any use and you can continue with it at a convenient moment.

That's fine. You have already spend some time on it. And I gave you a hard time, sorry for that :-).

I'll take over from here. Thank you for your contribution.

PR #147, opened 2020-04-26
Add some logging to the central server

pljones:
Mostly for debugging but servers now log when they're making the requests, so having the central server log when it's handling them seems "fair" :).

### DIFF
--- a/src/serverlist.h
+++ b/src/serverlist.h
@@ -207,6 +207,8 @@ protected:
     // count of registration retries
     int                     iSvrRegRetries;
 
+    QTextStream& tsConsoleStream;
+
 public slots:
     void OnTimerPollList();
     void OnTimerPingServerInList();
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -36,7 +36,8 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
       bCentServPingServerInList ( bNCentServPingServerInList ),
       pConnLessProtocol         ( pNConLProt ),
       eSvrRegStatus             ( SRS_UNREGISTERED ),
-      iSvrRegRetries            ( 0 )
+      iSvrRegRetries            ( 0 ),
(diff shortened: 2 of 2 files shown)


corrados:
With "logging" you mean on the command line, right? You are not talking about the logging file which stores all the connections to the server.

pljones:
Yep - just to stdout.

src/serverlist.h
```diff
@@ -207,6 +207,8 @@ class CServerListManager : public QObject
     // count of registration retries
     int                     iSvrRegRetries;
 
+    QTextStream& tsConsoleStream = *( ( new ConsoleWriterFactory() )->get() );
```
corrados, line 210: Shouldn't we do the initialization in the constructor? What you implemented is C++ 11, right. I am not so familiar with this construction methods. Does this has advantages over initialization in the constructor?

pljones: Oh.  I'm not sure -- I don't mind.

corrados: I would prefer to have it here: https://github.com/corrados/jamulus/blob/master/src/serverlist.cpp#L34
What do you think?

pljones: OK, rebased and initialised in the constructor...

Plus moved the code _outside_ the Mutex locks, which is where I'd originally put it, before removing it for a different branch and adding it back, here in the wrong place.

I've only tested register and unregister, not the expiry.

corrados:
> Plus moved the code outside the Mutex locks, which is where I'd originally put it, before removing it for a different branch and adding it back, here in the wrong place.

Do you think the mutex is an issue in this context?

pljones:
> Do you think the mutex is an issue in this context?

I moved the mutex so it's only done inside the if as it's only needed then.  It seem likely it's using CPU cycles to handle the lock and then more CPU cycles to release the lock if the condition isn't met.  (In fact, as the method should only be called if the condition is met, it's questionable why it's there but good to leave it.)

The mutex certainly prevented the output appearing, though.

pljones:
Rebased off latest master.

PR #162, opened 2020-04-30
--clientname command line option lets you set the Jamulus window title and jack client name.

newlaurent62:
--clientname command line option lets you set the Jamulus window title and jack client name.
Use example : 
Jamulus --clientname "Jamulus Paula"

if the --clientname is not set, APP_NAME is used.

Jamulus --help added the --clientname description.

I tested only Linux binaries.

### DIFF
--- a/android/sound.cpp
+++ b/android/sound.cpp
@@ -26,11 +26,12 @@
 
 
 /* Implementation *************************************************************/
-CSound::CSound ( void       (*fpNewProcessCallback) ( CVector<short>& psData, void* arg ),
-                 void*      arg,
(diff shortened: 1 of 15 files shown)


corrados:
Ok, thanks.

WolfganP:
Based on this request https://github.com/corrados/jamulus/issues/412 and the current implementation, wouldn't it useful to expand this feature to temporarily overwrite the Client Name for the session when the client is launched as well? (as in profile and reported to the server)

Use cases in sight: launch of multiple clients via shell scripts for load testing or automation of a streaming setup.

Toughts?

PR #166, opened 2020-05-01
Command line help sections 

newlaurent62:
I added three sections:
- Server only, 
- Client only 
- General purpose 
in the help for command line options.


corrados:
Thanks for your code. You branch is too far away from the master branch. I implemented it now on the master branch so we do not need this pull request anymore.

PR #175, opened 2020-05-05
Add Portuguese Translation

Snayler:
I finished the Portuguese translation.

I noticed some things are missing from the translation files, as follows:

**Main Window:**
- Disconnect Button

**About Window:**
- Text with list of libraries, resources and code snippets
- Contributors text

**Settings, My Profile and Server GUI:**
- Most dropdown menu items (except the instrument list, those were included)

I basically copied one of the other files, changed the language attribute to `"pt_PT"` and proceeded the translation with Qt Linguist.

I also copied the .qm file, was unsure of it's purpose but translated it anyway.


corrados:
Thank you.

corrados:
> Disconnect Button

I will fix that in my next commit.

> About Window:

Will be changed in the near future.

> Settings, My Profile and Server GUI:

There were a lot of strings missing. I try to fix them in the next Git commit.

corrados:
I did a lot of changes in the translation strings now. Please check if the changes to your translations are ok. I tried to remove as many HTML code as possible. In that process I also removed some <i> from your code. If you really want to have them, you have to re-introduce them but I think having less of these is better.
I also added missing strings. So you will find some new "unfinshed" texts in the translation file.

corrados:
Regarding the About dialog: My plan is to start the changes this weekend.

PR #180, opened 2020-05-06
deb packaging: build translations at build time

trebmuh:
This allows to build the translation right at build time, then ensuring that the qm files are always up-to-date with the ts files.

### DIFF
--- a/distributions/debian/control
+++ b/distributions/debian/control
@@ -11,6 +11,7 @@ Build-Depends:
  make,
  qt5-default,
  qtdeclarative5-dev,
+ qttools5-dev-tools,
 Standards-Version: 3.9.5
 Homepage: http://llcon.sourceforge.net/
 Vcs-Git: git://github.com/corrados/jamulus.git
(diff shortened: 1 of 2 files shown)


corrados:
Thanks for your contribution. But my plan is to update the .qm files manually everytime when somebody had changed the .ts file. So the .qm files in the Git repo should always be up-to-date.

A better solution would be this: https://doc.qt.io/qt-5/qmake-variable-reference.html#translations
"_You can use CONFIG += lrelease to automatically compile the files during the build_"
I tried that out but it did not work for me. Are you interested in trying to get this working? I think that would be a better solution than to do the update in the debian scripts, do you agree?

trebmuh:
> Are you interested in trying to get this working?

I'm not a coder, hence haven't such knowledge.

> I think that would be a better solution than to do the update in the debian scripts, do you agree?

In a ideal world yes. So far, better to have it in the debian scripts rather than not at all. Isn't it?

corrados:
Yes, it does not hurt. Thank you.

PR #183, opened 2020-05-07
Update pt_PT Translation

Snayler:
Updated the pt_PT translation with the missing values

Already released the translation in QT Linguist, so I believe the qm file is also updated.


corrados:
Wow, you are quick. Thank you!

PR #191, opened 2020-05-09
Updated Spanish translation and translator name

ignotus666:


corrados:
Thanks for the update.

PR #195, opened 2020-05-10
Handle server application closure more cleanly

pljones:
OK, I've had this running for a while and it seems to be working okay here.

I'd like to get it merged as I'm starting another change based off it...

### DIFF
--- a/src/recorder/jamrecorder.cpp
+++ b/src/recorder/jamrecorder.cpp
@@ -314,21 +314,30 @@ void CJamRecorder::Init( const CServer* server,
         throw std::runtime_error( (recordBaseDir.absolutePath() + " is a directory but cannot be written to").toStdString() );
     }
 
-    QObject::connect((const QObject *)server, SIGNAL ( Stopped() ),
-                     this, SLOT( OnEnd() ),
-                     Qt::ConnectionType::QueuedConnection);
+    QObject::connect( (const QObject *)server, SIGNAL ( Stopped() ),
+                      this, SLOT( OnEnd() ),
+                      Qt::ConnectionType::QueuedConnection );
 
-    QObject::connect((const QObject *)server, SIGNAL ( ClientDisconnected(int) ),
-                     this, SLOT( OnDisconnected(int) ),
-                     Qt::ConnectionType::QueuedConnection);
+    QObject::connect( (const QObject *)server, SIGNAL ( ClientDisconnected ( int ) ),
+                      this, SLOT( OnDisconnected ( int ) ),
+                      Qt::ConnectionType::QueuedConnection );
 
     qRegisterMetaType<CVector<int16_t>>();
-    QObject::connect((const QObject *)server, SIGNAL ( AudioFrame(const int, const QString, const CHostAddress, const int, const CVector<int16_t>) ),
-                     this, SLOT(  OnFrame(const int, const QString, const CHostAddress, const int, const CVector<int16_t>) ),
-                     Qt::ConnectionType::QueuedConnection);
+    QObject::connect( (const QObject *)server, SIGNAL ( AudioFrame( const int, const QString, const CHostAddress, const int, const CVector<int16_t> ) ),
+                      this, SLOT(  OnFrame (const int, const QString, const CHostAddress, const int, const CVector<int16_t> ) ),
+                      Qt::ConnectionType::QueuedConnection );
+
+    QObject::connect( QCoreApplication::instance(),
+                      SIGNAL ( aboutToQuit() ),
+                      this, SLOT( OnAboutToQuit() ) );
 
     iServerFrameSizeSamples = _iServerFrameSizeSamples;
+
+    thisThread = new QThread();
+    moveToThread ( thisThread );
+    thisThread->start();
 }
+
 
 /**
  * @brief CJamRecorder::OnStart Start up tasks when the first client connects
@@ -337,7 +346,7 @@ void CJamRecorder::OnStart() {
     // Ensure any previous cleaning up has been done.
     OnEnd();
 
-    currentSession = new CJamSession(recordBaseDir);
+    currentSession = new CJamSession( recordBaseDir );
     isRecording = true;
 }
 
@@ -346,7 +355,7 @@ void CJamRecorder::OnStart() {
  */
 void CJamRecorder::OnEnd()
 {
-    if (isRecording)
+    if ( isRecording )
     {
         isRecording = false;
         currentSession->End();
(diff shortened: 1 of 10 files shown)


src/server.cpp
```diff
@@ -908,6 +915,54 @@ void CServer::OnCLDisconnection ( CHostAddress InetAddr )
...
+void CServer::CleanShutdown() {
+    Mutex.lock();
+    {
+        bool oldDAC = bDisconnectAllClients;
+
+        // This is to prevent new connections
+        bDisconnectAllClients = true;
+
+        // This is to disconnect all connected channels
+        for ( int i = 0; i < iMaxNumChannels; i++ )
```
corrados, line 949: This is critical because of two reasons:
I sometimes have to restart the Central Server even if musicians are connected to the server. Since the server restarts very quickly, the musicians only have a short period of time when they do not hear each other but after that they can continue their jam as if nothing happend. If you initiate a disconnect, all musicians would have to re-connect. I prefer the current behavior.
Second issue: The disconnect message was introduce by me just to disconnect clients which are connected over a long period of time to my server eating up my bandwidth. With the disconnect functionality I could restart the server with the --disconnectall and then I could restart the normal server operation and the client connection was gone. But the way the disconnect message from the server is treated in the client is not good right now. It was only intended for an "emergency case" and not for normal operation.

src/serverdlg.cpp
```diff
@@ -342,18 +342,6 @@ lvwClients->setMinimumHeight ( 140 );
     Timer.start ( GUI_CONTRL_UPDATE_TIME );
 }
 
-void CServerDlg::closeEvent ( QCloseEvent* Event )
```
corrados, line 345: The closeEvent should also be removed in the header.

src/server.cpp
```diff
@@ -466,6 +465,14 @@ CServer::CServer ( const int          iNewMaxNumChan,
        SIGNAL ( SvrRegStatusChanged() ),
        this, SLOT ( OnSvrRegStatusChanged() ) );
 
+    QObject::connect ( QCoreApplication::instance(),
```
corrados, line 468: Is this called when the server GUI is closed?

src/server.h
```diff
@@ -257,6 +258,8 @@ class CServer :
     int GetNumberOfConnectedClients();
     CVector<CChannelInfo> CreateChannelList();
 
+    CSignalHandler* pSignalHandler = CSignalHandler::getSingletonP();
```
corrados, line 261: For consistency, can you move the initialization in the constructor?

src/signalhandler.cpp
```diff
@@ -0,0 +1,146 @@
+/******************************************************************************\
```
corrados, line 1: General comment to this file: It would be good if you could add some references from where you have all the information how to handle the signals. I guess you have used a lot of Qt help pages. Maybe other sources. It would be beneficial if you could add all these links in this file.

src/signalhandler.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+BOOL WINAPI CSignalWin::signalHandler ( _In_ DWORD sigNum )
+{
+    auto self = getSelf<CSignalWin>();
+    QReadLocker lock ( self->getLock() );
+    return self->pSignalHandler->emitSignal ( static_cast<int>( sigNum ) );
+}
+
+#else
+int CSignalUnix::socketPair[2];
```
corrados, line 98: Isn't this already declared in the header file?

src/server.cpp
pljones: Mmm, yes.  I see the conflict.

I don't actually need the client disconnect here.  (It didn't help the jam recorder, anyway.)  I was thinking it was "tidier" but it doesn't really let the client know what's going on anyway.  If the server has gone, the client will find out the same way whether or not disconnect is used.

src/server.cpp
pljones: Yes, tested on Linux and Windows GUI mode.

src/signalhandler.cpp
pljones: Very, very annoyingly, I've actually lost my main source.

I went to check up on it again and simply couldn't find it.  It took some tracking down originally.  Most of it comes from fragments on Sourceforge.  Windows-only and unix-only examples abound - but I found an example that did both.  Then lost it - I actually got all the main coding done in an afternoon.  Came back, cleaned everything up, went to add references...  And couldn't find it.  Checked again before raising the pull request.  Still couldn't.

Basically, it's the same solution as in many threads across Sourceforge: Windows and Unix work differently.  The one "key" bit is the metadata trigger of the Qt call back.  I'll do another round of searches.

src/signalhandler.cpp
pljones: No - it's defined there.  This is the actual socket pair instance.  Two `int`s in an array, not "an array of `int`s".

src/signalhandler.cpp
pljones: Ah, found it.  My memory is useless.  I'd actually put it somewhere for safe keeping.  I should have also written that down somewhere I couldn't miss it.

My "main" references - as it brought together several commonly available sets of ideas, was `QCtrlSignals` (https://github.com/Skycoder42/QCtrlSignals.git).  It's BSD 3-clause.

Personally, I think my limited use would count as "fair use" but I'm not prepared to push it - I'd rather give the acknowledgement.  It simplified my life.  I then took that code apart as it had too much complexity.  (Look at all those files!)

pljones:
Updated and retested.

corrados:
Thank you for the code and the adaptations.

src/recorder/jamrecorder.cpp
```diff
@@ -314,30 +314,39 @@ void CJamRecorder::Init( const CServer* server,
         throw std::runtime_error( (recordBaseDir.absolutePath() + " is a directory but cannot be written to").toStdString() );
     }
 
-    QObject::connect((const QObject *)server, SIGNAL ( Stopped() ),
-                     this, SLOT( OnEnd() ),
-                     Qt::ConnectionType::QueuedConnection);
+    QObject::connect( (const QObject *)server, SIGNAL ( Stopped() ),
+                      this, SLOT( OnEnd() ),
+                      Qt::ConnectionType::QueuedConnection );
```
corrados, line 319: I was wondering why you are using a "QueuedConnection" for your connections. Is this really required?

pljones: Possibly not here as it's the shut down.  But generally, the idea was that the main server thread just queues the request and the recorder thread reads the queue when it "gets around to it".  My understanding is that this is the most asynchronous approach to the signal/slot connections.  I could be misunderstanding it, of course! :)

corrados: Since you created a speparate thread for the recorder, the connections should automtically be a queued connection. But it does not hurt to set it explicitly here. Thanks for the clarification.

PR #200, opened 2020-05-11
Added Oboe support for android implementation (part)

sthenos:
Added Oboe support for android implementation (part).   Integration with Oboe works, however, audio rendering (playback and recording) is not quite right with audio sounding dithered many artefects, more work required on buffer copying to fix this.


sthenos:
Pull request for merge android oboe support into master.  One conflict in sound.cpp on the constructor arguments which I've resolved

corrados:
Thank you for this contribution.

WolfganP:
Thank you, great contribution to open up to mobile platforms! (not many musicians have their computers set up, so an easier to use platform may be a good intro to jamulus)

PR #216, opened 2020-05-14
Converted icon to be used as Android resource.

andreassch:


corrados:
Thanks for your contribution.
I find that you have inserted "-- %%INSERT_VERSION_NAME%% --" -> is this replaced by the version by some tool?
I find that you have added icons but I cannot see any reference to these in the code. Is that intentional?

andreassch:
This is replaced in the build process (I am using Qt Creator). The same holds for the other variables %%INSERT_APP_NAME%% etc. Actually, these variables have been inserted by Qt Creator. (This goes back to when I had to re-create AndroidManifest.xml with Qt Creator to make deploying to my phone work some weeks ago.)
The icon is set in AndroidManifest.xml with android::icon="@drawable/icon"

corrados:
Thank you for the clarification.

Issue #77, opened 2020-04-16
Allow internationalization


trebmuh:
I'd be really happy to provides the project with a French translation for the GUI if a translation mechanism was made available.

gilgongo:
Same applies to the Wiki/docs ideally. I don't think it's really feasible with Github's wiki though. Perhaps if we moved to Github Pages, but that might cause other issues.

trebmuh:
If the docs should be moved elsewhere for being translatable, then I'd encourage to make use of https://hosted.weblate.org/ (it's free software and free to use for free software). Note that, you can use the hosted.weblate to make the GUI component translation too.

gilgongo:
Interesting. How would this work in Jamulus's case? For example I notice from Weblate's [own docs](https://docs.weblate.org/en/latest/) that they appear to be hosted on Github (there's an "edit" button in the top right which takes you to what I assume is a docs repo with .rst files in it). What I don't understand is how to switch to another language (or is that being auto-detected?), or where the web app is being hosted (Is it Github pages, or perhaps some 3rd party that you can deploy to via Github?) Is this the default if you opt for weblate hosting?

trebmuh:
I don't understand very well what you wrote @gilgongo . Not a native English speaker here.

Not being a coding person neither, I can tell you how I understand it since I'm using it as a translator for several project.

- the dev(s) does ask to hosted.weblate if they're ok to host the project's translation for free (being a free software)
- weblate says OK
- the dev set up a translation mechanism (let's say the Qt translation system here)
- then the dev or/and weblate people (I don't know) make a kind of a "hook" thing which allow to copy the translation file from github to weblate, and back
- the translators translates online (on hosted.weblate) and the hook back take care of integrating things back into the github repo

I hope this helps.

If you need more information, I can maybe ask someone from a project I'm translating on weblate to talk with you on what are the details of the installation of that.

gilgongo:
@trebmuh My apologies - I was referring to how weblate might work with the localisation of our documentation, not the application itself. Does weblate help with localising a collaboratively-edited website?

trebmuh:
@gilgongo as far as I know, you just feed Weblate with files (.pot/.po file or .ts file) and translators can start working. I'm not sure with what you call a _ collaboratively-edited website_. As an example, zrythm GUI interface and website is translated using hosted.weblate.

gilgongo:
[...]

PR #221, opened 2020-05-14
Translation of qt messages

jpcima:
Hi, this change allows the Qt messages to be translated in the application.
#77


corrados:
Thank you.

trebmuh:
Nice, just rebuild from master head, and it works as expected. :+1:

Issue #71, opened 2020-04-15
Increase LED luminance in UI?


gilgongo:
With the addition of the UV meters, the issue of contrast in the UI has become more important. Suggest increasing the luminescence of the colours as below. However, in this mockup the lights are 2D, whereas in the UI they are 3D. This means I can only approximate a good luminance for them.

Here the Red is #FF3300, Green is #66FF33 and Yellow is #FFFF00 while the default is #F2F2F2

![image](https://user-images.githubusercontent.com/4561747/79324569-b700c100-7f07-11ea-85db-0468770d6969.png)

corrados:
I'll take care of it.

corrados:
I started with the green LED (see the "Delay" LED on the following screen shot):
![grafik](https://user-images.githubusercontent.com/46655886/79359925-79ba2480-7f43-11ea-83e1-1c3b419efae5.png)
How do you like it?

trebmuh:
I need to see the whole picture to be able to see if I'm seeing everything correctly. Most of the time, the color _differences_ is the key.

corrados:
What do you mean by "whole picture"? Do you mean you must see it live in the software when its running?
Here is another screen shot, now the green and red round LEDs are modified:
![grafik](https://user-images.githubusercontent.com/46655886/79367445-35805180-7f4e-11ea-9747-6377d3405ac7.png)

gilgongo:
@trebmuh I'm not colour blind, but I think I'm right in saying that we'd never be able to get good contrast between red, green and yellow for everyone on the Delay and Buffers indicator, so that's why I raised #57 The VU meters are a slightly different matter though, since the difference between those colours is only nice to have - the main utility is to see them moving.

trebmuh:
> What do you mean by "whole picture"? Do you mean you must see it live in the software when its running?

Usually yes. Or at least a picture with all the _final_ colors.

On the picture above, I've got difficulties to see the differences on the Input L&R horizontal segments which are colored and the other in gray. With a quick look at it, it seems that the 3 last of each columns are colored on green, but if I look closer, the first column could have only 2 greens horizontal segments, and the R colum probably 3 but maybe 4. Note that if I'm looking too long at it, everything gets mixed and I'm loosing ground.

trebmuh:
@gilgongo of course it is possible to get a good contrast. That's why greenlight on the street are the colors they are. If not, colorblind would be car-less :wink:
[...]

PR #222, opened 2020-05-15
Make HLED black if off

fleutot:
This is more a proof of concept, or a hidden feature request.
One of the reasons I haven't been using the fancy skin is that I find the level meter with LED disturbing. The LED that are off have too much contrast with the background, and they are too hard to distinguish from green and yellow, in my opinion.

Here is a comparison screenshot of this branch (left is this PR, right is master for reference):
jam-led-off-cmp

I made the black LED hastily, manually, in GIMP, I don't expect you'd be happy to use it directly.

Unfortunately, since the other round LED are used in the non-fancy skin, I don't think it would be a good idea to make them black when off.


corrados:
This topic is a critical one. The original "off LED" was some dark grey. But I have changed it because of: https://github.com/corrados/jamulus/issues/71 and https://github.com/corrados/jamulus/issues/57
So it is not only a design desicion but there is some more behind it.

fleutot:
Oh OK! I don't know much about color blindness, but from the tickets you linked, it seems that the issue was seeing if the LED was lit at all, and then the difference between red and green mostly.

I suppose that the most critical when it comes to VU meters is to be able to see if a LED is off or on, and second what color it has. If that is the case, a big difference in luminance between on and off (whatever the color) must be preferable?

fleutot:
Of course, we have people here who are more concerned with these matters! @gilgongo @trebmuh : Is this change bad for (your) color blindness?

corrados:
To avoid further discussions on color blindness, etc, I would like to close this pull request without merging. I think you have a workaround by using the "non fancy" skin instead. Would you agree to that?

Snayler:
I understand color blindness takes priority over user interface cosmetic issues, so how about expanding the settings to allow users to select color blindness mode? This would allow for accommodation of different color blindness types and would allow each person to specify how they want the skin to be presented.

fleutot:
@corrados: I'm ok with that! I can also use my fork ;)

fleutot:
Although it would have been nice to have feedback on the darker off-LED, from the people who've raised those concerns in the first place.

gilgongo:
>  feedback on the darker off-LED

I'm not actually colour-blind, but in principle the black LED is definitely an improvement as it's a better contrast and therefore makes the state of the meter clearer.  

In terms of the ideal solution, @Snayler suggestion would be best overall though because not only are there different types of colour blindness, there are also people with non-chromatic vision issues as well as those with just sub-optimal screens.

corrados:
Now I am a bit confused. In your Issue: https://github.com/corrados/jamulus/issues/71 you made the LEDs more white than they were before. Now you say that black gives better contrast?

gilgongo:
It isn't about the colours on their own as far as I understand (particularly in the case of colour blindness), but about the contrast between them. So if @fleutot has a better way of getting _more_ contrast, then I'm all for it.  I wasn't actually looking at the contrast between the "off" state of the LEDs and the R/G/B lights in #71, but I should have done. 

But I see that the contrast ratio between #FDFF22 and #000000 is 19.5 while the ratio between #FDFF22 and #E9E9E9 is 1.12

Good work, @fleutot  :-)

WolfganP:
Just googling, various articles with considerations on designing an UI with color blindness in mind appear (ie https://www.secretstache.com/blog/designing-for-color-blind-users/)

Many of the tips (ie textures) won't apply to the leds icons due to small size, but what about the shape beyond color selection? (ie small round dot when off, current rectangle) Maybe that "dynamic" shape isn't hard to implement and will cater better to users with visual impairments.

corrados:
> but what about the shape beyond color selection?

I would not like to change the shape. The "fancy skin" is just one skin out of two. For the normal skin we have bars instead of LEDs which is perfectly suited for color blind users.
In the future we will even have one more skin: https://github.com/corrados/jamulus/issues/73
We should have color blindness in mind for that skin from the beginning on.

corrados:
> I made the black LED hastily, manually, in GIMP, I don't expect you'd be happy to use it directly.

Ok, so if the experts say that black is better, I am fine with that decision. But I can only merge this merge request if you have the final black design ready.

gilgongo:
@WolfganP 

> shape beyond color selection? (ie small round dot when off, current rectangle) 

That was the intention behind [this ticket](https://github.com/corrados/jamulus/issues/57)

fleutot:
I see that there are large png files for the LED, which are then reduced to low resolution.

I suppose that there are original images in another format than png, and that the right way to do it would be to start from this format. I can't find these original files, help?

corrados:
Here you can find the sources: https://openclipart.org/search/?p=4&query=led
There are actually already black versions available :-).

fleutot:
Wow, look at that! I guess the down-scaling is only a matter of apply a low res in gimp or the like?

corrados:
These images are available as svg so you can scale them as you like. As far as I remember when I created the LED pixmaps long ago, I think I have used gimp.

fleutot:
Closing to re-open a new implementation: https://github.com/corrados/jamulus/pull/318

trebmuh:
> Of course, we have people here who are more concerned with these matters! @gilgongo @trebmuh : Is this change bad for (your) color blindness?

I'm seeing only right now that you asked me about feedback here 3 weeks ago. Apologize for the lag. My answer is: yes, black color for LED is fine as it brings more contrast to my eyes.

Thanks for caring about colorblind people when designing a GUI, it's much appreciated from here.

corrados:
What about making the black LEDs a bit darker? Here is an example:
grafik
What is your opinion?

trebmuh:
I prefer the "new" one. Not since the white contour isn't very nice to my eyes in the "old" one. Note that it might probably not be related to my colorblindness.

corrados:
Thanks for your feedback. Same with me, I just like it better if it is a bit darker. @fleutot and @gilgongo, what is your opinion? Should I change it in the Git code to the darker design?

gilgongo:
The darker design looks good from a contrast perspective, certainly.

fleutot:
It's better with the darker edge, I agree. I kept the edge from the original png for consistency with green/yellow/red, but usability goes first. I like the reduced reflection too.

corrados:
Thanks for your feedback. I have already changed it on the Git master.

PR #230, opened 2020-05-15
debian/watch: Adapt to the tags on github

tormodvolden:
It will download the upstream package as e.g.
jamulus-r3_5_3.tar.gz (respecting upstream versioning)
and link as jamulus_3.5.3.orig.tar.gz

Signed-off-by: Tormod Volden <debian.tormod@gmail.com>


corrados:
Thanks

PR #240, opened 2020-05-17
Change link

gilgongo:
To avoid confusion as per https://github.com/corrados/jamulus/issues/238

### DIFF
--- a/src/global.h
+++ b/src/global.h
@@ -107,7 +107,7 @@ LED bar:      lbr
 #define CENTSERV_GENERAL_NORTHAMERICA    "jamulus.fischvolk.de:22224"
 
 // getting started and software manual URL
-#define CLIENT_GETTING_STARTED_URL       "https://github.com/corrados/jamulus/wiki/Software-Manual"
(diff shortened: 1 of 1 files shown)


corrados:
Dies this means that the Help menu links oft the release Version 3.5.3 will no longer work?

gilgongo:
Should still work because 3.5.3 will be pointing at a copy of the "Getting Started" page [here](https://github.com/corrados/jamulus/wiki/Software-Manual).

corrados:
That is not good. Links in older versions should work for a long time, otherwise it is useless to implement links in the software at all. But now we have a problem that we do not have single source anymore if we keep the old page.

gilgongo:
Unfortunately, the Github wiki doesn't let you change the H1/title without altering the URL. Although I didn't mean we'd delete the old page **immediately** once 3.5.4 is out, it's a good point about the older versions. Maybe it's blessing in disguise: for anyone hitting the old page, we can show them a message encouraging them to upgrade :-)

trebmuh:
What about leaving a link pointing to the new wiki page in the old wiki page?

gilgongo:
@trebmuh That's what I'm doing, yes.

Issue #95, opened 2020-04-18
MIDI Controller


nicolopozzato:
I have read that from the 3.4.4 (2020-03-25) version there is MIDI support.

`- added support for controlling the audio mixer faders with a MIDI controller (MacOS and Linux)`

How can i map my MIDI Controller to the mixer fader, solo and mute button? 
Also there is a way to map the rever effect and the mute my input?

I don't have find anything about it. Only some thing about it in the source code...

For info i have a Korg NanoKontrol2

corrados:
Only the faders are supported.

corrados:
The MIDI control is just experimental. Not really an end-user feature right now. Right now it is implemented for the Behringer X-Touch, see https://github.com/corrados/jamulus/blob/master/src/soundbase.cpp#L289
There is some offset of 70 to support that device.

nicolopozzato:
Ok, could some sort of midi lerning interface be inserted? In order to make all devices compatible.

I imagine a settings page where you select a fader and then physically move it, in this way going to check which channel moves on MIDI you map it to the first channel of the mixer. So on for the next ones.

gilgongo:
@nicolopozzato Dunno if this is relevant, but there's something in the docs on enabling MIDI on [this page](https://github.com/corrados/jamulus/wiki/Tips,-Tricks-&-More) (scroll down, GH wiki doesn't do anchors).

nicolopozzato:
I'm working on trying to change the MIDI support...
This is a test to begin to help this incredible software!

Now channel70 is fader0 in jamulus

I was thinking about a new midi mapping...
1-20 General Purpuse command example:
```
1 -> Mute Stream (button)
2 -> Reverb Slider
```

Then group of 3 midi channel for every Jamulus Channel, so:
```
21 -> Channel Fader 0
22 -> Mute Button 0
23 -> Solo Button 0
24 -> Channel Fader 1
25 -> Mute Button 0
. . . . . . .
```

In this way everyone can map the note nedded in their MIDI Controller... (my friend has a Korg NanoKontroller2)

Does anyone know a midi program for linux that simulate midi presses? I don't have a midi controller, and my friend has a mac and we can't compile in his pc because we get error after following the guide in the wiki.

If anyone have idea how to integrate the midi support also for windows would be very nice!

dcocharro:
Would be awesome to have MIDI support to control faders, solos and mutes integrated with a user friendly interface .

[...]

PR #241, opened 2020-05-17
Feature/show channel numbers

tanjeff:
When controlling the mixer strip levels via MIDI, the channel number of the strip must be known. But when people leave the jamulus server, their channel numbers become unused and are reused for people join thereafter. Thus, the channel number for each strip is not obvious to the user.

I added a visible channel number to the mixer. This eases the use of a MIDI controller.

I kindly ask you to pull my change. However, if you are not satisfied, please let me know. I will try to update my code accordingly.

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -148,9 +148,9 @@ CChannelFader::CChannelFader ( QWidget*     pNW,
     plblCountryFlag->setWhatsThis ( strFaderText );
     plblCountryFlag->setAccessibleName ( tr ( "Mixer channel country flag" ) );
 
-    QString strChannelNumberText = tr ( "<b>Channel Number:</b> "
-        "Control this channel by sending MIDI CC messages with control "
-        "number 70+channelNumber." );
+    QString strChannelNumberText = "<b>" + tr ( "Channel Number" ) + ":</b> " +
+         tr ("Control this channel by sending MIDI CC messages with control "
+             "number 70+channelNumber." );
 
     plblChannelNumber->setWhatsThis ( strChannelNumberText );
     plblChannelNumber->setAccessibleName (
@@ -528,7 +528,7 @@ void CChannelFader::SetChannelInfos ( const CChannelInfo& cChanInfo )
     plblCountryFlag->setToolTip   ( strToolTip );
     plblInstrument->setToolTip    ( strToolTip );
     plblLabel->setToolTip         ( strToolTip );
-    plblChannelNumber->setToolTip ( "Channel Number" );
+    plblChannelNumber->setToolTip ( tr ("Channel Number") );
 }
 
 double CChannelFader::CalcFaderGain ( const int value )
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -49,6 +49,8 @@ CChannelFader::CChannelFader ( QWidget*     pNW,
     plblInstrument              = new QLabel            ( pFrame );
     plblCountryFlag             = new QLabel            ( pFrame );
 
+    plblChannelNumber           = new QLabel            ( "" );
+
     QVBoxLayout* pMainGrid      = new QVBoxLayout       ( pFrame );
     QHBoxLayout* pLevelsGrid    = new QHBoxLayout       ( pLevelsBox );
     QVBoxLayout* pMuteSoloGrid  = new QVBoxLayout       ( pMuteSoloBox );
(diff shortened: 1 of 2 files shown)


src/audiomixerboard.cpp
```diff
@@ -495,9 +525,10 @@ void CChannelFader::SetChannelInfos ( const CChannelInfo& cChanInfo )
         strToolTip.prepend ( "<h3>" + tr ( "Musician Profile" ) + "</h3>" );
     }
 
-    plblCountryFlag->setToolTip ( strToolTip );
-    plblInstrument->setToolTip  ( strToolTip );
-    plblLabel->setToolTip       ( strToolTip );
+    plblCountryFlag->setToolTip   ( strToolTip );
+    plblInstrument->setToolTip    ( strToolTip );
+    plblLabel->setToolTip         ( strToolTip );
+    plblChannelNumber->setToolTip ( "Channel Number" );
```
pljones, line 531: Needs to be translatable.

src/audiomixerboard.cpp
```diff
@@ -140,6 +148,13 @@ CChannelFader::CChannelFader ( QWidget*     pNW,
     plblCountryFlag->setWhatsThis ( strFaderText );
     plblCountryFlag->setAccessibleName ( tr ( "Mixer channel country flag" ) );
 
+    QString strChannelNumberText = tr ( "<b>Channel Number:</b> "
```
pljones, line 151: HTML code shouldn't be translated.  See, for example, `strFaderText`.

pljones:
Overall, I'd suggest this is only shown for people who are actually using MIDI automation.  Others musicians won't want it.

tanjeff:
> Overall, I'd suggest this is only shown for people who are actually using MIDI automation.

Agreed, but this involves a little more work. I'll try to do this soon.

I can implement it in the following ways:
* Bind the functionality to the --ctrlmidich argument, because without it, no MIDI control is possible anyway.
* Add another cmdline parameter
* Add an option to the settings dialog

I prefer the first one (use --ctrlmidich for it). Is this okay?

corrados:
Guys, first: We have a conflicting pull request: https://github.com/corrados/jamulus/pull/145
This will be worked on first and this will take some time.
Second:
> When controlling the mixer strip levels via MIDI, the channel number of the strip must be known. But when people leave the jamulus server, their channel numbers become unused and are reused for people join thereafter. Thus, the channel number for each strip is not obvious to the user.

Before you implement code it would be wise to first make the specification of your functionality public so we can discuss it before you start coding.

Is it the ideal solution to just display the channel numbers? I guess not. What MIDI controller are you using? Does it have motor faders? Maybe it would be better to implement MIDI out messages to control the MIDI hardware?

tanjeff:
You are right, there is much room for improvement around MIDI. We can discuss the topic, but I will not be able to implement the stuff anytime soon. Where is the best place to discuss this?

To me the visible Channel numbers are of great help. For the short run, they might still be a valuable improvement :)

corrados:
The ideal case would be to create an issue at https://github.com/corrados/jamulus/issues and write the specification of the new feature in there (what is the current situation, what shall the feature improve, etc.). Then we can discuss the specification before starting any coding.

corrados:
I can read in your code: "_Control this channel by sending MIDI CC messages with control number 70+channelNumber._". So you have a hardware MIDI controller with faders which have a MIDI CC message assigned. Now there are clients connected to the server but you do not know the server channel ID which is assigned to these faders. So if you want to control a certain fader for one client, you first have to look up the ID in Jamulus, then you have to find the fader using the corresponding MIDI CC message. Hmm, does not sound like a very user-friendly solution. Ok, the situation right now is that you do not know the ID so you have to try it out to find the correct fader.

First, we should make some brainstorming what could be a good (and simple) solution to the problem.

So if you, e.g., have 3 clients connected and client 1 disconnects so that in the Jamulus mixer panel, the two others are now the first and second fader where they were before the second and third. So wouldn't it be better that now you also control these two clients with your first two faders on your MIDI hardware? But the problem with that would be that the fader setting in your MIDI hardware would not reflect the actual level of that client anymore. You would have to move the fader to assign a new value and from then you are consistent again. Hmm, also not very nice.

Any other ideas?

corrados:
One more comment: Even if we use the fixed IDs from the server, it can be that in the above example you change the fader level of client 1, then client 1 disconnects and another client enters the server. In that case your fader position of the external MIDI device will also be incorrect. So I guess a clean solution can only be to have motor faders which are controlled by Jamulus.

tanjeff:
There is already an issue on that topic: #95. I suggest to continue discussion there.

corrados:
@tanjeff I just implemented to show the channel IDs in case the MIDI controller is used:
grafik
So you get the IDs prepend on the channel name.
The code is applied to the Git master. Here is the commit: https://github.com/corrados/jamulus/commit/33073658c6c277b71cc2a0dd46765e1bed03c35d
It you like you could try out the code and give feedback if it works ok for you.

tanjeff:
@corrados The solution works for me. However, I would prefer the number shown in a separate row. Example:
Screenshot_jamulus_2020-06-21_16-01-35
This would increase the readability of the number (especially when the screen is a bit distant), and the number wouldn't steal space from the name.

PR #247, opened 2020-05-18
optional INSTALL_PATH qmake flag (on unix)

seb314:
I packaged Jamulus for nix/nixos and it would
be neat if there was an option to specify a
path at which to install the binary. I'm not
really familiar with qmake -- is this a good
way to achieve this?

(nixpkgs pull request: https://github.com/NixOS/nixpkgs/pull/87914)

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -302,11 +302,15 @@ DISTFILES_OBOE += libs/oboe/AUTHORS \
         android/sound.h \
         android/sound.cpp
 
-    # could this also be done for all platforms?
-    defined(INSTALL_PATH, var) {
-        INSTALLS += target
-        target.path = $$INSTALL_PATH
+    isEmpty(PREFIX) {
+        PREFIX = /usr/local
     }
+    isEmpty(BINDIR) {
+        BINDIR = bin
+    }
+    BINDIR = $$absolute_path($$BINDIR, $$PREFIX)
+    INSTALLS += target
+    target.path = $$BINDIR
 }
 
 RCC_DIR = src/res
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -301,6 +301,12 @@ DISTFILES_OBOE += libs/oboe/AUTHORS \
         android/AndroidManifest.xml \
         android/sound.h \
         android/sound.cpp
+
+    # could this also be done for all platforms?
+    defined(INSTALL_PATH, var) {
+        INSTALLS += target
+        target.path = $$INSTALL_PATH
+    }
 }
 
 RCC_DIR = src/res


jtojnar:
Maybe something like this might be better:

```qmake
isEmpty(PREFIX) {
 PREFIX = /usr/local
}
isEmpty(BINDIR) {
 BINDIR = bin
}
BINDIR = $$absolute_path($$BINDIR, $$PREFIX)

target.path = $$BINDIR
```

Nixpkgs qmake setup hook already passes `PREFIX` to `qmake`.

corrados:
Thanks for your contributions. What shall I add now? The proposal from jtojnar seems to be more suited, right? So if you agree, I would just add these code lines to the pro-file.

seb314:
> Thanks for your contributions. What shall I add now? The proposal from jtojnar seems to be more suited, right? So if you agree, I would just add these code lines to the pro-file.

That would be great! 
Updated the PR with jtojnar's version. (I assume the `INSTALLS += target` line is still required to enable the install target, right?)

corrados:
> I assume the INSTALLS += target line is still required to enable the install target, right?

Lets clarify that first. @jtojnar Can you please comment?

jtojnar:
I am not familiar with qmake at all but I would expect it to be required.

Jamulus.pro
```diff
@@ -301,6 +301,16 @@ DISTFILES_OBOE += libs/oboe/AUTHORS \
...
+
+    isEmpty(PREFIX) {
+        PREFIX = /usr/local
+    }
+    isEmpty(BINDIR) {
+        BINDIR = bin
+    }
+    BINDIR = $$absolute_path($$BINDIR, $$PREFIX)
+    INSTALLS += target
+    target.path = $$BINDIR
```
jtojnar, line 313: This will likely need to be swapped:

```suggestion
    target.path = $$BINDIR
    INSTALLS += target
```

Jamulus.pro
```diff
@@ -301,6 +301,16 @@ DISTFILES_OBOE += libs/oboe/AUTHORS \
...
         android/sound.h \
         android/sound.cpp
+
+    isEmpty(PREFIX) {
+        PREFIX = /usr/local
+    }
+    isEmpty(BINDIR) {
+        BINDIR = bin
+    }
+    BINDIR = $$absolute_path($$BINDIR, $$PREFIX)
```
jtojnar, line 311: I would move this to the top of unix conditional. It will be useful when we decide to install other files like `.desktop` entry or icon.

Jamulus.pro
jtojnar: Not sure if this is valid syntax but it might also be nicer to use more descriptive target name:

```suggestion
    jamulus_bin_target.path = $$BINDIR
    INSTALLS += jamulus_bin_target
```

seb314: Build & install worked when I checked, but I'd be surprised if swapping them broke it

Jamulus.pro
jtojnar: Oh, these are actually available in the repo and AUR package installs them:

https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=jamulus#n26

```qmake
isEmpty(DATADIR) {
    DATADIR = share
}
DATADIR = $$absolute_path($$DATADIR, $$PREFIX)
ICONDIR = $$DATADIR/icons/hicolor/48x48/apps # https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#install_icons
DESKTOPDIR = $$DATADIR/applications
```

We should probably do that too.

Jamulus.pro
jtojnar: Reading https://doc.qt.io/qt-5/qmake-advanced-usage.html#installing-files, `target` seems to be a magic name that will choose something in unspecified manner. For the other installed files, you will need to specify `.files`.

tormodvolden:
Isn't this what DESTDIR is for? https://cmake.org/cmake/help/latest/envvar/DESTDIR.html

I believe this is how the Debian package build installs to a temporary directory before copying everything into package files.

jtojnar:
@tormodvolden No, `DESTDIR` only serves to change the temporary installation root, it does not affect the directory structure or final destination where things will be installed. For example, Debian would set `DESTDIR=/tmp/build-foo` and `PREFIX=/usr`. The separately overridable `BINDIR` is a GNU convention (see https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html).

Also this project does not use CMake but qmake, and does not install any files at all (there is no `install` make target). This PR attempts to fix that.

corrados:
Ok, so I take it that the change in that current pull request is the final code to be added? So I will merge it now. If anything is open or shall be changed, just create a new pull request.

jtojnar:
We would still want to add installation of icon and desktop files as described above: https://github.com/corrados/jamulus/pull/247#discussion_r428560108

seb314:
> We would still want to add installation of icon and desktop files as described above: [#247 (comment)](https://github.com/corrados/jamulus/pull/247#discussion_r428560108)

I tried those extra changes ([here](https://github.com/seb314/jamulus/tree/submit/install_add_icon_and_desktop)) but (at least when building with nix) there are some issues that make it not straighforward:
- the existing .desktop file assumes a lowercase binary "jamulus", so the .desktop install target will break for everyone who doesn't use `CONFIG+=noupcasename`
- the icon doesn't work (at least not the gnome I tested it with `desktopManager.gnome3.enable = true;` in nixos). Tried copying `fronticon.png` to `jamulus.png`, but no success so far. The AUR package does `install -Dm644 src/res/fronticon.png $pkgdir/usr/share/pixmaps/jamulus.png`, didn't try that yet for the nix package.

Anyway, the install target for the binary works fine with changes that are already merged

jtojnar:
Looking at https://github.com/seb314/jamulus/blob/submit/install_add_icon_and_desktop/distributions/debian/rules, we might want to install the icon from distributions instead. And also the service file, thought it would require more changes. But changing

https://github.com/seb314/jamulus/blob/f0d4e0a4f28b1b769c17bcd20ff899eba4c9c1bf/distributions/jamulus-server.service#L11

to something like

```
ExecStart=/usr/bin/jamulus -s -n --servername %u -l %L/jamulus -e jamulus.fischvolk.de -g -o "%u;;"
```

might be more useful anyway. See <https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=>, <https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines> and <https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Specifiers> for some info.

Ideally the desktop and service file would contain `@BIN_PATH@` placeholder that would be substituted based on the installation path but I was not able to find a way to do that in qmake. (Autotools have [`AC_CONFIG_FILES`](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Configuration-Files.html#Configuration-Files), CMake has [`configure_file`](https://cmake.org/cmake/help/latest/command/configure_file.html) and Meson has [`configure_file`](https://mesonbuild.com/Reference-manual.html#configure_file).)

Maybe it would be fine to port the project to a different build system as qmake will apparently not be supported by Qt 6.

jtojnar:
Just [found](https://martin.rpdev.net/2017/05/10/creating-config-header-files-using-qmake.html) `QMAKE_SUBSTITUTES`. Of course it is not even mentioned in the qmake documentation.

corrados:
> Maybe it would be fine to port the project to a different build system as qmake will apparently not be supported by Qt 6.

To my knowledge qmake will still be supported by Qt6 but cmake is the preferred tool to use. But I may be wrong. I also read that Qt will create a script to convert qmake files to cmake files. Let's wait until this script is ready and tested. Then we can try out to convert the Jamulus qmake file.

PR #254, opened 2020-05-21
Corrections to source language

ignotus666:


src/res/translation/translation_fr_FR.ts
```diff
@@ -663,7 +663,7 @@
     </message>
     <message>
         <location filename="../../clientdlg.cpp" line="172"/>
-        <source>The sound card buffer delay (buffer size) is set to a too small value.</source>
+        <source>The sound card buffer delay (buffer size) is set to too small a value.</source>
```
trebmuh, line 666: You're 100% sure here? It seems very weird to me.

ignotus666: Yes.

ignotus666: I just realised I also have to make those changes to quite a few files... ugh!

trebmuh: > Yes.

You're right.

ignotus666: That would also be correct but I don't understand your problem with how I rephrased it. I don't think a native speaker would find anything strange about it.

ignotus666: BTW I'm going to close this pull request and make a new one with all the relevant files corrected.

trebmuh: @ignotus666 : because it's sort of "_high level language_" and probably _"the sound card's buffer is too small - increasing it's size would fix the problem_" is the way usually people will speak about it. For a translator, it's usually easier to read from a "_common language_" text rather than a "_high level language_" one.

streaps:
I have to agree it sounds very strange to me (as a non-native speaker).

Keep in mind that there are many people who use English as the default language on their computer even if it's not their native language. Just because translations are often a mess and it's easier to search the web and ask for help in English.

streaps:
btw, it's usually called "buffer size" (number of sample frames) and "latency" (time in ms).

ignotus666:
It says buffer size there in brackets. And it could be argued that 'delay' is appropriate as its the time the audio data is 'held up' while the buffer gets filled, but we're getting into nitpicking territory here... Anyway, as I've said elsewhere my intention was just to correct errors, not change the style; that may happen in the future.

If a sentence is incorrect - it's incorrect. I find it bizarre that anyone should suggest that a syntactically incorrect sentence should be left as-is because non-native speakers might (or might not) understand it better. It just doesn't make sense.

trebmuh:
>  I find it bizarre that anyone should suggest that a syntactically incorrect sentence should be left as-is because non-native speakers might (or might not) understand it better. It just doesn't make sense.*

I didn't say so, I said, because the correct English sentence is less understandable for non-natives, why not change the sentence to be understood by everyone? And I've made a proposition.

ignotus666:
Sorry, I should have said who I was replying to; I know what you said trebmuh, it was in response to streaps.

To be clear: I have nothing against rewriting text to make it more understandable to a wider audience.

trebmuh:
OK

streaps:
@ignotus666 I didn't mean we should use a grammatically incorrect sentence or that I don't appreciate your work, but for some reason the sentence still looks wrong to my eyes.

streaps:
Using the terms that everyone else uses is not nitpicking. A "delay" is usually used for an "effect" (as in audio processing). You don't set a "buffer delay" on the sound card, you set the "buffer size". The latency depends on the sample rate of the audio device.

"buffer delay" is just the wrong term in this context. Please feel free to try google for "buffer delay" and "buffer size"

PR #256, opened 2020-05-21
Corrections to source language

ignotus666:


trebmuh:
As explained [here](https://github.com/corrados/jamulus/pull/254#pullrequestreview-416118604), please keep it simple English if there is no need to go to a higher level language.

ignotus666:
"Higher level language"? Sorry, but that's a bit of a stretch... The article 'a' was simply in the wrong place and I moved it to where it would be syntactically correct. I find it hard to believe I magically transformed the sentence into some 'higher level language' that is harder to understand than the original ‒I repeat, **incorrect**‒ wording simply by moving an article. There are many, many other sentences in this software that, while grammatically/syntactically correct, need to be read several times to get your head around them. 
Still, I'm looking at it from a native speaker's perspective and if you think a simplified version of the sentence would make life easier for non-native English speaking translators, I have no problem with changing it. I was just trying to make my changes as non-invasive as possible to avoid completely re-writing sentences. If I were to start being picky I could have gone much further.

corrados:
Thank you for your efforts. The texts are mainly from me and I am definitely not a native speaker.
As you might know I did the German translation and during that task I found out that some sentences are really confusing (as you also noticed) and also that I often use the "APP_NAME" and the text which can make the translation harder. I was considering removing some of them but that would have influence on all translations and is a lot of work (as you noticed in this pull request, a change in the english source has influence on all other translations...). Please also keep in mind, that the manual at:
https://github.com/corrados/jamulus/blob/master/src/res/homepage/manual.md
has very similar text than what we have in the code as the Whats This help. So if we do changed in the code, we should also consider this file.

Maybe a native speaker should review all the texts. Maybe we should change some texts so that they are a bit less technical.

corrados:
Hmm, I noticed that we know have a lot of "vanished" texts in the translations after I updated with Qt lupdate and lrelease. This is not good since all translators must fix that now...

trebmuh:
> "Higher level language"? Sorry, but that's a bit of a stretch... The article 'a' was simply in the wrong place and I moved it to where it would be syntactically correct. I find it hard to believe I magically transformed the sentence into some 'higher level language' that is harder to understand than the original ‒I repeat, incorrect‒ wording simply by moving an article. There are many, many other sentences in this software that, while grammatically/syntactically correct, need to be read several times to get your head around them.

I hope you didn't take it personally. I didn't say you "magically transformed" anything. There is 2 points in what I've tried to express here: first is that if syntactically wrong, the original wording was easier for me to understand (because this "wrong" English syntax match the "correct" French syntax), that's why I asked if you were 100% sure, second I've proposed to put the same idea in a way which is easier to understand for anyone : _"the sound card's buffer is too small - increasing it's size would fix the problem"_ which matches the same meaning and is probably easier to understand.



> Maybe a native speaker should review all the texts. Maybe we should change some texts so that they are a bit less technical.
> Still, I'm looking at it from a native speaker's perspective and if you think a simplified version of the sentence would make life easier for non-native English speaking translators, I have no problem with changing it. I was just trying to make my changes as non-invasive as possible to avoid completely re-writing sentences.

OK, thank you for explaining.

> If I were to start being picky I could have gone much further.

_and_

> Maybe a native speaker should review all the texts. Maybe we should change some texts so that they are a bit less technical.

It could be good.


> Hmm, I noticed that we know have a lot of "vanished" texts in the translations after I updated with Qt lupdate and lrelease. This is not good since all translators must fix that now...

As a translator, if the changes are improving the software quality, I'm all good with it. C'est la vie!

ignotus666:
I translate and proofread for a living so if you want I could have a go at revising the text. As I said, when doing these corrections I didn't want to change too much because it might have seemed too intrusive - my intention was just to correct obvious errors. In general it's fine but there are some sentences that I'd probably have to check with you because I'm not sure I understand what they actually mean :)

 Regarding the vanished texts I think it's just a matter of copying and pasting the "vanished" translation because in essence they haven't changed at all.

ignotus666:
No worries @trebmuh ;) re-reading myself I see I came across as a bit defensive. I know where you're coming from when you say certain syntax structures might seem more understandable because they follow the pattern of other languages - but if they're incorrect... they're incorrect. I think each language should adhere to its own idiomatic grammar and syntax and we should not be twisting them to adapt to other languages.

ignotus666:
BTW, slightly OT, but I'm also a conference interpreter and these days I've been doing online interpreting jobs where real time audio is crucial. My experience so far has been that the tools available are... pretty crap. Very high latency and terrible audio quality, which makes my job infinitely more complicated and frustrating. I've almost been tempted to walk them through installing Jamulus, only it's not entirely suited to this application because of the more complicated audio routing that has to be done... but there's a market there, and Jamulus would smash the competition to pieces!

PR #258, opened 2020-05-21
Get rid of a warning

pljones:
QtCreator was annoying me with its little yellow markers...

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -746,7 +746,7 @@ void CServer::Stop()
 void CServer::OnTimer()
 {
     int                i, j, iUnused;
-    int                iClientFrameSizeSamples;
+    int                iClientFrameSizeSamples = 0;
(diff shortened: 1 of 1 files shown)


corrados:
Wouldn't it be sufficient to do:
`void CServer::OnTimer()
{
    int                i, j, iUnused;
    int                iClientFrameSizeSamples = 0;`

pljones:
Oh... well, could do - I just felt this was more explicit about when the value is being "unused".  There are four places it's currently set and two places it's currently not set.  So you could (potentially) end up setting it in the first if and then not clearing it in the second (okay, we can see it's not used... but you have to look carefully for what the nullptr pointer value is used for).

corrados:
Yes, but just to remove the warning, I like my simple solution a bit better :-).

corrados:
I just fixed it myself with my last commit. So no need for a merge.

PR #259, opened 2020-05-21
#70 Exit client cleanly on signal

### LINKED ISSUE #70: Would save settings on Linux cmdline termination signal
For scripting integration with qjackctl it would be nice if pkill jamulus (which defaults to send SIGTERM) would save current settings as from GUI Ctrl+Q|View/Exit.
Same issue example for LibreOffice : https://ask.libreoffice.org/en/question/159189/close-libreoffice-gracefuly-from-command-line/

pljones:
Here's an attempt to fix #70 -- appears to work on a quick test.


corrados:
Perfect, thanks. I'll try that out on my Linux laptop, too.

PR #283, opened 2020-05-25
Add banjo and mandolin as instruments

atsampson:
Two more instruments that are now getting played on the public servers...

The icons are edited from photos I took myself of a Leedy Collegian banjo banjo and a Freshwater octave mandola mandolin, and are available under CC0 as well as Jamulus's normal license. I've put the original images [here](https://stuff.offog.org/jamulus/), also under CC0.

This will need an lrelease translation update to rebuild the .qm files after merging.


Snayler:
Nice to see you already translated to each language 👍

corrados:
Thank you!

Issue #228, opened 2020-05-15
GUI Interface for Recording


pljones:
Currently the Server GUI has no support for recording.

As a first step, I'd like to add
- whether recording is enabled / happening
- ability to start a new recording (similar to SIGUSR1 support under #170)

I've got a little bit of code started for this, so it'll probably be the next thing I pick up.
https://github.com/pljones/jamulus/tree/feature/228-gui-triggered-recording-cuts

corrados:
I wonder if you can apply Labels to a Github Issue when you create it? Or is it only allowed for me to apply labels?

Snayler:
@corrados only the owner of the project can apply labels, as far as I can tell from the issues I created.

WolfganP:
https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/permission-levels-for-a-user-account-repository collaborators can work with labels and other projects´admin tasks.
https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/inviting-collaborators-to-a-personal-repository shows how to invite certain users to collaborate
The other option is to convert the account to an organization which allows for better control of roles and permissions, but it has it´s own drawback https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/converting-a-user-into-an-organization

corrados:
I was just asking because of curiosity. There is no need for getting help with applying the labels. Adding the labels is kind of project organization for me.

trebmuh:
@corrados : there can be some sort off automatic labeling. LMMS github is doing that as far as I remember.

corrados:
No, actually I prefer to set the label myself. There are not too many new Issues created and it seems the first wave of Corona is through and things go back to normal slowly:
![grafik](https://user-images.githubusercontent.com/46655886/82122616-02bbba00-9795-11ea-9eb7-a3fdbda4ddb7.png)

trebmuh:
That was just a "FYI", it's up to you.

corrados:
Yes, I know. Thanks.
But now we should continue in the discussion about the actual topic of the Issue and not only off-topic things :-).

pljones:
I'm sure many governments would like a Corona graph like yours :).

corrados:
Well, this is what we have in Germany regarding Corona:
![grafik](https://user-images.githubusercontent.com/46655886/82123970-90e76e80-979c-11ea-9f73-05242e4b1733.png)
Looks somewhat similar :-).

pljones:
[...]

PR #290, opened 2020-05-25
Initial Recording GUI

pljones:
Issue #228.

Display the recorder state and latest recording directory and
allow a new recording to be requested.

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -470,7 +470,7 @@ CServer::CServer ( const int          iNewMaxNumChan,
 
     QObject::connect( &JamRecorder,
         SIGNAL ( RecordingSessionStarted ( QString ) ),
-        this, SLOT ( OnRecordingSessionStarted ( QString ) ) );
+        SIGNAL ( RecordingSessionStarted ( QString ) ) );
 
     QObject::connect ( QCoreApplication::instance(),
         SIGNAL ( aboutToQuit() ),
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -468,6 +468,10 @@ CServer::CServer ( const int          iNewMaxNumChan,
        SIGNAL ( SvrRegStatusChanged() ),
        this, SLOT ( OnSvrRegStatusChanged() ) );
 
+    QObject::connect( &JamRecorder,
+        SIGNAL ( RecordingSessionStarted ( QString ) ),
+        this, SLOT ( OnRecordingSessionStarted ( QString ) ) );
+
     QObject::connect ( QCoreApplication::instance(),
         SIGNAL ( aboutToQuit() ),
         this, SLOT ( OnAboutToQuit() ) );
--- a/src/serverdlg.cpp
+++ b/src/serverdlg.cpp
@@ -474,6 +474,12 @@ void CServerDlg::OnCentServAddrTypeActivated ( int iTypeIdx )
     UpdateGUIDependencies();
 }
 
+void CServerDlg::OnServerStopped();
+{
+    UpdateSystemTrayIcon ( false );
+    UpdateRecorderStatus ( QString::null );
+}
+
 void CServerDlg::OnSysTrayActivated ( QSystemTrayIcon::ActivationReason ActReason )
 {
     // on double click on the icon, show window in fore ground
--- a/src/serverdlg.cpp
+++ b/src/serverdlg.cpp
@@ -336,8 +336,8 @@ lvwClients->setMinimumHeight ( 140 );
         this, SLOT ( OnCentServAddrTypeActivated ( int ) ) );
 
     // push buttons
-    QObject::connect( pbtNewRecording, SIGNAL ( released() ),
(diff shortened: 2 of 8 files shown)


src/server.cpp
```diff
@@ -468,6 +468,10 @@ CServer::CServer ( const int          iNewMaxNumChan,
        SIGNAL ( SvrRegStatusChanged() ),
        this, SLOT ( OnSvrRegStatusChanged() ) );
 
+    QObject::connect( &JamRecorder,
```
corrados, line 471: You can directly connect a SIGNAL to a SIGNAL, see, e.g., here: https://github.com/corrados/jamulus/blob/master/src/client.cpp#L156
So you do not need the OnRecordingSessionStarted function.

src/serverdlg.cpp
```diff
@@ -319,6 +335,10 @@ lvwClients->setMinimumHeight ( 140 );
     QObject::connect ( cbxCentServAddrType, SIGNAL ( activated ( int ) ),
         this, SLOT ( OnCentServAddrTypeActivated ( int ) ) );
 
+    // push buttons
+    QObject::connect( pbtNewRecording, SIGNAL ( released() ),
+            this, SLOT ( OnNewRecordingClicked() ) );
```
corrados, line 340: there are additional spaces

src/serverdlg.cpp
```diff
@@ -451,6 +474,11 @@ void CServerDlg::OnCentServAddrTypeActivated ( int iTypeIdx )
     UpdateGUIDependencies();
 }
 
+void CServerDlg::OnNewRecordingClicked()
```
corrados, line 477: These "mini function bodies" I usually implement right in the header, see, e.g., here: https://github.com/corrados/jamulus/blob/master/src/connectdlg.h#L111

src/serverdlg.cpp
```diff
@@ -504,6 +532,18 @@ void CServerDlg::OnTimer()
     ListViewMutex.unlock();
 }
 
+void CServerDlg::OnServerStarted()
```
corrados, line 535: see above comment

src/serverdlg.h
```diff
@@ -90,9 +91,10 @@ public slots:
     void OnLocationCityTextChanged ( const QString& strNewCity );
     void OnLocationCountryActivated ( int iCntryListItem );
     void OnCentServAddrTypeActivated ( int iTypeIdx );
+    void OnNewRecordingClicked();
     void OnTimer();
-    void OnServerStarted() { UpdateSystemTrayIcon ( true ); }
```
corrados, line 94: Oh, I see. You moved that to the cpp file. So you like it better to have it in the cpp?

src/server.cpp
pljones: Yeah - I tried that.  It wasn't working -- but when I worked out why, I forgot to put this back again.  (I'd had some debugging in the slot handler.)  I'll fix it.

src/serverdlg.cpp
pljones: Tsk!  Missed them.

src/serverdlg.cpp
pljones: Good point - I've been doing that, don't know why this one slipped through.

src/serverdlg.h
pljones: Well, there's _two_ lines of code now... It doubled in size!  I thought it might be better there.  If you're happy with it staying, I'll keep it in the header.

pljones:
OK, updated.

src/serverdlg.h
corrados: Yes, that's ok. If it is more than one line, I also prefer to have it in the cpp file.

src/serverdlg.h
```diff
@@ -91,8 +92,14 @@ public slots:
     void OnLocationCountryActivated ( int iCntryListItem );
     void OnCentServAddrTypeActivated ( int iTypeIdx );
     void OnTimer();
+
     void OnServerStarted() { UpdateSystemTrayIcon ( true ); }
-    void OnServerStopped() { UpdateSystemTrayIcon ( false ); }
+    void OnServerStopped()
+    {
+         UpdateSystemTrayIcon ( false );
```
corrados, line 99: Oh, I see you have changed it. I should have been more precise in my initial post. I only prefer to put it in the header file if it is a single command. Just in some very special cases I put more than one in the header. But in this case, it should belong to the cpp file as you had it in your original commit. Sorry for the confusion...

corrados:
One final note: Can you please also update the ChangeLog for your modifications?

src/serverdlg.h
pljones: No problem -- Both initially changed.  Then I changed Started not to change.  But left it in the cpp.  I'll split them up :).

pljones:
Should all be there now.

corrados:
Thank you.

PR #302, opened 2020-05-28
Check the size of PROTMESSID_ACKN messages

atsampson:
The Evaluate* functions that parse other types of messages all check the body size before starting to read data, but the special code for acknowledgements didn't do this, so an ACKN message on an existing connection that had a valid checksum but no body would result in an out-of-bounds read.

Found by fuzzing the protocol parser with [afl-fuzz](https://lcamtuf.coredump.cx/afl/), using messages from CTestbench as an initial corpus and disabling the CRC check. This was the only problem it found in a 24h fuzzing run.


corrados:
Very good that you found it. Thank you!

PR #303, opened 2020-05-28
Update CTestbench for new protocol messages

atsampson:
I was using the CTestbench code to generate sample messages for a fuzzer, and spotted that it didn't generate all the message types that are now possible. Add the new message types, and connect up the signal for sending connectionless messages.

(As before, this won't actually do anything unless you enable the testbench code by editing main.cpp.)


corrados:
Thank you.

PR #304, opened 2020-05-28
Fixes issue with AppNap causing stuttering/silence in background

AronVietti:
Adds a mac only Activity class that controls the NSProccessInfo
Activity options. This exposes a C++ interface for the
Objective C calls and handles the Activity Id lifecycle. Currently
sets the Activity to background, latency critical, sleep disabled.

Added the Activity calls to main.cpp, starting it in the same
location that Windows priority is set, and ending it right before
the app exits.

Updated Jamulus.pro to add the new files and required lib.


corrados:
Thank you!

PR #312, opened 2020-05-30
#228 Enable/Disable recording from command line

pljones:
This is the second step of #228 - wiring SIGUSR2 and the GUI to enable/disable recording...

Unfortunately, both SIGUSR1 and SIGUSR2 sometimes cause the application to exit, which isn't great.  I'm not sure why.  (Of course, SIGINT and SIGTERM are _meant_ to cause the application to exit; I've yet to see it happen _first_ time, either.)

### DIFF
--- a/src/recorder/jamrecorder.cpp
+++ b/src/recorder/jamrecorder.cpp
@@ -302,7 +302,7 @@ QMap<QString, QList<STrackItem>> CJamSession::TracksFromSessionDir(const QString
  * @brief CJamRecorder::Init Create recording directory, if necessary, and connect signal handlers
  * @param server Server object emiting signals
  */
-void CJamRecorder::Init( const CServer* server,
+bool CJamRecorder::Init( const CServer* server,
                          const int      _iServerFrameSizeSamples )
 {
     QFileInfo fi(recordBaseDir.absolutePath());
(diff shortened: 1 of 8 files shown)


pljones:
It's _so_ long since I wrote a signal handler I can't remember what they're meant to look like, so I can't remember if the code I've (now) got it is reasonable.

I'm wondering whether I should SIG_IGN within the signal handler, writing to the pipe, then re-init afterwards.  Can't think _why_ that would help, though.

It's failing before the Qt emit gets handled, though, so it's somewhere in signalhander.{h,cpp}.

atsampson:
I'm not sure if either of these are causing the problem you're seeing, but:

- `signalHandler` ought to save and restore `errno` (since `write` may change it, and the signal might have arrived in the middle of some other piece of code that's using it).

- `setSignalHandled` should initialise `sa_flags` to 0; currently it'll have some random value off the stack initially.

You shouldn't need to `SIG_IGN` within the signal handler; the signal you're handling is blocked automatically, unless you specify the `SA_NODEFER` flag.

pljones:
Initialising the `sa_flags` could help, true.

atsampson:
Excellent!

`SA_RESTART` can't reset `errno`, because the latter's part of libc rather than the kernel. Richards' "Unix Systems Programming" gives this example:

errno

corrados:
Is this PR ready to be merge or do you want to do further modifications?

pljones:
This one's ready.

storeilly:
Just FYI.. Tried this on 3.5.6 server (on Ubuntu 18.04.4) with a single 3.5.5 client and it appears to create a new recording on SIGUSR1 or SIGUSR2 but doesn't suspend/stop recording. And it did stop the server once. Any logs you would like to assist diag?

pljones:
OK, I've reproduced this.

There's definitely something flaky with the signal handler.  The checkbox on the GUI works fine.

PR #318, opened 2020-06-01
Make level meter LED black when off

fleutot:
Implementation of [this proof of concept](https://github.com/corrados/jamulus/pull/222).

- Should I remove the png files for grey?
- I considered renaming the `enum` value and `QPixmap` identifier to `*Off`, because it's the logical meaning, whereas the actual color is just the implementation. Do you think it's a good idea?
- if yes, one could argue that `Green/Yellow/Red` could become `Low/Mid/High`. This would feel out of scope for this PR though.


corrados:
Thank you.
> Should I remove the png files for grey?

No, that is not necessary.

> I considered renaming the enum value and QPixmap identifier to *Off, because it's the logical meaning, whereas the actual color is just the implementation. Do you think it's a good idea?

It would be an improvement but not really necessary.

> if yes, one could argue that Green/Yellow/Red could become Low/Mid/High. This would feel out of scope for this PR though.

I prefer that the name reflects the color. If we change the color in the future, we may change it to low/mid/high but for now I would keep the names.

PR #319, opened 2020-06-02
instruments: Add ukulele and bass ukulele

dos1:
The icons are derived from an Openclipart image by qubodup:
https://openclipart.org/detail/214439/simple-ukulele-no-outline


corrados:
Thanks. I'll take a look at your pull request this weekend. Right now I do not have access to my laptop.

PR #323, opened 2020-06-03
opus: fix equivalent bitrate calculation for <20ms frame sizes

marcan:
This is a bugfix that I submitted to upstream Opus a few days ago. It
has been merged, but not released yet. Since it affects <20ms frame
sizes, which is what Jamulus uses, and affects audio quality
(particularly mono downmixes), it makes sense to merge this into
Jamulus' vendored Opus tree early.

Upstream issue: https://gitlab.xiph.org/xiph/opus/-/issues/2323


corrados:
Yes, I agree, thank you.

Issue #295, opened 2020-05-26
Create server recording indicator


Snayler:
As discussed on #276, it would be useful to have an indicator on the client to let the user know when a server is recording.

pljones:
Picking this up now.

pljones:
Raised #324 - see comments there.

Snayler:
This wasn't implemented yet, was it?

corrados:
No, it wasn't.

pljones:
Yep, sorry I've been busy at work and simply exhausted (I'm a programmer and my brain can't cope...).  I was hoping to get some more done over this weekend, just not had a chance yet.  But this is what I'll be doing next.

pljones:
Yep, getting to this now - #363 done above, #366 with @corrados and I've the next one lined up with the client side changes, so we could be good to go.

pljones:
#366 done and #368 ready now.  All done once that's in.

corrados:
merged -> closed

pljones:
Looks like I'm not clearing the client idea of reccorder state.  I switched from my 3.5.7git server with recording active to a 3.4.5git server - predates recording and can't notify anyway! - and was told recording was active..!

corrados:
I already fixed that in the Git repo. Just make a pull and everything should work as expected :-)

PR #324, opened 2020-06-03
Feature/295 client side recording indicator

pljones:
Note that this is branched off 24f972c #228 feature/228-ui-for-recording-dir -- that pull request must go in first.  (There may not be clashes but that's the way around that I tested it.)

This is #295 - a status indication in the client of when the server is recording.  Unfortunately, prior to this version of the server, the client can't know for sure.  (I could have a server version test - below 3.4.4 it can't be recording.  There _are_ still 17 such servers out there on the Default central server...)

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -121,7 +121,7 @@ CChannelFader::CChannelFader ( QWidget*     pNW,
     pMainGrid->addWidget ( pLabelInstBox );
 
     // add fader frame to audio mixer board layout
-    pParentLayout->addWidget( pFrame );
+    pParentLayout->addWidget ( pFrame );
 
     // reset current fader
     Reset();
@@ -603,18 +603,19 @@ double CChannelFader::CalcFaderGain ( const int value )
 * CAudioMixerBoard                                                             *
 \******************************************************************************/
 CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent, Qt::WindowFlags ) :
-    QGroupBox            ( parent ),
-    vecStoredFaderTags   ( MAX_NUM_STORED_FADER_SETTINGS, "" ),
-    vecStoredFaderLevels ( MAX_NUM_STORED_FADER_SETTINGS, AUD_MIX_FADER_MAX ),
-    vecStoredPanValues   ( MAX_NUM_STORED_FADER_SETTINGS, AUD_MIX_PAN_MAX / 2 ),
-    vecStoredFaderIsSolo ( MAX_NUM_STORED_FADER_SETTINGS, false ),
-    vecStoredFaderIsMute ( MAX_NUM_STORED_FADER_SETTINGS, false ),
-    iNewClientFaderLevel ( 100 ),
-    bDisplayPans         ( false ),
-    bIsPanSupported      ( false ),
-    bNoFaderVisible      ( true ),
-    iMyChannelID         ( INVALID_INDEX ),
-    strServerName        ( "" )
+    QGroupBox               ( parent ),
+    vecStoredFaderTags      ( MAX_NUM_STORED_FADER_SETTINGS, "" ),
+    vecStoredFaderLevels    ( MAX_NUM_STORED_FADER_SETTINGS, AUD_MIX_FADER_MAX ),
+    vecStoredPanValues      ( MAX_NUM_STORED_FADER_SETTINGS, AUD_MIX_PAN_MAX / 2 ),
+    vecStoredFaderIsSolo    ( MAX_NUM_STORED_FADER_SETTINGS, false ),
+    vecStoredFaderIsMute    ( MAX_NUM_STORED_FADER_SETTINGS, false ),
+    iNewClientFaderLevel    ( 100 ),
+    bDisplayPans            ( false ),
+    bIsPanSupported         ( false ),
+    bNoFaderVisible         ( true ),
+    iMyChannelID            ( INVALID_INDEX ),
+    strServerName           ( "" ),
+    strServerRecordingState ( "" )
 {
     // add group box and hboxlayout
     QHBoxLayout* pGroupBoxLayout = new QHBoxLayout ( this );
@@ -740,6 +741,19 @@ void CAudioMixerBoard::SetPanIsSupported()
     SetDisplayPans ( bDisplayPans );
 }
 
+void CAudioMixerBoard::SetServerRecordingState ( QString state )
+{
+    strServerRecordingState = state;
+    if ( !strServerName.isEmpty() )
+    {
+        QString strTitle = this->title().remove ( QRegExp ( "^\\[.\\] " ) );
+        if ( strTitle.mid ( 0, 1 ) != "\u2588" )
+        {
+            setTitle ( strServerRecordingState + " " + strTitle );
+        }
+    }
+}
+
 void CAudioMixerBoard::HideAll()
 {
     // make all controls invisible
@@ -769,7 +783,7 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
     // in the audio mixer board to show a "try to connect" before
     if ( bNoFaderVisible )
     {
-        setTitle ( tr ( "Personal Mix at the Server: " ) + strServerName );
+        setTitle ( strServerRecordingState + " " + tr ( "Personal Mix at the Server: " ) + strServerName );
     }
 
     // get number of connected clients
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -185,6 +185,12 @@ CClientDlg::CClientDlg ( CClient*        pNCliP,
     // reset mixer board
     MainMixerBoard->HideAll();
 
+    MainMixerBoard->setAccessibleName ( tr ( "Personal Mix group box" ) );
+    MainMixerBoard->setWhatsThis ( "<b>" + tr ( "Personal Mix" ) + ":</b> "
+        + "Displays your local, per-channel settings of all channels on a server. "
+        + "The group box title displays an indicator of whether the server is recording, "
+        + "where this is known, along with the server name." );
+
     // restore channel level display preference
     MainMixerBoard->SetDisplayChannelLevels ( pClient->GetDisplayChannelLevels() );
 
@@ -489,6 +495,10 @@ CClientDlg::CClientDlg ( CClient*        pNCliP,
     QObject::connect ( pClient, &CClient::VersionAndOSReceived,
         this, &CClientDlg::OnVersionAndOSReceived );
 
+    QObject::connect ( pClient,
+        SIGNAL ( RecorderStateChange ( ESvrRecState ) ),
+        this, SLOT ( OnRecorderStateChange ( ESvrRecState ) ) );
+
 #ifdef ENABLE_CLIENT_VERSION_AND_OS_DEBUGGING
     QObject::connect ( pClient, &CClient::CLVersionAndOSReceived,
         this, &CClientDlg::OnCLVersionAndOSReceived );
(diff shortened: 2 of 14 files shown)


src/audiomixerboard.cpp
```diff
@@ -741,6 +742,16 @@ void CAudioMixerBoard::SetPanIsSupported()
     SetDisplayPans ( bDisplayPans );
 }
 
+void CAudioMixerBoard::SetServerRecordingState ( QString state )
+{
+    strServerRecordingState = state;
+    QString strTitle = this->title().remove ( QRegExp ( " : ~~ .* ~~" ) );
```
corrados, line 748: Wow, this looks like a hack...

src/audiomixerboard.cpp
```diff
@@ -741,6 +742,16 @@ void CAudioMixerBoard::SetPanIsSupported()
...
+    if ( strServerRecordingState != "" && strTitle.mid ( 0, 1 ) != "\u2588" )
+    {
+        setTitle ( strTitle + " : ~~ " + strServerRecordingState + " ~~" );
```
corrados, line 751: These ~~ are at at least three places in the code. A single source would be preferred.

src/clientdlg.cpp
```diff
@@ -777,6 +781,23 @@ void CClientDlg::OnVersionAndOSReceived ( COSUtil::EOpSystemType ,
 #endif
 }
 
+void CClientDlg::OnRecorderStateChange ( uint8_t state )
+{
+    if ( state == 2 )
```
corrados, line 786: Why is state a number and no enum?

src/clientdlg.cpp
```diff
@@ -1083,6 +1104,9 @@ void CClientDlg::Connect ( const QString& strSelectedAddress,
         // change connect button text to "disconnect"
         butConnect->setText ( tr ( "D&isconnect" ) );
 
+        // set default recording state
+        MainMixerBoard->SetServerRecordingState( tr( "May be recording" ) );
```
corrados, line 1108: What is this "may be recording" thing? Is it or is it not recording? This message just confuses the users.

src/clientdlg.cpp
```diff
@@ -1132,6 +1156,9 @@ OnTimerStatus();
 
     // clear mixer board (remove all faders)
     MainMixerBoard->HideAll();
+
+    // clear recording state
+    MainMixerBoard->SetServerRecordingState( "" );
```
corrados, line 1161: There are a lot of places where a space is missing before the (.

src/protocol.h
```diff
@@ -78,6 +78,7 @@
 #define PROTMESSID_CLM_REQ_CONN_CLIENTS_LIST  1014 // request the connected clients list
 #define PROTMESSID_CLM_CHANNEL_LEVEL_LIST     1015 // channel level list
 #define PROTMESSID_CLM_REGISTER_SERVER_RESP   1016 // status of server registration request
+#define PROTMESSID_CLM_RECORDER_STATE_CHANGED 1017 // new recording state of server
```
corrados, line 81: This is not good. Why is this a connection less message? That must certainly a normal message instead.

src/server.cpp
```diff
@@ -400,12 +400,8 @@ CServer::CServer ( const int          iNewMaxNumChan,
...
     }
 
-    // enable jam recording (if requested) - kicks off the thread
-    if ( !strRecordingDirName.isEmpty() )
-    {
-        bRecorderInitialised = JamRecorder.Init ( this, iServerFrameSizeSamples );
-        bEnableRecording     = bRecorderInitialised;
-    }
+    // Enable jam recording (if requested) - kicks off the thread
+    SetRecordingDir ( strRecordingDirName );
```
corrados, line 404: See my comment in the other pull request.

src/server.cpp
```diff
@@ -606,6 +598,11 @@ void CServer::OnNewConnection ( int          iChID,
     // send version info (for, e.g., feature activation in the client)
     vecChannels[iChID].CreateVersionAndOSMes();
 
+    // A new client connected to the server and needs to know whether
+    // its audio is being recorded or not.  Subsequent changes
+    // will be sent out as and when they happen.
+    ConnLessProtocol.CreateCLRecorderStateChange ( RecHostAddr, GetRecorderState() );
```
corrados, line 604: Do not use a connection less message.

src/server.cpp
```diff
@@ -1342,6 +1313,35 @@ void CServer::CreateOtherMuteStateChanged ( const int  iCurChanID,
...
+    // get the recorder state
+    uint8_t state = GetRecorderState();
+
+    // now send connected channels list to all connected clients
+    for ( int i = 0; i < iMaxNumChannels; i++ )
+    {
+        if ( vecChannels[i].IsConnected() )
+        {
+            // send message
+            ConnLessProtocol.CreateCLRecorderStateChange ( vecChannels[i].GetAddress(), state );
```
corrados, line 1327: see above

src/server.cpp
```diff
@@ -1342,6 +1313,35 @@ void CServer::CreateOtherMuteStateChanged ( const int  iCurChanID,
...
+        }
+    }
+}
+
+uint8_t CServer::GetRecorderState()
+{
+    if ( !bRecorderInitialised )
+    {
+        return 0;
```
corrados, line 1336: use enums instead

src/server.cpp
```diff
@@ -1572,6 +1572,80 @@ void CServer::WriteHTMLChannelList()
     streamFileOut << "</ul>" << endl;
 }
 
+void CServer::RequestNewRecording()
```
corrados, line 1575: see my comment in the other pull request

src/server.h
```diff
@@ -198,9 +198,12 @@ class CServer :
                           CVector<int>&          veciNetwFrameSizeFact );
 
     bool GetRecorderInitialised() { return bRecorderInitialised; }
+    QString GetRecorderErrMsg() { return strRecorderErrMsg; }
     bool GetRecordingEnabled() { return bEnableRecording; }
     void RequestNewRecording();
     void SetEnableRecording ( bool bNewEnableRecording );
+    QString GetRecordingDir() { return strRecordingDir; }
+    void SetRecordingDir( QString newRecordingDir );
```
corrados, line 206: space missing

src/server.h
```diff
@@ -349,16 +356,19 @@ class CServer :
...
-    recorder::CJamRecorder     JamRecorder;
-    bool                       bRecorderInitialised;
-    bool                       bEnableRecording;
-
     // HTML file server status
     bool                       bWriteStatusHTMLFile;
     QString                    strServerHTMLFileListName;
     QString                    strServerNameWithPort;
 
+    // recording thread
```
corrados, line 364: see other pull request

src/clientdlg.cpp
pljones: It's explicitly true.

If you connect to a server and that server does not tell you it's recording - because it's too old to do so, that does not mean it's not recording.  It does not mean it is recording.  It means there is no way to know.  It may or may not be.

May be a preference for the indicator ("I don't care to be told").

src/audiomixerboard.cpp
pljones: I'd rather it wasn't done like this at all but as a separate indicator somewhere on the UI.  I'd like a little red "led" on in the top left-hand corner of the server group box, really.  But it would need to be tri-state.

src/clientdlg.cpp
corrados: :-)
You can also argument that old client versions do not have the indicator. So it is just a matter of old versions. You need both, a new client and a new server to get that information. But I am fine with that.

src/audiomixerboard.cpp
corrados: Adding the information in the title of the audio mixer board is totally fine. My complain was just about the single source problem.

src/protocol.h
pljones: It happens at any time under the person sitting at the server window's control.  It doesn't relate to anything any client did.

src/server.cpp
pljones: The message is sent in response to a server state change, not in response to a packet from the client.  What constitutes a connectionless message?  `PROTMESSID_CLM_CHANNEL_LEVEL_LIST`, `PROTMESSID_CLM_CONN_CLIENTS_LIST` are connectionless.  I can see `PROTMESSID_CHAT_TEXT` as about the only "broadcast" message that's not a CLM.

src/protocol.h
corrados: But it only only applies to all connected clients. If they are connected, they can receive normal messages.

corrados: If you use connection less messages, the message can get lost and the indicator is incorrect. The normal message makes sure the indicator is always correct.

src/server.cpp
corrados: There are two versions of conn_clients_list. One connection less intended for the server list and one for actual connected clients. The channel level is a special message which does not hurt if it gets lost, same with the ping message.

src/protocol.h
pljones: See question on other comment.

src/server.cpp
corrados: Your message must be a normal message to work properly.

pljones: OK, so like `PROTMESSID_CHAT_TEXT` then?

Does that mean I don't need the ultra-low frequency updates?

corrados: What do you mean by "ultra-low frequency updates"? The recording status changes only once in a while, right?

src/protocol.cpp
```diff
@@ -361,6 +361,18 @@ CONNECTION LESS MESSAGES
           (standard re-registration timeout).
 
 
+- PROTMESSID_CLM_RECORDER_STATE_CHANGED: Server recorder state has changed
+
+    +--------------+
+    | 1 byte state |
+    +--------------+
+
+    - "state":
```
pljones, line 370: This is now wrong.  Use the enum names.

src/server.cpp
pljones: See
https://github.com/corrados/jamulus/pull/324/files#diff-ad60d65b34e16a3319296bb5d683acd6R971-R978

corrados: Hm, this is a link to all changed files in your pull request. The question is, shall I look at one specific line? What line?

pljones: Hm.  It's actually a link to server.cpp lines 971 to 978 but github, whilst highlighting them, doesn't actually navigate you to them when you click the link.  No helpful.

src/server.cpp
```diff
@@ -956,8 +968,16 @@ JitterMeas.Measure();
...
     if ( iNumClients > 0 )
     {
+        // ultra-low frequency updates
+        if ( iULFFrameCount > ULTRA_LOW_FREQUENCY_LIMIT )
+        {
+            iULFFrameCount = 0;
+
+            // things can go missing, so send this every now and then
+            CreateAndSendRecorderStateForAllConChannels();
+        }
```
pljones, line 978: This is what I meant by the ultra-low frequency updates.

corrados: I see. Have not recognized that. That ULTRA_LOW_FREQUENCY_LIMIT and also the LOW_FREQUENCY_UPDATE_INTERVAL has to be removed. We need normal messages. The status indicator is only for connected clients, not intended for the server list. So only the status change is transmitted, no continuous updates.

corrados: You could take a look at PROTMESSID_MUTE_STATE_CHANGED. This is something similar. The Mute Icon status is only transmitted on a change.

src/audiomixerboard.cpp
pljones: Hopefully the latest version is less intrusive yet just as informative.

src/clientdlg.cpp
pljones: Hopefully all fixed now (in files touched by this patch, at least...).

src/server.cpp
pljones: Done :)

src/server.cpp
pljones: 👍 OK.

PR #330, opened 2020-06-04
Updated/reviewed help texts and Spanish translation

ignotus666:
Most changes have been to merge the texts reviewed by Gilgongo. All suggestions are welcome.


src/clientdlg.cpp
```diff
@@ -115,64 +112,67 @@ CClientDlg::CClientDlg ( CClient*        pNCliP,
...
+        "reverb level can be modified. For example, if "
+        "a microphone signal is fed in to the right audio channel of the "
+        "sound card and a reverb effect needs to be applied, set the "
         "channel selector to right and move the fader upwards until the "
-        "desired reverberation level is reached." ) + "<br>" + tr (
-        "The reverberation effect requires significant CPU so it should "
-        "only be used on fast PCs. If the reverberation level fader is set to "
-        "minimum (which is the default setting), the reverberation effect is "
+        "desired reverb level is reached." ) + "<br>" + tr (
+        "Reverb requires significant CPU so it should "
```
corrados, line 123: I think we can remove this. This is a relic from old times. With modern computers, the CPU requirement should be negligible.

src/clientdlg.cpp
```diff
@@ -115,64 +112,67 @@ CClientDlg::CClientDlg ( CClient*        pNCliP,
...
+        "<li>" "<b>" + tr ( "Red" ) + ":</b> " + tr ( "The delay is too large for "
+        "jamming." ) + "</li>"
+        "</ul>";
 
     lblDelay->setWhatsThis ( strLEDDelay );
     ledDelay->setWhatsThis ( strLEDDelay );
     ledDelay->setToolTip ( tr ( "If this LED indicator turns red, "
-        "you will not have much fun using the " ) + APP_NAME +
-        tr ( " software." ) + TOOLTIP_COM_END_TEXT );
+        "you will not have much fun using the application" ) +
```
corrados, line 159: There is a point missing after the application.

corrados:
There are just minor things to change. I'll do it myself after merging.
Thanks.

ignotus666:
I made those couple of changes and also removed the CPU stuff from manual.md

corrados:
I just noticed that there is a lot of work to do to fix the translation for your changes. I fear that the other contributors will not be happy to have to spend a lot of time again. Maybe they don't do it at all...

ignotus666:
I know, but if we wanted to bring the app's help texts up to date with the manual I don't see what else could have been done... I tried to modify as little as possible apart from the changes that were required to bring the help texts in line with manual.md and sort out some of the +APP NAME+ stuff. 

That said, it took me 1/2 an hour to update the translation - there's lots of text that can be 'recycled' from the 'vanished' segments. Editing the src files was a far bigger pain in the ar$e...

corrados:
Ok. Let's see when we get updates from the other translators.

Issue #223, opened 2020-05-15
Status of Jamulus integration in Linux distributions


corrados:
This issue is basically just a list of URLs which relate to possible integrations of Jamulus in Linux distributions.

Several years ago, there was an attempt to include Jamulus in Redhat:
https://bugzilla.redhat.com/show_bug.cgi?id=1118963
The CONFIG options "noupcasename" and "opus_shared_lib" were a result of it.

More recently, mirabilos did some work for including Jamulus in Debian:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=958146
We also have Jamulus Issues related to that, e.g., https://github.com/corrados/jamulus/issues/100

It seems that there is a Jamulus package in OpenSuse available:
https://software.opensuse.org/package/Jamulus

If you know of other Linux distributions or have updated informations, please post them here.

<a href="https://repology.org/project/jamulus/versions">
    <img src="https://repology.org/badge/vertical-allrepos/jamulus.svg" alt="Packaging status">
</a>

trebmuh:
Not a mainstream distribution, but one focused on music tools, its is actually in its testing phase for [LibraZiK](https://librazik.tuxfamily.org/doc3/). I'm waiting to finish the whole French translation to get it out of the tester repo and in the users repo.

Snayler:
ArchLinux (and any Arch based distro) has the latest version available in the AUR.
https://aur.archlinux.org/packages/jamulus/

pljones:
Just noticed the AUR has its own `jamulus.desktop`.  Maybe the Linux distros could agree on a `jamulus.desktop` entry that can be included or have one generated that's distro-specific by the build process?

trebmuh:
Yep, distros tends to do that from time to time. Sometimes with a very good reason, but usually not. Maybe someone can drop them a message about the [existence of a upstream desktop file](https://github.com/corrados/jamulus/blob/master/distributions/jamulus.desktop) ?

Oh, after checking, I know the packager and will report that to him.

jujudusud:
Hi all, Co-maintainer of AUR Archlinux package here.
just finishing to update the package in order to use upstream jamulus.desktop file.
Thanks a lot trebmuh and of you guy for all that stuff ! :-)

jtojnar:
[...]

PR #331, opened 2020-06-04
Install jamulus.desktop

atsampson:
As suggested in #223.

Because jamulus.desktop assumes the binary is called jamulus, and qmake doesn't have a mechanism for string expansion, only install the file so if noupcasename is set.

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -295,12 +295,24 @@ win32 {
     isEmpty(PREFIX) {
         PREFIX = /usr/local
     }
+
     isEmpty(BINDIR) {
         BINDIR = bin
     }
     BINDIR = $$absolute_path($$BINDIR, $$PREFIX)
-    INSTALLS += target
     target.path = $$BINDIR
+    INSTALLS += target
+
+    isEmpty(APPSDIR) {
+        APPSDIR = share/applications
+    }
+    APPSDIR = $$absolute_path($$APPSDIR, $$PREFIX)
+    desktop.path = $$APPSDIR
+    desktop.files = distributions/jamulus.desktop
+    # the .desktop file assumes the binary is called jamulus
+    contains(CONFIG, "noupcasename") {
+        INSTALLS += desktop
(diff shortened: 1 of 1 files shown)


Jamulus.pro
```diff
@@ -295,12 +295,24 @@ win32 {
...
     target.path = $$BINDIR
+    INSTALLS += target
+
+    isEmpty(APPSDIR) {
+        APPSDIR = share/applications
+    }
+    APPSDIR = $$absolute_path($$APPSDIR, $$PREFIX)
+    desktop.path = $$APPSDIR
+    desktop.files = distributions/jamulus.desktop
+    # the .desktop file assumes the binary is called jamulus
```
jtojnar, line 312: It should be possible to rename the desktop file end with `.in`, put `$$TARGET` in the `Exec` file and then use `QMAKE_SUBSTITUTES += distributions/jamulus.desktop.in`.

See https://martin.rpdev.net/2017/05/10/creating-config-header-files-using-qmake.html or https://github.com/detrout/debian-signon-ui/blob/c03c68965f92285d3d5341b45414473081ed5c1b/src/signon-ui.pro#L105-L109.

Jamulus.pro
```diff
@@ -295,12 +295,24 @@ win32 {
...
+    contains(CONFIG, "noupcasename") {
+        INSTALLS += desktop
+    }
 }
```
jtojnar, line 316: We probably also need to install the icon, see https://github.com/corrados/jamulus/pull/247#discussion_r428560108.

seb314:
> @seb314 could you please test this?

seems to work! 
more specifically: when using this patch with the nix jamulus package, Jamulus is found in i3 or gnome3 like other apps via the .desktop file (without icon though)

Jamulus.pro
atsampson: Ahh! That is... extremely well hidden in the qmake documentation.

PR #343, opened 2020-06-08
Update Debian packaging

tormodvolden:
With qmake now doing a more standard installation, there is less
left to do in our debian/rules.

Signed-off-by: Tormod Volden <debian.tormod@gmail.com>


corrados:
Sounds good. Thanks.

PR #349, opened 2020-06-10
French translation update for 3.5.6

trebmuh:
HTH.

BTW @corrados if possible, it'd be great if we could have a system where a few days (1 week seems fine to me) before a release you ping the translators, then the translations are in a good shape when a new version is released. (aka: it's a bit sad-ish that the French translation isn't 100% updated for the 3.5.6).

PS: jamulus 3.5.6 is the first one to officially enters in LibraZiK.


melcon:
Not trying to hijack, but how one would do to create a new translation for pt_BR? Would that make sense?

trebmuh:
@melcon open another issue, you're hijacking the thread

Issue #351, opened 2020-06-11
SIGUSR2 handling is (sometimes?) not working


pljones:
I've hesitated in raising this because it's taken me some time to convince myself it actually happens...

SIGUSR2 should toggle "Enable jam recording" on and off.

When you start the server with recording initialised, any client connection will, by default, get recorded as part of the current session.

SIGUSR2 should terminate that session and prevent further sessions.

----

* start the server with the GUI: `./Jamulus -s -p 55850 -R /tmp/recorder`
* connect a client to the server, note that the current session directory is displayed on the server GUI
* disconnect the client, note that the server console displays the paths of the recording project files
* get the process ID: `ps -ef | grep 55850 | grep Jamulus` (depending on your `ps`, it's the second column; MacOS may need different options and, of course, no chance on Windows).
* send SIGUSR2 to the process: `kill -SIGUSR2 <pid>` (turns off recorder)

With the GUI open, the "Enable jam recording" check box toggles and (possibly _because of that_) the system appears to work as expected.  Leave the application running and try:
* connect a client to the server, note that the current session directory is displayed on the server GUI
* disconnect the client, note that the server console displays the paths of the recording project files

No current session should have been displayed on the GUI and the server does not list the project files.  Now, one more test:
* send SIGUSR2 to the process: `kill -SIGUSR2 <pid>` (turns on recorder)
* connect a client to the server, note that the current session directory is displayed on the server GUI
* send SIGUSR2 to the process: `kill -SIGUSR2 <pid>` (turns off recorder)

At this point, the client is still connected but the server has emitted the current project and the field becomes disabled.  So, finally:
* disconnect the client, note that the server console displays the paths of the recording project files

No project files are created as no session was in progress.  All is well.

----

Without the GUI the tests are not **ALWAYS** successful.
* start the server without the GUI: `./Jamulus -s -n -p 55850 -R /tmp/recorder`
* connect a client to the server
* disconnect the client, note that the server console displays the paths of the recording project files
* get the process ID: `ps -ef | grep 55850 | grep Jamulus` (depending on your `ps`, it's the second column; MacOS may need different options and, of course, no chance on Windows).
* send SIGUSR2 to the process: `kill -SIGUSR2 <pid>` (_should_ turn off recorder)
* connect a client to the server, note that the current session directory is displayed on the server GUI
* disconnect the client, note that the server console displays the paths of the recording project files

At this point, as we expected the recorder to have been disabled, no recording project files should have been displayed on the console.

[...]

PR #352, opened 2020-06-11
Add recorder state to console

pljones:
#351 Improve recorder logging

The big bits of re-doing the changes to the recorder are still awaiting me having time to give to them, so I've just put this in server.cpp for now.  I'll sort it out with the rest.

Does this need anything in Changelog?

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -240,6 +240,7 @@ CServer::CServer ( const int          iNewMaxNumChan,
     Logging                     ( iMaxDaysHistory ),
     iFrameCount                 ( 0 ),
     JamRecorder                 ( strRecordingDirName ),
+    bEnableRecording            ( false ),
     bWriteStatusHTMLFile        ( false ),
     HighPrecisionTimer          ( bNUseDoubleSystemFrameSize ),
     ServerListManager           ( iPortNumber,
@@ -405,7 +406,7 @@ CServer::CServer ( const int          iNewMaxNumChan,
     if ( !strRecordingDirName.isEmpty() )
     {
         bRecorderInitialised = JamRecorder.Init ( this, iServerFrameSizeSamples );
-        bEnableRecording     = bRecorderInitialised;
+        SetEnableRecording ( bRecorderInitialised );
     }
 
     // enable all channels (for the server all channel must be enabled the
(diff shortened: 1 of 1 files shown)


src/server.cpp
```diff
@@ -705,7 +707,12 @@ void CServer::SetEnableRecording ( bool bNewEnableRecording )
 {
     if ( bRecorderInitialised )
     {
+        // note that this block executes regardless of whether
+        // what appears to be a change is being applied, to ensure
+        // the requested state is the result
+
         bEnableRecording = bNewEnableRecording;
+        qInfo() << "Recording state" << ( bEnableRecording ? "enabled" : "disabled" );
```
corrados, line 715: You need to check the Qt version to use qInfo: https://doc.qt.io/qt-5/qtglobal.html#qInfo

corrados:
No need for ChangeLog entry.

src/server.cpp
corrados: See, e.g., here: https://github.com/corrados/jamulus/blob/master/src/clientdlg.cpp#L763

src/server.cpp
```diff
@@ -405,7 +406,7 @@ CServer::CServer ( const int          iNewMaxNumChan,
     if ( !strRecordingDirName.isEmpty() )
     {
         bRecorderInitialised = JamRecorder.Init ( this, iServerFrameSizeSamples );
-        bEnableRecording     = bRecorderInitialised;
+        SetEnableRecording ( bRecorderInitialised );
```
corrados, line 409: Why call the function? If bRecorderInitialised is false, the function body in SetEnableRecording is not entered. If it is true, the special implementation in the if-condition is not reached. So there is no difference in just applying the new value to the boolean variable. Do you have something else in mind why you changed this? Do you want to put new functionality in the SetEnableRecording function?

src/server.cpp
pljones: Ah right, will do.

src/server.cpp
pljones: Ah...  Ahh.....  AAHH!!! Yes... very good point.  Oops.

pljones: ...except if it's true, I want the message, and I'd rather have the code for the message only once...

corrados: Ok, makes sense. Thanks for the explaination.

Issue #178, opened 2020-05-06
Possibility to change channel/user position ordering


sthenos:
For a number of different events now happening (Choirs, world jams, public gigs on jamulus), the ordering of channels for users is quite important.   At the moment, I understand the ordering of channels/users is based on the time when a user arrives in a new room, with the server completely resetting its "memory" of user positions when the last person leaves the server.   

I'm wondering how much effort would it be so that channels/users are sorted based on the alphanumeric name of the user.   With a --channelsortalpha argument or something similar passed to the server which enables this feature?  

This way we can ask our friends/choir members to add a numeric value in front of their name to control the order they appear in the mixerboard,

corrados:
> I'm wondering how much effort would it be so that channels/users are sorted based on the alphanumeric name of the user.

It is not trivial. The server assigns an ID to the connected client and all properties (like name, instrument, gain, etc.) are bound to that ID. Of course you could introduce an ID translation in the client which maps the sorted IDs to the IDs to the server but this will require a significant amount of new code. So I would estimate that if you do it correctly, it will require at least two working days to support this.

sthenos:
ok fair enough, thanks for clarifying on the amount of time required.    It's certainly not an urgent thing, but more of a nice to have.   Maybe park this one for a while?

corrados:
I think I found a fairly simple solution to make it possible to re-order the faders. I am currently looking at it.

corrados:
Ok, it is implemented now. We have a new menu item (sorry for German language):
![grafik](https://user-images.githubusercontent.com/46655886/84050849-92c4dc00-a9ae-11ea-92b3-e6d23fdb6b97.png)

Before sorting:
![grafik](https://user-images.githubusercontent.com/46655886/84050744-70cb5980-a9ae-11ea-980e-a9b4c2a64891.png)
After sorting:
![grafik](https://user-images.githubusercontent.com/46655886/84050777-7b85ee80-a9ae-11ea-8657-ccc6d12242c2.png)

Please note that you can use the keyboard shortcut Ctrl+n which simplifies the usage.

sthenos:
Ok thanks Volker.  Works well so far, will do some more testing with larger groups.

One point, thing the english menuitem name should say "Sort channel users by name".   Also it's unclear if it's a checkbox and maintiains the state for all channels or if new people join you need to hit the menu item again to resort people?

corrados:
> should say "Sort channel users by name"

Thanks, I will change this.

[...]

PR #356, opened 2020-06-12
Addded sort by instrument and Edit menu option

Alberstein8:
- Added a new sort option in Edit menu, Sort by instruments. Usefull for choirs for example, to gather same voices together. Based in Sort by name #178 code added in 3.5.6 original release.
- Translation to German and Spanish of new Edit menu option. 
- Changed &amp letter of existing Edit menu option "&Sort Channel Users by Name" by "Sort Channel Users by &Name" as new menu item is "Sort Channel Users by &Instrument"


corrados:
Thanks for your code. Your "iReceivedInstrument = cChanInfo.iInstrument;" will not work correctly since you have to reset it correctly. But I want to store the complete info instead of the items separately.
I will do the implementation of storing the complete info now, then you can change your code to use this.

Another thing: I do not want to have another sort function. Please use "CAudioMixerBoard::ChangeFaderOrder ( const bool bDoSort )" but add a new function parameter telling the function to what parameter it shall do the sorting.

corrados:
Ok, I have now changed the function:
void CAudioMixerBoard::ChangeFaderOrder ( const bool        bDoSort,
                                          const EChSortType eChSortType )
But not tested.

PR #358, opened 2020-06-13
Update for sort order

gilgongo:
Also a slight clarification


gilgongo:
BTW the wording is generic so it will cover multiple sort options in the future (eg https://github.com/corrados/jamulus/pull/356)

PR #363, opened 2020-06-14
#295 Part 1: Define protocol message for server jam recorder state publishing

pljones:
The enum will retain "0 undefined" for where the client never receives the protocol message.

The next change will be to the jam recorder (actually to the server - until I get back to adding the last bit of the Jam Recorder GUI changes, under which a some of what's currently there gets rewritten) to get it to emit values (along with adding them to the enum - carefully allowing for forwards and backwards compatibility by defining a complete and stable set of values......).

### DIFF
--- a/src/protocol.cpp
+++ b/src/protocol.cpp
@@ -216,6 +216,17 @@ MESSAGES (with connection)
     note: does not have any data -> n = 0
 
 
+- PROTMESSID_RECORDER_STATE: notifies of changes in the server jam recorder state
+
+    +--------------+
+    | 1 byte state |
+    +--------------+
+
+    state is a value from the enum ERecorderState:
+    - 0 undefined (not used by protocol messages)
+    - tbc
+
+
 CONNECTION LESS MESSAGES
 ------------------------
 
@@ -658,6 +669,10 @@ if ( rand() < ( RAND_MAX / 2 ) ) return false;
 
             case PROTMESSID_VERSION_AND_OS:
                 bRet = EvaluateVersionAndOSMes ( vecbyMesBodyData );
+                break;
+
+            case PROTMESSID_RECORDER_STATE:
+                bRet = EvaluateRecorderStateMes ( vecbyMesBodyData );
                 break;
             }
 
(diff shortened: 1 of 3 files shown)


src/protocol.cpp
```diff
@@ -1544,6 +1559,37 @@ bool CProtocol::EvaluateVersionAndOSMes ( const CVector<uint8_t>& vecData )
...
+
+    CreateAndSendMessage ( PROTMESSID_RECORDER_STATE, vecData );
+}
+
+bool CProtocol::EvaluateRecorderStateMes(const CVector<uint8_t>& vecData)
+{
+    int iPos = 0; // init position pointer
+
+    // check size
+    if ( vecData.Size() != 2 )
```
corrados, line 1579: Why check for size 2? Shouldn't it be 1?

src/protocol.cpp
```diff
@@ -1544,6 +1559,37 @@ bool CProtocol::EvaluateVersionAndOSMes ( const CVector<uint8_t>& vecData )
...
+    {
+        return true; // return error code
+    }
+
+    // server jam recorder state
+    const ERecorderState eRecorderState = static_cast<ERecorderState> ( GetValFromStream ( vecData, iPos, 1 ) );
```
corrados, line 1585: We should introduce a check for valid enum values. I don't know what the static_cast does, if no enum is defined for the received integer number.

pljones: It's "undefined" -- i.e. the compiler is at liberty to do the easiest thing possible - just put the value into the storage "as is".

C++ hasn't got introspection so you can't check directly.

You'd have to have "known minimum" and "known maximum" values for each enum ("RS_VALID_MIN", "RS_VALID_MAX" etc) and then require that enum ranges not have intermediate _undefined_ values, I guess.

Maybe then have a macro that takes the "RS" (for example) and the result and checks for `xx_VALID_MIN` and `xx_VALID_MAX` using on the values passed.

src/protocol.cpp
pljones: Fixed, thanks.

PR #366, opened 2020-06-14
#295 Part 2: Define and send recorder state

pljones:
Most of this currently lives in `server.cpp` as that's the "controller" for the state of the jam recorder.  Once the rewrite gets done, much of this will move out into the new `JamController` class...

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -610,6 +610,9 @@ void CServer::OnNewConnection ( int          iChID,
     // send version info (for, e.g., feature activation in the client)
     vecChannels[iChID].CreateVersionAndOSMes();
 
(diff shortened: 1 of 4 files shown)


src/server.cpp
```diff
@@ -610,6 +610,9 @@ void CServer::OnNewConnection ( int          iChID,
     // send version info (for, e.g., feature activation in the client)
```
pljones, line 610: Shouldn't this have been made one of the earliest messages sent by the server?  The client _may_ want to use it to react to the other messages that currently come earlier?

corrados: Yes, I could be moved up.

PR #368, opened 2020-06-15
#295 Part 3: Client side changes for recorder state - the end

pljones:
I've added the recording indicator like last time.  Code looks a little tidier, I think...  

I was wondering about the wiring of signals into CAudioMixerBoard - this is handled in CClientDlg by calling a public method, instead.  Is that OK?

Also, the "What's this" for the group box still feels rather heavily-weighted to explaining the indicator...

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -619,13 +619,22 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent, Qt::WindowFlags ) :
     bIsPanSupported      ( false ),
     bNoFaderVisible      ( true ),
     iMyChannelID         ( INVALID_INDEX ),
-    strServerName        ( "" )
+    strServerName        ( "" ),
+    eRecorderState       ( RS_UNDEFINED )
 {
     // add group box and hboxlayout
     QHBoxLayout* pGroupBoxLayout = new QHBoxLayout ( this );
     QWidget*     pMixerWidget    = new QWidget(); // will be added to the scroll area which is then the parent
     pScrollArea                  = new CMixerBoardScrollArea ( this );
     pMainLayout                  = new QHBoxLayout ( pMixerWidget );
+
+
+    setAccessibleName ( "Personal Mix at the Server groupbox" );
+    setWhatsThis ( "<b>" + tr( "Personal Mix at the Server" ) + "</b>: " +
+        tr ( "When connected to a server, the controls here allow you to set " ) +
+        tr ( "your local mix without affecting what others hear from you.") + "<br/>" +
+        tr ( "The title shows the server name and, when known, "
+             "whether it is actively recording." ) );
 
     // set title text (default: no server given)
     SetServerName ( "" );
@@ -807,13 +816,32 @@ void CAudioMixerBoard::ChangeFaderOrder ( const bool        bDoSort,
     }
 }
 
+QString CAudioMixerBoard::GetTitle()
+{
+    QString myTitle = "";
+    if ( eRecorderState == RS_RECORDING )
+    {
+        myTitle = "[" + tr ( "RECORDING ACTIVE" )  + "] ";
+    }
+    return myTitle + tr ( "Personal Mix at: " ) + strServerName;
+}
+
+void CAudioMixerBoard::SetRecorderState ( const ERecorderState newRecorderState )
+{
+    eRecorderState = newRecorderState;
+    if ( !strServerName.isEmpty() && !bNoFaderVisible )
+    {
+        setTitle ( GetTitle() );
+    }
+}
+
 void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInfo )
 {
     // we want to set the server name only if the very first faders appear
     // in the audio mixer board to show a "try to connect" before
     if ( bNoFaderVisible )
     {
-        setTitle ( tr ( "Personal Mix at the Server: " ) + strServerName );
+        setTitle ( GetTitle() );
     }
 
     // get number of connected clients
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -740,7 +740,7 @@ void CServer::SetEnableRecording ( bool bNewEnableRecording )
 
 #if QT_VERSION >= QT_VERSION_CHECK(5, 5, 0)
 // TODO we should use the ConsoleWriterFactory() instead of qInfo()
-        qInfo() << "Recording state " << ( bEnableRecording ? "enabled" : "disabled" );
+        qInfo() << "Recording state" << ( bEnableRecording ? "enabled" : "disabled" );
 #endif
 
         if ( !bEnableRecording )
--- a/src/audiomixerboard.h
+++ b/src/audiomixerboard.h
@@ -164,6 +164,8 @@ public:
 
     void SetChannelLevels ( const CVector<uint16_t>& vecChannelLevel );
 
+    void SetRecorderState ( const ERecorderState newRecorderState );
+
     // settings
     CVector<QString> vecStoredFaderTags;
     CVector<int>     vecStoredFaderLevels;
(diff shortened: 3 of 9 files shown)


src/audiomixerboard.cpp
```diff
@@ -619,14 +619,24 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent, Qt::WindowFlags ) :
...
 {
     // add group box and hboxlayout
     QHBoxLayout* pGroupBoxLayout = new QHBoxLayout ( this );
     QWidget*     pMixerWidget    = new QWidget(); // will be added to the scroll area which is then the parent
     pScrollArea                  = new CMixerBoardScrollArea ( this );
     pMainLayout                  = new QHBoxLayout ( pMixerWidget );
 
+
+    setAccessibleName ( "Personal Mix at the Server groupbox" );
+    setWhatsThis ( "<b>" + tr( "Personal Mix at the Server" ) + "<b>: " +
```
pljones, line 633: Oops, closing `<b>` should be `</b>`... fixing.

src/audiomixerboard.cpp
```diff
@@ -807,13 +817,27 @@ void CAudioMixerBoard::ChangeFaderOrder ( const bool        bDoSort,
     }
 }
 
+QString CAudioMixerBoard::GetTitle()
+{
+    return "[" + csRecorderStateToString ( eRecorderState ) + "] " + tr ( "Personal Mix at the Server: " ) + strServerName;
```
corrados, line 822: I would prefer that if the recorder does not record, no additional string is show. Just in case the recorder actually records, a clear message apears, preferably in capital letters like "RECORDING ACTIVATED".

src/audiomixerboard.cpp
```diff
@@ -619,14 +619,24 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent, Qt::WindowFlags ) :
...
     QWidget*     pMixerWidget    = new QWidget(); // will be added to the scroll area which is then the parent
     pScrollArea                  = new CMixerBoardScrollArea ( this );
     pMainLayout                  = new QHBoxLayout ( pMixerWidget );
 
+
+    setAccessibleName ( "Personal Mix at the Server groupbox" );
+    setWhatsThis ( "<b>" + tr( "Personal Mix at the Server" ) + "</b>: " +
+        tr ( "When connected to a server, the controls here allow you to set " ) +
+        tr ( "your local mix without affecting what others hear from you.") +
+        tr ( "<br/>The title shows - when known - the server name and whether it is recording, "
```
corrados, line 636: It seems most users do not read help texts. They eventually just ask in the forum. Therefore it is no good idea to use these symbols. As written below, just show a clear text message and everybody knows whats going on. Then you do not need to explain anything here in the help text (or at least just a very short hint).

src/util.h
```diff
@@ -572,6 +572,21 @@ enum ERecorderState
     RS_RECORDING = 3
 };
 
+inline QString csRecorderStateToString ( ERecorderState eRecorderState )
```
corrados, line 575: I think this is not necessary. In GetTitle() just check for RS_RECORDING.

src/audiomixerboard.cpp
corrados, line 636: BTW: <br> should not be in the tr (  ) function.

pljones: Ah, yeah, `<br/>` will move :).

src/util.h
pljones: Yes, now it's just shown or not shown  Wasn't even sure when I had three outputs.

src/audiomixerboard.cpp
pljones: Done.

src/audiomixerboard.cpp
pljones: Done.

src/util.h
pljones: Removed.

src/audiomixerboard.cpp
pljones: Done.

src/audiomixerboard.cpp
```diff
@@ -807,13 +816,32 @@ void CAudioMixerBoard::ChangeFaderOrder ( const bool        bDoSort,
...
 }
 
+QString CAudioMixerBoard::GetTitle()
+{
+    QString myTitle = "";
+    if ( eRecorderState == RS_RECORDING )
+    {
+        myTitle = "[" + tr ( "RECORDING ACTIVE" )  + "] ";
+    }
+    return myTitle + tr ( "Personal Mix at the Server: " ) + strServerName;
```
pljones, line 826: Maybe (just to shorten it) remove "at the Server"?

corrados: Good idea. Yes, you can do it if you like.

src/server.cpp
```diff
@@ -740,7 +740,7 @@ void CServer::SetEnableRecording ( bool bNewEnableRecording )
 
 #if QT_VERSION >= QT_VERSION_CHECK(5, 5, 0)
 // TODO we should use the ConsoleWriterFactory() instead of qInfo()
-        qInfo() << "Recording state " << ( bEnableRecording ? "enabled" : "disabled" );
+        qInfo() << "Recording state" << ( bEnableRecording ? "enabled" : "disabled" );
```
corrados, line 743: Why did you remove the space here?

pljones: Because qInfo() adds a space.

src/audiomixerboard.cpp
pljones: Done.

src/server.cpp
corrados: Oh, I did not know that. Thanks for the info.

src/audiomixerboard.cpp
```diff
@@ -807,13 +816,32 @@ void CAudioMixerBoard::ChangeFaderOrder ( const bool        bDoSort,
...
+    {
+        myTitle = "[" + tr ( "RECORDING ACTIVE" )  + "] ";
+    }
+    return myTitle + tr ( "Personal Mix at: " ) + strServerName;
+}
+
+void CAudioMixerBoard::SetRecorderState ( const ERecorderState newRecorderState )
+{
+    eRecorderState = newRecorderState;
+    if ( !strServerName.isEmpty() && !bNoFaderVisible )
```
corrados, line 832: I have changed this in my "merge fix". Reason: The recorder state message cannot be received before the new connected client list so this check is not necessary. And if the server name is empty, we still should show the recorder state. Do you agree?

src/audiomixerboard.cpp
```diff
@@ -807,13 +816,32 @@ void CAudioMixerBoard::ChangeFaderOrder ( const bool        bDoSort,
     }
 }
 
+QString CAudioMixerBoard::GetTitle()
```
corrados, line 819: FYI: In my "merge fix" I renamed this to UpdateTitle(). The functionality should still be the same as yours.

src/audiomixerboard.cpp
corrados: I just saw in the code that indeed, the recorder state can be received before the connected client list. Anyway, it does not hurt to show the recorder state even if we do not yet have received the connected client list. What happens is that for a very short time you first see only the recorder state and then you see the recorder state plus the server title. So everything should be fine.

pljones: Hm... it should work.  If you connect to a non-existent server and have the "trying to connect" title, it's not going to get a recording status message...

(In the previous version of my code, at one point I was setting "Unknown" as the status and had added it to the "trying to connect", then realised what the bNoFaderVisible check did, so included it.  It shouldn't be necessary in this version.)

corrados: Ok, thanks for the confirmation. Do you have plans to use that new code for your drealm server (which is recording all the time) so that we can test the implementation?

pljones: Yes, just about to deploy it.

pljones: Done.

pljones:
Hm, just tried connecting to a 3.4.5git server with the 3.5.7git client and got told it was recording.  I may not be clearing the state in the client correctly.

PR #379, opened 2020-06-20
New Select switch to change several faders like one

Alberstein8:
With the new Select switch, faders of several channels can be adjusted at the same time to same level


corrados:
Thank you

corrados:
Wow, I was too quick with the merge. A lot of work to do to fix the feature up. Next time I spend more time on the review... ;-)

Alberstein8:
😢

corrados:
No problem, my fault. Thanks again for the feature :-).

Alberstein8:
I’m thinking to extend it to apply to pan as well but I’ll wait for your “adjustments” first.

corrados:
Hmm, does it make sense to apply it to the Pan as well? What is the use case?

Alberstein8:
Well, in a choir, for example, you could think about hearing male voices more on your right and female on your left, kind of stereo by voice/instruments.

Enviado desde mi iPhone

storeilly:
Would it be useful to mention here the use of separate ini files? Jamulus
stores pan, mute, solo, and fader levels for the last x (200 I think)
clients met. If you use a different ini file for your choir, the faders
will be the way you left them last time.

Best regards
Stephen

Issue #360, opened 2020-06-13
Active links in welcome message 


corrados:
I like to post a link to music and/or the video service in my server welcome message - but links don't work. So users then have to copy/paste the URL into a browser. Could links be made active? Thanks.

Copied from https://sourceforge.net/p/llcon/discussion/developerforum/thread/c4ae1cf17b

corrados:
I have raised a question in the following issue: #314 (comment) regarding this feature request.

corrados:
See: https://github.com/corrados/jamulus/issues/314#issuecomment-644191787

corrados:
Regarding security issues, please read also https://github.com/corrados/jamulus/pull/380#issuecomment-647042975

ann0see:
I‘d at least add an option somewhere to disable active links. Furthermore I‘d add a message which asks the user if he „really wants to visit https://www.example.com“

corrados:
This is solved by the following Issue: https://github.com/corrados/jamulus/issues/591 -> closed.

PR #380, opened 2020-06-20
HTML-escape chat messages at the server

atsampson:
At present, whatever text you include in your user name or chat messages is passed through directly to the chat display window and interpreted as HTML, which I don't think is intentional. This commit fixes it by having the server escape the message when it adds the channel name and timestamp.

It's not an ideal fix because the client is still trusting arbitrary HTML sent by the server (as it does for server welcome messages, and while external links don't work, cf #360, it can include things like `` elements that reference local files, which isn't very nice). So maybe it'd be worth thinking about introducing a new type of protocol message where the channel name and text are separate fields?

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1333,7 +1333,8 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int      iCurChanID
 
     const QString strActualMessageText =
         "<font color=""" + sCurColor + """>(" +
-        QTime::currentTime().toString ( "hh:mm:ss AP" ) + ") <b>" + ChanName +
+        QTime::currentTime().toString ( "hh:mm:ss AP" ) + ") <b>" +
+        ChanName.toHtmlEscaped() +
         "</b></font> " + strChatText;
 
 


corrados:
> At present, whatever text you include in your user name or chat messages is passed through directly to the chat display window and interpreted as HTML, which I don't think is intentional.

Actual it is intentional. E.g. a lot of users make use of HTML in their server welcome message.

Why do you want to change this behaviour? Have you seen issues when running your fuzzing tests? BTW: It is really great that you spend time on checking the security of the Jamulus software, thanks for that!

Regarding the channel name (i.e. your change `ChanName.toHtmlEscaped()`) I think it would be ok to strip out the HTML because here I really don't see a need for it.

atsampson:
I've updated the patch so it only escapes the channel name. (What got me looking at this was that I set my name to `<i>AdamS` to test the status file change, then tried all the other places where the name is shown and found it appeared in italics in the chat too...)

I'm not sure that QTextBrowser is really intended to be safe against people sending arbitrary bits of HTML to it, but at least [the subset of HTML it currently understands](https://doc.qt.io/qt-5/richtext-html-subset.html) is pretty small, so most of the things you can do look annoying rather than dangerous. 

One thing that concerns me is that you can make it open arbitrary local paths as resources - e.g. with something like `` or `<p style="background: url(/usr/share/pixmaps/xterm_48x48.xpm)">hello` (or indeed a never-ending device file, which hangs Jamulus). If it's possible to disable loading resources entirely then that would be a good idea - is it possible to override [loadResource](https://doc.qt.io/qt-5/qtextbrowser.html#loadResource) with a dummy implementation, or can you think of an easier way?

corrados:
Right now the chat window QTextBrowser does not have links or external links activated. So I assume that we do not have a problem right now. But we have a feature request: https://github.com/corrados/jamulus/issues/360 which want external links to work. I already have written in the comments that I have security concerns with it. I'll reference your comment in that Issue.

corrados:
Sorry, I was not reading your comment correctly. You are concerned not about the links but resources. Can you please create a separate Issue for that?

PR #381, opened 2020-06-20
Escape server and user names in the status HTML

atsampson:
Escape server and user names in the status HTML - currently if I set my username to something like `<i>AdamS` then it'll be passed through as is.

The security impact of this is limited by the name size limit that CProtocol enforces, and by this feature not being used on most servers anyway. (At worst, though, if somebody's ignored the sensible advice in the Tips page and directly included the output in a PHP file, it might be possible to execute arbitrary code on the webserver using crafted channel names.)


corrados:
Thank you!

PR #382, opened 2020-06-20
Fix two uninitialised reads

atsampson:
Found using valgrind with fuzzer-generated testcases.

### DIFF
--- a/src/signalhandler.cpp
+++ b/src/signalhandler.cpp
@@ -174,6 +174,7 @@ bool CSignalUnix::setSignalHandled ( int sigNum, bool state )
     else
     {
         sa.sa_handler = SIG_DFL;
+        sa.sa_flags = 0;
     }
 
     return ::sigaction ( sigNum, &sa, nullptr ) == 0;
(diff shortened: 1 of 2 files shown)


src/signalhandler.cpp
```diff
@@ -165,11 +165,12 @@ bool CSignalUnix::setSignalHandled ( int sigNum, bool state )
 {
     struct sigaction sa;
     sigemptyset ( &sa.sa_mask );
+    sa.sa_flags = 0;
 
     if ( state )
     {
         sa.sa_handler = CSignalUnix::signalHandler;
-        sa.sa_flags = SA_RESTART;
+        sa.sa_flags |= SA_RESTART;
```
pljones, line 173: Why?  It's set to zero.  0 | anything is the thing.  So you may as well leave this line unchanged.

atsampson: Will do if you prefer - I'm more used to seeing this with |= so you can add other flags earlier/later.

corrados: pljones, you are right. I also agree to atsampson that usually "|=" is used for these flags. If you ask me I would prefer atsampsons implementation. But I think since this is your code, pljones, you can decide on this :-)

corrados: Ups, it's already gone?

corrados:
Thank you.

PR #383, opened 2020-06-20
Fixed a bug that made the Server window stop updating

AronVietti:
Minimizing the Server Window on OSX stopped it from ever being
refreshed. Added a line to call show() when a changeEvent is
recieved and the window is not minimized.


corrados:
Wow, the code in changeEvent() is very old. Surprises me that no other user complains about this came in in the past. So it seems to work on Windows (and maybe Linux) but obviously not an MacOS. Thank you for fixing!

PR #386, opened 2020-06-21
Fix some more Valgrind warnings

atsampson:
Fix some more uninitialised reads and memory leaks reported by Valgrind and ASan in normal use of the client and server GUIs.

(The memory leak fixes don't really make a difference in practice because they're all objects that are only allocated once anyway, but it does make it easier to see legitimate problems.)


corrados:
Thanks.

PR #389, opened 2020-06-21
Updated jamulus.desktop.in file to include Spanish translation

ignotus666:


trebmuh:
@ignotus666 @corrados you don't need `Comment[es]=Jam Session` if you don't translate it. Without a localised translation, it'll take and display the default one (`Comment=Jam Session`).

PR #391, opened 2020-06-22
Feature/228 refactor jam recorder

pljones:
First part of the last part of #228 - this moves the Jam Recorder code out, as much as possible, from the server itself.  The part that needs to be in the same thread is in JamControl; JamRecorder is the recorder thread.

I'm currently running this version on jamulus.drealm.info - seems to work okay both with the existing UI and with `-n`.


corrados:
I am planning the next release to be on upcoming Sunday. I have some things in the queue which I want to include in that release. My plan is that I do a code review of your code as soon as this release is out. I hope this delay is ok for you.

pljones:
Sure, I'll keep rebasing as things change.

PR #395, opened 2020-06-24
Catch a CGenErr by const reference rather than by value

stweil:
See https://lgtm.com/projects/g/corrados/jamulus/alerts?mode=list for more alerts.


corrados:
Thank you.

Issue #342, opened 2020-06-08
Add a man page


ycollet:
A man page is always a good thing to have :)

gilgongo:
Bearing in mind only about 5% of downloads are Linux though :-)

ycollet:
Argh .... :)

trebmuh:
jamulus.1 will be the EN man page
jamulus.fr.1 will be the FR man page
...etc

pljones:
Ideally, the source code and translation strings that already exist could be used to generate the man pages.

There's also the signal handling I'd expect to be covered in the man page.

PR #405, opened 2020-06-28
UI to set (and clear) recording directory

pljones:
The UI change for the recording directory spec'd in #228.  I did, eventually, decide to put it above the session controls as you need the overall directory first.

I've extended the "What's this" on the status label, too, to be a bit more helpful.

----

Aside: `man jamulus` doesn't work.  I guess that's something else to think about.  Just basic command line options and signal handling, no description of how the UI works (that's the online manual's job).

^^
Please, no one reply to that - it's just so I've written it down _somewhere_...

### DIFF
--- a/src/serverdlg.cpp
+++ b/src/serverdlg.cpp
@@ -123,28 +123,63 @@ CServerDlg::CServerDlg ( CServer*         pNServP,
     cbxLocationCountry->setAccessibleName ( tr (
         "Country where the server is located combo box" ) );
 
+    // recording directory
+    pbtRecordingDir->setAccessibleName ( tr ( "Display dialog to select recording directory button" ) );
+    pbtRecordingDir->setWhatsThis ( "<b>" + tr ( "Main Recording Directory" ) + ":</b> "
+        +  tr ( "Click the button to open the dialog that allows the main recording directory to be selected."
+                "The chosen value must exist and be writeable (allow creation of sub-directories "
+                "by the user Jamulus is running as). " ) );
+
+    edtRecordingDir->setAccessibleName ( tr ( "Main recording directory text box (read-only)" ) );
+    edtRecordingDir->setWhatsThis ( "<b>" + tr ( "Main Recording Directory" ) + ":</b> "
+        +  tr ( "The current value of the main recording directory. "
+                "The chosen value must exist and be writeable (allow creation of sub-directories "
+                "by the user Jamulus is running as). "
+                "Click the button to open the dialog that allows the main recording directory to be selected." ) );
+
+    tbtClearRecordingDir->setAccessibleName ( tr ( "Clear the recording directory button" ) );
+    tbtClearRecordingDir->setWhatsThis ( "<b>" + tr ( "Clear Recording Directory" ) + ":</b> "
+        +  tr ( "Click the button to clear the currently selected recording directory. "
+                "This will prevent recording until a new value is selected.") );
+
     // enable recorder
     chbEnableRecorder->setAccessibleName( tr ( "Checkbox to turn on or off server recording" ) );
-    chbEnableRecorder->setWhatsThis( "<b>" + tr ( "Enable Recorder" ) + ":</b>"
+    chbEnableRecorder->setWhatsThis( "<b>" + tr ( "Enable Recorder" ) + ":</b> "
         + tr ( "Checked when the recorder is enabled, otherwise unchecked. "
                "The recorder will run when a session is in progress, if (set up correctly and) enabled." ) );
 
     // current session directory
     edtCurrentSessionDir->setAccessibleName( tr ( "Current session directory text box (read-only)" ) );
-    edtCurrentSessionDir->setWhatsThis( "<b>" + tr ( "Current Session Directory" ) + ":</b>"
-        +  tr ( "Enabled during recording and holds the current recording session directory. "
-                "Disabled after recording or when the recorder is not enabled." ) );
+    edtCurrentSessionDir->setWhatsThis( "<b>" + tr ( "Current Session Directory" ) + ":</b> "
+        + tr ( "Enabled during recording and holds the current recording session directory. "
+               "Disabled after recording or when the recorder is not enabled." ) );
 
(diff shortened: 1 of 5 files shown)


gilgongo:
That man page thought it here too #342

src/serverdlg.cpp
```diff
@@ -123,29 +123,64 @@ CServerDlg::CServerDlg ( CServer*         pNServP,
...
+    edtCurrentSessionDir->setWhatsThis( "<b>" + tr ( "Current Session Directory" ) + ":</b> "
+        + tr ( "Enabled during recording and holds the current recording session directory. "
+               "Disabled after recording or when the recorder is not enabled." ) );
 
     // recorder status
     lblRecorderStatus->setAccessibleName ( tr ( "Recorder status label" ) );
-    lblRecorderStatus->setWhatsThis ( "<b>" + tr ( "Recorder Status" ) + ":</b>"
-        +  tr ( "Displays the current status of the recorder." ) );
+    lblRecorderStatus->setWhatsThis ( "<b>" + tr ( "Recorder Status" ) + ":</b> "
+        + tr ( "Displays the current status of the recorder.  The following values are possible:"
```
corrados, line 160: Sorry for the delay...
You should avoid any HTML code to be in the tr() function. That simplifies the translation and makes things more consistent. Also, in the past I have used more HTML code but after doing the German translation I found out that the more HTML code you have the more difficult is the translation. So I recommend to use as little HTML as possible. E.g. I removed the <b> and other not absolutely required HTML from the translation files.

src/serverdlg.cpp
```diff
@@ -539,6 +571,30 @@ void CServerDlg::OnStopRecorder()
...
+    if ( newRecordingDir != currentValue )
+    {
+        pServer->SetRecordingDir ( newRecordingDir );
+        UpdateRecorderStatus ( QString::null );
+    }
+}
+
+void CServerDlg::OnClearRecordingDirClicked()
+{
+    if ( pServer->GetRecorderErrMsg() != QString::null || pServer->GetRecordingDir() != "" )
```
corrados, line 591: Why do you mix QString::null and ""?

src/serverdlg.cpp
pljones: Argh!  Sorry, yes.  The `<dl>` was a late addition...

src/serverdlg.cpp
pljones: The recording directory is an empty string because it should never be null.

The error message being null means "there is no error message" explicitly - not that the error message is an empty string.

src/serverdlg.cpp
```diff
@@ -157,24 +157,25 @@ CServerDlg::CServerDlg ( CServer*         pNServP,
...
     lblRecorderStatus->setWhatsThis ( "<b>" + tr ( "Recorder Status" ) + ":</b> "
-        + tr ( "Displays the current status of the recorder.  The following values are possible:"
-               "<dl>"
-               "<dt>Not initialised</dt>"
-               "<dd>No recording directory has been set or the value is not useable</dd>"
-               "<dt>Not enabled</dt>"
-               "<dd>Recording has been switched off" )
+        + tr ( "Displays the current status of the recorder.  The following values are possible:" )
+        + "<dl>"
+        + "<dt>" + tr ( SREC_NOT_INITIALISED ) + "</dt>"
```
pljones, line 162: These are defined in `serverdlg.h` untranslated and translated in the two places each string is used.  I couldn't immediately spot a similar example to copy, so I'm not certain this is the best approach - open to suggestions.

src/serverdlg.cpp
pljones: See update.

src/serverdlg.cpp
corrados: The Qt Linguist tool automatically finds the same text and uses the same translation. So please do not use the #defines but simply put the same text on the different locations in the code.

corrados: BTW: What does <dl> and <dt> do? In the other places in Jamulus we always use <ul> and <li>. Would that be an option for you, too? Just for consistency.

pljones: The use of the #define makes sure it actually _is_ the same text.  It gets expanded and translated (by Qt Linguist) in each place.  I'd rather keep it for consistency.  (Had there been an _actual_ status, then there would have been a method returning the string based on the status and each status could have been passed to that retrieve the string.)

`<dl>` is hypertext for "definition list".  This is a list of definitions of what the statuses mean.  So it's the appropriate list type.  It's not an "unordered list" (`<ul>`), which just gives a series of bullet points, or a "simple list" (`<sl>`).  `<dt>` is the term being defined and `<dd>` is the definition of the term.  The formatting this results in makes it easier to see and any accessibility tools will find it easier to handle the content.

corrados: So do you recommend to replace all the `<ul>` by `<dd>` then?

pljones: It depends on context.  If something is "just a list of items", then an unordered list makes sense.

For instance, saying features are:
* it does this
* it also does this

that's an unordered list.  But if you're explaining a list of terms, `<dl>` seems more appropriate than, say
* this term
means this
* another term
means something else

which is what you might do with `<ul>` to get the same effect.  The bullet doesn't feel right in the context.

The "Buffers" LED uses `<ul>` to supply a list of reasons - so it makes sense there - it's not defining anything.

The "Audio Channels" control uses `<ul>` -- but it's neither a definitions of terms from the dropdown nor even really a list of things.  So, whilst there are three entries in the dropdown and three bullets, the bullets aren't directly addressing each in turn, they are just a collection of comments generally addressing what they mean.  The use of `<ul>` at all here is almost uncalled for - just separate paragraphs would do.

Currently, on the server dialog, the server registration status doesn't list the strings that could appear at all (I hadn't thought of any way to keep them consistent easily, so didn't do it!), just makes a vague comment about what the field is for (which may help or may not...).

corrados: Ok, makes sense.

PR #407, opened 2020-06-28
Added code to log disconnection of users

storeilly:
Straightforward logging for disconnected users. Issue # 277


storeilly:
I missed something stupid, it's logging multiple times on unix. Reviewing....

atsampson:
The client can send multiple `PROTMESSID_CLM_DISCONNECTION` messages, since it'll do so whenever it gets a packet from the server it's not expecting. Guard the call to `AddDisconnected` with `if ( vecChannels[iCurChanID].IsConnected() )`?

(At the moment this'll only log when a client tells the server it's disconnecting; do you also want to log disconnections resulting from timeouts?)

corrados:
See my comment in the Issue regarding this pull request: https://github.com/corrados/jamulus/issues/277#issuecomment-651162973

corrados:
Since https://github.com/corrados/jamulus/issues/277 is closed, shouldn't we close this pull request as well?

storeilly:
Thanks Volker

PR #429, opened 2020-07-04
"Clear" symbol for recording dir

gilgongo:
As per discussion: https://github.com/corrados/jamulus/issues/228#issuecomment-653781054


src/serverdlg.cpp
```diff
@@ -330,7 +330,7 @@ lvwClients->setMinimumHeight ( 140 );
     pbtNewRecording->setAutoDefault ( false );
     pbtRecordingDir->setAutoDefault ( false );
     edtRecordingDir->setText( pServer->GetRecordingDir() );
-    tbtClearRecordingDir->setText ( "(X)" ); // someone make me a nice icon?
+    tbtClearRecordingDir->setText(QString::fromUtf8("\u232B"));
```
pljones, line 333: image
`<[X]` Hm.  I'd have left the comment... :D  Better than the (X), though.

PR #453, opened 2020-07-11
update Spanish translation

ignotus666:


corrados:
I saw that you have removed the "vanished" tags. After calling the Qt tool lupdate, all these were added again. So these texts are not existing anymore in the code.

ignotus666:
I didn't really realise that so many lines had been added with my commit because of the removed "vanished" tags. I just translated the few new strings in QTlinguist and didn't see anything strange there. So does anything need to be done?

corrados:
No, I don't think so. I just run the lupdate and that modified your .ts file. But I think everything should be just fine. I just wanted to tell it to you so that you know why these "vanished" tags are put back in the file.

PR #459, opened 2020-07-12
Update translation_sv_SE.ts

genesisproject2020:
Updated with new texts.


genesisproject2020:
Updated with new texts.

PR #460, opened 2020-07-14
Add 2 new instruments: Baritone and Lead vocals

AdrienDS:
Used in Barbershop and TTBB ensemble, as well as a-capella groups.

I left "Lead" in english in each translation because it's a common word, with a meaning too broad for google translate to translate it properly in each language, and an english word is better than a bad translation.

- For the Baritone image, I used the key from the Alto Vocal, and the face from the Bass Vocal.
- For the Lead image, I used the face from the Bass Vocal and the notes from [this copyright-free image (wikimedia)](https://commons.wikimedia.org/wiki/File:Africa_Music_ZP_8th_notes.svg)

Screenshot of the new voices option in the Profile dropdown


src/util.cpp
```diff
@@ -1107,6 +1107,8 @@ CVector<CInstPictures::CInstPictProps>& CInstPictures::GetTable ( const bool bRe
         vecDataBase.Add ( CInstPictProps ( QCoreApplication::translate ( "CMusProfDlg", "Congas" ), ":/png/instr/res/instruments/congas.png", IC_PERCUSSION_INSTRUMENT ) );
         vecDataBase.Add ( CInstPictProps ( QCoreApplication::translate ( "CMusProfDlg", "Bongo" ), ":/png/instr/res/instruments/bongo.png", IC_PERCUSSION_INSTRUMENT ) );
         vecDataBase.Add ( CInstPictProps ( QCoreApplication::translate ( "CMusProfDlg", "Vocal Bass" ), ":/png/instr/res/instruments/vocalbass.png", IC_OTHER_INSTRUMENT ) );
+        vecDataBase.Add ( CInstPictProps ( QCoreApplication::translate ( "CMusProfDlg", "Vocal Baritone" ), ":/png/instr/res/instruments/vocalbaritone.png", IC_OTHER_INSTRUMENT ) );
```
corrados, line 1110: This is not allowed. You always have to append new instruments in the list. They never can be added in the middle of the list. Please correct.

src/res/translation/translation_de_DE.ts
```diff
@@ -2461,36 +2461,46 @@
     </message>
     <message>
         <location filename="../../util.cpp" line="1110"/>
+        <source>Vocal Baritone</source>
+        <translation>Gesang Bariton</translation>
+    </message>
+    <message>
+        <location filename="../../util.cpp" line="1111"/>
+        <source>Vocal Lead</source>
+        <translation>Gesang Lead</translation>
```
corrados, line 2470: Good that you added the translations, too. But you have used words like "Lead" in the translations. I would translate it differently. The problem is that the translators will not easily see that there is an issue in the translation since only for missing translations you get a question mark in the Qt Linguist. For all translations which are marked as ready you will not easily find them. So for translations where you are not 100 % sure that they are correct, you should leave them empty instead so that the translators will find them easily.

corrados:
Thanks for your contribution. I have added two comments to your code.

src/res/translation/translation_de_DE.ts
AdrienDS: @corrados Thanks, so just leaving: `<translation></translation>`

src/util.cpp
AdrienDS: @corrados Ok, I will move them to the bottom of `vecDataBase`.

Just to be sure, the order doesn't matter in `src/resources.qrc`, right ?

corrados: Yes, it is correct that the order in resources.qrc does not matter.

src/res/translation/translation_de_DE.ts
melcon: I believe Qt Linguist will only show the question mark if the _translation_ element has the attribute _type_ set to _unfinished_: `<translation type="unfinished"></translation>`

corrados: Oh, yes. Thanks for that hint. That's right.
@AdrienDS Can you please fix that?

PR #469, opened 2020-07-19
connect dialog: remove spaces from the parsed string in the the "Serv…

dingodoppelt:
remove any spaces from the parsed network address from the combo box to improve usability


dingodoppelt:
With these little additions Jamulus will connect to an existing server even if the user accidentally puts whitespace into the address by removing any whitespace from the user input.

corrados:
Good idea. Thanks.

PR #477, opened 2020-07-26
change windowTitle when locally muted

### LINKED ISSUE #476: Show if the client is locally muted in the window title
Sometimes I forget to unmute myself, especially when the client window is not visible or I sometimes just want to know, if I am muted or not without bringing the client window to the front to be able to check.
I've created a quick fix for this and created a pull request for everybody to review.
cheers, nils

dingodoppelt:
fixes #476


dingodoppelt:
this code produces not the desired effect, I'll have to go over it again.

PR #485, opened 2020-07-31
Adding KeyEvent for execute Mute Myself Function By pressing "M" button on Keyboard

dzpex:
Mute Myself Function is very useful for me.
I had thinked to introduce the possibility to use a keyboard shortcut to call this function rapidly without the mouse.
I hope that this function are useful for others Jamulus users.


corrados:
Thanks but this is already been worked on, see https://sourceforge.net/p/llcon/discussion/533517/thread/f3252db0f5/?limit=25#242c/56e5/5971/873b/a953/7921 and https://github.com/corrados/jamulus/commit/ec378919d69975895bae1ae99f0c9f09688bc9b6

gilgongo:
Just so I'm clear: v3.5.10 will have the following shortcut keys, is that right?

"Mute myself" Alt+m (Cmd+m Mac)
"Settings" window Alt+s (Cmd+s)
"Chat" windows Alt+c (Cmd+c)

corrados:
> "Chat" windows Alt+c (Cmd+c)

This is "h" instead of "c" -> "Chat" windows Alt+h (Cmd+h)

Under Windows you can press the "Alt" key and then you can see that the corresponding shortcut character is underlined in the text of the check box.

PR #491, opened 2020-08-04
Set thread names, drop JamRecorder priority

pljones:
Add the thread names to make the `ps -o spid,comm` output easier to follow.


src/server.cpp
```diff
@@ -167,7 +171,12 @@ void CHighPrecisionTimer::Start()
 #endif
 
         // start thread
+#ifndef SCHED_RR
         QThread::start ( QThread::TimeCriticalPriority );
+#else
+        // priorty is "ignored" on linux: actually broken
```
corrados, line 177: But you set "QThread::NormalPriority" in your jam recorder. Why is that working and the "QThread::TimeCriticalPriority" not? That may be an issue with user rights? Maybe the Linux user running the Jamulus server does not have enough rights to get this priority?
But maybe this is a bug in Qt. Have you tried to use a newer Qt version?

src/server.h
```diff
@@ -34,6 +34,9 @@
 # include <QtConcurrent>
 # include <QFutureSynchronizer>
 #endif
+#ifndef _WIN32
+# include <sched.h>
```
corrados, line 38: For what purpose do we need this new header?

src/server.cpp
pljones: It's not "working" as such -- passing _any_ value drops real time and assumes SCHED_OTHER with nice zero (i.e. default scheduling values).

There's a statement in the QThread documentation that it doesn't work on Linux.
https://doc.qt.io/qt-5/qthread.html#start
>  In particular, the priority will be ignored on systems that do not support thread priorities (such as on Linux, see the sched_setscheduler documentation for more details).
Now, I'd say it is a _bug_ in Qt because it's causing an undesirable change rather than being "ignored".

I'm investigating workarounds.

https://man7.org/linux/man-pages/man2/setpriority.2.html works on threads, as far as I understand it -- pass the thread ID and it'll set the nice value (for SCHED_OTHER and SCHED_BATCH).
https://man7.org/linux/man-pages/man2/sched_setparam.2.html works on threads -- but allows SCHED_FIFO and SCHED_RR priorities to be set

I've not worked with these calls before, so I'm struggling to get a _working_ example -- and I want it to _not_ require "root", so it should work with CAP_SYS_NICE:
https://man7.org/linux/man-pages/man7/capabilities.7.html

src/server.h
pljones: For the `#ifndef SCHED_RR` check.  `global.h` pulls in `unistd.h`, so I might be better protecting this include with `#ifdef POSIX` or something.

src/server.cpp
```diff
@@ -58,6 +58,8 @@ CHighPrecisionTimer::CHighPrecisionTimer ( const bool bNewUseDoubleSystemFrameSi
     veciTimeOutIntervals[1] = 1;
     veciTimeOutIntervals[2] = 0;
 
+    setObjectName ( "CHighPrecisionTimer" );
```
corrados, line 61: I don't think we need this for Windows.

src/server.cpp
corrados: > passing any value drops real time and assumes SCHED_OTHER with nice zero

That is very interesting... According to the Qt description it should ignore the value if it is not supported on the OS. So that seems really be a Qt issue. What Qt version do you currently use?

src/server.h
corrados: Why not using #ifdef POSIX at the place where you want to avoid setting the "TimeCriticalPriority"?

src/soundbase.cpp
```diff
@@ -41,6 +41,8 @@ CSoundBase::CSoundBase ( const QString& strNewSystemDriverTechniqueName,
     lNumDevs          = 1;
     strDriverNames[0] = strSystemDriverTechniqueName;
 
+    setObjectName ( "CSoundBase::" + strSystemDriverTechniqueName );
```
corrados, line 44: Do we see this as a separate thread under linux? I thought you will only see the Jack audio thread?

WolfganP:
In https://github.com/corrados/jamulus/issues/455#issuecomment-669357386 test with this patch included, one of the threads' name is `Thread (pooled)` and appear several times.
May be some other thread is still unlabeled?

pljones:
> In [#455 (comment)](https://github.com/corrados/jamulus/issues/455#issuecomment-669357386) test with this patch included, one of the threads' name is `Thread (pooled)` and appear several times.  May be some other thread is still unlabeled?

This was written before the `multithreading` patch, so no threads relating to that have been named.

src/server.cpp
pljones: Qt 5.9 here.

src/server.h
pljones: Because I was planning deep, dark magic if SCHED_RR were defined to use pthread_setpriority.  I've not got that working, though.

src/server.cpp
pljones: I wonder what QThread does on Windows... and how you monitor threads -- surely there's some way?

src/soundbase.cpp
pljones: Isn't this thread created from linux/sound.cpp?  This then should say "CSoundBase::jack" or whatever gets passed in.  Client-side isn't currently so interesting, though.

src/server.cpp
pljones: Looks like someone is maintaining backports for Ubuntu 18.04 LTS:
https://launchpad.net/~beineri
with 5.14.2 being the latest:
https://launchpad.net/~beineri/+archive/ubuntu/opt-qt-5.14.2-bionic

(Later ones for later versions of Ubuntu look like they were only released recently, so I'm hopeful for 18.04 LTS.  My 18.04 "end of life" plan kicks in later this year.)

src/server.h
corrados: Ok, it sound like there is some more work todo. Maybe it is a good idea that I create a branch from master called "multithreading_tests" where I merge your code to so that we have some kind of "playground". The reason why I am a bit careful is because changing thread priorities could have unwanted side effects. I would like to change thread priorities only if we can say for sure that the server really runs better/more stable with the new code.

src/server.cpp
corrados: Qt 5.9 is quite old. Just yesterday I upgraded my Ubuntu laptop to the newest Ubuntu version without any problems. I can recommend to do it.

src/server.cpp
corrados: We should focus on Linux for now. Interestingly, even without your patch I get the information about the highprecisiontimer in my htop:
grafik
So maybe there is no need to set it at all (even not for Linux).

src/soundbase.cpp
corrados: Yes, it's client only. We should focus on the server for now. So no need to set an object name here.

src/server.cpp
pljones: I'll be taking it but not until I've a a good couple of free days - which means not for a couple of months, which fits nicely with my preference for leaving it a while after the LTS .1 has been out.

src/server.h
pljones: Definitely - with the new QtConcurrent model, I'm wondering if it's worth re-doing the QThread approach entirely.  QtConcurrent::run() seems fairly similar to moving a task to a thread and calling QThread::start() on that thread -- but much clearer.  I'll continue reading about it to see if I can get a clear idea on the QtConcurrent view on priority.

src/server.cpp
pljones: Ah... maybe it's something like this:
```
QObject qObject; // or whatever
QThread qThread;
qObject->moveToThread(qThread);
qThread->start();
```
will give you a default thread name of `QThread` (your thread's class).  But `CHighPrecisionTimer` extends `QThread`, more like (from memory):
```
CHighPrecisionTimer timerThread;
timerThread.start();
```
So I guess my setObjectName here to the same thing is rather redundant :).

src/server.h
corrados: Ok, I have just created a new branch for multithreading_improvements and will merge your branch in that one.

corrados: > Because I was planning deep, dark magic if SCHED_RR were defined to use pthread_setpriority. I've not got that working, though.

Are you still progressing on that? Or is it on hold? I ask because I want to get rid of the # include<sched.h> for now. If you plan to continue your implementation, you may add it later on. So again, would it make sense to replace SCHED_RR by POSIX?

Issue #498, opened 2020-08-07
Pan Values Are Not Announced by Screen Reader


chigkim:
In audiomixerboard.cpp, the description is set `pPan->setAccessibleName ( tr ( "Local panning position of the current audio channel at the server" ) );`.
However, it never gets updated, so when keyboard focus lands on a pan control, screen reader just reads the above phrase even if you change the value with left/right arrows.
Need to fix CChannelFader::OnPanValueChanged to include `pPan->setAccessibleName ( QString::number(iPan) );`
This should trigger screen reader to announce new value when you change a pan with left/right arrows.
Also need to fix when initial pan value is set upon joining a server.
There are a number of blind Jamulus users who would love to see more accessibility improvement. Right now I'm just hacking things quickly and providing a custom build for them, so they could participate.
Theoretically, I could fork and send pull requests, but I'm blind myself, so I don't know if my hacks would alter the UI visually as well.
Mean time, I'll continue to open more issues and provide related snippets like this.
Happy to discuss on Jamulus if anyone is interested or has more ideas.

corrados:
That is strange. I would have thought the Qt deals with such thing. Why is the application developer responsible for adding the QDial values to the accessible name. Maybe a Qt bug? Or something is wrong with the way Jamulus uses the Qt accessibility feature of Qt.

chigkim:
Yes, it is strange because the volume control is working fine, but not pan.

chigkim:
I finally got my windows environment to build. On windows, you can't even focus on the pan control with tab for some reason. It just skips it.

corrados:
I just merged your pull request -> closed

PR #511, opened 2020-08-10
Accessibility: Pan Value

chigkim:
This is for issue #498.

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -432,6 +432,7 @@ void CChannelFader::SetPanValue ( const int iPan )
         // we set the new fader level in the GUI (slider control) which then
         // emits to signal to tell the server about the change (implicitly)
         pPan->setValue ( iPan );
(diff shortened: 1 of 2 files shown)


src/audiomixerboard.cpp
```diff
@@ -432,6 +432,7 @@ void CChannelFader::SetPanValue ( const int iPan )
         // we set the new fader level in the GUI (slider control) which then
         // emits to signal to tell the server about the change (implicitly)
         pPan->setValue ( iPan );
+        pPan->setAccessibleName( QString::number( iPan ) );
```
corrados, line 435: Why is it setAccessibleName and not setAccessibleDescription here?

chigkim:
If you use setAccessibilityDescription, users have to refocus voiceover 
(VO) cursor to read the newly changed value.

However, if you use setAccessibilityName, VO announces as the value 
changes without having to refocus.

PR #513, opened 2020-08-11
Store recording directory

pljones:
One problem.  Once set, there's no way from the command line to say "clear the base recording dir".  You can change it, but not clear it. (Same goes for some other strings.)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -100,7 +100,7 @@ int main ( int argc, char** argv )
         {
             bIsClient = false;
             tsConsole << "- server mode chosen" << endl;
-            CommandLineOptions += "--server";
+            CommandLineOptions << "--server";
             continue;
         }
 
@@ -113,7 +113,7 @@ int main ( int argc, char** argv )
         {
             bUseGUI = false;
             tsConsole << "- no GUI mode chosen" << endl;
-            CommandLineOptions += "--nogui";
+            CommandLineOptions << "--nogui";
             continue;
         }
 
@@ -127,7 +127,7 @@ int main ( int argc, char** argv )
             // right now only the creative commons licence is supported
             eLicenceType = LT_CREATIVECOMMONS;
             tsConsole << "- licence required" << endl;
-            CommandLineOptions += "--licence";
+            CommandLineOptions << "--licence";
             continue;
         }
 
@@ -140,7 +140,7 @@ int main ( int argc, char** argv )
         {
             bUseDoubleSystemFrameSize = false; // 64 samples frame size
             tsConsole << "- using " << SYSTEM_FRAME_SIZE_SAMPLES << " samples frame size mode" << endl;
-            CommandLineOptions += "--fastupdate";
+            CommandLineOptions << "--fastupdate";
             continue;
         }
 
@@ -161,7 +161,7 @@ int main ( int argc, char** argv )
             tsConsole << "- maximum number of channels: "
                 << iNumServerChannels << endl;
 
-            CommandLineOptions += "--numchannels";
+            CommandLineOptions << "--numchannels";
             continue;
         }
 
@@ -174,7 +174,7 @@ int main ( int argc, char** argv )
         {
             bStartMinimized = true;
             tsConsole << "- start minimized enabled" << endl;
-            CommandLineOptions += "--startminimized";
+            CommandLineOptions << "--startminimized";
             continue;
         }
 
@@ -187,7 +187,7 @@ int main ( int argc, char** argv )
         {
             bCentServPingServerInList = true;
             tsConsole << "- ping servers in slave server list" << endl;
-            CommandLineOptions += "--pingservers";
+            CommandLineOptions << "--pingservers";
             continue;
         }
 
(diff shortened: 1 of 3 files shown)


corrados:
Please use recordingdir instead of recording_dir. I am only using _ to show that it is base 64.

> You can change it, but not clear it.

Do you want to fix it before I merge it?

pljones:
> Do you want to fix it before I merge it?

I'd need to be changing `main.cpp`, too.  Should I change all the settings that say
* "If the target value is not empty, use the inifile value"
so they say 
* "If the target value is not empty and the command line did not set it to empty, use the inifile value"?

(I was thinking of first looping through the command line for an inifile setting, else using the default, then looping through the other known command line options, setting the target value if present else using the inifile value; then any _other_ inifile settings - i.e. ones with no matching command line option - would get loaded.  Cumbersome...)

pljones:
Alternative approach:
* in `main.cpp`, when looping over the command line arguments, a "found" list is maintained (list of strings), with each matched entry being added by its long option name (e.g. `--recording`)
* when calling Settings.Load, you'd pass the list and not use the inifile value for anything overridden on the command line

corrados:
Good idea with the string list, thanks. I'll add some minor comments in your code now.

src/main.cpp
```diff
@@ -99,6 +100,7 @@ int main ( int argc, char** argv )
         {
             bIsClient = false;
             tsConsole << "- server mode chosen" << endl;
+            CommandLineOptions += "--server";
```
corrados, line 103: Should we only set the strings for command line arguments which are actually used in the settings class? I actually prefer to add them all as you did but the other solution would give less code and would make clear that the string is actually used. What is your opinion?

src/main.cpp
```diff
@@ -111,6 +113,7 @@ int main ( int argc, char** argv )
         {
             bUseGUI = false;
             tsConsole << "- no GUI mode chosen" << endl;
+            CommandLineOptions += "--nogui";
```
corrados, line 116: At the places in Jamulus where I have used the QStringList class, I have used the operator<< instead of the +=. Does the += give a different result? If not, for consistency, we should use << here, too.

src/settings.cpp
```diff
@@ -843,13 +849,18 @@ if ( GetFlagIniSet ( IniXMLDocument, "server", "defcentservaddr", bValue ) )
...
+         if ( GetFlagIniSet ( IniXMLDocument, "server", "autostartmin", bValue ) )
+         {
+             pServer->SetAutoRunMinimized ( bValue );
+         }
     }
 
     // licence type (command line overwrites setting file)
-    if ( pServer->GetLicenceType() == LT_NO_LICENCE )
+    //if ( pServer->GetLicenceType() == LT_NO_LICENCE )
+    // TODO: command line needs to become boolean, at least...
```
corrados, line 862: Why should the command line be a boolean value? It is actually a boolean value. Do you want the command line to _not_ be a boolean value instead?

src/settings.cpp
```diff
@@ -859,14 +870,22 @@ if ( GetFlagIniSet ( IniXMLDocument, "server", "defcentservaddr", bValue ) )
...
+    if ( !CommandLineOptions.contains( "--welcomemessage" ) )
     {
         pServer->SetWelcomeMessage ( FromBase64ToString ( GetIniSetting ( IniXMLDocument, "server", "welcome" ) ) );
     }
 
     // window position of the main window
     vecWindowPosMain = FromBase64ToByteArray (
         GetIniSetting ( IniXMLDocument, "server", "winposmain_base64" ) );
+
+    // base recording directory (command line overwrites setting file)
```
corrados, line 883: Since now all settings are overwritten by the command line, maybe we should remove all the "_(command line overwrites setting file)_", what do you think?

src/main.cpp
pljones: I think there should be "greater equality" between command line and GUI - at the moment, there's a lot that the UI can do (particular the client) that can't be set on start up from the command line.  I'd also like to see most if not all of what can be set from the command line available in the UI.  And then the UI settings should all get loaded / saved to the inifile.

So it seemed logical to help that process by putting this in consistently, at least.

src/main.cpp
pljones: Oh - I was actually surprised `+=` worked.  My C backgound -- `<<` is very C++ ;).  Yes, `<<` as append seems much better!  I'll fix that now.

src/settings.cpp
pljones: Maybe I misread it -- I thought it was only checking for presence, rather than value.

src/settings.cpp
pljones: It could go as a comment on the call in `main.cpp` - I think it needs mentioning.

src/main.cpp
corrados: Ok, agreed.

src/settings.cpp
corrados: Ok, let's do that later on. I'll now merge your code.

corrados: I just saw that you already put it in main.cpp -> done.

PR #529, opened 2020-08-18
Fix for crash when using the JACK backend and quickly reconfiguring

hselasky:
Found some bugs here and there and managed to improve overall jitter with my audio card.

### DIFF
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -240,6 +240,9 @@ int CSound::process ( jack_nframes_t nframes, void* arg )
     CSound* pSound = static_cast<CSound*> ( arg );
     int     i;
 
+    // make sure we are locked during execution
+    QMutexLocker locker(&pSound->MutexAudioProcessCb);
+
     if ( pSound->IsRunning() && ( nframes == static_cast<jack_nframes_t> ( pSound->iJACKBufferSizeMono ) ) )
     {
         // get input data pointer
--- a/src/soundbase.cpp
+++ b/src/soundbase.cpp
@@ -75,15 +75,15 @@ void CSoundBase::Stop()
     // set flag so that thread can leave the main loop
     bRun = false;
 
-    // give thread some time to terminate
-    if ( !bIsCallbackAudioInterface )
-    {
-        wait ( 5000 );
-    }
+    // drain the audio process callback
+    QMutexLocker locker(&MutexAudioProcessCb);
 }
 
 void CSoundBase::run()
 {
+    // make sure we are locked during execution
+    QMutexLocker locker(&MutexAudioProcessCb);
+
     // main loop of working thread
     while ( bRun )
     {
--- a/src/soundbase.h
+++ b/src/soundbase.h
@@ -26,6 +26,7 @@
 
 #include <QThread>
 #include <QString>
+#include <QMutex>
 #ifndef HEADLESS
 # include <QMessageBox>
 #endif
@@ -137,6 +138,7 @@ protected:
 
     void run();
     bool bRun;
+    QMutex MutexAudioProcessCb;
 
     void             ParseMIDIMessage ( const CVector<uint8_t>& vMIDIPaketBytes );
 


corrados:
Thanks for your code. I'll put some comments to your new code.

src/client.h
```diff
@@ -368,10 +368,15 @@ class CClient : public QObject
     int                     iServerSockBufNumFrames;
 
     // for ping measurement
-    CPreciseTime            PreciseTime;
+    QElapsedTimer           PreciseTime;
```
corrados, line 371: Why did you change this? Was the ping time incorrect? How does this change improve the functionality?

src/client.cpp
```diff
@@ -762,7 +766,12 @@ void CClient::Init()
     const int iPrefMonoFrameSize = iSndCrdPrefFrameSizeFactor * SYSTEM_FRAME_SIZE_SAMPLES;
 
     // get actual sound card buffer size using preferred size
-    iMonoBlockSizeSam = Sound.Init ( iPrefMonoFrameSize );
+    const int iSystemMonoSizeSam = Sound.Init ( iPrefMonoFrameSize );
+
+    // prevent any early audio callbacks during configuration
```
corrados, line 771: That should never happen. The Init() function is only called if the Sound.Stop() was called before. Can you please give me more information in which case you saw a crash here?

src/client.cpp
```diff
@@ -939,6 +948,37 @@ void CClient::AudioCallback ( CVector<int16_t>& psData, void* arg )
     // get the pointer to the object
     CClient* pMyClientObj = static_cast<CClient*> ( arg );
 
+    QMutexLocker locker(&pMyClientObj->MutexAudioProcess);
+    if ( pMyClientObj->Sound.IsRunning() == false )
+    {
+        return;
+    }
+
+    // In common audio subsystems, there may be multiple audio sizes in use.
```
corrados, line 957: Interesting concept. Is this Linux-specific or does this apply to all supported OSs?

src/main.cpp
```diff
@@ -741,6 +741,9 @@ QString UsageArguments ( char **argv )
         "  -j, --nojackconnect   disable auto Jack connections\n"
         "  --ctrlmidich          MIDI controller channel to listen\n"
         "  --clientname          client name (window title and jack client name)\n"
+#ifndef HEADLESS
```
corrados, line 744: I actually don't want to this to be shown in the official command line options. It is just a "pro tool" for developers which want to optimize internal algorithm parameters. For "normal" users this would just confuse them. It is what it it: a hidden development feature.

src/util.h
```diff
@@ -1299,34 +1299,6 @@ class MathUtils
 };
 
 
-// Precise time ----------------------------------------------------------------
-// required for ping measurement
-class CPreciseTime
```
corrados, line 1304: What about the precision of the QElapsedTimer? Have you tested the precision on all supported platforms (Windows, Linux and MacOS)? As far as I remember (it is a very long time ago when I implemented that), I think I experimented with the Qt internal classes but the precision was not good enough. That was the reason I have implemented my own class.

src/client.h
hselasky: I saw QElapsedTimer used another place and thought there is no need to duplicate the OS checks for various timing implementations. No functional difference.

src/client.cpp
hselasky: I think the JACK backend doesn't have any stop function implemented. It just lazily sets a flag, while yet the function may still be processing. I found this when working on another patch, where I could change the buffersizes used by Jamulus from the GUI when using JACK audio backend, and it core-dumped over and over again :-(

hselasky:
> Thanks for your code. I'll put some comments to your new code.

Thank you!

src/client.h
hselasky: I checked: src/corelib/kernel/qelapsedtimer_mac.cpp, src/corelib/kernel/qelapsedtimer_win.cpp

Mac implementation is the same like yours (nanoseconds) Windows is no worse either (milliseconds).

src/util.h
hselasky: I checked: src/corelib/kernel/qelapsedtimer_mac.cpp, src/corelib/kernel/qelapsedtimer_win.cpp

Mac implementation is the same like yours (nanoseconds) Windows is no worse either (milliseconds).

src/client.cpp
hselasky: It applies to all OS'es where Jamulus is not the only client of the sound card subsystem.

Usually sound mixers have a pre-defined and fixed interval for the buffering. 

Also further, USB hardware only supports small buffers of size of 2ms 3ms 4ms and so on. When Jamulus specifies 64 samples of buffer, the USB hardware cannot technically program that! So you end up with audio bursts, that in turn affect the timing of the UDP packets, and I found this to be the root cause of all jitter problems during my tests with FreeBSD. Too bad Windows doesn't have a nanosecond clock. I'm not fully sure how this patch will work there. Maybe someone can help testing?

src/main.cpp
hselasky: OK. I'm fine with that. Should I drop this chunk then?

src/client.cpp
corrados: In CSound::process there is a check implemented: pSound->IsRunning(). If pSound->IsRunning() is false, pSound->ProcessCallback is not called. So there is already a protection implemented. We have to find out why this protection did not work for your case. Can you please give more information of what you did when it crashed? If you say "I found this when working on another patch", maybe you have changed something in the Jamulus code so that it crashed.

corrados: I still think that "MutexAudioProcess" is not necessary for the current Jamulus source code on the Git master branch.

src/client.cpp
corrados: I can do it when I am back at home in a week. If Windows does not support it, we should move your code in the Jack Sound.cpp file for now. Actually with my audio devices I tested, I did not have any higher jitter from the audio driver. What audio hardware are you using? And how did you measure the jitter?

src/util.h
corrados: Ok, so there was no issue with the functionality but you just want to improve the code. Maybe QElapsedTimer has improved since I tested it a long time ago. But I have to check the ping timer precision when I have access to my PC again (in about a week) to verify that the functionality does not changed by this replacement.

src/main.cpp
corrados: Yes, please.

src/client.cpp
hselasky: I have a X32-RACK audio device configured with 8-channel audio, where I use 2 for Jamulus.  I measured the jitter by using wireshark and looking at the time between data packets going on the wire.

src/util.h
hselasky: Sure.

src/client.cpp
hselasky: The problem is not new audio callbacks, but what if an audio callback is already pending when doing init? Then the check is ineffective.

hselasky: At leas the JACK driver does not drain the audio callback, and if there is a callback process, then there is just simple sleep of 5 ms.

hselasky: I fixed this in the sound base class instead like you suggested. When we stop() then aquire the new mutex, and drop it. After that the callback should always see the correct value of the running flag! See updated patch.

hselasky:
I added in one more patch. I hope it is not too much :-)

WolfganP:
Switching from int to float wouldn't add much processing time? (usually int arithmetic is way faster than float)

corrados:
> I added in one more patch. I hope it is not too much :-)

Well, actually it is. I would prefer smaller pull requests.

corrados:
> Switch all Jamulus audio sample processing to use floats instead of...

Can you please give some more background information why you did this?

corrados:
This is what I get on my Lexicon Omega under Windows:
grafik
So it gives me a jitter of about 1 ms.

hselasky:
Something like that is expected. It likely means the USB stack in Windows is using a 1ms audio buffer, which is a requirement by the XHCI USB host controller.

hselasky:
@corrados : I have the plans ready for a cool jitter correction algorithm. All I need is one bit per frame, and I see the first bit of every OPUS frame appears unused, so I'm asking for permission to use that bit for something!

corrados:
Can you please give some more info for what reason you need the bit?

hselasky:
@corrados : I want to implement a "noise-based" (not sure what you would call it) sequence number for every OPUS frame. That means there is for example a 13-bit frame number incrementing one by one, then via a formula only one of the 13-bits is or'ed into every OPUS frame.

The reason we need a frame number is because it is very difficult to estimate the packet loss accurately when there is not timing information provided by every UDP packet, from the sender side.

Also 48kHz is not 48kHz. Over time the clocks will drift (sender and receiver side), but if we had a sequence number, this clock drift could easily be eliminated by libsamplerate on the client side!

Without a sequence number it is impossible to know how many packets were dropped on the wire.

--HPS

corrados:
This goes in a wrong direction. The idea of Jamulus is to keep it simple. About 10 years back I also had done tests with a resampling but that did not give any improvement. The idea is: if a packet is lost, it is lost. No need to track the packet number. You can read more about this in http://llcon.sourceforge.net/PerformingBandRehearsalsontheInternetWithJamulus.pdf

If you want to include a resampler, you have a problem that we are working on blocks of data. If you have a clock difference, you would have to resample and get, e.g., 129 samples out of the resampler instead of 128. But you have to transmit 128. So what to do with the additional sample? You would have to introduce a new buffer. So you will have additional latency. To only meaningful clock correction would be hardware based (maybe GPS-based). But this is out of scope for the Jamulus software.

> it is very difficult to estimate the packet loss

For what reason do you need the packet loss for improving the sound card jitter?

hselasky:
I understand that Jamulus doesn't care about:

a) reordering of packets
b) lost packets

Using the OPUS codec to conceal jitter might be more time-clever than using libsamplerate.
OK, agreed, let's forget about libsamplerate.

> For what reason do you need the packet loss for improving the sound card jitter?

When you know the exact clock rate differences, you can then subtract or add that to the sound card jitter, to reduce the size of the jitter buffer needed.

Right now in the "perfect" link scenario all clock differences are accounted as jitter. It is like Jamulus needs packet loss to keep the jitter buffer down!

You asked about testing a Raspberry PI, and what I found is that the system clock is not that accurate.

The RPI for example generates 751 64-sample OPUS packets per second, but the client only consumes 750 per second.
So every second with "perfect" transmission the jitter buffer grows by 1 packet . This also adds up for the latency - right?

Even for 1 second of continous data, we could reduce the latency by 1.33 ms, if we knew some sequence numbers here and there, which is possible by filling the first bit of every OPUS frame with a predictable "noise" pattern.

I think it is simply wrong to assume that "the internet connection" will always have packet drops.

corrados:
> When you know the exact clock rate differences, you can then subtract or add that to the sound card jitter, to reduce the size of the jitter buffer needed.

Right, you could do that. But then you have to drop a packet from time to time to correct for the different clock rates. The jitter buffer just does exactly this anyway.

> So every second with "perfect" transmission the jitter buffer grows by 1 packet . This also adds up for the latency - right?

This depends on the selected jitter buffer size. If the jitter buffer size is set so that it corrects 1 packet per second, then you have exactly the same behavior.

> I think it is simply wrong to assume that "the internet connection" will always have packet drops.

Sure, you cannot say that the network always drops packet (do I do that in my paper? ;-) ). It depends on what type of network you have. There are more reliable connections and less reliable connections. E.g. over WLAN you will get packet loss in most cases.

Usually Jamulus users will have network issues by other devices overloading the network with updates, video streams, etc. In that case you will not be able to estimate the clock offset reliably. Just for the special case for a near perfect network condition you will have a chance to estimate it and drop some audio blocks at the right time to compensate for the clock offsets. But again, this is what the jitter buffer already does.

corrados:
Here is the jitter I get if I use the Lexicon Omega under MacOS with 128 samples block size:
grafik
And this is with 64 samples block size:
grafik
As you can see, I have very low jitter.
When I do my Jamulus jam sessions, I usually do it with 64 samples under MacOS. That gives me a very good performance.

hselasky:
I see. Let me do a proof of concept on an own branch, and see if it makes any difference, hopefully w/o breaking too much of the established concepts.

hselasky:
> over WLAN you will get packet loss in most cases.

Yeah. Been there done that. Does not work, even with expensive ones like from Ubiquity.

hselasky:
I reduced this pull request to only contain two small fixes. Hopefully increasing the chance of being merged.

corrados:
Can you please split the pull request further? That makes the integration and testing easiert. One pull request for each "topic" like e.g.  "replacing the presice timer for ping" would be great. Thanks.

hselasky:
@corrados : Yes.

PR #538, opened 2020-08-25
Added code to get jack latency

bflamig:
After studying and using TheStk/RtAudio, I figured out how to get the latencies for audio devices when using Jamulus under Linux/Jack. The changes here were tested using two separate audio interfaces connected to a Raspberry Pi.

### DIFF
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -25,6 +25,7 @@
 \******************************************************************************/
 
 #include "sound.h"
+//#include <iostream> // For debugging total latency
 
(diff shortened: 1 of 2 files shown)


bflamig:
Great! 

Noobie question: By "delete this fork" you mean use the "danger zone" of the settings dialog for that fork (blfamig/jamulus) and tell it to delete the fork's repository?

corrados:
I have not used that but I assume you are right, the fork would be deleted togther with all your work. So you should keep your fork.

PR #539, opened 2020-08-25
Implement new RX jitter buffer for use with Jamulus.

hselasky:
The new jitter buffer uses a timing histogram to compute the actual
jitter relative to the audio device or system clock timer.

A bit sequence has been added to all transmitted frames to improve
the jitter histogram computation.


hselasky:
Just wanted to get one more thing on the Radar. No rush for reviewing @corrados . This is a followup to the TX timing shape patch, after #529 .

hselasky:
Screenshot showing the new analyzer console.
image

Basically what you see is the distribution of packets relative to the audio device "interrupt". In this case a jitter value of 3 is configured.

hselasky:
Some tests users have made on the server side shows an improvement in the RX path when it comes to resolving jitter. Less pops and gets rid of "the increasing delay" issue. Try it for yourself :-)

corrados:
To have a chance to understand your code, I need much much more comments in your new code.

corrados:
I guess you would like to do more modifications to the Jamulus source code. The recommended procedure would be that before you start implementing, you create an Issue here: https://github.com/corrados/jamulus/issues. In that Issue you specify what you want to do. Then we can discuss the specification before you start the coding.

I have created Issues for your pull requests to better track the changes.

corrados:
In the meantime the Jamulus protocol has changed so that a packet counter is included. So no need for using just one bit anymore.

I have seen that you have started your own jam software project: https://github.com/hselasky/hpsjam. I cannot find any documentation about it. Do you plan to add some documentation/project description?

hselasky:
Hi @corrados ,
You can probably abandon this change. Yes, I started my own JAM tool. It works similar to Jamulus, though it has redundancy in the protocol using plain old good XOR. There were so many things I wanted to change, so I figured out that a new project was probably the best, than hammering you with patches. The program is QT based and has a server and client in one like yours.
You just run qmake and make on Linux or MacOSX and it should work out of the box.
--HPS

corrados:
Sounds great. Good luck with your project. I'll then close your other pull requests as well since you will most probably use floating point signal processing in hpsjam as well.

corrados:
BTW: I have just posted the link to your project on the Facebook Jamulus group. I hope this is ok for you. If not, I can delete that post.

hselasky:
No problem. Thank you!

PR #541, opened 2020-08-26
Replace CPreciseTime class with QElapsedTimer class.

hselasky:
Avoid having so many different ways to obtain time differences.


corrados:
I compiled your branch under Windows. This is what I get:
grafik
Using the current Git master code:
grafik
Obviously, the new ping measurement does not give the correct results. Can you please check what goes wrong?

hselasky:
Should be fixed now. Can you try again. Thank you!
(Took a while to reproduce)

corrados:
Now it works, thanks.

PR #542, opened 2020-08-26
Use magic for unspecified length when storing WAV files

hselasky:
They can easily be bigger than the supported maximum of 4GBytes.

Tested using:
Audacity
VLC
SOX

Quote from SOX documentation:

```
/* Magic length writen when its not possible to write valid lengths.
 * This can be either because of non-seekable output or because
 * the length can not be represented by the 32-bits used in WAV files.
 * When magic length is detected on inputs, disable any length
 * logic.
 */
#define MS_UNSPEC 0x7ffff000
```

### DIFF
--- a/src/recorder/cwavestream.cpp
+++ b/src/recorder/cwavestream.cpp
@@ -123,26 +123,32 @@ void CWaveStream::waveStreamHeaders()
 
 void CWaveStream::finalise()
 {
-    static const uint32_t hdrRiffChunkSize = sizeof(uint32_t) + sizeof(uint32_t) + sizeof(uint32_t);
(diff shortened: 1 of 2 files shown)


corrados:
@pljones This one is for you.

pljones:
I'd prefer if the magic length were overwritten if possible and only used:
> either because of non-seekable output or because the length can not be represented by the 32-bits used in WAV files.

So write in place of the padding value when starting but overwrite when the value is known and not too big.

hselasky:
OK, I'll have a look at it.

hselasky:
@pljones : Updated code to only store values if they fit within 32-bits.

pljones (approved):

Issue #551, opened 2020-08-28
Change stop server message when idling


bflamig:
When I'm scanner the Jamulus server logs, I'm always thrown for a loop when I see the message "server stopped". This makes me think the server has shutdown, when in fact, most of the time it is idling because there are no more clients connected.

I propose the message be changed to "server idling" when there are no clients. My branch bflamig:change_stop_server_msg has proposed code to make this change.

corrados:
Why not simply exchange the string "server stopped" at https://github.com/corrados/jamulus/blob/master/src/serverlogging.cpp#L62 by "server idling"? I could do that quickly on Git master.

bflamig:
Is the call to stop the server always for when no clients are currently connected? I thought maybe it was called when the server was normally shutting down too. If it's the former, then yeah, it's an easy fix for you if you want.

corrados:
Yes, it's just when no clients are currently connected. I'll change this quickly on master and close your pull request afterwards.

corrados:
It's applied -> closed

PR #552, opened 2020-08-28
Change server log message when idling

bflamig:
When a server has no more clients connected, the server log will show the message "server stopped". I propose this be changed to "server idling" for this case. When I see "server stopped" it makes me think the server has shutdown for some reason and is no longer running. So I go to restart it only to see it's already running.


corrados:
As discussed in #551, I'll close this pull request. Thanks.

PR #554, opened 2020-08-28
Daily Update Checker (on start up)

pljones:
Whilst I was doing the update checker, I couldn't stop myself making a few other small changes:
* Mention use of "-e localhost"
* Move saving settings to CSettings
* Move defaulting of centralserver into CSettings
* Server translations might be used even headless (logging)

The bulk of the changes is in the one commit.  It relies on the change to make CSettings responsible for saving and I used `tr()` liberally, even headless, so loading the translations might be useful.  The other two were very much "in passing" - the first triggered by some comment on Facebook, the third following my other changes (should have been done there).  Anyway...
* Add an update checker

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -547,13 +547,6 @@ int main ( int argc, char** argv )
 
 
     // Dependencies ------------------------------------------------------------
-    // per definition: if we are in "GUI" server mode and no central server
-    // address is given, we use the default central server address
-    if ( !bIsClient && bUseGUI && strCentralServer.isEmpty() )
-    {
-        strCentralServer = DEFAULT_SERVER_ADDRESS;
-    }
-
     // adjust default port number for client: use different default port than the server since
     // if the client is started before the server, the server would get a socket bind error
     if ( bIsClient && !bCustomPortNumberGiven )
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -684,19 +684,20 @@ int main ( int argc, char** argv )
                              bUseMultithreading,
                              eLicenceType );
 
+            // load settings from init-file (command line options override)
+            CServerSettings Settings ( &Server, strIniFileName );
+            Settings.Load ( CommandLineOptions );
+
+            // load translation
+            if ( bUseGUI && bUseTranslation )
+            {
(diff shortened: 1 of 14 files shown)


src/main.cpp
```diff
@@ -755,11 +792,13 @@ QString UsageArguments ( char **argv )
         "  -n, --nogui           disable GUI\n"
         "  -p, --port            set your local port number\n"
         "  -t, --notranslation   disable translation (use englisch language)\n"
+        "      --updatetool      'daily' or 'never' check for a new " + APP_NAME + " version\n"
         "  -v, --version         output version information and exit\n"
         "\nServer only:\n"
         "  -a, --servername      server name, required for HTML status\n"
         "  -d, --discononquit    disconnect all clients on quit\n"
         "  -e, --centralserver   address of the central server\n"
+        "                        (or 'localhost' to be a central server)\n"
```
corrados, line 801: Good idea. I have just applied this to master, thanks.

src/settings.h
```diff
@@ -36,25 +36,34 @@
...
+    Q_OBJECT
+
 public:
     CSettings() :
         vecWindowPosMain ( ), // empty array
         strLanguage      ( "" ),
-        strFileName      ( "" ) {}
+        strFileName      ( "" )
+    {
+        QObject::connect ( QCoreApplication::instance(), &QCoreApplication::aboutToQuit,
```
corrados, line 49: That's a good idea. That improves the code. I've just added this commit to the Git master, thanks.

src/main.cpp
```diff
@@ -533,13 +533,6 @@ int main ( int argc, char** argv )
 
 
     // Dependencies ------------------------------------------------------------
-    // per definition: if we are in "GUI" server mode and no central server
```
corrados, line 536: We need this here in case of a headless server since no settings are used in that case.

pljones: The inifile is used if specified and settings are _always_ loaded.  If no GUI is given settings are no _written_ to the inifile.  So the load is always called.

pljones: Rebase moved this to commit 2267a4d .

pljones:
I'm going to close this off.

PR #556, opened 2020-08-30
debian: Add headless package

tormodvolden:
Build two binary packages:
- "jamulus" with the full-fledged GUI build
- "jamulus-headless" with the headless build suitable for servers (includes systemd unit)
Both can be installed at the same time.


corrados:
Thank you.

trebmuh:
@tormodvolden am I right to understand that previoulsy, any installation of the "jamulus" package was auto-launching a jamulus server?

tormodvolden:
No, it ships the systemd unit, but you have to enable it (and start it if you want to run immediately).

trebmuh:
That is very strange then, because here, it was run automatically when the `jamulus` package was installed. Now, with the split between `jamulus` and `jamulus-headless`, if one only installs the `jamulus` package, then the service file isn't installed and it's not automatically launched.
_I'm on Debian Buster._

tormodvolden:
First to comment on the last part: "if one only installs the jamulus package, then the service file isn't installed and it's not automatically launched" - this definitely the intention. BTW, when I release this into my PPA by 3.5.11, I will also add a versioned conflict there so nobody keeps an old jamulus with its service and installs jamulus-headless on top with another service. However for the repo packaging (distribution/debian/) I don't think this is necessary since it's just an example, and something that is working towards official packaging at some point.

The first part I am not sure about. I compiled and ran this on Buster first and don't remember if I enabled the service manually. At some point I started downloading the PPA .deb instead to "dog-food" it. But I had a unit overlay and I don't know if that makes a difference. I don't have this Buster server running any longer so I mainly test on Ubuntu 16.04 now. Surely installing -headless doesn't enable it here. I will have to check if installing the old jamulus package (from scratch with no unit files around) would make it run automatically.

I guess it would make sense that the -headless package enables it out of the box, since the primary use-case is for running a server. Although it can be used as a headless client also, if I understand correctly. There is a guy here making headless RPi's for his choir fellows!

trebmuh:
>  Surely installing -headless doesn't enable it here.

It does here. Can you please check on your installation.

PR #557, opened 2020-08-30
Load settings even without gui

pljones:
Two minor changes:

In the _client_ side, there's a check for `bUseGUI && bUseTranslation` _immediately_ before a check for `bUseGUI`.   I've just moved the code inside the later `if`.

Now, that code is _after_ the client inifile load on line 631 - the client _always_ loads the inifile, headless, nogui, whatever.

So, the second change is a little bigger in effect, although the code looks smaller (side by side diff).  Line 695 was _already_ inside a check for `bUseGUI`, so drop the duplication.  Then move the inifile load outside the `#ifndef` to match the client.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -630,16 +630,16 @@ int main ( int argc, char** argv )
             CClientSettings Settings ( &Client, strIniFileName );
             Settings.Load ( CommandLineOptions );
 
-            // load translation
-            if ( bUseGUI && bUseTranslation )
-            {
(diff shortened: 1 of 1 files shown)


corrados:
What do we do with this pull request? As discussed, I would like to keep the existing code and do not make changes in this regard.

pljones:
OK, now the help message has been made clear, I'm happy to ditch all the changes I was making.

PR #563, opened 2020-09-01
update d/rules (workaround issue in desktop file)

trebmuh:
The `exec` field is replaced with jamulus-headless which isn't what's expected by the user to be executed.

Pinging @tormodvolden since you've been working on the debian/ files recently.

@corrados : maybe it could be a good idea to generate a desktop file for jamulus-GUI ("jamulus.desktop") and another desktop file for jamulus-headless ("jamulus-headless.desktop").

### DIFF
--- a/distributions/debian/rules
+++ b/distributions/debian/rules
@@ -15,8 +15,8 @@ override_dh_auto_build:
 	cd build-nox && make
 
 override_dh_auto_install:
-	cd build-gui && make install INSTALL_ROOT=../debian/tmp
 	cd build-nox && make install INSTALL_ROOT=../debian/tmp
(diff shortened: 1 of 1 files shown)


corrados:
> @corrados : maybe it could be a good idea to generate a desktop file for jamulus-GUI ("jamulus.desktop") and another desktop file for jamulus-headless ("jamulus-headless.desktop").

Yes, good idea.

tormodvolden:
Thanks! I hadn't noticed that the desktop file was autogenerated and didn't check it. This probably explains some Lintian warnings, I thought it was only confused...

A desktop file for jamulus-headless is IMO pointless.

Issue #555, opened 2020-08-29
Interaction between command line arguments and ini file


corrados:
For me, the headless Jamulus server on the Linux command line should behave like any other GNU tool. You configure it with command line arguments and that's it. No hidden ini file changes the settings.

The situation is different if I run the client/server with a GUI on, e.g., Windows. For that operation my expectation is that the software stores/recovers my settings I set in the GUI.

pljones and I have a different opinion on that. See the discussion here https://github.com/corrados/jamulus/issues/370#issuecomment-683264655 and here https://github.com/corrados/jamulus/issues/370#issuecomment-683267114 and here https://github.com/corrados/jamulus/issues/370#issuecomment-683275779.

I think we need some more opinions on that matter. Please discuss the expected behaviour in this Issue.

WolfganP:
My take is similar to the scattered comments around:
* For GUI users, I believe there's the expectation that an ini/cfg file will store the session settings (be loaded at pgm start time, and stored there if modified)
* For CLI users, I sense them more technical oriented, and willing to play with documented features and settings priorities. ie, for me:
- CLI arguments are king and override everything else (hardwired defaults or ini file import)
- if an ini/cfg file exists, it will be imported to complement (not overwrite) the set of settings for the session. The actions (ini file imported, settings adopted) should be announced on stdout/stderr.
- in headless mode, there's no interaction with the program (except server's basic recording control) so no expectation of ini file being rewritten/updated.

A complement to certain jamulus/system announcements may be the use of the chat window to push for important notifications (ie server outdated)

pljones:
Just to be clear, the scenarios being consider have the following use cases:

* If I run _without_ an ini file, and _without_ a gui, settings _should not_ be written.
  (Pass -e Rock, no inifile updates.)

* If I run _with_ an ini file, but _without_ a gui, settings _should not_ be written.
  (Load an inifile with Central Server Jazz, pass -e Rock, the inifile stays as Jazz.)

* If I run _without_ an ini file, but _with_ a gui, settings _should_ be written.
  (Pass -e Rock, default inifile gets updated.)

* If I run _with_ an ini file, and _with_ a gui, settings _should_ be written.
  (Load an inifile with Central Server Jazz, pass -e Rock, the inifile gets update to Rock.)

The first two cases I have no issue with.

[...]

PR #570, opened 2020-09-05
#555 Interaction between command line and inifile

pljones:
These changes are based on the discussion in #555 and make what I believe was stated as the target implementation effective.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -116,6 +116,15 @@ int main ( int argc, char** argv )
             bUseGUI = false;
             tsConsole << "- no GUI mode chosen" << endl;
             CommandLineOptions << "--nogui";
+
+            if ( CommandLineOptions.contains ( "--inifile" ) )
+            {
+                tsConsole << argv[0] << ": ";
+                tsConsole << "Cannot specify inifile"
+                    " without GUI -- use '--help' for help" << endl;
+                break;
+            }
+
             continue;
         }
 
@@ -456,6 +465,7 @@ int main ( int argc, char** argv )
         }
 
 
+#ifndef HEADLESS
         // Initialization file -------------------------------------------------
         if ( GetStringArgument ( tsConsole,
                                  argc,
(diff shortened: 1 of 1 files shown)


corrados:
I think the goal you try to reach is to have consistency. But your changes break a use case: Running Jamulus headless on a Raspberry Pi zero: https://github.com/corrados/jamulus/blob/master/distributions/raspijamulus.sh#L111 As you can see, I am using the inifile for the client.

So what I wrote in #555 means, that for the server we have the situation that in headless mode the inifile is never used. But the situation is different for the client. For the client I need the possibility to use the inifile even in headless mode. That is the reason the code is implemented as it is in the current Git master. So that is on purpose, actually.

The other thing is that we should keep the support of all command line arguments which we have in normal mode also for the headless mode. There was a discussion going on when the HEADLESS compile mode was introduced. I was also wanting to remove command line arguments for the headless mode but the develooper of the HEADLESS compilation convinced me that it is better to keep the complete set of command line arguments but just show a warning (as you get if you compiled Jamulus with HEADLESS but do not use the -n command line argument, then you get: "No GUI support compiled. Running in headless mode.".

corrados:
I have just added more description to the help text: https://github.com/corrados/jamulus/commit/8b7ccde7d64f739e96fa0b6415b6840eb5eda3f3

And also a warning if the inifile is used in headless server mode: https://github.com/corrados/jamulus/commit/0a0f21d90c4f248de237a763b2c03c264e54ad06

pljones:
It sounds like there's something missing from the command line for the RPi usecase.  Having to edit the ini file by hand sounds far from ideal, compared with passing command line options.

corrados:
This script is a special case since I want to support the Pi Zero and also better performing Raspberry Pi versions as well. The recommended way of doing it is to start the client with a GUI and using the --inifile (to specify the file you want to use for the Raspberry Pi), change all the settings you want and exit the client. Then you have everything stored in that ini file which you can then use for your headless Jamulus client on the Raspberry Pi.

pljones:
>  The recommended way of doing it is to start the client with a GUI and using the --inifile

That would work for the server, too, of course, if it was allowed to load the inifile when headless.

Issue #576, opened 2020-09-06
Jamulus, privacy, IP addresses and GDPR/legal


gilgongo:
If you are interested in the issues in the title of this ticket, please read this summary and discuss here if you want.

### TL;DR executive summary

If the Jamulus project were to treat IP addresses as personally identifiable information (PII) under GDPR, we would need to have a feature that allowed server operators optionally to turn on a notice that said something like, “By using Jamulus your IP address will be made available to third parties”. If the user clicked “no” they would not be able to use the system.

The argument for this would seem to be ultra-cautious, since web browsers or mail server software doesn’t do this. But is Jamulus different?

We could also have a notice on the wiki that tells people how their IP is used by Jamulus. But it’s not clear what good that would do without also giving server operators the option of informing users and allowing them to stop using Jamulus. But we could just be nice and explain for clarity’s sake.

### The very long boring version

The following are, I believe, accurate and uncontroversial points (if not, we should discuss them on this ticket):

1. As of now, the only thing that Jamulus collects from users that in theory could be personally identifiable information (PII) is the user’s IP address. The IP address is either that of the user’s server if they are running one, or their client. Note that we assume profile “names” are not PII because if the user doesn’t want to “freely give” their real name, they don’t have to.

1. All public server IPs are available to see on a 3rd party website (for debugging and other purposes). In the case of Jamulus client IPs, it’s harder to find out which IP addresses are connected to a given server, but it’s not impossible (Jamulus does not attempt to obscure them). (EDIT: This is not not true of later versions of Jamulus.)

1. In order for Jamulus to work, it also sends user’s IP addresses to a 3rd party (currently Google but it could be anyone).

**Do we need to treat IP addresses as PII?**

1. Firstly, it’s important to note that the Jamulus project itself has no responsibility toward users to advise them on legal matters, since a) the person responsible under law is the person(s) who runs a Jamuls server and b) not all Jamulus users are bound by the GDPR and c) we are not lawyers. We can be transparent about how Jamulus works though. This compares to a project such as the Apache web server for example, which has no privacy policy about its use of IP addresses in serving web pages.

[...]

PR #577, opened 2020-09-06
Don't automatically log full ip adresses on server log

ann0see:
My proposal in #576 No. 3
To be discussed.


pljones:
Sorry, as a server operator, I *strongly* object to removing this information about people using my facilities.  If I get abuse, I want to know as much information as possible about them.

PR #578, opened 2020-09-06
Remove last digits of IP in server logging (default)

ann0see:
My proposal in https://github.com/corrados/jamulus/issues/576#issuecomment-687860878
To be discussed.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -8,16 +8,16 @@
  *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software
- * Foundation; either version 2 of the License, or (at your option) any later 
+ * Foundation; either version 2 of the License, or (at your option) any later
  * version.
  *
- * This program is distributed in the hope that it will be useful, but WITHOUT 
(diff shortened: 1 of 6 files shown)


pljones:
Sorry, I strongly object to having this information removed.  If I'm providing a public service and someone chooses to use it, they choose to expose their IP address.

pljones:
It's also a potential security flaw, as it anonymises clients, enabling DDoS attacks more easily.

I can't see that it provides any benefits at all.

atsampson:
I also don't think this is a good idea - it's not effective anonymisation, and it just makes the server operator's life harder. Better to document that the logs contain addresses and you should treat them like logs from any other server, i.e. use something like logrotate to manage retention.

ann0see:
What does @corrados think? Should it be closed or should the default behavior be changed to "show full IPs in log except if server is started with e.g. —anonymizeip“?

corrados:
I actually agree with the others. I think the IP address is an essential part of the server operation. If a user does not want to see and log any IP address, he can run the server headless with std-out piped to the null device and do not use the logging feature.

ann0see:
Ok. If you think that’s ok I‘ll close the PR

PR #582, opened 2020-09-10
Split messages

corrados:
Large Jamulus protocol messages may hit a limit where UDP packets are blocked by some fragmentation issues (see https://github.com/corrados/jamulus/issues/255). Recently, when a lot of clients are connected to the server, we do not only see this UDP blocking issue for the server list but also at normal operation of Jamulus (i.e. the procol mechanism gets stuck), see https://github.com/corrados/jamulus/issues/547.

Since the protocol stack in Jamulus is a very critical part regarding stability and security:

@softins Since you are the wireshark/protocol expert, could you please review the code to make sure it works correctly, does not break compatibility and does not crash.

@atsampson I saw that you have done a lot of stability/fuzzing testing for the Issue https://github.com/corrados/jamulus/issues/314. Since with this new code a lot of the protocol mechanism is changed, I may have introduced new security issues. Could you please review the code and/or test the new code with your fuzzing tools to make sure it still is stable and secure?

### DIFF
--- a/src/protocol.cpp
+++ b/src/protocol.cpp
@@ -419,20 +419,6 @@ CONNECTION LESS MESSAGES
 /* Implementation *************************************************************/
 CProtocol::CProtocol()
 {
-
-
-// TEST
-const int iMaxNumParts = 200;//25;
-iPartSize = 600;
-vecvecbySplitMessageStorage.Init ( iMaxNumParts );
-
-for ( int i = 0; i < iMaxNumParts; i++ )
-{
-    vecvecbySplitMessageStorage[i].Init ( iPartSize );
-}
-
-
-
     Reset();
 
 
@@ -533,17 +519,17 @@ int iPointer = 0;
 qDebug() << "CreateAndSendMessage vecData.Size(): " << vecData.Size();
 
 // TEST split all messages by half for testing
-if ( vecData.Size() > iPartSize )
+if ( vecData.Size() > MESS_SPLIT_PART_SIZE_BYTES )
 {
-    const int iNumParts = static_cast<int> ( ceil ( static_cast<double> ( vecData.Size() ) / iPartSize ) );
+    const int iNumParts = static_cast<int> ( ceil ( static_cast<double> ( vecData.Size() ) / MESS_SPLIT_PART_SIZE_BYTES ) );
 
 qDebug() << "CreateAndSendMessage message spilt in " << iNumParts << " parts";
 
     for ( int iSplitCnt = 0; iSplitCnt < iNumParts; iSplitCnt++ )
     {
-        int iCurPartSize = iPartSize;
+        int iCurPartSize = MESS_SPLIT_PART_SIZE_BYTES;
 
-        if ( vecData.Size() - iPointer < iPartSize )
+        if ( vecData.Size() - iPointer < MESS_SPLIT_PART_SIZE_BYTES )
         {
             iCurPartSize = vecData.Size() - iPointer;
         }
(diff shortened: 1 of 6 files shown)


src/protocol.cpp
```diff
@@ -495,22 +524,74 @@ void CProtocol::CreateAndSendMessage ( const int               iID,
 {
     CVector<uint8_t> vecNewMessage;
     int              iCurCounter;
+    const int        iDataLen = vecData.Size();
 
-    Mutex.lock();
+    // check if message has to be split because it is too large
+    if ( ( iDataLen > MESS_SPLIT_PART_SIZE_BYTES ) && bSplitMessageSupported )
```
pljones, line 530: I'd swap the two sides of the `&&` around as bool truth is an easier computation than comparator truth.  Compiler might do it here as there's no possible side effects.

src/protocol.cpp
```diff
@@ -51,6 +51,19 @@ MAIN FRAME
...
 
 
+SPLIT MESSAGE CONTAINER
+-----------------------
+
+    +------------+------------------------+------------------+--------------+
+    | 2 bytes ID | 1 byte number of parts | 1 byte split cnt | n bytes data |
+    +------------+------------------------+------------------+--------------+
+
+- ID is the message ID of the message being split
```
pljones, line 61: `number of parts` - total number of parts comprising the whole message

pljones (commented):
I can't see anything seriously wrong structurally.  I can't comment on performance aspects just from looking at the code, though.

softins:
Sure, happy to have a look, but it will be on Friday.

src/protocol.cpp
corrados: Good idea, I'll change this.

src/protocol.cpp
corrados: done

src/protocol.cpp
```diff
@@ -51,6 +51,19 @@ MAIN FRAME
...
+- split cnt is the counter which is incremented for each part of the message
```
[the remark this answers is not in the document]
corrados: done

src/protocol.cpp
```diff
@@ -51,6 +51,19 @@ MAIN FRAME
...
+- data is the data part of the original message which were split
```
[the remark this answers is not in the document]
corrados: done

corrados:
> I can't see anything seriously wrong structurally. I can't comment on performance aspects just from looking at the code, though.

Thank you for your review.

softins:
It looks ok to me so far. I have compiled both server and client on local machines from the branch `split_messages`, with the split threshold reduced to 100 bytes so I can observe the split protocol without needing too many clients.

I have updated my Wireshark dissector at https://github.com/softins/jamulus-wireshark to display the new messages and the split container fields. I have yet to implement re-assembly of the split packet data parts into a virtual packet for display.

softins:
Maybe you also need to implement a `CLM_REQ_SERVER_LIST_SPLIT`, which a newer client could use instead of `CLM_REQ_SERVER_LIST`, to allow the server to split a `CLM_CONN_CLIENTS_LIST` in the same way?

corrados:
> split threshold reduced to 100 bytes so I can observe the split protocol without needing too many clients.

You can also simply use a long Chat text message. This will force the message split as well.

> CLM_REQ_SERVER_LIST_SPLIT

Yes, something similar will be the next step. But first things first. Let's for now fix the problems with multiple clients.

corrados:
I have run this code on the Central server for days now and it seems to be stable. So I will merge this pull request to master now. You can still put your reviews here and I will consider them on the master branch.

atsampson:
> Could you please review the code and/or test the new code with your fuzzing tools to make sure it still is stable and secure?

I've had a look through these changes, and they look OK to me so far. I'll try and add support for these to my fuzzing branch when I get a chance.

Issue #591, opened 2020-09-15
Hyperlink access to chords/lyrics/mp3... from the chat window


jc-Rosichini:
Hi Volker,

Over the past months I've noticed that it takes often a fair amount of time for musicians/singers during a jam to share and get access to chords/lyrics.

Usually one person ends up dropping a link in the chat for a guitar-tabs or similar page.
The point is that even if the link is now visible in the chat window, it takes some additional time to cut it & paste it into Firefox/Chrome or whatever.

Considering this very practical recurring User Interface request, I made some tiny changes in the code to be able to merely highlight the depicted hyperlink and then, if clicking on it, open up a browser page. 

This feature has been « unofficially » tested by a group of Jamulus friends over the last few days -> all of them  are considering that it's a MUST have:)
and are insisting to get it included in the regular release.

**modified source code included**
[JessY_code_mods.zip](https://github.com/corrados/jamulus/files/5223965/JessY_code_mods.zip)

Kind regards
Jess£   (jean-Claude)

corrados:
Thanks for your code. I created a pull request for your code: https://github.com/corrados/jamulus/pull/592

We have a similar Issue as yours: https://github.com/corrados/jamulus/issues/360. There was already a discussion about the security here: https://github.com/corrados/jamulus/pull/380#issuecomment-647042975

I have already played a bit with the `openLinks ` and `openExternalLinks ` flags a while ago. The problem is that you could easily confuse the chat window by sending an invalid link. You get in a weird state. So I think we need some more discussion and testing before we can add your patch to the official Jamulus code.

jc-Rosichini:
Thanks Volker,

Actually assuming the link is invalid, this should not create confusion within the chat window which has nothing to do with the link processing. The link being sent to the external browser, this is it that will detect that the link is not a valid one. I've tested this on Windows and this seems ok.

Wrt security issue, this shouldn't either interact with Jamulus itself. It's up to the user to decide if he wants or not click on the link and pass it to its browser (not the chat window). 
And btw this is what people are doing on almost every Jamulus session but with a more primitive and tedious way: doing cut & paste (and getting access to chords when the song is almost completed by others musicians :-) ) 

Last but not the least, if wished we can limit the url to https (see the code).

Kinds regards
Jess£

pljones:
Jamulus would be enabling this.  Therefore it's Jamulus's problem.

NickVeit:
[...]

PR #592, opened 2020-09-15
Hyperlink access to chords/lyrics/mp3... from the chat window (#591)

corrados:
Pull request with code provided in Hyperlink access to chords/lyrics/mp3... from the chat window #591

### DIFF
--- a/src/chatdlg.cpp
+++ b/src/chatdlg.cpp
@@ -106,8 +106,26 @@ void CChatDlg::OnClearChatHistory()
 void CChatDlg::AddChatText ( QString strChatText )
 {
     // add new text in chat window
-    txvChatWindow->append ( strChatText );
+	
+    // analyze strChatText to check if hyperlink
+	if (strChatText.contains("https://", Qt::CaseInsensitive) | strChatText.contains("http://", Qt::CaseInsensitive)) 
+	{
+		int indx_http  = strChatText.indexOf("http", 0);
+		int indx_space = strChatText.indexOf(" ", indx_http); 
+		if (indx_space==-1) indx_space=strChatText.length();
 
-    // notify accessibility plugin that text has changed
-    QAccessible::updateAccessibility ( new QAccessibleValueChangeEvent ( txvChatWindow, strChatText ) );
+		int cutslash = (strChatText.at(indx_space-1)=='/') ? 1:0;		// drop "/" if last character of url text
+
+		QString new_strChatText = strChatText.left(indx_http)+"<a href=\""+strChatText.mid(indx_http,indx_space-indx_http-cutslash)+"\">"+strChatText.mid(indx_http,indx_space-indx_http-cutslash)+"</a> "+strChatText.right(strChatText.length()-indx_space);
+
+		txvChatWindow->append (new_strChatText);	
+		// notify accessibility plugin that text has changed
+		QAccessible::updateAccessibility ( new QAccessibleValueChangeEvent ( txvChatWindow, new_strChatText) );
+	}	
+	else 
+	{
+		txvChatWindow->append ( strChatText );
+		// notify accessibility plugin that text has changed
+		QAccessible::updateAccessibility ( new QAccessibleValueChangeEvent ( txvChatWindow, strChatText ) );
+	}
 }
--- a/src/chatdlg.cpp
+++ b/src/chatdlg.cpp
@@ -107,18 +107,26 @@ void CChatDlg::AddChatText ( QString strChatText )
 {
     // add new text in chat window
 	
-    // analyze strChatText to check if hyperlink
-	if (strChatText.contains("https://", Qt::CaseInsensitive) | strChatText.contains("http://", Qt::CaseInsensitive)) 
+    // analyze strChatText to check if hyperlink (limit ourselves to https:://)
+	int indx_http  = strChatText.indexOf("https://", 0);
+	if (indx_http != -1)
 	{
-		int indx_http  = strChatText.indexOf("http", 0);
 		int indx_space = strChatText.indexOf(" ", indx_http); 
 		if (indx_space==-1) indx_space=strChatText.length();
 
 		int cutslash = (strChatText.at(indx_space-1)=='/') ? 1:0;		// drop "/" if last character of url text
+		
+		QString URL_name = strChatText.mid(indx_http,indx_space-indx_http-cutslash); // as entered by the user
+		QUrl URL = QUrl::fromUserInput(URL_name);
+		
+		QString new_strChatText;		
+		if ( URL.isValid() )
+		{
+			new_strChatText.append( strChatText.left( indx_http )+"<a href=\""+URL.toString()+"\">"+URL_name+"</a> "+strChatText.right( strChatText.length()-indx_space ) );
+		}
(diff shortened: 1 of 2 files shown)


src/chatdlg.cpp
```diff
@@ -106,8 +106,26 @@ void CChatDlg::OnClearChatHistory()
...
-    QAccessible::updateAccessibility ( new QAccessibleValueChangeEvent ( txvChatWindow, strChatText ) );
+	
+    // analyze strChatText to check if hyperlink
+	if (strChatText.contains("https://", Qt::CaseInsensitive) | strChatText.contains("http://", Qt::CaseInsensitive)) 
+	{
+		int indx_http  = strChatText.indexOf("http", 0);
+		int indx_space = strChatText.indexOf(" ", indx_http); 
+		if (indx_space==-1) indx_space=strChatText.length();
+
+		int cutslash = (strChatText.at(indx_space-1)=='/') ? 1:0;		// drop "/" if last character of url text
```
pljones, line 117: Why?  That trailing slash _can_ cause different site behaviour (in badly behaved sites).

src/chatdlg.cpp
```diff
@@ -106,8 +106,26 @@ void CChatDlg::OnClearChatHistory()
...
     // add new text in chat window
-    txvChatWindow->append ( strChatText );
-
-    // notify accessibility plugin that text has changed
-    QAccessible::updateAccessibility ( new QAccessibleValueChangeEvent ( txvChatWindow, strChatText ) );
+	
+    // analyze strChatText to check if hyperlink
+	if (strChatText.contains("https://", Qt::CaseInsensitive) | strChatText.contains("http://", Qt::CaseInsensitive)) 
+	{
+		int indx_http  = strChatText.indexOf("http", 0);
```
pljones, line 113: So if I say "the http protocol is indicated by a 'http://' or 'https://' prefix on a url", this will find the first "http" (between "the" and "protocol").

src/chatdlgbase.ui
```diff
@@ -33,11 +33,14 @@
...
-      <set>Qt::LinksAccessibleByKeyboard|Qt::LinksAccessibleByMouse|Qt::NoTextInteraction|Qt::TextBrowserInteraction|Qt::TextSelectableByKeyboard|Qt::TextSelectableByMouse</set>
+      <set>Qt::LinksAccessibleByKeyboard|Qt::LinksAccessibleByMouse|Qt::TextBrowserInteraction|Qt::TextSelectableByKeyboard|Qt::TextSelectableByMouse</set>
      </property>
      <property name="openLinks">
-      <bool>false</bool>
+      <bool>true</bool>
      </property>
+     <property name="openExternalLinks">
+      <bool>true</bool>
+     </property>
```
pljones, line 43: It would be useful, if we retain the licence popup as a separate dialog, for these properties to be set there, too.

src/chatdlg.cpp
jc-Rosichini: Hi, the trailing slash (if any) would prevent for instance Firefox to open the site
I noticed that when entering as an hyperlink https://www.google.com/ versus https://www.google.com, this is the reason of cutting this trailing slash. 
This is not a real issue anyway since any invalid url will be rejected by the browser :-)

src/chatdlg.cpp
jc-Rosichini: That's correct. 
A better way would be to find the index of strChatText.indexOf("http://", 0) or strChatText.indexOf("https://", 0) and include some more checks.
But in any case if the link appears to be invalid the browser will deal with it and depict it as non sense.

pljones: Firefox opens (from your post above) both equally well.  If it's a defect in the QTextBrowser widget's URL handling, that's a different matter.  (And another reason to take care with using it for active links.)

src/chatdlg.cpp
pljones: Firefox opens (from your post above) both equally well.  If it's a defect in the QTextBrowser widget's URL handling, that's a different matter.  (And another reason to take care with using it for active links.)

src/chatdlg.cpp
pljones: I don't think relying on the browser to clean up is a safe option.

jc-Rosichini: Hi Peter
IMHO The browser will merely not open the wrong url and issue a 404 page.

pljones: Take a look at Qt's [QUrl](https://doc.qt.io/qt-5/qurl.html).  A simple check like `QUrl(text).isValid()` would go a long way.

src/chatdlg.cpp
```diff
@@ -106,8 +106,26 @@ void CChatDlg::OnClearChatHistory()
...
+		int indx_space = strChatText.indexOf(" ", indx_http); 
+		if (indx_space==-1) indx_space=strChatText.length();
+
+		int cutslash = (strChatText.at(indx_space-1)=='/') ? 1:0;		// drop "/" if last character of url text
+
+		QString new_strChatText = strChatText.left(indx_http)+"<a href=\""+strChatText.mid(indx_http,indx_space-indx_http-cutslash)+"\">"+strChatText.mid(indx_http,indx_space-indx_http-cutslash)+"</a> "+strChatText.right(strChatText.length()-indx_space);
+
```
pljones, line 120: `strChatText.mid(indx_http,indx_space-indx_http-cutslash` is repeated.  It would be easier to read if it were put into a variable (`strUrl`, perhaps).

I'd also suggest, to improve readability and reduce the likelihood of dubious URLs being misread, to use a fixed width font at 110% size (or 1.1rem).

jc-Rosichini: Here is a short code modif prior I leave for a few days trip
I'll try to get connected while I'm out.
Cheers

code at:  https://we.tl/t-ZGz7CK0XqG

src/chatdlg.cpp
pljones: Really, just simplify it.  If I enter something as simple as `http://example.com/?key=value`, your slash stripper fails.  I'd dump all that code and just find the index of either "http://" or "https://" through to space and pass that string.

src/chatdlg.cpp
```diff
@@ -106,8 +106,34 @@ void CChatDlg::OnClearChatHistory()
...
+	int indx_http  = strChatText.indexOf("https://", 0);
+	if (indx_http != -1)
+	{
+		int indx_space = strChatText.indexOf(" ", indx_http); 
+		if (indx_space==-1) indx_space=strChatText.length();
+
+		int cutslash = (strChatText.at(indx_space-1)=='/') ? 1:0;		// drop "/" if last character of url text
+		
+		QString URL_name = strChatText.mid(indx_http,indx_space-indx_http-cutslash); // as entered by the user
+		QUrl URL = QUrl::fromUserInput(URL_name);
```
pljones, line 120: Nice.  With the check for http/https, this should mean at least half-way sane values go in.  If it comes back invalid, the check below then catches it.  I'd still like the text more readable than "normal" so it catches the reader attention and is clearer.

corrados:
I'll merge this now. Any open issue shall be resolved on the master branch.

PR #594, opened 2020-09-17
Manual.md in Spanish, updated gui translation

ignotus666:
I've put a Spanish version of manual.md along with its pics inside a folder (ES) in `src/res/homepage`. I've also added a language selector at the top of `manual.md` so that although Spanish-speaking users are led to the English version when they click on it via the application, they can easily switch to the Spanish version. I thought it would be less intrusive than trying to make the application lead directly to the translated manual.


corrados:
Thank you!

Issue #574, opened 2020-09-06
Recording on server using command line


sthenos:
This may already be possible so feel free to close, but is there a way to specify a server to have a recording directory but to not start up initially with recording enabled?   I.e. i'd like my servers to setup without recording turned up, but then use SIGUSR2 to enable recording once it has started?

gilgongo:
I think this came up on the forums the other day: 

https://sourceforge.net/p/llcon/discussion/server/thread/698e6ee735/#b6f5

Assuming it's the same issue, I suggested an (untested) workaround to simply start the server and immediately issue a SIGUSR2.

Maybe.

pljones:
I don't run Jamulus directly from the service file - I use a script.  So, for me, it it would be relatively easy to add `& (sleep 1; systemctl start toggleRecording)`.  I'm not sure if that'd work cleanly directly from the service file.  (And it assumes you have the toggleRecording service... it's just a `systemctl kill -s SIGUSR2 Jamulus`.)

sthenos:
Yes that's what I'm currently doing, I issue `kill -12` to the process, it's just not very clean.   It would be great if there was a `--enable-recording` flag as well as the `--recording` flag that would enable recording if `--enable-recording` is also passed, otherwise it leaves recording off. and then I can start recording at a later date using `kill -12`

pljones:
It'd likely be a `--disable-recording` or an `--enable-recording (yes|no)` option, defaulting to yes.  Coz it's mine, all mine, I tell you... ;)

sthenos:
lol, yes that logic makes more sense.     and yes I guess my use case is probably less frequent for most! ;)

pljones:
OK, just to write here what I wrote in #536...

Passing `--recording <dir>` will have the server log that recording is enabled... unless...
`--norecord` is passed as well, in which case the it will log that recording is disabled.

Passing neither should not log anything about recording.

I'm probably going to have to investigate why Jamulus Server with GUI isn't saving the recording directory (for me, at least).

pljones:
#596 raised to address this.

corrados:
The code is merged -> closed.

pljones:
I'm sure this would have been picked up but I really need to get into the habit of updating the change log.

#600 is now there to do that.

pljones:
[Wiki](https://github.com/corrados/jamulus/wiki/Command-Line-Options) updated.

PR #597, opened 2020-09-18
#574 Disable recording on start up

pljones:


corrados:
Thanks. Someone should update the Wiki page, too.

Issue #606, opened 2020-09-20
Make the "Enable recording" sticky in the GUI inifile


pljones:
From https://sourceforge.net/p/llcon/discussion/server/thread/698e6ee735/#495c

I thought I'd already got the GUI/inifile remembering the state of "Enable recording" but in tests just now, it seems not.

As with other settings, passing `--norecord` on the command line would override the inifile.

pljones:
#608 pull request raised.

corrados:
Can this one be closed now?

pljones:
Ah yes!

PR #608, opened 2020-09-20
#606 Persist state of recorder

pljones:
"norecord" value of "1" stored in inifile when Enable recorder is not checked in GUI, else stored as "0".

If --norecord is passed, stored value is not read (hence recorder remains disabled).

If --norecord is not passed, stored value is read and passed to recorder.

Recorder amended only to log (apparent) _change_ in state.

(Because CServer is running before its configuration has been fully determined, this means that the recorder will still start up enabled if --norecord, issuing its state message; if the inifile then sets it to norecord, it will issue a second message for the change in state.)

### DIFF
--- a/src/recorder/jamcontroller.cpp
+++ b/src/recorder/jamcontroller.cpp
@@ -48,15 +48,20 @@ void CJamController::SetEnableRecording  ( bool bNewEnableRecording, bool isRunn
 
     if ( bRecorderInitialised )
     {
+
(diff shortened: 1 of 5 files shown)


corrados:
Do you think we need a ChangeLog entry for this small change?

pljones:
The general wording in the last one should have covered it.  And I thought I'd done it.  So... it's more of a fix, really.

PR #614, opened 2020-09-21
Do not allow an empty chat message

dingodoppelt:
This is just a tiny usability fix

### DIFF
--- a/src/chatdlg.cpp
+++ b/src/chatdlg.cpp
@@ -93,7 +93,7 @@ void CChatDlg::OnLocalInputTextTextChanged ( const QString& strNewText )
 void CChatDlg::OnSendText()
 {
     // send new text and clear line afterwards, do not send an empty message
-    if ( edtLocalInputText->text() != "" )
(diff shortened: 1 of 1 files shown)


src/chatdlg.cpp
```diff
@@ -92,9 +92,12 @@ void CChatDlg::OnLocalInputTextTextChanged ( const QString& strNewText )
 
 void CChatDlg::OnSendText()
 {
-    // send new text and clear line afterwards
-    emit NewLocalInputText ( edtLocalInputText->text() );
-    edtLocalInputText->clear();
+    // send new text and clear line afterwards, do not send an empty message
+    if ( edtLocalInputText->text() != "" )
```
corrados, line 96: I think there is an .empty() function for QString. I would use that instead.

dingodoppelt: Thanks for pointing that out. I changed the code. I hope that is how you intended it

corrados:
Looks good, thank you.

PR #624, opened 2020-09-25
Linux Install Scripts added to Repository

niebert:
The Linux Install Scripts are added to the repository unter `linux/installscripts` for 3 distros.


corrados:
Thanks. But I think the correct place for your scripts is the "distributions" directory and not the "linux" directory. Do you agree?

linux/installscripts/install4debian.sh
```diff
@@ -0,0 +1,78 @@
+
+#!/bin/sh
+# set DISTRO either to "Ubuntu", "Debian" or "Fedora"
+DISTRO="Debina"
```
trebmuh, line 4: Debian/Debina

niebert: @trebmuh  Thank you for fixing the typo. @corrados yes please move the scripts to the most appropriate place in the repo. Users will find the script mainly with a reference in the `README.md` or a reference in Wiki, because the majority will install and not contribute code.

Issue #568, opened 2020-09-04
Connection problems after configuring input channels while connected to server


bflamig:
This issue is related to #428 and #547.

**Scenario summary:** Messing around with client audio input connections while connected to a  server seems to cause the server to go haywire, causing the client to sometimes never make it past the "Trying to connect" stage when reconnecting, and to sometimes having the message (from somewhere) showing up in the server logs saying "UPD channel severely degraded." Often this is accompanied by having no public servers show up in the list, regardless of genre. The "trying to connect" problem often survives a client reboot. Only rebooting the server clears this problem up.

**Scenario details:**

1. I have Windows client computers A and B, and one Linux client C, at home on the same local network. 

2. In preparation for a band practice, I started up client A and connected successfully to remote Jamulus server (AWS). No other clients were yet connected.
 
3. After having connected, I was having trouble getting the input channels configured right in the audio device (a Scarlett 4i4). So I was messing around trying to get the right channels specified and direct monitoring turned off 
4. During this messing around, I noticed that somewhere along the way, the client was no longer connected. Either the client disconnected itself (I didn't do it), or the server did.

5. After this, I was no longer able to connect to the AWS server with client A. It would hang on "Trying to connect." However, I was able to connect to other servers (like a few of the public servers) just fine.

6. Tested whether client B could connect to the AWS server. Had no problem doing so.

7. Went back to client A, and after trying multiple times, I decided to check the remote Jamulus server logs. I noticed that client A was indeed showing as being connected via a server log message, but the client was still saying "Trying to connect" on the client. After pushing the disconnect button on the client, the remote server log showed that the client was now disconnected.

8. Step 7 was repeated many times, including shutting down the client A program and restarting, and even rebooting the machine. No matter, client A could not get past the "Trying to connect" message. I noticed that the public server lists were now empty all the time.

9. Gave up and moved to client C, a Linux computer (Rpi4). Connected no problem to the server. But I noticed the input connections weren't set up right and messed around with the Jack settings in qjackctl. Somewhere along the way, Jack gave up and quit and so I shutdown the Jamulus client, got the Jack settings right, and then fired up the client to connect to the server. No dice. Not able to get past the "Trying to connect" message. Server log showed it connecting, and disconnecting after I pushed the client disconnect button.

10. Gave up on client's A and C and went client B (a Windows machine). Decided I would just use it for both my fiddle and my wife's fiddle on the two input channels. It connected just fine to the AWS server. 

[...]

PR #625, opened 2020-09-25
Slightly randomize starting port

bflamig:
Issue #568 documents problems connecting to a server. The symptoms were narrowed down to a scenario where one client is connected to the server, using port (default + 10 = 22134), and then disconnected. A second client on another machine is started and attempts to connect to the server, and is given the same port number (22134) as a matter of course, since it's the first port number to try. Due to what is probably a faulty router (or some other mechanism?), that port is in fact still associated with the first machine, even though Jamulus is not running there, having been already disconnected from the server. The result is the UDP packets for the second machine are wrongly sent to the first machine. This results in the behavior where the server thinks the new client is connected, but the new client does not, because it is not receiving packets from the server. This behavior persists over an indeterminate amount of time, leading to much frustration on the part of the user trying to connect.

One workaround was suggested in #568 where if a client is not able to reconnect, the initial port used in finding an available can be manually changed from the standard 22134 starting port number to some other number, by specifying the starting port number on the command line using the -p option.  While that seems to work, this pull request suggests a more automatic way to work around the issue: Instead of always starting with default + 10 (22134) port number, start instead with default + 10 + some small random number. Chances are then good that any problem due to a "sticky port" on a faulty router (or wherever it happens) is bypassed.

NOTE: Since the assignment of starting port number is done in the initial phases of the client startup, if the client fails to connect, the user should shut down Jamulus and try again, upon which a different starting port number will be used.

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -67,12 +67,19 @@ void CSocket::Init ( const quint16 iPortNumber )
         {
             // if the port is not available, try "NUM_SOCKET_PORTS_TO_TRY" times
             // with incremented port numbers
+
+            // Randomize the start by 5 ports, in case a faulty router gets
+            // stuck and confused by a particular port (like the starting port).
+            // Might work around frustrating "cannot connect" problems.
+
+            quint16 startingPortNumber = iPortNumber + rand() % 5;
+
             quint16 iClientPortIncrement = 0;
             bSuccess                     = false; // initialization for while loop
 
             while ( !bSuccess && ( iClientPortIncrement <= NUM_SOCKET_PORTS_TO_TRY ) )
             {
-                UdpSocketInAddr.sin_port = htons ( iPortNumber + iClientPortIncrement );
+                UdpSocketInAddr.sin_port = htons ( startingPortNumber + iClientPortIncrement );
 
                 bSuccess = ( ::bind ( UdpSocket ,
                                     (sockaddr*) &UdpSocketInAddr,


corrados:
There is actually already some randomization supported by Jamulus. Can you please try to start Jamulus with:

Jamulus --port 0

And see if that solves your problem? I am curious.

bflamig:
I had seen the if statement in the code that checked for the port number being 0 and then randomizing, but hadn't realized the conditions it occurred under. You are saying it's from when the --port 0 option is used. If so, I don't see why that wouldn't solve the problem. I'll try it. But since the problem is intermittent (the vast majority of the time I don't have any connection problems) it may be a while for it shows up and I can test it.

I am wondering, for information purposes, what's the reason for not randomizing all the time? I can think of a few reasons myself: more consistency in the ports being chosen makes for easier debugging down the line, etc. Easier to administrate, etc.

corrados:
The problem with randomizing is that you cannot do these multithreading tests anymore which we performed a lot recently. If you start a bunch of clients at the same PC, by randomizing, you will most probably get the same port for two clients and the second client will not start up. If you just increment the ports, that cannot happen. That was the reason why I incremented it. It would be good if your problem was solved by using the --port 0.

bflamig:
"by randomizing, you will most probably get the same port for two clients"

Not understanding this. How are they likely to get the same port? 

On the other hand, by **not** randomizing, they are both going to start at 22134, right? So both clients will likely get port 22134 (if the socket bind() function messes up and says that port is available to both). I must be missing something.

How does the bind() function actually work? How does it know whether a port is "available" or not? Is the algorithm different between Windows and Linux? Just curious.

My proposal is in between the two approaches of full randomization and randomizing (within a few ports) where the port scanning range starts.

corrados:
> I must be missing something.

No, I was missing something actually... I did not understand my own code anymore ;-).

But I have still one concern. You create a random offset of 0 to 5. So in 20 % probability your two clients will use the same port and you still have issues with your router. How to solve this?

I think by using --port 0 you will get a random port in a much wider range which makes it almost unprobable to get the same port. Do you agree?

bflamig:
I agree there is a 20% probability that the two clients will do the port scanning from the same starting port number. But I was figuring that was better than a 100% probability. Chances are that if a client is seeing the connect failure due to a "sticky port" -- that is, a port that is in some kind of limbo as described in the opening comments, that they can just restart the client and chances are good it will work, since there is an 80% chance a different starting port will be used. 

By doing it this way, the client doesn't have to set the --port 0 option on the command line. They don't have to know anything about it. Just tell them in the instructions to restart Jamulus if they are having issues. That's what I was intending.

I had used such a small random range because I was being conservative. I didn't want to change the code behavior too much from what was already there. I was thinking "let's do something in between complete random and not random at all." The range could be bigger. I figured you would be better in knowing what that should be.

But I ask the question again: Why not select the port completely randomly all the time anyway? What was the reason for always starting at default_port + 10 , (and then incrementing the number if selected port isn't "available.")

UPDATE: I just tested using --port 0. On my Windows 10 machine it selected port 56038 one time, and 50282 another. I checked the documentation for the bind socket function in Windows, and it said the range of ports selected from is 49152 - 65535, if running Windows Vista or later. So that's far away from the 22134 range. I say this just in case you didn't know.

I do have a question: If I have two machines on the local network running Jamulus clients, how does one machine know what ports the other machine is using? I assumed the router, but in the issue thread #568 someone said that the machine you are running on decides what to use -- not the router. So I am a bit befuddled the moment :)

bflamig:
I changed my last comment a few times trying to get the language right.

bflamig:
UPDATE: Near the latest comments on the thread at #568, @atsampson explains very well how the port assignment works.

corrados:
> What was the reason for always starting at default_port + 10 , (and then incrementing the number if selected port isn't "available.")

No specific reason. I'll merge your code now and will slightly modify it. Thanks.

bflamig:
Ok thanks

Issue #618, opened 2020-09-22
segmentation fault (server 3.5.11)


mawess:
Got a segmentation fault in the server tonight (my first! all previous version the since April has worked (almost) perfectly)

server version 3.5.11
replaced some possibly sensitive data with [...]

6 clients connected. denoted as a b, c , x and y.

Last login: Sun Sep 20 10:47:29 on ttys000
[...] ~ % ./JamulusServer.txt
Password:
- maximum number of channels: 50
- using 64 samples frame size mode
- selected port number: 22125
- recording directory name: ./jamulusrec
Recording state enabled
2020-09-22 16:55:24, yyy.yyy.yyy.yyy, connected (1)
2020-09-22 16:55:35,, server idling -------------------------------------
Session RPP: "/Users/ [...].rpp"
Session LOF: "/Users/ [...].lof"
2020-09-22 17:26:37, aaa.aaa.aaa.aaa, connected (1)
2020-09-22 17:29:04,, server idling -------------------------------------
Session RPP: "/Users/[...].rpp"
Session LOF: "/Users/ [...].lof"
2020-09-22 17:58:25, aaa.aaa.aaa.aaa, connected (1)
2020-09-22 18:20:15, bbb.bbb.bbb.bbb, connected (2)
2020-09-22 18:30:41, ccc.ccc.ccc.ccc, connected (3)
2020-09-22 18:38:42, yyy.yyy.yyy.yyy connected (4)
2020-09-22 18:51:07, xxx.xxxx.xxx.xxx, connected (5)
2020-09-22 19:09:14, yyy.yyy.yyy.yyy, connected (6)
2020-09-22 19:09:47, yyy.yyy.yyy.yyy, connected (5)
2020-09-22 19:09:47, yyy.yyy.yyy.yyy, connected (5)
2020-09-22 19:09:47, yyy.yyy.yyy.yyy, connected (5)
./JamulusServer.txt: line 1: 46920 Segmentation fault: 11  sudo nice -n -20 /Applications/JamulusServer.app/Contents/MacOS/JamulusServer -u 50 -F -p 22125 -R ./jamulusrec
[...] ~ %

corrados:
This seems to be the same as: https://github.com/corrados/jamulus/issues/480.

I was hoping we had fixed that. Anyway, I think this bug is very hard to reproduce. But here are similarities I can find in the settings/logs:

- in both cases the recording was enabled
- in both cases a user was connecting/disconnecting quickly in a short period of time

[...]

PR #626, opened 2020-09-26
#618 Make recorder thread-safe

pljones:
I had a look through all the signals handled by the jam recorder and, as far as I can tell, I've covered the places where there could be contention.

I've tested that a single client can connect successfully and recording still works...  I had a go at recreating the race condition by stopping the server, rather than having the client disconnect.  Couldn't get it to crash (but I've done that before locally and it's not crashed, so I doubt it's a "good test").

So, any ideas on reproducing the crash conditions would be appreciated! :)

### DIFF
--- a/src/recorder/jamrecorder.cpp
+++ b/src/recorder/jamrecorder.cpp
@@ -348,8 +348,13 @@ void CJamRecorder::Start() {
     // Ensure any previous cleaning up has been done.
     OnEnd();
 
-    currentSession = new CJamSession( recordBaseDir );
-    isRecording = true;
+    // needs to be after OnEnd() as that also locks
+    ChIdMutex.lock();
+    {
+        currentSession = new CJamSession( recordBaseDir );
+        isRecording = true;
+    }
+    ChIdMutex.unlock();
 
     emit RecordingSessionStarted ( currentSession->SessionDir().path() );
 }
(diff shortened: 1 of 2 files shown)


corrados:
Since it does not hurt to lock the entire functions, you should use the mutex locker class as I do on many places in the code, e.g.:

QMutexLocker locker ( &Mutex );

You just place this one line at the beginning of the function and the locking/unlocking is done automatically by the QMutexLocker.

Additionally, I think the CJamRecorder::Start() function needs also to be mutexed since the object currentSession is created in that function. Do you agree?

pljones:
Is QMutexLocker recursive-safe?  There's at least one place a method that needs the lock calls another method that needs the lock, where that second method is also called from elsewhere (i.e. `OnEnd()`).

corrados:
No, I think it isn't. In that case I would only secure the:

    currentSession = new CJamSession( recordBaseDir );
    isRecording = true;

with a mutex and not the entire function. And in OnFrame you should keep your implementation (i.e. do not include the     // Make sure we are ready
    if ( !isRecording )
    {
        Start();
    }
in the mutex as you already did).

pljones:
OK, updated it a bit.  `Start()` locks after `OnEnd()` and `OnDisconnected()` is now fully protected by the mutex.

corrados:
Looks good. Hopefully we'll not get a dead lock. I'll merge your code now and it would be good if you could update your server (which has recording enabled) with the newest code to see if it runs stable.

pljones:
OK, up and running now.

PR #627, opened 2020-09-26
Integration branch of: Switch all Jamulus audio sample processing to use floats instead of a mix of double and int16_t

corrados:
This is basically the pull request for the floating point support from hselasky. I did:

- code style changes
- replaced old casts by static_cast
- variable renaming
- fixed the clipping indicator
- fixed compilation problems

TODO is still the testing on all platforms.

This is work in progress and the Pull Request will most probably updated with new commits in the near future.

Feel free to review the code. If you have time to test the new code, please tell us your test results (audio quality, stability, are all existing features work as they did before the change, etc.).


corrados:
I think the recording should be tested since there were also changes needed in that module.

pljones:
> I think the recording should be tested since there were also changes needed in that module.

I want to get the Mutex changes tested first.  Didn't have much happening tonight, so I'd like to wait further trials.

corrados:
Wow, there are so many things which do not work as expected anymore with the float code... I think it will take quite some time until we can finally merge that code to the master. I'm just in progress of fixing the level meter.

windows/sound.cpp
```diff
@@ -663,8 +665,9 @@ struct sample24MSB
     }
 };
 
+// Might want to comment that use of double in factors below avoids nasty conversion surprises.
```
pljones, line 668: Could also mention that FACTOR32 and FACTOR24 are the same because...

corrados:
@pljones I have just merged the code to the master. hselasky  has changed your jam recorder to record in 24 bit instead of the previous 16 bit resolution. Can you please check if the new code still works ok?

Since the jam recorder already creates very large files, with 24 bit the amount will even be higher now. If it is too much data, we could consider going back to 16 bit. But this is your decision (since I do not use the jam recorder).

corrados:
@pljones The Windows audio interface is not working correctly so I had to hard reset the master to a previous commit and therefore undo the merge to master. So if you want to test, you could do it on the new branch: https://github.com/corrados/jamulus/pull/660

PR #634, opened 2020-09-28
fix png file which caused libpng warning: iCCP: known incorrect sRGB …

kdoren:
resource file src/res/instruments/aguitar.png had a bad sRGB profile, causing warning:
"libpng warning: iCCP: known incorrect sRGB profile"

removed with pngcrush to eliminate the warning.


corrados:
Thanks.

PR #653, opened 2020-10-02
Concurrency tweaks

kraney:
See #455 

With the included changes, I'm able to accomodate 96 clients on an 8-core server before there is packet loss and sound quality degrades. (mono/stereo, 128, no recording enabled.) Same system maxed out at 65 without these changes, all other things equal.

There are still some low-hanging fruit for improvements. There's some unnecessary work building a channel list inside the main decode loop; that could instead be updated only when a new channel joins or leaves (similar to the hashmap added in this changeset.)

Key changes
* Created a hashmap-based lookup of channel by hostaddress for use in PutPacket
* Increased the size of the thread pool
* Moved decode data into structs and made two pages of them, so decode can write one page while mix reads the other (then they trade.)
* Moved the futuresync wait to take advantage of this paging
* Split decode work into multiple threads by blocks similar to how mixing is done.

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -927,7 +927,7 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
             MixFutureSynchronizer.clearFutures();
 
             // Each thread must complete within the 1 or 2ms time budget for the timer.
-            const int iMaximumMixOpsInTimeBudget = 900;  // Approximate limit as observed on GCP e2-standard instance
+            const int iMaximumMixOpsInTimeBudget = 500;  // Approximate limit as observed on GCP e2-standard instance
                                                           // TODO - determine at startup by running a small benchmark
             const int iMTBlockSize = iMaximumMixOpsInTimeBudget / iNumClients; // number of ops = block size * total number of clients
             const int iNumBlocks   = ( iNumClients - 1 ) / iMTBlockSize + 1;
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1037,9 +1037,13 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
         // processing with multithreading
         if ( bUseMultithreading )
         {
+            // make sure all concurrent run threads from prior call have finished before we start again
+            FutureSynchronizer.waitForFinished();
+            FutureSynchronizer.clearFutures();
+
             // Each thread must complete within the 1 or 2ms time budget for the timer.
-            const int iMaximumMixOpsInTimeBudget = 500;  // Approximate limit as observed on GCP e2-standard instance
-                                                         // TODO - determine at startup by running a small benchmark
+            const int iMaximumMixOpsInTimeBudget = 1800;  // Approximate limit as observed on GCP e2-standard instance
+                                                          // TODO - determine at startup by running a small benchmark
             const int iMTBlockSize = iMaximumMixOpsInTimeBudget / iNumClients; // number of ops = block size * total number of clients
             const int iNumBlocks   = static_cast<int> ( std::ceil ( static_cast<double> ( iNumClients ) / iMTBlockSize ) );
 
@@ -1060,10 +1064,6 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
                                                                    iNumClients,
                                                                    iChannelDataPage ) );
             }
-
-            // make sure all concurrent run threads have finished when we leave this function
-            FutureSynchronizer.waitForFinished();
-            FutureSynchronizer.clearFutures();
         }
     }
     else
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -827,7 +827,7 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
         // process connected channels
         if (bUseMultithreading) {
             const int iDBlockSize = 20;
-            const int iNumBlocks   = static_cast<int> ( std::ceil ( static_cast<double> ( iNumClients ) / iDBlockSize ) );
+            const int iNumBlocks   = ( iNumClients - 1 ) / iDBlockSize + 1;
 
             for ( int iBlockCnt = 0; iBlockCnt < iNumBlocks; iBlockCnt++ )
             {
@@ -930,7 +930,7 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
             const int iMaximumMixOpsInTimeBudget = 900;  // Approximate limit as observed on GCP e2-standard instance
                                                           // TODO - determine at startup by running a small benchmark
             const int iMTBlockSize = iMaximumMixOpsInTimeBudget / iNumClients; // number of ops = block size * total number of clients
-            const int iNumBlocks   = static_cast<int> ( std::ceil ( static_cast<double> ( iNumClients ) / iMTBlockSize ) );
+            const int iNumBlocks   = ( iNumClients - 1 ) / iMTBlockSize + 1;
 
             for ( int iBlockCnt = 0; iBlockCnt < iNumBlocks; iBlockCnt++ )
             {
@@ -1089,6 +1089,7 @@ unsigned short CServer::DecodeDataBlocks ( const int iStartChanCnt,
                     {
                         emit ClientDisconnected ( currentChan.iChanID ); // TODO do this outside the mutex lock?
                     }
+                    hashChannelIndex.remove(vecChannels[currentChan.iChanID].GetAddress());
 
                     bChannelIsNowDisconnected = true;
                 }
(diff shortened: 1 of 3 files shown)


corrados:
Thanks for the code. We have some other changes in the queue for using float type for signal processing. This will give merge conflicts. The plan is to make a release this weekend. Then I integrate the float code and then I'll look at your code. In the meantime, people who need support for many clients at their servers can clone the code from your repo. That way well will get some feedback.

WolfganP:
Interesting changes @kraney, thx for the code. When you reach the sound degradation, did you find the root cause? (ie which resource became scarce for the app to fail?)

@corrados maybe is a better way forward to merge this PR to the existing multithreading_improvements branch to have all multithread efforts consolidated?

kraney:
@WolfganP I don't have a definitive answer but it seems clear it's a matter of exhausting the time budget for OnTimer causing a delay for the next timer. There are still available CPU cycles and memory on the system, and network bandwidth is not exhausted.

Observed behavior when the number of clients just exceeds the limit is that the network latency number begins to gradually increase, by something like 5-15 ms with each statistics update. Left unchanged it will keep going up until it shows ">500". That does not seem to be a true representation of the delay based on what I hear, however. But it does correspond to reduced sound quality. I ultimately settled on using that gradual increase as the definitive indicator that I'd exceeded the practical capacity of the server. The trigger seems to be around adding "yet another thread" some point after I have more threads than CPU cores. I am interpreting it as some threads being delayed by waiting for execution time, causing them to exceed the time budget - and thus delaying the next timer.

When I catch it creeping up, I can stop one or two clients and the latency number begins to gradually wind downwards at a similar pace. Sound quality becomes normal.

It seems likely there are incremental gains to be won just by getting very careful about exactly what work takes place in the decode / mix threads. And possibly some smaller gains by making other threads more efficient so they compete less for CPU time. There's a lot of "iterate over all channels (even disconnected ones)" taking place that I think could be avoided. The list of currently active channels in particular could be preserved between timers and maintained in a more efficient way.

kraney:
The approach I've taken in my changes was chosen to make the smallest, most incremental change to what currently exist that gets the job done. It isn't necessarily how I'd approach the problem starting from scratch.

What I would envision for a "from scratch" solution would be to have a "decode queue", a "record queue", a "frame queue", a "mix queue", and a "send queue". Each of these would ideally be a lockless circular buffer.

One or more decode threads would take items from the decode queue, do the decoding, and put them on the frame queue and optionally the record queue. The correct unit of work for the decode queue seems to be an audio frame from an individual channel, so the current PutAudio would be putting work on this queue.

One or more record threads would pull items off the record queue and write them to disk. This again seems to be doable at single audio frame / single channel, so that's why I put it at this stage.

There would be a thread to gather items from the framing queue and aggregate them until all channels are represented, then put that result on the mix queue. This "decoded data for one instant of time across all channels" is what I'm referring to as a "frame" and seems to be the correct unit of work for the mix threads. I suspect one thread will always be plenty for this task. It could probably be done with the exiting OnTimer mechanism. It would roll into the data cross-channel things like the mix levels and pan.

One or more mix threads would pick up a frame representing one audio buffer across all channels. In the 'multiple thread' case this is multiple readers of the same frame. They would do the mixing and then put each mixed result on the send queue.

In the end, this is essentially the same work that is already happening. But the separation provided by the queues allows the number of threads per task to be scaled independently, probably based on queue depth for each input queue. Threads would stick around rather than being transient. And a slow record write would not delay the mix pipeline (although at a certain point the record thread might not be able to keep up just based on disk bandwidth.)

src/server.cpp
```diff
@@ -798,10 +800,16 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
     bool bUpdateChannelLevels      = false;
     bool bSendChannelLevels        = false;
 
+    // flip to the idle page for the next call, in case it overlaps with the previous
+    iChannelDataPage = 1-iChannelDataPage;
```
corrados, line 804: I was also thinking of using a similar implementation as your "page". But this only makes sense if the decoding is done in one thread. The idea was that if the single decoding thread requires the complete time, the mixing/encoding could be done in parallel when the next frame is decoded. But as far as I understood, you now also spilt the decoding in multiple threads. Then it does not make any more sense to use the pages. You then use 100 % of the CPU for decoding and after that 100 % of the CPU for mixing and encoding. So you would not gain anything from doing decoding and mixing/encoding in parallel. Do you agree?

corrados:
> There are still some low-hanging fruit for improvements. There's some unnecessary work building a channel list inside the main decode loop; that could instead be updated only when a new channel joins or leaves (similar to the hashmap added in this changeset.)

I don't think the current implementation for the channel list building will consume any meaurable amount of time compared to all the decoding/mixing/encoding done for each channel. I would only change the existing code if it significantly gains speed to use the hashes. Have you measured the improvement by this change?

src/server.cpp
kraney: The FutureSync no longer occurs after the max threads are started, so they can continue to run after the OnTimer function has returned. That's when you get concurrency between decode and mix.

Instead, the FutureSync is called just _before_ launching new mix threads. So at that point it will block waiting for the last timer to completely finish before starting new mixes.

kraney:
I have not yet measured it. I don't know how much it contributes, I only know it's unnecessary work currently inside the most time sensitive path. Decoding and mixing have to be there. This does not.

kraney:
I implemented a quick and dirty check, and it seems your intuition is correct; removing that loop did not provide a significant gain.

kraney:
> > There are still some low-hanging fruit for improvements. There's some unnecessary work building a channel list inside the main decode loop; that could instead be updated only when a new channel joins or leaves (similar to the hashmap added in this changeset.)
> 
> I don't think the current implementation for the channel list building will consume any meaurable amount of time compared to all the decoding/mixing/encoding done for each channel. I would only change the existing code if it significantly gains speed to use the hashes. Have you measured the improvement by this change?

I think I originally misinterpreted the question.

Actually, I do think there is still benefit from pipelining decode in parallel with mix. The decode does not use 100% of the CPU, it uses a few cores at most. Less than half of them generally speaking. Then there are more cores that can be doing mixing at the same time; mix claims a higher proportion of cores in general. The mix from a prior timer can be wrapping up while the decode begins for the next one.

It's an assertion that's relatively easy to test, though, by just moving the MixFutureSync wait after the mix instead of before.

kraney:
I just ran a quick test with two changes to test this hypothesis.
* I changed the thread count to simply always be idealThreads() - one per core. It just divides the decode across all cores, then same for mix
* I removed the change that increases the threadpool size beyond idealThread()
* I moved the future sync after the mix, so it won't run at the same time as decode.

With that change, the max number of clients dropped to 86.

Next I tried keeping everything except moving the sync back so that mix can continue even as the next decode begins. This restored the limit to 92.

It's not a *huge* gain, so it suggests there's some truth to what you say.

Here what I believe to be the explanation - 
The mix threads are actually "mix and send" threads and are not actually compute bound, at least not in the final phase when it's time to send packets. They go into a wait state for the system calls to transmit the packets. While the thread sleeps waiting for its system call to complete, the decode threads are free to make use of the CPU. So a limited amount of overlap is beneficial. For the same reason, there's a small benefit to allocating more threads than cores - but not too many.

corrados:
I don't think we should use the "pages" at all. You found out that the decoding is a bottle neck. So we should parallelize that and try to use the CPU as much as we can. BTW, by using the pages and parallel processing of encoding/decoding we introduce additional audio latency. It is better to fulfill the decoding task as fast as possible, finish that task (by using the future-sync) and then do the mixing/encoding as fast as possible.

Please also read this post: https://github.com/corrados/jamulus/issues/596. You are working on 3) which is good :-). But you are also working on the core of Jamulus. So points 1) and 2) apply. Changes to that part of Jamulus are critical. I would like to make as little changes to the existing core code as possible. Just do changes which are an absolute necessity to improve things. Therefore I would not like to do changes on, e.g., the for-loop which finds connected channels since this code is proven to be stable and just works. If we do not gain speed, just keep the existing code.

kraney:
I understand your preference to minimize the changes.

I will point out that although overlapping decode  with mix can increase latency, a) it is by definition smaller than the time budget, and b) it only occurs once the number of clients has exceeded what can be handled without it. If the mix can finish before the next timer pops, it will still do so.

corrados:
Let's start with using multithreading on the decoding since we now know that this is a bottle neck. The task is to try to optimize this as much as we can. Then we can compare that result to what you have on your branch with the "pages".

I already started working on that on the master. I am doing it consistently with the multithreading implementation for the mixing/encoding. I'll manually take parts from your branch for that (i.e. I will not merge your branch for now). This work will be done in the next days so I expect it to be done until next weekend.

corrados:
I found out that the work was simpler than expected. I posted my test results here: https://github.com/corrados/jamulus/issues/455#issuecomment-703804934.

corrados:
@kraney I would like to close this Pull Request. I think most of your implementation is already taken over to the master branch. If there is a need to improve the multithreading performance further, we should base our modifications on the current code in the master branch. But as I read in the forums, the maximum number of musicians at a Jamulus server which were reported was about 50 which should be easily be supported by the current code on master. I would only start another round of improving the multithreading performance if there is an actual request for it.

kraney:
OK. I'm aiming for more but I can already achieve it from my fork. Eventually I'll rebase my remaining changes to the tip of master, so it'll be possible to easily reevaluate then if there is a desire to do so

corrados:
Ok, let's do it this way. Closed for now.

Issue #83, opened 2020-04-16
Android support


corrados:
_This issue has been migrated from Sourceforge.  Created: 2015-12-13
Creator: Volker Fischer_

Add support for the Android operating system. The software can already be compiled for Android but the audio interface does not work right now. So the first step is to finish the audio interface implementation. As a second step, all GUI functionality must be checked that it is possible to use it with a touch screen only.

corrados:
Some progress reported here: https://sourceforge.net/p/llcon/discussion/533517/thread/cfd165e0/#452b/6e3b

corrados:
See here: https://www.facebook.com/groups/507047599870191/permalink/521878088387142/

j-santander:
Is there any instructions on how to build this? or any specific branch/forked repository I could refer to? From the previous links there seems to be some implementation working to some extent. I might be able to give a hand, at least testing
Thanks

corrados:
The code on master is from Simon. But it does not work correctly. Recently, I have started a branch where I do some testing. But also, nothing working yet.

j-santander:
Ok, thanks. I got it to compile and started testing. 

There's an mistake on android/sound.cpp: on `CSound::onAudioRead()` when converting from incoming float buffer to int16_t:

```c++
                pSound->vecsTmpAudioSndCrdStereo[frmNum * oboeStream->getChannelCount() + channelNum] =
                    (short) floatData[frmNum * oboeStream->getChannelCount() + channelNum] * _MAXSHORT;
```
should be
```c++
                pSound->vecsTmpAudioSndCrdStereo[frmNum * oboeStream->getChannelCount() + channelNum] =
                    (int16_t)(floatData[frmNum * oboeStream->getChannelCount() + channelNum] * _MAXSHORT);
```

otherwise the precedence of the casting operator will make the input always 0.

Changing that if I write to a file the content `pSound->vecsTmpAudioSndCrdStereo` vector I got a correct signal.

To continue progressing I might need to understand better how the whole things work (I'm not at all familiar with neither your code or even the oboe library). A few doubts:

- Is it assumed that the sound card is always stereo? It seems so, but the oboe library seems to be configured in Mono. Changing that I seemed to started receiving a more-or-less-ok signal on another client. What other assumptions should be made for the signal coming from `CSound` -> `CClient` (sampling rate, frame size....)
- You got oboe setup for `float` samples. Is there any reason to not use `int16_t` samples? (we're doing the conversion by hand).
[...]

PR #661, opened 2020-10-05
Initial working Android Code

j-santander:
This pull request contains code changes to the already existing android code solving some of the issues. It is linked to Issue #83.

It has been minimally tested on a Google Pixel 3a but I could listen to other participants and they could hear me.

Key things are:

- Initialize the Oboe's FramesPerCallback to the proposed `iNewPrefMonoBufferSize` in `CSound::Init()`
- Only set up with a callback the input/recording stream. The output/playing stream will be written directly (non-blocking) everytime we get a callback on the input. This matches how CClient does things.
- Fixed the converstion from float to int16_t that previously resulted in the input being zero.
- Added a second vector (float) to contain the reformatted output samples. Once the default format becomes float, this could go away.

I guess there's still a lot of work to do in the handling of error cases and possible variations in the oboe behaviour across different devices, but I hope this could be a starting point.

Please comment.

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -1019,3 +1019,5 @@ contains(CONFIG, "disable_version_check") {
     message(The version check is disabled.)
     DEFINES += DISABLE_VERSION_CHECK
 }
+
+ANDROID_ABIS = armeabi-v7a arm64-v8a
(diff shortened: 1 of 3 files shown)


corrados:
Thanks. I'll merge it now and try it out on my Android phone.

PR #672, opened 2020-10-12
Implement simple audio compressor for all input and output.

hselasky:

### DIFF
--- a/src/compressor.h
+++ b/src/compressor.h
@@ -1,0 +1,91 @@
+/*
+   ZynAddSubFX - a software synthesizer
+
+   Compressor.h - simple audio compressor macros
+   Copyright (C) 2016 Hans Petter Selasky
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU General Public License
+  as published by the Free Software Foundation; either version 2
+  of the License, or (at your option) any later version.
+*/
+
+#pragma once
+
+static inline bool
+floatIsValid(const float x)
+{
+    const float __r = x * 0.0f;
+    return (__r == 0.0f || __r == -0.0f);
+}
+
+#define	stereoCompressor(div,pv,l,r) do {	\
+  /*						\
+   * Don't max the output range to avoid	\
+   * overflowing sample rate conversion and	\
+   * equalizer filters in the DSP's output	\
+   * path. Keep one 10th, 1dB, reserved.	\
+   */						\
+  static constexpr float __limit =		\
+      1.0f - (1.0f / 10.0f);			\
+  float __peak;					\
+						\
+  /* sanity checks */				\
+  __peak = (pv);				\
+  if (!floatIsValid(__peak))			\
+	__peak = 0.0;				\
+  if (!floatIsValid(l))				\
+	(l) = 0.0;				\
+  if (!floatIsValid(r))				\
+	(r) = 0.0;				\
+  /* compute maximum */				\
+  if ((l) < -__peak)				\
+    __peak = -(l);				\
+  else if ((l) > __peak)			\
+    __peak = (l);				\
+  if ((r) < -__peak)				\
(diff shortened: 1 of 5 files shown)


src/compressor.h
```diff
@@ -0,0 +1,89 @@
...
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU General Public License
+  as published by the Free Software Foundation; either version 2
+  of the License, or (at your option) any later version.
+*/
+
+#pragma once
+
+#define	floatIsValid(x) ({			\
```
atsampson, line 15: `({ ... })` is a GNU C extension, not standard C++, so it won't work with all compilers. It'd be better to make this a (static inline) function.

hselasky: Fixed.

corrados:
Introducing a compresser to solve a Windows audio interface bug is no option for me. The actual bug has to be fixed.

The preferred way of inserting audio effects to Jamulus is to insert the effect(s) in the audio path before the Jamulus client. E.g. in Jack Audio this is very simple but there exist tools for Windows and MacOS, too.

hselasky:
@corrados : I think it is not a problem in Jamulus. Regardless, can you test this patch?

hselasky:
@corrados : Please beware that some audio devices need additional space in the samples to do EQ // audio processing, and don't support using the full 24-bit range for example.

hselasky:
> The preferred way of inserting audio effects to Jamulus is to insert the effect(s) in the audio path before the Jamulus client. E.g. in Jack Audio this is very simple but there exist tools for Windows and MacOS, too.

On the server side this is not an option. I think the compressor is cheap enough that we simply put it into the Jamulus audio path. This replaces the simple range-clipping, which is _not_ working.

hselasky:
@corrados : In the old code everything was cast to 16-bit, so you never reached the limit of 24-bit. How can you say that this is a problem in the new code and not with your audio device?

I'm sorry, but I don't have any Windows machines here to test. I tried a VM on a MacBookPro, but that is terribly slow. Can you explain exactly what is clipping? Is the audio going through the OPUS codec via a server or is this for local loopback?

--HPS

corrados:
Sorry, there is an urgent request (https://github.com/corrados/jamulus/issues/619) which I want to work on with highest priority. So I don't have time to assist you here. Maybe bflamig can help.

bflamig:
**@hselasky**: I have to agree with **@corrados** that using the proposed compressor isn't addressing the real issue, whatever that is.  I also note that as far as I can tell, no other audio library that I can find needs such tweaks for 24 bit audio data.  Do you have documentation or links that describe this issue for certain devices that have this problem? They would be interesting to read.

By the way, we don't know that the current range clipping is not working. It's entirely possible it's not being called all the times it needs to be.

I don't have access to an audio device that uses 24 bit samples, so I can't debug this directly. A small test program needs to be written that uses a buffer of 24 bit data having the range extremes, and passing them through the conversion routines to see if there are any inherent problems. My personal belief is that the Jamulus code isn't really keeping the floating point values between -1.0 & 1.0. I think it's going over that at times, and that *will* cause overflow problems when converting to integer samples. That would sound nasty and perhaps that's what corrados is seeing.

Or, it's a simple mistake in the code, that we're just not realizing yet.

I'm currently working on an audio project of my own, and during the course of developing that code, I will be testing such conversions. Am in the process of doing that now, but I don't have a good estimate of when that will be accomplished. Only so many hours in a day! Perhaps I'll have some results over the weekend. But the only real test is to use a 24 bit device and see it in action with Jamulus, in a debugger.

corrados:
I'll close this pull request since the compressor does not fix the source of the problem and since HPS now has started his own jam software project (see https://github.com/corrados/jamulus/pull/539#issuecomment-727191851).

PR #675, opened 2020-10-12
Additional changes to the initial Android support

j-santander:
Related to Issue #83
Testing with different devices I found that the previous approach, where the output stream was written synchronously and non-blocking on the callback from the input stream, was not working for devices using the OpenSL backend.

So I've changed the implementation to:
- Have a circular buffer to store the output samples received from the CClient callback.
- Have the output stream set with a callback too that will extract samples from the circular buffer.

Added a Stats struct to keep track of some counters while experimenting.

I've tested this build in three devices:
- Google Pixel 3XL on Android 10.
- Google Pixel 3a on Android 11.
- BQ Aquaris X5 Plus on Android 7.1.1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ Makefile
 .project
 .settings
 .idea
+.vscode
 *.user
 *.user.*
 *.o
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ Makefile
 .cproject
 .project
 .settings
+.idea
 *.user
 *.user.*
 *.o
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -350,6 +350,7 @@ FORMS_GUI = src/clientdlgbase.ui \
     src/aboutdlgbase.ui
 
 HEADERS += src/buffer.h \
+    android/ring_buffer.h \
     src/channel.h \
     src/client.h \
     src/global.h \
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -1025,4 +1025,4 @@ contains(CONFIG, "disable_version_check") {
     DEFINES += DISABLE_VERSION_CHECK
 }
 
-ANDROID_ABIS = armeabi-v7a arm64-v8a
+ANDROID_ABIS = armeabi-v7a arm64-v8a x86 x86_64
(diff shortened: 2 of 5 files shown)


pljones:
Have you got the android builds somewhere they can be pull and sideloaded?  ARM and x86?  I've a few Android 8 devices (including a FireOS one... that'll be fun, I bet) I could try it on:
- Moto G5 Plus (Android 8.1.0)
- Lenovo Tab8 (Android 8.1.0)
- Nexus 7 (Android 7.whatever it got to)
- Fire 8 7" (who knows... 7 or 8 I think?)

(Yep, last two not to hand.)

j-santander:
Just uploaded it to my Google Drive:
https://drive.google.com/drive/folders/1LDgpKbqmNb2z42IhSHebuC-CpHNioqbW?usp=sharing
These are debug versions, unsigned of my latest build. Just checked and the
same apk seems to contain arm64, arm, x86 and x86_64 versions
Not sure how to setup anything more stable and reproducible.

pljones:
OK, I'll try more later.  It started up fine on my MotoG.  The connect dialog layout could be better (the column widths didn't use the screen too well).  Probably see other things too.  Not got as far as actual audio..! :)

pljones:
* The app should prevent sleep - even if I'm not powered, by default, I don't want the screen disappearing whilst I'm jamming.
* The ping on my MotoG is about 12-20ms worse than my Windows PC - I guess that's not unexpected as it's WiFi; same on the Tab8.
* At some point of trying to switch between the main window and connection window, the app got "confused"... if it's trying to be multi-window, that's not working for me.  It's repeatable, too: hit Connect, then use the connect window's menu to open View->Connect, then go back to the main window and hit Connect.  Neither the connect button and the main window menu View->Connect get me to the connection screen, I have to exit.

j-santander:
@pljones thanks for testing. The UI part is something I haven't modified, it's probably straight from the desktop version (not sure if there's any customization for Android).

I'll look into the device change issue.

pljones:
I think "Back" is an Android-only concept -- so it will need Android-specific handling.  I doubt that's been written at all.  It's probably being handled like "Alt-TAB" on Windows.  But there's no way (on Android 8, at least) of seeing all the in-app windows Jamulus has open and switching between them.  I can't think of a "natural" way to handle it that works for Android and other operating systems.

dzpex:
J-santander can you upload a new version of the apk....? i'm not ale to compile it i received more errors.....thanks

PR #686, opened 2020-10-22
Add missing #include <QVersionNumber> in classes that use it

atsampson:
It's already included in `clientdlg.h`.

This fixes building with Qt 5.7 on Debian stable -- it's OK with Qt 5.13 (presumably because one of the other headers transitively includes `QVersionNumber`).

### DIFF
--- a/src/channel.h
+++ b/src/channel.h
@@ -28,6 +28,7 @@
 #include <QDateTime>
 #include <QFile>
 #include <QTextStream>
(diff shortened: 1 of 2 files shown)


corrados:
Can you please take a look at this: https://travis-ci.org/github/corrados/jamulus/jobs/738215226
You must click on View Log. Thanks.

PR #691, opened 2020-10-23
Add version guards for QVersionNumber includes.

atsampson:
This header was introduced in Qt 5.6.

This should fix the build problem commented on in #686.


corrados:
Thanks.

PR #715, opened 2020-11-05
Updated and completed Polish translation

SeeLook:
- according to current git state
- on behalf of Martyna27 (former translator) and with her permission of
course


corrados:
Thank you!

PR #740, opened 2020-11-22
Avoid selecting IPv6 results from hostname lookup (#722)

### LINKED ISSUE #722: ParseNetworkAddress might select IPv6 addresses over IPv4 when a hostname has both
Hi, thanks a lot for Jamulus. I'm trying to take a look at the feasibility of adding IPv6 support, and I believe I found a bug along the way.

If a hostname has both A and AAAA records, `NetworkUtil::ParseNetworkAddress()` might pick the address from the AAAA record instead of the one from the A record, and fail to connect to a server running on that host.

When given a hostname, `NetworkUtil::ParseNetworkAddress()` uses `QHostInfo::fromName()` to perform the lookup, and currently picks the first result from the list. The [documentation of this method](https://doc.qt.io/qt-5/qhostinfo.html#fromName) does not specify anything about the order of the list for cases where there are multiple addresses. Looking a few layers down in the implementation, [it calls `getaddrinfo()`](https://github.com/qt/qtbase/blob/215400e153b6017ae33aadfa7a319c6daeed4f85/src/network/kernel/qhostinfo.cpp#L463), whose behaviour seems to be system-dependent (see e.g. https://stackoverflow.com/a/11326970).

On my computer, `getaddrinfo()` returns IPv6 addresses before IPv4 addresses (when both are available). So when I try to connect to a hostname that has both A and AAAA records, Jamulus selects the IPv6 address, and fails to connect to the server running on that host.

(curiously, what happens when an IPv6 `CHostAddress` makes its way down to `CSocket::SendPacket()` is that `toIPv4Address()` fails and returns 0, which is interpreted as `0.0.0.0`, so it tries to connect to `localhost`)

My suggestion is to make `NetworkUtil::ParseNetworkAddress()` look for the first IPv4 address in the list of returned addresses. The question is what to do if the returned result is non-empty, but only contains IPv6 addresses. Even though the rest of the function has some support for IPv6, I think it would make the most sense to return false when this happens. Right now, there's nothing useful that can be done with an IPv6 address anyway. If that sounds good, I could produce a pull request.

jarmar:
This is especially important for hostnames that have both IPv4 and IPv6 addresses, as `NetworkUtil::ParseNetworkAddress()` previously might have returned IPv6 addresses for such cases.

This solves connecting to localhost on IPv6-enabled systems.

fixes #722

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -982,18 +982,26 @@ bool NetworkUtil::ParseNetworkAddress ( QString       strAddress,
         // that the string contains a valid host name string
         const QHostInfo HostInfo = QHostInfo::fromName ( strAddress );
 
-        if ( HostInfo.error() == QHostInfo::NoError )
-        {
-            // apply IP address to QT object
-             if ( !HostInfo.addresses().isEmpty() )
-             {
-                 // use the first IP address
-                 InetAddr = HostInfo.addresses().first();
-             }
-        }
-        else
+        if ( HostInfo.error() != QHostInfo::NoError )
         {
             return false; // invalid address
+        }
+
+        // use the first IPv4 address, if any
+        bool bFoundIPv4 = false;
+        foreach ( const QHostAddress HostAddr, HostInfo.addresses() )
+        {
+            if ( HostAddr.protocol() == QAbstractSocket::IPv4Protocol )
+            {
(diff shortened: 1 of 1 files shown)


corrados:
Thank you.

PR #780, opened 2020-12-12
Retire manual.md

gilgongo:
See https://github.com/corrados/jamulus/issues/779


ann0see (commented):
From my POV this should be translatable

src/global.h
```diff
@@ -104,7 +104,7 @@ LED bar:      lbr
 // getting started and software manual URL
 #define CLIENT_GETTING_STARTED_URL       "https://jamulus.io/wiki/Getting-Started"
 #define SERVER_GETTING_STARTED_URL       "https://jamulus.io/wiki/Running-a-Server"
-#define SOFTWARE_MANUAL_URL              "https://github.com/corrados/jamulus/blob/master/src/res/homepage/manual.md"
+#define SOFTWARE_MANUAL_URL              "https://jamulus.io/wiki/Software-Manual"
```
ann0see, line 107: From my POV this should be translatable

src/res/homepage/ES/manual-es.md
```diff
@@ -1,209 +1,5 @@
...
-### Parámetro de estado de conexión actual
-
-![Indicators](indicators.png)
-
-El Ping es el tiempo que requiere el flujo de audio para viajar desde el cliente al servidor y volver. Este retardo lo determina la red y debería ser de unos 20-30 ms. Si este retardo es mayor de unos 50 ms, la distancia al servidor es demasiado grande o tu conexión a internet no es óptima.
-
-El retardo total se calcula con el ping y el retardo ocasionado por la configuración de buffers.
-
-La Tasa de Subida de Audio depende del tamaño actual de los paquetes de audio y la configuración de compresión de audio. Asegúrate de que la tasa de subida no es mayor que la velocidad de subida disponible (comprueba la tasa de subida de tu conexión a internet, por ejemplo con [speedtest.net](http://speedtest.net)).
+## Movido a https://jamulus.io/wiki/Software-Manual
```
ann0see, line 5: I'd link  https://jamulus.io/es/wiki/Software-Manual

src/global.h
ann0see, line 107: From my POV this should be translatable

src/res/homepage/ES/manual-es.md
ann0see, line 5: The spanish version should be linked, I think

gilgongo: OK I think I've done that (not sure).

src/global.h
gilgongo: What do you mean?

ann0see: It should automatically link to the language which is set in Jamulus.

gilgongo: Oh ok. Is that something we can do on this PR then?

pljones: Wouldn't that been done by the Wiki itself?

ann0see: No. The wiki doesn't redirect automatically. Nevertheless, it does show a red message over the navigation if the browser language is available.

gilgongo: I'm afraid I'd have no idea how to automate that.

ann0see: I don't even suggest to automate it. We could maybe make the message a bit more prominent.

pljones: What should happen if there's a translation set for Jamulus but not for that language on the wiki?  Just use the default pages?

I'm pretty sure the path could appear in the translation strings and have a `translate("wiki", url)` wrapper where used to get the translated value.

ann0see: > What should happen if there's a translation set for Jamulus but not for that language on the wiki? Just use the default pages (i.e. same as for any string not translated)?


Yes. I agree.

trebmuh: This string should be translatable, then Jamulus GUI translators can do their job when [translating their po files](https://github.com/corrados/jamulus/tree/master/src/res/translation).

PR #799, opened 2020-12-29
generate compiled translation files correctly

mirabilos:
- restore qmake’s `lrelease` option
- let the resource compiler pick up the compiled files correctly
- remove the precompiled binaries from the repository

Tested in Debian.

A possible improvement on this is to set `CONFIG += embed_translations` as well and **don’t** add them to `src/resources.qrc` (but let qmake do that); we’d have to redo the part that lists the available translations in `src/util.cpp:CLocale::GetAvailableTranslations()` then (but given it maps all languages from *xx_XX* to *xx* except Portuguese is doubled, some more insightful algorithm might be used there anyway, or the consumers could just deal with having the countries there as well).

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -668,6 +668,16 @@ DISTFILES += ChangeLog \
     README.md \
     distributions/jamulus.desktop.in \
     distributions/jamulus.png \
+    src/res/translation/translation_de_DE.qm \
+    src/res/translation/translation_fr_FR.qm \
+    src/res/translation/translation_pt_PT.qm \
+    src/res/translation/translation_pt_BR.qm \
+    src/res/translation/translation_es_ES.qm \
+    src/res/translation/translation_nl_NL.qm \
+    src/res/translation/translation_pl_PL.qm \
+    src/res/translation/translation_it_IT.qm \
+    src/res/translation/translation_sv_SE.qm \
+    src/res/translation/translation_sk_SK.qm \
     src/res/CLEDBlack.png \
     src/res/CLEDBlackSmall.png \
     src/res/CLEDDisabledSmall.png \
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -8,6 +8,7 @@ contains(CONFIG, "noupcasename") {
 
 CONFIG += qt \
     thread \
+    lrelease \
     release
 
 QT += network \
@@ -21,6 +22,7 @@ contains(CONFIG, "headless") {
     QT += widgets
 }
 
+LRELEASE_DIR = src/res/translation
 TRANSLATIONS = src/res/translation/translation_de_DE.ts \
     src/res/translation/translation_fr_FR.ts \
     src/res/translation/translation_pt_PT.ts \
@@ -666,16 +668,6 @@ DISTFILES += ChangeLog \
     README.md \
     distributions/jamulus.desktop.in \
     distributions/jamulus.png \
-    src/res/translation/translation_de_DE.qm \
-    src/res/translation/translation_fr_FR.qm \
-    src/res/translation/translation_pt_PT.qm \
-    src/res/translation/translation_pt_BR.qm \
-    src/res/translation/translation_es_ES.qm \
-    src/res/translation/translation_nl_NL.qm \
-    src/res/translation/translation_pl_PL.qm \
-    src/res/translation/translation_it_IT.qm \
-    src/res/translation/translation_sv_SE.qm \
-    src/res/translation/translation_sk_SK.qm \
     src/res/CLEDBlack.png \
     src/res/CLEDBlackSmall.png \
     src/res/CLEDDisabledSmall.png \


mirabilos:
… wait, this fails on buster (but succeeds on sid, hmm)…

mirabilos:
I think for a proper fix we _must_ use `embed_translations` as well :|

mirabilos:
qmake in buster doesn’t seem to call lrelease at all‽

mirabilos:
Right, buster is missing `/usr/lib/«MULTIARCH»/qt5/mkspecs/features/lrelease.prf` from `qt5-qmake`… Qt too old, I guess…

mirabilos:
Ah right, Qt 5.12 and newer…

corrados:
Thanks but Jamulus should support all Qt 5 versions. How do we solve this?

mirabilos:
Turns out one cannot simply add `lrelease.prf` to the buster build. I’ve been trying to track down why. I *believe* it is this:

```
97d7a00989f8d5df15eb68d90c1af547b67b3fa4
Author: Oswald Buddenhagen <oswald.buddenhagen@qt.io>
Date:   Wed Mar 21 18:03:10 2018 +0100

    rcc: don't drop entries with missing files in -list mode
    
    the files may not exist _yet_. this change allows dynamically generating
    the resource contents.
    
    the ignoreErrors parameter is renamed to listMode and made less
    aggressive, to better reflect the actual usage.
    
    Change-Id: I2f6a75a23f1ef903f0d957f9a09f4df0ce2a2b35
    Reviewed-by: Thiago Macieira <thiago.macieira@intel.com>
    Reviewed-by: hjk <hjk@qt.io>
    Reviewed-by: Kai Koehne <kai.koehne@qt.io>
```

Since this changes the C++ part of Qt itself… bad luck.

mirabilos:
Let’s see whether this works. Let users either add `CONFIG+=force-lrelease` to the qmake invocation or run lrelease manually or use the precompiled versions.

Update:

Hmm, considering that `CONFIG+=lrelease` doesn’t do anything in Qt < 5.12, do we even need to make it conditional? I think not…

corrados:
How is your new code supposed to work? If I clone from your repo/branch and delete the *.qm files in the translation directory, under Windows I then get compiler errors. I thought that with lrelease the *.qm files are created on every build?

mirabilos:
Volker Fischer dixit:

>How is your new code supposed to work? If I clone from your repo/branch
>and delete the *.qm files in the translation directory, under Windows I
>then get compiler errors. I thought that with lrelease the *.qm files
>are created on every build?

Hm weird, it works for me, and yes they are.

Are you sure you’re using Qt 5.12 or newer?

If using Qt 5.11 or older, you have to either not delete the .qm files
or run lrelease manually before the build. (So no change to before.)

corrados:
I have Qt version 5.15.2 on Windows.

corrados:
How do we proceed with this pull request?

mirabilos:
Volker Fischer dixit:

>How do we proceed with this pull request?

Hm, good question. Someone should probably look at why this
doesn’t work on Windows®, but that one is not me, I think
the win2k installation I have on an old laptop somewhere is
not… exactly helpful.

If you commit this change but do NOT delete the .qm binary
files for now, will it break anything? In that case, I’d
expect the lrelease run to update the .qm files if it works
and the build to use the committed files if not.

This would allow me to exclude the binary files from the
Debian source and not need to patch the build system, as
lrelease “just works”, in bullseye/sid anyway, and for the
backport, I just run it manually before building:

https://evolvis.org/plugins/scmgit/cgi-bin/gitweb.cgi?p=alioth/jamulus.git;a=commitdiff;h=refs/heads/buster-backports;hp=refs/heads/master

corrados:
> If you commit this change but do NOT delete the .qm binary files for now, will it break anything?

I don't think so. To make sure I'll try that once more.

I am still a bit concerned about this change: https://github.com/corrados/jamulus/pull/799/files#diff-3b059d403dbbf0c4c127642ee3318bc2c3b10f70f929b71e2746c4a6ef907244L654. This will influence the tar.gz release of Jamulus. If someone uses a Linux distrubition with an old Qt version, the .qm files will not be created and since they are no longer distributed, he will not have them at all. What do you think?

mirabilos:
Volker Fischer dixit:

>> If you commit this change but do NOT delete the .qm binary files for now, will it break anything?
>
>I don't think so. To make sure I'll try that once more.

OK, thanks.

>This will influence the tar.gz release of Jamulus. If someone uses a
>Linux distrubition with an old Qt version, the .qm files will not be
>created and since they are no longer distributed, he will not have them
>at all. What do you think?

Leave that part out (or, maybe better, revert it with an explicit
commit saying that the files need to stay part of that .tar.gz for
now until we can either assume working lrelease on users’ parts or
require them to run lrelease manually like I did in the buster
branch — this will allow later “git revert”ing the partial revert
once no longer needed).

If you wish I can add such a commit.

corrados:
> If you wish I can add such a commit.

Yes, please. Thanks.

corrados:
> > If you commit this change but do NOT delete the .qm binary files for now, will it break anything?

> I don't think so. To make sure I'll try that once more.

I just checked it without deleting the .qm files and it compiles and runs fine on my Windows.

mirabilos:
Volker Fischer dixit:

>> If you wish I can add such a commit.
>
>Yes, please. Thanks.

I just did so, and rebased on latest master while at it.

corrados:
Thank you. Do you need a official Jamulus release soon or is it enough for now to have it in the Git repo?

mirabilos:
Volker Fischer dixit:

>Thank you. Do you need a official Jamulus release soon or is it enough for now to have it in the Git repo?

In the repo is enough, unless you changed much since 3.6.2 which is
the version I recently uploaded hoping it’ll make into the next
stable release, and it’s carrying this as a local patch.

Thanks!

PR #800, opened 2020-12-29
New windows installer now checks if old version of Jamulus is wrongly installed

ann0see:
Only tested on 64 bit Windows 10. I don't know how it behaves on 32 Bit Windows but it should work. Not 100% sure though. 

Originally requested by
@drummer1154


ann0see:
Also see https://github.com/ann0see/jamulus/pull/11

Issue #764, opened 2020-12-05
iOS Support


abbanatic:
Hi,

please be so kind to think about providing iOS support for Jamulus in the future.

Due to Apple's change from Intel to ARM, universal binaries for all of Apple's platforms became reality.

So it would be great for Jamulus to support the New Apple Silicon AND iOS devices, with their great built-in capabilities regarding Audio and MIDI.

Thank you very much!
Thomas

ann0see:
Agree. Porting should be possible but there would still be the same two problems like for Android (which is in development):

1. how to deal with WiFi (of course you could use USB to LAN adapters)
2. How to deal with Audio Interfaces (iOS uses CoreAudio, so does macOS)

Something iOS specific:

Submitting to the AppStore would require Jamulus to be dual licensed. Apple doesn't like the GPL.

ann0see:
First of all, I don't have any knowledge of iOS development. Therefore somebody experienced should look at this issue. I created a new iOS folder, edited the .pro file and copied the macOS files in there. I made some tests:

I tried to compile Jamulus with QT Creator for iOS and it (of course) failed. 

`In file included from /.../jamulus/ios/sound.cpp:25:
/.../jamulus/ios/sound.h:70:5: error: unknown type name 'AudioDeviceID'; did you mean 'AudioFileID'?` 

Will probably be one of the main issues. 

also in client.h the include statement has to be changed to ask if it's being built for iOS.

gene96817:
Apple announced at WDC2020 some kind of merging macOS and iOS for 2021. I don't follow the details in this area closely. This suggests people interested in iOS support should look at the developer's tools for macOS 12.x.x and this "merger" with iOS. Of course, this doesn't solve the needs of devices that run old iOS release.

ann0see:
I got a headless Jamulus server up and running on my iPhone SE 1st gen (although it's connected via wifi, the quality is ok). Everything GUI related doesn't work and also I assume files can't be written because of the iOS sandbox.

ann0see:
https://developer.apple.com/documentation/avfoundation/avaudiosession?language=objc in https://developer.apple.com/documentation/avfoundation/avaudiosessionmodemeasurement mode could be a good choice for audio in iOS. Does anybody know anything about iOS Audio/iOS CoreAudio?

ann0see:
Although I have no idea how to write the sound.cpp or sound.h files or StoryBoard files etc. have a look at my changes: 

[...]

PR #802, opened 2020-12-30
Add iOS compilation support

ann0see:
#764 

Maybe the first steps for iOS support. (Mainly copied files from the macOS version)
What works at the moment: "nothing" except compilation for iOS and a headless server: I was able to run on my iPhone SE 1st gen

Todo:
A lot:

1. Audio support (there's nothing working at the moment). I only ask for microphone access
2. GUI support (once I run the compiled app file, I only get a black screen)

I'm still testing this (especially if it really compiles since I made some changes to it)

How to compile: Execute /path/to/qmake Jamulus.pro -spec macx-ios-clang CONFIG+=iphonesimulator CONFIG+=simulator CONFIG+=qml_debug and open the resulting Xcode project file.

Disclaimer: iOS development and Objective C are new to me and I would love if somebody had a look at iOS support.

Could somebody please verify that this doesn't break compiling for macOS?


ios/activity.h
```diff
@@ -0,0 +1,46 @@
+/******************************************************************************\
+ * Copyright (c) 2004-2020
+ *
+ * Author(s):
+ *  AronVietti
```
corrados, line 5: Is this "activity.h" and "activity.m" a copy from the files in mac?

ios/ios_app_delegate.mm
```diff
@@ -0,0 +1 @@
+
```
corrados, line 1: We need an empty file?

ios/sound.h
```diff
@@ -0,0 +1,108 @@
+/******************************************************************************\
+ * Copyright (c) 2004-2020
+ *
+ * Author(s):
+ *  Volker Fischer
```
corrados, line 5: I have not written this file.

ios/sound.mm
```diff
@@ -0,0 +1,232 @@
+/******************************************************************************\
+ * Copyright (c) 2004-2020
+ *
+ * Author(s):
+ *  Volker Fischer
```
corrados, line 5: I have not written this file. BTW: What is the ending ".mm"?

src/main.cpp
```diff
@@ -577,9 +580,13 @@ int main ( int argc, char** argv )
 #ifdef HEADLESS
     QCoreApplication* pApp = new QCoreApplication ( argc, argv );
 #else
+# if defined (Q_OS_IOS)
```
corrados, line 583: I know I am picky ;-) -> Two spaces missing.

ios/activity.h
ann0see: Yes.

ios/ios_app_delegate.mm
ann0see: It was a test to overcome a message during runtime (there was and still is a warning during runtime that you need to set a property in the app delegate for ui view or similar. Not exactly sure how I can solve this therefore I've just created this file since we'll probably need it.)

ios/sound.h
ann0see: As I said, it's mainly a copy from macOS.

ios/sound.mm
ann0see: .mm means that it's a objective c++ file --> you can mix c++ and objective c in this file. This is needed (as far as I understand) to access the iOS sound API

src/main.cpp
ann0see: Should be solved

ios/activity.h
corrados: I do not want to have exact copies. Why not using the existing files?

ios/sound.h
corrados: Still, I have not created this file. You have created it. So you should write your name in this file.

src/main.cpp
```diff
@@ -580,7 +580,7 @@ int main ( int argc, char** argv )
 #ifdef HEADLESS
     QCoreApplication* pApp = new QCoreApplication ( argc, argv );
 #else
-# if defined (Q_OS_IOS)
+# if defined (  Q_OS_IOS  )
```
corrados, line 583: Have you now used two spaces on each side of the bracket? :-) So you inserted a total of 4 spaces but only 2 are needed ;-)

ann0see: Ok. Is this ok now?

ios/activity.h
ann0see: We could use the Mac ones, but in my opinion this would be confusing since we include Mac files on iOS (while both are different operating systems). The cleanest way would be to create an "Apple" folder with only these files. Not even 100% sure if they are needed on iOS since the server stopped once I locked my device.

ios/sound.h
ann0see: Ok. I've added my name. But since you wrote most of it for MacOS (at least that's what I assume), I've now written "ann0see based on code from Volker Fischer". I hope that's ok?

ios/activity.h
corrados: Please check if it is really required. If not, just remove it. Otherwise you should use the one in the mac directory. It is not the cleanest way but better than having two identical files.

ios/ios_app_delegate.mm
corrados: Ok, so let's keep it for now.

ios/sound.h
corrados: Yes, that's ok.

src/main.cpp
corrados: Yes, thanks.

ios/activity.h
ann0see: Should be gone now. The behaviour of the headless server is the same without this. So I think we don't need this file. I've now added the typical app delegate code to the app delegate files, but it doesn't seem to change anything (except that the warning I mentioned has now disappeared)

ann0see:
The app now automatically starts as server (see main.cpp)

ann0see:
The gui problem might be related to this: https://forum.qt.io/topic/118060/the-opengl-display-stays-black-under-ios/10

Jamulus.pro
```diff
@@ -146,6 +146,22 @@ win32 {
...
+} else:ios {
+    QMAKE_INFO_PLIST = ios/Info.plist
+    QT += macextras
+    OBJECTIVE_SOURCES += ios/ios_app_delegate.mm
+    HEADERS += ios/ios_app_delegate.h
+    HEADERS += ios/sound.h
+    OBJECTIVE_SOURCES += ios/sound.mm
+    QMAKE_TARGET_BUNDLE_PREFIX = com.corrados.jamulus
+    QMAKE_APPLICATION_BUNDLE_NAME. = $$TARGET
+‚
```
corrados, line 158: What is this comma for?

ann0see: Probably it's not needed.

corrados:
Thanks!

PR #806, opened 2021-01-02
Win installer: make it possible to unselect desktop Shortcut 

ann0see:
This should also fix the $PWD issue

### DIFF
--- a/windows/installer.nsi
+++ b/windows/installer.nsi
@@ -77,15 +77,13 @@ LangString RUNNING_APP_MSG ${LANG_ENGLISH} \
 ; LangString RUNNING_APP_MSG ${LANG_ITALIAN} \
 ;   "${APP_NAME} è in esecuzione. Chiudere l'applicazione prima di eseguire l'installazione."
 
-LangString OLD_WRONG_REG_FOUND_EMPTY ${LANG_ENGLISH} \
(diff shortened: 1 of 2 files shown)


windows/installer.nsi
```diff
@@ -281,7 +288,11 @@ Section "un.Install"
     ${EndIf}
 
     ; Remove the Start Menu and desktop shortcuts
-    Delete   "$DESKTOP\${APP_NAME}.lnk"
+    IfFileExists "$DESKTOP\${APP_NAME}.lnk" deletedtshortcut continueremove
+    deletedtshortcut:
```
pljones, line 292: `deletedtshortcut` -- should be `deletetshortcut` (i.e. not past tense).

windows/installer.nsi
```diff
@@ -281,7 +288,11 @@ Section "un.Install"
     ${EndIf}
 
     ; Remove the Start Menu and desktop shortcuts
-    Delete   "$DESKTOP\${APP_NAME}.lnk"
+    IfFileExists "$DESKTOP\${APP_NAME}.lnk" deletedtshortcut continueremove
```
pljones, line 291: The `deletedtshortcut` is visually hard to parse...  Perhaps just `deleteshortcut`.  Also, I'd prefer `continueremove` to be `skipshortcut`, as it's the "else" to this "if".

ann0see: Should be done now

Issue #795, opened 2020-12-23
Automatic build of official Debian packages


ann0see:
As far as I know, Jamulus is not included in the official Debian repository.

Compiling Jamulus needs a bunch of dependencies (which in my case resulted in a GNOME desktop install on my server) which I rather want to avoid on my server.

It would be great if there were official Debian packages which are released via GitHub releases. These packages could be built automatically by GitHub Actions on every release.

We already have the necessary scripts in this repository, so it should be easy to implement.

chrisrimple:
I'm running Jamulus Server on Debian at Vultr.com.  I followed [Andrew Ferguson's install instructions](https://andrewferguson.net/2020/07/08/jamulus-and-temporally-hyper-near-servers/), then update to new Jamulus versions with the steps shown in my [Remote Band Rehearsals doc](https://docs.google.com/document/d/1smcvsxdaaViPQvGMQHmah_6BQeqowhmGSFMHfnlY2FI/edit#heading=h.ttnch1lxf6f).  Anything that makes it easier is a good thing, in my mind.  :)

ann0see:
https://github.com/marketplace/actions/automatic-releases
this could be modified.

ann0see:
My work:

https://github.com/ann0see/jamulus/blob/main/.github/workflows/autorelease.yml

https://github.com/ann0see/jamulus/releases/tag/latest

38github:
ann0see: Thank you for your effort. I greatly appreciate it. I managed to install it on Debian 10 and it runs but apt complained about libgcc10-s1 not being available:
```
Unpacking jamulus (3.6.2git-0) over (3.6.2-0~~xenial1) ...
dpkg: dependency problems prevent configuration of jamulus:
 jamulus depends on libgcc-s1 (>= 3.0); however:
  Package libgcc-s1 is not installed.
```

I tried your deb because when I was using the Xenial build from PPA I could not see any pan pots but after installing your build they appeared. Yes, I did have stereo and everything configured correctly with xenial build but still no pan pots.

Questions:
* Can you build it for latest stable (Debain 10)?
* If so, also the latest stable releases (right now 3.6.2)?

ann0see:
@38github I must admit, that I didn't do much (all the debian packaging tools etc. were already there. I just did the GH-Actions stuff) and I haven't had the time to install and check the packages yet on a real machine. I just looked at the content and it looked fine. 

[...]

PR #807, opened 2021-01-03
Automatically build Debian packages and set latest tag to release

ann0see:
#795 

This will make the release process slightly different (mainly the latest tag will be set automatically).

If somebody adds a tag with r and pushes it to the repo, GitHub actions will compile the .deb files from the latest code and upload it to a release: https://github.com/ann0see/jamulus/releases/tag/latest

Hacks/Bugs:

I currently compile Jamulus on Ubuntu 18.04 but use the 16.04 dependencies. This seems to allow us to have the pan function on Debian 10 but doesn't show the muted icon in the main mixer. Hopefully it doesn't break anything else

### DIFF
--- a/.github/workflows/autobuild_deb.yml
+++ b/.github/workflows/autobuild_deb.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   pre-release:
     name: "Pre Release"
-    runs-on: "ubuntu-20.04"
+    runs-on: "ubuntu-16.04"
     steps:
       - name: Checkout Project
         uses: actions/checkout@v1
--- a/.github/workflows/autobuild_deb.yml
+++ b/.github/workflows/autobuild_deb.yml
@@ -1,4 +1,4 @@
-name: "pre-release"
+name: "Autobuild deb"
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


corrados:
Does this mean that everytime I set a new label (i.e. any label), the "latest" tag will be moved and a release will be created? This is not intended since sometimes I apply some labels to some intermediate states of the code. I do not want the "release" label to point to that tag.
How can we solve this?

corrados:
I just saw that you already filter the tags by "r*". So I only have to make sure that the inofficial tags do not start with an "r", right?

corrados:
> This will make the release process slightly different (mainly the latest tag will be set automatically).

So you mean that instead of tagging the repo and afterwards move the latest tag -> now I only set the release tag and the movement of the latest tag is done automatically? So it means that I have less work to do (which would be good :-) ).

ann0see:
> So I only have to make sure that the inofficial tags do not start with an "r", right?

Yes. That's the case

> So you mean that instead of tagging the repo and afterwards move the latest tag -> now I only set the release tag and the movement of the latest tag is done automatically? So it means that I have less work to do (which would be good :-) ).

Yes.

ann0see:
In theory macOS and Windows builds could be possible with GH actions too. But I haven't managed it yet (and would like to focus on one OS first)

corrados:
Great, thanks.

corrados:
> In theory macOS and Windows builds could be possible with GH actions too.

If you could figure out how to do it, that would be great.

Issue #803, opened 2021-01-01
Flatpak build


LeandroStanger:
Please, consider supporting [Flatpak](http://docs.flatpak.org/en/latest/introduction.html#reasons-to-use-flatpak) and publishing Jamulus on [Flathub](https://github.com/flathub/flathub/wiki/App-Submission).

corrados:
Can you do it?

38github:
Finally this is soon a reality and we don't have to wait for distros to update their repositories. Even old dstros with out-dated dependies will probably be able to run latest versions.

corrados:
Great. Any volunteers?

38github:
If I had the knowledge I would be on it right away!

ranfdev:
I have a christmas present for you
```json
{
  "app-id": "io.jamulus.Jamulus",
  "runtime": "org.kde.Platform",
  "runtime-version": "5.15",
  "sdk": "org.kde.Sdk",
  "command": "Jamulus",
  "rename-desktop-file": "jamulus.desktop",
  "rename-icon": "jamulus",
  "finish-args": [
    "--share=ipc",
    "--socket=x11",
    "--socket=wayland",
    "--filesystem=host",
    "--device=dri"
  ],
  "modules": [
    {
      "name": "jack2",
      "buildsystem": "simple",
      "build-commands": [
      	"./waf configure --prefix=/app --htmldir=/app/share/doc/jack/ --classic",
      	"./waf build -j $FLATPAK_BUILDER_N_JOBS",
      	"./waf install"
      ],
      "config-opts": [
      ],
      "sources": [
        {
          "type": "archive",
          "url": "https://github.com/jackaudio/jack2/archive/v1.9.16.tar.gz",
          "sha256": "e176d04de94dcaa3f9d32ca1825091e1b938783a78c84e7466abd06af7637d37"
        }
      ]
    },
    {
      "name": "jamulus",
      "buildsystem": "qmake",
      "config-opts": [
      ],
      "sources": [
        {
          "type": "archive",
          "url": "https://github.com/corrados/jamulus/archive/r3_6_2.tar.gz",
          "sha256": "6bea992f0e4b6a9d08104f8ad9d42b630d523da258e201f70418cecda1c11dac"
        }
      ]
    }
  ]
}
```

It's the flatpak manifest needed to build Jamulus. You can save it as `io.jamulus.Jamulus.json`
Then:
```sh
# To build and install
[...]

PR #810, opened 2021-01-04
Add flatpak build support, with CI

ranfdev:
Warning: I can say it compiles and runs, but I don't have jack installed and my sound card isn't working, so I can't test anything.
Before this PR gets accepted, someone should test this.

To manually build the flatpak release, you can run
```
flatpak-builder build distributions/io.jamulus.Jamulus.json --force-clean --install --user
```

Then Jamulus should show up in your app list.
If you want to launch it from the terminal

```
flatpak run io.jamulus.Jamulus
```

This solves #803 
Enjoy

### DIFF
--- a/.github/workflows/autobuild_flatpak.yml
+++ b/.github/workflows/autobuild_flatpak.yml
@@ -2,6 +2,7 @@ on:
   push:
     tags:
       - "r*"
+  workflow_dispatch:
 name: "Autobuild Flatpak"
 jobs:
   flatpak-builder:
--- a/.github/workflows/autobuild_flatpak.yml
+++ b/.github/workflows/autobuild_flatpak.yml
@@ -1,0 +1,18 @@
+on:
+  push:
+    tags:
+      - "r*"
+name: "Autobuild Flatpak"
+jobs:
+  flatpak-builder:
+    name: "Flatpak Builder"
+    runs-on: ubuntu-latest
+    container:
+      image: docker.io/bilelmoussaoui/flatpak-github-actions
+      options: --privileged
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


ann0see:
Great. Does this upload the build anywhere?

ranfdev:
Yes, you can find the build artifact when viewing the github action

ann0see:
Ok. But you can't download it after the build artifact is deleted? We might need to make modifications to this file:

https://github.com/corrados/jamulus/blob/master/.github/workflows/autobuild_deb.yml

to allow automatic upload to a release.

ann0see:
I haven’t tested it. So probably this is untested code.

corrados:
Does it make sense to merge it anyway? As far as I understood, if I apply a release tag, everything works automtically, right?

ann0see:
>  As far as I understood, if I apply a release tag, everything works automtically, right?

Yes. Everything will run in GH actions and you can have a look at it in the actions tab.

corrados:
Good that you mention that. Github Actions are currently disabled for the corrados/jamulus repo. Do you have it set to "Allow all actions" in your repo?

ann0see:
No. Actually I think it was already enabled for me.

corrados:
Ok, Github Actions are now enabled. Since this Pull Request is just about adding an additional action which may or may not work but does not break anything existing, I'll apply this now and when the next release is done, we see if it worked or not. If it not worked, then we can fix the issue for the upcoming Jamulus version.

ann0see:
Ok. I could only speak for my action. 

Have a look at this: https://github.community/t/workflow-error-action-not-allowed/16788/4

since the workflow seems to fail.

Issue #811, opened 2021-01-05
Channel Layout Minor Tweaks


jdrage:
Couple of minor layout bugs I noticed:

1. When unmuting and muting, the appearance of the info icon causes the layout to jitter a little -- the fader bounces up and down a few pixels. (Repros on all skins.)
2. On the Fancy skin, the fader handle isn't centered over the slot.

jdrage:
I have fixes for these, will create a PR shortly.

corrados:
Pull Request is merged.

corrados:
I actually don't like the new implementation. This is the spacings we had before:
![grafik](https://user-images.githubusercontent.com/46655886/104815238-a2918200-5813-11eb-808f-dcc887134182.png)
And this is with the new code:
![grafik](https://user-images.githubusercontent.com/46655886/104815249-b50bbb80-5813-11eb-928c-0d72c947f036.png)
This is definitely too much wasted space. I'll undo the change for now.

corrados:
I just saw that the compact skin was also affected by the change:
![grafik](https://user-images.githubusercontent.com/46655886/104854262-fb414780-5905-11eb-977d-240e6778477c.png)
So, it is better to disable it and find a better solution...

gene96817:
Would it help to rotate the text 90 degrees counterclockwise?

corrados:
There is no issue with the compact skin. This is how it is supposed to look like: https://github.com/corrados/jamulus/issues/339#issuecomment-661153946.

The issue was just with the patch provided by jdrage in his Pull Request https://github.com/corrados/jamulus/pull/812. But as I said, I reverted that patch.

corrados:
With my latest commit, at least on Windows the layout does not jump on Mute now:
![xwGQgOT8nx](https://user-images.githubusercontent.com/46655886/105066593-3fa42300-5a7f-11eb-8b9f-512607983bd8.gif)

PR #812, opened 2021-01-05
Fix for #811 - Channel Layout Minor Tweaks

jdrage:
Minor tweaks for the channel layout

1. Set a minimum size on the info widget so that if the text content is empty it still occupies space in the layout.
2. Adjust the horizontal alignment of the fader handle to center it over the slot.

**How built:** Windows Qt/VS2019 x86
**How verified:** Windows Desktop
- Server: FamJam
- Normal/Fancy/Compact skins

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -42,6 +42,7 @@ CChannelFader::CChannelFader ( QWidget* pNW ) :
     pPan                        = new QDial       ( pLevelsBox );
     pPanLabel                   = new QLabel      ( tr ( "Pan" ), pLevelsBox );
     pInfoLabel                  = new QLabel      ( "", pLevelsBox );
+    pInfoLabel->setMinimumSize(36, 36); // prevents jitter when muting/unmuting
 
     pMuteSoloBox                = new QWidget     ( pFrame );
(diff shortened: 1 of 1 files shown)


src/audiomixerboard.cpp
```diff
@@ -42,6 +42,7 @@ CChannelFader::CChannelFader ( QWidget* pNW ) :
     pPan                        = new QDial       ( pLevelsBox );
     pPanLabel                   = new QLabel      ( tr ( "Pan" ), pLevelsBox );
     pInfoLabel                  = new QLabel      ( "", pLevelsBox );
+    pInfoLabel->setMinimumSize(36, 36); // prevents jitter when muting/unmuting
```
pljones, line 45: Please follow the coding standards - spaces before and after braces.  (I think also a blank line before line 44 to keep these two lines grouped would be good, too.)

corrados:
> When unmuting and muting, the appearance of the info icon causes the layout to jitter a little -- the fader bounces up and down a few pixels.

I just checked it with the current code and running Jamulus on Linux. It does not bounce on Linux. But on MacOS the icon is even larger than on Windows. I'll test your code on MacOS, soon, to make sure that it does not bounce even with your change.

corrados:
Thanks.

PR #819, opened 2021-01-08
Win Installer: Rewrite 32 bit OS install detection

ann0see:
See https://github.com/corrados/jamulus/issues/792#issuecomment-756425457

### DIFF
--- a/windows/installer.nsi
+++ b/windows/installer.nsi
@@ -83,7 +83,7 @@ LangString RUNNING_APP_MSG ${LANG_ENGLISH} \
 ;   "${APP_NAME} è in esecuzione. Chiudere l'applicazione prima di eseguire l'installazione."
 
 LangString OLD_WRONG_VER_FOUND ${LANG_ENGLISH} \
-    "Due to a bug, an old version of Jamulus might be installed to a wrong path on your computer. Do you want to remove it before installing this new version (we strongly recommend this)?"
+    "Due to a bug, a 64 Bit version of Jamulus might be installed to your 32 Bit Program Files folder. Do you want to remove it before installing this new version (we strongly recommend this)?"
 
 LangString OLD_WRONG_VER_FOUND_CONFIRM ${LANG_ENGLISH} \
     "If you continue without removing it, your installation might be broken! Are you sure you don't want to remove the old version?"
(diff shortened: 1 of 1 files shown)


ann0see:
Don’t merge yet

corrados:
I saw that the installer script always downloads all the needed tools. Maybe it would be better if the script checks if the directory is already existent and if it detects that the directory is present, it skips the download. Do you think this is feasable?

ann0see:
Ok. It seems as if my changes were almost ok. https://stackoverflow.com/questions/28740320/how-do-i-check-if-a-powershell-module-is-installed

> I saw that the installer script always downloads all the needed tools.

That shouldn't be the case. Line 82 in deploy_windows.ps1 should stop the installation if the folder exists.

What you see is the installation of VSSetup and Nuget probably. We might be able to check if they are installed:

https://stackoverflow.com/questions/28740320/how-do-i-check-if-a-powershell-module-is-installed

There's still one bigger TODO: Detection of missing ASIO drivers in the installer and telling the user to install them.

corrados:
> There's still one bigger TODO: Detection of missing ASIO drivers in the installer and telling the user to install them.

I would wait with that feature. Maybe in the near future it is not required anymore. See: https://github.com/corrados/jamulus/pull/821

ann0see:
Actually, It's not that much work (I think we should inform the user but not block him from installing Jamulus (Does the server work without ASIO? In my opinion it must). So right after the Welcome page, this page will be shown if no ASIO driver was found):

WinInstallerASIO

jamulus.io will then have to briefly explain ASIO and tell the user that he probably needs ASIO4ALL. @gilgongo any thoughts on this?

gilgongo:
If the installer can reliably detect the absence of ASIO on the system then it seems like a good idea (and when/if #821 happens it can be retired of course).

My only comments on the screen would be the wording (which would need some work), and also what happens if you hit "Next". I assume the installer quits? Would be good to have a closing message at that point, since I assume some people will just hit "Next" and not read what's on the screen. In fact that's pretty common :-)

ann0see:
> If the installer can reliably detect the absence of ASIO on the system then it seems like a good idea

At the moment it can only detect if a specific key in the registry is present (SOFTWARE/ASIO).

> and also what happens if you hit "Next". I assume the installer quits? 

No. It in my opinion, it shouldn’t. It could nevertheless check if the key now exists and ask if the user is really sure that he wants to continue without an ASIO driver. What if you just want to install the server? We should **not** require the user to have ASIO installed. 

> My only comments on the screen would be the wording 

Sure. You can make improvements to it.

ann0see:
I think we should continue discussing this in #824

Issue #822, opened 2021-01-09
Enabling code scanning for Jamulus source code


corrados:
Github supports code scanning since a while now:

https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/enabling-code-scanning-for-a-repository (see also this https://github.blog/2020-09-30-code-scanning-is-now-available/ and this https://github.com/corrados/jamulus/issues/593).

We should make use of that in the Jamulus project.

corrados:
@ann0see Is this something you could take care of?

ann0see:
It sounds interesting, but I would prefer to focus um the installer, maybe automatic building for windows and Mac and potentially a bit more iOS support first. So I think it would be better if someone else had a look at this.

corrados:
No hurry with that feature :-)

pljones:
OK, it looks like there's no Qt support.  The scanner tries to run a build -- the built in auto-build fails to identify the project as using Qt/qmake/Jamulus.pro and running qmake directly fails.

pljones:
Oh... `runs-on: ubuntu-latest` -- if there's a Qt image, maybe it can use that...

pljones:
It looks like we'd need a [self-hosted runner](https://docs.github.com/en/free-pro-team@latest/actions/hosting-your-own-runners/about-self-hosted-runners) running the appropriate build environment(s).

It's possibly going to take a little while to work out what would be involved -- it's probably _not_ a good idea to run the self-hosted runner on a Jamulus server box to avoid CPU contention.

pljones:
I just realised I did the change on corrados/jamulus, not pljones/jamulus!  Sorry..!

pljones:
> [Self-hosted runner security with public repositories](https://docs.github.com/en/free-pro-team@latest/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories)

That points out a few potential issues with a self-hosted runner, from a security perspective.  Using a single-use, on demand VM would probably alleviate them, I think.  Anyone got recommendations for providers?  (Ideally with free-forever plans, of course...)

corrados:
Is it not possible to run in entirely on Github? Do you need external resources to run it? In this case I actually would not spend time on it. I thought that you simply setup a configuration script with your setting and then a job is run at Github which checks your code. So, it seem I had not made my homework...

ann0see:
Isn't it possible to install qt via apt?

[...]

PR #823, opened 2021-01-09
Create codeql-analysis.yml

pljones:
Per #822 .


pljones:
Default GitHub scanner appears to have no Qt support.  I'll try one of the market place ones.

pljones:
See if there's a ubuntu-latest with Qt...

PR #824, opened 2021-01-09
Win installer: Add ASIO detection

ann0see:
Since #821 is not yet ready, I think it makes sense to detect a missing ASIO driver. 

Current behaviour:

If there's no ASIO key in the registry --> Show a page which links to jamulus.io

If the user exits this page and the key is still missing --> Show a warning where the user can descide if he wants to continue to install jamulus. This is needed since the server doesn't need an ASIO driver installed.

### DIFF
--- a/windows/deploy_windows.ps1
+++ b/windows/deploy_windows.ps1
@@ -66,6 +66,34 @@ Function Get-RedirectedUrl {
     }
 }
 
+function Load-Module ($m) { # see https://stackoverflow.com/a/51692402
+
+    # If module is imported say that and do nothing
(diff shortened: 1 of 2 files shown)


ann0see:
https://github.com/corrados/jamulus/pull/824/commits/f865124dd976a7f0d22ac463b6a01316aa31d148 will also skip the download of nuget and VSSetup if they are not installed. I couldn't test what happens if they are not installed since I only have a Windows 8 VM which doesn't seem to support some commands used in this script. Can anybody confirm that this the installation of these dependencies still works? (I don't see any reasons why it shouldn't)

PR #829, opened 2021-01-10
Activate CodeQL analysis

pljones:
Per #822 .


ann0see:
I‘m quite new to GitHub Actions too and don’t know anything about CodeQL.

Just one comment (I will look through this PR tomorrow and comment on some other things if I find something):

You can also use a matrix all three files look quite similar. This is basically a way to loop through specific steps with a variable filled with the content of the matrix. Think of it like a foreach loop.

See my PR #828 for the automatic build of Jamulus binaries.

.github/workflows/codeql-analysis-linux.yml
```diff
@@ -0,0 +1,48 @@
+name: "CodeQL for Linux"
+
+on:
+  push:
+    branches: [ master ]
```
ann0see, line 5: So we only do scanning on master?

.github/workflows/codeql-analysis-linux.yml
```diff
@@ -0,0 +1,48 @@
...
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '23 11 * * 6'
```
ann0see, line 10: Not sure why you need a cron job here (if this line does what I think it does.)

.github/workflows/codeql-analysis-linux.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  analyze-linux:
+    name: Analyze Linux (Ubuntu 18.04)
+    runs-on: ubuntu-latest
```
ann0see, line 15: You could use a [matrix](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix) to run the steps on every OS here.

.github/workflows/codeql-analysis-linux.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
```
ann0see, line 19: That's a matrix ;-)

.github/workflows/codeql-analysis-linux.yml
```diff
@@ -0,0 +1,48 @@
...
+        # Learn more:
+        # https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#changing-the-languages-that-are-analyzed
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Install Qt
+      run: |
+        sudo apt-get update
```
ann0see, line 32: In my script I did an upgrade too. Not sure if this is needed.

.github/workflows/codeql-analysis-macos.yml
```diff
@@ -0,0 +1,49 @@
...
+        #language: [ 'cpp', 'javascript', 'python' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
+        # Learn more:
+        # https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#changing-the-languages-that-are-analyzed
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Install Qt
```
ann0see, line 30: You could install Qt via brew too: https://jamulus.io/wiki/Compiling#requirements I don't really trust this action since it might break in future. brew shouldn't.

.github/workflows/codeql-analysis-macos.yml
```diff
@@ -0,0 +1,49 @@
...
+      uses: jurplel/install-qt-action@v2
+      with:
+        version: '5.15.2'
+        dir: '${{ github.workspace }}'
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: ${{ matrix.language }}
```
ann0see, line 40: In theory you could just use cpp here, since we only check C++

.github/workflows/codeql-analysis-windows.yml
```diff
@@ -0,0 +1,62 @@
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Install Qt
+      uses: jurplel/install-qt-action@v2
+      with:
+        version: '5.15.2'
+        dir: '${{ github.workspace }}'
+        arch: 'win64_msvc2019_64'
```
ann0see, line 35: Do we only need 64 bit?

ann0see (commented):
Although I don't really have an idea of CodeQl just some comments.

pljones:
> 
> 
> I‘m quite new to GitHub Actions too and don’t know anything about CodeQL.
> 
> Just one comment (I will look through this PR tomorrow and comment on some other things if I find something):
> 
> You can also use a matrix all three files look quite similar. This is basically a way to loop through specific steps with a variable filled with the content of the matrix. Think of it like a foreach loop.
> 
> See my PR #828 for the automatic build of Jamulus binaries.

Yes, I'd seen some of that and wasn't clear how it worked.  The original (auto-generated by Github) had three separate sections, one for each platform.  I split it to make it easier to compare the results.

In fact, there's a lot of overlap with the build process -- the CodeQL part appears to be just two extra steps: `Initialize CodeQL` and `Perform CodeQL Analysis`.

Anyway... here goes - four jobs now, one merged, three separate.  If the merged works, I'll delete the separate ones.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,66 @@
...
+      uses: github/codeql-action/init@v1
+      with:
+        languages: ${{ matrix.language }}
+
+    - name: Build
+      run: |
+        [ -x /Users/runner/work/jamulus/jamulus/Qt/5.15.2/clang_64/bin/qmake ] && \
+          /Users/runner/work/jamulus/jamulus/Qt/5.15.2/clang_64/bin/qmake || qmake
+        make
+      if: matrix.os != 'ubuntu-windows'
```
ann0see, line 55: do you mean windows-latest here?

ann0see (changes requested):

ann0see:
> the CodeQL part appears to be just two extra steps: Initialize CodeQL and Perform CodeQL Analysis.

I think so. But the commands in my workflow are different since I build executable files (therefore I run mac/deploy_mac.sh and in theory windows/deploy_windows.ps1)

pljones:
> 
> 
> > the CodeQL part appears to be just two extra steps: Initialize CodeQL and Perform CodeQL Analysis.
> 
> I think so. But the commands in my workflow are different since I build executable files (therefore I run mac/deploy_mac.sh and in theory windows/deploy_windows.ps1)

This workflow also builds the executables -- it only produces one for each platform, not for each arch for each platform plus headless/non-headless/nosound/non-nosound combinations...

.github/workflows/codeql-analysis-windows.yml
pljones: For the CodeQL?  I don't really know -- I'm assuming it's not _that_ dependent on the OS architecture.

Otherwise we would need to do each OS, on each ARCH, plus perhaps headless and nosound combinations.  (I run my server headless+nosound, so if that affects the CodeQL, I'd want it covered.)

.github/workflows/codeql-analysis-macos.yml
pljones: Yeah, I left the generated code -- it has a matrix on the language and supports other languages, so I left it.

.github/workflows/codeql-analysis-macos.yml
pljones: I've never used `brew`.  What is it?

.github/workflows/codeql-analysis-linux.yml
pljones: No, I had the upgrade and it took ages on every build.  So I took it out and it still worked fine.

.github/workflows/codeql-analysis-linux.yml
pljones: Is done :)

.github/workflows/codeql-analysis-linux.yml
ann0see: Yes. Same issue here. I will remove it from my pr too.

.github/workflows/codeql-analysis-macos.yml
ann0see: It's like apt for macOS: https://brew.sh/

.github/workflows/codeql-analysis-windows.yml
ann0see: I don't know either. So probably that's ok.

pljones:
Nothing other than what the compiler might be throwing out already, so far:
https://github.com/corrados/jamulus/security/code-scanning/20?query=ref%3Arefs%2Fpull%2F829%2Fmerge
> Multiplication result converted to larger type 

src/server.cpp 1100, 1317
src/client.cpp 1142, 1092, 1084
src/buffer.cpp 281, 280, 253, 252
and various libs/opus ones

.github/workflows/codeql-analysis-windows.yml
pljones: In fact, from the results, I'm wondering if it needs to be run multi-platform even.

ann0see: If you run it multi-platform, issues in the sound.cpp and sound.h files can be checked for every OS.

.github/workflows/codeql-analysis-linux.yml
pljones: I guess it's an extra "catch-all" - as this is a security check, it's presumed you do actually want it run and prove it's still going to work, at least once a week.

.github/workflows/codeql-analysis-macos.yml
pljones: I've never used `choco` either...  Makes me wonder if there's an equivalent there.

.github/workflows/codeql-analysis-linux.yml
pljones: Either on Push to master or on Pull Request to master.

I'm not sure what (other people's) forks will do once they update after this gets merged.  The documentation implies each repo needs to enable the scan -- but it doesn't specifically mention forks.

.github/workflows/codeql-analysis-windows.yml
pljones: Possibly - I don't know how it chooses which files to scan.  If it's based on the object files created by the build, then that would do it.

corrados:
Great to see first code analysis results. Is this pull request then ready to be merged?

pljones:
> Great to see first code analysis results. Is this pull request then ready to be merged?

Could go for it.  It's doing the full "desktop"-style builds for all three platforms.  I'm pretty sure `ARCH` isn't important -- and cutting out code with `headless` or `nosound` probably goes against the point..!

corrados:
Ok, thanks.

Is it possible to support such a label which we have from Travis for your new Github code analysis:
grafik
Do you know if something like this is supported?

ann0see:
Yes. That's supported. It should be somewhere in the actions tab

ann0see:
It's this md code: `CodeQL`

pljones:
And this is what the result of a pull request check looks like:
https://github.com/corrados/jamulus/pull/833/checks?check_run_id=1691291583

"No new issues" is okay.  (Better is to fix some ;) )

pljones:
> 
> 
> Ok, thanks.
> 
> Is it possible to support such a label which we have from Travis for your new Github code analysis:
> grafik
> Do you know if something like this is supported?

Did you mean something like this project?
https://github.com/technikhil314/auto-badger

ann0see:
You can click on the three dots in the actions tab and generate a badge. This should be an option. 

I don't think we need something like the auto badge project.

corrados:
I found it, thanks. Since the "CodeQL" action does compile Jamulus for all OSs, doesn't it replace the Travis check which we have implemented right now? If this is the case, I would like to exchange the badge and use Github only tools and remove the Travis .yml file. What do you think?

ann0see:
> doesn't it replace the Travis check

It's slightly different, but it could replace it.

PR #832, opened 2021-01-12
Feature/android

nefarius2001:
fix for bug ui on android
contributes to #83

Bug description:
Dialogs like the Settings- or Chat-Window pop ob in Foreground and fullscreen. After closing them via system's backbutton, they do not show propperly again, but still obstruct the main-window when opened. Closing those dialogs via close() makes them behave normally and reusable (noticed with the Connect-dialog)

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,0 +1,24 @@
+name: Android CI
+
+on:
+  push:
+    branches: [ feature/android ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Make gradlew executable
+        run: chmod +x ./gradlew
+
+      - name: Build with Gradle
+        run: ./gradlew build
+
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -47,6 +47,7 @@ DEFINES += APP_VERSION=\\\"$$VERSION\\\" \
     USE_ALLOCA \
     CUSTOM_MODES \
     _REENTRANT
+DEFINES += "BOOL_CLOSE_DIALOGS_ON_BACKBUTTON=0" # fixes android ui issue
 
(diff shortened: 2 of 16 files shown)


nefarius2001:
solved bug when debugging.
did not solve but when deployed as release via apk

will reopen after further testing

nefarius2001:
working as expected
(I confused branches before)

pljones:
Is there a Gihhub automated build for Android?  (And per supported CPU arch - and, I guess, possibly per version?)

corrados:
Thanks for your code but I intend to implement it a bit different. I'll look at it when I have time (probably upcoming weekend) and create a branch for it. When I am ready it would be good if you could test my code.

corrados:
It was easier than I thought. I just create a new branch **android_back_button_fix** with my version of the fix: https://github.com/corrados/jamulus/compare/android_back_button_fix?expand=1

Can you please checkout that branch and test if it works ok for you on android?

nefarius2001:
>  Hm. This seems too easy but appears to be "the way":

too easy indeed, it fails for me: https://github.com/nefarius2001/jamulus/runs/1697511257?check_suite_focus=true
Thanks a lot for pointing out!

> Can you please checkout that branch and test if it works ok for you on android?

will do.
@corrados can you say why you disabled the escape key? It was disabled for server-selection and chat only, now it would be for all dialogs. Is that intended?

corrados:
> can you say why you disabled the escape key?

@pljones requested and implemented this some time ago. He was used to press the Esc key very often and unexpectably closed the dialog by this. It does not hurt if now other dialogs also do not close on the Esc key.

nefarius2001:
android_back_button_fix  works for me. 
Thanks, and nice solution.

Maybe on mobiles disabling the back-button in main menu would be desirable, to not accidentially close the program (lets do that only if/when desired)

pljones:
> Maybe on mobiles disabling the back-button in main menu would be desirable, to not accidentially close the program (lets do that only if/when desired)

It might be nice not to _exit_ the app - it should still be running and clicking its icon should reopen the main dialogue.  But it _should_ close the "window".  That's pretty much how other apps seem to work on Android, anyway.

PR #833, opened 2021-01-12
Get changelog contents automatically on autobuild

ann0see:
This will automatically get the changes (based from the version in Jamulus.pro) from the ChangeLog file (done by perl script).


ann0see:
qjackctl might need a version number since it doesn't automatically install.

corrados:
Wow, that is cool stuff, thanks.

ann0see:
Please keep in mind that it only creates one "latest" release. This means that once you push a new r* tag, the old release will be deleted.

corrados:
What do you mean by deleted? What will be deleted? The tag in the Git repo? The release on Github?

ann0see:
This: https://github.com/corrados/jamulus/releases/tag/latest
Will be deleted and recreated. The description of the release will be filled with the content of the ChangeLog file fitting to the version which it extracts from Jamulus.pro


https://github.com/corrados/jamulus/releases/tag/r3_6_2 and similar
is not effected via the GitHub action at all (at least not yet)

corrados:
Ok, that makes sense.

> and similar is not effected via the GitHub action at all (at least not yet).

Do you plan to support this, too?

ann0see:
Not yet 100% sure but it should be possible. If you think that this should be possible, I can look into it. We might need to get the tag which triggered the workflow, get the release url and then upload the files to the release. https://stackoverflow.com/questions/58177786/get-the-current-pushed-tag-in-github-actions

ann0see:
I think the easiest way to do this would be to add a second yaml file which also runs if you tag something with beta_

corrados:
These "https://github.com/corrados/jamulus/releases/tag/r3_6_2" are not beta versions but official release versions. Why do you propose a new yml for beta_?

The idea would be that I just create the release tag and everything else (except of the upload to Sourceforge) is done automatically, including building the binaries for each OS, moving the latest tag, creating a Github release and copy the ChangeLog into that (these are all steps which I did manually by hand for each release).

ann0see:
Ok. Let me tell you what is working at the moment: 

1. Building binaries for each OS: macOS and Ubuntu: done and tested. Windows compiles and runs on 32 Bit, 64 Bit doesn't seem to work yet. But the automatic Windows build should build the new installer. Once it's finished (separate issue), we should be able to solve this issue somehow
2. Moving the latest tag: This already works. The release tagged "latest" will now have the content of the changelog. 
3. Creating a GitHub release [based on the tag you set] (this is not working yet). Although the tag does get created, I still have to find a way to upload files and the content of the ChangeLog. Since two releases get published (the latest one and the one based on your tag) it's quite a struggle to get it to work with one file. But I'm still trying to get it done in one file.

ann0see:
I would also like to support "beta" releases which don't add the "latest" tag and create a "latest" release.

Issue #792, opened 2020-12-22
Installer should use public start menu (and desktop) for links 


nico-dr:
if you are logged in as non administrator user (hopefull the most are) and you install Jamulus on a Win 10 PC, you will be asked for an administrator account and the installer will run under this (different) user with the result that the current logged in user will not find the usefull links to start Jamulus.

ann0see:
Same issue here. (Thought of this a while ago). I was also a bit confused that it didn't show up in my start menu. Should be an easy fix.

pljones:
The installer needs a good rewrite.

It should allow the individual user to decide how to install it.

I want to have an "all users" install.  Some people might not want that, though.

I don't want desktop shortcuts.  Some people might want them, though.

I only want the 64bit version installed.  And I want the default install location to be correct (not the 32bit location).

But some people may still need the 32bit version, as well, even when running on 64bit.

I want to specify what the install locations are.

Most of all, I want the installer to remember my choices.

ann0see:
> The installer needs a good rewrite.

Probably yes. 

Nevertheless my draft pr would be a fast fix for that. 

I got really annoyed by this bug and if you don't find Jamulus on your desktop it's a very bad UX. Experienced users can always delete desktop icons, inexperienced users would be annoyed by the current behavior ("No, I can't see Jamulus!!! It's not installed!!!" "Yes it is! Go to Program Files>..." "How do I do that?"...)

corrados:
> The installer needs a good rewrite.

Can someone try out the current implementation on https://github.com/corrados/jamulus/tree/improve_win_ci? I am pretty sure that this solves the issues. Unfortunately, I did not succeed in running it.

ann0see:
I'm now trying it.

You need to set

` Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser`

and there seem to be multiple other issues

ann0see:
Now it does start compiling but throws an error after a few minutes. Nur sure, but there might be something wrong in the Jamulus Source code. 

We could port at least a few changes from this branch. Not 100% sure how to fix the power shell script yet, but it should be possible.

ann0see:
Got it to work (run on external harddisk and admin and my patch).

![jamulusinstaller](https://user-images.githubusercontent.com/20726856/102988652-86503d00-4514-11eb-9460-402df1fe4867.png)



[...]

PR #839, opened 2021-01-14
Create automatic release on tag push with content of ChangeLog (not only latest release)

ann0see:
As discussed in my last PR. 

https://github.com/corrados/jamulus/pull/833#issuecomment-759727238

In a next step we should make this less messy:

My proposal:

Create one script for each OS in e.g. /linux named autorelease_linux.sh which builds installs dependencies, builds Jamulus and moves the content to /deploy

Getting the dependencies (Qt) seems to be difficult. Therefore we can't download them for Windows.

@pljones 

Another question: Why is the /distributions folder not in /linux?

### DIFF
--- a/.github/actions_scripts/get_release_vars.sh
+++ b/.github/actions_scripts/get_release_vars.sh
@@ -2,5 +2,6 @@
 
 VERSION=$(cat ${1}/Jamulus.pro | grep -oP 'VERSION = \K\w[^\s\\]*')
 echo "::set-output name=JAMULUS_VERSION::${VERSION}"
-
+# get the tag which pushed this
+echo "::set-output name=PUSH_TAG::${GITHUB_REF#refs/*/}"
 perl ${1}/.github/actions_scripts/getChangelog.pl ${1}/ChangeLog ${VERSION} > ${1}/autoLatestChangelog.md
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -1,142 +1,142 @@
-name:                         Publish Release
 on:
   push:
     tags:
       - "r*"
   workflow_dispatch:
+name:                               Publish Release
 jobs:
   create_release:
-     name:                    Create release
-     runs-on:                 ubuntu-latest
+     name:                          Create release
+     runs-on:                       ubuntu-latest
      outputs:
-      upload_url:             ${{ steps.create_release.outputs.upload_url }}
-      upload_latest_url:      ${{ steps.create_latest_release.outputs.upload_url }}
-      version:                ${{ steps.jamulus-build-vars.outputs.JAMULUS_VERSION }}
+      upload_url:                   ${{ steps.create_release.outputs.upload_url }}
+      upload_latest_url:            ${{ steps.create_latest_release.outputs.upload_url }}
+      version:                      ${{ steps.jamulus-build-vars.outputs.JAMULUS_VERSION }}
      steps:
-         - name:              Checkout code
-           uses:              actions/checkout@v2
-         - name:              Get Jamulus build info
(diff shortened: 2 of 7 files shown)


ann0see:
Ok. I've now added the autobuild scripts to mac and linux. Windows still depends on the installer.

pljones:
> Getting the dependencies (Qt) seems to be difficult. Therefore we can't download them for Windows.

```
    - name: Install Qt on Windows
      uses: jurplel/install-qt-action@v2
      with:
        version: '5.15.2'
        dir: '${{ github.workspace }}'
        arch: 'win64_msvc2019_64'
        tools: 'tools_qtcreator,4.14.0-0,qt.tools.qtcreator'
      if: matrix.os == 'windows-latest'
```
I've not looked into `choco` -- that might have an install for Qt.  Installing Qt is needed on Linux and MacOS, so it's not really extra on Windows.
```
    - name: Install ASIOSDK
      shell: pwsh
      run: |
        Invoke-WebRequest -Uri "http://www.steinberg.net/asiosdk" -OutFile "asiosdk.zip"
        Expand-Archive -Path "asiosdk.zip" -DestinationPath "ASIOSDK"
        Get-ChildItem -Path "ASIOSDK/asiosdk_2*/*" -Recurse | Move-Item -Destination "windows/ASIOSDK2"
      if: matrix.os == 'windows-latest'
```
ASIOSDK is only used in Windows, so this is only needed there.  I'm not sure how recently Steinberg made the ASIOSDK this easy to get.  But it's unrestricted, basically - no challenge, sign up or anything.
```
    - name: Visual Studio shell
      uses: egor-tensin/vs-shell@v2
      if: matrix.os == 'windows-latest'
```
This one's the one I'm less happy with.  I've no idea why this seems to be necessary -- I'd have expected the `windows-latest` image, which comes with the tools installed, to have the PATH set up right...
```
    - name: Build on Windows
      shell: cmd
      run: |
        D:\a\jamulus\jamulus\Qt\5.15.2\msvc2019_64\bin\qmake.exe Jamulus.pro -spec win32-msvc
        D:\a\jamulus\jamulus\Qt\Tools\QtCreator\bin\jom.exe qmake_all
        D:\a\jamulus\jamulus\Qt\Tools\QtCreator\bin\jom.exe
      if: matrix.os == 'windows-latest'
```
Using jom rather than nmake but it could just as well use nmake.  Why `qmake_all`?  Not sure...  On other platforms, it's just `qmake; make`, subject to knowing where `qmake` is installed.

> Another question: Why is the /distributions folder not in /linux?

I don't know that - I've not been involved in the deployment side of things...

ann0see:
>  jurplel/install-qt-action@v2

This is the action I currently use. But I fear that it might break in future since it seems to find ways to get around the need of creating and logging in to a Qt account.


>  ASIOSDK

The new installer and powershell script will automatically download it. Therefore I don't see a need of downloading it before. 

 >  D:\a\jamulus\jamulus\Qt\5.15.2\msvc2019_64\bin\qmake.exe Jamulus.pro -spec win32-msvc

The new deploy_windows.ps1 script should also run qmake. Your case seems to be different since you don't need to create the installer. Therefore all the stuff in the deploy_windows file (especially everything around the installer) is not needed.

The current problem with Windows is that I can't open Jamulus after it's being installed. It throws an error message #792

pljones:
> This is the action I currently use. But I fear that it might break in future since it seems to find ways to get around the need of creating and logging in to a Qt account.

The last open source version (i.e. that can be freely downloaded) is 5.12.10...  (I was going to write a choco install after finding one with the version hardcoded but it seemed a bit pointless -- I gave up having got this far:
https://github.com/mjscosta/qt-silent-installer/compare/master...pljones:master
)

It could probably be done with `pwsh` just as easily as the ASIO4ALL download.

Issue #785, opened 2020-12-15
Windows SmartScreen: Jamulus Code Signing


ann0see:
Currently, if you install Jamulus on Windows, SmartScreen warns you on Windows 10 (Error: This software might harm your PC). MS Edge even blocks the download.

This evokes a bad impression especially for new users. 

Due to the frequent release cycle there’s no way Jamulus might get on SmartScreen‘s list of known ~~hardware~~ software. 

There are two options to solve this issue:

1. Buy an overly expensive code signing certificate although there‘s something like https://www.certum.eu/en/code-signing-certificates/
2. Make one installer which downloads the latest version of Jamulus and after some time will be downloaded enough times to get on the whitelist. The installer of Jamulus should be able to do online installs, I think. This would also allow us to do a few pre-checks (e.g. is an ASIO Driver installed?)

Related
#550 macOS

pljones:
If this kicks off (point 2, I suppose), then it and #738 should be done together, I suppose.  No point getting it right in one then binning it and getting it wrong in the other...

My Windows scripting is too ancient to pick this up, unfortunately.  (Even for #738 it should be redone in a more modern manner whilst retaining the flexibility.)

gilgongo:
Is there a suggestion that Microsoft may have whitelisted the installer now?

ann0see:
Probably. But only the current version. Once we have a new release it starts again

corrados:
Isn't this very much related to #550? I don't want to have two issue with the same topic. Can we close one of these?

ann0see:
This issue focuses on Windows, while the other focuses on macOS. We might have to handle it differently

gilgongo:
Please continue with any further thoughts here https://github.com/jamulussoftware/jamulus/discussions/1042

Issue #710, opened 2020-11-02
Add ASIO4All setup aid on first setup


ann0see:
Follow up https://github.com/corrados/jamulus/issues/598#issuecomment-697825195 (this idea will not solve the problem but rather mitigate it)

It would be great to add a link to the ASIO4All setup on the new homepage if ASIO4All was detected on the first start of Jamulus/if Jamulus detects that something is wrong to facilitate the setup of ASIO4All.

Furthermore I experienced a hang on "Trying to connect..." if the soundcard was wrongly setup.

(For me the Trying to connect... message is too general at the moment, but that's another issue)

corrados:
> Furthermore I experienced a hang on "Trying to connect..." if the soundcard was wrongly setup.

This is already covered here: https://github.com/corrados/jamulus/issues/710

ann0see:
@corrados I think you linked the wrong issue

ann0see:
My proposal would be to check the ASIO driver name on Windows and show a message next to the ASIO setup Button which links to the wiki. 
I can look into it.

corrados:
You are right. This is the correct link: https://github.com/corrados/jamulus/issues/129

gene96817:
@ann0see  I like your idea. How hard would it be to indicate that there is no audio-in or audio-out configured?

ann0see:
Unfortunately I don't know. How would you check if the wrong input/output was set up? We don't know which the right ones are.

I think the main mixer board should look different if you are not connected to a server.
My proposal looks ugly:
![proposalFirstStart](https://user-images.githubusercontent.com/20726856/98994544-474fc480-2530-11eb-81ed-1f1c59147c5c.png)

But something like that including a link to settings with has a small "play sound" button and a light indicating that the microphone detects something would be great.

I don't think only ASIO4All would profit with that.

gene96817:
Of course, we cannot know what is the right connection for the user. All we can tell the user is the connection is not defined (not connected to a device).

ann0see:
I don't even think that we know that (since ASIO4ALL is automatically selected).

ann0see:
I came up with another (probably bad) idea: We know that Jamulus needs exclusive access to the sound card. What if we start a program which wants to claim the soundcard (mic and output) and look for red crosses/yellow symbols? It could cause even more problems if the program accessing the soundcard doesn't give it back to jamulus but it might also help in short therm. I will look at it

gene96817:
[...]

PR #841, opened 2021-01-14
New windows installer

### LINKED ISSUE #792: Installer should use public start menu (and desktop) for links

ann0see:
Created by @doloopuntil and @ann0see 

There are still some open TODOs:

From #792 

# Blocking
~~1. Replace the graphics: The installer/uninstaller still uses the outdated logo --> Design team @geheimerEichkater @mulyaj @Erioldoesdesign (done)~~

# Non blocking
https://github.com/corrados/jamulus/issues/792#issuecomment-749710366:
> I want to have an "all users" install. Some people might not want that, though.

This is not yet done. 

~~> I don't want desktop shortcuts. Some people might want them, though.~~

This is possible now.

~~> I only want the 64bit version installed. And I want the default install location to be correct (not the 32bit location).~~

This is done now

> But some people may still need the 32bit version, as well, even when running on 64bit.

Would probably need a larger rewrite of the installer.nsi

~~> I want to override what the install locations are.~~

This should be possible now

> Most of all, I want the installer to remember my choices.

To discuss. Is this blocking? Not yet sure how to do it exactly. (we remember if the user wants to install the desktop icon now)

****

Issue with spaces in build path: deploy_windows.ps1 doesn't work correctly if you run it in a path which contains spaces (not urgent); no idea how to solve it properly

~~> Does the updated installer remove previous misplaced Jamulus x64 installations from C:\Program Files (x86)\Jamulus?~~

Yes. Now it does after asking a user if he wants it. 

> https://github.com/corrados/jamulus/issues/792#issuecomment-756425457

There is still a bug which displays that the Jamulus installation is bigger than it in fact is. Probably it's the size of the 32 Bit and 64 Bit installation together

> Detection of ASIO Driver

Works. @gilgongo: We need to add a short description of ASIO on the homepage. Furthermore we need to improve the wording

> Windows installer builds on GH Actions but the produced .exe file doesn't work on Windows 10 64 Bit 

Maybe this issue is gone (probably this is due to a broken VC Redistributable). Edit: no, it isn't see this screenshot:

WinError

> Further language support:

Still todo. But that's out of scope. 

> Travis cli doesn't build on Windows

Is this still needed? @pljones I think your codeql action is enough

#785 

Not sure if can/should download the binaries. In the NSIS docs it says that a download in an installer might trigger anti-virus scanning software which is even worse.

Closes #792 #710


pljones:
> > Most of all, I want the installer to remember my choices.
> 
> To discuss. Is this blocking? Not yet sure how to do it exactly.

It's possible if you write to the registry.  There's a path for installers to store stuff.  See
https://github.com/pljones/Sims4Tools/blob/master/s4pe/mknsis.nsi#L80-L91
It's intended for the uninstaller to use to uninstall the app -- but the the installer can read the values, of course -- see
https://github.com/pljones/Sims4Tools/blob/master/s4pe/mknsis.nsi#L253-L274

windows/installer.nsi
```diff
@@ -1,213 +1,387 @@
...
+      goto skipshortcut
+    skipshortcut:
+    RMDir /r "$SMPROGRAMS\${APP_NAME}"
+
+    ; There may be an auto run entry in the registry for the server, remove it
+    DeleteRegValue HKCU "${AUTORUN_KEY}" "${AUTORUN_NAME}"
+
+    ; Remove the remaining registry keys
+    DeleteRegKey HKLM "${APP_UNINSTALL_KEY}"
+    DeleteRegKey HKLM "${APP_INSTALL_KEY}"
```
ann0see, line 364: @pljones it should remove all the registry content (including the path) here since I run the uninstaller on every new install of Jamulus (in my opinion it's the cleanest way for a update).

We might write the APP_INSTALL_KEY and your choices to a temp file or variable and load it with the installer after the uninstaller is finished. What do you think?

pljones: If you've uninstalled, you're not expected to remember anything.

When I say remember my choices, I mean from one update to the next update.  The installer should read and store the values itself, run the uninstaller (which will clean up the registry), the use the values it previously read as the defaults, where applicable.  If the install completes, those values will be saved for the next run.

ann0see: Sorry, I think my answer was not clear.

Currently on an update of Jamulus, the installer runs the uninstaller of the last version if it finds one. Therefore, also the registry entries will be deleted. My proposal would be something like this:

```
if (oldUninstaller.exists()) {
    oldChoices = registry.getChoicesOfLastInstall();
    oldUninstaller.remove(); // will remove the old Jamulus version including the keys we just got
    registry.setChoicesOfInstall(oldChoices); // but since we saved them, we can rewrite them
}
```

pljones: Yes, something like that.  If the installer reads the registry first, it should still have the values it read, even if it runs the uninstaller.  So "oldUninstaller.remove()" can still run the uninstaller.  Your installer is still running, just waiting for the uninstaller to finish.

corrados:
This pull request contains the nsProcess.dll. This file should not be added to the repo, see this discussion a while ago: https://github.com/corrados/jamulus/issues/341#issuecomment-650569987

windows/installer.nsi
ann0see: Ok. Now the state of "create desktop icon" is remembered #845 Can you please have a look at these changes and comment on the PR if this is enough?

ann0see:
Ok. So the question is if deploy_windows.ps1 should download this file then?

Of course, we can remove it and add it to the gitignore file.

ann0see:
Ok. I've tried to download the nsProcess.dll file from https://nsis.sourceforge.io/NsProcess_plugin but unfortunately the zip file seems to be broken (at least the inbuilt windows extractor thinks so. 7Zip can extract it). I'm not sure, but maybe we can rehost it on a page of ours? An issue was already commented in the nisis forum by me: http://forums.winamp.com/showthread.php?t=230998&page=2

But I'm not sure if anybody really sees and then fixes it.

pljones:
> Ok. I've tried to download the nsProcess.dll file from https://nsis.sourceforge.io/NsProcess_plugin but unfortunately the zip file seems to be broken (at least the inbuilt windows extractor thinks so. 7Zip can extract it). I'm not sure, but maybe we can rehost it on a page of ours? An issue was already commented in the nisis forum by me: http://forums.winamp.com/showthread.php?t=230998&page=2

Can you point me to the commit where you're seeing the problem?

ann0see:
It’s not yet committed. Just try to download the zip file on windows and try to extract it via right click + extract all. It should throw an error

pljones:
image

ann0see:
Yes, 7zip works. But we have to use the integrated windows extraction tool which doesn’t seem to work.

pljones:
image
Ah, yeah - someone's naughtily named a 7z file "zip".

ann0see:
So I guess we can‘t do anything against that. Do you think powershell can unzip 7zip files natively? This is needed since we can’t just install 7zip on the machine of a user who wants to compile jamulus.

pljones:
You could try renaming to 7z and seeing if the extractor can then deal with it...  It may just be believing the extension rather than the data content.

pljones:
Sadly I can't see much in the way of "licence" information regarding the plug in - it's not clear if repackaging and hosting would be allowed.

ann0see:
I doubt that windows supports 7zip natively (I read that it doesn't). You can follow the thread in the forum, I think it's possible to even make source code changes to it. But the license is not clear.

corrados:
Is it possible to download 7z in your script?

ann0see:
Downloading is possible, but not unpacking. So it doesn't make sense. I think we can wait let's say a day or more and wait if there are comments in the forum post. I think we are allowed to host it somewhere else (just have a look at the forum post where someone even changed the code).

ann0see:
Also have a look here:

https://nsis.sourceforge.io/License

> All NSIS source code, plug-ins, documentation, examples, header files and graphics, with the exception of the compression modules and where otherwise noted, are licensed under the zlib/libpng license.

I hope this counts as NSIS plug-in. But it was written by 3 individuals, I think.

pljones:
> I hope this counts as NSIS plug-in.

The page title, "NsProcess plugin", pretty strongly indicates that it's a plug in.  (Subject to random adjustments in hyphenation...)

ann0see:
Ok. I've now uploaded a real zip file to the forum: http://forums.winamp.com/attachment.php?attachmentid=54705&d=1610882327

I don't really like the idea to host it there and would hope they allow us to upload it to jamulus.io or similar.

ann0see:
I've now opened a PR. I hope we can host it on our own.

pljones:
> Ok. I've now uploaded a real zip file to the forum: http://forums.winamp.com/attachment.php?attachmentid=54705&d=1610882327
> 
> I don't really like the idea to host it there and would hope they allow us to upload it to jamulus.io or similar.

Yeah, seems a bit weak to have to have it in a forum post.  Assuming the license is as stated - https://opensource.org/licenses/Zlib - then repackaging without alteration seems permissible.  It's very open about what you can do.

corrados:
Should I merge this branch now?

ann0see:
I think this should work since the most important todos are done. This will also fasciliate the windows autobuild. We can continue working on it of course.

Issue #276, opened 2020-05-24
Create chatlog on the client


Snayler:
I was thinking it could be useful to create a chatlog file on the server (and maybe on the client) so that people can keep a log of what was said in chat.

Sometimes we share links with each other and it would be nice to be able to go back and consult these links at a later time, without fear of closing Jamulus and loosing the chat log. One example use-case, a covers band where people share youtube links to songs for everyone to study for the next rehearsal. Sure one could just copy the chatlog during the session and paste it in some file, but having the server/client do this automatically would be a plus.

corrados:
I would not store any chat on the server. If a log shall be implemented, it should be done on the client side. Because if I am in a session with another musician and I want to write something to him, I do not want others to see my text if any other connects to that same server after we both have left.

Snayler:
Ok, that makes sense from a privacy point of view. Yes, only client side during a session could work.

corrados:
I just changed the title to reflect our discussion.

Snayler:
> I do not want others to see my text if any other connects to that same server after we both have left.

I re-read this and realized something. My initial idea was not to have a "persistent" chat, where messages from an older session could be read when someone connected after. The idea was for the server to keep a text log file.

As for privacy concerns, I also realized that's gone out the window as soon as recording was implemented. We never know if a server is recording, in which case our voice conversations can be recorded. Jamulus as it is right now and on public servers, should not be regarded as a privacy oriented means of communication, from a client point of view.

corrados:
Ok, but for whoom is that text log file for? For the server operator to find out if someone has texted something illegal?

corrados:
> We never know if a server is recording

I already thougt about this issue. We may implement an indicator that recording is enabled. Then every client can clearly see whats going on and can disconnect from that server if he does not want his audio to be recorded.

Snayler:
> Ok, but for whoom is that text log file for? For the server operator to find out if someone has texted something illegal?

Yes, I want to police my own server</joke>. I was thinking more in private server terms. Someone forgets to turn on the chat log during rehearsal, can ask server operator to send him the log. But now that I gave more than 1 second thought on this, the log can be sent from another member connected at the time. And people might want to text stuff not meant for the server operator to know, so privacy outweighs convenience.

> I already thougt about this issue. We may implement an indicator that recording is enabled. Then every client can clearly see whats going on and can disconnect from that server if he does not want his audio to be recorded.

[...]

PR #844, opened 2021-01-15
Support chat history storage in file (#276)

corrados:
This code is about to implement  #276. This functionality is enabled by default, i.e., you do not get an empty chat dialog on startup of the Jamulus software. The only way you can get rid of old messages is to explicitely clear the text with the given menu entry. Actually, that is the intention of a chat history.

I think, I just have to get used to that new functionality. If you are able to compile and test this code, feedback is welcome.

I was thinking of adding a limit of number of lines which can be stored in the history to make sure we do not get memory issues. If someone never clears the chat history, that would be a problem at some point. So this is still missing.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,8 @@
 
 ### 3.6.2git <- NOTE: the release version number will be 3.6.3 ###
 
+- support a chatlog on the client (#276)
+
 - show --clientname as first word in title to avoid clipping in Windows task bar (#789)
 
 
--- a/src/chatdlg.cpp
+++ b/src/chatdlg.cpp
@@ -26,8 +26,10 @@
 
 
 /* Implementation *************************************************************/
-CChatDlg::CChatDlg ( QWidget* parent ) :
-    CBaseDlg ( parent, Qt::Window ) // use Qt::Window to get min/max window buttons
+CChatDlg::CChatDlg ( CClientSettings* pNSetP,
+                     QWidget*         parent ) :
+    CBaseDlg  ( parent, Qt::Window ), // use Qt::Window to get min/max window buttons
+    pSettings ( pNSetP )
 {
     setupUi ( this );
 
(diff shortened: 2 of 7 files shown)


corrados:
I just added a limit for the history file. I set it to 1000 lines which should be sufficient.

pljones:
I think both #276 and #846 asked for a different feature (one that I wouldn't want to have, to be honest).  They wanted it so that when a client joined an active session, all the chat that had happened for that session was sent to the new client.  (To me, this would need a limit of, say, "last 30 minutes" or something).

But this loads up chat from Server A into a session on Server B - which seems to have no meaningful purpose.

ann0see:
> This functionality is enabled by default

Would it be possible to NOT enable this by default? Some people (including myself) would not like the idea that everybody automatically saves my chat messages.

pljones:
> > This functionality is enabled by default
> 
> Would it be possible to NOT enable this by default? Some people (including myself) would not like the idea that everybody automatically saves my chat messages. Just a small checkbox or entry in the chat dialog could be added which lets the user save the content of the chatlog.

The worse problem is that User 1 and User 2 can be having a chat on a server and User 3 is there with logging turned on, capturing User 1's and User 2's chat.  Now, if they've turned off logging, they may well think that means they aren't being logged by anyone.  `CChatDlg::AddChatText` should only capture _your own_ chat, and then only if you've turned on logging.

Like I say, I don't see a lot of use in this.

corrados:
Thanks for your comments.

> They wanted it so that when a client joined an active session, all the chat that had happened for that session was sent to the new client.

This is a no-go for me. When I write  a chat message, I want to write it to the people currently connected to the server and nobody else. Also, storing messages on the server is a data protection issue which we should avoid.

> Would it be possible to NOT enable this by default?

I agree with pljones, it does not make sense to introduce a switch. Either we have this functionality in Jamulus or not.

> Some people (including myself) would not like the idea that everybody automatically saves my chat messages.

In theory they could do it right now. You can simply copy and paste the chat history in an editor and save it.

> CChatDlg::AddChatText should only capture your own chat, and then only if you've turned on logging.

That does not make any sense to me. The chat history is a nice feature if you were in a session and, e.g., someone send a link to a song text. If you would only store your own messages, you will not save this link. This would make the chat history meaningless.

> Like I say, I don't see a lot of use in this.

I see the following usecases: Someone sends his email address or telephone number to you for further communication. If you accidently close Jamulus, the information is gone. Also, as I said above, links to songs or song texts may be useful to have stored in the history. A server welcome message may contain links to video sessions. So you would have that information available without logging in again.

ann0see:
What about adding a "save current chat history to file" button?

corrados:
I this feature turns out to be useless and nobody likes it, I'll close this pull request now. It does not make sense to implement such features in Jamulus, I agree.

ann0see:
The feature to allow an easier way to save the content of the chat log to a file if you click a button would be good. But not automatically please.

PR #849, opened 2021-01-16
qmake: avoid unnecessary runtime dependencies

### LINKED ISSUE #741: New dependency qtconcurrent not needed?
I ran build-debian-package.sh and dpkg-shlibdeps says that there is an unneeded dependency ("package could avoid a useless dependency").

See:

```
dpkg-shlibdeps: Warnung: Das Paket könnte eine nutzlose Abhängigkeit vermeiden, falls debian/jamulus-headless/usr/bin/jamulus-headless nicht gegen libQt5Concurrent.so.5 gelinkt wäre (es verwendet keines der Symbole der Bibliothek).
dpkg-shlibdeps: Warnung: Das Paket könnte eine nutzlose Abhängigkeit vermeiden, falls debian/jamulus/usr/bin/jamulus nicht gegen libGL.so.1 gelinkt wäre (es verwendet keines der Symbole der Bibliothek).
dpkg-shlibdeps: Warnung: Das Paket könnte eine nutzlose Abhängigkeit vermeiden, falls debian/jamulus/usr/bin/jamulus nicht gegen libQt5Concurrent.so.5 gelinkt wäre (es verwendet keines der Symbole der Bibliothek).
```

which translates to something like if jamulus-headless or jamulus wouldn't be linked to libQt5Concurrent.so.5 and jamulus wouldn't be linked to libGL.so.1 and libQt5Concurrent.so.5 we could avoid a debian (headless) dependency. Furthermore it says that the binary doesn't use any symbols of the library. I'm not quite sure what that exactly means. 

server.cpp uses QTConcurrent here: https://github.com/corrados/jamulus/blob/d8a5c6fdc7ec9bceaf6fb8569dce1a2962c64597/src/server.cpp#L833, so I don't understand this warning.

hoffie:
Most g++ versions already seem to use --as-needed in the ld invocations by default. Some versions such as Debian buster's don't. This leads to an unnecessary runtime dependency on libQt5Concurrent.so.

This PR ensures that qmake passes --as-needed in all cases to avoid this.

Fixes #741.
Verified build on: Arch Linux, g++ 10.2.0
Verified build on: Debian buster, g++ 8.3.0

@ann0see Can you verify that this works in your setup/build as well?

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -309,6 +309,12 @@ win32 {
     # we want to compile with C++11
     CONFIG += c++11
 
+    # --as-needed avoids linking the final binary against unnecessary runtime
+    # libs. Most g++ versions already do that by default.
+    # However, Debian buster does not and would link against libQt5Concurrent
+    # unnecessarily without this workaround (#741):
+    QMAKE_LFLAGS += -Wl,--as-needed
+
     HEADERS += linux/sound.h
     SOURCES += linux/sound.cpp
 


ann0see:
I hope this doesn’t break anything. 

I can test it tomorrow.

pljones:
Looks like it didn't break anything in the automated checks.

corrados:
Thank you. If it does break anything, we can remove it on Git master again. So let's merge it.

pljones:
> If it does break anything, we can remove it on Git master again. So let's merge it.

That's meant to be part of the idea of having checks - making the decision on merges easier.  If the checks aren't good enough, we can improve the checks :).  (Yes, and revert the change... ;) )

ann0see:
>  If the checks aren't good enough,

As far as I see, they only check if the code compiles and doesn't have serious security issues which could be found by codeql. But that doesn't mean the code really works.

ann0see:
https://github.com/corrados/jamulus/issues/741#issuecomment-761802611

In case you didn’t read it. The build seems to work.

PR #852, opened 2021-01-17
Remove nsProcess and download it from forum

ann0see:
See #841


corrados:
Thanks. Is the file installer-banner.xcf still used? This seems to be an old one. The ending looks like a Gimp file but geheimerEichkater does not use Gimp. So maybe that one should be removed.

ann0see:
Puh. I don't know. If nobody uses the file anymore, we can remove it, of course

PR #867, opened 2021-01-19
Change ASIO warning

ann0see:
The current warning wasn’t too helpful and might sound too harsh. 
The new one should politely tell the user that he should install an ASIO driver and not evoke the impression that he did something wrong.

Different wording highly appreciated


ann0see:
Strange: CodeQL failed while installing Qt. Travis seems to say this is ok.

PR #830, opened 2021-01-11
mute warnings for depreciated macros

nefarius2001:
fix #827
not breaking any old-qt-version build
supress heaps of distracting, unimportant warnings in new-qt-version builds

PR #870, opened 2021-01-20
Attempt to use Portaudio (v2)

npostavs:
Installer for Windows, as of 2021-06-17: https://github.com/jamulussoftware/jamulus/suites/3016078650/artifacts/68302467

Use the "Jamulus WASAPI" shortcut to test WASAPI mode.

For Linux (e.g., to try ALSA), you currently need to compile yourself with `qmake CONFIG='portaudio portaudio_shared_lib'` and have the relevant Portaudio dev package installed (it's called `portaudio19-dev` on Debian).

Start Jamulus with `Jamulus --api ALSA` to test ALSA.

----

Second round (first attempt was #821).

The current code successfully replaces the ASIO backend (as far as I can tell), except for channel selection. It's not entirely clear to me how the channel stuff works from the Jamulus side. It looks like Jamulus is treating certain channel number specially, e.g. `// special case with 4 input channels: support adding channels` in `windows/sound.cpp` and `GetSelCHAndAddCH` in `src/soundbase.h`?


corrados:
Thanks, looks much better now. As soon as I have time, I'll review and test your code.

npostavs:
> The current code successfully replaces the ASIO backend (as far as I can tell), except for channel selection

I was missing the ASIO control panel, added now.

corrados:
I'll do just some quick comments on your code. A more deep analysis of the code will take place later on.

Jamulus.pro
```diff
@@ -50,12 +57,14 @@ DEFINES += APP_VERSION=\\\"$$VERSION\\\" \
...
     DEFINES -= UNICODE # fixes issue with ASIO SDK (asiolist.cpp is not unicode compatible)
-    DEFINES += NOMINMAX # solves a compiler error in qdatetime.h (Qt5)
-    HEADERS += windows/sound.h
-    SOURCES += windows/sound.cpp \
-        windows/ASIOSDK2/common/asio.cpp \
-        windows/ASIOSDK2/host/asiodrivers.cpp \
-        windows/ASIOSDK2/host/pc/asiolist.cpp
+    !CONFIG(portaudio) {
+        DEFINES += NOMINMAX # solves a compiler error in qdatetime.h (Qt5), breaks portaudio wasapi&wdmks though
+        HEADERS += WINDOWS/SOUND.H
```
corrados, line 62: Please do not use all capital letters for the paths/file names. Windows is not case sensitive but Jamulus runs on many OSs so we should keep all files names/pathes as they are.

Jamulus.pro
```diff
@@ -1048,10 +1103,40 @@ contains(CONFIG, "opus_shared_lib") {
     DISTFILES += $$DISTFILES_OPUS
 }
 
+
+CONFIG(portaudio) {
+    DEFINES += USE_PORTAUDIO
+    HEADERS += src/portaudio-sound.h
```
corrados, line 1109: I do not like to have a "-" in a filename. Please use portaudiosound.h instead

Jamulus.pro
```diff
@@ -1048,10 +1103,40 @@ contains(CONFIG, "opus_shared_lib") {
...
+    SOURCES += src/portaudio-sound.cpp
+    mingw* {
```
corrados, line 1111: I have never used mingw compiler. I usually use the free Visual Studio compiler. Why do you use mingw? Any special reason?

Jamulus.pro
```diff
@@ -1048,10 +1103,40 @@ contains(CONFIG, "opus_shared_lib") {
...
+
 # disable version check if requested
 contains(CONFIG, "disable_version_check") {
     message(The version check is disabled.)
     DEFINES += DISABLE_VERSION_CHECK
 }
 
 ANDROID_ABIS = armeabi-v7a arm64-v8a x86 x86_64
+
+CONFIG += silent
```
corrados, line 1140: What is this for?

Jamulus.pro
```diff
@@ -1048,10 +1103,40 @@ contains(CONFIG, "opus_shared_lib") {
...
+QMAKE_CXXFLAGS += -Wno-deprecated-declarations
```
corrados, line 1141: Hmm... Do we have deprecated warnings in the code? Where do they come from? From the portaudio library or the Jamulus code?

src/client.h
```diff
@@ -44,7 +44,9 @@
 # include "vstsound.h"
 #else
 # if defined ( _WIN32 ) && !defined ( JACK_REPLACES_ASIO )
-#  include "../windows/sound.h"
+//#  include "../windows/sound.h"
```
corrados, line 47: Here you disable the old Windows sound interface?

src/portaudio-sound.cpp
```diff
@@ -0,0 +1,276 @@
+/******************************************************************************\
+ * Copyright (c) 2021
+ *
```
corrados, line 3: The Author is missing and has to be added. You can use your Github name if you like.

src/portaudio-sound.h
```diff
@@ -0,0 +1,81 @@
+/******************************************************************************\
+ * Copyright (c) 2021
+ *
```
corrados, line 3: As above, the author is missing.

.gitmodules
```diff
@@ -22,3 +22,6 @@
 [submodule "tools/jamulus-server-remote"]
 	path = tools/jamulus-server-remote
 	url = https://github.com/vdellamea/jamulus-server-remote
+[submodule "libs/portaudio"]
+	path = libs/portaudio
+	url = https://github.com/npostavs/portaudio.git
```
corrados, line 27: You have referenced your fork. I would like to have to original repo referenced here.

corrados: The repo commit you have used is 3988fd. Is this an official release version of the portaudio library?

Jamulus.pro
npostavs: Mostly habit I guess. I didn't have Visual Studio installed.

Jamulus.pro
npostavs: It changes the output of `make` from 

    g++ -I...etc etc -D...etc etc <enough args to wrap around several lines>...

to

    compiling ../src/main.cpp

Jamulus.pro
npostavs: I was getting a bunch of deprecation warnings from Qt, I guess you might be compiling with an older Qt version? I have the this and the `CONFIG += silent` thing in a separate commit so it's easy to drop and/or add it separately. Mainly these are there to keep myself sane.

    ../src/util.h:1303:109: warning: 'QTextStream& QTextStreamFunctions::endl(QTextStream&)' is deprecated: Use Qt::endl [-Wdeprecated-declarations]
     1303 |                             streamFile << i << " " << static_cast<double> ( vElapsedTimes[i] ) / 1000000 << endl;
          |                                                                                                             ^~~~
    In file included from C:/msys2/mingw64/include/QtCore/qdebug.h:49,
                     from C:/msys2/mingw64/include/QtNetwork/qabstractsocket.h:47,
                     from C:/msys2/mingw64/include/QtNetwork/qhostaddress.h:48,
                     from C:/msys2/mingw64/include/QtNetwork/QHostAddress:1,
                     from ../src/client.h:27,
                     from ../src/client.cpp:25:
    C:/msys2/mingw64/include/QtCore/qtextstream.h:293:75: note: declared here
      293 | Q_CORE_EXPORT QT_DEPRECATED_VERSION_X(5, 15, "Use Qt::endl") QTextStream &endl(QTextStream &s);
          |                                                                           ^~~~

src/client.h
npostavs: Yeah, I should have ifdef'd it though.

.gitmodules
npostavs: > The repo commit you have used is 3988fd. Is this an official release version of the portaudio library?

No, I added a couple of commits to last stable release (hence why I used my fork instead of upstream):
1. [Backport to let the wasapi code build](https://github.com/PortAudio/portaudio/commit/d7e79160a3670dc89ad35417ea511d9196964927)
2. [Modification to handle ASIO buffer size changes](https://github.com/PortAudio/portaudio/commit/3988fd9ebc10cd7b1834ec6e0c0c01c9f6dbf27a)

Jamulus.pro
npostavs: right, not sure how I ended up upcased filenames

Jamulus.pro
hoffie: Sounds like your branch could use a rebase against master? :) The latter includes #830 which probably fixes the same issue (although with a slightly different approach).

(Glad to see the work on Portaudio, btw! :))

npostavs: > The latter includes #830 which probably fixes the same issue (although with a slightly different approach).

Ah, yes. Looks like a better approach too since it's targeted only at Qt.

pljones:
Note that the CodeQL checks are failing with
> Project ERROR: portaudio only tested on mingw for now

.gitmodules
corrados: Can you please create pull requests for your changes in the official repo?

Jamulus.pro
corrados: I would like to stick to Visual Studio for now if possible. Is there a chance that you install the free Visual Studio or is this no option for you?

npostavs:
Okay, I seem to have it building with msvc now. By the way, I can't figure out how to build the debug version from QtCreator:

```
18:36:05: Starting: "C:\Qt\Tools\QtCreator\bin\jom.exe" 
	C:\Qt\5.15.2\msvc2015_64\bin\qmake.exe -o Makefile ..\Jamulus.pro -spec win32-msvc "CONFIG+=debug" "CONFIG+=qml_debug"
	C:\Qt\Tools\QtCreator\bin\jom.exe -f Makefile.Release
```
Even though the debug configuration is selected as evidenced by the `CONFIG+=debug`, it insists on using `-f Makeilfe.Release`. At least with mingw I can use `make debug` from the command line.

Jamulus.pro
pljones: Why?

npostavs: It's a personal preference mainly; I find it helps me pick out important error and warnings messages from the output. I don't have a problem dropping it before merging, but it's inconvenient to omit it from the PR in the meantime, due to the way Github works .

src/portaudiosound.cpp
```diff
@@ -0,0 +1,279 @@
...
+    }
+    else
+    {
+        iPrefMonoBufferSize = iNewPrefMonoBufferSize;
+    }
+
+    vecsAudioData.Init ( iPrefMonoBufferSize * 2 );
+    if ( deviceStream && deviceIndex >= 0 )
+    {
+        LoadAndInitializeDriver ( deviceIndex );
```
corrados, line 142: Calling this function is not supposed to be done in the Init function. LoadAndInitializeDriver should only be called if a new sound card device is selected. The Init function is called very often on any configuration change of the current sound card (like mono/stereo, buffer sizes, etc.).

src/portaudiosound.cpp
```diff
@@ -0,0 +1,279 @@
...
+        strDriverNames[i] = devInfo->name;
+    }
+}
+
+int CSound::Init ( const int iNewPrefMonoBufferSize )
+{
+    if ( deviceIndex >= 0 )
+    {
+        long minBufferSize, maxBufferSize, prefBufferSize, granularity;
+        PaAsio_GetAvailableBufferSizes ( deviceIndex, &minBufferSize, &maxBufferSize, &prefBufferSize, &granularity );
```
corrados, line 103: Interesting. This looks very similar to the native ASIO interface function.

src/portaudiosound.cpp
npostavs: I guess I could rename `LoadAndInitializeDriver(int)` (as `ReloadCurrentDriver` perhaps?) to make things clearer, but since the portaudio stream needs to be re-opened to apply a new buffer size, that call is necessary.

pljones:
> Even though the debug configuration is selected as evidenced by the `CONFIG+=debug`, it insists on using `-f Makefile.Release`. At least with mingw I can use `make debug` from the command line.

Hm.  Not only that but the created EXEs are the same size, which indicate it is, actually, the same build.  Manually overriding to Makefile.Debug gives me
```
16:46:51: Starting: "D:\Qt\Tools\QtCreator\bin\jom.exe" -f Makefile.Debug
Error: dependent 'D:\Qt\5.12.2\msvc2017_64-static\lib\Qt5Widgetsd.lib' does not exist.
16:46:51: The process "D:\Qt\Tools\QtCreator\bin\jom.exe" exited with code 2.
```
I'm guessing QtCreator has seen that the debug libraries were not installed and decided not to try to use them by default - so forcing it fails.

.gitmodules
npostavs: The first is already from portaudio's master branch. I've just made a pull request for the second one: PortAudio/portaudio#473

npostavs:
Okay, I've rebased and have it working with the codeql thing.

ann0see:
Great! Not sure how we handle this, but I'm really happy if this works one day!

ann0see:
@npostavs is it ready to be tested/does it work somehow?

npostavs:
> @npostavs is it ready to be tested/does it work somehow?

It works for me locally (ASIO devices only right now). Looks like I have to rebase again though.

pljones:
1. Needs the conflict resolved
2. It's still draft

npostavs:
> * Needs the conflict resolved

Done :)
 
> * It's still draft

The main reason I still have it as draft is due to the lack of channel support: it's currently losing features. Apart from that, it is working.

npostavs:
At least the standard channel selection seems to be working now. There's still the 4-channel mixing thing to do, although I'm not sure how necessary that is; it looks like the linux backend doesn't support it anyway...

.gitmodules
pljones: I can see you've not had much progress over on the PortAudio PR.  Had any luck finding out about the thread safety issues?

npostavs: > Had any luck finding out about the thread safety issues?

No. I guess the fact that Jamulus' current Windows code has `ASIOMutex` is suggestive, although it's not used for `CSound::LoadAndInitializeDriver`(!?)

pljones: OK, quick review of the threading model from memory - I've not looked at it closely on the client:
- the client main thread starts (single thread)
- if there's an audio driver, the audio thread should start (single thread)
- the network thread starts (single thread)
- if there's a GUI, the GUI thread should start (single thread)

I _think_ that's it.  I may be mistaken about the separate audio thread - it could be the main thread (which would otherwise only be co-ordinating network and audio and responding to and informing the GUI).

Where would multi-threading cause contention in audio driver processing in the client?  I'd guess it can't before the driver is running, at least.

ann0see:
I think this should not be part of the next release and probably needs some beta testing before we can merge it?

pljones:
Also it's now got conflicts.  I'd have to agree it's missed the 3.7.0 cut now.

nefarius2001:
I can look into the conflicts, bit not sure how to contribute them to this. Maybe a  pull request to the author?

npostavs:
Oh, I think the conflicts are mostly trivial anyway, I can fix them up.

Jamulus.pro
pljones: I've no problem with it being used in your personal builds but I like to see exactly what Github is building, so I'd rather this weren't done.

npostavs: Right, I've meanwhile realized that I can pass `CONFIG += silent` on the `qmake` command line, so this isn't actually needed in the .pro file anyway.

.gitmodules
npostavs: So if the device's buffer size gets changed, it sends a `kAsioResetRequest` message in the audio thread. Jamulus' handler calls `SoundBase::EmitReinitRequestSignal` which does `emit ReinitRequest`. That's connected to the `CClient::OnSndCrdReinitRequest` slot, which would run in the GUI thread?

.github/workflows/autobuild.yml
```diff
@@ -48,6 +48,8 @@ jobs:
          # Checkout code
          - name:                    Checkout code
            uses:                    actions/checkout@v2
+         - name: Checkout portaudio submodule
+           run: git submodule update --init -- libs/portaudio
```
nefarius2001, line 52: the first action does not need submodules, please remove these lines

(in the second action its needed, and included by the "with:   submodules:   true" of the checkout

nefarius2001 (changes requested):
I just checked the yml-changes

ann0see:
I'm really looking forward to having this integrated in one of the next versions on Jamulus. How is latency with PortAudio compared to the native API?

ann0see:
Sorry: I've stopped the CI to get the windows artifacts since I'd like to test it. It should be running now.

nefarius2001:
@ann0see did you succeed in downloading the artifact? I cannot open the zip from https://github.com/jamulussoftware/jamulus/actions/runs/603986607. Windows 10 says "Operation can not complete because the file contains a virus or potentionally unwanted software"

ann0see:
Yes, I could download and install it. Not sure what this warning you got means? Is it SmartScreen?

nefarius2001:
lets discuss that in #1129

ann0see:
Just tried to set a device with a wrong sample rate in ASIO4ALL. Jamulus doesn't complain if I set it during run time. Once I close Jamulus, ASIO4ALL can't be opened.

npostavs:
> Just tried to set a device with a wrong sample rate in ASIO4ALL. Jamulus doesn't complain if I set it during run time.

I don't have such a device handy so I can't reproduce this, but I had tried to handle buffer size changes only so it makes sense that things wouldn't work right. I've just pushed a new version that should fully reinitialize the library on any config change, I hope that will do the trick.

ann0see:
Ok. Some more feedback:

1. If there is no ASIO driver, only an error message "No ASIO devices found" is shown. It would be great if the message was the same as it used to be (The one which explained that you'll need one like ASIO4ALL,...). It should be somewhere in the old code.
2. If I misconfigure ASIO4ALL the message asking if I want to open the driver setup shows up, but it doesn't open ASIO4ALL if I click on "Yes".
3. For me it seems that connecting to a server takes more time now. 
NoPortaudio
4. In fact, the latency feels a bit higher. But that could also be due to ASIO4ALL. 
5. If I change the Buffer Size in ASIO4ALL without changing it in Jamulus, the audio stutters (especially on high buffer size settings in ASIO4ALL and low settings in Jamulus)

npostavs:
>     5\. If I change the Buffer Size in ASIO4ALL without changing it in Jamulus, the audio stutters (especially on high buffer size settings in ASIO4ALL and low settings in Jamulus)

Does the ASIO4ALL buffer size slide do anything? I had never noticed any effect from moving it.

ann0see:
> Does the ASIO4ALL buffer size slide do anything?

With the old code it doesn't do anything. But with your new code it seems to do something.

pljones:
Jamulus ASIO code was telling the driver "I want you to do 48KHz and (user selected buffer size)".  The ASIO driver then went into slave mode - "I've been opened by an application that's telling me what to do, I just do it".

Does the new code work the same way?  Directly passing through the request from Jamulus to the underlying driver?  If not, that will change behaviour significantly.

The Jamulus "Driver Setup" button should open the PortAudio settings.  PortAudio should allow selection of and configuration of the driver _it's_ using from there, I'd have thought.

Otherwise, what benefit is there adding a code layer?  It's just more CPU cycles for no functionality gain.

npostavs:
> Does the new code work the same way? Directly passing through the request from Jamulus to the underlying driver?

I thought it should be, according to the code. But what I'm seeing now with ASIO4ALL is that the selected buffer size ends up larger (more than 512 samples) instead of the selected one, and I'm not sure why.

One odd thing I noticed when looking at the Jamulus' current ASIO code is that it doesn't initialize `driverInfo` before passing it to `ASIOInit` which seems bad? (I noticed this because one of the things that need to be set is a handle to the application's main window, which is not straightforward to retrieve in Jamulus).

https://github.com/jamulussoftware/jamulus/blob/r3_6_2/windows/sound.cpp#L65

> The Jamulus "Driver Setup" button should open the PortAudio settings. PortAudio should allow selection of and configuration of the driver it's using from there, I'd have thought.

No, there isn't any such thing as "PortAudio settings" in the way that ASIO has them.

> Otherwise, what benefit is there adding a code layer?

The main point is add support for non-ASIO APIs such as WASAPI or WDM. This means that the extra layer of ASIO4ALL would no longer be needed. I view the current ASIO-only implementation as just a stepping stone. I was hoping to get equivalent functionality though.

pljones:
So if I'm not using ASIO4ALL and have a direct hardware ASIO driver, I don't get blocked by this?  I can switch from it to a standard ASIO driver?

ann0see:
>  I can switch from it to a standard ASIO driver?

As far as I know, yes.

Before this gets merged and implemented, some other tests will need to be made (especially latency measurements and wrong user configurations with ASIO4ALL).

npostavs:
> So if I'm not using ASIO4ALL and have a direct hardware ASIO driver, I don't get blocked by this? I can switch from it to a standard ASIO driver?

Previously, I tested this code using a device that comes with its own ASIO driver; it seemed to work fine. Testing with with ASIO4ALL it looks like the problem is that PortAudio ends up choosing a larger buffer size when I pass `deviceInfo->defaultLowInputLatency` as the `suggestedLatency` of the input parameters. I've just pushed an update that passes `0` instead.

ann0see:
Great! Thank you and sorry for all of the problems I report which are mainly related to ASIO4ALL.

The main problems with ASIO4ALL I see is that we can't be sure about anything:
It can change the sample rate during runtime, numbers/names of in/outputs change, buffer sizes are suddenly higher. An input which is somehow blocked by another application (red cross) is selected which leads to no sound... Can you think of anything else?

Therefore I suggest to mis-configure as much as possible with this driver (which seems to happen in a real world scenario). That's how I found some bugs in the old code.

npostavs:
> An input which is somehow blocked by another application (red cross) is selected which leads to no sound...

I'm somehow not even able to reproduce this anymore, though I'm sure I ran into it in the past. Could a Windows update have changed this? Somehow Jamulus manages to get priority and grab the microphone from other applications now!? (This is with Jamulus 3.6.2, so it's definitely nothing that I've done).

ann0see:
Unfortunately I doubt this. But who knows?

pljones:
> Previously, I tested this code using a device that comes with its own ASIO driver; it seemed to work fine. Testing with with ASIO4ALL it looks like the problem is that PortAudio ends up choosing a larger buffer size when I pass deviceInfo->defaultLowInputLatency as the suggestedLatency of the input parameters. I've just pushed an update that passes 0 instead.

What I mean is that I don't have to use this new layer at all.  If it's adding an extra layer of code between Jamulus and the driver, and it's been seen to change the behaviour, then it should be optional.  Currently in Jamulus, I get presented with a list of all my registered ASIO drivers.  To ask a simple yes/no:

Does this get listed in the list with my other ASIO drivers?
image

npostavs:
> Does this get listed in the list with my other ASIO drivers?

No, it's not an alternate driver. It's an alternative to the layer of code in Jamulus that connects to the ASIO driver (i.e., `windows/sound.cpp`) . Ideally there would be no change in behaviour and `windows/sound.cpp` could be completely dropped.

pljones:
OK, so back to the original question: what problem is this solving?

At the moment, I click the Device driver button and can select between ASIO device drivers.  I need this functionality.

Will putting PortAudio in between Jamulus and all my ASIO devices make them work better?  Worse?  Differently?

Why do it?

I originally thought this was to provide a replacement for ASIO4ALL - a layer over WDM presented to Jamulus as an ASIO interface but it seems not.  So I don't know what the benefit is intended to be.

npostavs:
> I originally thought this was to provide a replacement for ASIO4ALL - a layer over WDM presented to Jamulus as an ASIO interface but it seems not. So I don't know what the benefit is intended to be.

It is ultimately intended to allow Jamulus to layer over WDM/WASAPI (so that ASIO4ALL would no longer be needed). However, I've implemented ASIO right now. See also https://github.com/jamulussoftware/jamulus/pull/821#issuecomment-757110225

pljones:
OK, I'd rather see it done as a separate option - _ideally_ at run time but I don't think the code lends itself to that, so it probably needs to be a built time option.

npostavs:
It is a build time option, see first hunk in `Jamulus.pro`.

pljones:
Somehow I'd missed the `Jamulus.pro` changes entirely...  OK, initially the default should stay "as is".  That way, the code changes can be put in place and the build won't need to change to keep behaviour as is whilst this is getting more testing.  Then the code can get to master sooner and people can choose to build it if they want to test it out.  I think it's going to need extensive testing.

pljones:
> Okay, I can't figure out how to add more configs to the autobuild though.

Don't worry - I can't figure out what it's doing now, either.

nefarius2001:
>>   Okay, I can't figure out how to add more configs to the autobuild though.
> Don't worry - I can't figure out what it's doing now, either.

It basically calls the windows/deploy_windows.ps1
Let me look into this, should not be hard - but take a bit since iterations on autobuild are slow

I will give npostavs a pull request, once the autobuild-action is tested. Feel free to peek here: https://github.com/nefarius2001/jamulus/tree/autobuild/portaudio.v2

nefarius2001:
Beginners questions:
1) how can I check if Build was done with the portaudio? In Jamulus itself, I mean.
2) what do i need to select to use the portaudio ? Generic, or asio4all? I think I am given the same choices in the settings as in normal build

npostavs:
> 1. how can I check if Build was done with the portaudio? In Jamulus itself, I mean.

Ah, there wasn't a way to do that; I've just added PortAudio to Jamulus' "Libraries" tab of the "About" dialog.

> 2\. what do i need to select to use the portaudio ? Generic, or asio4all? I

Currently you wouldn't do anything differently; if Jamulus is compiled with PortAudio it will use PortAudio. I'm not sure what you mean by "Generic", but any ASIO device you select will be accessed using PortAudio routines.

npostavs:
I've added initial support for WDM-KS and WASAPI. When I tried using my internal sound-card with WASAPI, it had a bit more reported delay than ASIO4ALL (~50ms vs ~40ms) though the sound was louder and clearer with WASAPI. The slowness is expected since I don't yet pass the exclusive flag in the WASAPI version (it's possible the improved sound quality gets lost in exclusive mode too, since it's supposed to bypass the OS software mixer machinery).

When I tried with WDM-KS (which always runs in exclusive mode), it failed with some inscrutable error message :(.

ann0see:
Wow, thank you. WASAPI support would be a great step forward!

nefarius2001:
@npostavs do I need to install Portaudio separately as a basis, or is it "compiled into" Jamulus in that flavor?

npostavs:
> do I need to install Portaudio separately as a basis, or is it "compiled into" Jamulus in that flavor?

The latter; the source is pulled in via submodule.

ann0see:
Running the latest code, from the artifact removed the ASIO-Device-Setup button for me. Will try WSAPI soon.

npostavs:
> Running the latest code, from the artifact removed the ASIO-Device-Setup button for me. Will try WSAPI soon.

I think the artifact build isn't using PortAudio anymore since it's not the default (check Help>About>Libraries); which yeah, I made the button conditional on using ASIO and forgot about the `windows/sound.cpp` code which always uses ASIO.

ann0see:
Ah. Ok. I just found out that it didn't support --api. Therefore I'm compiling it now with portaudio support.

ann0see:
Sorry for the late reply here. I'll probably fork your branch, add PortAudio support to the .pro file and then use the artifact. 

Not sure what's still missing here except for my comment from above (especially concerning the "No ASIO driver found" warning.

In general, I think the whole Audio API (also parts not concerning PortAudio) must be re thought. 

Especially it shouldn't close Jamulus if you choose a 44.1 kHz only device in Jamulus, shouldn't freeze (which seems to happen some times with ASIO4ALL and the native API, not PortAudio) and should have a reload button in settings. I know that's out of scope here but maybe also related to the initial problem: 

First setup of unpredictable audio hardware which is be different for everybody. 

Since you now have experience of the audio API, maybe you can look I to a few problems with ASIO4ALL and the current implementation?

npostavs:
> In general, I think the whole Audio API (also parts not concerning PortAudio) must be re thought.
> 
> Especially it shouldn't close Jamulus if you choose a 44.1 kHz only device

Yes, Jamulus' current insistence on having a correctly setup sound device before it agrees to run makes for an inconvenient interface.

> Since you now have experience of the audio API, maybe you can look I to a few problems with ASIO4ALL and the current implementation?

I can't reproduce problems with ASIO4ALL itself, but I seem to have reproduced a crash which I think corresponds to #872 using voicemeeter set to 44.1kHz. I'll comment there soon.

ann0see:
Ok Thanks for having a look at #872 I'm still struggeling to find the cause of it.

ann0see:
Finally compiled it! WSAPI works and is a huge step forward! Yes, latency is high ... ASIO works and WDM-KS gives a huge error message. "Could not open Portaudio stream: Unanticipated host error, WdmSyncloctl: DeviceloControl".

npostavs:
> WDM-KS gives a huge error message.

I think I've fixed this, although I'm not getting good sound when using WDM. I've also added support for WASAPI exclusive mode which seems to be giving similar results as using ASIO4ALL.

ann0see:
> I've also added support for WASAPI exclusive mode which seems to be giving similar results as using ASIO4ALL.

Great.

ann0see:
Ok. Just tried it and latency in exclusive mode is much better. Could you maybe make exclusive mode default and add (high latency) behind shared mode?

WDM-KS seems to work with my interface now (although I hear x-runs also with higher buffer sizes). But thank you very much for doing all this!!

pljones:
> I think I've fixed this, although I'm not getting good sound when using WDM. I've also added support for WASAPI exclusive mode which seems to be giving similar results as using ASIO4ALL.

You will be thanked by generations of Jamulus users!!

ann0see:
I think we should publish an (un) official PortAudio build to get to know issues.

dcorson-ticino-com:
How is this used and PortAudio configured ?
I looked at both Win10 artifacts and see only the ASIO drivers I have installed,
no mention of WDM or WASAPI.
There is a difference between the two versions, though.  The Win portAudio version doesn't show the Jack server I have installed.
I would be grateful for more info how to get this up and running.

npostavs:
> I looked at both Win10 artifacts and see only the ASIO drivers I have installed,
> no mention of WDM or WASAPI.

By default, PortAudio will use only ASIO devices. You can choose different APIs using the `--api` command line option, e.g., `--api wasapi`.

> The Win portAudio version doesn't show the Jack server I have installed.

Hmm, not sure how the current Windows version could be doing that. Unless you are building with the `jackonwindows` config option?

dcorson-ticino-com:
Thanks for your quick reply Noam.  I will try it out using the options, but I think this will need to be a part of the GUI when released.
You can not put off Win10 users faster than by saying "command line". :)

dcorson-ticino-com:
Because I think this is an important development I did some testing today to see where we are.

My baseline is using my NI KompleteAudio2 with its native ASIO driver (Win10).  The audio is good and I have an overall delay of 10-11ms with 64 sample buffer and small buffers enabled.  All testing on my home network, ping = 0.

Using WDM-KS with the NI interface I got 13-14ms with 64 sample buffer and small buffers enabled, good audio.
Using WASAPI with the NI interface in shared mode my best was ~45ms with 64 sample buffer and small buffers disabled, Audio OK.
Using WASAPI with the NI interface in exclusive mode my best was ~35ms with 64 sample buffer and small buffers enabled, but the audio is distorted and full of static.  No change of buffer size, or jitter buffer size made any difference for the audio quality.  Also something appears to be happening that the auto jitter buffer doesn't do any adjustments at all.

A problem is that with WDM and WASAPI I was not able to use any of the local interfaces of my laptop.  I tried all the possibilities and every time got a response something like the following.
PAError

I don't know if this is not yet implemented, but for me getting rid of ASIO4ALL for those using the internal audio devices is the most important reason for using PortAudio.  I have never been able to get ASIO4ALL to work.

If there is any way I can help with this development please let me know.

pljones:
> > The Win portAudio version doesn't show the Jack server I have installed.
> 
> Hmm, not sure how the current Windows version could be doing that. Unless you are building with the `jackonwindows` config option?

@dcorson-ticino-com, whilst PortAudio _can_ support JACK on Windows, @npostavs hasn't built PortAudio as a library, including the JACK libraries, and then linked Jamulus to it.  He's integrated parts of the PortAudio codebase into Jamulus - but not the JACK support.  I have to say, having the JACK support would *hugely* improve Jamulus's flexibility on Windows, so I'm with you on having that, too, if possible.

dcorson-ticino-com:
@pljones I think I wasn't clear.  
Jamulus 3.7.0 sees the JackServer as an ASIO device.  
The PortAudio Jamulus doesn't see it at all.

ann0see:
Can't you compile Jamulus with "jackonwindows" to add JACK support?

blaisegassend:
I would be very interested in trying out the new portaudio version with my chorus (I'd rather have a bit more latency, than somebody who is computer challenged unable to figure out why the microphone can't open in exclusive mode). Is there a place to get an unofficial binary? I'm not set up to compile on Windows.

I'm also curious how much extra latency non-exclusive access is piving with portaudio.

dcorson-ticino-com:
> Can't you compile Jamulus with "jackonwindows" to add JACK support?

My comment is that:
1) using old Jamulus versions on Win10 without any special compile options the Jack server manifests itself as an ASIO interface.  This has been working without problems since many Jamulus versions.  
2) The PortAudio Jamulus version does not see this ASIO interface.

This is, however, a tiny problem compared to the non-functioning internal PC audio hardware.

ann0see:
> have a bit more latency

Unfortunately it's more than just a bit. At least for me WASAPI in shared mode is close to unusable. However exclusive mode works. 

You can get the binary if you click on the checks tab here and then on assets. Afterwards download the PortAudio asset and start Jamulus with a CLI argument like `--api WASAPI` or `--api WDM-KS`

dcorson-ticino-com:
This is the complete list of errors I get when I start Jamulus PortAudio without an ASIO device plugged in.
PortAudioErrors

npostavs:
> Ok. Just tried it and latency in exclusive mode is much better. Could you maybe make exclusive mode default and add (high latency) behind shared mode?

I think this might leave Jamulus completely unusable in some cases, until #1338/#1378 is resolved. (That is, if devices can't be used in exclusive mode, Jamulus would quit before the user can change to shared mode)

> My baseline is using my NI KompleteAudio2 with its native ASIO driver (Win10).

Might be interesting to compare what happens when you access it through ASIO4ALL.

> A problem is that with WDM and WASAPI I was not able to use any of the local interfaces of my laptop.

Yeah, various devices seem to give some inscrutable errors. It's possible playing with PortAudio's `suggestedLatency` parameter could help (at least, I was able to get WDM-KS working on my machine after changing it; though the sound isn't great).

> Can't you compile Jamulus with "jackonwindows" to add JACK support?

I believe that's mutually exclusive with the current ASIO and also this PortAudio Jamulus `CSound` implementations.

> I would be very interested in trying out the new portaudio version with my chorus (I'd rather have a bit more latency, than somebody who is computer challenged unable to figure out why the microphone can't open in exclusive mode). Is there a place to get an unofficial binary? I'm not set up to compile on Windows.

There is an installer built by the CI, but this PR isn't ready for computer challenged people to use yet (since it requires command line options to enable the other APIs). Perhaps adding an extra "Jamulus WASAPI" shortcut would be enough for testing?

> This is the complete list of errors I get when I start Jamulus PortAudio with command line option --api WDM-KS with no ASIO device plugged in.

I know what to do about the "Less than 2 channels not supported" thing, the others I'm less sure about (possibly fiddling with `suggestedLatency` as I mentioned above is the answer).

pljones:
> @pljones I think I wasn't clear.
> Jamulus 3.7.0 sees the JackServer as an ASIO device.
> The PortAudio Jamulus doesn't see it at all.

OK, you're not talking about JACK support.  You're talking about native ASIO support, with JACK supplying an ASIO driver (JackRouter).

dcorson-ticino-com:
> OK, you're not talking about JACK support. You're talking about native ASIO support, with JACK supplying an ASIO driver (JackRouter).

Exactly

pljones:
And yes, I'd expect that still to work if PortAudio support for ASIO is to be considered "working".

npostavs:
> There is an installer built by the CI, but this PR isn't ready for computer challenged people to use yet (since it requires command line options to enable the other APIs). Perhaps adding an extra "Jamulus WASAPI" shortcut would be enough for testing?

> I know what to do about the "Less than 2 channels not supported" thing,

I've added WDM and WASAPI shortcuts & mono devices should work now.

Installer: https://github.com/jamulussoftware/jamulus/suites/2494709738/artifacts/53853352

---

It turns out when I install the JACK ASIO router thing, Jamulus hangs on startup when I run it under msys2, meaning I can't debug it. I'm going to leave this on the backburner for now.

Jamulus.pro
```diff
@@ -25,6 +25,16 @@ CONFIG += qt \
     lrelease \
     release
 
+CONFIG (noportaudio) {
+    CONFIG -= portaudio
+}
+
+CONFIG (portaudio) {
+    message("driver WINDOWS portaudio")
+}else{
```
hoffie, line 34:
```suggestion
} else {
```

Jamulus.pro
```diff
@@ -25,6 +25,16 @@ CONFIG += qt \
     lrelease \
     release
 
+CONFIG (noportaudio) {
+    CONFIG -= portaudio
+}
```
hoffie, line 30: Any specific reason for this? I don't think we have `no*` CONFIG values for everything else either, probably just in those places where we need to override defaults.

hoffie (commented):
Sadly, I haven't been able to follow this PR closely. It is currently tagged for 3.8.0 which we plan to feature-freeze tomorrow.

Is there a benefit of merging this to make it easier for people to try this mode? As far as I understand, the default stays unchanged and it requires specific build options to be included (which is handled via autobuild). At the same time, this PR also adds code which is not guarded by an `#ifdef`, but I guess it's guarded in other ways.

I'm open to merging it if we can be really sure that it does not break the standard mode, if it is basically working and if doing so would help in testing.

Needs a rebase, btw.

@npostavs @jamulussoftware/maindevelopers What do you think?

Jamulus.pro
npostavs: Oh, that was leftover from when I had portaudio as the default. Dropping it.

npostavs:
> Is there a benefit of merging this to make it easier for people to try this mode?

Perhaps not so much, since this is only useful for Windows at the moment. We can post the auto-built exe just as easily regardless of whether it's merged or not. I think it would make sense to extend it to other platforms too (e.g., there was someone who wanted to use ALSA instead of JACK on Linux), but I don't know if I could get that done by tomorrow. Probably it's okay to wait for the next version.

> At the same time, this PR also adds code which is not guarded by an `#ifdef`, but I guess it's guarded in other ways.

I left in the gui code that makes the wasapi mode controls since it's simpler that way and it should be harmless (the controls stay hidden when not in wasapi mode). And the decision whether to hide the ASIO control panel button is now made at runtime instead of compile time. The rest of the code is ifdef'd or excluded by the project file.

hoffie:
Ok, thanks. Based on your feedback, I'm de-tagging this from 3.8.0 now. Feel free to post to the Discussions when it makes sense that more people test this.

jujudusud:
Hi here, testing on Windows with Presonus Audiobox usb. Working with WDM version but not with WASAPI.
WASAPI-001

pljones:
Windows PortAudio Installer requests:
- be a different install from the standard installer
- do offer to use the same locations, though

npostavs:
Hmm, I've never seen that dll problem, and I can't see how it could be caused by my changes.

pljones:
Hm, interestingly [Microsoft Answers](https://answers.microsoft.com/en-us/windows/forum/all/msvcp1401dll-missing-even-after-installing/9ffe1025-000c-4dda-b686-08b3ae00d074)'s first step gives:
image
And looking a little closer
```
19/05/2021  17:06           565,640 msvcp140.dll
10/10/2017  20:52           980,624 msvcp140d.dll
07/12/2019  10:10           622,832 msvcp140_clr0400.dll
```
19th May could have been a recent MS Update, I guess...

pljones:
Indeed... just as I'd typed that my PC rebooted.  Checking the event log, the last thing it had tried was an update.  Manually attempting a "Check for updates" got me a repeated fail.  DSIM failed to fix it.  sfc claimed to fix it...  I'm currently trying a manually WSUS update.

pljones:
OK... just to clear that one up -- I had to **uninstall** the Visual C++ Runtime 2015-2019.  Jamulus 3.8.0 then reinstalled it and that install is okay.  Now, let's try this one again...

pljones:
Hm.  Same again.  After the Windows Installer installs the runtime, it's no longer usable...

pljones:
So, outstanding tasks:
- See if anyone else can reproduce the problem I had above with VC++ Runtime 2015-2019
- Rebase to latest jamulussoftware/jamulus master, squashing commits

npostavs:
> It's very weird. The VC++ runtime in both 3.8.0 and this is 14.29.30037. But the 3.8.0 installer makes it work and this installer breaks it.

Huh. Seems quite impossible. But the installer stuff is all black magic to me, I'm afraid.

> * Rebase to latest jamulussoftware/jamulus master, squashing commits

It is rebased to latest master, as far as I know. The non-squashing is intentional, but if you guys want to squash-merge the PR I'm fine with that.

windows/deploy_windows.ps1
```diff
@@ -221,9 +221,19 @@ Function Build-App
         [string] $BuildArch
     )
 
-    Invoke-Native-Command -Command "$Env:QtQmakePath" `
+	if($Env:jamulus_build_config){
```
pljones, line 224: Even though clang-format hasn't caught you, could you please follow the style used elsewhere in the file.

dcorson-ticino-com:
A quick summary of the audio I get with this new version on Win10:
WASAPI
Using USB soundcard (NI Komplete Audio 2)
Exclusive mode, sound can be heard, but is covered by static and clicking with all buffer and audio selections.
Shared mode, noise and clicking with 64 sample buffer with and without Small network buffers.  
128 and 256 sample buffers sound usable.
In all cases the jitter is very high, 11 or above.  With ASIO normally I have jitter of 2 or 3.  This means that the total delay is also very high 50ms or above.

The PC internal mic has a lot of static and is not usable with all buffering and audio possibilities.  
The jitter is also 11 or more.

WDM
Using USB soundcard (NI Komplete Audio 2)
Seems to be working fine, jitter just slightly more than using ASIO.
All of the other sound sources and sinks make an error message as with the earlier version of port audio.

Things are moving in the right direction, but we still have a way to go.
Let me know if there is any way I can help with further testing.

pljones:
> 
> 
> > It's very weird. The VC++ runtime in both 3.8.0 and this is 14.29.30037. But the 3.8.0 installer makes it work and this installer breaks it.
> 
> Huh. Seems quite impossible. But the installer stuff is all black magic to me, I'm afraid.
> 

Well, I'm going to have to put it down to something local getting broken and now being fixed sort of fixed.  But it seems to happen intermittently.  I can reliably fix it by uninstalling the runtime before installing Jamulus... but it also seems to work now anyway.  So I don't quite know what changed.  Definitely local, though, as I used the same installer!!

src/soundbase.h
```diff
@@ -103,7 +117,9 @@ class CSoundBase : public QThread
 
     virtual float GetInOutLatencyMs() { return 0.0f; } // "0.0" means no latency is available
 
-    virtual void OpenDriverSetup() {}
+    virtual void           OpenDriverSetup() {}
+    virtual EPaApiSettings GetExtraSettings() { return PaApiNoExtraSettings; }
```
pljones, line 121: Why do this line and the next need to be here?  `src/soundbase.h` shouldn't have driver or underlying API-specifics in.

src/soundbase.h
```diff
@@ -51,6 +51,20 @@ enum EMidiCtlType
     None
 };
 
+enum EPaApiSettings
```
pljones, line 54: Not particularly happy with PortAudio code being in `src/soundbase.h`.  Definitely not happy with WASAPI code being in here.

src/main.cpp
```diff
@@ -602,8 +613,18 @@ int main ( int argc, char** argv )
...
             // actual client object
-            CClient
-                Client ( iPortNumber, iQosNumber, strConnOnStartupAddress, strMIDISetup, bNoAutoJackConnect, strClientName, bMuteMeInPersonalMix );
+            CClient Client ( iPortNumber,
+                             iQosNumber,
+                             strConnOnStartupAddress,
+                             strMIDISetup,
+                             bNoAutoJackConnect,
+                             strClientName,
+#ifdef USE_PORTAUDIO
```
pljones, line 622: I'd drop the `#else` block and have the client not have the parameter except when using PortAudio, rather than have an empty string going nowhere.

src/clientsettingsdlgbase.ui
```diff
@@ -358,6 +358,31 @@
              </property>
             </spacer>
            </item>
+           <item>
```
pljones, line 361: Hm.  This doesn't look right.  I prefer JACK's approach of listing all the devices as "API: device driver name".  Then the existing panel layout can remain and the command line argument isn't needed, making it easy to use.

src/clientsettingsdlg.h
```diff
@@ -87,6 +88,7 @@ public slots:
     void OnNewClientLevelEditingFinished() { pSettings->iNewClientFaderLevel = edtNewClientLevel->text().toInt(); }
     void OnInputBoostChanged();
     void OnSndCrdBufferDelayButtonGroupClicked ( QAbstractButton* button );
+    void OnWasapiModeButtonGroupClicked ( QAbstractButton* button );
```
pljones, line 91: As elsewhere, not happy with this.

src/clientsettingsdlg.cpp
```diff
@@ -367,13 +367,33 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     chbDetectFeedback->setWhatsThis ( strDetectFeedback );
     chbDetectFeedback->setAccessibleName ( tr ( "Feedback Protection check box" ) );
 
+    if ( pNCliP->GetExtraSettings() != PaApiWasapiModes )
```
pljones, line 370: There shouldn't be hardware-specific code in here.

src/client.cpp
```diff
@@ -31,6 +31,7 @@ CClient::CClient ( const quint16  iPortNumber,
                    const QString& strMIDISetup,
                    const bool     bNoAutoJackConnect,
                    const QString& strNClientName,
+                   const QString& strApiName,
```
pljones, line 34: `#ifdef` around this.

src/client.h
```diff
@@ -113,6 +107,7 @@ class CClient : public QObject
               const QString& strMIDISetup,
               const bool     bNoAutoJackConnect,
               const QString& strNClientName,
+              const QString& strApiName,
```
pljones, line 110: `#ifdef` around this.

src/client.h
```diff
@@ -194,6 +189,9 @@ class CClient : public QObject
     int     GetSndCrdLeftOutputChannel() { return Sound.GetLeftOutputChannel(); }
     int     GetSndCrdRightOutputChannel() { return Sound.GetRightOutputChannel(); }
 
+    EPaApiSettings GetExtraSettings() { return Sound.GetExtraSettings(); }
```
pljones, line 192: As elsewhere, not happy seeing hardware-specific code here.

src/clientsettingsdlgbase.ui
npostavs: This doesn't account for the shared vs exclusive of WASAPI mode though (which is what the code you are commenting is about)?

I also dispute that multiplying the number of device entries by (up to) 3 will make things easier. Having n^2 entries due to the Cartesian product of input and output devices is already bad enough.

pljones: Why can't they be returned as a single driver's list of inputs and outputs, anyway, rather than having to be returned each as a separate driver as they are now?

pljones:
OK, so now I've tried using it without the VC++ runtime issue.

The PA-ASIO is annoying.  For a driver it doesn't like, it doesn't simply ignore it, it issues repetitive errors.  And it does it on every change to anything driver-related, of course.

I also found it thought one of my drivers was in use when I selected it, whereas the standard Jamulus ASIO code had no problem.

The WDM-KS mode didn't work - it listed the devices, let me select my sound card in/out pair but had no sound in or out.  (That's the device the PA-ASIO code thought was in use.)

The WASAPI mode didn't work - it didn't list _any_ devices.

(One way around or the other for WDM-KS / WASAPI - I forgot to note which was which.)

This was with `jamulus_portaudio_3.8.0dev-7d7e464_win.exe` from the autobuild.

src/clientsettingsdlgbase.ui
npostavs: I don't understand what you mean.

pljones: OK, go about it more list ASIO - but, rather than separating out each hardware device, have the PortAudio API be the device.  This way you could choose the API without having to have `--api`.

Then you have Input 1, Input 2, Output 1 and Output 2 boxes.  For the chosen API, you'd return all the inputs to populate the input boxes and all the outputs to populate the output boxes.

Benefits:
* No cartesian product.
* Similar to the existing user experience.
* No command line option required.

npostavs: Oh, I see. I guess that works as long as the devices themselves don't have multiple channels, I think this is usually the case for non-ASIO devices. I'm wouldn't be too happy about special-casing the code like that. Also it still doesn't solve the shared vs exclusive WASAPI mode thing.

> * Similar to the existing user experience.

The Cartesian product thing is the existing user experience on macOS, by the way.

pljones: I'm explicitly referring to ASIO here as the primary target for this change is Windows ASIO4ALL users.

ann0see:
I also had the crackling sound. I think it’s similar to the sound I hear if I choose pull or push (?) mode in ASIO4ALL.

ann0see:
@npostavs I think this should go to a feature branch (not master) then? What do you think?

npostavs:
Yeah, I'm not feeling very optimistic about completing this right now. It might make more sense to work out the kinks with PortAudio using KoordASIO (#1936) and then maybe revisit this more integrated approach later.

ann0see:
Your work is now here: https://github.com/jamulussoftware/jamulus/tree/feature/np/portaudio 
At least it's somehow official now ;-)

Issue #475, opened 2020-07-25
Version number of unreleased version


fleutot:
Version numbering for untagged commits creates confusion. People see 3.5.9git and wonder where they can get version 3.5.9, before it is released.

Suggestion: use the `git describe --long --tags --dirty --always`
See [here](https://softwareengineering.stackexchange.com/a/141986/6363)

The result of this command can be used in a makefile to define a C macro (to pass to gcc with - D), which generates a relevant version number everytime you run make.

The result is:

    <last tag>[-g<SHA1>][-dirty]

, or something along that line. Last tag is used, instead of upcoming version.

If another commit is checked out, you know which one. If there are local changes for that build, you know that too.

If that looks like a good idea, I can work on that next week.

pljones:
I was thinking along these lines (I think there may be another issue that mentioned it in passing?).  I do agree it would cut the confusion.

fleutot:
@pljones we had that discussion in the Facebook group.

corrados:
I don‘t think we need a complicated versioning. If 3.5.10git is confusing, I could use the last version instead: 3.5.9git

fleutot:
How do you update til `*.*.*git`? Manually? The described scheme is not that complex, once implemented everything happens automatically.

corrados:
I only update it once I make a new release and keep it in sync with the ChangeLog.

corrados:
I have now changed the version number from 3.5.10git to 3.5.9git. I'll do this strategy for the next versions, too. I'll close this Issue now. If the confusion will not be solved by this, just re-open this issue again or create a new one.

storeilly:
Would it be possible to include optional, detailed commit id's in the project file? It would help to identify exe's where there are multiple versions undergoing testing. I have the following in my Jamulus.pro file which might help. It only affects the make when I include "versionifo" in the config string.

================================================
VERSION = 3.5.12git

#use target name which does not use a capital letter at the beginning
contains(CONFIG, "noupcasename") {
    message(The target name is jamulus instead of Jamulus.)
    TARGET = jamulus
}

#allow detailed version info in target executable
contains(CONFIG, "versioninfo") {
    message(Using custom target and version)
    GIT_CURRENT_SHA1=$$system(git describe --tags)
    VERSION = $$VERSION-$$GIT_CURRENT_SHA1
    TARGET = $$TARGET-$$VERSION
    message(Version is $$VERSION)
    message(Target is $$TARGET)
}

CONFIG += qt \
[...]

PR #883, opened 2021-01-24
#475 git hash & status in verison info

nefarius2001:
my answer to #475 

for all "normal" builds a git hash will be appended, separated to the version with a "-"
by specifying "official_release_version" in CONFIG this is not happening

"3.6.2"  << official release
"3.6.2-37f993aa" << intermediate release
"3.6.2-37f993aa-dirty" << intermediate release

tested on windows looking at the "about" info

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -1,4 +1,4 @@
-VERSION = 3.6.2
+VERSION = 3.6.2dev
 
 # use target name which does not use a captital letter at the beginning
 contains(CONFIG, "noupcasename") {
@@ -8,7 +8,12 @@ contains(CONFIG, "noupcasename") {
 
 #allow detailed version info for intermediate builds
 contains(CONFIG, "official_release_version") {
+	#for creating a new release:
+	# - count the VERSION up 
+	# - create a new commit
+	# - run build with "official_release_version" on it
     message(building an official release version)
+	VERSION = ($$replace(VERSION, dev, ))
 } else {
     GIT_DESCRIPTION=$$system(git describe --match=xxxxxxxxxxxxxxxxxxxx --always --abbrev --dirty) # the match should never match
     VERSION = "$$VERSION"-$$GIT_DESCRIPTION
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -7,17 +7,12 @@ contains(CONFIG, "noupcasename") {
 }
 
 #allow detailed version info for intermediate builds
-contains(CONFIG, "official_release_version") {
-	#for creating a new release:
-	# - count the VERSION up 
-	# - create a new commit
-	# - run build with "official_release_version" on it
-    message(building an official release version)
-	VERSION = ($$replace(VERSION, dev, ))
-} else {
+if (contains(VERSION, .*dev.*)) {
     GIT_DESCRIPTION=$$system(git describe --match=xxxxxxxxxxxxxxxxxxxx --always --abbrev --dirty) # the match should never match
     VERSION = "$$VERSION"-$$GIT_DESCRIPTION
     message(building an intermediate version: $$VERSION)
+} else {
+    message(building a final release version: $$VERSION)
 }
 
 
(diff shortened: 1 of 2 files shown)


nefarius2001:
@corrados  please state if these version names are fine to you, or if you want to keep the “git” appendix. To me a hash is more intuitive to indicate a intermediate state, than the name of a tool/a location.

@ann0see : good point about change log detection, will check

ann0see:
We could also make your proposal compatible with the current one (just add git to the version). 

Or another proposal:

3.6.2 official release

3.6.2dev during development (unofficial)? This wouldn't break the changelog detection and be less confusing than git.

corrados:
> 3.6.2dev during development (unofficial)

That is a good idea.

corrados:
@fleutot and @storeilly Do you also like to comment on this pull request?

storeilly:
I think that detailed version info needs to be available somewhere, especially for headless operation.
For me the dev / git extension would only be significant to show official release versions and I've removed them in favour of the "rev-parse --short HEAD" command in all of my builds as they are currently different from the official release.

fleutot:
@storeilly I think using "git describe" might make things easier than "git rev-parse".

storeilly:
> 
> 
> @storeilly I think using "git describe [options]" alone suffices. It should look for the latest annotated tag and you wouldn't need to manually change "3.6.2".
> 
> What do you think of the behavior of `git describe --dirty=-dev --always`?

It's very clever and probably the correct method for the main code base, I'm not qualified to dispute it :) On my fork the hex string is enough for me. (Team of 1)

nefarius2001:
> What do you think of the behavior of git describe --dirty=-dev --always?
I see a danger for confusion " --dirty=-dev", because dev means between releases, and dirty means between commits.

Also I think a hash is "safer" for implementing, known length and uncritical characters:
`git describe --match=xxxxxxxxxxxxxxxxxxxx --always --abbrev --dirty`

Jamulus.pro
```diff
@@ -1,11 +1,26 @@
...
+VERSION = 3.6.2dev
 
 # use target name which does not use a captital letter at the beginning
 contains(CONFIG, "noupcasename") {
     message(The target name is jamulus instead of Jamulus.)
     TARGET = jamulus
 }
 
+#allow detailed version info for intermediate builds
+contains(CONFIG, "official_release_version") {
```
corrados, line 10: Why to use a config flag for it? Couldn't you parse the VERSION and if you find a "dev" then you know it is a developer version and if no "dev" is present, you know it is an official release version.

corrados: What I do when I create a new release is that I modify the ChangeLog and then I update the VERSION in the .pro file and create a Git commit from it. Then I apply the release tag. Then ann0see's automation detects that it is an official release version and starts his scripts. In that Git commit, the VERSION is set correctly to a release version number (which matches the entry in the ChangeLog).

ann0see: > Then ann0see's automation detects that it is an official release version and starts his scripts [...]

Exactly. If you merge @nefarius2001 changes, the latest tag will not be applied to this commit and you'll still need to add it automatically. (Just wanted to add this).

nefarius2001: @corrados I had concerns, that somehow it might be forgotten/screwed up to add the "dev" appendix after removing it for the release. But it makes the job much easier, so lets go that road

corrados: @ann0see "the latest tag will not be applied to this commit" -> What do you mean by this? Do you refer to your Git automation scripts which do a release if a release tag is applied to the repo? But here we are talking about the version number shown in the Jamulus software. If the "dev" is not there, nothing happens (at least when looking at the code after the latest change).

corrados: > I had concerns, that somehow it might be forgotten/screwed up to add the "dev" appendix after removing it for the release.

The is the very first thing I am doing after finishing a release. You can take a look at the Git history if you like.

ann0see: >  Do you refer to your Git automation scripts which do a release if a release tag is applied to the repo?

Yes. That was just an addition to your answer. I wasn't sure you were aware of this.

corrados:
Thanks.

Issue #875, opened 2021-01-21
Server lists getting full


pljones:
Default: 150 - every time I've checked the last three days
All Genres: 150 - every time I've checked the last three days
Rock: 51 - over 50 most days
Jazz: 59 - over 50 most days
Classical/Folk/Choir: 136 - over 120 most days... I think over half of of these are Choir or choral...

I think we need to start thinking about what happens next.  The original plan was lock down would end by last October at latest and life would go back to normal...  It's unlikely to be back to normal this year.

So people can't set up "Default" or "All Genres" servers - which is irritating.  I'm sure many people running servers don't think of them as having a specific genre.  I don't - and I really don't want my server dropping of All Genres, having lost its Default list listing already... :)

DavidSavinkoff:
Maybe the server list could leave room for temporary servers, rather than encouraging servers to stay connected 24/7 with a Bold font after 3 days. Otherwise, you will have server squatting (maybe you can make money with this feature).

corrados:
Servers with version >=3.6.1 sending the version number when registering. So one solution would be to only allow servers with version >=3.6.1 to register in the Default and All Genres lists. That would also solve the issue that so many very old Jamulus server versions are present. But some server operators may be very disappointed by that.

Opinions?

corrados:
@pljones maybe you could also start a thread about this >=3.6.1 in the Facebook group to get some feedback from there.

corrados:
> Classical/Folk/Choir: 136 - over 120 most days... I think over half of of these are Choir or choral...

We could split this list: one list for Classical/Folk and one for Choir. Would you be able to host the new list on your server?

pljones:
Yes.  Classical/Folk/Choir just hit 150..!  (It was 146 and then a couple of seconds later, 150!)

pljones:
> @pljones maybe you could also start a thread about this >=3.6.1 in the Facebook group to get some feedback from there.

The joy of herding cats...  I don't expect any input from server operators running older versions as I don't expect they're paying attention.

pljones:
I've set up server list servers on 22124 through 22824 on my machine now.

corrados:
Thanks. So let's go with a split. But the question is: Who has to move to the new list? Choral or the other? I would wote for Choral to stay and the others get a new list. But the problem is that the server operators choose which list they want to be. So the process of moving to the new list will take long...

ann0see:
[...]

PR #885, opened 2021-01-24
Introduced new server lists: Any Genre 2, Choral/Barbershop (#875)

corrados:


ann0see:
Sorry to be that guy, but why don't you use a for loop and loop through an array which contains all the lists/names? This would make adding new lists much easier.

corrados:
What place in the code you are referring to?

> This would make adding new lists much easier.

I don't think we need more lists in the near future.

src/connectdlg.cpp
```diff
@@ -74,10 +74,12 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP,
     // central server address type combo box
     cbxCentServAddrType->clear();
     cbxCentServAddrType->addItem ( csCentServAddrTypeToString ( AT_DEFAULT ) );
```
ann0see, line 76: You could loop through an array containing the lists starting from here. 

Could this be done with a compiler macro or something like this: https://stackoverflow.com/questions/24392000/define-a-for-loop-macro-in-c#24392091

src/serverdlg.cpp
```diff
@@ -242,10 +242,12 @@ lvwClients->setMinimumHeight ( 140 );
     // central server address type combo box
     cbxCentServAddrType->clear();
     cbxCentServAddrType->addItem ( csCentServAddrTypeToString ( AT_DEFAULT ) );
```
ann0see, line 244: Same here

corrados:
I hate compiler macros. Let's keep the code as is right now and the next time we add new lists, we'll change the code to be a loop.

ann0see:
Ok. Sounds good

corrados:
Ok, so let's merge it. Thanks.

ChangeLog
```diff
@@ -2,7 +2,9 @@
 
 
 
-### 3.6.2git <- NOTE: the release version number will be 3.6.3 ###
+### 3.6.2git <- NOTE: the release version number will be 3.7.0 ###
+
+- introduced new server lists: Any Genre 2, Choral/Barbershop (#875)
```
pljones, line 7: This comment is wrong, by the way.  What the change does is https://github.com/jamulussoftware/jamulus/pull/885/files#diff-40ab92941994ec9418dbb76ae46d5aef0e4cb8b034f2be2b31562bb9aca8070f
- rename "Default" to "Any Genre 1"; rename "All Genres" to "Any Genre 2"; rename "Genre Classical/Folk/Choral" to "Genre Classical/Folk"
- add "Any Genre 3" and "Genre Choral/Barbershop"

Issue #314, opened 2020-05-31
Discussing potential security issues


corrados:
The number of Jamulus servers in the server list grew significantly in the last three month. If the Jamulus software has security issues, a lot more people are affected now as it was in the past. Therefore I open this Issue to start a discussion about potential security issues.

Since Jamulus opens a network port and listens for UDP packets, this is the entry point for possible attacks. Basically Jamulus detects if the received UDP packet is a protocol message and parses it. If it detects that it is no protocol message and the size of the packet is correct, it puts it directly into the OPUS decoder. Assuming that the OPUS decoder is stable, the most critical part is the protocol parser. To test the parser with random messages, a test bench is already implemented in Jamulus: https://github.com/corrados/jamulus/blob/master/src/testbench.h. More advanced tests would be to use tools like this: https://lcamtuf.coredump.cx/afl

I read posts about considering encryption or sign the network packets. But actually I do not see a need for that. Jamulus is an open system so encryption of the audio packets is not necessary in my opinion since someone who wants to hear the stream can simply connect to the server. Signing will also not work since an arbitrary client can connect to any available server.

Since I am no security expert, with this Issue I just want to collect your opinions on that matter. If this discussion leads to specific implementation need, a new Issue shall created for it. But this Issue shall only be for the discussion (therefore I added the "documentation" label to it).

pljones:
This was going to be vaguely coherent but...

Whilst privacy is one of many forms of security it shouldn't be the main focus of anything required for Jamulus, as Jamulus does not publish personal ("person"-related) information without user consent.  Cryptographic technologies mostly relate to privacy ("tampering" isn't always a privacy issue but it's usually taken as one when cryptographic solutions are suggested).  Like you, I'd discount them.  (They're also expensive in CPU cycles.)

Currently I find it "hard" to connect to certain servers (e.g. mojolounge and nylonmojo that open port 22120).  Whilst I get some protocol messages (I get the list of other participants, for example), I don't get audio protocol messages (and may well not get others).  I think others may have similar difficulty with my server (some people do okay...).  I'm guessing this relates to firewalls - so I think they do seem to be an area that needs consideration here.  They don't all work the same way, some aren't able to do bi-directional port forwarding, sometimes the source port on a packet is re-written by the forwarding, sometimes it seems the firewall port-forwarding gets ... lost on the way back (there's no connection...).

[...]

PR #886, opened 2021-01-24
Chat: Make http:// URLs clickable

### LINKED ISSUE #879: Support http links in the chat dialog
Links http: not clickable, only https: becomes clickable.
Is there a reason for it or a bug?

hoffie:
Previously, only https:// URLs were whitelisted as part of a security
hardening to prevent problematic handlers such as file:// (#314, #360).

This PR adds http:// to the whitelist. This is helpful as some resources
are still not available using encrypted https:// connections.

Fixes #879.

### DIFF
--- a/src/chatdlg.cpp
+++ b/src/chatdlg.cpp
@@ -114,14 +114,14 @@ void CChatDlg::AddChatText ( QString strChatText )
     // notify accessibility plugin that text has changed
     QAccessible::updateAccessibility ( new QAccessibleValueChangeEvent ( txvChatWindow, strChatText ) );
 
-    // analyze strChatText to check if hyperlink (limit ourselves to https://) but do not
+    // analyze strChatText to check if hyperlink (limit ourselves to http(s)://) but do not
     // replace the hyperlinks if any HTML code for a hyperlink was found (the user has done the HTML
     // coding hisself and we should not mess with that)
     if ( !strChatText.contains ( QRegExp ( "href\\s*=|src\\s*=" ) ) )
     {
-        // searches for all occurrences of https and cuts until a space (\S matches any non-white-space
+        // searches for all occurrences of http(s) and cuts until a space (\S matches any non-white-space
         // character and the + means that matches the previous element one or more times.)
-        strChatText.replace ( QRegExp ( "(https://\\S+)" ), "<a href=\"\\1\">\\1</a>" );
+        strChatText.replace ( QRegExp ( "(https?://\\S+)" ), "<a href=\"\\1\">\\1</a>" );
     }
 
     // add new text in chat window
@@ -130,8 +130,8 @@ void CChatDlg::AddChatText ( QString strChatText )
 
 void CChatDlg::OnAnchorClicked ( const QUrl& Url )
 {
-    // only allow https URLs to be opened in an external browser
-    if ( Url.scheme() == QLatin1String ( "https" ) )
+    // only allow http(s) URLs to be opened in an external browser
+    if ( Url.scheme() == QLatin1String ( "https" ) || Url.scheme() == QLatin1String ( "http" ) )
     {
         if ( QMessageBox::question ( this, APP_NAME, tr ( "Do you want to open the link" ) + " <b>" + Url.toString() +
                                      "</b> " + tr ( "in an external browser?" ), QMessageBox::Yes | QMessageBox::No ) == QMessageBox::Yes )


corrados:
Hmm. What is about the security issue? @atsampson, @ann0see, @gilgongo, @pljones What is your opinion on the security risk?

ann0see:
If you allow https://, I don't really see a problem with http:// (But I'm not an expert on this). Since there's a message asking the user if he really wants to open <url>, I don't see a big security issue.

gilgongo:
I guess in so far as all security is a balance between utility and safety, we think the utility of having clickable links outweighs the possibly insecurity of clicking them.

pljones:
Given this was previously specifically decided against, I'd expect there to be an overriding argument presented before changing anything that addresses the security concerns, rather than just "it's not nice".

gene96817:
https is supposed to help block (mitigate?) spoofing. In another thread, I was warned about jamulus.io.fakehost.com. My browsers default behavior is to block http. It is not a problem for me, but for the naive users I am onboarding, it would be good to eliminate the use of http. Fewer things to go wrong with naive users.

gilgongo:
@gene96817 It seems pretty (probably increasingly) unlikely that people will legitimately post links to non-SSL sites. Almost to the point that if they do, then that's a marker for illegitimacy now?

hoffie:
> Given this was previously specifically decided against, I'd expect there to be an overriding argument presented before changing anything that addresses the security concerns, rather than just "it's not nice".

It's a usability thing. I work with people who hardly know what a browser or what copy and paste is. They had someone else set up their machine and their Jamulus instance. For those people, it's way easier to have a clickable link. Would also be interesting to know what @AndyMcProducer's use case was (cf. #879).

I fully agree that `https://` is better than `http://`. Browser vendors are pushing in that direction as well. However, let's take a common example for choirs: a huge public score collection such as cpdl.org. While the original site does provide `https://`, there are also lots of links to third-party websites which still support `http://` only. When wanting to link to those, a usability hurdle appears: Either the users have to copy and paste the link into their browsers or some technically knowledgeable person has to download and re-upload this score to some `https://`-enabled website.

Regarding security, I think it would be important to understand what threat we are trying to mitigate:
- The original suggestion in https://github.com/corrados/jamulus/issues/314#issuecomment-644191787 was about practical vulnerabilities regarding the `file://` protocol and maybe other protocols where their exploitability for malicious actions is not even known by now (with examples from Pidgin/Zoom). As far as I understand, this is about (pseudo-)protocols with unintended side-effects, i.e. especially those which are not typically opened by browsers. I fully agree that links with such protocols must be filtered and I think this issue is real because [openURL()](https://doc.qt.io/qt-5/qdesktopservices.html) will apparently open/execute anything it is passed.  I'm also fully in favor of the approach of @corrados who chose a whitelist-based approach instead of a blacklist-based one. This is basically about the **threat of potential security vulnerabilities in Jamulus**. I don' think that whitelisting `http://` would change anything in this regard.
- There is a **potential threat about leading unknowning users to bad websites** (I think that's what @gene96817 meant and what @corrados hinted at in https://github.com/corrados/jamulus/issues/314#issuecomment-643641443). I have too little experience to tell if this is a real problem on public Jamulus servers (i.e. bad users joining and posting bad URLs to the chat), but I am pretty confident that the decision about whitelisting `http://` changes little to nothing in this regard. `https://` (or TLS and its X.509 certificates) does not tell anything about the reputation of a website. Give me some minutes and I'll set up a fake jamulus.io.example.org website with or without https on some domain I actually own (no, I'm not intending to do that ;)). `https://` solely ensures that the connection is authenticated and encrypted. Basically, it adds confidence that you are talking to the website you intended to visit.
- There is a **potential threat about leading users to bad websites via phishing**, i.e. via creating a link which displays as `example.org/good.html` but points to `example.org/bad.html` instead. Jamulus already protects against this issue by asking the user about the actual URL (and not the displayed text). This protection should work for both `http://` and `https://`.

The only security argument which remains is: "people should only be using authenticated and encrypted communication". I generally like this sentiment. However, applying this argument in this http/https discussion sounds a bit strange to me, considering that Jamulus is supposed to be an open ecosystem with no authentication or encryption builtin.

Therefore, I do not see any specific security downsides with whitelisting `http://`. If you have some, I'm happy to hear about them.

gene96817:
For Jamulus, the main benefit of https is to ensure the user is connected to the site we intended.

atsampson:
I'd prefer it if Jamulus treated `http:` and `https:` links the same - the decision about whether to allow regular HTTP (or automatically upgrade it to HTTPS, in the cases where that's possible through HSTS or a plugin like HTTPS Everywhere) is one that the browser should make, not Jamulus.

corrados:
Thanks for all your comments! I'll merge it now.

AndyMcProducer:
I'd have thought some where there would be a list of blacklisted sites or a way to check realtime.

Issue #861, opened 2021-01-18
Grammar and wording checks installer.nsi


ann0see:
@gilgongo I think this is something for you. Could you please check the new [installer.nsi](https://github.com/corrados/jamulus/blob/master/windows/installer.nsi) file for language mistakes?

Furthermore, we should link an ASIO section on jamulus.io in the new installer.

corrados:
This: https://github.com/corrados/jamulus/blob/1ce54ba0566c9064939f88ea13f35999cb97ae61/windows/installer.nsi#L31 is from the "old days of Jamulus". Should we keep it or use an updated text instead?

ann0see:
I think this depends on what you think. The website uses "Make music online. With friends. For free." but I think it could be changed to "Jam online. With friends. For free" or maybe we could open a community voting?

ann0see:
Also the ASIO detection warning must explain a bit more.

dcorson-ticino-com:
I would go with "Make music".  I have noticed that may musicians around me (French part of Switzerland) don't know the term "to jam".  Thus they find the name Jamulus strange.
I thought "to jam" was well known and international, but apparently it isn't.

corrados:
> The website uses "Make music online. With friends. For free." but I think it could be changed to "Jam online. With friends. For free"

I would use the same as on the homepage. If you change the text on the homepage, you should then change it in the installer as well.

> https://github.com/corrados/jamulus/blob/a3238e960741e08526d1482c2a5efc6e56bdd897/windows/installer.nsi#L92

I would not use "bug". I would remove the sentence "Most likely, this is due to a bug in the old installer of Jamulus.". If the users read about a bug right when they want to install Jamulus, that might confuse them. The sentence does not give a Jamulus user any useful information for him.

ann0see:
Ok. The user should nevertheless know that it's 99% not a problem he introduced.

corrados:
Well, he actually did introduce it by installing the old Jamulus version. Anyway, we should keep the text as short and informative as possible. The info we want to ship: For proper runtime behavior of the Jamulus software, it is recommended to uninstall the old Jamulus version.

ann0see:
Yes. But the problem of this is that it is not only an old version of Jamulus but it is a wrongly installed version of Jamulus.

If one time somebody would introduce a 32Bit installer for 64Bit windows (I wouldn't really see why this would be needed but there was a feature request somewhere) this message should still work and explain the "issue".

[...]

PR #904, opened 2021-01-27
Fit slogan to website

ann0see:
As proposed by @corrados in #861


corrados:
FYI: I have just added a protection to the master branch so that all merge requests needs a review. I think this makes sense. For a test I'll try out if I can overwrite this as an administrator (e.g. if we need a quick fix) and this seems to be possible.

ann0see:
Great.

PR #920, opened 2021-01-29
 Delete release on action start

ann0see:
If we don't remove the old release (especially the one tagged with latest) only a draft release will be created.

See: https://github.com/ann0see/jamulus/commit/66e0b6eeb1298b33aef8abec53c5349231abb5f7


ann0see:
I think once this is merged, a beta release should be published.

corrados:
Let's wait until the repo is moved to jamulussoftware and also wait for fischvolk.de to be replaced by jamulus.io.

Issue #923, opened 2021-01-29
Memory leak in the server


corrados:
I just observed the following on the Central Server: Right now we have heavy traffic on that server. The server run for about a month and the Jamulus server instance used about 500 MB. When I reboot the server, Jamulus just consumes about 4 MB.

I think the following code might be the problem:
https://github.com/jamulussoftware/jamulus/blob/ab785cb8d5e722663bb7887c8566a29a442ff86e/src/serverlogging.cpp#L65
We do a "new" all the time but do not delete that instance. Maybe it is better to completely remove the ConsoleWriterFactory() class and use QInfo instead.

pljones:
Yes, I'd back that, as we've moved to a dependency where `qInfo()` will always exist.  It'll need a little refactoring (as the messages format differently).

corrados:
Would you like to do it or shall I do this change?

pljones:
👍 I'll do it now - shouldn't take long.

corrados:
Thanks.

pljones:
Pull request: https://github.com/jamulussoftware/jamulus/pull/926

hoffie:
@pljones Now that #926 is merged, this could be closed?

softins:
> @pljones Now that #926 is merged, this could be closed?

If indeed that was the source of the leak. Time will tell!

pljones:
Yes, it's probably wise to monitor a server - preferably one that's long running and heavily used - to see if this the problem has gone away.

pljones:
Can we tag this as needing some kind of post-3.7.0 review (and hopefully then closing)?  I'd expect checking the process size on the Any Genre 1 directory server immediately after installing 3.7.0, then at intervals of a week for three weeks should provide enough information.  If the size is stable, it's fixed.

genesisproject2020:
I compiled the code yesterday (3.6.2dev-6d79bf1b) and it has now been running on "Swedish Open Jam 1" for 25h. During that time the amount of allocated memory went from 40Mb to, at the moment, 1Gb. I've also changed it on a private instance where a group had a rehearsal this evening with 12 members and after the 2h rehearsal the service have allocated 1,7Gb of memory and it is stuck there now when the instance is idle again. I do believe something is still eating the memory. Can I assist in someway to help finding the leak?

My environment: Ubuntu 20.04 LTS (5.4.0-65-lowlatency)

genesisproject2020:
[...]

PR #926, opened 2021-01-30
#923 Remove ConsoleWriterFactory

pljones:
All instances of `ConsoleWriterFactory` have been removed.

In addition, on Windows (when not headless), the console has been attached once on start up to ensure messages appear.  This only happens when run from the command line - otherwise the messages don't appear.

Parameterised messages use `QString` formatting but that's needed an extra `qUtf8Printable` call to pass the result to `qInfo()` etc.

A few of the start up messages were warnings, so they're now `qWarning()`, rather than `qInfo()`.  Start up failure is now displayed as `qCritical()` messages.

As we've generally moved to `qInfo()` etc, I removed the QT version check where it was in place.

----
Unrelated bits...
* I changed the help text for `--centralserver` to use the term "server list".
* I fixed the formatting of `--ctrlmidich` and `--clientname` to be consistent with the other long opt-only options.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -24,7 +24,6 @@
 
 #include <QCoreApplication>
 #include <QDir>
-#include <QTextStream>
 #include "global.h"
 #ifndef HEADLESS
 # include <QApplication>
@@ -48,7 +47,6 @@
 int main ( int argc, char** argv )
 {
 
-    QTextStream&   tsConsole = *( ( new ConsoleWriterFactory() )->get() );
     QString        strArgument;
     double         rDbleArgument;
     QList<QString> CommandLineOptions;
@@ -89,6 +87,13 @@ int main ( int argc, char** argv )
     QString      strWelcomeMessage           = "";
     QString      strClientName               = "";
 
+#if !defined(HEADLESS) && defined(_WIN32)
+    if (AttachConsole(ATTACH_PARENT_PROCESS)) {
+        freopen("CONOUT$", "w", stdout);
+        freopen("CONOUT$", "w", stderr);
+    }
+#endif
+
     // QT docu: argv()[0] is the program name, argv()[1] is the first
     // argument and argv()[argc()-1] is the last argument.
     // Start with first argument, therefore "i = 1"
@@ -101,7 +106,7 @@ int main ( int argc, char** argv )
                                "--server" ) )
         {
             bIsClient = false;
-            tsConsole << "- server mode chosen" << endl;
+            qInfo() << "- server mode chosen";
             CommandLineOptions << "--server";
             continue;
         }
@@ -114,7 +119,7 @@ int main ( int argc, char** argv )
                                "--nogui" ) )
         {
             bUseGUI = false;
-            tsConsole << "- no GUI mode chosen" << endl;
+            qInfo() << "- no GUI mode chosen";
             CommandLineOptions << "--nogui";
             continue;
         }
@@ -128,7 +133,7 @@ int main ( int argc, char** argv )
         {
             // right now only the creative commons licence is supported
             eLicenceType = LT_CREATIVECOMMONS;
-            tsConsole << "- licence required" << endl;
+            qInfo() << "- licence required";
             CommandLineOptions << "--licence";
             continue;
         }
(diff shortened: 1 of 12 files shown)


pljones:
Tests:

- tested start up with every parameter (`-v` last) and all the messages formatted ok
- started a linux server and a Windows client and connected to it - all messages formatted ok
- sent a "start new recording" signal to the server - messages formatted ok
- sent a "toggle recording" signal to the server - messages formatted ok
- registered a server at a serverlist (found a bug...) - messages formatted ok (now)

Anything else anyone can think of?

pljones:
@softins @ann0see - can you review?

softins:
Will do, but it will either be later today or else tomorrow.

ann0see:
I can look into it tomorrow

pljones:
Thanks :)

src/main.cpp
```diff
@@ -89,6 +87,13 @@ int main ( int argc, char** argv )
     QString      strWelcomeMessage           = "";
     QString      strClientName               = "";
 
+#if !defined(HEADLESS) && defined(_WIN32)
```
ann0see, line 90: I assume this is the first time we change coding style?

ann0see (commented):
It's quite a big change. It does compile, I think but should we test it before a merge again?

ann0see (approved):
Still to be tested (real world test), but since you couldn't find issues, I will approve it

softins:
I saw in several places the comment `// TODO we should use the ConsoleWriterFactory() instead of qInfo()`, so I'm trying to understand the reasons for firstly that comment, and now the opposite decision. I see `ConsoleWriteFactory` says it was implemented by @pljones 

What environments has compilation been tested on so far? I can test on Linux(RPi) and Mac(Mojave), but not (yet) PC.

ann0see:
The CodeQL check makes sure that it compiles on all 3 OS. So this should be ok. 

Raspberry Pi is untested but I‘m quite sure if it compiled on Debian and Ubuntu (both amd64) armhf should work too n

softins:
> The CodeQL check makes sure that it compiles on all 3 OS. So this should be ok.

I have much to learn! Not sure what this is.

> Raspberry Pi is untested but I‘m quite sure if it compiled on Debian and Ubuntu (both amd64) armhf should work too.

Just trying this now.

softins (approved):
Compiles and runs fine on RPi too, so approving.

Would still like to understand the reason for originally moving to `ConsoleWriterFactory` and why this no longer applies.

softins:
It would appear that my review doesn't count - presumably I don't have write access:
> At least 2 approving reviews are required by reviewers with write access.

ann0see:
Does it work now? I just changed the role maindeveloper role to have write access and made you an organisation admin. 

I think we can merge this now

softins:
> Does it work now? I just changed the role maindeveloper role to have write access and made you an organisation admin.

Thanks, github is now happy with my approval.

> I think we can merge this now

I agree.

ann0see:
Merged.

ann0see:
> I have much to learn! Not sure what this is.

It basically tests if the code compiles and to some extent checks it for vulnerabilities.

pljones:
> Would still like to understand the reason for originally moving to `ConsoleWriterFactory` and why this no longer applies.

Qt4 compatibility.  We gave up on that some time last year, I think.

pljones:
> 
> 
> Merged. @pljones I think we should still do some real world tests.

I did
```
./Jamulus -n -i foo.ini -p 55850 -t -d -e localhost -f '[1.2.3]' -l /dev/stdout -L -m /dev/stdout -o 'here;here;23' -R /tmp --norecord -s -T -u 1 -w 'Hello world' -z -M --mutemyown -c 1 -j --ctrlmidich 13 --clientname foo -l /dev/stdout
```
before the push and it seemed to work okay -- all the main.cpp messages appeared - also adding -v stopped start up still.

I also did
```
./Jamulus -n -p 55850 -d -e localhost -f '[3.0.0]' -R /tmp -s -T -u 1 -c 1
```
and tested the various recorder messages with SIGUSR1, SIGUSR2 and clients connecting.

softins:
> > Would still like to understand the reason for originally moving to `ConsoleWriterFactory` and why this no longer applies.
> 
> Qt4 compatibility. We gave up on that some time last year, I think.

Ah, cool, understood. Thanks! I remember that happening.

PR #930, opened 2021-01-31
Updated download links from corrados to jamulussoftware 

vdellamea:
Updated download links from corrados to jamulussoftware  in installscripts


ann0see (approved):
Although I didn't test it, I think it should work ok. There might be some changes needed to these files (quoting could be improved). But that's another issue

jerogee (approved):
Tested tag name extraction and fetching with wget for new links. Works as expected.

Other than that: having 3 scripts that are 99% identical does not make sense. Should be merged. But that is another for another issue to improve.

pljones (approved):

Issue #890, opened 2021-01-25
Chat allows impersonation


hoffie:
The chat permits users to use HTML/CSS for formatting. This formatting can be used to post messages which perfectly resemble the look of chat messages sent by other people.

![chat-impersonation](https://user-images.githubusercontent.com/762915/105739000-ee40db80-5f37-11eb-9267-2d294026b20c.png)

Possible choices:
1. Accept the risk and document that it is supposed that all Jamulus users are well-behaved. One might argue that this is expected in other places as well (such as people not disturbing other group's sessions or not attempting to cause confusion by deliberately duplicating usernames). The main difference may be that the audio of bad users can be muted, while chat messages can't (and maybe shouldn't, especially when someone is muted due to audio issues and the chat is being used to diagnose/help).
2. Try preventing such misuse. I think this is hard and prone to be subject to be circumvented again. One could try forbidding all variations of `<br>`, `<p>`, `<div>` and `<anyelement style="display:block">` in order to disallow line breaks within one chat message. I think it's hard to catch all methods which cause line breaks. It will even be harder to do that with a regexp-based approach, without parsing HTML.
3. Disable HTML for ordinary chat messages.
    - Personally, my impression is that HTML is often used for servers' welcome messages to provide some nicely structured header/information. This is no problem because the welcome message can only be changed by the server admin who has to be trusted anyway. Therefore, this could be left as-is.
    - I have seldomly had the need to use HTML in my chat messages. I did not even know that this is possible until reading the relevant code. I don't know how many people currently rely on this feature and would be sad to see it going away.
    - Nevertheless, this would be a breaking change. The simplest implementation would probably be to force this server-side by HTML-escaping all user-supplied text.
    - If deemed necessary, it might still be useful to permit some formatting (bold text or something). This could be implemented by permitting Markdown or specific whitelisted HTML tags.
    - This option would also help to avoid the formatting-leaking behavior mentioned in #869.

Personally, I do not care too much. I mainly opened this issue because it looks unintended and because other discussions revealed that some people do care a lot about preventing abuse of Jamulus features especially for less computer-addicted persons. :)

This issue is what I hinted at in #887.

corrados:
Interesting observation, thanks for sharing. Blocking HTML for normal messages would be a solution. Let's open a discussion on that: @pljones, @ann0see, @gilgongo, others?

ann0see:
[...]

PR #939, opened 2021-02-01
Escape chat messages at the server

atsampson:
Update my patch to strip HTML from chat messages from #380, as discussed in #890 (@hoffie).

One thing to test with this before merging is that URLs containing `&` still get link-ified correctly.

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1395,7 +1395,7 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int      iCurChanID
         "<font color=""" + sCurColor + """>(" +
         QTime::currentTime().toString ( "hh:mm:ss AP" ) + ") <b>" +
         ChanName.toHtmlEscaped() +
-        "</b></font> " + strChatText;
+        "</b></font> " + strChatText.toHtmlEscaped();
 
 
     // Send chat text to all connected clients ---------------------------------
--- a/src/protocol.cpp
+++ b/src/protocol.cpp
@@ -167,6 +167,9 @@ MESSAGES (with connection)
     | 2 bytes number n | n bytes UTF-8 string |
     +------------------+----------------------+
 
+    - "UTF-8 string": the chat message (plain text from client to server;
+      HTML from server to client)
+
 
 - PROTMESSID_NETW_TRANSPORT_PROPS: Properties for network transport
 


hoffie:
> One thing to test with this before merging is that URLs containing `&` still get link-ified correctly.

Uh, good point. I just tried and they still work.

In fact, I think your PR fixes another (hypothetical) problem. Before your PR, `https://example.org?foo=1&amp;bar=1` incorrectly linked to `https://example.org?foo=1&bar=1` (this is probably true for all html entities). With your PR, this problem no longer occurs.

ann0see:
Hmm. Escaping server sided isn‘t really a secure way to fix the issue in my opinion. A quick fix, yes. But not the final way to solve the original problem.

hoffie:
> Escaping server sided isn‘t really a secure way to fix the issue in my opinion. A quick fix, yes. But not the final way to solve the original problem.

I assume you are suggesting to handle it on the target client instead?
I agree it would be better, but this would either require stopping server-side message formatting (i.e. prefixing the user name) or it would require transferring the individual parts of the message (user, text) to the client seperately. The latter would probably equal some (possibly incompatible) change in the protocol, which is probably not desired. :)

Which remaining security issues do you see after merging this PR? "Just" the general UDP-related spoofing things?

ann0see:
> I assume you are suggesting to handle it on the target client instead?

Yes. Exactly. Send e.g. json to the client, validate that and then display the message.

> Which remaining security issues do you see after merging this PR? "Just" the general UDP-related spoofing things?

That's also an issue. I could also imagine that somebody sets a welcome message with "maliscious" html which we do not escape.

hoffie:
> Send e.g. json to the client, validate that and then display the message.

Agreed, but as far as I can see this would require a protocol change/extension. If we were to do that, I guess it would make sense to combine it with some other possible improvements in that area.

> I could also imagine that somebody sets a welcome message with "maliscious" html which we do not escape.

Agreed, but I think we will always have to trust the server to a specific level. Clients should not have to trust the server as much as that it could run arbitrary code or something, but regarding the content I think there is nothing one could do. An attempt at comparing it with some other area: Facebook could always impersonate anyone on their platform because they have the required access. No need for any formatting hacks, cross site scripting or anything. They (or here: the Jamulus server) could just change the content.
So, I think we have to live with this part. I would vote for simply trusting Welcome messages as they can only be changed by server operators. Server operators are always in a position to cause denial of service (just stop the server) or do impersonation (run a modified Jamulus server code or something).

passing:
> I am more concerned that by now a valid use case for allowing (some) HTML has been mentioned by @passing in [#890 (comment)](https://github.com/jamulussoftware/jamulus/issues/890#issuecomment-771754260). I don't know yet how to keep such things still working.

@hoffie 
a different way could be a possibility to send multi-line messages that are monospaced. So maybe some button to switch from the "default mode" (single line, proportional font) to "tab mode" (multiline, monospaced). With that, a converter that creates html would also no longer be needed.

pljones:
OK, can we summarise:
1. This _does_ prevent any client of any version - when connected to a server with this fix - using CSS spoofing in chat
2. This _does not_ address protocol level attacks of a client

Protection from 2 is not, in my view, the scope of this ticket and shouldn't block adoption of the a solution for a known issue.

softins (approved):
Just tried this out and it works fine, as expected.

src/server.cpp
```diff
@@ -1395,7 +1395,7 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int      iCurChanID
         "<font color=""" + sCurColor + """>(" +
         QTime::currentTime().toString ( "hh:mm:ss AP" ) + ") <b>" +
         ChanName.toHtmlEscaped() +
-        "</b></font> " + strChatText;
+        "</b></font> " + strChatText.toHtmlEscaped();
```
ann0see, line 1398: I assume it works as intended. Should I test it again?

softins: I think the change is trivial enough just to approve. I tested it:
image

ann0see (approved):

ann0see:
BTW: is this already documented in the ChangeLog?

softins:
> BTW: is this already documented in the ChangeLog?

It appears not. Historically, who added items to the ChangeLog? The contributor of a change, or Volker?

ann0see:
> If we need contributors to add their own entries, we need to document this in the contribution guidelines.

Agree. Probably the contributing file must be updated too.

ann0see:
Just merged it and documented it in the changelog.

PR #941, opened 2021-02-02
Add feature request template

ann0see:
Same as #924

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature request
+about: Suggest a new idea for Jamulus (not an improvement of an existing function)
+title: ''
+labels: 'feature request'
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+<!--A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] -->
+
+**Describe the solution you'd like**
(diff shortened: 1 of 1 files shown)


gilgongo (approved):

pljones:
Hm.

I think the template should have a guidance section (with "remove this before submitting", of course) that explains what is likely to be _considered_, let alone accepted.

That is - Jamulus does what Jamulus does.  It avoids doing what other software does deliberately to avoid both processing overhead and maintenance overhead.  So if there is another way to solve the problem, that's likely a better solution.

ann0see:
What about:

"Thanks for opening a new feature request! Please keep in mind that Jamulus achieves low latency audio by reducing complexity and following the "Keep it simple and stupid" approach. Adding new features may result in more processing and maintenance overhead so please be sure that your feature can not be achieved by different approaches/software too."

pljones:
Yep, that's the idea.  Maybe add "Make it clear why here." on the end.

ann0see:
I wouldn't stress it too much since it might sound intimidating. 

If we go too far it could sound like:

Thanks for your feature request and all the work you put into thinking about it. Since we don't like to introduce too many new features, you can be quite sure that it will not be implemented.

ann0see:
I've now added your suggestion, but my comment is stil valid ;-).

pljones:
Looks better now :)

pljones (approved):

ann0see:
Great! merged now.

PR #942, opened 2021-02-02
Add the keyboard shortcut simple control the Mute Stream checkbox

jardous:
This simple change allows controlling the mute easily by pressing the M key on the keyboard (when the main window has focus).

The motivation is to use a modified keyboard (with just the M key) as a control pedal to easily mute/unmute myself between listening/playing.

### DIFF
--- a/src/clientdlgbase.ui
+++ b/src/clientdlgbase.ui
@@ -325,6 +325,9 @@
           <property name="text">
            <string>&amp;Mute Myself</string>
           </property>
+          <property name="shortcut">
+           <string>M</string>
+          </property>
          </widget>
         </item>
         <item>


ann0see (approved):
I hope there isn't any other feature which uses the m shortcut.

jardous:
> I hope there isn't any other feature which uses the m shortcut.
Currently this looks like it is the only shortcut used in the whole project

ann0see:
I thought there were some more what happens if you use CTRL+S or CTRL+M?

jardous:
> I thought there were some more what happens if you use CTRL+S or CTRL+M?

Nothing. Those are not defined at all. 
It will however break the "ALT+M" accelerator that is defined in the string "<string>&amp;Mute Myself</string>" and it depends on loaded translation.

ann0see (changes requested):
Ah. So it was "ALT+M". Can we somehow improve this to still allow the alt+m shortcut?

pljones:
If we're going to start having shortcuts like this, I think there's going to be a need to have the ability to disable/enable them in settings / menu option.  I don't mind accelerator keys - Alt-M, etc - as I have control (well, Alt) over them.  Suddenly finding the M key does something might be inconvenient...

ann0see:
@jardous Although I like this feature, would it be possible for your modified keyboard to send alt+m? This feature shouldn't break another one.

jardous:
with a custom keyboard it would be possible. Let's reject this for now.

ann0see:
Ok. So I'll close this.

PR #938, opened 2021-02-01
Add Dutch translation for Windows installer.

jerogee:

PR #953, opened 2021-02-04
Move installer translation scripts to /src/res/translation/wininstaller

### LINKED ISSUE #944: Move installer translation scripts to /src/res/translation/wininstaller
Currently the translators need to jump between /windows/ and /src/res/translation this could be unified.

ann0see:
Fixes #944 

Includes #938


ann0see:
To Jerogee: this PR includes your changes too. I think if you approve it (and another maindeveloper does too) we can merge this.

pljones (approved):

jerogee (approved):
Looks fine to me. I could only not test whether the new relative paths resolve correctly, but in principle they should.

ann0see:
Sorry @pljones for removing the review. Wanted to test something: I assigned a team and would like to know what/who receives a message.

pljones:
Looks like my review stayed.

softins:
I received review request emails for all the requests to maindevelopers.

jerogee:
@ann0see , I only got a "review requested" label added to the issue notification that was already there because I was mentioned/active. Not part of maindevelopers, so my approval does not count for the needed 2 approving reviews:
image

@softins , like to help out and tick this one off? ;-)

windows/installer.nsi
```diff
@@ -75,9 +75,9 @@ Page Custom ASIOCheckInstalled ExitASIOInstalled
 !insertmacro MUI_UNPAGE_FINISH
 
 ; Supported languages configuration
-; Additional languages can be added in the file installerlng.nsi in the translate folder, see https://nsis.sourceforge.io/Examples/Modern%20UI/MultiLanguage.nsi
+; Additional languages can be added in the file installerlng.nsi in the wininstaller folder, see https://nsis.sourceforge.io/Examples/Modern%20UI/MultiLanguage.nsi
 
-!include "translate\installerlng.nsi"
+!include "${ROOT_PATH}\src\res\translation\wininstaller\installerlng.nsi"
```
softins, line 80: Comparing with installerlng.nsi, that uses `..\src\res\...`, so should this be `..\src\...` instead of `${ROOT_PATH}\src\...`?

ann0see: I think ${ROOT_PATH} should also be used in the other file which would then make it an absolute path (as far as I remember these are the only relative paths in the installer).

ann0see:
> I received review request emails for all the requests to maindevelopers. 

Made a few changes to the team in the organisation and I hope now it doesn't assign the whole team.

windows/installer.nsi
softins: I wasn't sure, because all those paths were previously relative too (`translate\...`). I agree consistency would be good, one way or the other.

softins: Looking through the `.nsi` files, it appears the only other relative paths are referring to items not in the tree, such as `.nsh` files. So I would assume they are found via some include path. In view of that, I would agree that it would be safer for `installerlng.nsi` to refer to the individual language `.nsi` files using `${ROOT_PATH}\src\...` instead of `..\src\...`. It would avoid any ambiguity between "relative to the working directory" or "relative to the directory of the including file".

softins (approved):

Issue #888, opened 2021-01-25
The central server presents internal IP


genesisproject2020:
I'm trying to run my own central server together with a few local servers reporting to my central server to make it easy for a group to have "break out sessions". All servers exist locally with local ip addresses.

Problem... With the current configuration alternatives the central server presents its public IP correct the outside but the servers registered will be presented with its internal ip to the outside. In a standard senario a server will register it self to a central server elsewhere and the public will be correctly presented. When building everything in a local environment and then publish it to the world the current options will not work.

Request... Be able to send the IP (or url) that the server should present to be able to build local environments. Each server will require a unique port when the servers use the same external IP.

gene96817:
There is some information missing in this scenario. How are the internal servers addressed from the Internet? (Meaning how are you managing the router, NAT, and firewall functions?) How are the addresses and port mapping determined? This is a scenario where using domain names for each server can help. For the described scenario to be stable, you would have to manually manage your (internal) IP addresses. For long-term convenience, it is best to (find a solution) using DNS and DHCP.

genesisproject2020:
One basic example and use case:

Internet - Firewall with NAT - an internal server
"real ip" - port mapping - internal server with a few Jamulus servers running on different ports
193.x.x.x - port 23000 - 10.0.0.10:23000 (jamulus central server)
193.x.x.x - port 23001 - 10.0.0.10:23001 (jamulus server registered --centalserver 10.0.0.10:23000)
193.x.x.x - port 23002 - 10.0.0.10:23002 (jamulus server registered --centalserver 10.0.0.10:23000)
193.x.x.x - port 23003 - 10.0.0.10:23003 (jamulus server registered --centalserver 10.0.0.10:23000)

The server in the example above the central server will present it self with ip 193.x.x.x and the rest will be presented with 10.0.0.10. There is no need for a dns name (but it makes it easier for the users). Everything is manually handled but I do not see that as a problem for most people.

gene96817:
1- what is your public IP address?
2- How do you want clients on the public network to address these internal addresses?
     You need to create a mapping  between your public IP:port to internal IP:port for every portl in your internal network.
3- for clients on the local network, do you want your users to use those internal (numerical) addresses? You can use host names if you enter the host name/IP address mapping in each computer's host table.

DNS was invented (long ago) to eliminate the need of managing all those host tables.

genesisproject2020:
[...]

PR #954, opened 2021-02-04
Add support for central servers behind NAT

hoffie:
It is currently not possible to run a central server + slaves behind NAT and have external clients connect to such servers (#888). This is due to the fact that a the central server will only know the slave server's internal IP address. It will therefore also hand out this internal IP address to external clients. External clients will then fail to connect.

This PR solves this via two server-side modifications. No protocol changes or client modifications are necessary.
1. Add a new command line flag `--serverpublicip`. This flag can be used by slave servers to specify their external IP address which is reported in the slave server registration message to the central server.
2. Extend the server list logic: If a request by a non-local client comes in, the IP address of the returned slave servers is run through an additional check. If an entry's IP is an internal IP, we know it won't work for the external client. We therefore check the "alternative" IP of the slave server entry (which comes from `--externalserverip`). If this is a public IP, we use this instead.

@genesisproject2020 initially requested this feature and tested the approach taken in this PR. This code is slightly modified/cleaned up (flag name changed), so it would be cool if you could retest and report back in this PR.


src/main.cpp
```diff
@@ -814,6 +830,9 @@ QString UsageArguments ( char **argv )
         "  -u, --numchannels     maximum number of channels\n"
         "  -w, --welcomemessage  welcome message on connect\n"
         "  -z, --startminimized  start minimizied\n"
+        "      --serverpublicip  specify your public IP address when\n"
```
hoffie, line 833: If you have any other suggestions how the flag should be called, let me know.

src/serverlist.cpp
```diff
@@ -443,6 +455,18 @@ void CServerListManager::CentralServerQueryServerList ( const CHostAddress& Inet
                 {
                     vecServerInfo[iIdx].HostAddr = ServerList[iIdx].LHostAddr;
                 }
+                else if ( !NetworkUtil::IsPrivateNetworkIP(InetAddr.InetAddr) && NetworkUtil::IsPrivateNetworkIP(vecServerInfo[iIdx].HostAddr.InetAddr) && !NetworkUtil::IsPrivateNetworkIP(ServerList[iIdx].LHostAddr.InetAddr) )
```
hoffie, line 458: This condition is really long. If you have any ideas how to simplify / change formatting, let me know. There seem to be other instances of such long conditions though.

src/util.cpp
```diff
@@ -1059,6 +1059,29 @@ QString NetworkUtil::FixAddress ( const QString& strAddress )
...
+        QPair<QHostAddress, int>(QHostAddress("172.16.0.0"), 12),
+        QPair<QHostAddress, int>(QHostAddress("192.168.0.0"), 16),
+    };
+
+    int addressesSize = addresses.size();
+    for ( int iIdx = 0; iIdx < addressesSize; iIdx++ ) {
+        if (qhAddr.isInSubnet(addresses[iIdx])) {
+            return true;
+        }
+    }
```
hoffie, line 1081: I have little to zero C++/Qt experience. There may well be better approaches to handle such list-based checks. I'm open to suggestions.

src/util.cpp
```diff
@@ -1059,6 +1059,29 @@ QString NetworkUtil::FixAddress ( const QString& strAddress )
     return strAddress.simplified().replace ( " ", "" );
 }
 
+// Return whether the given HostAdress is within a private IP range
+// as per RFC 1918 & RFC 5735.
+bool NetworkUtil::IsPrivateNetworkIP ( const QHostAddress &qhAddr )
```
hoffie, line 1064: Is `qhAddr` the proper naming for a `QHostAddress`?

src/serverlist.cpp
```diff
@@ -41,7 +42,18 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
     SetCentralServerAddress ( sNCentServAddr );
 
     // set the server internal address, including internal port number
-    SlaveCurLocalHostAddress = CHostAddress( NetworkUtil::GetLocalAddress().InetAddr, iNPortNum );
+    QHostAddress qhaServerPublicIP;
```
hoffie, line 45: Is `qhaServerPublicIP` the proper naming scheme?

ann0see:
Interesting approach.


So there's no way to automatically get the external IP from the slave servers automatically?

Would https://github.com/jamulussoftware/jamulus/blob/373e1f48f3f4e2209bc4dba5e7c6342238f00fb1/src/util.cpp#L1021

work?

I'm not yet sure about security problems which might be introduced by this approach. What would happen if someone intentionally sets a wrong IP?

hoffie:
> Would
> https://github.com/jamulussoftware/jamulus/blob/373e1f48f3f4e2209bc4dba5e7c6342238f00fb1/src/util.cpp#L1021
> work?

I don't think so because this is the code which [runs](https://github.com/jamulussoftware/jamulus/blob/da07b893a2fedfbc0d4907461b194276d244c0e5/src/serverlist.cpp#L44) without my PR. This code can detect the external interface of the machine (i.e. it will yield some 192.168.x, 10.x.y.z, etc. IP), but not the external IP of the network/NAT. To accomplish that automatically, some kind of external "reflector" would be needed such as a web service or using a protocol such as [STUN](https://en.wikipedia.org/wiki/STUN).

> I'm not yet sure about security problems which might be introduced by this approach. What would happen if someone intentionally sets a wrong IP?

Yes, setting an intentionally wrong IP could allow one to create misleading names for existing servers (for example). However, this can not be prevented by not merging the PR. Bad actors could just modify Jamulus themselves or send plain, modified protocol messages using some other means.

gene96817:
There are two possible solutions. (a) a solution where the firewall is configured to use addresses and ports determined by the server and (b ) a solution where the firewall and server are independent on each other.
To make (a) work you have to assign port mappings to both the NAT and the server or the server dynamically picks ports and configures the NAT. If you don't have control of the NAT, then STUN can be used. for (b) to function, you could have all connections initiated by the server. Then the outbound initial packet will carry the port assigned by the NAT and the client will have the IP:port pair to send the reverse traffic.

For the inbound traffic to find the right destination, the mapping of the public IP:port pair and then internal IP:port pair has to be known. If the path is initiated by the inbound traffic, then the port mappings have to be predetermined and communicated to the client. If the path is initiated by the outbound traffic, then the NAT can dynamically create the mapping and the remote client only needs to know the public IP:port pair. You can add layers of complication but this is the fundamental machinery for packets crossing the NAT.

I am not able to follow the discussion to know this principle has been handled properly.

src/util.cpp
atsampson: You can simplify the loop using Qt's [foreach](https://doc.qt.io/qt-5/containers.html#foreach) - something like:

```
foreach ( auto item, addresses ) {
    if ( qhAddress.isInSubset( item ) ) {
        return true;
    }
}
```

(In new code I'd use standard C++'s [range-based for](https://en.cppreference.com/w/cpp/language/range-for), but Jamulus tends to use Qt's facilities.)

It'd also be good to declare `addresses` as static [constexpr](https://en.cppreference.com/w/cpp/language/constexpr), so it doesn't get constructed anew every time this bit of code runs.

src/serverlist.cpp
atsampson: Break it into multiple lines by adding a newline after each `&&` (see `protocol.cpp` for examples of this).

hoffie:
> There are two possible solutions.

I'm not sure I'm following. Are you saying that other techniques should be used instead of the proposed PR? Or are you listing ways for finding out one's own public IP? Or just some generic explanation regarding NAT?

This PR is intended for users who run a central server and one or more slave setups in a private network behind NAT. The NAT has to be set up with proper port forwardings in this case. @genesisproject2020 can probably elaborate what his setup looks like exactly.

src/serverlist.cpp
hoffie: Thanks for the `protocol.cpp` pointer. I have adjusted this part to match the coding style there.

src/util.cpp
hoffie: > You can simplify the loop using Qt's foreach - something like:

Thanks, done!

> It'd also be good to declare addresses as static constexpr, so it doesn't get constructed anew every time this bit of code runs.

Hrm, I understand the intention (and was even thinking about such a thing), but `constexpr` doesn't seem to work (`src/util.cpp:1068:54: error: the type 'const QList<QPair<QHostAddress, int> >' of 'constexpr' variable 'addresses' is not literal`). Per the docs it only works on literals (which I read as: int, char, etc.?) while I'm using a QList. I have now changed it to just `static`. Wouldn't this accomplish the same?

genesisproject2020:
Works great!

genesisproject2020:
@ann0see I have a hard time to see that the solution opens upp for a security problem. If someone would like to redirect traffic there is a lot of ways to do that. If someone would redirect the traffic to the "wrong" IP where a server does not answer it will just not work.

The solution makes it possible to build an internal solution and easily make it available to the outside world. I've created a main central server with 4 "sub servers", all within my local network, to be used for a choir to use when practicing so the each group can be by them selves when needed and all see where they are connected. I've asked my groups on my servers to test it and I've got a great response.

@gene96817 I'm sorry but as I've said before I have a hard time to follow you in this case.

ann0see:
> I have a hard time to see that the solution opens upp for a security problem. If someone would like to redirect traffic there is a lot of ways todo that.

Agree. Probably I'm over cautious (there's no encryption either). I don't object this idea, please don't misunderstand me. 

Nevertheless, I'd like to hear a few other opinions/ideas since the best way would be automatic external IP detection.

hoffie:
> the best way would be automatic external IP detection

I certainly think this is possible, but I tried to adhere to KISS principles and attempted to make this as simple as possible. ;)
As said, we could get the external IP from somewhere, for example by using STUN. For reference, here is a quick example from someone who has implemented this in Ruby: https://gist.github.com/bmuller/a1525bd8f7799812867f
We would either have to implement this ourselves or find/include a library which does this.
This would probably also start discussions about auto-detecting changed IP addresses (reconnect of dialup connections, etc.). This is all possible, but would require code which has to be maintained. That's why I went the simple route first. If demand comes up, we could still extend this.

By the way, I do not think such autodetection is possible without using external infrastructure. This either implies relying on third parties (e.g. Google or some other public STUN server provider) or running something on jamulus.io.

genesisproject2020 (approved):

ann0see:
Ok. Going to approve it now (I didn’t  test the myself but I assume since the code compiles everything is ok). Please also add a short explanation of this feature to the documentation (open a PR to the command line page) and the ChangeLog file here.

ann0see (approved):

gene96817:
@genesisproject2020  We are having understanding each other. Since you have this working, my comments don't apply. I just don't understand why this is working. There is probably something "obvious" that is not explained in how this is working.

hoffie:
@ann0see:
> Please also add a short explanation of this feature to the documentation (open a PR to the command line page) and the ChangeLog file here.

Both done.



@gene96817:
> I just don't understand why this is working. There is probably something "obvious" that is not explained in how this is working.

I think you may have been talking about the general network aspects (NAT, port forwarding, etc.). Proper NAT configuration is the basis for all this. The PR is about the application-level protocol changes which are required in addition. It's not meant to be an alternative. Maybe this clears up some confusion.

src/serverlist.cpp
pljones: There generally isn't consistent Microsoft-style class name initial prefixing of variable names.  So these aren't inherently "right" (and I bristle against them) but there are enough type-prefixed names in the code to make them not inherently "wrong", either.

src/protocol.cpp
```diff
@@ -303,6 +303,9 @@ CONNECTION LESS MESSAGES
       NOTE: In the PROTMESSID_CLM_SERVER_LIST list, this field will be empty
       as only the initial IP address should be used by the client.  Where
       necessary, that value will contain the server internal address.
+      When running a central server and a slave server behind the same NAT,
+      this field is used the other way round: It will contain the public
+      IP in this case which will be served to clients from the Internet.
```
pljones, line 308: There's a certain natural flow to that which I like :)

pljones (approved):

gilgongo:
@hoffie Just a question about the documentation for this, which I note you have added [here](https://github.com/jamulussoftware/jamuluswebsite/commit/721a04e0173ba3b713613ee695f0d5ef73409dbd#diff-20d394e0a707647386e4afab7787878f6876e5e7b01f1d06e4ab9fa01bc4b0bf).

Your explanation introduces the concept of a "slave server", which we don't use anywhere else, and isn't explained. If I understand correctly and you mean a "normal" server which is being advertised by the central server, then would the explanation of the serverpublicip option be as follows? 


> When running a public central server behind a NAT firewall on a private network, use the `--serverpublicip`  option to specify the public IP address of the server(s) being listed by your central server. This is necessarily to allow clients on the public Internet to connect to them via NAT. Note that for the servers using this option, you will still need proper port forwarding in your router/firewall.

Also as a general point, I think it might be better to move this explanation to the other docs for central servers [here](https://jamulus.io/wiki/Choosing-a-Server-Type#Central) and link to that from the configuration page, as we do for other things. This is to avoid duplication in future, and cluttering up the configuration page. Would it be OK if I did that?

hoffie:
> Your explanation introduces the concept of a "slave server", which we don't use anywhere else, and isn't explained.

Hrm, I'm pretty sure I haven't invented it. Grepping through the code I see that it is being used in multiple places such as the `ChangeLog` and `src/serverlist.cpp`. This is probably where I've got this from. But I assume that you are mainly referring to documentation where it isn't used.

> If I understand correctly and you mean a "normal" server which is being advertised by the central server, then would the explanation of the serverpublicip option be as follows?

Yes, this is what is meant and your proposal sounds fine.

> Also as a general point, I think it might be better to move this explanation to the other docs for central servers here and link to that from the configuration page, as we do for other things. This is to avoid duplication in future, and cluttering up the configuration page. Would it be OK if I did that?

Sure.

> Oh and also BTW, please don't put translations onto the changes branch. That's just for the English, which then gets merged to translation before site release. I'll back that out if that's OK with you.

I thought about this and read the [CONTRIBUTING.md](https://github.com/jamulussoftware/jamuluswebsite/blob/changes/CONTRIBUTING.md) document multiple times, but could not find any information whether it would be a good idea or not. Feel free to back it out. Maybe I'll raise a PR for adding "Only change English docs in changes branch" later.

pljones:
One point on "central server", "slave server", "normal server", etc.  I've been trying to encourage "server list" and "server list server" rather than "central server" -- as there's nothing central about having seven server lists, plus as many "custom" lists as you like.  So a server that hosts a server list is a "server list server".  What it's doing is "hosting a server list".  Then all the servers on that list are just "servers on that server list".  That eliminates jargon.

gilgongo:
>  I've been trying to encourage "server list" and "server list server"

"Server list server" is a bit .... palindromic (in English at least). I think "directory server" has been suggested before too. BTW I take it "server list" refers to the thing in the UI that the server provides to the client, in which case I don't think we need to be too strict about the usage there since the context will explain it ("A list of servers will appear", "Select from the list"). The main confusion is between servers and central servers, where context sometimes doesn't help ("To connect to the server", "The address of the server").

Also, we can easily change all this in the docs, but I don't know if might cause confusion in the code (which has been the main objection in the past)?

gilgongo:
@hoffie Ah I see what you mean about the contributing guidelines:

>  A fix which needs to be translated (e.g. if you want to make a sentence or paragraph clearer) --> Commit to changes since others will need to translate it.

That's not clear :-) I think it should say "a fix IN THE ENGLISH which needs to be translated". Otherwise all the language files get out of sync with the "root" English file, unless you are volunteering to translate the fix for ALL languages ;-)

pljones:
> I think "directory server" has been suggested before too.

"Something needs to be done" is really as far as I'd got -- "central" is wrong - was only right whilst there was only one.  I'd rather the term was as non-technical as it can be: I guess people still understand "directory" to be something you look things up in?

gilgongo:
> I guess people still understand "directory" to be something you look things up in?

I think that's a fairly safe assumption. Might we even retire the "server" word from it and just call them "Directories"? So "Private server", "Public server", "Directory"? Replacing "Central Server" with "Directory" in the docs actually reads quite well and allows the word "Server" to be disambiguated in all contexts.

pljones:
The only issue is that a "directory" _server_ is still able to function as a public or private Jamulus _server_, so I wouldn't drop _server_ entirely.  It's essentially an additional function on top of being a Jamulus server.

gilgongo:
Well, by that logic we'd be calling clients "client/servers" since you can use the `-s` option. I think referring simply to a "Directory" gets it out of the hair of the vast majority of users have no in interest in central servers, and therefore disambiguates things nicely.

pljones:
A client without "-s" can't be used as a server.  You can't run a server at all without "-s" -- "-e localhost" isn't allowed for a client.  So you have to be a server to have "-e localhost".

"-e" just whether you're server is a list / directory server ("-e localhost"), a registered server ("-e something" -- because you can register with a custom directory and hence not be pubic just because you're registered) or an unregistered server (no "-e").

gilgongo:
> A client without "-s" can't be used as a server. You can't run a server at all without "-s" -- "-e localhost" isn't allowed for a client. So you have to be a server to have "-e localhost".

Yes, but I don't see how any of that is conveyed by calling a thing a "client" and more than `-e` is conveyed by the term "server". All I'm saying is that by naming two things "... server" it makes the docs less clear perhaps. 

We could adopt a long form of "Directory Server" (for headings or UI labels, say) and a short form of just "Directory" (eg "Can you connect to any of the directories?", "When running a directory." etc. )

Again I have no idea if this will cause confusion in the code.

pljones:
I'm saying that _anything_ with `-s` is a server.  The value of `-e` determine what kind of server it is.  Hosting a directory is an additional function provided by directory servers.

So, I'd use the phrase "Can you display any of the directories?" (i.e. the results of querying directory servers) - they're different things: the server is what supplies the directory.  The same server can also be used for jamming, of course, so connecting to the directory server could be for that.

gilgongo:
> The docs need to make it clear that different kinds of server 

Every relevant page in the docs has a header asking you whether you've read the [explanation](https://jamulus.io/wiki/Choosing-a-Server-Type) before continuing. Short of throwing up a server quiz and allowing only those supplying correct answers to read the next page, I highly doubt we could be any clearer :-) 

> it's currently obvious many people get confused about this,

Which is why (given the above) my hypothesis is that calling two things a "server" may not helpful. 

But I think the ability simply to refer to "directory" when we used to say "central" would be fine. So a straight find/replace of that term in the docs would be OK I think. That would mean we'd be saying "directory server" in the docs, but also could use just "directory" in casual conversation without mishap (eg "Can you display any of the directories?").

pljones:
> Do we need to display the port numbers of the directory servers in the docs here? That is, if I set jamulusrock.jamulus.io on my command line, will connections be forwarded to 22424 automatically or do I need to explicitly add :22424 to the end of the host name? If so, can we just remove those ports from that table?

Host name resolves to an IP number.  On its own, that will always get you the default port number of 22124.  So yes, you always need to supply the port number when it's not 22124, i.e. not the default port number.

softins:
I'm a bit late to the party here, but have only just had time to test multiple servers behind NAT.

I tested with the 3.6.2 release. I set up inbound port forwarding for ports 22120-22123 to my RPi.

I then created a central server on my RPi on port 22120. It was visible to Explorer on my public IP port 22120.

I then created three more servers on the RPi, using ports 22121-22123, and specifying that they should register with my PUBLIC IP port 22120 (not my LAN IP, nor 127.0.0.1). They all appeared correctly in Explorer.

I then started up a client on another computer on the LAN, and set the custom central server to my PUBLIC IP port 22120. I was able to connect successfully to any of the four servers via the acustom central server.

All of this worked without the new feature, so it leads me to wonder why it was needed and when it would actually be used?

hoffie:
> I then created three more servers on the RPi, using ports 22121-22123, and specifying that they should register with my PUBLIC IP port 22120 (not my LAN IP, nor 127.0.0.1). They all appeared correctly in Explorer.
> 
> I then started up a client on another computer on the LAN, and set the custom central server to my PUBLIC IP port 22120. I was able to connect successfully to any of the four servers via the custom central server.

Hrm, that's indeed one of the proposed solutions I've posted here: https://github.com/jamulussoftware/jamulus/issues/888#issuecomment-768171934
I've just noticed that the initial requester (@genesisproject2020) did not comment on that part (?). @genesisproject2020 Had you tested this? I ran a quick test and couldn't get it to work, but that might also be related to my specific network setup/NAT environment.

> All of this worked without the new feature, so it leads me to wonder why it was needed and when it would actually be used?

If it turns out that the requested behavior is possible without any code changes, then, it should probably be discussed if the new code should be removed again. Would be sad to say as there's been some effort not only by you and me but also by @genesisproject2020, @pljones, @gilgongo and others, but I think it would be in the interest of the project to only add code which is really needed.

I suggest waiting for some feedback by @genesisproject2020.

softins:
@hoffie thanks for your comments! I can imagine it might be possible that some routers do not allow internal hosts to route to the external address for forwarding back in. In that situation, this new option may help. Just to confirm, it would only be needed on the central server, not on the other local servers registering to it?

hoffie:
> I can imagine it might be possible that some routers do not allow internal hosts to route to the external address for forwarding back in.

Yes, might be possible. In my case it was Linux/iptables, so guess it would be possible to hack this somehow. Let's wait what @genesisproject2020 says.

> Just to confirm, it would only be needed on the central server, not on the other local servers registering to it?

Not quite. If you need this feature (i.e. you run a central server and regular servers behind NAT and want public clients to connect via the central server), then both the central server and the regular server need to run the new code. The central server does not need any configuration, it will auto-detect if the new code branch should be triggered. The regular server needs the new `--serverpublicip` parameter.

If the central server uses the old code, nothing bad will happen, it just won't work (as it used to).
If the regular server uses the old code, `--serverpublicip` will not work and startup will fail.

genesisproject2020:
Some firewalls and broadband routers have support for hairpinning (using external IP to communicate within a NAT environment). Mostly this is not supported and therefore another solution is needed to be able to present the external IP.

Another scenario where a solution to choose the presenting IP is if you have more then one IP on the external interface. The solution implemented to get the external IP will probably choose the first IP or the NAT IP.

The solution to override the found external IP solves both scenarios in an easy and effective way. 

I've used the code for a while with no complications. As said before, the updated code has to be on the directory instance as well as the server instances that connect to the directory server. The clients can be whatever version so it's backward compatible.

gene96817:
A usual way to avoid the need for hairpinning, is to use the router's DNS service. Then internal network devices can use external hostnames and get internal IP addresses. This allows all clients to use the same hostnames; external clients get external IP addresses from an external DNS server and internal clients get internal IP addresses from the internal (in the router) DNS server. (The internal DNS server table is very short. Only the internal host needs an entry. All other domain names are proxied to an external DNS server.)

genesisproject2020:
The suggestion to use an internal DNS will present the internal IP to the directory server and will not work.

There will probably be a long row of techie solutions to solve this but none of them will be an easy way for someone to do. The solution @hoffie created overcomes the need of being a techie and having special equipment. I hope we can leave it as it is.

gene96817:
> The suggestion to use an internal DNS will present the internal IP to the directory server and will not work.

Why?

softins:
> > The suggestion to use an internal DNS will present the internal IP to the directory server and will not work.
> 
> Why?

Because the issue is not DNS, whether split horizon, independent internal/external or otherwise.

The issue here is purely related to IP addresses and NAT routing. If a router supports hairpinning, then internal hosts A and B can communicate with each other using the external IP and the appropriate port numbers. If they are just commnicating with each other, there is no point and it’s wasteful, but if host A is a central server and needs to publicise host B to the outside world, the fact that B registered from its external IP, by hairpinning, means that the correct public IP and port get publicised. This works in all versions of Jamulus without needing `--serverpublicip`.

If the router does not support hairpinning, host B is only able to register with host A directly on its LAN IP address. In that case, the central server on host A would list the wrong IP for B when sending it to an external client. The new option `--serverpublicip` is intended to solve this particular problem, by telling the central server the external IP to present instead of the LAN IP.

gene96817:
oooo.... are you saying that a core problem is many Jamulus servers do not have domain names?

softins:
> oooo.... are you saying that a core problem is many Jamulus servers do not have domain names?

No, I’m not saying that’s a problem at all. I’m saying it isn’t relevant to this issue.

PR #960, opened 2021-02-06
Improve multi threading

menzels:
Hi Jamulus People,

Running jamulus for my choir on Hetzner servers, i stumbled on some multi threading problems.
Connecting more than about 17 people i got really high ping and i saw the cores were not used uniformly.

This PR fixes the issue i had and distributes load equally among all available cores.

I also removed multithreading for the decode step, because profiling showed that this code only takes about 6-7% of jamulus' cpu time (at 21 connected clients)

This is my first work on the jamulus project, feel free to discuss and suggest changes.


ann0see:
@menzels Thank you for your first PR. Improving multi threading is great! I'm not an expert on this, but this sounds great! Could you please have a look at why the translation files get deleted with this PR?

menzels:
> 
> 
> @menzels Thank you for your first PR. Improving multi threading is great! I'm not an expert on this, but this sounds great! Could you please have a look at why the translation files get deleted with this PR?

@ann0see i am not sure why translation files are modified here, this was not intended, i will revert these changes.

softins:
I will try and review this tomorrow. Thanks for the PR!

atsampson:
The change to encoding looks sensible.

I'm not sure about the change to decoding - 6% is enough that it's probably worth parallelising, and there are people who run servers with very large numbers of clients. Is it substantially worse than the original code for small numbers of clients if you parallelise it the same way you've parallelised encoding?

It'd be good to put the rationale from the pull request into the commit message, for people looking at this change in the Git history later.

menzels:
> I'm not sure about the change to decoding - 6% is enough that it's probably worth parallelising, and there are people who run servers with very large numbers of clients. Is it substantially worse than the original code for small numbers of clients if you parallelise it the same way you've parallelised encoding?

Thanks for the comment. 
I did not think about big numbers of clients, for <50 parallelizing the decoding won't make a difference i think.
But the drawback of parallelizing, is also not that big, ill add that again.

src/server.cpp
```diff
@@ -411,10 +411,20 @@ CServer::CServer ( const int          iNewMaxNumChan,
...
-    // introduced by kraney (#653): "increased the size of the thread pool"
-    if ( bUseMultithreading )
+    int iAvailableCores = QThread::idealThreadCount();
+
+    // setup QThreadPool if multithreading is active and possible
+    if ( bUseMultithreading && iAvailableCores > 1 )
     {
-        QThreadPool::globalInstance()->setMaxThreadCount ( QThread::idealThreadCount() * 4 );
+        // set maximum thread count to available -1 to keep one thread for the main event loop
+        iMaxNumThreads = iAvailableCores - 1;
```
pljones, line 420: The ` - 1` may vary.  The GUI adds a thread, as does the recorder.  Whilst neither of those is real time process priority as the audio threads should be, they may need to be taken into account.

src/server.cpp
```diff
@@ -1304,26 +1314,26 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
...
+        // OPUS encoding
+        if ( pCurOpusEncoder != nullptr )
         {
-            // OPUS encoding
-            if ( pCurOpusEncoder != nullptr )
-            {
-// TODO find a better place than this: the setting does not change all the time so for speed
-//      optimization it would be better to set it only if the network frame size is changed
-opus_custom_encoder_ctl ( pCurOpusEncoder, OPUS_SET_BITRATE ( CalcBitRateBitsPerSecFromCodedBytes ( iCeltNumCodedBytes, iClientFrameSizeSamples ) ) );
+            // TODO find a better place than this: the setting does not change all the time so for speed
```
pljones, line 1320: Please revert this spacing change.  It was deliberately formatted the way it was.

src/server.cpp
```diff
@@ -1304,26 +1314,26 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
...
+            //      optimization it would be better to set it only if the network frame size is changed
+            opus_custom_encoder_ctl ( pCurOpusEncoder, OPUS_SET_BITRATE ( CalcBitRateBitsPerSecFromCodedBytes ( iCeltNumCodedBytes, iClientFrameSizeSamples ) ) );
 
+            for ( int iB = 0; iB < vecNumFrameSizeConvBlocks[iChanCnt]; iB++ )
```
pljones, line 1324: Can `int iB` now be changed to the same data type as the widest of `SYSTEM_FRAME_SIZE_SAMPLES` and `vecNumAudioChannels[iChanCnt]`, to avoid the overflow warning below?

pljones:
@kraney in #653 is referenced in part of the code changed, so input from Kris might be useful, too.

kraney:
Since this strategy tunes the number of blocks to the number of CPU cores (-1) it should be more efficient / more fairly allocated - provided the computation is truly CPU bound and doesn't have any system calls or IO waits in it. If those are present, where this would simply block and wait, in the other case having more threads than cores, the waiting thread can go to sleep while a different thread makes progress. OTOH that thread swap comes at the cost of cache consistency.

That reasoning suggests this might be a good idea for DecodeReceiveDataBlocks, but probably not so hot for MixEncodeTransmitDataBlocks. I haven't run any experiments to validate it, just speaking academically at this point.

src/server.cpp
```diff
@@ -818,9 +828,11 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
...
         {
             // processing with multithreading
-// TODO optimization of the MTBlockSize value
-            const int iMTBlockSize = 10; // every 10 users a new thread is created
-            const int iNumBlocks   = ( iNumClients - 1 ) / iMTBlockSize + 1;
+            // spread work equally among available threads
+            // since decoding is a much lighter workload, only create new tasks for every 10 clients
+            // measurements by menzels suggest it is ~6% (decode) vs 60% (encode) of process cpu time
+            const int iNumBlocks   = std::min(iNumClients / 10 + 1, iMaxNumThreads);
+            const int iMTBlockSize = (iNumClients - 1) / iNumBlocks + 1;
```
softins, line 835: I think this is probably not correct. If `iNumClients` is 1, then `iMTBlockSize` will be 0, and further down (not in the diff), `iStopChanCnt` will be -1, which doesn't sound right to me.

I'm not sure the reason for swapping `iNumBlocks` and `iMTBlockSize`.

src/server.cpp
```diff
@@ -901,11 +913,9 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
...
         if ( bUseMultithreading )
         {
-            // introduced by kraney (#653): each thread must complete within the 1 or 2ms time budget for the timer
-// TODO determine at startup by running a small benchmark
-            const int iMaximumMixOpsInTimeBudget = 500; // approximate limit as observed on GCP e2-standard instance
-            const int iMTBlockSize = iMaximumMixOpsInTimeBudget / iNumClients; // number of ops = block size * total number of clients
-            const int iNumBlocks   = ( iNumClients - 1 ) / iMTBlockSize + 1;
+            // spread work equally among available threads
+            const int iNumBlocks   = std::min(iNumClients, iMaxNumThreads);
+            const int iMTBlockSize = (iNumClients - 1) / iNumBlocks + 1;
```
softins, line 918: Same comment as for line 835 above.

src/server.cpp
```diff
@@ -1304,26 +1314,26 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
...
-            }
+                                            &vecsSendData[iB * SYSTEM_FRAME_SIZE_SAMPLES * vecNumAudioChannels[iChanCnt]],
+                                            iClientFrameSizeSamples,
+                                            &vecvecbyCodedData[iChanCnt][0],
+                                            iCeltNumCodedBytes );
 
             // send separate mix to current clients
             vecChannels[iCurChanID].PrepAndSendPacket ( &Socket,
                                                         vecvecbyCodedData[iChanCnt],
                                                         iCeltNumCodedBytes );
```
softins, line 1335: This statement need re-indenting due to the change of nesting.

softins (changes requested):

src/server.cpp
softins: Sorry, I inferred brackets where there were none. My mistake.

menzels:
> 
> 
> Since this strategy tunes the number of blocks to the number of CPU cores (-1) it should be more efficient / more fairly allocated - provided the computation is truly CPU bound and doesn't have any system calls or IO waits in it. If those are present, where this would simply block and wait, in the other case having more threads than cores, the waiting thread can go to sleep while a different thread makes progress. OTOH that thread swap comes at the cost of cache consistency.
> 
> That reasoning suggests this might be a good idea for DecodeReceiveDataBlocks, but probably not so hot for MixEncodeTransmitDataBlocks. I haven't run any experiments to validate it, just speaking academically at this point.

thank you for this comment, i will test if tuning the threads above the core count makes a difference. 
it really seems like the threads are waiting on sending to the socket, so that might be the case.
in the end though the whole process has to wait until every thread did send to socket, so i dont know if the difference is really notable. i will check.

src/server.cpp
menzels: I think one core is enough for gui, recording, and such. 
I did more tests with up to 80 clients and it works quite nicely, without gui though in my case.

menzels:
I did more testing and found that a simple c++11 threadpool implementation adds less overhead compared to qtconcurrent, i added this implementation for your consideration.
Its ok to remove that again if you see issues with it.

What do you think about adding the possibility to set the number of decode/encode threads as command line parameter? While providing a sensible default.

pljones:
> What do you think about adding the possibility to set the number of decode/encode threads as command line parameter? While providing a sensible default.

It might be worthwhile for test purposes - i.e. on a development branch - but I'd shy away from this level of detailed control in the code.  I'd expect, once we have determined a reasonable approach, to be able to pick the right number from available information.

dingodoppelt:
I tried to test this code a few times now and couldn't get it to work. Is there anything I might have overlooked? I don't get any sound but the overall delay is about twice as high as on the regular server and the buffer LED stays red constantly. Same thing on the cloud server I tested it on. I'm very interested in getting this to work since we found that at > 18 clients our cloud servers show the same behaviour as described by @menzels 
For my tests I launched the server with -s -n -F -T

menzels:
> I tried to test this code a few times now and couldn't get it to work. Is there anything I might have overlooked? I don't get any sound but the overall delay is about twice as high as on the regular server and the buffer LED stays red constantly. Same thing on the cloud server I tested it on. I'm very interested in getting this to work since we found that at > 18 clients our cloud servers show the same behaviour as described by @menzels
> For my tests I launched the server with -s -n -F -T (compiled with "nosound headless")

We used a build from this branch for our last choir rehersal and it worked great. i have no idea right now what could be wrong with yours. i will test sunday with the artifact from the auto build.

hoffie:
> I'd expect, once we have determined a reasonable approach, to be able to pick the right number from available information.

What about running multiple Jamulus servers (or other time-critical software) on the same machine? I'd be in favor of choosing a sensible default and making it still possible to override the number of usable cores.

Side note: I did not have any troubles with ~25 users with the current code, but I'd welcome any optimizations which make this more stable/scalable.

gene96817:
what is the related setting being used for the choruses with over 100 voices?

dingodoppelt:
> We used a build from this branch for our last choir rehersal and it worked great. i have no idea right now what could be wrong with yours. i will test sunday with the artifact from the auto build.

@menzels 
I just tried again on my 4-vcore cloud server (running two other jamulus servers at the same time on an ubuntu 20.04.1 lowlatency kernel) and it shows the same behaviour as before: no sound but higher delay and a buffer LED constantly red.
The other test was on my local machine: Not able to connect at all
Has anybody else tested the code yet and may be able to reproduce my issues?

pieterbos:
@dingodoppelt I compiled this branch on Armbian 21.02.2 Focal, started a server and connected, no problems whatsoever so far.

menzels:
We just had rehearsal again and i could not get my code to run this time. really strange, there was just no sound.
I will look into this more this week.

menzels:
there were some bad errors in my code. 
now it should work again.

Jamulus.pro
```diff
@@ -402,6 +402,7 @@ HEADERS += src/buffer.h \
     src/global.h \
     src/protocol.h \
     src/recorder/jamcontroller.h \
+    src/ThreadPool.h \
```
pljones, line 405: Please keep to code style.  All source file names are lower case (or should be).

src/ThreadPool.h
```diff
@@ -0,0 +1,99 @@
...
+#include <queue>
+#include <memory>
+#include <thread>
+#include <mutex>
+#include <condition_variable>
+#include <future>
+#include <functional>
+#include <stdexcept>
+
+class ThreadPool {
```
pljones, line 14: Could you please review this whole file for code style.

src/server.cpp
```diff
@@ -22,7 +22,8 @@
  *
 \******************************************************************************/
 
-#include "server.h"
+#include "server.h"
```
pljones, line 25: You've added trailing whitespace.  Could you please ensure you avoid this.

src/server.cpp
```diff
@@ -413,10 +414,25 @@ CServer::CServer ( const int          iNewMaxNumChan,
...
     }
 
-    // introduced by kraney (#653): "increased the size of the thread pool"
-    if ( bUseMultithreading )
-    {
-        QThreadPool::globalInstance()->setMaxThreadCount ( QThread::idealThreadCount() * 4 );
+    int iAvailableCores = QThread::idealThreadCount();
+
+    // setup QThreadPool if multithreading is active and possible
+    if(bUseMultithreading) {
```
pljones, line 420: Please stick to the project code style conventions.

src/server.cpp
```diff
@@ -552,12 +568,12 @@ inline void CServer::connectChannelSignalsToServerSlots()
 template<>
 inline void CServer::connectChannelSignalsToServerSlots<0>() {}
 
-void CServer::CreateAndSendJitBufMessage ( const int iCurChanID,
+void CServer::CreateAndSendJitBufMessage ( const int iCurChanID,
```
pljones, line 571: More trailing whitespace.  Could you review all the changes before committing to avoid this, please.

src/server.cpp
```diff
@@ -831,16 +845,23 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
...
 
-                FutureSynchronizer.addFuture ( QtConcurrent::run ( this,
-                                                                   &CServer::DecodeReceiveDataBlocks,
-                                                                   iStartChanCnt,
-                                                                   iStopChanCnt,
-                                                                   iNumClients ) );
+                Futures.push_back ( tpThreadPool->enqueue ( CServer::DecodeReceiveDataBlocks, 
+                                                    this, iStartChanCnt, iStopChanCnt, iNumClients ) );
+
+                // FutureSynchronizer.addFuture ( QtConcurrent::run ( &CServer::DecodeReceiveDataBlocks,
```
pljones, line 851: If this is a possible alternative, please explain that.  Otherwise if it's redundant code, please remove.

pljones, line 851: Please delete code no longer used.  Git allows it to be recovered easily, if needed.  Commenting it out makes the code unclear.

src/server.cpp
```diff
@@ -938,25 +959,27 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
     }
 }
 
-void CServer::DecodeReceiveDataBlocks ( const int iStartChanCnt,
+void CServer::DecodeReceiveDataBlocks ( CServer*  sServer,
```
pljones, line 962: Will these pointers (here and the other calls) be thread-safe?  The loop above uses locking around its scope but it then kicks off multiple threads, passing each a reference to the same instance.  Have all the server methods used been checked for thread safety?

src/server.cpp
```diff
@@ -1305,26 +1328,26 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
...
 opus_custom_encoder_ctl ( pCurOpusEncoder, OPUS_SET_BITRATE ( CalcBitRateBitsPerSecFromCodedBytes ( iCeltNumCodedBytes, iClientFrameSizeSamples ) ) );
 
+            for ( size_t iB = 0; iB < vecNumFrameSizeConvBlocks[iChanCnt]; iB++ )
+            {
                 iUnused = opus_custom_encode ( pCurOpusEncoder,
-                                               &vecsSendData[iB * SYSTEM_FRAME_SIZE_SAMPLES * vecNumAudioChannels[iChanCnt]],
-                                               iClientFrameSizeSamples,
-                                               &vecvecbyCodedData[iChanCnt][0],
-                                               iCeltNumCodedBytes );
+                                            &vecsSendData[iB * SYSTEM_FRAME_SIZE_SAMPLES * vecNumAudioChannels[iChanCnt]],
```
pljones, line 1341: The code alignment looks wrong here.

dingodoppelt:
> now it should work again.

it does! thanks for your fixes :) i'm running the code on my vserver and it really improved things. thank you very much!

Helondeth:
I'm running your code on my server for choir rehearsal and it works perfectly for me :)

softins:
Hi @menzels, I want to resurrect this code so it can be merged into the master branch. I have cloned your branch into my own repo, rebased it to the latest master, and have done some of the cleanup mentioned in the comments above. You can see what I've done at https://github.com/softins/jamulus/commits/improve-multi-thread

I also made a [comment](https://github.com/jamulussoftware/jamulus/pull/960#discussion_r630324969) above, wondering about the use of `-1` when calculating the number of threads to use. I'd be interested in your opinion on that.

How would you like to proceed with this? There are several options:
1. You could pull from my branch;
2. As a maintainer, I could force-push back to your branch, to update this PR;
3. I could raise a replacement PR from my branch (which still contains all your commits attributed correctly to you).

menzels:
> 
> 
> Hi @menzels, I want to resurrect this code so it can be merged into the master branch. I have cloned your branch into my own repo, rebased it to the latest master, and have done some of the cleanup mentioned in the comments above. You can see what I've done at https://github.com/softins/jamulus/commits/improve-multi-thread
> 
> I also made a [comment](https://github.com/jamulussoftware/jamulus/pull/960#discussion_r630324969) above, wondering about the use of `-1` when calculating the number of threads to use. I'd be interested in your opinion on that.
> 
> How would you like to proceed with this? There are several options:
> 
>     1. You could pull from my branch;
> 
>     2. As a maintainer, I could force-push back to your branch, to update this PR;
> 
>     3. I could raise a replacement PR from my branch (which still contains all your commits attributed correctly to you).

Hi, thank you for taking this, i unfortunately did not have the time yet to finish this work. i will take a look at your branch. Feel free to force push to my branch or open your own PR as you like.

I like your reasoning about the -1 topic, feel free to remove it.

softins:
> Hi, thank you for taking this, I unfortunately did not have the time yet to finish this work.

No problem. What work do you feel is still outstanding on this? Or was it just responding to the review comments?

> I will take a look at your branch. Feel free to force push to my branch or open your own PR as you like.

I force-pushed to your own branch, to retain the original PR number and conversation history. Please have a look when you have the time.

> I like your reasoning about the -1 topic, feel free to remove it.

Thanks. I also removed the check for `>= 10`, as it is just as good, if not better, to share out the work between the threads from the outset. I did various tests locally to observe the behaviour with different numbers of clients.

With the number of threads set to the actual number of cores, the main audio processing is nicely divided between those threads, with the other threads just taking up a small amount.

src/threadpool.h
```diff
@@ -0,0 +1,123 @@
...
+#include <queue>
+#include <memory>
+#include <thread>
+#include <mutex>
+#include <condition_variable>
+#include <future>
+#include <functional>
+#include <stdexcept>
+
+class ThreadPool {
```
hoffie, line 38:
```suggestion
class ThreadPool
{
```

As this is a completely new file, maybe just run `clang-format` with the config from the relevant PR on it?

hoffie (commented):
I just had a very quick look at the code and haven't yet tried to fully follow it through.

I have successfully tested the current code (2ba6d567ce13d7deb48993ef39e402d9e1d40dfe) with 10 people and multithreading enabled today without any negative side effects.

As per @softins's comment in the chat, we may try to get it into 3.8.0 if it is ready until ~Sunday (where rc1 is supposed to be tagged and no more code changes can get in).

src/threadpool.h
softins: Probably, but I would do it along with all the others, when it's time, rather than on its own in advance.

Maybe also `class ThreadPool` should be `class CThreadPool` for consistency with the rest of Jamulus.

hoffie (commented):
Some more comments as I'm starting to dive into the relevant logic.

src/server.cpp
```diff
@@ -966,25 +971,27 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
     }
 }
 
-void CServer::DecodeReceiveDataBlocks ( const int iStartChanCnt,
+void CServer::DecodeReceiveDataBlocks ( CServer*  sServer,
```
pljones, line 974: Isn't the general naming convention for a pointer `pServer`?

src/server.cpp
```diff
@@ -966,25 +971,27 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
...
     // loop over all channels in the current block, needed for multithreading support
     for ( int iChanCnt = iStartChanCnt; iChanCnt <= iStopChanCnt; iChanCnt++ )
     {
-        DecodeReceiveData ( iChanCnt, iNumClients );
+        sServer->DecodeReceiveData ( iChanCnt, iNumClients );
     }
 }
 
-void CServer::MixEncodeTransmitDataBlocks ( const int iStartChanCnt,
+void CServer::MixEncodeTransmitDataBlocks ( CServer*  sServer,
```
pljones, line 986: And `pServer` again?

src/server.cpp
```diff
@@ -1390,26 +1397,26 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
...
 opus_custom_encoder_ctl ( pCurOpusEncoder, OPUS_SET_BITRATE ( CalcBitRateBitsPerSecFromCodedBytes ( iCeltNumCodedBytes, iClientFrameSizeSamples ) ) );
 
+            for ( int iB = 0; iB < vecNumFrameSizeConvBlocks[iChanCnt]; iB++ )
+            {
                 iUnused = opus_custom_encode ( pCurOpusEncoder,
-                                               &vecsSendData[iB * SYSTEM_FRAME_SIZE_SAMPLES * vecNumAudioChannels[iChanCnt]],
-                                               iClientFrameSizeSamples,
-                                               &vecvecbyCodedData[iChanCnt][0],
-                                               iCeltNumCodedBytes );
+                                            &vecsSendData[(long)iB * SYSTEM_FRAME_SIZE_SAMPLES * vecNumAudioChannels[iChanCnt]],
```
pljones, line 1410: Alignment / spacing.

src/server.h
```diff
@@ -310,11 +311,13 @@ class CServer :
     void WriteHTMLChannelList();
     void WriteHTMLServerQuit();
 
-    void DecodeReceiveDataBlocks ( const int iStartChanCnt,
+    static void DecodeReceiveDataBlocks ( CServer* sServer,
```
pljones, line 314: `pServer`

src/server.h
```diff
@@ -310,11 +311,13 @@ class CServer :
...
+                                   const int iStartChanCnt,
                                    const int iStopChanCnt,
                                    const int iNumClients );
 
-    void MixEncodeTransmitDataBlocks ( const int iStartChanCnt,
+    static void MixEncodeTransmitDataBlocks ( CServer* sServer,
```
pljones, line 319: `pServer`

src/server.cpp
```diff
@@ -966,25 +971,27 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
...
+void CServer::DecodeReceiveDataBlocks ( CServer*  sServer,
+                                        const int iStartChanCnt,
                                         const int iStopChanCnt,
                                         const int iNumClients )
 {
     // loop over all channels in the current block, needed for multithreading support
     for ( int iChanCnt = iStartChanCnt; iChanCnt <= iStopChanCnt; iChanCnt++ )
     {
-        DecodeReceiveData ( iChanCnt, iNumClients );
+        sServer->DecodeReceiveData ( iChanCnt, iNumClients );
```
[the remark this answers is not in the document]
softins: I think it's because `CServer::DecodeReceiveDataBlocks` and `CServer::MixEncodeTransmitDataBlocks` can only be passed to `tpThreadPool->enqueue()` as static functions, not instance methods, and therefore do not automatically inherit a `this` pointer when called. So it has to be passed manually as the first argument, and used explicitly when calling the class methods `DecodeReceiveData` and `MixEncodeTransmitData`. I don't think it's a problem, but it possibly deserves an explanatory comment in the code.

src/server.cpp
```diff
@@ -936,16 +942,15 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
...
+                                                    this, iStartChanCnt, iStopChanCnt, iNumClients ) );
             }
 
             // make sure all concurrent run threads have finished when we leave this function
-            FutureSynchronizer.waitForFinished();
-            FutureSynchronizer.clearFutures();
+            for ( auto& fFuture : Futures ) {
+                fFuture.wait();
+            }
+            Futures.clear();
```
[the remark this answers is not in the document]
softins: I think I prefer it as it is, with all uses of `Futures` localised to the same area, rather than some of it hidden away in another function.

src/server.cpp
```diff
@@ -826,21 +839,22 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
             }
         }
 
+        // use multithreading for any non-zero number of clients
+        // (was previously 10 clients or over, but is worth doing for all numbers)
+        bUseMT = bUseMultithreading && iNumClients > 0;
+
         // prepare and decode connected channels
-        if ( !bUseMultithreading )
+        if ( !bUseMT )
```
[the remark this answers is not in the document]
softins: If a `then` and an `else` are a lot different in length, particularly if one is only a line or two, I find it more readable to have ths shorter one first, and construct the condition accordingly. So I prefer the way it is.

src/server.cpp
```diff
@@ -826,21 +839,22 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
...
         {
-            for ( int iChanCnt = 0; iChanCnt < iNumClients; iChanCnt++ )
-            {
-                DecodeReceiveData ( iChanCnt, iNumClients );
-            }
+            DecodeReceiveDataBlocks ( this, 0, iNumClients - 1, iNumClients );
```
pljones, line 849: Is it worth a comment here to mirror the one at line 853, saying "run the OPUS decoder for all blocks" or whatever?

src/server.cpp
```diff
@@ -849,16 +863,18 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
...
+                                                           this,
+                                                           iStartChanCnt,
+                                                           iStopChanCnt,
+                                                           iNumClients ) );
             }
 
             // make sure all concurrent run threads have finished when we leave this function
-            FutureSynchronizer.waitForFinished();
-            FutureSynchronizer.clearFutures();
+            for ( auto& future : Futures ) {
```
pljones, line 874: I'm not a fan of `auto` - it means I have to think "so what does this actually mean" - it's not self-documenting.  Given `Futures` isn't defined locally, I think it would be nice to declare `future` explicitly.

src/server.cpp
```diff
@@ -936,16 +945,18 @@ static CTimingMeas JitterMeas ( 1000, "test2.dat" ); JitterMeas.Measure(); // TE
...
+                                                           this,
+                                                           iStartChanCnt,
+                                                           iStopChanCnt,
+                                                           iNumClients ) );
             }
 
             // make sure all concurrent run threads have finished when we leave this function
-            FutureSynchronizer.waitForFinished();
-            FutureSynchronizer.clearFutures();
+            for ( auto& fFuture : Futures ) {
```
pljones, line 956: As above.

src/threadpool.h
```diff
@@ -0,0 +1,124 @@
...
+#include <functional>
+#include <stdexcept>
+
+class CThreadPool
+{
+public:
+    CThreadPool() = default;
+    CThreadPool(size_t);
+    template<class F, class... Args>
+    auto enqueue(F&& f, Args&&... args)
```
pljones, line 44: More `auto`...

src/threadpool.h
```diff
@@ -0,0 +1,124 @@
...
+    std::mutex queue_mutex;
+    std::condition_variable condition;
+    bool stop;
+};
+
+// the constructor just launches some amount of workers
+inline CThreadPool::CThreadPool(size_t threads)
+    :   stop(false)
+{
+    for(size_t i = 0;i<threads;++i)
```
pljones, line 63: Spacing

pljones: And missing `{}` for the for `loop`.

src/threadpool.h
```diff
@@ -0,0 +1,124 @@
...
+        workers.emplace_back(
```
pljones, line 64: And I don't think the `[this]` and `{` should each be on new lines part of an argument (for `emplace_back(` (with a space)).

pljones: And generally spacing.

src/threadpool.h
```diff
@@ -0,0 +1,124 @@
...
+                        this->tasks.pop();
+                    }
+
+                    task();
+                }
+            }
+        );
+}
+
+// add new work item to the pool
```
pljones, line 87: And formatting throughout, basically.

src/server.cpp
softins: So is `auto` just a shorthand, or does it make something possible to express that can't otherwise be expressed succinctly?

src/server.cpp
```diff
@@ -421,10 +421,24 @@ CServer::CServer ( const int          iNewMaxNumChan,
...
+    if ( bUseMultithreading ) {
+        if ( iAvailableCores == 1 )
+        {
+            qDebug() << "found only one core, disabling multithreading";
+            bUseMultithreading = false;
+        }
+        else
+        {
+            // set maximum thread count to available cores; other threads will share at random
+            iMaxNumThreads = iAvailableCores;
```
hoffie, line 436: Should this be configurable, btw?
If not all cores are usuable by this Jamulus instance, more threads will just add more overhead, I'd assume?

See also https://github.com/jamulussoftware/jamulus/pull/960#issuecomment-782758759

The previous multithreading logic was not configurable either, so this should not be a blocker for this PR, I guess.

hoffie (approved):
It would be good to find the source of the overflow warning and fix it (or be confident that it is safe to ignore).

Side note: I have used this code today again (~12 users). There was one user with strange problems (sporadical 500ms ping times), but this may also be explained by bad Internet connectivity. I'm mainly mentioning this in case someone has observed similar things.

If this PR still had bugs, I'd expect all clients to be affected the same (or none at all).

src/server.cpp
softins: I've had a think about this, and I don't think reserving cores actually gains us anything. Certainly there seems no point in dividing the audio processing across more threads than there are cores. But using fewer threads would mean it is usually not using cores optimally. Using all the cores means they will have approximately equal amounts of spare capacity, which can be used by other lighter threads or by other programs, even other Jamulus instances. If there was, say, an additional Jamulus instance, it too could make use of the spare capacity on all the cores. If we "reserved" a core by using fewer threads, I don't think there is any way for the second instance to specifically use the less busy core instance (I suspect it just uses any free core, but this is way deeper than I know), so we may as well allow the free time to be distributed evenly.

src/server.cpp
softins: > I'm not a fan of `auto` - it means I have to think "so what does this actually mean" - it's not self-documenting. Given `Futures` isn't defined locally, I think it would be nice to declare `future` explicitly.

The use of `auto` seems to be a standard C++11 idiom to be able to declare the loop variable succinctly and in a DRY (don't repeat yourself) way. So I would strongly favour leaving these as they are.

src/server.cpp
pljones: I think it's possible (in Linux, anyway), to limit the number of cores visible to a process, so it's not really needed in the code itself.  (You can also essentially tell a process to use a specific core by only letting it see that core.)

pljones (approved):
I'm sure clang-format won't like it :)

softins:
> I'm sure clang-format won't like it :)

I haven't actually looked at clang-format yet. Is it a fixer, or just a checker?

Issue #742, opened 2020-11-24
Pan setting presets


rickrowan:
Especially for choral work, setting individual users to different L/R balance positions makes a huge difference in ones ability to hear the other parts.  The experience is akin to standing spatially in the group.  With the new capability to host many users, the task of setting individual balance levels can be daunting.

Suggstion:  Having a preset or auto set pan-all button such as:  a) random setting for each client or b) set each client to vary from Left to Right as they appear from left to right across the client window.  This is probably best, especially if clients are sorted by part.

passing:
To solve the issue described, I could also imagine an alternative interface where you would not have musicians in columns with control sliders, but instead have a rectangular space that corresponds to the pan setting left-to-right and the volume top-to-bottom. Then the clients would be dots/circles on that space that you could move around individually or in groups to change their volume and balance positions.
Could be a quite intuitive way to control the different channels by just dragging them around...

menzels:
I am working on an implementation suggestion for this suggestion:

> Suggstion: Having a preset or auto set pan-all button such as: a) random setting for each client or b) set each client to vary from Left to Right as they appear from left to right across the client window. This is probably best, especially if clients are sorted by part.

menzels:
Working on this feature, i found that i like to keep my own signal in the middle. 
What do you think? is that a good default or do we need to make that configurable?

rickrowan:
I usually find I move myself a bit off-centre but not a lot.  Typically, I don't want anyharmonic sound centred.  I find they need to be just a little off centre do remain nicely distinguishable.  That said, it is not hard to adjust just yourself either way.  I's go with your idea and leave yourself alone.

gilgongo:
Migrating to discussion until such time as a desired spec/PR can be arrived at.

PR #963, opened 2021-02-06
Pan setting presets #742 add panleftright and panrandom

menzels:
This is an implementation suggestion for the easy part of #742 

MyChannel is kept in the middle. it might be desirable to make that configurable too.


ann0see:
@menzels Thanks for the PR! I can’t really say much to this PR (wasn’t involved), but It seems as if this breaks compilation on Linux. Can you please have a look at the checks page in this PR and look what’s wrong?

ann0see:
@menzels, any updates here? I assume the random generator on linux is missing. Probably it needs to be added somehow?

src/audiomixerboard.cpp
```diff
@@ -1264,6 +1268,52 @@ void CAudioMixerBoard::SetAllFaderLevelsToNewClientLevel()
...
+            iCount += 2;
+        }
+
+        vecpChanFader[iChannelID]->SetPanValue ( iCount++ * AUD_MIX_PAN_MAX / iNumConnectedClients );
+    }
+}
+
+void CAudioMixerBoard::SetAllPanRandom()
+{
+    QMutexLocker locker ( &Mutex );
```
ann0see, line 1305: Not sure about that (not a Qt expert), but do you need to unlock it?

src/audiomixerboard.cpp
```diff
@@ -1075,12 +1076,15 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
...
 
+    veciChannelOrder.clear();
+
     for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
     {
         const int iCurFaderID = PairList[i].second;
 
         if ( vecpChanFader[iCurFaderID]->IsVisible() )
         {
+            veciChannelOrder.push_back(iCurFaderID);
```
ann0see, line 1087: I think, you need to change the code style here?

ann0see (changes requested):

src/audiomixerboard.cpp
menzels: The mutexlocker unlocks the mutex automatically when it goes out of scope (RAII)

menzels:
> 
> 
> @menzels, any updates here? I assume the random generator on linux is missing. Probably it needs to be added somehow?

I have to check how to add the random generator on linux, hope to find time this week.

ann0see:
Great! Also please have a look at the conflict.

ann0see:
Any updates here?

menzels:
> 
> 
> Any updates here?

It seems we are using Qt 5.9.5 for linux and Qt 5.9.9 for MacOS.
QRandomGenerator is only available since Qt 5.10.

What is the reason for using such and old version? Would it be possible to update?
The current version would be Qt 5.15.3.

ann0see:
Yes. This old version of Qt is needed to support older versions of macOS. I think similar reasons could apply for Linux.

src/audiomixerboard.cpp
```diff
@@ -619,7 +620,7 @@ void CChannelFader::UpdateSoloState ( const bool bNewOtherSoloState )
 
 void CChannelFader::SetChannelLevel ( const uint16_t iLevel )
 {
-    plbrChannelLevel->SetValue ( iLevel );
+    plbrChannelLevel->SetValue ( iLevel * dPreviousFaderLevel * 0.01 );
```
hoffie, line 623: Hrm, I don't understand why this is necessary and whether this is safe to do as this touches a generic function unconditionally of the usage of the new feature?

hoffie (commented):
> What is the reason for using such and old version? Would it be possible to update?
> The current version would be Qt 5.15.3.

Autobuild from CI is only one aspect. Jamulus can be built on newer Qt5 (and often is), but we also have to keep supporting older versions.

I wonder if one could use the C `rand()` (which may be harder to use, requires `srand()` and might be biased when using modulo?) or C++11 features such as [uniform_int_distribution](https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution). Sadly, I'm not sure if we can assume C++11 to be supported (for Android, we require C++14 according to Jamulus.pro, for "unix" we require C++11; not sure about Mac/Windows). Maybe @pljones or @softins know?

This feature looks really interesting and I'm looking forward to having it. Thanks for working on this! :)

softins:
There are various systems in use at different versions of Qt. The binaries for Mac need to use 5.9 until we decide to drop support for macOS 10.10. We have no way of knowing how many users are still on such an old version of macOS, until we drop support and they complain!

Of the systems I have:

CentOS 7 uses Qt 5.9.7
CentOS 8 uses Qt 5.12.5
Amazon Linux 2 uses QT 5.9.2
Raspberry Pi OS uses Qt 5.11.3
Ubuntu 18 uses Qt 5.9.5
Ubuntu 20 uses Qt 5.12.8

Regarding C++11 support, I think we assume that on all platforms now. My recent PR for sanitising recording filenames used the C++11 iterator syntax `for (auto &x : vector)` without any problems on any of our supported platforms.

pljones:
> I wonder if one could use the C `rand()` (which may be harder to use, requires `srand()` and might be biased when using modulo?) or C++11 features such as [uniform_int_distribution](https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution). Sadly, I'm not sure if we can assume C++11 to be supported (for Android, we require C++14 according to Jamulus.pro, for "unix" we require C++11; not sure about Mac/Windows). Maybe @pljones or @softins know?

My emergency lights start flashing when anyone mentions random number generation around Jamulus.  Bad move.  There is generally a lot of real time needed for good random number generation, either to get "randomness" from the environment or to run complex calculations on the CPU.

Just saying... :)

src/audiomixerboard.cpp
softins: It appears to be part of an independent feature, and was merged from another of @menzels branches called "change level indicater to post-fader". Should probably be in a separate PR.

hoffie:
Are there practical experiences regarding the random mode?
In my experience, people choose one of the following:
- Set a random pan to create a more spacial experience. People choose random, because it's hard to equally space channels manually. However, it seems like the technical implementation for the latter would be simpler (SetAllPanLeftRight), so I'm wondering if this would be sufficient already? Even if not, it would already be an improvement, so maybe we could reduce the scope of the PR to just that for now?
- Set a similar pan for similar groups (i.e. "i want the basses left of me, the altos in front of me, etc."). So maybe a "auto-distribute spacially, but keep groups together" would be helpful here. It would just be another "nice to have", so I'm not sure of this is really required.

pljones:
I was actually thinking about your second example when I was implementing the increased number of groups change.  Being able to set the groups, then say "pan them apart" would be nice (keeping the basses in the middle, of course, by leaving them ungroup or something).  Do the pan knobs adjust all grouped channels the same way the faders do their adjustment?  I'd not even got as far as checking that...

ann0see:
Closing since it seems to break Qt 5 compatibility and no solution has been found yet. Feel free to open a new PR in future!

Issue #146, opened 2020-04-26
Server pipe out the audio to some DAW software for mixing


chaolue:
Firstly, this software is great. We've been using it to practice and record performances together. 

I have a request for a feature that isn't quite in Jamulus but may not need too much to make it happen, but I really don't know. And I want to see if this is something that others are interested in too (while I think this feature would be cool others might have better ideas).

I had the thought that it would be good if there was a way to have the server pipe out the audio it receives from clients to some DAW software for mixing either for a live stream, and/or in order to send it back to clients processed. 

Would it take much to add this in? Initially it could be for Jack on Linux. It may be similar to how the server pipes out audio for recording. 

Again really impressed with the software and development. Looking forward to what plans your have for it.

pljones:
I stream audio to and from Reaper to Jamulus now.  It's not something that needs adding to Jamulus.

chaolue:
Thanks @pljones, just to clarify; are you saying that it is already possible in the Jamulus server to pipe all channels of the clients audio to a DAW like Reaper?

It sounds like you're describing how a client could send audio to a DAW before sending it through Jamulus on to the server. I'm talking about the server itself using a DAW with all channels of the clients audio. Does that make sense? Or even sound like a good idea?

pljones:
The server has no audio I/O itself.  It can record audio to files that can be read by a DAW (indeed, it writes a Reaper RPP to go along, which lines all the files up in the time line correctly - the file names also encode the position).

Remember, time is _critical_ for Jamulus.

The problem with passing the data from one client to each other client, unmixed is that each time a new client joins, the volume of data uploaded _by the server_ increases exponentially, so the delay in sending to each client will similarly increase.  Currently, the increase is linear - as it's always only a stereo pair being added.  That exponentially increase will exceed the available time (i.e. the duration of a packet) very rapidly, meaning packets start getting dropped constantly to keep up.

elliotclee:
pljones: It's true, but mixing the incoming client audio also takes time. It would be nice to have the per-client option for the server to send the streams synchronized but unmixed back to that client. In my situation (fast symmetrical Internet connection, and running my client on the same computer as the server) it wouldn't be a performance problem. I understand the exponential increase would be a problem for some but I'd like at least the option of a client getting all the client streams back unmixed. Then whoever is producing the livestream of the performance could mix it down in their DAW.

[...]

PR #967, opened 2021-02-07
Streamer2

dingodoppelt:
This code calls ffmpeg to stream a stereo mix directly from the server either into a streaming service like icecast or a file.


pljones (commented):
OK, it will need to be made conditional compile not WIN32.
```
#ifndef WIN32
....code....
#endif
```
throughout.

The `.qm` files...  Can you try `git reset HEAD~ whatever.qm` and see if that fixes it?

src/streamer/jamstreamer.cpp
```diff
@@ -0,0 +1,35 @@
...
+}
+
+void CJamStreamer::Init( const QString strStreamDest ) {
+    this->strStreamDest = strStreamDest;
+}
+
+void CJamStreamer::OnStarted() {
+    // create a pipe to ffmpeg called "pipeout" to being able to put out the pcm data to it
+    pcm = 0; // it is necessary to initialise pcm
+    QString command = "ffmpeg -y -f s16le -ar 48000 -ac 2 -i - " + strStreamDest;
```
hoffie, line 29: It might be worth considering hardening the ffmpeg call as it executes a command and includes a variable in the invocation.  fail to find a `popen()` alternative which takes an array of arguments instead of a string (i.e. like `exec*()` to `system()`). Unless I'm missing something, the only remaining hardening measure would be proper quoting + escaping of the variable. One might argue that the variable is currently supposed to come from a trusted context (command line), but things may change in unexpected ways sometimes. Improving this would also ensure that special chars will not break things even in well-intentioned environments.

hoffie (changes requested):

dingodoppelt:
> The `.qm` files... Can you try `git reset HEAD~ whatever.qm` and see if that fixes it?

that fixed it, thanks!

src/streamer/jamstreamer.cpp
npostavs: Perhaps use [`QProcess`](https://doc.qt.io/qt-5/qprocess.html)?

hoffie: > Perhaps use [`QProcess`](https://doc.qt.io/qt-5/qprocess.html)?

Good idea! This should solve the potential security impact and might also get us cross-platform support for free. :)

pljones: Hm.  I'm thinking you're going to want to _know_ the path the ffmpeg binary.  Ideally from either a setting or from a command line option.  Better not just to launch whatever executable happens to get picked up from "ffmpeg" getting run.

dingodoppelt: > the only remaining hardening measure would be proper quoting + escaping of the variable

How would I go about and do this? At the moment I just read the output part of ffmpeg from the commandline (which can be anything that ffmpeg understands like "-f mp3 icecast://source:<password>:<port>/<mountpoint>" or something completely different. I don't know in advance how many arguments will be given in different usecases)

> Perhaps use QProcess?

Thanks for the advice! I looked into it and can't see how I could alter my code (I'm not a programmer ;) to do the same as before using QProcess. Any hints on what I would be looking for?

> Better not just to launch whatever executable happens to get picked up from "ffmpeg" getting run.

I've thought about that before and come to the conclusion that it won't matter for security reasons since you can call anything named "ffmpeg". Allowing the user to specify which program to call doesn't make it more secure in my opinion, since you'd be able to call anything you want. It'd only be useful if you wanted to call a certain build of ffmpeg besides the one already installed system-wide one, but I couldn't think of a usecase where you would want to do that.

npostavs: > I looked into it and can't see how I could alter my code (I'm not a programmer ;) to do the same as before using QProcess.

I haven't done this before, but it should work like: `FILE* pipeout` is replaced by `QProcess qproc`, the `popen` call would be replaced by

```
QStringList argumentList = { "ffmpeg", "-y", "-f", "s16le", "-ar", "48000", "-ac", "2", "-i", "-", strStreamDest };
qproc.start ( "ffmpeg", argumentList, QIODevice::WriteOnly ); // Note that program name is also repeated as first argument
```

Then the `fwrite(..., pipeout)` would be replaced by `qproc.write(...)`. And `pclose(pipeout)` becomes `qproc.closeWriteChannel()`

src/streamer/jamstreamer.cpp
```diff
@@ -0,0 +1,37 @@
...
+void CJamStreamer::process( int iServerFrameSizeSamples, CVector<int16_t> data ) {
+    if (pcm == 0) // thanks to hselasky
+    {
+        pcm = new int16_t [2 * iServerFrameSizeSamples];
+    }
+    for ( int i = 0; i < ( 2 * iServerFrameSizeSamples ); i++ )
+    {
+        pcm[i] = data[i];
+    }
+    fwrite(pcm, 2, ( 2 * iServerFrameSizeSamples ), pipeout);
```
npostavs, line 20: The copy to `pcm` is redundant since vector contents are guaranteed to be contiguous. So you can put `&data[0]` instead of `pcm` here and skip the copy loop.

src/streamer/jamstreamer.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "jamstreamer.h"
+
+using namespace streamer;
+
+// --- CONSTRUCTOR ---
+CJamStreamer::CJamStreamer() {}
+
+// --- PROCESS ---
+// put the received pcm data into the pipe to ffmpeg
+void CJamStreamer::process( int iServerFrameSizeSamples, CVector<int16_t> data ) {
```
npostavs, line 11: `data` should be a reference to avoid copying all its data per function call: `CVector<int16_t>& data`

src/streamer/jamstreamer.cpp
dingodoppelt: thanks for the advice! did it and it worked :)

src/streamer/jamstreamer.cpp
dingodoppelt: i can't get this one to work though. is it in any way possible for you to fork my code and raise a pull request? (that whole pointer business drives me mad, i'm still not a programmer)

npostavs: Seems like it worked: https://github.com/dingodoppelt/jamulus/pull/2

pljones:
Is this ready for final review?

src/streamer/jamstreamer.cpp
dingodoppelt: almost there :) the QProcess.write call still complains about the given arguments.
could you try compiling yourself and see what the output is? I've tried a few things and it only made it worse...

npostavs: Oh yeah, `QProcess.write` takes `const char*` rather than `const void*` so some more casting is needed:

        qproc.write ( reinterpret_cast<const char*> ( &data[0] ), sizeof (int16_t) * ( 2 * iServerFrameSizeSamples ) );

dingodoppelt: thanks! this compiles now but crashes the server with this message:
```
QObject: Cannot create children for a parent that is in a different thread.
QSocketNotifier: Socket notifiers cannot be enabled or disabled from another thread
```
i'm creating a thread and connecting it with the signals in server.cpp and i'm totally new to all this.
sorry for asking too many questions :/

npostavs: Oh, hmm. I'm not really familiar with how threading works for Qt either, but maybe the fix is to make `qproc` into a pointer, so that the object can be constructed only after the `moveToThread` finishes? That way it would be constructed and used in the same thread, I think.

dingodoppelt: making it a pointer gets rid of the error messages but the server still just silently segfaults. we're close, i guess ;)
could you or anybody else ( maybe @hoffie , @pljones ) give this code a test run to see what might be happening there?
you can launch the server with:
```
--streamto "-f mp3 output.mp3"
```

npostavs: Well, it doesn't crash for me, but I'm not able to get any output either with the `QProcess*` version. It looks like `ffmpeg` never even got started, although I didn't manage to catch any `QProcess::FailedToStart` error. Maybe we should just stick with the `popen` version...
https://github.com/dingodoppelt/jamulus/compare/streamer2...npostavs:streamer2.np.qproc

dingodoppelt: > Well, it doesn't crash for me

It does for me only when I connect a client but yes, ffmpeg never gets started it seems.
> Maybe we should just stick with the popen version...

I think so, too

src/main.cpp
```diff
@@ -369,6 +370,22 @@ int main ( int argc, char** argv )
...
+                                 argv,
+                                 i,
+                                 "--streamto", // no short form
+                                 "--streamto",
+                                 strArgument ) )
+        {
+            strStreamDest = strArgument;
+            qInfo() << qUtf8Printable( QString( "- stream destination: %1" )
+                .arg( strStreamDest ) );
+            CommandLineOptions << "--streamto";
```
hoffie, line 384: The new command line argument should probably also added to the `--help` text.

src/streamer/jamstreamer.cpp
hoffie: I'm not seeing a segfault either, but I have found three problems which I have fixed in a PR to @npostavs' branch against this PR. :)

https://github.com/npostavs/jamulus/pull/1

dingodoppelt: Awesome! Thanks a lot for your work :)
Would that mean I can get rid of the conditional compile for win32 as well? I can't check since I have no windows machine here.

src/streamer/jamstreamer.cpp
```diff
@@ -0,0 +1,74 @@
...
+        qproc = new QProcess;
+        QObject::connect ( qproc, &QProcess::errorOccurred, this, &CJamStreamer::onError );
+        QObject::connect ( qproc, QOverload<int, QProcess::ExitStatus>::of( &QProcess::finished ), this, &CJamStreamer::onFinished );
+        qproc->setStandardOutputFile ( QProcess::nullDevice() );
+    }
+    QStringList argumentList = { "-loglevel", "error",
+                                 "-y", "-f", "s16le",
+                                 "-ar", "48000", "-ac", "2",
+                                 "-i", "-" };
+    argumentList += strStreamDest.split( QRegExp("\\s+") );
```
npostavs, line 30: Perhaps we should use [`QProcess::splitCommand`](https://doc.qt.io/qt-5/qprocess.html#splitCommand) for this?

src/streamer/jamstreamer.cpp
hoffie: > Would that mean I can get rid of the conditional compile for win32 as well?

I think so, yes, but one should test.

> I can't check since I have no windows machine here.

Neither can I (Linux-only here). A first smoke test would probably be removing the ifdefs. CodeQL should probably catch if compiling works. Someone should still run a real test with ffmpeg if possible.

src/streamer/jamstreamer.cpp
hoffie: > Perhaps we should use [`QProcess::splitCommand`](https://doc.qt.io/qt-5/qprocess.html#splitCommand) for this?

Absolutely! I didn't know about that. What I did just fixed the immediate issue (`--streamto "-f mp3 foo.mp3"`) but now that you mention it, it would probably break when attempting to escape things to ffmpeg (e.g. `--streamto "-f mp3 '/tmp/path with spaces'"`) or something. So my proposal makes things better, but is still not complete. I'm in favor of using `splitCommand`.

The only problem I see is that this function is only available as of Qt 5.15. As far as I can see, Jamulus is supposed to support earlier versions (5.9?). Therefore, a compromise might be a version check which uses `splitCommand` if available and falls back to the basic regexp for older versions.

Are you (@npostavs) going to continue with this? Or is @dingodoppelt?
I honestly only looked at this whole PR due to the potential security impact and I only continued investigating after the troubles with QProcess seemed stuck a bit.

Personally, I currently have no need for this feature at all so I would prefer if one of you could continue. :)

src/streamer/jamstreamer.cpp
dingodoppelt: It doesn't compile on windows whatsoever. I've put the conditional compile back in

src/streamer/jamstreamer.cpp
dingodoppelt: @hoffie 
I'd really love to, but I obviously lack the necessary skills (I'm still overly happy that my first implementation did what it did but that's about what I can contribute since you guys are accomplished programmers and I still don't know what I'm doing here ;)

npostavs: Ah, didn't realize it 5.15 specific. 

> Therefore, a compromise might be a version check which uses `splitCommand` if available and falls back to the basic regexp for older versions.

The compromise sounds like a recipe for confusion to me. How necessary are the multiple arguments? E.g., `-f mp3` is redundant if you already give a `.mp3` filename anyway, as far as I've understood.

```
       -f fmt (input/output)
           Force input or output file format. The format is normally auto
           detected for input files and guessed from the file extension for
           output files, so this option is not needed in most cases.
```

hoffie: > E.g., -f mp3 is redundant if you already give a .mp3 filename anyway, as far as I've understood.

Yes, my example was hypothetical. No idea why I added that. The help text talks about "ffmpeg argument**s**", but not sure if or in what circumstances this would be needed. If the requirement could be dropped, this would be the simplest fix. @dingodoppelt?

dingodoppelt: @hoffie 
I use it to stream to icecast and that won't work unless the format is specified. Since ffmpeg takes a lot of arguments (even multiple outputs are possible) this is very much needed so you get the full functionality that ffmpeg offers.
The recording to mp3 was just for test purposes so you wouldn't have to set up an icecast server to test the code. I wasn't aware, that the -f option is redundant for I haven't used it to record into a file (I use the awesome jamrecorder for that purpose)

hoffie: Not sure how we could move this forward.

> The compromise sounds like a recipe for confusion to me.

I generally agree. However, we are talking about edge cases I think (e.g. filenames/urls with spaces). It seems like the half-broken hack I've implemented works for @dingodoppelt's use case. If we can improve that code to be less half-broken on modern Qt versions, I think that's a benefit.

So I'd still vote for the compromise as I don't see any better alternatives?

npostavs: Hmm, the other possibilities would be for `--streamto` to take multiple arguments, a la `find -exec ... \;`. Or just require a single `--streamto` argument which is a shell script that calls `ffmpeg`?

hoffie: > the other possibilities would be for --streamto to take multiple arguments, a la find -exec ... \;

Not sure I'm following. Do you mean repeated `--streamto` arguments which would be concatenated to an argument list by Jamulus? This would delegate the escaping/parsing part to the shell, so we would get it for free. However, usability-wise it does not sound optimal to me.

> Or just require a single --streamto argument which is a shell script that calls ffmpeg?

Jamulus-wise, this would be the simplest form. @dingodoppelt Would that work for you as well?
It would have additional advantage that non-ffmpeg use cases could be enabled by this feature as well. It would have the downside that the input details (sample rate, etc.) would have to be specified in such a script.

dingodoppelt:
Hey everybody,
I converted this PR to "draft" since I still have problems with the server segfaulting since the implementation of QProcess.
I don't know why it works on some servers. For "production use" I fell back to the implementation without QProcess and it works fine. I don't know at all why this happens. Any ideas?
It seems that the problem only occures on machines > 2 cores, but that's just a guess since I can just test on a limited number of machines. It would be great if somebody else could test the code. I really don't want it merged if it introduces unexpected crashes of the server.

hoffie:
> I converted this PR to "draft" since I still have problems with the server segfaulting since the implementation of QProcess.

Any chance you can capture a core dump of such a crash? Haven't observed one here.

dingodoppelt:
> > I converted this PR to "draft" since I still have problems with the server segfaulting since the implementation of QProcess.
> 
> Any chance you can capture a core dump of such a crash? Haven't observed one here.

i hope it is what you were asking for:

https://gist.github.com/dingodoppelt/a381f37411f0f0e92e91fb8c02ae54aa

hoffie:
> i hope it is what you were asking for:
> 
> https://gist.github.com/dingodoppelt/a381f37411f0f0e92e91fb8c02ae54aa

This helps, I think this is the problematic part:
```==7780== Conditional jump or move depends on uninitialised value(s)
==7780==    at 0x1A45B0: streamer::CJamStreamer::OnStarted() (in /home/nils/test/jamulus/Jamulus)``` 

However, I fail to come up with an idea what this is about exactly. Sounds like a NULL pointer reference. The likeliest candidate would be `qproc`, but this seems guarded/initialized properly.

Can you try running with `gdb` instead of `valgrind` and capture a backtrace using `bt` after the crash? (I'm not too familiar with gdb either, but I hope for some more hints)

I assume you are using the latest code from this PR, right?
@npostavs Any idea?

**Comment by dingodoppelt** [comment]
After a few tests I found that connecting to the server works when it is not registered on a central server list. Joining via the connection dialog while it is registered on a central server crashes the server (??)
Could it be a bug in the central server code then?

**Comment by pljones** [comment]
OK, unfortunately - and I really wanted this to make it - I think it'll have to wait till after 3.7.0 now.  Maybe I'll get a chance to look into it, too.

**Comment by dingodoppelt** [comment]
@pljones 
Good decision and thanks for your kind note :) As I said before, I really don't want it merged too as long as it breaks the server.
It is no problem for me since I'm using a working version already which is doing fine for my usecase.
Maybe we can concentrate on testing the existing code before changing it any more to see if there are any other undiscovered bugs.

**Comment by npostavs** [comment]
> After a few tests I found that connecting to the server works when it is not registered on a central server list. Joining via the connection dialog while it is registered on a central server crashes the server (??)

Yep, I reproduced it when registering with the server. Looks like `qproc` isn't initialized to NULL automagically? (my C++ knowledge is a bit rusty) The following patch seems to fix it.

```diff
diff --git c/src/streamer/jamstreamer.cpp w/src/streamer/jamstreamer.cpp
index 310d0653..23e32452 100644
--- c/src/streamer/jamstreamer.cpp
+++ w/src/streamer/jamstreamer.cpp
@@ -3,7 +3,7 @@
 
 using namespace streamer;
 
-CJamStreamer::CJamStreamer() {}
+CJamStreamer::CJamStreamer() : qproc ( NULL ) {}
 
 // put the received pcm data into the pipe to ffmpeg
 void CJamStreamer::process( int iServerFrameSizeSamples, const CVector<int16_t>& data ) {

```

dingodoppelt:
Just another update of what is not working since the qproc implementation:
I can just connect to the icecast stream as a listener once. The second time I try to connect it won't start playing but ffmpeg seems to be streaming nonetheless. Can anybody check to see if this behaviour is reproducible? The old implementation works but doesn't feature the error logging which I'd really like to have. Is it possible to do this without QProcess as well?

gilgongo:
Please can post (or point me to) a quick summary of how this function would be used so that we can add that to the docs at some point? Thanks.

dingodoppelt:
@gilgongo The usecases might be:
1. Being able to stream the server's stereo output via ffmpeg to a streaming service like icecast to have less listeners on the jamulus server itself blocking slots for players. (listeners blocking slots are becoming a problem more and more)
2. Being able to hand out the streaming link to people that can't run jamulus but need to hear what's happening on the server (choirs are using it already for that purpose)
3. Preview a server via the streaming link without the need to fire up jamulus itself (I'm thinking of a play button in the jamulus explorer for example)
4. To show the possibilities of using jamulus to people not yet convinced to try it by handing them a link they can play on any device like a smartphone, TV set and so on to get an idea of whats happening on jamulus
5. Being able to perform on jamulus and embed the stream in a webpage with an accompanying video stream so listeners (non-musicians) don't have to install jamulus to just being able to listen.

gilgongo:
> How to use it:
> just pass the --streamto argument any output argument ( beginning from {[outfile options] outfile} ) that ffmpeg understands (ffmpeg --help)

Cool. Sounds like it might need its own entry in Tips & Tricks to cover all that. As to the ffmpeg usage, can you give an example of that? so `jamulus -s --streamto <what would be here? a url? socket?>`

dingodoppelt:
@gilgongo 
This is what I pass to --streamto for streaming an ogg stream to an icecast server:

```--streamto "-codec:a libvorbis -qscale:a 3 -content_type 'audio/ogg' -vn -f ogg icecast://source:<icecast_source_pw>@<icecast_server_url_or_IP>/<icecast_mountpoint>"```

When calling ```ffmpeg --help``` the usage is described as: 
```usage: ffmpeg [options] [[infile options] -i infile]... {[outfile options] outfile}...```
--streamto takes the part which is ```{[outfile options] outfile}...```
You can find more info on the options ffmpeg accepts here: https://ffmpeg.org/ffmpeg.html#Main-options

**Comment by pljones** [comment]
Quick question:

Once a session ends, is the `ffmpeg` stream ending?  If not, is the same stream used for the next session or does another get started?

**Comment by dingodoppelt** [comment]
> Quick question:
> 
> Once a session ends, is the `ffmpeg` stream ending? If not, is the same stream used for the next session or does another get started?

The stream should be ending as it is receiving the server stopped signal which triggers CJamStreamer::OnStopped in the streamer code. The next time a client joins (after the server idling) it receives the server started signal and executes CJamStreamer::OnStarted which calls ```ffmpeg``` again to set up a new stream.
With the QProcess implementation not being code I wrote I can't tell you exactly and I'm still investigating why it doesn't work which is why I'm still using my old implementation without QProcess.

**Comment by hoffie** [comment]
So, a short summary of the current state:
- The initially proposed code seems to work stable, but has some things which could be optimized:
  - Doesn't use the cross-platform Qt primitives.
  - Might make it easy to introduce security vulnerabilities later (command injection), although not a concrete problem right now.
  - Lacks error handling regarding ffmpeg startup
- The updated code based on QProcess does have error handling, but is not production-ready yet:
  - Seems like there's still an unresolved stability issue. I will also try to reproduce this and see if I can find the problem.
  - The current code makes assumptions which may not hold true in all cases ("no ffmpeg argument will require quoting or will contain spaces") as the argument splitting is regexp-based, which seems wrong. The proper solution would be a function (`splitCommand`) which is not available in all Jamulus-supported Qt versions
- Besides that, there is an open usability question: Should this feature be "ffmpeg-based streaming" or "generic PCM-on-stdin streaming"? In other words: Should `ffmpeg` be hardcoded? My tendency would be towards the generic solution. This would have the huge advantage that it would circumvent the `splitCommand` problem and would keep the implementation smaller. We could still supply an example streaming script with ffmpeg hardcoded. Any opinions?

@gilgongo The last part will also be important for the documentation. Maybe you have an opinion on that as well?

**Comment by pljones** [comment]
> "generic PCM-on-stdin streaming"

Would it work on Windows?  The benefit of the "internal pipe" solution is that "it should", as you're managing the stream internally.  I'm not sure if the server stdout would be easily available to pipe, particularly in the GUI server.

**Comment by dingodoppelt** [comment]
>  Besides that, there is an open usability question: Should this feature be "ffmpeg-based streaming" or "generic PCM-on-stdin streaming"? In other words: Should `ffmpeg` be hardcoded? My tendency would be towards the generic solution. This would have the huge advantage that it would circumvent the `splitCommand` problem and would keep the implementation smaller. We could still supply an example streaming script with ffmpeg hardcoded. Any opinions?

I'd go with the generic solution as well. This wouldn't make ffmpeg compulsory and would allow for other usecases to be covered. Another advantage to the existing implementations is to be able to stop and start streaming without having to restart (reconfigure) the server to stop/start the stream or change its destination.

**Comment by hoffie** [comment]
> Another advantage to the existing implementations is to be able to stop and start streaming without having to restart (reconfigure) the server to stop/start the stream or change its destination.

Ah, right, didn't even think of it. Cool! :)

> > "generic PCM-on-stdin streaming"
> Would it work on Windows? The benefit of the "internal pipe" solution is that "it should", as you're managing the stream internally. I'm not sure if the server stdout would be easily available to pipe, particularly in the GUI server.

Not sure if we are talking about the same. With "generic PCM-on-stdin streaming" I meant that Jamulus opens a sub-process and streams PCM to stdin of that sub-process. I'm not referring to Jamulus' stdin. This approach is the same with or without `QProcess`.

Regarding Windows: When using `QProcess` I'd hope it could work as the code would not use anything which is documented as non-Windows-only. Should be tested, of course.

**Comment by pljones** [comment]
OK - that's good -- I just worry when I see "stdin" and "process" for Windows... :)

**Comment by hoffie** [comment]
> I can just connect to the icecast stream as a listener once. The second time I try to connect it won't start playing but ffmpeg seems to be streaming nonetheless. Can anybody check to see if this behaviour is reproducible?

So, you are using --streamto to make ffmpeg send the stream to an Icecast server. When you connect to the Icecast server, the first clients gets audio data, but the second client won't. Do I understand correctly?
I can't think of any reason how the Jamulus-side implementation or ffmpeg could influence this part...?

Do you have a specific way to reproduce this? Maybe even without Icecast but with local files or anything what one could easily replicate?

**Comment by dingodoppelt** [comment]
> Do you have a specific way to reproduce this? Maybe even without Icecast but with local files or anything what one could easily replicate?

@hoffie 
To replicate the icecast issue I can't think of how I could do that with a file locally.
With an icecast server running you just launch the jamulus server and connect a client so the stream starts. Then connect to the stream to listen. You'll find the first time you connect you'll hear sound. Then disconnect from the stream and rejoin to find the stream won't start anymore. It is just working once. To listen to the stream again you either disconnect all clients from the server (which closes ffmpeg on server idling and restarts ffmpeg on the first client connecting to the server) or restart the server altogether.

File: src/streamer/jamstreamer.cpp (hunk 5 above)
Comment (npostavs) [reply src/streamer/jamstreamer.cpp:28]: > The only problem I see is that this function is only available as of Qt 5.15. As far as I can see, Jamulus is supposed to support earlier versions (5.9?).

I just noticed that the Qt docs seem to say the split functions are only available as of Qt 5.14, so it might be trouble as well?

https://doc.qt.io/qt-5/qstring.html#split-6

**Comment by pljones** [comment]
So you're saying the set up is
* Running Jamulus instance with potential session stream to a per session ffmpeg instance
* Running Icecast instance listening for audio data and client connections
* ffmpeg instance configured to output its input from Jamulus to the Icecast instance

ffmpeg will only be running during a session.  Jamulus and Icecast are running all the time.

(1) Right so far?

Then a Jamulus client connects to the Jamulus server and ffmpeg connects to Icecast.

Then an Icecast client connects to the Icecast server.

At this point the Icecast client can hear the Jamulus session.

(2) Right so far?

The Icecast client disconnects.

Another Icecast client connects but now cannot hear audio.

(3) Right so far?

If that's all correct, I'd say "check the logs" to see what's reported as going on.

**Comment by hoffie** [comment]
@dingodoppelt In addition to what @pljones asked, can you confirm that this behavior does not occur when using some other, non-Jamulus icecast client?

I can't really think of anything which would influence this from within Jamulus or the QProcess vs. no-QProcess handling.

**Comment by dingodoppelt** [comment]
> can you confirm that this behavior does not occur when using some other, non-Jamulus icecast client?

@hoffie: yes, my code from the original pr works.

**Comment by hoffie** [comment]
I have just installed icecast and tried to use this feature (10400f54fbc62894445b6920fd0a49d1cd1785dc, the latest state of this PR). After some fiddling with icecast config/authentication, it worked, even with multiple clients on icecast and/or on Jamulus. So, I'm not able to reproduce the issue you are describing.

Can you share some more details (icecast/ffmpeg version and config)?

```
$ icecast -v
Icecast 2.4.4

$ ffmpeg -v
ffmpeg version n4.3.2
```

I used the default icecast config and configured a single listener mountpoint.
I used `./Jamulus -s -n --streamto "-f mp3 icecast://source:hackme@127.0.0.1:8000/stream"`

Some further things I've observed:
- [ ] `--streamto` does not fail when used in Jamulus as a client mode, yet, it does not work there. I think we should fail in `src/main.cpp`.
- [ ] When ffmpeg crashes (e.g. due to an icecast server restart), streaming will stop with no way to restart it besides re-connecting all clients and/or restarting the Jamulus server. Not sure, if/how this could be improved. We could think about restarting on failure, but this adds more complication and risks crash loops.
- [ ] The current code still uses the ffmpeg-focused approach which requires command line splitting. I think the plan was to go for the generic approach and add an appropriate example script (with a `ffmpeg` invocation exaxmple). This will still have to be done.
- [ ] We should find out why it worked for me and did not for you.

pljones:
> - [ ] `--streamto` does not fail when used in Jamulus as a client mode, yet, it does not work there. I think we should fail in src/main.cpp.
> - [ ] The current code still uses the `ffmpeg`-focused approach which requires command line splitting. I think the plan was to go for the generic approach and add an appropriate example script (with a `ffmpeg` invocation exaxmple). This will still have to be done.
> - [ ] We should find out why it worked for me and did not for you.

👍 

> - [ ] When ffmpeg crashes (e.g. due to an icecast server restart), streaming will stop with no way to restart it besides re-connecting all clients and/or restarting the Jamulus server. Not sure, if/how this could be improved. We could think about restarting on failure, but this adds more complication and risks crash loops.

Personally, I'd leave this out for now.  I need to get a proposal written for better server management. For example, at the moment, the SIGUSR1/2 for the jam recorder isn't portable.  If the jam server runs out of space, it simply raises an exception to alert the server operator (that is, it deliberately crashes the server).  Neither is particularly nice.

So I'm hoping to have something like a separate "management" thread that non-core service can communicate with.  That should enable better and deeper status monitoring and control without impacting the main server too much.

(I'm hoping a similar approach can be taken with the client, too, to allow a nicer "headless" mode there.)

(I live in hope my day job work load will ease a bit so I've brain-space left to work out the details, along with everything else!)

olibac:
Would be possible to redirect every client channel arriving to server to ReaRoute? (this is related to #1305 @ann0see mentioned a few days ago).
The same way every channel output is redirected to a file when recording maybe it could be redirected to ReaRoute (or any virtual cable) so you can use Reaper (or any DAW) to mix every channel, add effects (eq, compressor, etc) and deliver final streo mix to a streaming service or to record live session with a better sounding mix.
I think it could be useful for audio techs as they can have more control on the "tracks" in order to improve quality of the final mix.

pljones:
The client can direct all of it's two stereo outputs wherever you want.

It only _has_ the stereo mix.  That's not going to change in the lifetime of "Jamulus" as "Jamulus".  If you need a multi-channel, locally mixed solution, you'll need to look elsewhere.

olibac:
@pljones Thank you for the response. @ann0see explained to me the mixing is done in the server, that's why I am asking if it can be done in the server (not the client). The same way every channel in the server is redirected to a file when recording audio maybe it could be redirected to a virtual cable in order to be able to mix in a DAW.

pljones:
The server isn't attached to any audio hardware currently.  Doing so would slow it down.  Which would mean the number of clients a single server could support would drop.  The jam recorder output each audio file unsynchronised.  If you're trying to get multichannel realtime audio for mixing externally, you need it synchronised, which means keeping in time with an external clock.

Again, that's unlikely to happen.  I've already said, someone could take the existing jam recorder quite easily and replace each file with a stream.  What each stream was attached to would then be up to the user.  But they wouldn't be "in time with each other".

olibac:
Thanks for the clarification about synchronization. And thanks for the efforts too.

WolfganP:
I did some tests with Reastream VST plugins on the server side (see https://github.com/jamulussoftware/jamulus/issues/146#issuecomment-687891312 and subsequent comments) but never had the time to make a workable prototype with the server (I even worked out to develop a Wireshark dissector for the protocol https://github.com/WolfganP/reastream-wireshark but that was the end of it due other time consuming tasks)

hoffie:
@dingodoppelt Are you still planning to work on this? If not, would you be ok with someone else taking over?

dingodoppelt:
@hoffie 
I actually worked on the streamer by introducing a protocol message to turn streaming on and off on the fly. but that commit went into my old code since i have problems getting the changes done to this pull request to work. i am using my streaming code on about 20 servers and combined it with a feature to chat with the outside world for which I had to introduce a second protocol message. that is not something i'd like to have merged since it is too much of a "special usecase" and hacked together to have something working for me. I'd rather see something like an API (accessible through a management thread) and dedicated protocol messages for controlling things.
I really think that extensions like this are better integrated into something like @pljones proposed:

> So I'm hoping to have something like a separate "management" thread that non-core service can communicate with. That should enable better and deeper status monitoring and control without impacting the main server too much.

I thought that this pr might be better off being held back until something like a management thread is introduced. that would make the stream or a chat with the outside modules or plugins. I don't really know if it is wise to merge an in-between since ffmpeg would have to go first ( and i'm still not a programmer ;)
We should first make clear what is really necessary to have a stream and possible future extensions working without messing up the code too much.

hoffie:
Thanks for your reply. In my opinion, it should be possible to bring this into mergable state with little effort. https://github.com/jamulussoftware/jamulus/pull/967#issuecomment-803569540 outlines what should be done, IMO.

I don't know how much demand for this feature there is, but I think we should either move this forward for the next+1 (3.9.0) release or close it for now.

While I agree with @pljones that having such a management interface would be really nice, I doubt that it will be as simple as this PR, so it might make take some more time.

What do people think? Is there demand? (Maybe show via a thumbs-up reaction on this comment)

If there is demand and you (@dingodoppelt) don't plan to move this any further, then someone else could possibly step in and try to finish it?

ann0see:
I think this could go to a feature branch and be discussed later. A management interface etc. could be created - based on #1975

ann0see:
@dingodoppelt I now pushed it to a feature branch: https://github.com/jamulussoftware/jamulus/tree/feature/ffmpeg-streamer2 therefore closing. I think @dtinth is working on something related to this.

ann0see:
As soon as the management interface (JSON-RPC) is merged, the feature branch could be revived: https://github.com/jamulussoftware/jamulus/blob/feature/ffmpeg-streamer2/README.md

PR #968, opened 2021-02-07
feature-req925 recording indicator

dcorson-ticino-com:
New function SetMixerBoardDeco sets title color and background according
to recording state and skin.
Called on recorder state change, skin change and disconnect

### DIFF
--- a/src/audiomixerboard.h
+++ b/src/audiomixerboard.h
@@ -218,8 +218,8 @@ public:
 
     void        SetChannelLevels ( const CVector<uint16_t>& vecChannelLevel );
 
-    void        SetRecorderState ( const ERecorderState newRecorderState );
-    ERecorderState  GetRecorderState () { return eRecorderState; };
+    void            SetRecorderState ( const ERecorderState newRecorderState );
+    ERecorderState  GetRecorderState() { return eRecorderState; };
 
     void        SetAllFaderLevelsToNewClientLevel();
     void        StoreAllFaderSettings();
--- a/src/audiomixerboard.h
+++ b/src/audiomixerboard.h
@@ -219,6 +219,8 @@ public:
     void        SetChannelLevels ( const CVector<uint16_t>& vecChannelLevel );
 
     void        SetRecorderState ( const ERecorderState newRecorderState );
+    ERecorderState  GetRecorderState () { return eRecorderState; };
+
     void        SetAllFaderLevelsToNewClientLevel();
     void        StoreAllFaderSettings();
     void        LoadAllFaderSettings();
--- a/src/audiomixerboard.h
+++ b/src/audiomixerboard.h
@@ -196,34 +196,34 @@ public:
 
     virtual ~CAudioMixerBoard();
 
-    void        SetSettingsPointer ( CClientSettings* pNSet ) { pSettings = pNSet; }
-    void        HideAll();
-    void        ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInfo );
-    void        SetServerName ( const QString& strNewServerName );
-    QString     GetServerName() { return strServerName; }
-    void        SetGUIDesign ( const EGUIDesign eNewDesign );
-    void        SetDisplayPans ( const bool eNDP );
-    void        SetPanIsSupported();
-    void        SetRemoteFaderIsMute ( const int iChannelIdx, const bool bIsMute );
-    void        SetMyChannelID ( const int iChannelIdx ) { iMyChannelID = iChannelIdx; }
+    void            SetSettingsPointer ( CClientSettings* pNSet ) { pSettings = pNSet; }
+    void            HideAll();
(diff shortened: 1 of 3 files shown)


src/audiomixerboard.h
```diff
@@ -219,6 +219,8 @@ class CAudioMixerBoard :
     void        SetChannelLevels ( const CVector<uint16_t>& vecChannelLevel );
 
     void        SetRecorderState ( const ERecorderState newRecorderState );
+    ERecorderState  GetRecorderState () { return eRecorderState; };
```
ann0see, line 222: The style here's a bit odd

src/clientdlg.cpp
```diff
@@ -39,6 +39,8 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     pSettings           ( pNSetP ),
     bConnectDlgWasShown ( false ),
     bMIDICtrlUsed       ( !strMIDISetup.isEmpty() ),
+    eLastRecorderState  ( RS_UNDEFINED ),       // for SetMixerBoardDeco
+    eLastDesign         ( GD_ORIGINAL ),        //
```
ann0see, line 43: Not sure about an empty comment here.

src/clientdlg.cpp
```diff
@@ -1354,3 +1357,40 @@ rbtReverbSelR->setStyleSheet ( "" );
...
+
+void CClientDlg::OnRecorderStateReceived (  const ERecorderState newRecorderState )
+{
+    MainMixerBoard->SetRecorderState ( newRecorderState );
+    SetMixerBoardDeco( newRecorderState, pClient->GetGUIDesign() );
+}
+
+void CClientDlg::OnGUIDesignChanged()
+{
+    SetGUIDesign ( pClient->GetGUIDesign() );
```
ann0see, line 1369: The code style seems to be a bit inconsistent here.

ann0see (changes requested):
Thanks for the new PR! It seems to work now. I didn't analyse it yet, but from my first impression, there are some smaller style changes you should make. Will test the code if I have more time.

dcorson-ticino-com:
Just pushed with the corrected style (I hope)
Don

src/audiomixerboard.h
```diff
@@ -218,7 +218,9 @@ class CAudioMixerBoard :
 
     void        SetChannelLevels ( const CVector<uint16_t>& vecChannelLevel );
 
-    void        SetRecorderState ( const ERecorderState newRecorderState );
+    void            SetRecorderState ( const ERecorderState newRecorderState );
```
ann0see, line 221: Thanks for the changes! It still looks a bit odd: The function names are not aligned. So you'd need to add spaces to every other function so that all names are over each other.

src/clientdlg.cpp
```diff
@@ -1354,3 +1357,53 @@ rbtReverbSelR->setStyleSheet ( "" );
...
+    SetMixerBoardDeco ( MainMixerBoard->GetRecorderState(), pClient->GetGUIDesign() );
+}
+
+void CClientDlg::SetMixerBoardDeco(  const ERecorderState newRecorderState, const EGUIDesign eNewDesign  )
+{
+    // return if no change
+    if( (newRecorderState == eLastRecorderState) && (eNewDesign == eLastDesign) )
```
ann0see, line 1376: I think you'll need a space between if and ( and also between ( and newRecorderState etc.

ann0see (changes requested):
Thanks for your changes. I think the new linter action (which is in development) is needed since the jamulus code style is odd. 

I've added a few other comments.

ann0see:
The functionality is ok. Didn't follow the issue, so I can't comment on that. For me, the red tone looks a bit too bright. But that's probably something the UI team (you also belong to) already discussed.

ann0see (approved):
Seems to be good now.

src/clientdlg.cpp
```diff
@@ -1354,3 +1357,53 @@ rbtReverbSelR->setStyleSheet ( "" );
...
+{
+    SetGUIDesign ( pClient->GetGUIDesign() );
+    SetMixerBoardDeco ( MainMixerBoard->GetRecorderState(), pClient->GetGUIDesign() );
+}
+
+void CClientDlg::SetMixerBoardDeco(  const ERecorderState newRecorderState, const EGUIDesign eNewDesign  )
+{
+    // return if no change
+    if ( ( newRecorderState == eLastRecorderState ) && ( eNewDesign == eLastDesign ) )
+        return;
```
pljones, line 1377: Generally this would not be a bare naked `return` - it would have its own `{}` block around it.  Letting it pass.

pljones (approved):

ann0see:
@dcorson-ticino-com Could you please add yourself to the contributor list and add this feature to the Changelog?

Issue #434, opened 2020-07-05
Close Server Connection to Client - Feature request for moderating jam participant behaviour


alanhahn:
Allow the server admin to close individual client connections to stop disruptions. In other words, the jam coordinator can kick someone out of the jam for bad behaviour.

In the GUI I envision a right mouse click on the client's entry and, in addition to "What's This?", there is a "Close Connection" choice.

corrados:
I recommend to setup a private server. That solves the issue.

alanhahn:
It doesn't solve the issue if the location of the private server is made available, unintentionally, to those who become disruptive. Or, there is a good reason for maintaining the server as public even if it attracts bad behaviour.

corrados:
> private server is made available, unintentionally

That usually does not happen if you know your community well.

There are several other techniques to solve the issue:
- If your choir has, e.g., 12 members -> set the maximum allowed number of clients to 12 in the server. Then if all your choir members are connected, the server is full and nobody else can join.
- Set all your choir members to Solo. Then you will hear only those.
- Move the fader of the unwanted client to the bottom or activate the Mute button. Jamulus will store this setting.

pljones:
This request has been mentioned a few times, so I thought it might be worth raising.

The problem with "set the number of clients" is that someone disruptive might get in a busy out your server.  Yes, you can bounce with disconnectall (or, better, always run with disconnectonquit) but it's more disruption for everyone else.

Asking every member to mute the disruptive person doesn't stop them busying out the server.

Asking every member of a choir to solo every member of the choir every time they join the server is disruptive in itself.  And if someone new joins the choir, everyone has to remember to solo them so they can hear them saying "Hello".  (And still doesn't stop the server being busied out.)

A "block IP" from the server GUI would allow an admin to take direct action against a disruptive user and, in most cases, solve the problem very simply.  Anyone managing to bomb a server from multiple IPs is a more dangerous level of threat than I'd suggest needs covering here.

alanhahn:
Thanks, Peter for fleshing out what I am asking for. I did not think of the 'Block IP' option but that is better than a 'Disconnect' choice because it prevents the person from rejoining over and over again.

pljones:
[...]

Issue #299, opened 2020-05-27
Add protection for private servers


ann0see:
Would it be possible to add some kind of protection to private servers so that not everybody who knows the IP can join?

Adding a password for private servers would enable real private sessions and increase overall security and privacy.
Maybe it's not easy to implement with udp or would add some overhead?
Maybe a ip unblock after somebody has input the correct server password would be an option?

corrados:
Jamulus shall be a free jam session tool. Nothing is more frustrating if a server is listed in the server list but it is protected by a password.
If you have setup a private server, then this is enough protection. Nobody will guess the IP number and the port number you have chosen.

genesisproject2020:
Suggestion. A server listed has to be open. Adding a private flag and when it's set on a server the server will be not listed.

ann0see:
I totally agree that public servers shouldn't have a password. But especially for private servers some kind of protection would be good (since a portscan is easy). Maybe I'm a bit paranoid but I don't like the idea of somebody being able to connect to my private server and possibly interupt or disturb a jam session.

corrados:
I use the software since 14 years now. I regularily use private server for some rehearsals. I had never the case that some stranger entered that private server.

gilgongo:
> Adding a private flag and when it's set on a server the server will be not listed.

Interesting,  because it implies you wouldn't need to set up a private server (with port forwarding etc.) as you do today. On the other hand I don't think it would work with a 150 server limit to each Central Server. They'd probably just all fill up.

genesisproject2020:
> > Adding a private flag and when it's set on a server the server will be not listed.
> 
> Interesting, because it implies you wouldn't need to set up a private server (with port forwarding etc.) as you do today. On the other hand I don't think it would work with a 150 server limit to each Central Server. They'd probably just all fill up.

Could be a list that doesnt update that often and then a bigger list that will not show up in the client at all.

WolfganP:
If it's a private server, why listing and hiding it? I don't see any purpose on it if the intent is being non discoverable. Just host it without registering into a central server.

Regarding the port scan, I don't see what adding a password will add in terms of security of the session:
*if it's not an authorized user that received the IP:port via a side channel (sharing via mail or IM group), most likely s/he also received the password as well.
[...]

PR #974, opened 2021-02-09
eduTools (Features for education contexts): Waiting room; disable chat;...

ann0see:
DISCUSSION ONLY. Please do not merge this.

This PR includes the changes of my fork (eduTools branch) with some features added having educational needs in mind. (If we add some features of this, we can of course port them)

1. Support chat commands
2. Support waiting room (just hacked a simple way of authentication together)
3. Add support for disabling chat via chat commands
4. Do not automatically log IPs (= hide them on the server will probably not be ported to upstream. See #825)

I know that
@Brian-McBride and @gene96817 might be interested in this. 

Related to #299 #434

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -72,6 +72,8 @@ int main ( int argc, char** argv )
     bool         bNoAutoJackConnect          = false;
     bool         bUseTranslation             = true;
     bool         bCustomPortNumberGiven      = false;
+    bool         bLogIP                      = false;
+    bool         bEduModeEnabled             = false;
     int          iNumServerChannels          = DEFAULT_USED_NUM_CHANNELS;
     quint16      iPortNumber                 = DEFAULT_PORT_NUMBER;
     ELicenceType eLicenceType                = LT_NO_LICENCE;
@@ -85,6 +87,7 @@ int main ( int argc, char** argv )
     QString      strServerInfo               = "";
     QString      strServerPublicIP           = "";
     QString      strServerListFilter         = "";
+    QString      strEduModePassword          = "";
     QString      strWelcomeMessage           = "";
     QString      strClientName               = "";
 
@@ -136,6 +139,19 @@ int main ( int argc, char** argv )
             eLicenceType = LT_CREATIVECOMMONS;
             qInfo() << "- licence required";
             CommandLineOptions << "--licence";
+            continue;
+        }
+
+
+        // Log whole IP-Adresses on server ----------------------------------------------------
+        if ( GetFlagArgument ( argv,
+                               i,
+                               "--logip",
+                               "--logip" ) )
+        {
+            bLogIP = true;
+            qInfo() << qUtf8Printable( QString( "- Logging full IP adresses." ) );
+            CommandLineOptions << "--logip";
             continue;
         }
 
@@ -443,6 +459,20 @@ int main ( int argc, char** argv )
             continue;
         }
 
+        // Education mode password ---------------------------------------------
+        if ( GetStringArgument ( argc,
+                                 argv,
+                                 i,
+                                 "--edumodepassword", // no short argument
+                                 "--edumodepassword",
+                                 strArgument ) )
+        {
+            strEduModePassword = strArgument;
+            bEduModeEnabled    = true;
+            qInfo() << qUtf8Printable( QString( "- enabled Edu-Mode with password " ) );
+            CommandLineOptions << "--edumodepassword";
+            continue;
+        }
 
         // Server welcome message ----------------------------------------------
         if ( GetStringArgument ( argc,
@@ -715,10 +745,13 @@ int main ( int argc, char** argv )
                              strServerListFilter,
                              strWelcomeMessage,
                              strRecordingDirName,
+                             strEduModePassword,
                              bDisconnectAllClientsOnQuit,
                              bUseDoubleSystemFrameSize,
                              bUseMultithreading,
+                             bLogIP,
                              bDisableRecording,
+                             bEduModeEnabled,
                              eLicenceType );
 
 #ifndef HEADLESS
@@ -816,8 +849,10 @@ QString UsageArguments ( char **argv )
         "                        (or 'localhost' to be a server list)\n"
         "  -f, --listfilter      server list whitelist filter in the format:\n"
         "                        [IP address 1];[IP address 2];[IP address 3]; ...\n"
+        "  --edumodepassword     enable Edu-Mode and set following admin password\n"
         "  -F, --fastupdate      use 64 samples frame size mode\n"
         "  -l, --log             enable logging, set file name\n"
+        "   --logip              Log full IP adresses\n"
         "  -L, --licence         show an agreement window before users can connect\n"
         "  -m, --htmlstatus      enable HTML status file, set file name\n"
         "  -o, --serverinfo      infos of this server in the format:\n"
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -8,16 +8,16 @@
  *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software
- * Foundation; either version 2 of the License, or (at your option) any later 
+ * Foundation; either version 2 of the License, or (at your option) any later
  * version.
  *
- * This program is distributed in the hope that it will be useful, but WITHOUT 
+ * This program is distributed in the hope that it will be useful, but WITHOUT
  * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
- * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more 
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
  * details.
  *
  * You should have received a copy of the GNU General Public License along with
- * this program; if not, write to the Free Software Foundation, Inc., 
+ * this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
  *
 \******************************************************************************/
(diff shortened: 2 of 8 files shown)


pljones:
There's little point hiding IP addresses in the Jamulus log.  Routers and firewalls likely both log them.

ann0see:
> There's little point hiding IP addresses in the Jamulus log. Routers and firewalls likely both log them.

Yes sure. I think we had this discussion somewhere. This PR is basically just a diff and starting point for collecting ideas concerning the education topic.

gene96817:
@ann0see   Where and how do you want these topics discussed?

ann0see:
I would like to start a discussion on how to make Jamulus more education friendly. I know that there are discussions about adding passwords etc. but my way of "solving" the issue was that I implemented a waiting room feature. Surely some of the ideas/code can be included in some way in further projects etc. That's why I opened this PR: Basically discussing my changes and ways of implementing the waiting room, if you could/should include some here, what could be improved,...

pljones:
The first questions to be asked for any new feature: "Why inside Jamulus?  Can it not be done as well if not better elsewhere?  How does putting it inside Jamulus make it significantly more useful than the alternatives?"

ann0see:
Yes. I plan to add a few review comments here and maybe clean the history.

gene96817:
I'd like to avoid handling a series of single-purpose features. It would be helpful to identify clusters of similar or related requirements.

I believe the desire for waiting room, private sessions, scheduling private sessions, and authentication can/should be consolidated into a common solution, I believe these should be outside of core Jamulus. I think of these as connection/admission management. I also would like to see the needs behind these features implemented to encourage servers to be public so that the unscheduled or unused tine is available to the community. (For example, my server utilization is less than 5%. I was delighted that a band was using it because it was the only server within 3000km. I can have a reason to make the server private. That would be a loss to the community.)

src/server.cpp
```diff
@@ -1135,6 +1181,13 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
         // Mono target channel -------------------------------------------------
         for ( j = 0; j < iNumClients; j++ )
         {
+            // check if client is blocked.
+            // one can always hear himself --> if we currently iterate through the blocked client we do send him audio
+            if ( ( j != iCurChanID && vecChannels[j].IsBlocked() ) || ( j != iCurChanID && currentClientBlocked ) )
+            {
```
ann0see, line 1187: This is the interesting part: 

In short: 

If you're blocked, nobody can hear you and you can not be heared. You will only be able to hear yourself. This makes it possible to have real private sessions in terms of "although you're connected to a server, nobody can come into your rehearsal"

src/server.cpp
```diff
@@ -1394,20 +1464,147 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int      iCurChanID
...
-        QTime::currentTime().toString ( "hh:mm:ss AP" ) + ") <b>" +
+        "<small style='color:" + sCurColor + ";'>[" +
+        QTime::currentTime().toString ( "hh:mm" ) + "] <b>" +
         ChanName.toHtmlEscaped() +
-        "</b></font> " + strChatText;
+        "</b></small><br />" + strChatText;
 
+    // check if chat is disabled
+    if ( bEduModeEnabled && EduModeIsFeatureDisabled( EDU_MODE_FEATURE_CHAT ) )
+    {
```
ann0see, line 1474: If the chat was disabled by the admin --> send an error message and don't submit it to others.

Why? Some children might spam the chat (this also happens in Zoom etc). Currently nothing can be done against that.

src/server.cpp
```diff
@@ -1394,20 +1464,147 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int      iCurChanID
...
+                vecChannels[i].CreateChatTextMes ( strActualMessageText );
+            }
+        }
+    }
 
-    // Send chat text to all connected clients ---------------------------------
-    for ( int i = 0; i < iMaxNumChannels; i++ )
+}
+void CServer::InterpretAndExecuteChatCommand ( const int iCurChanID,
+                                               const QString& strChatText )
```
ann0see, line 1492: A function analyzing incoming chat messages and basically enabling the administrative functions

src/server.cpp
```diff
@@ -1394,20 +1464,147 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int      iCurChanID
...
+{
+    QString strChatCmd = strChatText;
+    strChatCmd.remove ( 0, 3 );
+    QString ChanName = vecChannels[iCurChanID].GetName(); // to greet the user
+
+    if ( strChatCmd == strEduModePassword && bEduModeEnabled )
```
ann0see, line 1498: Very insecure way of authentication (plain text,...).

src/server.cpp
```diff
@@ -1394,20 +1464,147 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int      iCurChanID
...
+        EduModeSetFeatureDisabled( EDU_MODE_FEATURE_CHAT, true );
+        vecChannels[iCurChanID].CreateChatTextMes ( "Disabled Chat." );
+    }
+    else if ( strChatCmd.startsWith ( "enableChat" ) && vecChannels[iCurChanID].IsAdmin() )
+    {
+        EduModeSetFeatureDisabled( EDU_MODE_FEATURE_CHAT, false );
+        vecChannels[iCurChanID].CreateChatTextMes ( "Enabled Chat." );
+    }
+    else if ( strChatCmd.startsWith ( "bl" ) && vecChannels[iCurChanID].IsAdmin() )
+    {   // block user
```
ann0see, line 1539: Manually block a user --> do not mix him into the mix of others & do not let him hear the others.

src/server.cpp
```diff
@@ -1394,20 +1464,147 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int      iCurChanID
...
+        }
+
+    }
+    else if ( strChatCmd.startsWith( "enableWaitingRoom" ) && vecChannels[iCurChanID].IsAdmin() )
+    {
+        EduModeSetFeatureDisabled( EDU_MODE_FEATURE_WAITINGRM, false );
+        vecChannels[iCurChanID].CreateChatTextMes ( "Enabled waiting room. New users will not be able to hear others." );
+    }
+    else if ( strChatCmd.startsWith( "disableWaitingRoom" ) && vecChannels[iCurChanID].IsAdmin() )
+    {
```
ann0see, line 1592: It should be possible to disable the waitingroom.

ann0see (commented):
Some comments on the code.

Why should some kind of moderation functions be part of Jamulus?

1. Every video conferencing tool supports features like this. Why? They have been abused, people behave badly,... 
2. Some educational institutions might block Jamulus because these features are missing
3. There's no other way I know in Jamulus to have a real private rehearsal. Although private servers provide some kind of protection, it's security through obscurity. If somebody shares the IP, everybody could join (= Zoom bombing but in Jamulus). Nowadays, in educational context, there must be some way of ensuring that children are safe and nobody unknown or unwanted can join.

So what's the intention behind these features?

It should be possible to have fully private sessions (= ensure that nobody else can join or listen to your rehearsal. This is needed in some places where children are involved). If we can solve this on public servers: This would be the best solution.

src/main.cpp
```diff
@@ -140,6 +143,19 @@ int main ( int argc, char** argv )
         }
 
 
+        // Log whole IP-Adresses on server ----------------------------------------------------
+        if ( GetFlagArgument ( argv,
```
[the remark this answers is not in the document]
gene96817: What does the IP address do for us? It is possible to create a proxy for IP that is anonymized.
Some form of IP address could be needed for diagnostics.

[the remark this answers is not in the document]
ann0see: > Some form of IP address could be needed for diagnostics.

Agree. The problem was already discussed in #576.

This can of course be removed (and probably will). I personally would like to have this (optional) function in Jamulus since it would then allow the server admin to decide if he wants to hide IPs or not. For me, this might still be a privacy issue.

gene96817:
How willing are we (the developer community) willing to use DTLS for secure UDP sessions? This could be part of a secure architecture. This would eliminate spoofing and hijacking of a session.

pljones:
OK, fundamental to this, as far as I can see, are two functions:
- similar to the licence, prevent a connection taking part on the server
Ideally, it would work the same way - display the chat message and block further action - just without the pop-up to accept the licence enabling access
- allow a certain keyword to be used to cause "magic" to happen - i.e. if the keyword is used the rest of the command can enable or disable a user's access

So on connecting, the "owner" would need to "magic enable owner" to enable themselves and then "magic enable other1" to enable someone who has a name of "other1".  Of course, if they are one of three "NoName" users, either enable isn't supported or their IP and port number would be needed.

Similarly, "owner" will be able to incant "magic disable other2" to prevent further use of the server.

Finally, you're suggesting two more "broad" commands "magic open" to prevent the need for "magic enable whoever" and "magic close" to re-enable it.

The "magic" keyword would be configurable at server start up.

pljones:
So, the trick.  We've said anyone connecting is blocked unless someone issues the magic command.  However, the first person to connect _is blocked_ and that means we need to be processing _everyone's_ chat incoming, in case it's the magic command to enable yourself.  Which means the client needs to be sending them.  I'm not sure it is when the licence is what's blocking.  But it could - just they'd not get echoed with this change, as the server would need to know to ignore the client.

ann0see:
> similar to the licence, prevent a connection taking part on the server
Ideally, it would work the same way - display the chat message and block further action - just without the pop-up to accept the licence enabling access


Yes. That's the idea behind this PR and way of solving the problem. 

After having thought a bit more about this, another option would simply be a button or setting which disables mixing your audio in the mix of clients you select (and maybe even hide them from the mixer board.) basically a forced mute on your side. Ideally, you could say "do not mix my audio unless I select that this person can hear me" and this state could be saved client side, like the levels. This would be a small, incremental feature which at least adds some protection.

gene96817:
I'd like to suggest a different framework.
Consider adding a "participation manager" based on OpenPAM. The logic and features of the participation manager is outside of Jamulus, although we could provide a very simple sample. There messages between Jamulus and the participation manager/OpenPAM would be:
1- user request (could be <null>, a string, or a n-tuple) to OpenPAM
2- message for display to user from OpenPAM
3- additional user response to OpenPAM
4- connection accept/reject from OpenPAM to Jamulus

The participation manager, at the simplest, is just a list of credentials (or UID). 
A more complicated participation manager would receive an n-tuple  (e.g. jam session name, UID). It could also make decisions based on time-of-day. All of this is managed by a different UI separate from Jamulus. Jamulus server owners can create simple or complicated systems independent of the Jamulus code.

This scheme can be configured to meet the different scenarios requested for private sessions and education applications.

ann0see:
I've just created a branch on this repo containing the changes of this PR so everybody can edit it. I will therefore close this PR now.

ann0see:
Since I deleted the branch a while ago (not sure why), I've rebased and pushed it again here: https://github.com/jamulussoftware/jamulus/tree/feature/a0/eduTools

I think with JSON-RPC at least some of the thoughts from this branch can be reused.

Issue #847, opened 2021-01-16
When not connected: "Please connect to a server to hear yourself and others"


dcorson-ticino-com:
I see repeated problems with new users not realising that it is necessary to connect to a server to hear ones self and set the input volume.  I see this both on the forum and among the groups I am helping to get up-and-running with Jamulus.
A startup wizard would be a neat solution, but an even simpler solution could be a message in the center of the mixer window when it is empty, i.e. not connected to a server, to the effect of "Please connect to a server to hear yourself and others".
A second Connect button could be placed under the message.
Further the Input level "LEDs" could be grayed out when not connected to pass on the message you need to do something before you can hear yourself.

Strangely enough the biggest problem I have seen with the new groups around me is setting the input volume correctly.  Not too quiet so you can't be heard, but not too loud to cause clipping distortion, which is always blamed on poor audio quality in Jamulus.  Once the input volume is set right everyone is smiling.
Don

ann0see:
> Further the Input level "LEDs" could be grayed ou

I already thought and implemented this once in my fork.

I think it's here: https://github.com/ann0see/jamulus/commit/fb5e038af87fd6094422005895b165ef1b470e45

corrados:
There is already a feature request for showing input levels when not connected, see: https://github.com/corrados/jamulus/issues/712

corrados:
> Further the Input level "LEDs" could be grayed out when not connected to pass on the message you need to do something before you can hear yourself.

Well, the LEDs are black in idle right now. If a new Jamulus user (i.e. a user which has not used Jamulus before at all) sees the current LED color, I think it does not make any difference for him if they are black or gray. Both colors could indicate that the LEDs are disabled.

dcorson-ticino-com:
Yes I saw 712, but don't really see the need to be able to set levels when not connected.
I think just adding this message would be simpler than adding a new data path for the sound when offline and be just as effective.

Your objection about the leds is true, and graying the whole field is maybe not so simple, I am a programmer, but have no experience in Windows object oriented systems.  Throw assembler and microcontrollers at me.

pljones:
> I am a programmer, but have no experience in Windows object oriented systems. Throw assembler and microcontrollers at me.

I was a mainframe programmer when I started looking at Jamulus code...  (Admittedly, I took a while to start looking at the UI code... but it's all just code.)

ann0see:
[...]

Issue #973, opened 2021-02-09
Move ASIO-Setup button above Driver selection 


ann0see:
Just had a thought about the ASIO Setup button. The current placement might not be obvious for the user. Therefore I think we should move (and maybe rename it to ASIO device Setup). This might make setting devices up more obvious.

pljones:
The action of the button is dependent upon selection of the device.  I've six ASIO drivers, each with their own set up.  I'd move it above the "Input Channel Mapping", perhaps, but still below the device selection.

ann0see:
Ok. I agree. That's probably a better position. I'll look into it now.

ann0see:
There's still the question about renaming the button. 

> The action of the button is dependent upon selection of the device

Do you have an idea about something generic which clearly tells the (new) user that in order to configure his in/outputs in ASIO4ALL he needs to click on this button?

mulyaj:
I think "ASIO Device Setup" works.  Maybe you could also box the device dropdown and setup together, under the label device? 

Something like this?
![image](https://user-images.githubusercontent.com/63881221/107885015-d8369300-6ec5-11eb-9c34-03c5601b0de4.png)

ann0see:
Yes. This looks logical. Can you open a PR to this branch with these changes?

PR #983, opened 2021-02-12
Add connect to server help message

ann0see:
Might solve #847

Todo: Wording of the sentence, translation (but that's something for the next release). 

Maybe we could use: Setup your audio, connect and start jamming!

Not Connected to a server:

ConnectServer

Connected:
Connected

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -710,6 +710,9 @@ void CClientDlg::OnConnectDlgAccepted()
         // get the address from the connect dialog
         QString strSelectedAddress = ConnectDlg.GetSelectedAddress();
 
+        // hide label connect to server
+        lblConnectToServer->hide();
+
         // only store new host address in our data base if the address is
         // not empty and it was not a server list item (only the addresses
         // typed in manually are stored by definition)
@@ -1238,6 +1241,9 @@ void CClientDlg::Disconnect()
     TimerSigMet.stop();
     lbrInputLevelL->SetValue ( 0 );
     lbrInputLevelR->SetValue ( 0 );
+
+    // show connect to server message
+    lblConnectToServer->show();
 
     // stop other timers
     TimerBuffersLED.stop();
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -80,10 +80,12 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     lbrInputLevelL->setAccessibleName        ( strInpLevHAccText );
     lbrInputLevelL->setAccessibleDescription ( strInpLevHAccDescr );
     lbrInputLevelL->setToolTip               ( strInpLevHTT );
+    lbrInputLevelL->setEnabled               ( false );
     lbrInputLevelR->setWhatsThis             ( strInpLevH );
     lbrInputLevelR->setAccessibleName        ( strInpLevHAccText );
     lbrInputLevelR->setAccessibleDescription ( strInpLevHAccDescr );
     lbrInputLevelR->setToolTip               ( strInpLevHTT );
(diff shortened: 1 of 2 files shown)


ann0see:
@dcorson-ticino-com I think this would be a "quick" fix for your feature request. I've now also added the "gray out input levels if not connected" feature. Probably a re-arrangement of the UI would be good, but I think that's far from finished now. Also I'd personally like to have incremental changes with this and #973 being small ones.

pljones:
"Setup" should be two words.
"Set up your audio" only happens once, so shouldn't be a "nag" like this - first time installation hand-holding only.

ann0see:
> I'd also prefer it as replacing "Server" in the group box title, rather than as a separate label like you've shown.

The problem with this would be that it's probably a lot less visible.

pljones:
The first time user process shouldn't be part of Jamulus at all.  It's only needed once.

dcorson-ticino-com:
To reduce the nagging a bit I would make the message only shown at program start-up before the first connection.
I agree that the audio setup should only be at installation time.  Would a bool in the ini file be an adequate way of seeing this?
D

gene96817:
Sometimes another application might make changes to the underlying system that affects Jamulus. Will there be an easy way to repeat the first time set up? This is also useful for a new user that didn't get the right results in their first connection. Then we/I need to handhold them through the first time set up (again) to get the process right.

ann0see:
@gene96817 unfortunately the first time setup doesn't exist yet. Can you open an issue and describe what you'd like? We can then discuss everything there.

ann0see:
Just changed Setup to Set up. Any chance this will be part of the next release?

pljones (approved):

ann0see:
@pljones I hope you know that this will also gray out the level meter if you're not connected.

pljones:
Yes, I saw that.  I think it makes sense.  When Jamulus isn't connect to a server, it _should_ have the sound card closed.  (On Windows, it doesn't have the ASIO driver open when it's not connected to the server.  It should _not_ connect to JACK except when connected to a server, either, to avoid putting load on the system unnecessarily.  No idea on MacOS.)

ann0see:
> I consider current behaviour ... unhelpful

That makes two of us. Probably a larger todo for the next release. 

What I would suggest:

If ASIO/Jack/... wasn't found or doesn't work (wrong sample rate,...): open Jamulus but give a clear warning what is wrong. Add a button which opens settings,... Do you want to open a new issue on this?

atsampson:
>  It should not connect to JACK except when connected to a server, either, to avoid putting load on the system unnecessarily

I'm not sure this is a huge worry - Jamulus doesn't do very much in its JACK callbacks when it's not connected to a server, so the additional load from this is pretty minimal. From a very rough measurement on my machine, jackd takes 0.7% CPU when nothing's connected, and Jamulus takes 0.3% when idling. For people like me who have Jamulus connected to non-default JACK ports, you'd also need some external mechanism for remaking your connections every time you connect to a server, which would be possible but a bit more work.

gene96817:
(I don't have a Windows machine so I can't reconfirm this behavior.)
It seems the Windows machines can only have one application access the microphone. This means it matters if Jamulus is open before Zoom or visa versa. If the closes its connection to the sound card, does this mean Jamulus might not be able to get a connection (because another application has taken control)? This is not an issue on Macs.

ann0see:
Did you try ASIO4ALL v2.13? For me, I think it partly "fixed" this error

softins (approved):

pljones:
> It seems the Windows machines can only have one application access the microphone.

This is completely dependent on the driver in use.

gene96817:
All those "dependent on" makes it so hard to give assistance.  :(

Issue #984, opened 2021-02-12
submodule checkout broken


nefarius2001:
**Describe the bug**
 PR #934 breaks git checkout

**reproduce**
git clone --recurse-submodules https://github.com/jamulussoftware/jamulus >>> FAIL
git clone --recurse-submodules git@github.com:jamulussoftware/jamulus.git >>> FAIL

This is mandatory for github autobuilds (#978), but slipped in Auto-Checks because those do not yet use submodules.

@pljones
What command is your PR aiming for? Can it be improved to fit your needs and have the above still work?

hoffie:
Thanks for the report. I can confirm. I think the `github.com/` part doesn't belong there. Gonna open a PR in a minute.

ann0see:
Thanks @hoffie!

pljones:
I shouldn't try to rush things so much...

PR #978, opened 2021-02-10
unified autobuild github-action

nefarius2001:
Work for #854:
single workflow 
- compiling on all platforms, including codeQL where possible => nice check for each PR
- separate codeQL-builds where necessary
- upload packs/binaries/installers as artifacts of the job (autodelete after 31 days) => low barrier for testing features on different platforms
- upload packs/binaries/installers to releases if triggered by tags starting with "r_" << still needs small tweaking
- updating packs/binaries/installers in a special release if triggered by tag named "latest" << still needs small tweaking

PR #986, opened 2021-02-12
Github Workflow CodeQL: Check out submodules

hoffie:
(As noticed by @nefarius2001)

- Submodules are part of official releases, so it might make sense
  to run them through CodeQL as well.
- This is supposed to detect breakage with submodule inclusion
  as observed in #984.

Docs: https://github.com/actions/checkout

I'm not sure if the modified CodeQL workflow will already trigger in this PR (I suspect it won't). If it does, it will probably (correctly) fail until #985 is merged. It would be great if it did because I could then rebase and we could check if the action works as intended and passes afterwards.

Note: I have never worked with Github actions before so it might be horribly wrong what I do.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -27,6 +27,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v2
+      with:
+        submodules: recursive
 
     - name: Install Qt on Linux
       run: |


hoffie:
> I'm not sure if the modified CodeQL workflow will already trigger in this PR (I suspect it won't). If it does, it will probably (correctly) fail until #985 is merged. It would be great if it did because I could then rebase and we could check if the action works as intended and passes afterwards.

Seems like it works -- CodeQL fails in this PR because the new CodeQL workflow seems to be used:
```  Error: fatal: repository 'https://github.com/github.com/google/oboe.git/' not found
  Error: fatal: clone of 'https://github.com/github.com/google/oboe.git' into submodule path '/home/runner/work/jamulus/jamulus/libs/oboe' failed
  Failed to clone 'libs/oboe' a second time, aborting
  Error: The process '/usr/bin/git' failed with exit code 1
```

I suggest waiting for #985 to be merged, will then rebase and then CodeQL should be green again.

nefarius2001:
#978 is close to ready, will fix qodeql

hoffie:
> #978 is close to ready, will fix qodeql

Ah, thanks for the pointer. Your PR seems to move the CodeQL part to the autobuild which already contains a submodule-enabled checkout. So, after your PR my PR becomes irrelevant.

If your PR is merged quickly, I'll close my PR as it becomes obsolete. If there is still some back and forth I'd be in favor of merging my PR temporarily to spot any problems in this area.

pljones:
Can this be closed?

hoffie:
Yes, it will be part of #978 already.

PR #989, opened 2021-02-13
Fix bad embedded quotes in strings.

softins:
Embedded quotes were included in strings by doubling them: `""`
This doesn't work in C++, so changed to `\"`


src/global.h
```diff
@@ -234,8 +234,8 @@ LED bar:      lbr
 // common tool tip bottom line text
 #define TOOLTIP_COM_END_TEXT             \
     "<br><div align=right><font size=-1><i>" + \
-    QCoreApplication::translate ( "global", "For more information use the ""What's " \
-    "This"" help (help menu, right mouse button or Shift+F1)" ) + \
+    QCoreApplication::translate ( "global", "For more information use the \"What's " \
```
dakhubgit, line 237: Do translations follow automatically via a different process or do they also need attention in this commit?

softins: Good call. I’m not yet familiar with the translation system, so will check.

ann0see (commented):
From my first impression, everything seems to be ok.

src/global.h
softins: Translation files updated to match and included in this PR. Tested on Linux.

ann0see (approved):

src/util.cpp
```diff
@@ -426,87 +426,87 @@ CAboutDlg::CAboutDlg ( QWidget* parent ) : CBaseDlg ( parent )
...
+        "<p>Melcon Moraes (<a href=\"https://github.com/melcon\">melcon</a>)</p>"
         "<p><b>" + tr ( "Dutch" ) + "</b></p>"
-        "<p>Jeroen Geertzen (<a href=""https://github.com/jerogee"">jerogee</a>)</p>"
+        "<p>Jeroen Geertzen (<a href=\"https://github.com/jerogee\">jerogee</a>)</p>"
         "<p><b>" + tr ( "Italian" ) + "</b></p>"
-        "<p>Giuseppe Sapienza (<a href=""https://github.com/dzpex"">dzpex</a>)</p>"
+        "<p>Giuseppe Sapienza (<a href=\"https://github.com/dzpex\">dzpex</a>)</p>"
         "<p><b>" + tr ( "German" ) + "</b></p>"
-        "<p>Volker Fischer (<a href=""https://github.com/corrados"">corrados</a>)</p>"
+        "<p>Volker Fischer (<a href=\"https://github.com/corrados\">corrados</a>)</p>"
```
pljones, line 502: Do we know if this is true?  Is anyone else looking to pick up the German if not (which I suspect is the case)?

ann0see: I assume this is no longer true. A german translator is missing. I don't want to do it myself since I'd like to focus on other things. I assume there are multiple other german/swiss/austrian people especially in the facebook group who could translate Jamulus. Could you please ask in the facebook group?

softins: I've just posted this in the FB group:
> The Jamulus development team is looking for a technical, native German speaker to maintain the German translations for the Jamulus program. The translation file already exists, so this only involves producing translations for new or updated text. Please PM me if you are interested.

ann0see: Thank you very much! If somebody PMs you (I hope so), he/she can just ping me somewhere here on GitHub or the organisation. Nevertheless, I can't help much since I'm only familiar with the translation of the wiki.

pljones (approved):
I'm assuming there's a reason the .qm files are there.

ann0see:
> I'm assuming there's a reason the .qm files are there.

Yes: see: https://github.com/jamulussoftware/jamulus/pull/989#discussion_r575677628

Once we do a new release, these files will probably need to be recreated. I saw that in the debian repo, some improvements have been made to this: https://sources.debian.org/patches/jamulus/3.6.2+dfsg1-2/use-lrelease.diff/

I will merge this now.

softins:
> I'm assuming there's a reason the .qm files are there.

The .qm files are generated from the .ts files, but there are no Makefile rules to do that (not sure why not) - it's a manual process by running `lrelease *.ts`, and the make process assumes the .qm files are already in place and up to date. Since the repo already contained the .qm files, I made sure they were updated and checked back in. That will need to happen again when we have the translations updated for the next release.

PR #1004, opened 2021-02-15
Add validation for --serverpublicip

hoffie:
https://github.com/jamulussoftware/jamuluswebsite/pull/290#issuecomment-778690300
cc @pljones

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -578,9 +578,17 @@ int main ( int argc, char** argv )
         qWarning() << "Mute my own signal in my personal mix is only supported in headless mode.";
     }
 
-    if ( !strServerPublicIP.isEmpty() && ( strCentralServer.isEmpty() || bIsClient ) )
+    if ( !strServerPublicIP.isEmpty() )
     {
-        qWarning() << "Server Public IP will only take effect when registering a server with a central server.";
+        QHostAddress InetAddr;
+        if ( !InetAddr.setAddress ( strServerPublicIP ) )
+        {
+            qWarning() << "Server Public IP is invalid. Only plain IP addresses are supported.";
+        }
+        if ( strCentralServer.isEmpty() || bIsClient )
+        {
+            qWarning() << "Server Public IP will only take effect when registering a server with a central server.";
+        }
     }
 
     // per definition: if we are in "GUI" server mode and no central server


src/main.cpp
```diff
@@ -578,9 +578,17 @@ int main ( int argc, char** argv )
         qWarning() << "Mute my own signal in my personal mix is only supported in headless mode.";
     }
 
-    if ( !strServerPublicIP.isEmpty() && ( strCentralServer.isEmpty() || bIsClient ) )
+    if ( !strServerPublicIP.isEmpty() )
     {
-        qWarning() << "Server Public IP will only take effect when registering a server with a central server.";
+        QHostAddress InetAddr;
+        if ( !InetAddr.setAddress(strServerPublicIP) )
```
ann0see, line 584: I think you'll need to add spaces before strServerPublicIP.

---

Another question:
.setAddress() probably works, but for me (not a Qt expert) it seems as if this isn't really a method which checks if an ip is valid?

ann0see (changes requested):

src/main.cpp
hoffie: > I think you'll need to add spaces before strServerPublicIP.

You're absolute right, thanks, fixed.

> .setAddress() probably works, but for me (not a Qt expert) it seems as if this isn't really a method which checks if an ip is valid?

Not a Qt expert by any means either, I've just used this method because it's documented to behave this way ([*"Returns true and sets the address if the address was successfully parsed; otherwise returns false."*](https://doc.qt.io/qt-5/qhostaddress.html#setAddress-5)) and because it's used in other [places](https://github.com/jamulussoftware/jamulus/blob/110da30a6b7b0211efbccd5acfe40dcc4245094f/src/protocol.cpp#L1966) in [Jamulus](https://github.com/jamulussoftware/jamulus/blob/110da30a6b7b0211efbccd5acfe40dcc4245094f/src/util.cpp#L989) for that as well. That's why I think it's fine. If someone has a better fit, I'm happy to change it. :)

ann0see: Ok. I‘d leave this conversation unresolved so that others can see it but will approve the PR.

ann0see (approved):

pljones (approved):
Very clear! :)

PR #1009, opened 2021-02-17
Client: Always update gain levels

### LINKED ISSUE #955: Restarting the server on the fly ignores set levels and mixes everybody at 100%
**Describe the bug**
Restarting the server sets levels to 100%

**To Reproduce**
restart a running server with clients on it and notice how it ignores the set gain levels and mixes everybody at 100%

**Expected behavior**
the server should mix everybody at the individually set levels

**Operating system**
linux, windows, macos

**Version of Jamulus**
3.6.2 and before

hoffie:
Previously, gain levels had only been updated when a channel's name changed. This lead to missing gain level updates when reconnecting to a restarted server. Therefore, always update gain levels when receiving a new channel list.

Fixes #955.

Note: This is an alternative approach to #1010. Both PRs are compatible and can be merged together.

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1182,30 +1182,27 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
                         }
                     }
 
-                    // restore gain (if new name is different from the current one)
-                    if ( vecpChanFader[i]->GetReceivedName().compare ( vecChanInfo[j].strName ) )
-                    {
-                        // the text has actually changed, search in the list of
-                        // stored settings if we have a matching entry
-                        int  iStoredFaderLevel;
-                        int  iStoredPanValue;
-                        bool bStoredFaderIsSolo;
-                        bool bStoredFaderIsMute;
-                        int  iGroupID;
+                    // restore gain:
+                    // the text has actually changed, search in the list of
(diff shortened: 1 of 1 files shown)


src/audiomixerboard.cpp
```diff
@@ -1174,30 +1174,27 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
                         }
                     }
 
-                    // restore gain (if new name is different from the current one)
-                    if ( vecpChanFader[i]->GetReceivedName().compare ( vecChanInfo[j].strName ) )
```
hoffie, line 1178: The diff looks larger than it is. Essentially, this `if` is replaced by an `if ( true ) `. `git show -w bd081647b2152f9a430c4c59bb0b98a035d2cade` will be more readable than Github. :)

pljones: Why doesn't Github have an "ignore whitespace".....?!

hoffie: Just learned that Github actually supports this! :)

https://github.blog/2018-05-01-ignore-white-space-in-code-review/

https://github.com/jamulussoftware/jamulus/pull/1009/files?w=1

softins (approved):

pljones (approved):

PR #1017, opened 2021-02-18
IPv6 support

jardous:
This PS add support for IPv6. Note that it is backwards compatible with IPv4.
Motivation is quite simple. I am running a server on AWS and  IPv4 ping takes >40ms, but IPv6 ping takes just 15ms! And usability of Jamulus all depends on the ping time.

The difference between IPv4 and IPv6 is so big, that it should be recommended to all users to switch/upgrade to IPv6.

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -25,6 +25,13 @@
 #include "socket.h"
 #include "server.h"
 
+#ifdef _WIN32
+#include <winsock2.h>
+#include <ws2tcpip.h>
+#else
+#include <arpa/inet.h>
+#endif
+
 
 /* Implementation *************************************************************/
 void CSocket::Init ( const quint16 iPortNumber )
@@ -39,15 +46,19 @@ void CSocket::Init ( const quint16 iPortNumber )
 #endif
 
     // create the UDP socket
-    UdpSocket = socket ( AF_INET, SOCK_DGRAM, 0 );
+    UdpSocket = socket ( AF_INET6, SOCK_DGRAM, 0 );
+    // The IPV6_V6ONLY socket option must be false in order for the socket to listen on both protocols.
+    // On Linux it's false by default on most (all?) distros, but on Windows it is true by default
+    const uint8_t no = 0;
+    setsockopt( UdpSocket, IPPROTO_IPV6, IPV6_V6ONLY, (const char *)&no, sizeof( no ) );
 
     // allocate memory for network receive and send buffer in samples
     vecbyRecBuf.Init ( MAX_SIZE_BYTES_NETW_BUF );
 
     // preinitialize socket in address (only the port number is missing)
-    sockaddr_in UdpSocketInAddr;
-    UdpSocketInAddr.sin_family      = AF_INET;
-    UdpSocketInAddr.sin_addr.s_addr = INADDR_ANY;
+    struct sockaddr_in6 UdpSocketInAddr;
+    UdpSocketInAddr.sin6_family = AF_INET6;
+    UdpSocketInAddr.sin6_addr   = in6addr_any;
 
     // initialize the listening socket
     bool bSuccess;
@@ -57,11 +68,11 @@ void CSocket::Init ( const quint16 iPortNumber )
         if ( iPortNumber == 0 )
         {
             // if port number is 0, bind the client to a random available port
-            UdpSocketInAddr.sin_port = htons ( 0 );
+            UdpSocketInAddr.sin6_port = htons ( 0 );
 
             bSuccess = ( ::bind ( UdpSocket ,
                                 (sockaddr*) &UdpSocketInAddr,
-                                sizeof ( sockaddr_in ) ) == 0 );
+                                sizeof ( struct sockaddr_in6 ) ) == 0 );
         }
         else
         {
@@ -77,11 +88,11 @@ void CSocket::Init ( const quint16 iPortNumber )
 
             while ( !bSuccess && ( iClientPortIncrement <= NUM_SOCKET_PORTS_TO_TRY ) )
             {
-                UdpSocketInAddr.sin_port = htons ( startingPortNumber + iClientPortIncrement );
+                UdpSocketInAddr.sin6_port = htons ( startingPortNumber + iClientPortIncrement );
 
                 bSuccess = ( ::bind ( UdpSocket ,
                                       (sockaddr*) &UdpSocketInAddr,
-                                      sizeof ( sockaddr_in ) ) == 0 );
+                                      sizeof ( struct sockaddr_in6 ) ) == 0 );
 
                 iClientPortIncrement++;
             }
@@ -92,11 +103,11 @@ void CSocket::Init ( const quint16 iPortNumber )
         // for the server, only try the given port number and do not try out
         // other port numbers to bind since it is important that the server
         // gets the desired port number
-        UdpSocketInAddr.sin_port = htons ( iPortNumber );
+        UdpSocketInAddr.sin6_port = htons ( iPortNumber );
 
         bSuccess = ( ::bind ( UdpSocket ,
                               (sockaddr*) &UdpSocketInAddr,
-                              sizeof ( sockaddr_in ) ) == 0 );
+                              sizeof ( struct sockaddr_in6 ) ) == 0 );
     }
 
     if ( !bSuccess )
@@ -181,19 +192,36 @@ void CSocket::SendPacket ( const CVector<uint8_t>& vecbySendBuf,
         // send packet through network (we have to convert the constant unsigned
         // char vector in "const char*", for this we first convert the const
         // uint8_t vector in a read/write uint8_t vector and then do the cast to
-        // const char*)
-        sockaddr_in UdpSocketOutAddr;
+        // const char *)
+        if (HostAddr.InetAddr.protocol() == QAbstractSocket::IPv4Protocol) {
 
-        UdpSocketOutAddr.sin_family      = AF_INET;
-        UdpSocketOutAddr.sin_port        = htons ( HostAddr.iPort );
-        UdpSocketOutAddr.sin_addr.s_addr = htonl ( HostAddr.InetAddr.toIPv4Address() );
+            sockaddr_in UdpSocketOutAddr;
+            UdpSocketOutAddr.sin_family      = AF_INET;
+            UdpSocketOutAddr.sin_port        = htons ( HostAddr.iPort );
+            UdpSocketOutAddr.sin_addr.s_addr = htonl ( HostAddr.InetAddr.toIPv4Address() );
 
-        sendto ( UdpSocket,
-                 (const char*) &( (CVector<uint8_t>) vecbySendBuf )[0],
-                 iVecSizeOut,
-                 0,
-                 (sockaddr*) &UdpSocketOutAddr,
-                 sizeof ( sockaddr_in ) );
+            sendto ( UdpSocket,
+                     (const char*) &( (CVector<uint8_t>) vecbySendBuf )[0],
+                     iVecSizeOut,
+                     0,
+                     (sockaddr*) &UdpSocketOutAddr,
+                     sizeof ( UdpSocketOutAddr ) );
+
+        } else {
+
+            sockaddr_in6 UdpSocketOutAddr;
+
+            UdpSocketOutAddr.sin6_family = AF_INET6;
+            UdpSocketOutAddr.sin6_port   = htons ( HostAddr.iPort );
+            inet_pton(AF_INET6, HostAddr.InetAddr.toString().toLocal8Bit().constData(), &UdpSocketOutAddr.sin6_addr);
+
+            sendto ( UdpSocket,
+                     (const char*) &( (CVector<uint8_t>) vecbySendBuf )[0],
+                     iVecSizeOut,
+                     0,
+                     (sockaddr*) &UdpSocketOutAddr,
+                     sizeof ( UdpSocketOutAddr ) );
+        }
     }
 }
 
@@ -223,18 +251,19 @@ void CSocket::OnDataReceived()
 */
 
     // read block from network interface and query address of sender
-    sockaddr_in SenderAddr;
+    struct sockaddr_storage addrstorage;
+    struct sockaddr_in6 *in6;
 #ifdef _WIN32
-    int SenderAddrSize = sizeof ( sockaddr_in );
+    int SenderAddrSize = sizeof ( struct sockaddr_storage );
 #else
-    socklen_t SenderAddrSize = sizeof ( sockaddr_in );
+    socklen_t SenderAddrSize = sizeof ( struct sockaddr_storage );
 #endif
 
     const long iNumBytesRead = recvfrom ( UdpSocket,
                                           (char*) &vecbyRecBuf[0],
                                           MAX_SIZE_BYTES_NETW_BUF,
                                           0,
-                                          (sockaddr*) &SenderAddr,
+                                          (sockaddr*) &addrstorage,
                                           &SenderAddrSize );
 
     // check if an error occurred or no data could be read
@@ -243,10 +272,14 @@ void CSocket::OnDataReceived()
         return;
     }
 
-    // convert address of client
-    RecHostAddr.InetAddr.setAddress ( ntohl ( SenderAddr.sin_addr.s_addr ) );
-    RecHostAddr.iPort = ntohs ( SenderAddr.sin_port );
-
+    in6 = reinterpret_cast<struct sockaddr_in6 *>(&addrstorage);
+    if (IN6_IS_ADDR_V4MAPPED(&(in6->sin6_addr))) {
+        RecHostAddr.InetAddr.setAddress ( ntohl (*((uint32_t*)in6->sin6_addr.s6_addr + 3) ) );
+        RecHostAddr.iPort = ntohs ( in6->sin6_port );
+    } else {
+        RecHostAddr.InetAddr.setAddress( in6->sin6_addr.s6_addr );
+        RecHostAddr.iPort = ntohs ( in6->sin6_port );
+    }
 
     // check if this is a protocol message
     int              iRecCounter;


ann0see:
Hi @jardous! Thanks for opening your first PR! I strongly support ipv6 support! IMHO, every modern software should support it ;-).

It seems as if this code fails on Windows. Could you please have a look at it. There's also #722 which you might need to take care of.

jardous:
thanks @ann0see. Unfortunately I am running Linux computers only so I'll just rely on the Windows's CodeQL/Analyze.

ann0see:
Ah. Ok. I assume you saw this error:

```
socket.cpp
src\socket.cpp(27): fatal error C1083: Cannot open include file: 'arpa/inet.h': No such file or directory
```

Probably arpa/inet.h doesn't exist on Windows.

ann0see:
@hoffie assigned you here. I hope that's ok for you. Although I know some networking stuff, I assume you're better here.

src/socket.cpp
```diff
@@ -25,6 +25,13 @@
 #include "socket.h"
 #include "server.h"
 
+#ifdef _WIN32
```
pljones, line 28: None of this should be in the `.cpp` file.  There's a `socket.h` file for socket-related includes.

We should only use Qt network support, where possible, also.

pljones:
What testing has been done with this?

IPv4 client <-> IPv4 server <-> IPv4 client
IPv6 client <-> IPv4 server <-> IPv4 client
IPv4 client <-> IPv6 server <-> IPv4 client
IPv6 client <-> IPv6 server <-> IPv4 client

jardous:
Thanks @pljones for list of test cases.

I will need help from the community to test all of those scenarios as I only have limited setup possibilities.
It is fresh - I have started writing the PS just few hours ago. I have just done some local testing.

BTW the Windows building issue is fixed.

hoffie:
Glad to see this being worked on! Sadly, I currently do not run IPv6 anywhere and will probably not be able to test with v6 right now. Nevertheless, the main point should probably that we don't break anything for existing users. So, I fully agree with what @pljones says regarding Qt and testing.

I want to add that I think there are places in the protocol where IP addresses are shared, e.g. https://github.com/jamulussoftware/jamulus/blob/master/src/protocol.cpp#L327

Some parts are likely related to central server handling, but there may be others. I'm not saying that all of this should be addressed within this first step towards full IPv6 support, but we should probably get an overview and see how things might fail. Such known shortcomings could then be documented somehow.

gene96817:
I just reviewed my router configuration and has both IPv4 and IPv6 running with my ISP. I guess I could help with some IPv6 testing but I have never tried it on my private network. Finally, IPv6 is coming to my home. I wonder how many other developers have IPv6 already enable in their routers and waiting to be used.

pljones:
Good point on the directory servers - I'd forgotten those... so we'd need four times the test cases again, then thin out the duplicates (i.e. all of the above with unregistered server, all with old code directory server, all with new code IPv4, all with new code IPv6).

DavidSavinkoff:
Maybe there should be a build option for ipv6... And a build option for ipv4 so that there are choices for all. The amount of code changed looks to be small enough to use #ifdef. This would probably be the fastest way forward. Note that I do not have ipv6, and will probably not have ipv6 untill the USA goes metric.

jardous:
@DavidSavinkoff I would prefer to change it so it supports both. I'll take look - give me some time.

atsampson:
Another case that will need dealing with is when you're trying to connect to a server that provides both IPv4 and IPv6 (presumably the directory servers would do this in an IPv6 future), but the local network provides only IPv4 - or only IPv6 - or both. It's entirely possible (and quite common on current networks) to have a local IPv6 address but not have connectivity to public IPv6.

There's a standard way of dealing with this: the splendidly named [Happy Eyeballs](https://tools.ietf.org/html/rfc8305). But it's not trivial to implement, and ideally we'd make use of existing facilities to do the job if possible. I'd agree with @pljones that using Qt's socket classes is likely to be a better bet than trying to extend the custom ones Jamulus currently uses.

And then there's all the fun with IPv6 and DNS - e.g. you can have users behind a DNS64 setup, where A records (containing IPv4 addresses) are automatically translated into AAAA records pointing at a NAT64 server. So you'd have local Jamulus thinking it's connected by IPv6, but actually talking through a proxy to an IPv4 server...

gene96817:
> And then there's all the fun with IPv6 and DNS - e.g. you can have users behind a DNS64 setup, where A records (containing IPv4 addresses) are automatically translated into AAAA records pointing at a NAT64 server. So you'd have local Jamulus thinking it's connected by IPv6, but actually talking through a proxy to an IPv4 server...

There are plenty of mistakes being made with IPv6. Eventually, implementations will automatically work properly without users needing to know about IPv4 or IPv6. We will be making mistakes, too, until it just works.

jardous:
server list fetch fixed

bjacke:
> And then there's all the fun with IPv6 and DNS - e.g. you can have users behind a DNS64 setup, where A records (containing IPv4 addresses) are automatically translated into AAAA records pointing at a NAT64 server. So you'd have local Jamulus thinking it's connected by IPv6, but actually talking through a proxy to an IPv4 server...

but whatever IPv6 connectivity is being used there, this should be transparent for userspace applications. I think checking the Jamulus protocol ping time both via v4 and v6 should be checked and the better latency connection being used then.

For those of you who don't have IPv6 at home yet, at https://www.tunnelbroker.net/ I think it's still possible to get ipv6 tunnels, I used them for a long time, very much recommended and helpful for testing things here.

jarmar:
I looked into adding IPv6 support last year, but didn't finish it when I realized how much work it would be, and that I could work around it on my end instead.

At that point, I investigated what protocol messages include IPv4 addresses. I found:

* `PROTMESSID_CLM_REGISTER_SERVER`/`_EX`, `PROTMESSID_CLM_SERVER_LIST`, `PROTMESSID_CLM_RED_SERVER_LIST`, `PROTMESSID_CLM_SEND_EMPTY_MESSAGE`: all related to the server list.
* `PROTMESSID_CONN_CLIENTS_LIST`, `PROTMESSID_CLM_CONN_CLIENTS_LIST`: these messages have fields for an address, _but_ since commit e7621af7, they are always dummy addresses.

So my impression was that making IPv6 work everywhere _except_ in the server list should be "simple", in the sense that it should be possible to do transparently and without modifying the protocol. But making the server list IPv6-aware would require a lot more up-front design.

src/socket.cpp
```diff
@@ -39,15 +46,15 @@ void CSocket::Init ( const quint16 iPortNumber )
 #endif
 
     // create the UDP socket
-    UdpSocket = socket ( AF_INET, SOCK_DGRAM, 0 );
+    UdpSocket = socket ( AF_INET6, SOCK_DGRAM, 0 );
```
jarmar, line 49: The `IPV6_V6ONLY` socket option must be false in order for the socket to listen on both protocols. On Linux it's false by default on most (all?) distros, but on Windows it is true by default, see https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options.

src/socket.cpp
```diff
@@ -182,18 +189,35 @@ void CSocket::SendPacket ( const CVector<uint8_t>& vecbySendBuf,
...
-        UdpSocketOutAddr.sin_port        = htons ( HostAddr.iPort );
-        UdpSocketOutAddr.sin_addr.s_addr = htonl ( HostAddr.InetAddr.toIPv4Address() );
-
-        sendto ( UdpSocket,
-                 (const char*) &( (CVector<uint8_t>) vecbySendBuf )[0],
-                 iVecSizeOut,
-                 0,
-                 (sockaddr*) &UdpSocketOutAddr,
-                 sizeof ( sockaddr_in ) );
+        if (HostAddr.InetAddr.protocol() == QAbstractSocket::IPv4Protocol) {
```
jarmar, line 192: Is this necessary? `toIPv6Address()` will convert IPv4 addresses to IPv6 mapped, so I believe it should be enough to only handle the IPv6 case.

jardous:
Hello guys,

I have just tested the changes as follows.

My network setup is:
PC (Linux, IPv4 + NAT, IPv6) -> Router (port forwarding to the PC's IPv6 address) -> IPv6 network (UnityMedia) -> AWS Debian (Frankfurt, static IPv4, IPv6)

Following scenarios:
1. server (old, IPv4) + client (new, connect to IPv4 address) --> PASS (ping >60ms, overall delay >80ms)
2. server (new, IPv4+IPv6) + client(new, connect to IPv6 address) --> PASS (ping 15ms, overall delay 35ms)
3. server (new, IPv4+IPv6) + client(new, connect to IPv4 address) --> PASS (ping 60ms, overall delay 80ms)
4. local server (new, IPv4+IPv6) + client(new, connect to ::1) --> PASS (ping 0ms, overall delay 13ms)

pljones:
> server (old, IPv4) + client (new, connect to IPv4 address) --> PASS (ping >60ms, overall delay >80ms)

Can you try
> server (old, IPv4) + client (old, connect to IPv4 address) --> ??

jardous:
> > server (old, IPv4) + client (new, connect to IPv4 address) --> PASS (ping >60ms, overall delay >80ms)
> 
> Can you try
> 
> > server (old, IPv4) + client (old, connect to IPv4 address) --> ??

yes - of course. That is how I did run it till now. Same results (ping >60ms, overall delay >80ms)

pljones:
OK - thanks for confirming.  Sounds like maybe your ISP is tunnelling IPv4 over IPv6 already, then, so native IPv6 gives you a great saving.

jardous:
> OK - thanks for confirming. Sounds like maybe your ISP is tunnelling IPv4 over IPv6 already, then, so native IPv6 gives you a great saving.
> 
> Looks like it's the opposite in the UK -- if you can find an ISP offering IPv6, it's tunnelled over IPv4...

yes. I live in Germany and have quite common cable connection from Unity Media (now Vodafone). The cable router has no IPv4 WAN address.

gilgongo:
BTW I assume if the client is on IPv6 it will try that before IPv4, is that right? So no need for any additions to the user docs on this, since it'll just happen automatically?

jardous:
> BTW I assume if the client is on IPv6 it will try that before IPv4, is that right? So no need for any additions to the user docs on this, since it'll just happen automatically?

It should be all transparent and it all depends on what format of the server address you specify

gilgongo:
> depends on what format of the server address you specify

So to be clear, it's not automatic if you explicitly enter an IPv4 address to connect to the server, but will be transparent if you enter anything else? I ask because I see we're not adding some option like `-4` and `-6` to Jamulus as part of this (AFAIKT)

jardous:
> g else? I ask because I see we're not adding some option like `-4` and `-6` to Jamulus as part of this (AFAIKT). Oh, and I assume if central servers are only on IPv4 then clients on IPv6 can still connect over IPv6 to compatible servers? Again, just making sure we don't need to hedge the definition of "Supporting IPv6" in the docs :-)

Yes, with the new version I am able to connect to any IPv4 server. If I use my own new server with IPv6 address, it will use IPv6.

gilgongo:
@jardous Sorry, I meant the _directory server_ (also called "central_ server") for public servers. So does IPv6 connectivity for public servers depend on the directory server **also** being on an IPv6 connection? In other words, if the directory is only on IPv4, but the server listed on it is on IPv6, will my client on IPv6 be able to connect to the server over IPv6? I'm thinking possibly not, since the directory server would only ping on IPv4...

bjacke:
I did some tests with the latest patch. The server site was working fine right from the beginning for me, my comments here are all related to the client part:

1) connecting to a DNS name that has v4 and v6 addresses, the client only connects to the v4 address, even if I block the v4 traffic on the server, the client still uses the non-working v4 address instread of the v6 address

2) a -4/-6 option would actually be nice and/or an option in the GUI whether IPv4 or IPv6 should be prefered

3) a user should be able to see which IP address she's connected to. I would suggest to display the connected IP address in the main window where currently is written "Personal mix at: <servername>" and just add (ipaddress) there.

Other things work well for me now, including the notation "[f00:faa::1]:22124" to use a IPv6 address with an alternative port number. Thank you!

pljones:
OK, here's a table:
```
| ClientA | ClientA ISP | Directory Server ISP | Directory Server | ServerA ISP | ServerA | ClientB ISP | ClientB | Status 3.6.2 | Status #1017 |
+---------+-------------+----------------------+------------------+-------------+---------+-------------+---------+--------------+--------------+
| IPv4    | native IPv4 | native IPv4          | IPv4             | native IPv4 | IPv4    | native IPv4 | IPv4    | "as is"      |              |
| IPv4    | native IPv4 | native IPv4          | IPv4             | native IPv4 | IPv4    | tunnel IPv4 | IPv4    |              |              |
| IPv4    | native IPv4 | native IPv4          | IPv4             | native IPv4 | IPv4    | tunnel IPv6 | IPv6    |              |              |
| IPv4    | native IPv4 | native IPv4          | IPv4             | native IPv4 | IPv4    | native IPv6 | IPv6    |              |              |
+---------+-------------+----------------------+------------------+-------------+---------+-------------+---------+--------------+--------------+
| IPv4    | native IPv4 | native IPv4          | IPv4             | tunnel IPv4 | IPv4    | tunnel IPv4 | IPv4    |              |              |
| IPv4    | native IPv4 | native IPv4          | IPv4             | tunnel IPv6 | IPv6    | tunnel IPv6 | IPv6    |              |              |
| IPv4    | native IPv4 | native IPv4          | IPv4             | native IPv6 | IPv6    | native IPv6 | IPv6    |              |              |
+---------+-------------+----------------------+------------------+-------------+---------+-------------+---------+--------------+--------------+
| IPv4    | native IPv4 | tunnel IPv4          | IPv4             | tunnel IPv4 | IPv4    | tunnel IPv4 | IPv4    |              |              |
| IPv4    | native IPv4 | tunnel IPv6          | IPv6             | tunnel IPv6 | IPv6    | tunnel IPv6 | IPv6    |              |              |
| IPv4    | native IPv4 | native IPv6          | IPv6             | native IPv6 | IPv6    | native IPv6 | IPv6    |              |              |
+---------+-------------+----------------------+------------------+-------------+---------+-------------+---------+--------------+--------------+
| IPv4    | tunnel IPv4 | tunnel IPv4          | IPv4             | tunnel IPv4 | IPv4    | tunnel IPv4 | IPv4    |              |              |
| IPv6    | tunnel IPv6 | tunnel IPv6          | IPv6             | tunnel IPv6 | IPv6    | tunnel IPv6 | IPv6    |              |              |
| IPv6    | native IPv6 | native IPv6          | IPv6             | native IPv6 | IPv6    | native IPv6 | IPv6    |              |              |
+---------+-------------+----------------------+------------------+-------------+---------+-------------+---------+--------------+--------------+
```
Yes... I'd *like* it all tested, that's just how I am...  Given we're unlikely to be *able* to test it all, I think we're going to have to take a best efforts approach _***BUT***_ in that case we *must* allow the the user to select the network addressing mode to use, in case the "default" turns out to be the tunnelled mode and they are able to use the native alternative (with a huge saving in latency).

jardous:
@pljones Thanks. That is an impressive table. I think we should narrow it down to check if we do not break compatibility. At least for the first iteration so we have the feature in.
For those who know, and can run an IPv6 server, would that be a nice "extra" and they can help with testing.

gene96817:
I am very happy to see all this work.
When this code is released, with the v4/v6 operating transparently according to the better ping time, it will be very interesting to watch which servers will be receiving client connections in v6. We could be surprised.

pljones:
> with the v4/v6 operating transparently according to the **better ping time**

This depends on which way the tunnelling is happening.  If the ISP is offering IPv4 and IPv6 addresses but tunnels over IPv4, then you could see degradation switching from IPv4 (i.e. native) to IPv6 (i.e. tunnelled).

We could, indeed, be surprised - but not necessarily pleasantly.

bjacke:
at https://www.google.com/intl/en/ipv6/statistics.html#tab=per-country-ipv6-adoption you can compare the avarage county v4/v6 latenty difference. IPv6 us almost always faster. I've never heard of a provider who offers IPv6 through tunneling.  Teredo tunnels of Windows machines might pop up often as (slow) tunneled connections but a latency check of Jamulus should be able sort those slow v6 tunnels out. Almost every Windows machine without native v6 connection has such a Teredo tunnel, maybe someone of you will be able to test that.
Talking about tunnels, I still don't really understand the point about the many different native vs. tunnel cases in the table. The connection technology is transparente to user space applications. I would rather make sure that the v4/v6 latency check works well DNS resolution works well and that connection also works from clients to servers that have both v4 and v6 addresses but one of them is not reachable via UDP on the Jamulus port.

pljones:
There are four components - Client A, Server, Directory Server and Client B.  All four need to support both IPv6 to either IPv4 or IPv6 on each of the other three components - with the exceptions that Client A and Client B don't connect directly and the Client B connections are the same as Client A - it's only there to ensure it's been considered as part of the overall signal path.

The tunnelling components _should_ be transparent.  They're the ones that are most likely to vary by ISP rather than anything in the user control.  Testing is more so there's _some_ knowledge about what to expect in some cases - it's here that I'm least worried about exhaustive coverage.

gilgongo:
@pljones 

> All four need to support both IPv6 to either IPv4 or IPv6 on each of the other three components

So if Server and Client are on IPv6 but Directory is not, then Server and Client will connect to each other using IPv4, is that right?

pljones:
> So if Server and Client are on IPv6 but Directory is not, then Server and Client will connect to each other using IPv4, is that right?

I hope not.

gene96817:
Where does the client get the Server address from? If it is from the directory server, then the server needs to have both IPv4 and IPv6 addresses from the Servers that register. If explorer.jamulus.io represents what is in the Directory Server, then I only see IPv4 addresses, (suggesting that servers will only get IPv 4 addresses from the directory). Maybe explorer.jamulus.io doesn't display the IPv6 but the addresses are there?

pljones:
IPv6 support doesn't exist yet, so nothing will use it.  This is a pull request with a new feature in that has yet to be accepted.

gilgongo:
> Where does the client get the Server address from? If it is from the directory server, then the server needs to have both IPv4 and IPv6 addresses 

This is why I asked :-) It would be a bit of a downer if we had to run two sets of directory servers to enable IPv6 connections on public servers.

pljones:
Part of the reason the PR hasn't been accepted is that it's not been shown to work transparently with the directory servers.  Let's leave it until the code exists to worry about whether it works.

bjacke:
you might either add IPv6 support to the directory server now, your provider is Linode and they support IPv6 on all hosts.

bjacke:
it would also be nice if the WELL_KNOWN_HOST 1.1.1.1 check would be removed from the code so that you have no IPv4 connectivity requirement there.

pljones:
> 
> 
> it would also be nice if the WELL_KNOWN_HOST 1.1.1.1 check would be removed from the code so that you have no IPv4 connectivity requirement there.

We have the same requirement for the public interface address, regardless of whether it's IPv4 or IPv6 - or doesn't IPv6 addressing provide a separate address per endpoint?

bjacke:
> We have the same requirement for the public interface address, regardless of whether it's IPv4 or IPv6 - or doesn't IPv6 addressing provide a separate address per endpoint?

ah, sorry, I didn't realize that this is the magic for finding out the outgoing IP(v4) address.

hoffie:
> @pljones Thanks. That is an impressive table. I think we should narrow it down to check if we do not break compatibility. At least for the first iteration so we have the feature in.
> For those who know, and can run an IPv6 server, would that be a nice "extra" and they can help with testing.

I also think it would be a valid approach to add basic, optional IPv6 support in the first iteration (i.e. no IPv6 by default, no support for central servers with IPv6). This would help people who do have better IPv6 connectivity than IPv4, it would help to get an idea what works/breaks and it would help getting the feature in while still ensuring that we don't break anything.

For full support I agree with @pljones that lots of things should be validated.

Hk1020:
A test I just did but with #1455, which ported to recent master on a private fork.

I have a server running with both ipv4 and ipv6 and a A+AAAA dns record. Connecting using the dns name establishes a ipv6 connection but connecting via the central server selects ipv4. So it looks like the client gets the IP address from the central server.

pljones:
The purpose of the Jamulus directory servers is to provide directory entry to IP/port mappings, keeping those details refreshed and keeping firewalls open by sending ping messages.  As nothing has been done to the directory servers to add IPv6 support, it's no surprise that using them gets an IPv4 address.

Hk1020:
So the client does not do a dns lookup of the server it wants to connect to. This is unlike to what happens with entering the server name in the little connect box. From there connection to ipV6 servers seems to work fine with #1455. I could not get anything working with this PR. And with server I mean the actual Jamulus server, not the directory server.

pljones:
The client will do a DNS lookup if one is needed (because it passes the values to Qt to deal with -- a dotted quad "resolves" without any look up).

 #1455 provides fallback in the case where IPv6 is not available - unless you tell it explicitly not to (with `--ip6only`).

This change currently _removes_ IPv4 support.  So unless you're connecting a client and server built from this branch, you cannot expect it to work.

pljones:
Would we be better closing this and concentrating on #1455?  That also needs further work, of course, but it makes no sense to have two open PRs for IPv6.

gene96817:
I don't understand why [https://github.com/jamulussoftware/jamulus/pull/1455](url) is so dependent on command flags. It should be straightforward to automate the dynamic selection of with IPv6 or IPv4.

pljones:
> I don't understand why [https://github.com/jamulussoftware/jamulus/pull/1455](url) is so dependent on command flags. It should be straightforward to automate the dynamic selection of with IPv6 or IPv4.

Probably better raised there.

DavidSavinkoff:
> I don't understand why [https://github.com/jamulussoftware/jamulus/pull/1455](url) is so dependent on command flags. It should be straightforward to automate the dynamic selection of with IPv6 or IPv4.

Some flags will be needed because 'automatic' doesn't work as it should.

IPv6 has the same problems as IPv4 by not acknowleging that the other version exists.
This theoretically should only be a IPv4 problem because IPv6 has a chance to correct the problem. IPv6 is so stupid that when IPvX happens, IPv6 will be in the same shoes as IPv4 was. This problem is in operating systems too. This is why gai.conf has to be manually tweeked.

gene96817:
> IPv6 has the same problems as IPv4 by not acknowleging that the other version exists.

I apologize for asking a very simple question. Are you saying that the following startup procedure will not work?

Initialize an IPv6 connection with the server...
     on IPv6 timeout, abandon IPv6 and initialize an IPv4 connection
           on IPv4 timeout, <declare the server down or start error recovery>

Seems to me, all servers support IPv4s, and ervers with the new code will support IPv6 but they may not have an IPv6 connection. If there is no response to an IPv6 initiation, we fallback to the (universally available) IPv4.  If we can make this process work then the Jamulus world will automatically migrate to IPv6 without additional intervention.



>

DavidSavinkoff:
> > IPv6 has the same problems as IPv4 by not acknowleging that the other version exists.
> 
> I apologize for asking a very simple question. Are you saying that the following startup procedure will not work?
> 
> Initialize an IPv6 connection with the server...
> on IPv6 timeout, abandon IPv6 and initialize an IPv4 connection
> on IPv4 timeout,
> 
> Seems to me, all servers support IPv4s, and ervers with the new code will support IPv6 but they may not have an IPv6 connection. If there is no response to an IPv6 initiation, we fallback to the (universally available) IPv4. If we can make this process work then the Jamulus world will automatically migrate to IPv6 without additional intervention.
> 
> >

This is exactly the problem. There shouldn't be timeouts for either IPv4 or IPv6. That is why you need command line options, and gai.conf modifications.
The time outs and fallbacks are poor design baked into IPv6 deployment. One would think that IPv6 would have solved this IPv4 problem. This is probably why IPv6 is so slow to be accepted.

gene96817:
There are lots of reasons that make IPv6 slow to adopt. My first IPv6 product rollout was in the 1990s... go figure.
Timeouts and fallbacks are hacks. Still, it is a pragmatic way to deal with the lack of version negotiations in IPv4 and IPv6 start-up procedures. Good implementation is automatic and permissive. Why inflict the deficiencies of IPv4/IPv6 on every Jamulus installation? Automagical functions are wonderful (as long as they are not buggy).

ann0see:
@all here. Please have a look at #1938 by @softins for review.

softins:
Closing this as it has been superseded by #1938

PR #1018, opened 2021-02-18
Use "CONFIG+=noupcasename" to fix desktop file

ann0see:
It seems as if the created desktop file didn't use the correct target and therefore wrong binary name. See: https://github.com/nefarius2001/jamulus/pull/6/commits/00a917fb5954ec5bb6814fe2a1941d288d3f3d67


gilgongo (approved):

hoffie (approved):
LGTM.

Issue #1016, opened 2021-02-18
set TOS/DSCP IP header bits for routers to prioritize packages


bjacke:
To signalize edge routers to prioritize Jamulus UDP traffic, it would be good if Jamulus would set the TOS header bits accordingly. https://doc.qt.io/qt-5/qabstractsocket.html describes TypeOfServiceOption which could probably used for that. A values of 0x10 (IPTOS_LOWDELAY) is an often used value, which is known to be very compatible with different kinds of routers without causing problems.
That would help many people get better latency even if they are not able to configure QoS manually on their routers.

gene96817:
A very good idea. However many (most?) ISPs (in the USA) ignore these bits.

bjacke:
It's imporant for routers before rhe packages arrive the internet. Many home router homor TOF bits and send the packages out earlier, even if othe bulk data is in the send queue

18.02.2021 17:53:14 gene96817 <notifications@github.com>:

gene96817:
Yes, individuals can configure their personal network to give Jamulus packets priority over competing traffic from other members of their household.  I would not give this idea to anyone that needed assistance installing Jamulus.

DavidSavinkoff:
This functionality should be added to the Linux version of Jamulus... and Apple (BSD ) if possible.
IPTOS in Linux sockets programming is obviously Not obsolete.
Otherwise, one will have to configure their router or firewall to set these bits in IP packets.
Maybe someone (MSFT ?) could write a Windows(tm) driver to add IPTOS funtionality

hoffie:
I like this idea. I've been running my Jamulus servers (and even clients) while adding TOS flags using iptables. Although I haven't done any objective tests, it seems that ping times are reduced.

Having this directly in Jamulus would simplify setups and make this accessible to more people and platforms.

@bjacke Do you know about possible negative side effects, such as routers blocking such packets entirely?
> which is known to be very compatible with different kinds of routers without causing problems

Any pointers regarding that maybe?

bjacke:
IPTOS_LOWDELAY (0x10) is working very well and I'm not aware of broken routers that block those packets openssh has been using IPTOS_LOWDELAY for a long time, then they changed that to af21 because they liked that more for some reason. That change however resulted in a number of boxes to drop those packages then. This is why Debian for example patched their ssh client back to use IPTOS_LOWDELAY insted of af21. See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=923880
[...]

PR #1066, opened 2021-02-19
Update feature_request.md

gilgongo:
I suggest we need to make sure the Issues list contains things that have been reasonably well agreed and specified in discussions first, otherwise Issues will just become discussions. And we can't project manage them into projects/milestones so will just spin our wheels on them.

BTW I'm not familiar with these templates not sure if I'm doing it right...


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -1,21 +1,17 @@
 ---
-name: Feature request
-about: Suggest a new idea for Jamulus (not an improvement of an existing function)
+name: Feature
+about: Agreed specificaiton for a feature request or enhancement
 title: ''
-labels: 'feature request'
+labels: 'feature'
```
ann0see, line 5: Are you sure? Is this the new label for feature requests?

pljones: This is what's been being discussed.  Feature requests, i.e. anything that isn't defined and accepted as a new feature that's going to be delivered, get talked about in Discussions.  Then a formal feature issue gets raised with the agreed spec.

gilgongo: Yes, looking at the Issues yesterday, I'd say maybe 80% of them were "Hey! Lets do X!" followed by people saying, "No! Why? What do you mean?" then after about 20-30 messages it would be "OK, so about we do Y then?". But meanwhile, the original "Let's do X!" was intact so it was impossible to really know what the "Issue" was, let alone how to label it for a release, tie it to a PR or whatever.

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -1,21 +1,17 @@
 ---
-name: Feature request
-about: Suggest a new idea for Jamulus (not an improvement of an existing function)
+name: Feature
+about: Agreed specificaiton for a feature request or enhancement
```
hoffie, line 3:
```suggestion
about: Agreed specification for a feature request or enhancement
```

.github/ISSUE_TEMPLATE/feature_request.md
hoffie: I fully agree with this direction, however I'm a bit unsure where to draw the line. If Github issues should be strictly limited to perfectly specified new features, then I guess this would already involve half of the ideas about the specific implementation. This would also mean that issues would practically be more or less limited to people who intend to write the actual code.

This has certainly benefits, but I wonder whether that would be a bit too strict.

There are issues where the intention is absolutely clear, but there is still some discussion/opinions regarding risk of implementation or specific details. A good example is https://github.com/jamulussoftware/jamulus/issues/1016. In that case I think I would have preferred for it to stay an issue as the "discussion" around it is valuable in the context of the implementation?

On the other hand, I fully agree that issues like the [big metronome one](https://github.com/jamulussoftware/jamulus/discussions/1064) should really be discussions until everything is sorted.

I guess the problem is that it will be hard to predict when issues go "out of control". I fear that we might direct too many to discussions for such requests. This would keep the issue list clean, but would make it harder to manage all the open requests which would then be in the discussions area?

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -1,21 +1,17 @@
...
-**Is your feature request related to a problem? Please describe.**
-<!--A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] -->
+**Has this feature been discussed and generally agreed?**
+<!-- Please ensure that your feature is a clearly specified piece of work that can be reasonably actioned by someone without too much help. If not, please post it in Discussions first so that it can become so. -->
 
 **Describe the solution you'd like**
-<! --A clear and concise description of what you want to happen. --> 
+<!-- A clear and concise description of what you want to happen. --> 
 
-**Describe alternatives you've considered**
```
hoffie, line 17: I wonder whether the alternatives part should be kept. When making technical decisions, it's useful to document why alternative solutions have not been implemented. Maybe it was intended to still have it but keep that part in the related discussion? Might be hard to "enforce" some guidance/structure there though.

.github/ISSUE_TEMPLATE/feature_request.md
gilgongo: > If Github issues should be strictly limited to perfectly specified new features,

Sorry, what I meant was that they should be at least reasonably stable. Not literally morphing into different things over time. 

> This would also mean that issues would practically be more or less limited to people who intend to write the actual code

I also didn't mean to imply that people can't just do PRs and ask for forgiveness later :-) It's just that if you use the word "request" it implies discussion, and then it all goes down hill from there. 

So - a rule of thumb perhaps is: does your issue come with some kind of implementation guidance such that somebody who wasn't party to the discussion might pick it up and do a PR for it? A question, as with #1016 like "can it be done with Qt without writing platform-specific code?" is exactly the sort of thing that may lead to "OK let's do Y then instead" (possibly after numerous posts making it very hard to work out what's happening to the ticket).

gilgongo: Thinking some more about the original template, I think there's some wisdom in perhaps reminding the user to list alternatives that have been considered, since that would be good to to effectively summarise the issues in the discussion that usually go "Let's do X!" "No, because [reason]" "OK let's do Y then". That might prevent people wanting to discuss it all over again in the Issue.

.github/ISSUE_TEMPLATE/feature_request.md
gilgongo: I agree. I'll add that back in (slightly re-worded to emphasise the hope that it's been discussed and alternatives thereby discounted).

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -1,21 +1,21 @@
...
+name: Feature
+about: Agreed specification for a feature or enhancement
 title: ''
-labels: 'feature request'
+labels: 'feature'
 assignees: ''
 ---
 
-<!-- Thanks for opening a new feature request! Please keep in mind that Jamulus achieves low latency audio by reducing complexity and following the "Keep it simple and stupid" approach. Adding new features may result in more processing and maintenance overhead, so please be sure that your feature can not be achieved by different approaches/software too. Make it clear, why here. -->
+<!-- Thanks for opening a new feature request! -->
```
hoffie, line 9: Should this skip "request" as well if we intentionally remove it from the label as well?
```suggestion
<!-- Thanks for suggesting a new feature! -->
```

Or
```suggestion
<!-- Thanks for specifying a new feature! -->
```
?

.github/ISSUE_TEMPLATE/feature_request.md
hoffie: I really like the current version of the PR! :)

> It's just that if you use the word "request" it implies discussion, and then it all goes down hill from there.

Agreed.

> A question, as with #1016 like "can it be done with Qt without writing platform-specific code?" is exactly the sort of thing that may lead to "OK let's do Y then instead" (possibly after numerous posts making it very hard to work out what's happening to the ticket).

I think we are heading in the same direction, but have different tendencies where to draw the line. "Can it be done with Qt" sounds highly implementation-specific to me.
My mental model would be essentially, sorted from vague to specific:
- **Discussions** start with vague ideas, include discussions about different approaches and get lots of (valuable!) side-discussions from other people chiming in. Example "We should have input gain adjustments in Jamulus"
- **Issues** would be the "Intent to implement", e.g. "I want to make Jamulus work for people who have no other means to adjust their input volume. I'm therefore going to allow the server to mix people's input above 100%".
- **PRs**  obviously show a concrete way to implement a given feature. They will still be the place where code-centric discussions will (and IMHO should) happen, e.g. "Why did you implement the gain adjustment in function X instead of Y?". A real-world example for the "should we use Qt or a native function" is #967.

I understand that the goal is to keep issues manageable and actionable in order to make it possible to do some project management on them. The current status is -- or was, thanks to your cleanup work! -- that there is a high number of open issues, some which receive lots of attention, some with almost none. So I fully understand that we should err on the side of keeping things a bit too strict for now.

Anyway, just wanted to share my thoughts and I'm certainly open to try it out as documented in this PR. We could still adjust later.

hoffie (approved):

.github/ISSUE_TEMPLATE/feature_request.md
pljones: >  "Can it be done with Qt" sounds highly implementation-specific to me.

Jamulus is based on Qt libraries for cross-platform support.  So it's the opposite of being implementation-specific, really.  The _aim_ is to use Qt facilities where available, to _avoid_ writing implementation-specific code.

hoffie: > Jamulus is based on Qt libraries for cross-platform support. So it's the opposite of being implementation-specific, really. The aim is to use Qt facilities where available, to avoid writing implementation-specific code.

I think the aim is clear. I was trying to refer to the stage where this is relevant: Such questions are tied to the actual implementation stage (most likely in the PR itself) and are unlikely to come up in the preceding discussion about a new future. All I'm saying is that pointing someone to use Qt instead of native APIs is not that groundbreaking that it would require an existing issue to be converted "back" to a discussion. In other words, I'd consider this to be a *coding detail* (in your words).

Just trying to clear up the misunderstanding. I really don't want to be the one bike-shedding over such details. :)

pljones: No, the discussion about the implementation _DOES NOT_ happen in the PR.  The PR is discussion about the specific code submitted.  The _proposed implementation_ should have already been agreed _before_ any code is written - otherwise it's a waste of time and potentially annoying and disappointing for contributors.

hoffie: Yes, I fully agree that this is how it is supposed to be. I was merely describing what I observe in issues/PR currently.

As this PR is about improving exactly this and defining how it should be, I should probably have raised my concerns elsewhere.

gilgongo: In @hoffie's defence though, I think we shouldn't really discourage "drive-by PRs" too much in the same way as we shouldn't discourage the posting of un-discussed Issues too much. They are both communications of intent after all. For all I know, it may be potentially annoying and disappointing for contributors of issues or discussions if those get tossed out too.

.github/ISSUE_TEMPLATE/feature_request.md
gilgongo: Yes I was wondering about that too. I guess it's logical :-)

.github/ISSUE_TEMPLATE/feature_request.md
pljones: I agreed - this is just so that the templates are clearer, though.

Ideally, there would be a PR template that explained the reviewer expectations of how code should have arrived ready for PR, with pointers to relevant documentation, too.

None of this prevents someone "just doing it".  It will then get handled appropriately.

gilgongo:
Are we OK on this now? Basically, we can't stop (and don't want to stop) people opening issues for feature requests as opposed to bugs, but we're saying that if they do this, they we **prefer** them to have had some discussion first. And to keep the issues sane we'll just move them to discussion if needed because there's a blurred line between what's "an actionable thing that we can put on a milestone" and "a rambling exploration of a possible idea".

ann0see (approved):
I hope that the label "feature" will be created. All in all, I think we can just merge this.

gilgongo:
> I hope that the label "feature" will be created.

Do you mean an issue label? We already have one called "feature request".

ann0see:
> Do you mean an issue label?

Yes. Let's wait and see what happens somebody opened an issue with the new template

Issue #966, opened 2021-02-07
Feature request: Automatic channel fader adjustments


JohannesBrx:
Adjusting the channel faders may be feasible for a small number of musicians but gets quite tedious for a **larger choir** with more than 20 channels. An automatic channel adjustment would make the setup much easier. It also would allow the less experienced users to initialize the faders to a reasonable basic setting. The automatic adjustment could be executed during a uniform part of the music piece or during warming up the voice.

I am aware of the feature loading and saving mixer configurations. However, non-reproducible client microphone configurations between different sessions will require frequent adjustments of the mixer configuration and thus require a more or less complicated exchange of such configuration. Finally, someone currently has to manually adjust the levels.

One idea would be computing a moving average of the channel levels transmitted by the server, e.g., in `CAudioMixerBoard::SetChannelLevels`. In my prototype, I use an exponential moving average like the following 

`vecAvgLevels[iChId] = (1. - alpha) * vecAvgLevels[iChId] + alpha * vecChannelLevel[i];`

to compute the volume over several frames, where `vecAvgLevels` is a `CVector` with the averaged channel levels. The parameter `alpha` controls averaging speed: With `alpha=1.0`, the channel level from the server is directly taken, while, e.g., `alpha=0.1` would mean an extended averaging. This kind of averaging would introduce almost no additional computational load and reduce the quantization of the levels as a side-effect.

The volume could then be set by a new menu entry like "Auto-Adjust all Faders". A basic approach could be setting the fader value anti-proportionally to the estimated channel level. In my current version, I would set a near overdriven level to a fader value of 0.4.

When I am interpreting the sources correctly, the server level update interval should be at 266.7 milliseconds. (Computed from a sample rate of 48 kHz, the `CHANNEL_LEVEL_UPDATE_INTERVAL` of 200 and 64 bytes frame size.)

ann0see:
@JohannesBrx

Thanks for opening this issue and sorry for the late reply here.
As far as I understand, you already have a prototype?

My choir also has problems with multiple microphones which are too loud/too quiet, so personally, I'd like to have a feature like this in the client, but since the server does all the mixing work, I assume it's not possible easily?

Please keep in mind that every feature could produce more latency and server load. I'm afraid that this might violate the KISS principle: https://github.com/jamulussoftware/jamulus/discussions/915#discussion-2242562

gene96817:
[...]

PR #1071, opened 2021-02-19
Auto-adjustment of channel faders

JohannesBrx:
This pull request (for issue #966) adds an automatic adjustment of the channel faders depending on the volume of each channel via a new menu entry `Auto-Adjust all Faders`. This shall specifically support larger choirs to get a reasonable mixer setting, although individual adjustments might still be necessary. The function could be called during voice warm up or a uniform part of the music piece. Adjusting the large number of channels manually is quite tedious and might be required for each session because the participants might not use the same microphone levels each time.

**Latest demo version**: https://github.com/jamulussoftware/jamulus/actions/runs/627676186
(Updated Mar, 6th 2021)

### Application

The channel adjustment is best used right before the end of a warm-up singing phrase. When being called, approximately the **previous** 5 seconds are considered for channel adjustment. The shortcut CTRL+F or CMD+F is handy for quick access.

When using channel groups, all groups are equalized such that their contribution is equal. In addition to that, individual channels within a group are also equalized to contribute equally to the group.

### Implementation details

The method averages the level meters for each channel, which get their input from the server. Averaging is particularly necessary to overcome the coarse quantization of the server values: The 4 bit meter level values are given in decibels ranging from -50dB to 0dB in 9 steps from 0...8, while each step is 6.25dB. (The meter value 9 is reserved for indicating clipping.) 

image

The adjustment computes the channel fader positions for a target volume of -30dB. Therefore, the channel fader for a channel with -30dB original volume is set to its maximum, thus not reducing its original volume. Louder channels are attenuated accordingly. Channels quieter than -30dB are too weak for a full compensation: Although the channel fader will be set its maximum for these channels, they cannot be fully compensated.

### Averaging the meter level

Averaging the current meter level `channelLevel` is currently done with code like
```
            avgLevel = (1.0f - alpha) * avgLevel + alpha * channelLevel
```
where `alpha=0.2` and `avgLevel` is the averaged level. The update of `avgLevel` can be seen in the following figure. The dots indicate the update intervals of 266 milliseconds. In practice, convergence might be faster because you would probably not start from silence.

image

### Accuracy improvement

So far, this implementation only affects the client.

However, the accuracy of this approach is negatively influenced by the low accuracy (4 bit) of the level meter value from the server, which has a resolution of 6.25dB. While (currently) not being implemented in this PR, I investigated using 8 bit instead 4 bit for the level value, resulting in a much better precision. However, this would also increase the data stream: The level is sent every 200 audio frames. Typical Jamulus UDP packages (medium quality, mono, 128 bytes audio buffer) are 88 bytes according to Wireshark. The channel level list package is 60 bytes for one channel. For a typical choir of 40 persons, the data rate would increase from 88*200+59+40/2=17679 to 88*200+59+40/2=17699, which means 0.11% increase. The bandwidth for one channel would increase from 369kbps to 369.42kbps, when I computed everything correctly.

Alternatively, one could think about a new protocol message which is used to request recommended channel fader values from the server, which would only be sent on demand.

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1316,15 +1316,15 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
     QMutexLocker locker ( &Mutex );
 
     // initialize variables used for statistics
-    float vecMinLevel[MAX_NUM_FADER_GROUPS + 1];
     float vecMaxLevel[MAX_NUM_FADER_GROUPS + 1];
     int   vecChannelsPerGroup[MAX_NUM_FADER_GROUPS + 1];
     for ( int i = 0; i < MAX_NUM_FADER_GROUPS + 1; ++i )
     {
-        vecMinLevel[i] = UPPER_BOUND_SIG_METER;
         vecMaxLevel[i] = LOW_BOUND_SIG_METER;
         vecChannelsPerGroup[i] = 0;
     }
+    CVector<CVector<float>> levels;
+    levels.resize ( MAX_NUM_FADER_GROUPS + 1 );
 
     // compute min/max level per group and number of channels per group
     for ( int i = 0; i < MAX_NUM_CHANNELS; ++i )
@@ -1344,29 +1344,37 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
 
             if ( leveldB >= AUTO_FADER_NOISE_THRESHOLD_DB )
             {
-                vecMinLevel[group] = fmin ( vecMinLevel[group], leveldB );
                 vecMaxLevel[group] = fmax ( vecMaxLevel[group], leveldB );
+                levels[group].Add ( leveldB );
             }
             ++vecChannelsPerGroup[group];
         }
     }
 
+    // sort levels for later median computation
+    for ( int i = 0; i < MAX_NUM_FADER_GROUPS + 1; ++i )
+    {
+        std::sort ( levels[i].begin(), levels[i].end() );
+    }
+
     // compute the number of active groups (at least one channel)
     int cntActiveGroups = 0;
-    for ( int i = 0; i < MAX_NUM_FADER_GROUPS + 1; ++i )
+    for ( int i = 0; i < MAX_NUM_FADER_GROUPS; ++i )
     {
         cntActiveGroups += vecChannelsPerGroup[i] > 0;
     }
 
     // only my channel is active, nothing to do
-    if ( cntActiveGroups == 0 )
+    if ( cntActiveGroups == 0 &&
+        vecChannelsPerGroup[MAX_NUM_FADER_GROUPS] == 0 )
     {
         return;
     }
 
     // compute target level for each group
     // (prevent clipping when each group contributes at maximum level)
-    float targetLevelPerGroup = -20.0f * log10 ( cntActiveGroups );
+    float targetLevelPerGroup = -20.0f * log10 (
+        std::max ( cntActiveGroups, 1 ) );
 
     // compute target levels for the channels of each group individually
     float vecTargetChannelLevel[MAX_NUM_FADER_GROUPS + 1];
@@ -1381,15 +1389,21 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
             targetLevelPerGroup - 20.0f * log10 ( vecChannelsPerGroup[i] ) :
             0.0f;
 
+        // get median level
+        int cntChannels = levels[i].Size();
+        if ( cntChannels == 0 )
+            continue;
+        float refLevel = levels[i][cntChannels / 2];
+
         // since we can only attenuate channels but not amplify, we have to
-        // check that the weakest (min) channel can be brought to the target
+        // check that the reference channel can be brought to the target
         // level
-        if ( vecTargetChannelLevel[i] > vecMinLevel[i] )
+        if ( refLevel < vecTargetChannelLevel[i] )
         {
             // otherwise, we adjust the level offset in such a way that
             // the level can be reached
             levelOffset = fmin ( levelOffset,
-                vecMinLevel[i] - vecTargetChannelLevel[i] );
+                refLevel - vecTargetChannelLevel[i] );
 
             // compute the minimum necessary fader setting
             minFader = fmin ( minFader, -vecMaxLevel[i] +
@@ -1418,7 +1432,17 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
 
             int group = vecpChanFader[i]->GetGroupID();
             if ( group == INVALID_INDEX )
-                group = 4;
+            {
+                if ( cntActiveGroups > 0 )
+                {
+                    // do not adjust the channels without group in group mode
+                    continue;
+                }
+                else
+                {
+                    group = 4;
+                }
+            }
 
             // do not adjust channels with almost zero level to full level since
             // the channel might simply be silent at the moment
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1317,37 +1317,28 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
     for ( int i = 0; i < MAX_NUM_CHANNELS; ++i )
     {
         // only apply to visible faders (and not to my own channel fader)
-        if ( vecpChanFader[i]->IsVisible() && ( true /* i != iMyChannelID */ ) )
+        if ( vecpChanFader[i]->IsVisible() && ( i != iMyChannelID ) )
         {
-            // get averaged level meter value (range 0...8)
-            float levelMeter = vecAvgLevels[i];
-
-            // map meter output level to deciber (range -50dB...0dB)
+            // map averaged meter output level to decibels
             // (invert CStereoSignalLevelMeter::CalcLogResultForMeter)
-            float leveldB = levelMeter *
+            float leveldB = vecAvgLevels[i] *
                 ( UPPER_BOUND_SIG_METER - LOW_BOUND_SIG_METER ) /
                 NUM_STEPS_LED_BAR + LOW_BOUND_SIG_METER;
 
-            // map decibels to linear level
-            // level = pow ( 10.0f, level / 20.0f );
-
             // do not adjust channels with zero level to full level since
             // the channel might simply be silent at the moment
-            if ( leveldB < -40.0f )
+            if ( leveldB < AUTO_FADER_NOISE_THRESHOLD_DB )
             {
-                // specify target level
-                float targetdBLevel = -30.0f;
-
                 // compute new level
-                float newdBLevel = -leveldB + targetdBLevel;
+                float newdBLevel = -leveldB + AUTO_FADER_TARGET_LEVEL_DB;
 
-                // map range from -35..0 dB to 0..1
-                float newVal = (newdBLevel + 35.0f) / 35.0f;
+                // map range from decibels to fader level
+                // (this inverts MathUtils::CalcFaderGain)
+                float newFaderLevel = (newdBLevel + 35.0f) / 35.0f *
+                    AUD_MIX_FADER_MAX;
 
                 // set fader level
-                vecpChanFader[i]->SetFaderLevel ( newVal * AUD_MIX_FADER_MAX, true );
-
-                // printf("%.2f dB => %.2f dB\n", leveldB, newdBLevel);
+                vecpChanFader[i]->SetFaderLevel ( newFaderLevel, true );
             }
         }
     }
@@ -1551,13 +1542,13 @@ void CAudioMixerBoard::SetChannelLevels ( const CVector<uint16_t>& vecChannelLev
 {
     const int iNumChannelLevels = vecChannelLevel.Size();
     int       i                 = 0;
-    float     alpha             = AUTO_FADER_ADJUST_ALPHA;
 
     for ( int iChId = 0; iChId < MAX_NUM_CHANNELS; iChId++ )
     {
         if ( vecpChanFader[iChId]->IsVisible() && ( i < iNumChannelLevels ) )
         {
             // compute exponential moving average
+            float alpha = AUTO_FADER_ADJUST_ALPHA;
             vecAvgLevels[iChId] = (1.0f - alpha) * vecAvgLevels[iChId] +
                 alpha * vecChannelLevel[i];
 
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -857,7 +857,7 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent ) :
     // create all mixer controls and make them invisible
     vecpChanFader.Init ( MAX_NUM_CHANNELS );
 
-    vecAvgLevels.Init ( MAX_NUM_CHANNELS );
+    vecAvgLevels.Init ( MAX_NUM_CHANNELS, 0.0f );
 
     for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
     {
@@ -1147,6 +1147,7 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
                     {
                         // the fader was not in use, reset everything for new client
                         vecpChanFader[i]->Reset();
+                        vecAvgLevels[i] = 0.0f;
 
                         // check if this is my own fader and set fader property
                         if ( i == iMyChannelID )
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1327,7 +1327,7 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
 
             // do not adjust channels with zero level to full level since
             // the channel might simply be silent at the moment
-            if ( leveldB < AUTO_FADER_NOISE_THRESHOLD_DB )
+            if ( leveldB > AUTO_FADER_NOISE_THRESHOLD_DB )
             {
                 // compute new level
                 float newdBLevel = -leveldB + AUTO_FADER_TARGET_LEVEL_DB;
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1314,6 +1314,18 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
 {
     QMutexLocker locker ( &Mutex );
 
+    // initialize variables used for statistics
+    float vecMinLevel[MAX_NUM_FADER_GROUPS + 1];
+    float vecMaxLevel[MAX_NUM_FADER_GROUPS + 1];
+    int   vecChannelsPerGroup[MAX_NUM_FADER_GROUPS + 1];
+    for ( int i = 0; i < MAX_NUM_FADER_GROUPS + 1; ++i )
+    {
+        vecMinLevel[i] = UPPER_BOUND_SIG_METER;
+        vecMaxLevel[i] = LOW_BOUND_SIG_METER;
+        vecChannelsPerGroup[i] = 0;
+    }
+
+    // compute min/max level per group and number of channels per group
     for ( int i = 0; i < MAX_NUM_CHANNELS; ++i )
     {
         // only apply to visible faders (and not to my own channel fader)
@@ -1325,20 +1337,105 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
                 ( UPPER_BOUND_SIG_METER - LOW_BOUND_SIG_METER ) /
                 NUM_STEPS_LED_BAR + LOW_BOUND_SIG_METER;
 
-            // do not adjust channels with zero level to full level since
+            int group = vecpChanFader[i]->GetGroupID();
+            if ( group == INVALID_INDEX )
+                group = 4;
+
+            if ( leveldB >= AUTO_FADER_NOISE_THRESHOLD_DB )
+            {
+                vecMinLevel[group] = fmin ( vecMinLevel[group], leveldB );
+                vecMaxLevel[group] = fmax ( vecMaxLevel[group], leveldB );
+            }
+            ++vecChannelsPerGroup[group];
+        }
+    }
+
+    // compute the number of active groups (at least one channel)
+    int cntActiveGroups = 0;
+    for ( int i = 0; i < MAX_NUM_FADER_GROUPS + 1; ++i )
+    {
+        cntActiveGroups += vecChannelsPerGroup[i] > 0;
+    }
+
+    // only my channel is active, nothing to do
+    if ( cntActiveGroups == 0 )
+        return;
+
+    // compute target level for each group
+    // (prevent clipping when each group contributes at maximum level)
+    float targetLevelPerGroup = -20.0f * log10 ( cntActiveGroups );
+
+    // compute target levels for the channels of each group individually
+    float vecTargetChannelLevel[MAX_NUM_FADER_GROUPS + 1];
+    float levelOffset = 0.0f;
+    float minFader = 0.0f;
+    for ( int i = 0; i < MAX_NUM_FADER_GROUPS + 1; ++i )
+    {
+        // compute the target level for each channel in the current group
+        // (prevent clipping when each channel in this group contributes at
+        // the maximum level)
+        vecTargetChannelLevel[i] = vecChannelsPerGroup[i] > 0 ?
+            targetLevelPerGroup - 20.0f * log10 ( vecChannelsPerGroup[i] ) :
+            0.0f;
+
+        // since we can only attenuate channels but not amplify, we have to
+        // check that the weakest (min) channel can be brought to the target
+        // level
+        if ( vecTargetChannelLevel[i] > vecMinLevel[i] )
+        {
+            // otherwise, we adjust the level offset in such a way that
+            // the level can be reached
+            levelOffset = fmin ( levelOffset,
+                vecMinLevel[i] - vecTargetChannelLevel[i] );
+
+            // compute the minimum necessary fader setting
+            minFader = fmin ( minFader, -vecMaxLevel[i] +
+                vecTargetChannelLevel[i] + levelOffset );
+        }
+    }
+
+    // take minimum fader value into account
+    // very weak channels would actually require strong channels to be
+    // attenuated to a large amount; however, the attenuation is limited by
+    // the faders
+    if ( minFader < -35.0f )
+        levelOffset += -35.0f - minFader;
+
+    // adjust all levels
+    for ( int i = 0; i < MAX_NUM_CHANNELS; ++i )
+    {
+        // only apply to visible faders (and not to my own channel fader)
+        if ( vecpChanFader[i]->IsVisible() && ( i != iMyChannelID ) )
+        {
+            // map averaged meter output level to decibels
+            // (invert CStereoSignalLevelMeter::CalcLogResultForMeter)
+            float leveldB = vecAvgLevels[i] *
+                ( UPPER_BOUND_SIG_METER - LOW_BOUND_SIG_METER ) /
+                NUM_STEPS_LED_BAR + LOW_BOUND_SIG_METER;
+
+            int group = vecpChanFader[i]->GetGroupID();
+            if ( group == INVALID_INDEX )
+                group = 4;
+
+            // do not adjust channels with almost zero level to full level since
             // the channel might simply be silent at the moment
-            if ( leveldB > AUTO_FADER_NOISE_THRESHOLD_DB )
+            if ( leveldB >= AUTO_FADER_NOISE_THRESHOLD_DB )
             {
                 // compute new level
-                float newdBLevel = -leveldB + AUTO_FADER_TARGET_LEVEL_DB;
+                float newdBLevel = -leveldB + vecTargetChannelLevel[group] +
+                    levelOffset;
 
                 // map range from decibels to fader level
                 // (this inverts MathUtils::CalcFaderGain)
-                float newFaderLevel = (newdBLevel + 35.0f) / 35.0f *
+                float newFaderLevel = ( newdBLevel + 35.0f ) / 35.0f *
                     AUD_MIX_FADER_MAX;
 
+                // limit fader
+                newFaderLevel = fmin ( fmax ( newFaderLevel, 0.0f),
+                    float ( AUD_MIX_FADER_MAX ) );
+
                 // set fader level
-                vecpChanFader[i]->SetFaderLevel ( newFaderLevel, true );
+                vecpChanFader[i]->SetFaderLevel ( newFaderLevel, false );
             }
         }
     }
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1340,7 +1340,9 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
 
             int group = vecpChanFader[i]->GetGroupID();
             if ( group == INVALID_INDEX )
-                group = 4;
+            {
+                group = MAX_NUM_FADER_GROUPS;
+            }
 
             if ( leveldB >= AUTO_FADER_NOISE_THRESHOLD_DB )
             {
@@ -1392,7 +1394,9 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
         // get median level
         int cntChannels = levels[i].Size();
         if ( cntChannels == 0 )
+        {
             continue;
+        }
         float refLevel = levels[i][cntChannels / 2];
 
         // since we can only attenuate channels but not amplify, we have to
@@ -1440,7 +1444,7 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
                 }
                 else
                 {
-                    group = 4;
+                    group = MAX_NUM_FADER_GROUPS;
                 }
             }
 
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1399,8 +1399,8 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
     // very weak channels would actually require strong channels to be
     // attenuated to a large amount; however, the attenuation is limited by
     // the faders
-    if ( minFader < -35.0f )
-        levelOffset += -35.0f - minFader;
+    if ( minFader < -AUD_MIX_FADER_RANGE_DB )
+        levelOffset += -AUD_MIX_FADER_RANGE_DB - minFader;
 
     // adjust all levels
     for ( int i = 0; i < MAX_NUM_CHANNELS; ++i )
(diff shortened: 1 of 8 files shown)


gene96817:
I love the idea of this feature.

1- when is this computation and adjustment done?
2- How do we freeze the settings? Some pieces have a wide dynamic range (ppp to fff). 
     We don't want this to be a continuous adjustment that kills the dynamic range and expression of the music.
3- How do we undo the freeze and get a new adjustment?
4- what is the recommended usage for a chorus where some sections are of different sizes? The balance is made (heard?) by sections and not by singers; meaning, each singer does not have the same weight in sound contribution.

src/audiomixerboard.cpp
```diff
@@ -1305,6 +1307,49 @@ void CAudioMixerBoard::SetAllFaderLevelsToNewClientLevel()
...
+{
+    QMutexLocker locker ( &Mutex );
+
+    for ( int i = 0; i < MAX_NUM_CHANNELS; ++i )
+    {
+        // only apply to visible faders (and not to my own channel fader)
+        if ( vecpChanFader[i]->IsVisible() && ( true /* i != iMyChannelID */ ) )
+        {
+            // get averaged level meter value (range 0...8)
+            float levelMeter = vecAvgLevels[i];
```
pljones, line 1320: This appears to be referenced only once.  I'd rather it didn't exist and `vecAvgLevels[i]` was referenced directly - it makes the code easier to follow.

src/audiomixerboard.cpp
```diff
@@ -1305,6 +1307,49 @@ void CAudioMixerBoard::SetAllFaderLevelsToNewClientLevel()
...
 }
 
+void CAudioMixerBoard::AutoAdjustAllFaderLevels()
+{
+    QMutexLocker locker ( &Mutex );
+
+    for ( int i = 0; i < MAX_NUM_CHANNELS; ++i )
+    {
+        // only apply to visible faders (and not to my own channel fader)
+        if ( vecpChanFader[i]->IsVisible() && ( true /* i != iMyChannelID */ ) )
```
pljones, line 1317: I'm guessing this is temporary `true /* i != iMyChannelID */`?

src/audiomixerboard.cpp
```diff
@@ -1305,6 +1307,49 @@ void CAudioMixerBoard::SetAllFaderLevelsToNewClientLevel()
...
+
+            // map decibels to linear level
+            // level = pow ( 10.0f, level / 20.0f );
+
+            // do not adjust channels with zero level to full level since
+            // the channel might simply be silent at the moment
+            if ( leveldB < -40.0f )
+            {
+                // specify target level
+                float targetdBLevel = -30.0f;
```
pljones, line 1336: This is a const by the look of it, right?  Definitely shouldn't be assigned this deep inside the method (okay, compiler probably sorts that out but style-wise, it's best defined somewhere as a constant).

src/audiomixerboard.cpp
```diff
@@ -1305,6 +1307,49 @@ void CAudioMixerBoard::SetAllFaderLevelsToNewClientLevel()
...
+
+    for ( int i = 0; i < MAX_NUM_CHANNELS; ++i )
+    {
+        // only apply to visible faders (and not to my own channel fader)
+        if ( vecpChanFader[i]->IsVisible() && ( true /* i != iMyChannelID */ ) )
+        {
+            // get averaged level meter value (range 0...8)
+            float levelMeter = vecAvgLevels[i];
+
+            // map meter output level to deciber (range -50dB...0dB)
```
pljones, line 1322: Is the "-50dB" the `targetdBLevel`?  Or the -40f or the 35f?  Ideally name all of these values then use the names.

src/audiomixerboard.cpp
```diff
@@ -1503,11 +1548,16 @@ void CAudioMixerBoard::SetChannelLevels ( const CVector<uint16_t>& vecChannelLev
 {
     const int iNumChannelLevels = vecChannelLevel.Size();
     int       i                 = 0;
+    float     alpha             = AUTO_FADER_ADJUST_ALPHA;
```
pljones, line 1551: Why do this?

src/global.h
```diff
@@ -146,6 +146,9 @@ LED bar:      lbr
 #define AUD_MIX_FADER_MAX                100
 #define AUD_MIX_PAN_MAX                  100
 
+// coefficient for averaging channel levels for automatic fader adjustment
+#define AUTO_FADER_ADJUST_ALPHA          0.2
```
pljones, line 150: `0.2` should be `0.2f`?

pljones:
OK, the code looks pretty good.  I reviewed _the code_, of course.  It needs just a few changes to capture what you've put in the (very good) description here.  (Plus a couple of minor tweaks.)

JohannesBrx:
> 1- when is this computation and adjustment done?

It is done via a menu currently called `Auto-Adjust all Faders` right below the existing `Set All Faders to New Client Level`. You have to start it manually.

> 2- How do we freeze the settings? Some pieces have a wide dynamic range (ppp to fff).
> We don't want this to be a continuous adjustment that kills the dynamic range and expression of the music.
> 3- How do we undo the freeze and get a new adjustment?

Since this feature is run once, you will not have a continuous adaptation.

> 4- what is the recommended usage for a chorus where some sections are of different sizes? The balance is made (heard?) by sections and not by singers; meaning, each singer does not have the same weight in sound contribution.

Good point! You could group the voices and adjust whole groups at once. We could also think about considering groups in the computation of the levels. However, I am not sure whether this might add a kind a secret logic which leads to confusion!?

pljones:
> We could also think about considering groups in the computation of the levels.

It would be useful if the grouping feature could be taken into account.  (So "per group" and "all groups" options, if groups are set?)

> However, I am not sure whether this might add a kind a secret logic which leads to confusion!?

People using grouping probably understand what they're using it for.  People not using it won't be affected.

gene96817:
#4 is hard. Very useful for choirs, especially barbershop. 

I've never noticed/seen the Auto-Adjust all Faders. I'll have to give it a try.

src/audiomixerboard.cpp
JohannesBrx: Yes, I will switch to the originally intended behavior such that it is the same as for `Set All Faders to New Client Level`.

pljones:
> I've never noticed/seen the Auto-Adjust all Faders. I'll have to give it a try.

You'll need to build JohannesBrx/jamulus/AutoAdjustChannelFaders to get it.

src/audiomixerboard.cpp
JohannesBrx: Yes, I will switch to the originally intended behavior such that it is the same as for `Set All Faders to New Client Level`.

src/audiomixerboard.cpp
JohannesBrx: This is just to be able to use the short `alpha` below instead the bulky define.

I will move `alpha` directly to the code line concerning the averaging:
```
            const float alpha = AUTO_FADER_ADJUST_ALPHA;
            vecAvgLevels[iChId] = (1.0f - alpha) * vecAvgLevels[iChId] +
                alpha * vecChannelLevel[i];
```

Or would you prefer removing `alpha` completely and using `AUTO_FADER_ADJUST_ALPHA` instead?

src/audiomixerboard.cpp
JohannesBrx: I will introduce constants for both the threshold level and the target level on `global.h`.

JohannesBrx:
> OK, the code looks pretty good. I reviewed _the code_, of course. It needs just a few changes to capture what you've put in the (very good) description here. (Plus a couple of minor tweaks.)

It will be important to gain some experience of the approach in practice, especially whether the accuracy using the current 4 bit level meter values is sufficient.

Up to know I verified the correctness of the implementation by using a sine wave input signal at different (specified) amplitude levels, applying the method and then recording the output in Audacity. I also did a quick test during our choir rehearsal. The adjusted faders looked at least _plausible_ to me ;-) I currently plan doing other tests with voice samples at different well-defined amplitude levels.

JohannesBrx:
> Oh, or here
> https://github.com/jamulussoftware/jamulus/actions/runs/582519157
> (can't remember how long the PR autobuild artifacts live for - 30 days, I think)

It should be available for 31 days according to the build log (`retention-days: 31`). It's a really great thing to automatically get preview versions for all operating systems.

JohannesBrx:
**Important note**: For live tests please use https://github.com/jamulussoftware/jamulus/actions/runs/583730608, which contains an important bug fix.

I performed a controlled test with multiple sine wave blocks at 440Hz and various amplitude levels. I fed this signal into Jamulus as the input or _microphone_ signal (see figure below). The amplitudes range from 0dB down to -40dB in steps of -2dB and each amplitude level lasts for 5 seconds. Jamulus was run in a special mode where the fader adjustment is constantly applied -- note that this is just for testing. Finally, I recorded the output signal (_earphone signal_) from Jamulus. The amplitude ranges are also denoted in the label track.

One can see that the input signal is constantly being adapted and the output signal remains in a certain corridor, which is limited by the quantization of the levels from the server. For approximately three neighboring amplitude level blocks (spanning 6dB) the level meter value is the same. Therefore, the fader cannot be adjusted in this small range. However, in practice, the accuracy might be better because typical music or voice signals will exhibit a larger amplitude range. Averaging the resulting different level meter values might then improve the accuracy.
image

hoffie:
@JohannesBrx Thanks for your work! Your tesst look really thorough, I'll give this a try with my choir later today and will report back how it works in practice.

JohannesBrx:
> @JohannesBrx Thanks for your work! Your tesst look really thorough, I'll give this a try with my choir later today and will report back how it works in practice.

Great! I'm looking forward to your feedback.

Be sure to get the latest version with the bug fix under https://github.com/jamulussoftware/jamulus/actions/runs/583730608.

gene96817:
As I understand the past discussion, this auto-adjustment is manually triggered.
It is thought that the adjustment can be made during warm-up.

Q1: how long is the sampling period?

Q2: What kind of "warm up" is best for the auto-adjustment?
        Should it be like the stereotypical orchestra all tuning up on A?
        Should the chorus have every voice be singing a chord tuning sequence?
        If an ensemble plays a section of their music, what happens if not every voice (line in the music) is equally active during the auto-adjustment period? This is a question on the impact of a pause (rest) in the music. What is the impact of the mean power between the voices are not equal? 
        What is the best sound sample from the ensemble for the auto-adjustment?

JohannesBrx:
> Q1: how long is the sampling period?

There is no explicit sampling period. Instead, I use an exponential moving average (I updated the first comment for the PR with a figure). In extreme cases where you start from silence to full signal it would take 5 seconds for the average to be converged or _stable_. 

In practice I would assume that _2 seconds_ should be sufficient.

> Q2: What kind of "warm up" is best for the auto-adjustment?
> Should it be like the stereotypical orchestra all tuning up on A?
> Should the chorus have every voice be singing a chord tuning sequence?
> If an ensemble plays a section of their music, what happens if not every voice (line in the music) is equally active during the auto-adjustment period? This is a question on the impact of a pause (rest) in the music. What is the impact of the mean power between the voices are not equal?
> What is the best sound sample from the ensemble for the auto-adjustment?

At the moment I would suggest that the whole choir sings an _Ahhh_ vocal for at least 3 seconds and the auto-adjustment is triggered a bit before the end of the sequence. The frequency should not play an important role for the adjustment itself. However, it should be a tone pitch which is comfortable for each singer.

Applying the auto-adjustment at a non-uniform section of the music will most probably not work correctly.

gene96817:
Thanks for the suggestion for giving the auto-adjustment a sample.
It sounds like the process would be:
(a) have a chorus tune up a chord at fortissimo
(b) repeat the tuning and adjust the sections for the balance desired
(c) repeat (b) and when the sound is stable, trigger the auto-adjustment and hold for 5 seconds (to ensure a stable 2-second sample).

pgScorpio:
I think Auto-adjustment for volume would be a great thing to have, but aren't you trying to solve it at the wrong side ??
Wouldn't be a (auto) volume control on the client side be a much better (and simpler) solution ?
I think **any** kind of mic. volume control at the client side would already be a **big** improvement anyway.

And a simple 1.5 or 3dB Step volume control would be easy to implement with just a few shift and add instructions.

src/audiomixerboard.cpp
```diff
@@ -1508,6 +1544,11 @@ void CAudioMixerBoard::SetChannelLevels ( const CVector<uint16_t>& vecChannelLev
     {
         if ( vecpChanFader[iChId]->IsVisible() && ( i < iNumChannelLevels ) )
         {
+            // compute exponential moving average
+            float alpha = AUTO_FADER_ADJUST_ALPHA;
```
pljones, line 1548: I think either you or I missed one of a minor comment :).  Can you just use `AUTO_FADER_ADJUST_ALPHA`, rather than creating a variable to hold a constant, please.

src/global.h
```diff
@@ -146,6 +146,16 @@ LED bar:      lbr
 #define AUD_MIX_FADER_MAX                100
 #define AUD_MIX_PAN_MAX                  100
 
+// coefficient for averaging channel levels for automatic fader adjustment
+#define AUTO_FADER_ADJUST_ALPHA          0.2f
+
+// target level for auto fader adjustment in decibels
```
pljones, line 152: I'm wondering if we're not going to want to make this configurable per client.  Is -30dB based on the number of other clients summed levels not clipping?  Say I had twice the number of clients or half the number of clients, am I going to want a different target?

hoffie:
@JohannesBrx So, the test went really well! We were some 16 singers and I tried the auto-adjustment multiple times within a (homophonic) piece (so, not even deliberate Aaahs or something). I'm not saying that I can confirm that it works flawlessly (did not have the time to check during singing). I'm merely saying it works and it is a huge improvement over the status quo. I'm pretty sure this will be a welcome addition especially for larger groups.

@pgScorpio 
Sounds a bit like https://github.com/jamulussoftware/jamulus/discussions/1030, maybe you want to comment there as well?

src/audiomixerboard.cpp
JohannesBrx: I will change that in the next commit. This answers my question above:

> Or would you prefer removing alpha completely and using AUTO_FADER_ADJUST_ALPHA instead?

JohannesBrx:
> It sounds like the process would be:
> (a) have a chorus tune up a chord at fortissimo
> (b) repeat the tuning and adjust the sections for the balance desired
> (c) repeat (b) and when the sound is stable, trigger the auto-adjustment and hold for 5 seconds (to ensure a stable 2-second sample).

Please not that the averaging is done continuously. When you start the auto-adjustment, the current average is taken. Therefore, the averaging considers audio samples _before_ starting the auto-adjustment. The advantage is that you have already heard the sequence which the auto-adjustment operates on. (It also makes implementation in this case easier.)

JohannesBrx:
> I think Auto-adjustment for volume would be a great thing to have, but aren't you trying to solve it at the wrong side ??
> Wouldn't be a (auto) volume control on the client side be a much better (and simpler) solution ?
> I think **any** kind of mic. volume control at the client side would already be a **big** improvement anyway.

I agree with you that having access to the microphone level at the client side would be a great improvement since this is a topic we stumble across each choir rehearsal session. As @hoffie mentioned, this would be a topic to be discussed in #1030.

However, I would not count on every choir (or band) member correctly setting up the microphone level when some of the less experienced participants even need help installing the software. Doing it automatically would require that everyone sings approximately the same _calibration_ tone. In my opinion it would not fit into the concept of Jamulus to automatically trigger the microphone level calibration at client side for all clients. Finally, several audio interfaces have a hardware control knob to adjust the microphone level.

src/global.h
JohannesBrx: I have chosen -30dB as the weakest channel we can support for adjustment. It does _not_ address the number of channels. I have chosen -30dB as a compromise between weaker channels we still want to support (seeing that the level meter value is at the lower half for some choir members)  and the loss of signal strength (and resulting quantization). 

The following table explains my thoughts about the minimum supported channel level:

Channel level | Attenuation | Comment
------------- | ------------ | -----------
0dB | -30dB | Maximum input signal gain
-10dB | -20dB
... | ...
**-30dB** | 0dB | Weakest input signal being supported, fader at its maximum (0dB)
-40dB | (+10dB) | Cannot be compensated, will still be weaker

We could also think about using -35dB as the target level since the faders range from -35dB to 0 dB.

The target level for preventing clipping would be `20 * log10(1. / cntChannels)`, e.g., -20dB for 20 channels or approximately -30dB for 32 channels. However, I think this would only apply when each channel is at its maximum level, which will seldom be the case in practice.

We could also think about setting the target level dynamically: We would take the minimum of the channel levels (weakest channel) and the required attenuation such that we do not get clipping.

gene96817:
@JohannesBrx  I was trying to be very careful. Note:

> Please note that the averaging is done continuously. When you start the auto-adjustment, the current average is taken. Therefore, the averaging considers audio samples before starting the auto-adjustment. 

Here you say "done continuously". I wanted to know the bounds of the measurements. In an earlier message, it was said it was manually triggered. 

Also said, "considers audio samples before starting the auto-adjustment". I want to know how long before the start.

The reason I made explicit what I thought would be the process for a chorus, you can't turn on the sound and amplitude instantly and accurately. So, I expect to ask the chorus to produce the sound they want to use for setting the levels starting before the samples used by the algorithm and held stably past the end of the measurement. The less expert the chorus, the longer it takes for turning on and off the desired sound.d

Thanks.

pgScorpio:
> > I think Auto-adjustment for volume would be a great thing to have, but aren't you trying to solve it at the wrong side ??
> > Wouldn't be a (auto) volume control on the client side be a much better (and simpler) solution ?
> > I think **any** kind of mic. volume control at the client side would already be a **big** improvement anyway.
> 
> I agree with you that having access to the microphone level at the client side would be a great improvement since this is a topic we stumble across each choir rehearsal session. As @hoffie mentioned, this would be a topic to be discussed in #1030.
> 
> However, I would not count on every choir (or band) member correctly setting up the microphone level when some of the less experienced participants even need help installing the software. Doing it automatically would require that everyone sings approximately the same _calibration_ tone. In my opinion it would not fit into the concept of Jamulus to automatically trigger the microphone level calibration at client side for all clients. Finally, several audio interfaces have a hardware control knob to adjust the microphone level.

"Doing it automatically would require that everyone sings approximately the same calibration tone"
I don't get this. what's wrong with just an auto mic level control you can turn on and off? 
Turn auto mic level on, sing a few seconds, mic volume is automatically set. turn auto level off and the auto level level stays set but control switches to manual again...
Just like in many other applications like Zoom.

gene96817:
> "Doing it automatically would require that everyone sings approximately the same calibration tone"
I don't get this. what's wrong with just an auto mic level control you can turn on and off?
Turn auto mic level on, sing a few seconds, mic volume is automatically set. turn auto level off and the auto level level stays set but control switches to manual again...
Just like in many other applications like Zoom.

There are two scenarios. 
In the Zoom case, a user just wants to have a convenient way to set their microphone setting. 

In the more complicated case, the user wants their mic volume set relative to the whole ensemble because later in the music with lots of texture and mood changes, each section has a different volume relative to other sections. To get the right relative volume, you need to hear the relative volume. 

In Zoom the control is more like a vox. You are on or off. If you try to speak softer relative to another speaker, Zoom will ignore you. In music, there is a lot of magic in the changes of relative volume.

JohannesBrx:
> Here you say "done continuously". I wanted to know the bounds of the measurements. In an earlier message, it was said it was manually triggered.
> 
> Also said, "considers audio samples before starting the auto-adjustment". I want to know how long before the start.
> 
> The reason I made explicit what I thought would be the process for a chorus, you can't turn on the sound and amplitude instantly and accurately. So, I expect to ask the chorus to produce the sound they want to use for setting the levels starting before the samples used by the algorithm and held stably past the end of the measurement. The less expert the chorus, the longer it takes for turning on and off the desired sound.d

@gene96817: 
The sampling and averaging of audio samples is done all the time -- even when you do not activate the adjustment. (It is a cheap operation which does not decrease performance.) This is what I meant with _continuously_. The averaging is designed in such a way that approximately the previous 5 seconds influence the averaging value.

At the moment where you start the channel adjustment, the current averaging level is taken and the faders are then directly adjusted. You could say that the current status is frozen and used for adjustment. Only audio samples _before_ starting the adjustment are considered by doing so.

However, there is no distinct time period from when the samples are taken. I tried to visualize that in the following figure: The blue curve is the input level from the server, where dots indicate the discrete sampling interval of 266ms. The orange curve is the averaged level value. You can see that the noise of the input values is reduced but it takes approximately 5 seconds to reach the final value. 

Roughly speaking, when you press CTRL-F or use the menu entry, samples from the last 5 seconds are taken. In practice, I would let the choir sing a phrase with different tones and perform the adjustment at the end of the sequence. The phrase should be at least 5 seconds long.

image

pgScorpio:
> > "Doing it automatically would require that everyone sings approximately the same calibration tone"
> > I don't get this. what's wrong with just an auto mic level control you can turn on and off?
> > Turn auto mic level on, sing a few seconds, mic volume is automatically set. turn auto level off and the auto level level stays set but control switches to manual again...
> > Just like in many other applications like Zoom.
> 
> There are two scenarios.
> In the Zoom case, a user just wants to have a convenient way to set their microphone setting.
> 
> In the more complicated case, the user wants their mic volume set relative to the whole ensemble because later in the music with lots of texture and mood changes, each section has a different volume relative to other sections. To get the right relative volume, you need to hear the relative volume.
> 
> In Zoom the control is more like a vox. You are on or off. If you try to speak softer relative to another speaker, Zoom will ignore you. In music, there is a lot of magic in the changes of relative volume.

Our choir is perfectly capable of keeping relative volume later in the music. The problem is setting the correct volume during setup. So just a simple auto level function to set the microphone level once before performance will do. Auto changing levels during a performance is even a no go!

src/global.h
JohannesBrx: @pljones 
I have now committed a version which takes groups into account, prevents clipping by mixing and limits the overall attenuation (the minimum slider value is -35dB). I have run some tests with test vectors in a separate file (not committed) to verify the implementation. However, I did not yet have the chance to test this version with multiple channels in a real environment.

src/audiomixerboard.cpp
```diff
@@ -1305,6 +1308,137 @@ void CAudioMixerBoard::SetAllFaderLevelsToNewClientLevel()
...
+
+    // compute the number of active groups (at least one channel)
+    int cntActiveGroups = 0;
+    for ( int i = 0; i < MAX_NUM_FADER_GROUPS + 1; ++i )
+    {
+        cntActiveGroups += vecChannelsPerGroup[i] > 0;
+    }
+
+    // only my channel is active, nothing to do
+    if ( cntActiveGroups == 0 )
```
pljones, line 1359: Style - always wrap in braces:
```
if ( X ) {
  return;
}
```

src/audiomixerboard.cpp
```diff
@@ -1305,6 +1308,137 @@ void CAudioMixerBoard::SetAllFaderLevelsToNewClientLevel()
...
+        return;
+
+    // compute target level for each group
+    // (prevent clipping when each group contributes at maximum level)
+    float targetLevelPerGroup = -20.0f * log10 ( cntActiveGroups );
+
+    // compute target levels for the channels of each group individually
+    float vecTargetChannelLevel[MAX_NUM_FADER_GROUPS + 1];
```
pljones, line 1367: I'd have to check elsewhere but I think the `[` and `]` need spacing.

src/audiomixerboard.cpp
JohannesBrx: Ok. I'll place the first bracket on the next line (after `if` line) since this seems to be the style in the majority of cases.

By the way: Shall I _resolve conversation_ or will you do so? (also applies for the other reviews)

JohannesBrx:
> @JohannesBrx So, the test went really well! We were some 16 singers and I tried the auto-adjustment multiple times within a (homophonic) piece (so, not even deliberate Aaahs or something). I'm not saying that I can confirm that it works flawlessly (did not have the time to check during singing). I'm merely saying it works and it is a huge improvement over the status quo. I'm pretty sure this will be a welcome addition especially for larger groups.

@hoffie Thanks for your feedback :-) In the meanwhile I have added support for groups. I also think we could get rid of the -30dB constant (which was the _target level_ for all channels), making an additional configuration item unnecessary. If you once again have a choir rehearsal this week I would appreciate you feedback for the new version. The current version is here:
https://github.com/jamulussoftware/jamulus/actions/runs/601973445

I would like to rebase this branch on the current master and I think there would be no conflicts. However, I am not sure whether this might cause problems with the current reviews. Any opinions?

src/audiomixerboard.cpp
pljones: Oops, yes - next line...  (yes... really... plenty of white space... it's cheap... :) )

src/global.h
pljones: OK... I _think_ I understand.

src/audiomixerboard.cpp
pljones: Have you pushed this change?  (For some reason Github hasn't flagged the file change since I viewed it.)

pljones:
Rebasing is for the good - Github _should_ only flag changed files as needing review again...

src/audiomixerboard.cpp
JohannesBrx: Now I have committed it after using rebasing from the current upstream master.

JohannesBrx:
> Rebasing is for the good - Github _should_ only flag changed files as needing review again...

Ok, now I have rebased it from the current upstream master. I needed to *force-push* it afterwards. If this causes any trouble, please let me know!

pljones (approved):
Just to note that _all_ of the work is being done in the UI - the client code that's doing the time-critical stuff isn't touched.

On the up-side, that means, should the two be split into separate threads, the time-critical thread hasn't the extra work.

On the down-side, it means the calculation code isn't available to a "headless" build.

So I'm approving -- but maybe we could think about moving the "hard work" to `client.cpp`, so it's accessible without the GUI being active.  (The most important part being "compute exponential moving average" -- that would then need to be stored and made accessible.)  The code in `AutoAdjustAllFaderLevels` perhaps could be split - `client.cpp` "knowing how" to compute the values and `audiomixerboard.cpp` still calling `SetFaderLevel` to make it happen.

hoffie:
> If you once again have a choir rehearsal this week I would appreciate you feedback for the new version.

Used 7276381d today and it still worked fine. I'm not using grouping though. My first use case (until this is available for the masses ;)) is a special Jamulus client which I use for streaming/recording. Until now I had to blindly set (or import) fader levels. Now I can use this feature and it works great. :)
This is just another one-shot observation and not systematic testing. When looking at your graphs, you seem to have done the latter perfectly anyway! :)

Still can't say if further tuning is necessary, but this is so much better than not even having this feature at all, that I'd really want to have it in one of the next versions. Will have a look at the code in the coming/days weeks, too.

Thanks for working on this! :)

JohannesBrx:
After some tests with choirs I have refined the adjustment strategy as follows:

Before the change, the channels were adjusted in such a way that the weakest channel contributes as much as the strongest channel. (At least for level differences smaller or equal than 35dB.) However, this caused the adjustment to be quite unstable because the weak channel might fluctuate a lot. In the new version, the median channel level is taken instead the weakest channel.

In addition to that, the algorithm does not perform adjustments for channels without group assignment when there are *any* groups. Doing so, e.g., the instructor (without group) will not be considered for adjustment. When there are no groups defined, adjustment is applied to all channels.

gene96817:
Using the mean level as the reference is good.
It could be a valid choice that any channel outside of one std dev needs to be resolved manually.

Would instructor = director?

From your discussion, I can see the implication that the person without a group assignment is the instructor. However, what if most attendees have group assignments and a "few" do not? In this case, for my choruses, these few could be visitors. We could choose to say (1) note, no assignment means no adjustments - good or bad, (2) no assignment = a default misc group, (3) no assignment = not a regular participant and treated as a listen-only visitor.

src/audiomixerboard.cpp
```diff
@@ -1321,15 +1321,15 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
...
-    float vecMinLevel[MAX_NUM_FADER_GROUPS + 1];
     float vecMaxLevel[MAX_NUM_FADER_GROUPS + 1];
     int   vecChannelsPerGroup[MAX_NUM_FADER_GROUPS + 1];
     for ( int i = 0; i < MAX_NUM_FADER_GROUPS + 1; ++i )
     {
-        vecMinLevel[i] = UPPER_BOUND_SIG_METER;
         vecMaxLevel[i] = LOW_BOUND_SIG_METER;
         vecChannelsPerGroup[i] = 0;
     }
+    CVector<CVector<float>> levels;
```
pljones, line 1331: Can this be declared once in class scope, rather than each time `AutoAdjustAllFaderLevels` is called, as it does the `resize`?

In fact, all the variables that set up arrays might be better at class scope to get created just the once.  Would that affect things adversely?

src/audiomixerboard.cpp
```diff
@@ -1349,29 +1349,37 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
...
             {
-                vecMinLevel[group] = fmin ( vecMinLevel[group], leveldB );
                 vecMaxLevel[group] = fmax ( vecMaxLevel[group], leveldB );
+                levels[group].Add ( leveldB );
             }
             ++vecChannelsPerGroup[group];
         }
     }
 
+    // sort levels for later median computation
```
pljones, line 1359: Any `sort` call makes me worry.  There really isn't much data here, however, so I'm going to presume it can never hurt...

src/audiomixerboard.cpp
```diff
@@ -1349,29 +1349,37 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
```
pljones, line 1349: On the previous line... that `4` should really be `MAX_NUM_CHANNELS` shouldn't it?  (Plus correct the `{}` style...)

src/audiomixerboard.cpp
```diff
@@ -1386,15 +1394,21 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
             targetLevelPerGroup - 20.0f * log10 ( vecChannelsPerGroup[i] ) :
             0.0f;
 
+        // get median level
+        int cntChannels = levels[i].Size();
+        if ( cntChannels == 0 )
+            continue;
```
pljones, line 1400: Please use the proper `{}` - else you'll have more work later :).

src/audiomixerboard.cpp
```diff
@@ -1423,7 +1437,17 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
...
-                group = 4;
+            {
+                if ( cntActiveGroups > 0 )
+                {
+                    // do not adjust the channels without group in group mode
+                    continue;
+                }
+                else
+                {
+                    group = 4;
```
pljones, line 1448: Again, should be `MAX_NUM_FADER_GROUPS`?

JohannesBrx:
> Using the mean level as the reference is good.
> It could be a valid choice that any channel outside of one std dev needs to be resolved manually.

How would we indicate the user calling the adjustment which channels have been adjusted and which ones not? Currently the faders of channels which cannot be adjusted will be at their maximum range. (It is the best we can do.)

> Would instructor = director?
> 
> From your discussion, I can see the implication that the person without a group assignment is the instructor. However, what if most attendees have group assignments and a "few" do not? In this case, for my choruses, these few could be visitors. We could choose to say (1) note, no assignment means no adjustments - good or bad, (2) no assignment = a default misc group, (3) no assignment = not a regular participant and treated as a listen-only visitor.

I would choose (1). Visitors, instructors, streaming clients etc. could have no group assignment to skip adjustments.

src/audiomixerboard.cpp
JohannesBrx: This has also been fixed with commit 66aa3fc...

src/audiomixerboard.cpp
JohannesBrx: I know... However, I do not see an alternative for the median, which I believe is the best choice in this case. And median requires sorting.

The worst case would be sorting an array with `MAX_NUM_CHANNELS` elements. Since this is a function which is not called periodically but by the user (CTRL+F or CMD+F), I think this should be acceptable.

By the way, `CAudioMixerBoard::ChangeFaderOrder` also uses sorting ;-)

src/audiomixerboard.cpp
JohannesBrx: > Can this be declared once in class scope, rather than each time AutoAdjustAllFaderLevels is called, as it does the resize? 

Because of potential performance issues? Since this function is only called when the user *explicitly* calls it via the menu or keystroke, I would think that this should not be a problem. (In contrast to the audio processing functions which are called periodically and with a high frequency.)

> In fact, all the variables that set up arrays might be better at class scope to get created just the once. Would that affect things adversely?

Most of the other (fixed-sized) arrays are placed on the stack (not the heap) and therefore should not hurt performance.

> (Iterating over to clean each time is probably quicker than the memory allocation. I note there's no actual initialisation currently - is this method guaranteed to have zeroed values?)

Since only the outer `CVector` is resized, the STL should call the constructor of the inner `CVector` element, which should initialize with an empty list. New elements are added dynamically via `CVector::Add`.

ann0see:
I used this in the last two rehearsals and it works good. 

The only problem might be the missing accuracy but I'd say that's an improvement for a new PR. Better to have this feature than not having it at all.

ann0see:
Would this also be ready to be merged?

JohannesBrx:
> Would this also be ready to be merged?

I think so. However, a second review is still missing.

ann0see:
Ok. I'll look into it tomorrow. BTW: thank you for all your input and PRs!

src/audiomixerboard.cpp
```diff
@@ -1310,6 +1313,167 @@ void CAudioMixerBoard::SetAllFaderLevelsToNewClientLevel()
...
+                vecTargetChannelLevel[i] + levelOffset );
+        }
+    }
+
+    // take minimum fader value into account
+    // very weak channels would actually require strong channels to be
+    // attenuated to a large amount; however, the attenuation is limited by
+    // the faders
+    if ( minFader < -AUD_MIX_FADER_RANGE_DB )
+        levelOffset += -AUD_MIX_FADER_RANGE_DB - minFader;
```
ann0see, line 1425: Not sure about the style here. Would we usually use brackets here?

JohannesBrx: Ok, I will fix that with the next commit.

ann0see (approved):
From my tests and a quick look over the code, I'll approve (but not merge) it yet. Softins or someone else should have a look at the code too before a merge.

softins:
> From my tests and a quick look over the code, I'll approve (but not merge) it yet. Softins or someone else should have a look at the code too before a merge.

I will look through the code and review asap.

softins (approved):
I haven't studied the algorithm in detail, but the code looks good, and I have built and tested it on my Pi.

hoffie (approved):
I finally took the time to scan the code and it looks good to me. I've already used it multiple times im practice and it does provide a huge benefit. A huge thanks for doing this!

@JohannesBrx Do you want to add yourself as a contributor in `src/util.cpp` as part of this PR?

Not merging yet for this reason, other than that, I think this can go in. :)

JohannesBrx:
> I finally took the time to scan the code and it looks good to me. I've already used it multiple times im practice and it does provide a huge benefit. A huge thanks for doing this!
> 
> @JohannesBrx Do you want to add yourself as a contributor in `src/util.cpp` as part of this PR?

I have added myself to the list. I am glad to be able to give something back in this great project.

ann0see:
I think you should also edit the changelog and give a shot description there? Afterwards this can be merged.

PR #1074, opened 2021-02-20
Don't hide mute indicator in mono mode

npostavs:
I'm not sure why the indicator would be hidden in mono mode, it's surely equally useful regardless.

Screenshots showing the problem:

mute-mono
mute-stereo

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -285,7 +285,6 @@ bool CChannelFader::GetDisplayChannelLevel()
 
 void CChannelFader::SetDisplayPans ( const bool eNDP )
 {
-    pInfoLabel->setHidden ( !eNDP );
     pPanLabel->setHidden  ( !eNDP );
     pPan->setHidden       ( !eNDP );
 }


gilgongo:
Yes, this does seem like a bug. I can only assume the display of the icon was somehow caught up in the display of the pan control perhaps. (BTW I'm not an engineer so I'll wait for others to review and ping them if they don't see it in the next few days).

pljones:
I think it we take the opportunity to move the "Has you muted" icon down to above the channel name (maybe next to the "Mute" checkbox), it would allow this to be fixed and the other UI request to be addressed.  I'm pretty sure you're right, @gilgongo - it'll be the Pan thing.

hoffie:
Hrm, @corrados deliberately added the hiding in [a7770e27aecf3f2827b53d27d0c63addbb90d8d5](https://github.com/jamulussoftware/jamulus/commit/a7770e27aecf3f2827b53d27d0c63addbb90d8d5?w=1), so it'd be worth double-checking why that was done? (Maybe this is what you mean by the "Pan thing", given that this commit came right after Pan adjustments and I just lack context)

pljones:
OK 4b8c7de1 through to a7770e27 then seem to be the commits...  Basically, it appears Volker put the `Speaker with cancellation stroke` into the same space as the (new) pan knob to "save space".  So all we'd be doing is moving it somewhere else to "save space" but differently - because the pan knob isn't always there (i.e. putting the icon into a hidden component means it's hidden - but using one that's never hidden means it "should just work" (tm)).

gene96817:
I get a bit itchy every time I see "hide" in the UI. I personally dislike the UI changing according to the mode. It is distracting and my brain isn't big enough to keep up with the changes. It is a lower load on my brain for the UI to have a constant layout. 

Greying out a feature is much better than hiding. It's there but it (the function) isn't there.

gilgongo:
> Greying out a feature is much better than hiding. It's there but it (the function) isn't there.

Yes, although there may be an issue in this case to do with [mode error](https://en.wikipedia.org/wiki/Mode_(user_interface)#Mode_errors) perhaps. If we showed, for example, an "OK" icon to indicate sound wasn't being muted, then the user has to work a bit harder to work out the mode ("Are they all muted or all OK?"). That and the fact that we don't have much space for the indicator to be very differentiated. So for now at least I'd say hide it even though I agree that's not best practice.

pljones:
> It's there but it (the function) isn't there.

For "Has me muted", there's no "function" to access, as such.  It's informational.  Leaving it greyed out would be like leaving all hover text visible but greyed out when not hovered...

I'd rather it didn't need fixing (it's a bit of a nasty oversight) but it does.  And I'd like it to be right this time -- AND to take into account what we'll do for any "Has themselves muted" info indicator.

ann0see:
Hmm any updates here? What's missing?

hoffie (approved):
After checking each code path which touches `pInfoLabel`, this looks good to me. I've verified that it fixes the bug. The icon is left aligned.

Thanks!

pljones (approved):

pljones:
Can we have screenshots for Fancy, Normal and Compact skins, please.  I've approved anyway, though, but I'd like to see the screenshots, too :).

npostavs:
> Can we have screenshots for Fancy, Normal and Compact skins, please. I've approved anyway, though, but I'd like to see the screenshots, too :).

Sure.
mono-mute-all-skins
stereo-mute-all-skins

gilgongo:
Looks good to me (well, in so far as the current UI looks good at least...)

dcorson-ticino-com:
Does this also fix the problem that the font used for the symbol is not always the same size as seen in https://github.com/jamulussoftware/jamulus/issues/1312  ?

gilgongo:
No that's a separate issue I believe. I'll merge this now though as I think all is in order - thanks @npostavs !

Issue #834, opened 2021-01-13
Turn of screensaver if connected to a server (macOS and LXQt)


ann0see:
If you run Jamulus on Debian 10 under LXQt the screensaver automatically starts if Jamulus is connected to a server. Same problem on macOS (probably BigSur). It should be possible to inform the system that the application is active see the windows docs on this: https://docs.microsoft.com/en-en/windows/win32/api/winbase/nf-winbase-setthreadexecutionstate?redirectedfrom=MSDN

It probably depends on the OS how to manage this.

chrisrimple:
I've experienced a similar issue on MacOS: the screen can go dark (power saver) while Jamulus is running.

Apple > System Preferences > Desktop & Screen Saver > Screen Saver > After After n Minutes
- MacOS isn't detecting "activity" for Jamulus, so screensaver is turned on after n minutes

Apple > System Preferences > Energy Saver > Power Adapter > Turn Display Off After n Minutes
- MacOS isn't detecting "activity" for Jamulus, so screen is turned off after n minutes

DavidSavinkoff:
Does activating the screen saver actually interrupt the audio in Jamulus? I think that if Jamulus can prevent the screen saver from activating, it should be an option that only affects that instance of Jamulus... The Reason is: Suppose the screen saver option is automatically activated by Jamulus, because it was saved, then you decide to decide to run 'only the Jamulus server. But the server grabs the 'deactivate screen saver setting' leaving you wondering why the screen saver stopped working.

ann0see:
I think that the screensaver is already not active if you run the server (#260). I haven't tested this yet.

DavidSavinkoff:
Maybe this screen saver issue needs to thoroughly investigated. Especially if Jamulus needs to have the screen saver disabled for performance reasons. I'm starting to think that Jamulus should leave the screen saver alone... Some other application should deal with it.

ann0see:
I think it might depend on which state Jamulus is in. So if you're connected to a server the screensaver should not start. But if you're not connected it should still be enabled.

chrisrimple:
From my own experience on MacOS, the Energy Saver setting is in effect when the Jamulus client is running.  On multiple occasions, I have been playing with other musicians on a Jamulus server and had my screen go dark.  A click or keypress awakes it, but also requires a login (depends on security settings).  If the client is running, it would be nice if it reported to the OS that it's "busy".

pljones:
Android
[...]

PR #1075, opened 2021-02-20
No display sleep for Jamulus on Mac

jerogee:
- This prevents the display from sleeping / screen saving on Mac while Jamulus is running. 
- Should work on iOS 7.0+ and  macOS 10.9+ (Mavericks and later). Tested on 10.14 (Mojave) 
- Addresses chrisrimple's situation described in #834, but does not close the issue completely as only for macOS with LXQt still tbd.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -42,7 +42,7 @@
 
 - bug fix: fix crash if settings are changed in ASIO4ALL during a connection (contained in #796). Reverts #727 for Windows
 
-
+- Avoid screen from sleeping or starting screen saver for Mac (#834)
 
 
 
--- a/mac/activity.mm
+++ b/mac/activity.mm
@@ -40,7 +40,7 @@ CActivity::~CActivity()
 
 void CActivity::BeginActivity()
 {
-    NSActivityOptions options = NSActivityBackground | NSActivityIdleSystemSleepDisabled | NSActivityLatencyCritical;
+    NSActivityOptions options = NSActivityBackground | NSActivityIdleDisplaySleepDisabled | NSActivityIdleSystemSleepDisabled | NSActivityLatencyCritical;
     
     pActivity->activityId = [[NSProcessInfo processInfo] beginActivityWithOptions: options reason:@"Jamulus provides low latency audio processing and should not be inturrupted by system throttling."];
 }


hoffie (approved):
Thanks for your PR! :)

This sounds like we should have it. The option is also [widely used](https://github.com/search?q=NSActivityIdleDisplaySleepDisabled) and officially [documented](https://developer.apple.com/documentation/foundation/nsactivityoptions/nsactivityidledisplaysleepdisabled), so this seems correct to me.

I **cannot** test this my self though.

chrisrimple:
If someone can compile a test MacOS build for me, I can test the change.

hoffie:
> If someone can compile a test MacOS build for me, I can test the change.

That would be cool! :)

Thanks to @nefarius2001 we have autobuilds on all PRs now, although they are bit hidden and I'm not sure if they are publicly accessible: PR > Checks > Artifacts (right corner)

This would be the direct link for MacOS:
https://github.com/jamulussoftware/jamulus/suites/2085329317/artifacts/42267420

If this link doesn't work for you, I could upload that file somewhere else.

nefarius2001:
@hoffie actions and artifacts are public and can be found by everyone in the „actions“ tab
https://github.com/jamulussoftware/jamulus/actions

hoffie:
> @hoffie actions and artifacts are public and can be found by everyone in the „actions“ tab
> https://github.com/jamulussoftware/jamulus/actions

Yeah, the listing works anonymously, but when clicking on one of the artifacts it leads to a 404, it seems. I guess it works for all logged-in users though. Here is some discussion around this: https://github.com/actions/upload-artifact/issues/51

```
$ curl "https://github.com/jamulussoftware/jamulus/suites/2085329317/artifacts/42267420" -I
HTTP/2 404 
```

If you think that something should (or rather: can) be done about it, I guess we should open a new issue for that?

jerogee:
These autobuilds are fantastic. Great work on that @nefarius2001.

ann0see (approved):
Confirmed to work on macOS Big Sur. It also to block darkening if you're not connected. Any thoughts on this?

--- 

I think this can be included in the next release too? If that's the case, a short comment should be added in the ChangeLog file

hoffie:
> I think this can be included in the next release too? If that's the case, a short comment should be added in the ChangeLog file

Will not interfere with translations, seems non-critical and we might want another beta anyway, so, yes from my point of view.

chrisrimple:
Sorry about the slow follow up, but confirmed working on MacOS 10.14.6 Mojave.  I left the computer unattended for more than an hour with Jamulus connected to a server and no other activity, and the screen never shut off.

Issue #1081, opened 2021-02-21
Security: Investigate buffer boundary warning in src/signalhandler.cpp


hoffie:
**Describe the bug**
Codacy emits a warning for [src/signalhandler.cpp line 85](https://github.com/jamulussoftware/jamulus/blob/80e1d2325a43aae00667849a41e6cbbdc4b4f163/src/signalhandler.cpp#L85).

From @ann0see in https://github.com/jamulussoftware/jamulus/issues/314#issuecomment-761890259

Note: I don't see a buffer around that line and was wondering whether the reference could be outdated or wrong. However, this file has not seen any recent updates, as far as I can see.

**To Reproduce**
Run a Codacy scan

**Expected behavior**

No warning.

**Screenshots**
```
"Check buffer boundaries if used in a loop including recursive loops (CWE-120, CWE-20)."

in src/signalhandler.cpp line 85
```

**Version of Jamulus**
Assuming that @ann0see had master checked, this would imply that the version was 80e1d2325a43aae00667849a41e6cbbdc4b4f163 or before.

@softins Assigning to you as you sounded like you were to look into this? :)

softins:
OK, it looks to be referring to the `::read()` in this function:
```
void CSignalHandler::OnSocketNotify( int socket )
{
    int sigNum;
    if ( ::read ( socket, &sigNum, sizeof ( int ) ) == sizeof ( int ) )
    {
        emitSignal ( sigNum );
    }
}
```
It is using the `int sigNum` as a buffer, and it should be ok because the read is limited to `sizeof(int)` bytes in size.

Maybe Codacy got confused because the read was not into a `char` buffer?

@hoffie, where is this Codacy? Is this something you ran on your system, or that Github ran in the CI?

pljones:
I _think_ it's a false positive.

hoffie:
> @hoffie, where is this Codacy? Is this something you ran on your system, or that Github ran in the CI?

Not sure, @ann0see did this. However, I just found that Codacy seems free for Open Source projects and I'm just trying it out in my personal Github account on my Jamulus fork. Will report back what it finds. We could discuss whether it makes sense to integrate it into the jamulussoftware org. I'll follow up on this later.

hoffie:
Codacy now lists 5 security-related things to check. The first is the one which @ann0see mentioned. I'm providing a screenshot of the detailed view below, but besides confirming that we've looked at the right file, it does not any useful details:

[...]

PR #1084, opened 2021-02-21
Codacy/Flawfinder: Document a false positive finding

hoffie:
Add an explaining comment and a special hint at Codacy/Flawfinder to ignore the (non-)problematic line.

Special comment effectiveness has been validated here: https://github.com/hoffie/jamulus/pull/3

Related: #1081

### DIFF
--- a/src/signalhandler.cpp
+++ b/src/signalhandler.cpp
@@ -82,7 +82,10 @@ bool CSignalHandler::emitSignal ( int sigNum )
 void CSignalHandler::OnSocketNotify( int socket )
 {
     int sigNum;
-    if ( ::read ( socket, &sigNum, sizeof ( int ) ) == sizeof ( int ) )
+    // The following read() triggers a Codacy/flawfinder finding:
+    // "Check buffer boundaries if used in a loop including recursive loops (CWE-120, CWE-20)"
+    // Investigation has shown that this is a false positive as the proper buffer size is enforced.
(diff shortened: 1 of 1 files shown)


softins (approved):

atsampson:
In comments like this, it's a good idea to say **which** warning it's safe to ignore, just in case the tools detect a different problem here in the future.

hoffie:
> In comments like this, it's a good idea to say **which** warning it's safe to ignore, just in case the tools detect a different problem here in the future.

I though my comment was already rather elaborate and `git blame` could be used to end up in this PR/the related issue, but I guess you are right that it will be better to make this explicitly. I have pushed an updated comment. Thanks for this suggestion! :)

pljones:
I don't like linters/analysers that don't let you mark clearly _to the linter_ that you've accepted the warning and don't want to hear about it again.  (They should be explicit markers, as close to the code as possible: `..., /* Codacity(#1432, accept) */&sigNum, ...` kind of thing.)  Few and far between, unfortunately.

hoffie:
> I don't like linters/analysers that don't let you mark clearly _to the linter_ that you've accepted the warning and don't want to hear about it again (in addition to human-readable comments). (They should be explicit markers, as close to the code as possible: `..., /* Codacity(#1432, accept) */&sigNum, ...` kind of thing.) Few and far between, unfortunately.

I fully agree. I don't think it's possible with Flawfinder though. I even checked [its code](https://github.com/david-a-wheeler/flawfinder/blob/master/flawfinder#L1464).

pljones (approved):

PR #1093, opened 2021-02-23
Simpler Recording Indication

dcorson-ticino-com:
In class audiomixerboard:
- added EGUIDesign eDesign to save GUI design
- in UpdateTitle added setStyleSheet
- in HideAll (in lieu of Disconnect) added setStyleSheet

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -837,6 +837,7 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent ) :
     iNumMixerPanelRows   ( 1 ),
     strServerName        ( "" ),
     eRecorderState       ( RS_UNDEFINED ),
+    eDesign              ( GD_ORIGINAL ),
     eChSortType          ( ST_NO_SORT )
 {
     // add group box and hboxlayout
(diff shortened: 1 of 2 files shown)


src/audiomixerboard.cpp
```diff
@@ -995,6 +998,9 @@ void CAudioMixerBoard::HideAll()
     // use original order of channel (by server ID)
     ChangeFaderOrder ( ST_NO_SORT );
 
+    // Reset recording indication styleSheet
```
pljones, line 1001: I was thinking of a class variable to hold the current stylesheet based on the GUI design, set in a switch here.

Then, in UpdateTitle, set the stylesheet from the class variable if recording is active, else set the stylesheet to `""`.

dcorson-ticino-com: There is no skin dependency when setting here in CAudioMixerBoard. When the styleSheet is reset with "" it reverts to the styleSheet of the underlying class which has already been set according the the skin.

pljones: OK - it was just the previous version seemed to have different strings per GUI design.  If it's not required, this is fine.

The only other thing might be to put the value at 1109 into a `const` or `#define` in the header, rather than in line.

pljones (approved):

softins (approved):

PR #1097, opened 2021-02-23
V370 german

rolamos:
PR for German Translation

PR #1099, opened 2021-02-23
Updated Polish translation for 3.7.0 release

SeeLook:


ann0see (approved):
Thanks! Unfortunately I can't comment much here. Can you test if the translation is correct by checking the artifact?

SeeLook:
Checked and even ironed out some kinks.

src/res/translation/translation_pl_PL.ts
```diff
@@ -952,7 +952,11 @@
     <message>
         <location filename="../../clientdlgbase.ui" line="584"/>
         <source>Set up your audio, connect to a server and start jamming!</source>
+<<<<<<< HEAD
```
pljones, line 955: This is a failed rebase which needs resolving.

SeeLook: Sorry for trouble, next shoot should be correct.

src/res/translation/translation_pl_PL.ts
```diff
@@ -118,7 +118,7 @@
     <message>
         <location filename="../../aboutdlgbase.ui" line="72"/>
         <source>Copyright (C) 2005-2021 The Jamulus Development Team</source>
-        <translation type="unfinished"></translation>
+        <translation>Copyright (C) 2005-2021 Zespół programistów Jamulus-a</translation>
```
hoffie, line 121: We should wait with the merge depending on what we settle on in https://github.com/jamulussoftware/jamulus/pull/1097#discussion_r581265793

src/res/translation/translation_pl_PL.ts
```diff
@@ -2973,7 +2973,7 @@
     <message>
         <location filename="../../global.h" line="237"/>
         <source>For more information use the &quot;What&apos;s This&quot; help (help menu, right mouse button or Shift+F1)</source>
-        <translation>Aby uzyskać więcej informacji użyj &quot;Co to&quot; (Pomoc, prawy przycisk myszy lub Shift+F1)</translation>
+        <translation>Aby uzyskać więcej informacji użyj „Co to jest” (z menu Pomoc lub prawy przycisk myszy lub Shift+F1)</translation>
```
hoffie, line 2976: I assume you have verified that these quotes work properly in the UI?

hoffie (approved):
Thanks for your work! Looks good to me. I don't speak Polish but I have at least checked punctuation, spacing and technical words.

I wonder why there is any empty `src/res/translation/translation_pl_PL.qm` file. Not sure if it causes any problems, I guess it will be overwritten later anyway.

src/res/translation/translation_pl_PL.ts
SeeLook: Yes. I have been using those for a long in another Qt project I'm translating and it appears correctly under any OS-es with Polish localization.

hoffie:
Just checked: There don't seem to be any untranslated things left, which is good! :)

SeeLook:
That was the plan :)

src/res/translation/translation_pl_PL.ts
pljones: Thanks!

src/res/translation/translation_pl_PL.ts
ann0see: I think we can just leave it like that. If the qm file isn't something we need to worry about I'd say this is ready to be merged then.

pljones:
> I wonder why there is any empty `src/res/translation/translation_pl_PL.qm` file. Not sure if it causes any problems, I guess it will be overwritten later anyway.

I don't remember the process -- @softins can you comment?

softins:
Yes, no need to worry about the .qm files. Once all the .ts files are in and merged, one of us (I'm happy to) can just run `lrelease Jamulus.pro`, which will generate all the .qm files and then check them in. I think really they ought to be generated by the makefile, but that doesn't happen now. I tried deleting the qms and runnng make, and it failed, complaining it couldn't find them.

PR #1100, opened 2021-02-23
Build deb files on Ubuntu 18.04

ann0see:
If we don't do this, we can't run the .deb files on debian buster. ~~(still to be tested)~~

cc @nefarius2001

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -102,7 +102,7 @@ jobs:
 
            - config_name:           Linux (artifacts+codeQL)
              target_os:             linux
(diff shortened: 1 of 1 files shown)


ann0see:
Ok. Can confirm that the headless build now works.

pljones (approved):

hoffie (approved):
Tested:
- Non-Headless build can be installed on Debian Buster and starts (`-s --nogui`, forgot to test the headless build explicitly)
- Both headless and non-headless can be installed on Debian Sid and start (`-s --nogui`)
- Both headless and non-headless can be installed on Ubuntu 20.04 and start (`-s --nogui`)

PR #1101, opened 2021-02-23
Hide connection label in connect method

### LINKED ISSUE #1076: UI: "Set up your audio..." messages sometimes remains
**Describe the bug**
The "Set up your audio..." message sometimes remains when connected to a server.

**To Reproduce**
Couldn't pin it down.  It could have been going straight from one server to another... not sure.  I'll try again tomorrow.

**Expected behavior**
The message should only be displayed when _not_ connected.

**Screenshots**
Unfortunately, wasn't able to save them whilst jamming.

**Operating system**
image
**Version of Jamulus**
image

ann0see:
Fixes #1076 (hopefully).

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -721,10 +721,6 @@ void CClientDlg::OnConnectDlgAccepted()
         // get the address from the connect dialog
         QString strSelectedAddress = ConnectDlg.GetSelectedAddress();
 
-        // hide label connect to server
-        lblConnectToServer->hide();
-        lbrInputLevelL->setEnabled ( true );
-        lbrInputLevelR->setEnabled ( true );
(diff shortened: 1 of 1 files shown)


pljones (approved):
Looks like it should work -- if that's the only place the button gets changed to Disconnect, of course :).

hoffie (approved):
> Looks like it should work -- if that's the only place the button gets changed to Disconnect, of course :).

I found three places which call `butConnect->setText()`:
- In the constructor (`lblConnectToServer`'s default state should be OK already)
- In `Connect()` which hides `lblConnectToServer` properly
- In `Disconnect()` which shows `lblConnectToServer`

So, looks sane to me. :)

ann0see:
Ok. I hope everything is ok then.

PR #1102, opened 2021-02-23
Remove "using namespace std;"

atsampson:
Importing everything from the `std` namespace is likely to cause problems, since it contains lots of common names and more get added with new C++ versions.

The Jamulus code usually qualifies `std::` names explicitly where needed; this also fixes the few places where it doesn't. (~~Only tested on Linux so far, so wait for the CI to check out the other platforms...~~ CI seems happy on all platforms.)

Related to #1098.


ann0see:
Thanks! Does this fix the issue/discussion you mentioned?

atsampson:
I don't know - probably best to ask whoever reported that problem originally.

ann0see:
I think it doesn’t since they use Qt6 and there’s another issue. Let’s wait and see what others say here.

pljones (approved):

softins (approved):

PR #1104, opened 2021-02-24
Update Slovak translation

jose1711:


hoffie (approved):
Thanks for your work! I don't know Slovak, but I have checked spacing, punctuation, etc. and it seems fine.

One question: I notice that there are still some untranslated strings. I guess it will take some time for those to be translated as well.

Are you planning to translate them as well?

I'm approving the PR for now anyway as getting this in is already a great thing to have! :)

Example:
```
    <message>
        <location filename="../../soundbase.cpp" line="197"/>
        <source> could not be started because of audio interface issues.</source>
        <translation type="unfinished"></translation>
    </message>
```

@jamulussoftware/maindevelopers I guess we should check all such PRs for stuff which still needs translating. Hadn't done that in the previous PRs, but will do now. `grep unfinished -5 src/res/translation/translation_$TRANSLTION*.ts` helps.

- [x] Listed in the Contributors UI window
- [x] Changes look consistent
- [ ] No untranslated strings

src/res/translation/translation_sk_SK.ts
```diff
@@ -91,7 +91,7 @@
     <message>
         <location filename="../../util.cpp" line="510"/>
         <source>Slovak</source>
-        <translation type="unfinished"></translation>
+        <translation>slovenčina</translation>
```
softins, line 94: Should this be capitalised? Slovenčina

src/res/translation/translation_sk_SK.ts
```diff
@@ -2790,12 +2790,12 @@
...
         <location filename="../../../linux/sound.cpp" line="64"/>
         <source>Make sure to set the Frames/Period to a low value like </source>
-        <translation type="unfinished"></translation>
+        <translation>Ubezpečte sa, že nastavenie Snímky/Perióda je nastavené na nízku hodnotu ako </translation>
     </message>
     <message>
         <location filename="../../../linux/sound.cpp" line="67"/>
         <source> to achieve a low delay.</source>
-        <translation type="unfinished"></translation>
+        <translation> , aby ste dosiahli nízke oneskorenie.</translation>
```
softins, line 2798: I suspect the space before the comma is not needed.

src/res/translation/translation_sk_SK.ts
jose1711: Depends, but in general language names in Slovak are not capitalized (see http://www.slovencina.vselico.com/velkepismena.html [Slovak])

jose1711:
> Thanks for your work! I don't know Slovak, but I have checked spacing, punctuation, etc. and it seems fine.
> 
> One question: I notice that there are still some untranslated strings. I guess it will take some time for those to be translated as well.
> 
> Are you planning to translate them as well?

Thank you. Yes, eventually. This commit was mainly stuff I've been working on and somehow forgot about it. I can't really predict when I have time to resume so pushing what I currently have.

pljones (approved):

PR #1105, opened 2021-02-24
Fix typo which dropped machine-specific opus sources

npostavs:
#85 updated the OPUS version and added these, but they're not being compiled. Actually I think fixing the `Jamulus.pro` typo isn't enough because the code is ifdef'd out. I'm not sure how to get the correct set of defines short of running `./configure` or `cmake`.

@doloopuntil since you wrote #85, any thoughts?

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -668,18 +668,18 @@ SOURCES_OPUS_X86 = libs/opus/celt/x86/celt_lpc_sse4_1.c \
 android {
     contains(ANDROID_ARCHITECTURE, arm) | contains(ANDROID_ARCHITECTURE, arm64) {
         HEADERS_OPUS += $$HEADERS_OPUS_ARM
-        SOURCE_OPUS += $$SOURCES_OPUS_ARM
+        SOURCES_OPUS += $$SOURCES_OPUS_ARM
     } else:contains(ANDROID_ARCHITECTURE, x86) | contains(ANDROID_ARCHITECTURE, x86_64) {
         HEADERS_OPUS += $$HEADERS_OPUS_X86
-        SOURCE_OPUS += $$SOURCES_OPUS_X86
+        SOURCES_OPUS += $$SOURCES_OPUS_X86
     }
 } else:win32 | unix | macx {
     contains(QT_ARCH, arm) | contains(QT_ARCH, arm64) {
         HEADERS_OPUS += $$HEADERS_OPUS_ARM
-        SOURCE_OPUS += $$SOURCES_OPUS_ARM
+        SOURCES_OPUS += $$SOURCES_OPUS_ARM
     } else:contains(QT_ARCH, x86) | contains(QT_ARCH, x86_64) {
         HEADERS_OPUS += $$HEADERS_OPUS_X86
-        SOURCE_OPUS += $$SOURCES_OPUS_X86
+        SOURCES_OPUS += $$SOURCES_OPUS_X86
     }
     win32 {
         HEADERS_OPUS += libs/opus/win32/config.h
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -1082,6 +1082,9 @@ contains(CONFIG, "opus_shared_lib") {
 
     contains(QT_ARCH, x86) | contains(QT_ARCH, x86_64) {
         msvc {
+            # According to opus/win32/config.h, "no special compiler
+            # flags necessary" when using msvc.  It always supports
+            # SSE intrinsics, but doesn't auto-vectorize.
             SOURCES += $$SOURCES_OPUS_ARCH
         } else {
             # Arch-specific files need special compiler flags, but we
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -655,7 +655,16 @@ SOURCES_OPUS = libs/opus/celt/bands.c \
 
 SOURCES_OPUS_ARM = libs/opus/celt/arm/armcpu.c \
     libs/opus/celt/arm/arm_celt_map.c \
-    libs/opus/silk/arm/arm_silk_map.c
+    libs/opus/silk/arm/arm_silk_map.c \
+    libs/opus/silk/arm/arm_silk_map.c \
+    libs/opus/silk/arm/biquad_alt_neon_intr.c \
+    libs/opus/silk/arm/LPC_inv_pred_gain_neon_intr.c \
+    libs/opus/silk/arm/NSQ_del_dec_neon_intr.c \
+    libs/opus/silk/arm/NSQ_neon.c \
+    libs/opus/celt/arm/celt_neon_intr.c \
+    libs/opus/celt/arm/pitch_neon_intr.c \
+    libs/opus/celt/arm/celt_fft_ne10.c \
+    libs/opus/celt/arm/celt_mdct_ne10.c
 
 SOURCES_OPUS_X86_SSE = libs/opus/celt/x86/x86cpu.c \
     libs/opus/celt/x86/x86_celt_map.c \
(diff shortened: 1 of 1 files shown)


pljones:
Given all builds have been compiling the OPUS source successfully for quite some time, can you explain what this is fixing?

npostavs:
They've been skipping compilation of OPUS' machine-specific optimizations files. The portable parts do already compile fine on their own.

pljones:
Oh right - so it's more a performance enhancement than fixing something that's actually broken.  Good to have!

Can you see from the checks (https://github.com/jamulussoftware/jamulus/pull/1105/checks) whether the change had any effect on the builds?

This will have to wait until 3.7.1 now, I'm afraid, though.

npostavs:
I think https://github.com/jamulussoftware/jamulus/pull/1105/checks?check_run_id=1970718068#step:5:369 (+ the next few lines) show the files which are newly being compiled. To actually get optimizations I think we might need some more `#define`s though.

pljones:
It looks like some tuning is required to stop the Android build breaking.
```
x86/NSQ_sse4_1.o:NSQ_sse4_1.c:function silk_NSQ_sse4_1: error: undefined reference to 'silk_noise_shape_quantizer'
x86/VAD_sse4_1.o:VAD_sse4_1.c:function silk_VAD_GetSA_Q8_sse4_1: error: undefined reference to 'silk_VAD_GetNoiseLevels'
```

npostavs:
Hmm, it's not getting the `DEFINES_OPUS` stuff.

npostavs:
Looks like the problem is that `ANDROID_ARCHITECTURE` isn't actually set by anyone.

npostavs:
Everything is building now, but we're still missing the optimizations for the arm architectures.

npostavs:
I think it should all be working now.

Jamulus.pro
```diff
@@ -1066,10 +1074,42 @@ contains(CONFIG, "opus_shared_lib") {
...
 } else {
+    DEFINES += $$DEFINES_OPUS
     INCLUDEPATH += $$INCLUDEPATH_OPUS
     HEADERS += $$HEADERS_OPUS
     SOURCES += $$SOURCES_OPUS
     DISTFILES += $$DISTFILES_OPUS
+
+    contains(QT_ARCH, x86) | contains(QT_ARCH, x86_64) {
+        msvc {
+            SOURCES += $$SOURCES_OPUS_ARCH
```
hoffie, line 1085: I understand this to mean that when building with MSVC, the sources are added as-is without any specific compiler flags. Can you add a clarifying comment why that is? Does MSVC figure them out automatically?

hoffie (approved):
Thanks! Looks good to me (besides a single place which might benefit from a comment?).
Compiles on x86_64, compiler flags look as expected.
Jamulus still works without hearable changes in quality.

```
gcc -c -pipe -O2 -D_REENTRANT -Wall -Wextra -fPIC -DAPP_VERSION=\"3.7.0dev-406acad7\" -DCUSTOM_MODES -D_REENTRANT -DQT_NO_DEPRECATED_WARNINGS -DHAVE_LRINTF -DHAVE_STDINT_H -DWITH_SOUND -DOPUS_X86_MAY_HAVE_SSE -DOPUS_X86_MAY_HAVE_SSE2 -DOPUS_X86_MAY_HAVE_SSE4_1 -DOPUS_X86_PRESUME_SSE=1 -DOPUS_X86_PRESUME_SSE2=1 -DCPU_INFO_BY_C -DOPUS_BUILD=1 -DUSE_ALLOCA=1 -DOPUS_HAVE_RTCD=1 -DHAVE_LRINTF=1 -DHAVE_LRINT=1 -DQT_NO_DEBUG -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_NETWORK_LIB -DQT_XML_LIB -DQT_CONCURRENT_LIB -DQT_CORE_LIB -I. -Isrc -Ilibs/opus/include -Ilibs/opus/celt -Ilibs/opus/silk -Ilibs/opus/silk/float -Ilibs/opus/silk/fixed -Ilibs/opus -I/usr/include/qt -I/usr/include/qt/QtWidgets -I/usr/include/qt/QtGui -I/usr/include/qt/QtNetwork -I/usr/include/qt/QtXml -I/usr/include/qt/QtConcurrent -I/usr/include/qt/QtCore -I. -I. -I/usr/lib/qt/mkspecs/linux-g++ -o init_decoder.o libs/opus/silk/init_decoder.c
```

Jamulus.pro
npostavs: `opus/win32/config.h` says `/* Can always compile SSE intrinsics (no special compiler flags necessary) */`

I think the the major difference is that gcc, given the flags, will use SSE instructions automatically (this happened to me a couple times before I worked out how to apply the flags to the correct files; I got illegal instruction crashes) even when no intrinsics are used.

npostavs:
It would be good to have someone to test the ARM version (assuming it's currently functional).

hoffie:
> It would be good to have someone to test the ARM version (assuming it's currently functional).

I can confirm that compilation and audio processing works on Raspberry Pi 3 Model B Plus Rev 1.3 (armv7l).

```
gcc -c -pipe -O2 -D_REENTRANT -Wall -Wextra -fPIC -DAPP_VERSION=\"3.7.0dev-406acad7-dirty\" -DCUSTOM_MODES -D_REENTRANT -DQT_NO_DEPRECATED_WARNINGS -DHAVE_LRINTF -DHAVE_STDINT_H -DHEADLESS -DOPUS_BUILD=1 -DUSE_ALLOCA=1 -DOPUS_HAVE_RTCD=1 -DHAVE_LRINTF=1 -DHAVE_LRINT=1 -DQT_NO_DEBUG -DQT_NETWORK_LIB -DQT_XML_LIB -DQT_CONCURRENT_LIB -DQT_CORE_LIB -I. -Isrc -Ilibs/opus/include -Ilibs/opus/celt -Ilibs/opus/silk -Ilibs/opus/silk/float -Ilibs/opus/silk/fixed -Ilibs/opus -I/usr/include/qt -I/usr/include/qt/QtNetwork -I/usr/include/qt/QtXml -I/usr/include/qt/QtConcurrent -I/usr/include/qt/QtCore -I. -I/usr/lib/qt/mkspecs/linux-g++ -o repacketizer.o libs/opus/src/repacketizer.c
```

softins (approved):
I've just tried this out on my Pi 4. I made two compilations: one of this PR, and one of the branch point fabe41cf, so it was testing the same source tree.

Connecting to a 3.6.2 server on my local LAN and playing music, I observed the following:

* Branch point version: overall delay was 22ms more than the ping time.
* This PR with machine-specific sources: overall delay was 15ms more than the ping time.

softins:
> * Branch point version: overall delay was 22ms more than the ping time.
> * This PR with machine-specific sources: overall delay was 15ms more than the ping time.

Hmm, I'm not sure how I achieved that, because I have been unable to repeat the improvement a second time, with what I thought is an identical procedure. :(

npostavs:
> Hmm, I'm not sure how I achieved that

Maybe you had different buffer sizes?

I wouldn't expect this change to affect delay. I thought it should affect CPU utilization, and thought I had seen it, but I can't reproduce that now either. I'm wondering if I accidentally compared against a debug build or something. So this change is overall looking a bit disappointing: everything still works, but no measurable benefit.

hoffie:
But I guess we should still merge this as it makes libopus behave as it was intended, right?
@pljones Do you want to have a look as well or should we merge?

pljones:
I think, before merging, I'd like to see some performance figures:

- run the **old** server on a physical machine with a single core (to keep things simple to start with)
- connect sufficient clients that the CPU is running at over 50% usage and fairly constant
- repeat with the optimised code in the server and see if there's any change
- repeat the above three steps with a two-core instance (again, getting the load over 50% usage on the old code)

Ideally the above would be repeated for all CPU flavours affected -- but x64 as a minimum.

softins:
> I think, before merging, I'd like to see some performance figures:
> 
> * run the **old** server on a physical machine with a single core (to keep things simple to start with)
> * connect sufficient clients that the CPU is running at over 50% usage and fairly constant
> * repeat with the optimised code in the server and see if there's any change
> * repeat the above three steps with a two-core instance (again, getting the load over 50% usage on the old code)
> 
> Ideally the above would be repeated for all CPU flavours affected -- but x64 as a minimum.

And to be certain there are no other variables, the old code being tested should be fabe41cfa85d6bb1f21c063b39483b05ddc08d08, the branch point for these changes.

softins:
I've just done a few tests, using a RPi4 as client, and an Intel dual-core Xeon 3065 @ 2.33GHz as server, and compared the performance of the server with and without the machine-specific code. The server was _not_ using multi-threading (-T) or fastupdate (-F).

On the client, I ran one client with GUI, and then 9 or 19 simultaneous headless clients, all connected to the Intel server. I played music from VLC via Jack into all the clients, so that I wasn't just testing Opus handling silence. I also listened to the sound coming back from the server to the GUI client. As an aside, the 4-core RPi was quite happy running 20 clients at once (only one of them with GUI), with each CPU running at around 70% according to `htop`.

According to `top` on the server, CPU usage with 10 clients was 33% with the generic code, and 32% with the machine-specific code. Going up to 20 clients, CPU usage was 69% with the generic code and 67% with the machine-specific code.

So there is a small gain using the machine-specific code. I didn't notice any difference in quality, nor would I have expected to.

pljones (approved):

ann0see:
This seems to break the macOS build as noted in https://github.com/jamulussoftware/jamulus/issues/1841

ann0see:
@npostavs Do you have any pointer to where you got the compiler flags from? I think the macx-xcode build doesn't like them...

npostavs:
I think they're ultimately from `libs/opus/configure.ac`, 

```
m4_define([DEFAULT_X86_SSE_CFLAGS], [-msse])
m4_define([DEFAULT_X86_SSE2_CFLAGS], [-msse2])
m4_define([DEFAULT_X86_SSE4_1_CFLAGS], [-msse4.1])
m4_define([DEFAULT_X86_AVX_CFLAGS], [-mavx])
m4_define([DEFAULT_ARM_NEON_INTR_CFLAGS], [-mfpu=neon])
```

ann0see:
Thanks @npostavs
Do you have some more context - maybe did you consider macOS in any of your tests or do you have documentation on these flags?

npostavs:
I don't know much about macOS, and I definitely have no idea of what the difference between macx and macx-xcode is.

These flags are to enable specific hardware instructions & corresponding instrinsic functions for various generations of x86 chips (except the neon one which is for some version of ARM chips, of course). They are listed in the gcc manual here: https://gcc.gnu.org/onlinedocs/gcc-13.1.0/gcc/x86-Options.html#index-mmmx

Are you perhaps able to run `autoreconf && `./configure` under the `libs/opus` directory with the macx-xcode toolchain to see what flags it selects? (assuming whatever this xcode thing is, it's able to run autotools)

What flags are being used by the current Jamulus.pro, and what sort of errors is the macx-xcode compiler emitting?

PR #1106, opened 2021-02-24
Prepare Swedish translation för V370

genesisproject2020:
V370

### DIFF
--- a/src/res/translation/translation_sv_SE.ts
+++ b/src/res/translation/translation_sv_SE.ts
@@ -119,7 +119,7 @@
     <message>
         <location filename="../../aboutdlgbase.ui" line="72"/>
         <source>Copyright (C) 2005-2021 The Jamulus Development Team</source>
-        <translation type="unfinished"></translation>
+        <translation>Upphovsrätt (C) 2005-2021 The Jamulus Development Team</translation>
     </message>
     <message>
         <location filename="../../aboutdlgbase.ui" line="106"/>
@@ -973,7 +973,7 @@
     <message>
         <location filename="../../clientdlgbase.ui" line="584"/>
         <source>Set up your audio, connect to a server and start jamming!</source>
-        <translation type="unfinished"></translation>
+        <translation>Ställ in ditt ljud, anslut till en server och börja jamma!</translation>
     </message>
     <message>
         <location filename="../../clientdlgbase.ui" line="606"/>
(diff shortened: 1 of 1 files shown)


ann0see:
Thanks @genesisproject2020 I'll look at this shortly!

- [x] Listed in the Contributors UI window [yes](https://github.com/jamulussoftware/jamulus/blob/1c21bd379466b3c96d59fcae4dcc0576211b7621/src/util.cpp#L509)
- [] Changes look consistent (will check this soon)
- [x] No untranslated strings

src/res/translation/translation_sv_SE.ts
```diff
@@ -119,7 +119,7 @@
     <message>
         <location filename="../../aboutdlgbase.ui" line="72"/>
         <source>Copyright (C) 2005-2021 The Jamulus Development Team</source>
-        <translation type="unfinished"></translation>
+        <translation>Upphovsrätt (C) 2005-2021 The Jamulus Development Team</translation>
```
ann0see, line 122: There's still an open discussion if this should be translated: https://github.com/jamulussoftware/jamulus/pull/1097#discussion_r581265793

src/res/translation/translation_sv_SE.ts
```diff
@@ -1165,47 +1165,48 @@
...
         <location filename="../../clientsettingsdlg.cpp" line="199"/>
         <source>ASIO Device Settings push button</source>
-        <translation type="unfinished"></translation>
+        <translation>ASIO enhetsinställningsknapp</translation>
     </message>
     <message>
         <location filename="../../clientsettingsdlg.cpp" line="294"/>
         <source>ASIO Device Settings</source>
-        <translation type="unfinished"></translation>
+        <translation>ASIO enhetsinställningar</translation>
```
ann0see, line 1209: Not sure since I don't speak swedish. But is this a translation of ASIO **Device** Setup? We decided to use "ASIO Device Setup" to make it clearer mainly for newbies (using ASIO4ALL) that this opens the settings dialog for their audio devices/sound card etc. if somebody doesn't know about ASIO adding device here might be clearer.

ann0see (commented):
Thank you! This looks great! Have a look at my comments.

src/res/translation/translation_sv_SE.ts
genesisproject2020: It's a correct translation but let's see what's the decision will be.

src/res/translation/translation_sv_SE.ts
genesisproject2020: I think we should decide if we like to translate it all or partly. I personally think a mix is a bad idea. Everything else is translated and my suggestion to continue to have it like that.

ann0see: Ok. Let's wait for some others to comment here. Personally I'd prefer to have ASIO Device Setup for everything what was ASIO Setup before.

hoffie: > Ok. Let's wait for some others to comment here. Personally I'd prefer to have ASIO Device Setup for everything what was ASIO Setup before.

Do you think it should be verbatim (i.e. untranslated)?
I agree it makes sense to distinguish between "device settings" and just "settings", even in translations. Google Translate lists "enhetsinställningar" for "device settings", so in my opinion this is fine?

ann0see: Ah ok. Sorry. Probably I just misunderstood it here.

hoffie (approved):

src/res/translation/translation_sv_SE.ts
ann0see: I think we can just leave it up to the translators. So you can decide what you want to do.

ann0see (approved):

src/res/translation/translation_sv_SE.ts
pljones: I think we've agreed to leave it to translators, I'm going to merge.

Issue #1086, opened 2021-02-22
Prepare release 3.7.0


hoffie:
- [x] Decide if this should be a bugfix release (e.g. 3.6.2 -> 3.6.3) or a feature release (3.6.2 -> 3.7.0). -> We have a feature release, so it's 3.7.0
- [x] Create a milestone and add all relevant issues (@ann0see did this)
- [x] Ensure that all issues/PR targeted for this release are done by checking the [milestone](https://github.com/jamulussoftware/jamulus/milestone/1)
- [x] Declare a code freeze. PRs can still be worked on and may get reviewed, but must not be merged unless agreed explicitly.
- [x] Create at least one beta release (@ann0see did that). **WARNING**: Do not deploy beta releases to the central server which is used for update checks.
- [x] Get feedback on the stability of the beta release (@pljones and @hoffie are running servers and/or clients on the beta level)
- [x] Website translations
  - [x] Create new branch `translate3_7_0` based on branch `release` (@ann0see did this)
  - [x] Squash-merge `changes` into `translate3_7_0`
  - [x] Ping translators. Translators should open PRs with their changes against the `translate3_7_0` branch (?) and reference this issue ("Add German translations for 3.7.0")
  - [x] Wait for all PRs to be merged
  - [x] Merge `translate3_7_0` into release
- [x] App installer translations
  - [x] Ping translators. @ann0see has done this individually?
  - [x] Review PRs, <details><summary>Checklist for the PRs</summary>
    ```
    - [ ] Translator listed in the `src/util.cpp` **optionally add link to PR or code**
    - [ ] Looks consistent
    - [ ] Passes `tools/check-wininstaller-translations.sh`
    ```
    </details>
  - [x] Wait for all PRs to be merged:
    - pl #1124
    - fr #1125
    - es #1152
- [x] App translations (@softins handles this currently, https://github.com/orgs/jamulussoftware/teams/translators/discussions/6/comments/4)
  - [x] Generate `.ts` files in master via `lupdate` (@softins, directly in master)
  - [x] Ping translators (@softins). Translators do their work using `linguist` and submit PRs with their changes against master. Get the list of translators from `src/utils.cpp` (some grep/sed would be nice; last one per language is the active one).
  - [x] Review PRs, <details><summary>Checklist for the PRs</summary>
     ```
     - [ ] Translator listed in the `src/util.cpp` **optionally add link to PR or code**
     - [ ] Punctuation and spacing consistent
[...]

PR #1070, opened 2021-02-19
Typofixes

kwetal:
Fixes #1069.

Issue #1069, opened 2021-02-19
Translation typos


kwetal:
**Describe the bug**
The Dutch translation of Jamulus has some typos.

**To Reproduce**
- `Welkomsbericht` in 2 places must be `Welkomstbericht`
- `Vaardigheidssniveau` must be `Vaardigheidsniveau`

**Operating system**
all

**Version of Jamulus**
Jamulus, Version 3.6.2dev-373e1f48

**Additional context**
N/A

pgScorpio:
There are a lot of missing or wrong translations in the Dutch language:

"Custom" translated as "Aangepast" in serverlist and as "Eigen" on the settings page, (Very confusing for many users!)
"Connect" not translated on the main screen, but is translated "Verbinden" on the connections page.
"Settings" not translated in menu, but is translated "Instellingen" on the main page.
"Exit" not translated in menu.
"Audio Stream Rate" not translated on the settings page.
"Pan" not translated on the main page.

jerogee:
@pgScorpio, thanks! I will include your suggestions and corrections in the upcoming release. I am only not sure about "Pan". I believe it is quite common in Dutch audio context and there is little space for the word on the main window. Do you have anything in mind?

pgScorpio:
Pan is normally called "Balans" in dutch. (Abbreviated "Bal.")
Exit (program) is (Programma) "Afsluiten" en Audio Stream Rate is "Audio bitstream snelheid" (Strange mix of dutch and english)

kwetal:
A possible Dutch translation for "pan" might be "balans".

jerogee:
@pgScorpio, @kwetal, thanks. "Balans" it is!

PR #1110, opened 2021-02-25
Update Dutch translation for 3.7.0 (see #1086)

jerogee:
With improved consistency. Also covers #1069 and #1070.


ann0see:
I'm sorry, but I don't see any changes here? Did you commit the correct file?

src/res/translation/translation_nl_NL.ts
```diff
@@ -417,7 +417,7 @@
     <message>
         <location filename="../../audiomixerboard.cpp" line="229"/>
         <source>PAN</source>
-        <translation>PAN</translation>
+        <translation>BAL.</translation>
```
ann0see, line 420: Are you sure about the dot here?

ann0see (commented):
Thanks! See my comment.

src/res/translation/translation_nl_NL.ts
```diff
@@ -1279,42 +1279,42 @@
...
         <location filename="../../clientsettingsdlg.cpp" line="185"/>
         <source>Hz. You will not be able to select a driver/device which doesn&apos;t. For more help see jamulus.io.</source>
-        <translation type="unfinished"></translation>
+        <translation>Hz. U kunt geen stuurprogramma/apparaat selecteren die dit niet ondersteunt. Voor meer hulp zie jamulus.io.</translation>
     </message>
     <message>
         <location filename="../../clientsettingsdlg.cpp" line="199"/>
         <source>ASIO Device Settings push button</source>
-        <translation type="unfinished"></translation>
+        <translation>ASIO Device Settings drukknop</translation>
```
ann0see, line 1317: Are you sure it's not ASIO Apparaatinstellingen here too?

ann0see:
Adding the official checklist here too.

- [ ] Listed in the Contributors UI https://github.com/jamulussoftware/jamulus/blob/04bba26d365b415f6b1bf646196d5b6c5b3ae8d3/src/util.cpp#L500
- [] Changes look consistent
- [x] No untranslated strings

hoffie (commented):
Thanks! Looks good to me. @ann0see's comment should to be checked before merge.

src/res/translation/translation_nl_NL.ts
jerogee: This is the abbreviated form, hence the dot. Full form is "BALANS". For some reason there is also abbreviation in the English (GRP), probably to avoid having long labels, so followed example there:
```
        pPanLabel->setText                  ( tr ( "PAN" ) );
        pcbMute->setText                    ( tr ( "MUTE" ) );
        pcbSolo->setText                    ( tr ( "SOLO" ) );
        strGroupBaseText =                    tr ( "GRP" );
```
If longer forms are fine, let me know and I update it to "BALANS".

ann0see: Thanks for the explanation. I think the short form is better here. I'm happy to approve it.

src/res/translation/translation_nl_NL.ts
jerogee: @ann0see, nice catch, thanks! Pushed a change.

ann0see (approved):
Great! Thanks for the change.

hoffie (approved):
Thanks!

PR #1124, opened 2021-02-26
Polish translation of Windows installer NSIS script

SeeLook:
I dared to translate that.
Probably it is not enough to make it work with all NSIS structure and the installer
but from translation angle it should be sufficient.
Let me know if I can do more.


src/res/translation/wininstaller/pl.nsi
```diff
@@ -0,0 +1,37 @@
+; Polish translation
+
+LangString DESKTOP_SET_SHORTCUT ${LANG_POLISH} \
```
ann0see, line 3: I hope the LANG_POLISH is correct here.

ann0see (changes requested):
To make it work, you'll need to edit the installerlng.nsi file too.

ann0see:
Thanks for this BTW!

src/res/translation/wininstaller/pl.nsi
SeeLook: I found from where it goes: `res/translation/wininstaller/installerlng.nsi`
So there is incantation for every supported installer lang:
```
!insertmacro MUI_LANGUAGE "Dutch"
!include "${ROOT_PATH}\src\res\translation\wininstaller\nl.nsi"
```

In NSIS this `!insertmacro MUI_LANGUAGE` thing adds suitable variable `${LANG_SOMETHING}` for any language.

so I may add the same for the Polish one.

ann0see (commented):
Can you check the artifact from the checks page on this PR to verify everything works as expected?

ann0see (approved):

ann0see:
The installer seems to work fine (I had another problem, but I think it's not related and just a windows bug.)

SeeLook:
I tested Polish texts in working Windows installer and there is a small glitch in the layout - a button overlaps the text.
Jamulus-installer

But this is not a translation problem.
The rest looks just fine.

ann0see:
Ok. Probably we need to make the label a bit bigger. It should be possible to do this here: 

https://github.com/jamulussoftware/jamulus/blob/f9b89738c7abbe224fe3e699123df8b34327077a/windows/installer.nsi#L350

ann0see:
Can you edit this line and check if e.g. 13u instead of 12u in this line works better?

SeeLook:
Can I check that through this pull request? Or through another one? - This one is about translation I guess.
I have not build environment for Jamulus under Windows, so artifact is only way I can get installer executable.

ann0see:
> Or through another one?

Yes, another PR would be better

SeeLook:
But until this pull request is not merged and I send a change to my fork it will automatically land exactly here I suppose.
Or there is some other way. This github quirks I don't know well.

ann0see:
If you change it on master, yes it will add the commit here. If you select to create a new branch via the GitHub web UI, you should be able to create a new PR.

hoffie (approved):
Looks good, thanks!

Not merging yet as I'm not sure if @ann0see wants to wait for the UI fix?

hoffie:
@SeeLook Could you please rebase? I guess one of the other translations added a conflict in `src/res/translation/wininstaller/installerlng.nsi`.

SeeLook:
Please give me a hint how can I do that rebase.
For the record, my fork is at state of this PR exactly.
And I will send installer UI patch with another PR.

hoffie:
> Please give me a hint how can I do that rebase.
> For the record, my fork is at state of this PR exactly.

I think it can be done in the Github UI. Do you see a [Resolve conflicts](https://github.com/jamulussoftware/jamulus/pull/1124/conflicts) button? This should lead to the right place. It will open an editor where you would have to edit the conflicting file (`installerlng.nsi`) to contain both your changes (Polish) and the changes from master (French). I think in this case this could simply be done by removing the lines with conflict markers (`<<<<<<< master`, `=======`, `>>>>>>> master`) and making sure that the number of empty lines looks right.

You can also do the same using the git command line, but it'd be some more steps.

Can you try?


> And I will send installer UI patch with another PR.

Sounds good, thanks!

PR #1125, opened 2021-02-26
Create fr.nsi

jujudusud:
Creating the french translation of windows installer.

PR #1126, opened 2021-02-26
Update installerlng.nsi

jujudusud:
Adding French macro.


npostavs:
I guess this should be combined with #1125 to work?

jujudusud:
> I guess this should be combined with #1125 to work?

Certainly, but I don't know how to do that with the github web interface.
I will search now.

jujudusud:
I think it's merged now ? @npostavs

PR #1141, opened 2021-02-27
Update pt_BR translation for 3.7.0 (see jamulussoftware#1086)

melcon:


src/res/translation/translation_pt_BR.ts
```diff
@@ -1306,42 +1307,44 @@
...
         <location filename="../../clientsettingsdlg.cpp" line="184"/>
         <source> currently only supports devices supporting a sample rate of </source>
-        <translation type="unfinished"></translation>
+        <translation> atualmente suporta apenas dispositivos com suporte à taxa de amostra de </translation>
     </message>
     <message>
         <location filename="../../clientsettingsdlg.cpp" line="185"/>
         <source>Hz. You will not be able to select a driver/device which doesn&apos;t. For more help see jamulus.io.</source>
-        <translation type="unfinished"></translation>
+        <translatorcomment>Shouldn&apos;t the URL have the https:// prefix?</translatorcomment>
```
hoffie, line 1341: Thanks for the hint regarding the `https://` prefix. I'm not sure. I suspect that this is shown within a tooltip, so it would never be clickable anyway. My tendency would be to keep it as jamulus.io as it is shorter?

hoffie (approved):
Looks good, thanks for your work!

- [x] Listed in the Contributors UI window https://github.com/jamulussoftware/jamulus/blob/f9b89738c7abbe224fe3e699123df8b34327077a/src/util.cpp#L499
- [x] Changes look consistent
- [x] No untranslated strings (except the copyright, which is fine I'd say)

ann0see (changes requested):
Look good to me too. I'd remove the comments concerning the https:// question since I doubt the tool tip is clickable

ann0see (approved):

PR #1144, opened 2021-02-27
Tools: Add installer translation checker

hoffie:
Compares Windows installer translations with the English version, checks for consistency and common mistakes.

This is helpful for reviewing installer translation PRs. We may consider putting it into CI runs at some point. Not suggesting this right now though.

### DIFF
--- a/tools/check-wininstaller-translations.sh
+++ b/tools/check-wininstaller-translations.sh
@@ -1,0 +1,75 @@
+#!/bin/bash
+set -eu
+
+BASE_DIR=src/res/translation/wininstaller/
+BASE_LANG=en
+INSTALLERLNG=installerlng.nsi
+BASE_LANGSTRINGS=$(grep LangString "${BASE_DIR}/${BASE_LANG}.nsi" | cut -d' ' -f2)
+EXIT=0
+LANGUAGE_FILES="$(ls -1 src/res/translation/wininstaller/??.nsi | grep -vF "${BASE_LANG}.nsi")"
+for LANGUAGE_FILE in ${LANGUAGE_FILES}; do
+    echo
+    echo "* ${LANGUAGE_FILE}"
+    echo -n "  - Checking language file is included in ${INSTALLERLNG}... "
+    if grep -qF $(tr '/' '\\' <<<"${LANGUAGE_FILE}") "${BASE_DIR}/${INSTALLERLNG}"; then
+        echo "ok"
+    else
+        echo "ERROR"
+        EXIT=1
+    fi
+
+    echo -n "  - Checking language file for empty strings... "
+    if grep -qF '""' "${LANGUAGE_FILE}"; then
+        echo "WARNING, found empty string(s)"
+        EXIT=1
+    else
+        echo "ok"
+    fi
+
+    echo -n "  - Checking for comment leftovers from original file... "
+    if head -n1 "${LANGUAGE_FILE}" | grep -qF English; then
+        echo "ERROR, found 'English' in header"
+        EXIT=1
+    else
+        echo "ok"
+    fi
+
+    echo -n "  - Checking for wrong macros... "
+    LANG_MACROS="$(grep -oP '\$\{LANG_[^}]+\}' "${LANGUAGE_FILE}")"
+    if grep ENGLISH <<<"$LANG_MACROS"; then
+        echo "ERROR, found LANG_ENGLISH"
+        EXIT=1
+    else
+        echo "ok"
+    fi
+
+    echo -n "  - Checking for multiple lang macros... "
+    if [[ $(sort -u <<<"$LANG_MACROS" | wc -l) != 1 ]]; then
+        echo "ERROR, found multiple LANG_ macros"
+        EXIT=1
+    else
+        echo "ok"
+    fi
+
+    echo -n "  - Checking if LANG_ macro is in ${INSTALLERLNG}..."
+    LANG_NAME="$(sort -u <<<"${LANG_MACROS}" | sed -rne 's/\$\{LANG_(.*)\}/\1/p')"
+    if grep -qi '!insertmacro MUI_LANGUAGE "'"${LANG_NAME}"'"' "$BASE_DIR/${INSTALLERLNG}"; then
+        echo "ok"
+    else
+        echo "ERROR, not found"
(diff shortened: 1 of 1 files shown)


tools/check-wininstaller-translations.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+set -eu
+
+BASE_DIR=src/res/translation/wininstaller/
+BASE_LANG=en
+INSTALLERLNG=installerlng.nsi
+BASE_LANGSTRINGS=$(grep LangString "${BASE_DIR}/${BASE_LANG}.nsi" | cut -d' ' -f2)
+EXIT=0
+for LANGUAGE_FILE in $(ls -1 src/res/translation/wininstaller/??.nsi | grep -vF "${BASE_LANG}.nsi"); do
+		echo
```
pljones, line 10: What are you using for indent?  It's huge!

tools/check-wininstaller-translations.sh
```diff
@@ -0,0 +1,75 @@
...
+		echo "* ${LANGUAGE_FILE}"
+		echo -n "  - Checking language file is included in ${INSTALLERLNG}... "
+		if grep -qF $(tr '/' '\\' <<<"${LANGUAGE_FILE}") "${BASE_DIR}/${INSTALLERLNG}"; then
+				echo "ok"
```
pljones, line 14: Two tabs, I see...  That's rather a lot for a single level indent.

tools/check-wininstaller-translations.sh
```diff
@@ -0,0 +1,75 @@
...
+		echo -n "  - Checking for multiple lang macros... "
+		if [[ $(sort -u <<<"$LANG_MACROS" | wc -l) != 1 ]]; then
+				echo "ERROR, found multiple LANG_ macros"
+				EXIT=1
+		else
+				echo "ok"
+		fi
+
+		echo -n "  - Checking if LANG_ macro is in ${INSTALLERLNG}..."
+		if grep -qi '!insertmacro MUI_LANGUAGE "'$(sort -u <<<"${LANG_MACROS}" | sed -rne 's/\$\{LANG_(.*)\}/\1/p')'"' "$BASE_DIR/${INSTALLERLNG}"; then
```
pljones, line 54: You could split then `; then` onto the next line - it looks nicer in bash, I think.

tools/check-wininstaller-translations.sh
```diff
@@ -0,0 +1,75 @@
...
+				EXIT=1
+		fi
+
+		echo -n "  - Checking for LangStrings diffs... "
+		OTHER_LANGSTRINGS=$(grep LangString "${LANGUAGE_FILE}" | cut -d' ' -f2)
+		diff=$(diff -u <(echo "$BASE_LANGSTRINGS") <(echo "$OTHER_LANGSTRINGS"))
+		[[ ! "$diff" ]] && echo "ok" && continue
+		echo "warning, differences found:"
+		echo "---"
+		echo "$diff"
```
pljones, line 67: How does the formatting from the diff look here (i.e. are all the line breaks still there)?  If it's lost, just rerun the diff.

tools/check-wininstaller-translations.sh
```diff
@@ -0,0 +1,75 @@
...
+		echo "---"
+		echo
+		EXIT=1
```
pljones, line 70: I'd do this immediately after the `if` on 64.

hoffie: Yep, done. Thanks for all your suggestions!

tools/check-wininstaller-translations.sh
hoffie: Line breaks are still there due to proper quoting :)

```
* src/res/translation/wininstaller/de.nsi
  - Checking language file is included in installerlng.nsi... ok
  - Checking language file for empty strings... ok
  - Checking for comment leftovers from original file... ok
  - Checking for wrong macros... ok
  - Checking for multiple lang macros... ok
  - Checking if LANG_ macro is in installerlng.nsi...ok
  - Checking for LangStrings diffs... warning, differences found:
--- /dev/fd/63	2021-02-28 10:06:27.394345680 +0100
+++ /dev/fd/62	2021-02-28 10:06:27.397679038 +0100
@@ -2 +2 @@
-INVALID_FOLDER_MSG
+INVALID_FOLDER_MSGx
```

tools/check-wininstaller-translations.sh
hoffie: > You could split then `; then` onto the next line - it looks nicer in bash, I think, and means the`;` is not required.

Hrm, I'd like to keep the `then` on the same line as other scripts do it the same way (e.g. `./linux/autorelease_linux.sh`, `./android/automated_build/install-qt.sh`). I think both styles are common though.

> (You could also split the long `$(...)` over a couple of lines with line breaks after and before the `(` and `)`, plus a break with `|\` before the `sed`. i.e. one command per line.)

I've moved the long call to a variable which makes the line shorter.

tools/check-wininstaller-translations.sh
hoffie: Hm, no idea where I've got those defaults from. Changed to 4 spaces now to match `./linux/autorelease_linux.sh`.

tools/check-wininstaller-translations.sh
hoffie: See above.

tools/check-wininstaller-translations.sh
```diff
@@ -0,0 +1,75 @@
...
+    if grep -qF $(tr '/' '\\' <<<"${LANGUAGE_FILE}") "${BASE_DIR}/${INSTALLERLNG}"; then
+        echo "ok"
+    else
+        echo "ERROR"
+        EXIT=1
+    fi
+
+    echo -n "  - Checking language file for empty strings... "
+    if grep -qF '""' "${LANGUAGE_FILE}"; then
+        echo "ERROR, found empty string(s)"
```
ann0see, line 23: I'm not sure if we want to output an error here or just a warning?

hoffie: Ok, changed to warning.

ann0see (approved):
I think, this can be merged soon.

pljones (approved):

PR #1145, opened 2021-02-28
Update util.h

DavidSavinkoff:
Wrong include file in /jamulus/src/util.h
Within the /jamulus/src/util.h file:
'# include <ws2ipdef.h>'
Should be replaced with:
'# include <ws2tcpip.h>'
"The Ws2def.h, Ws2ipdef.h, and Wsipv6ok.h header files should never be used directly."
This bug fix was tested ok on Windows in this thread:
[https://github.com/jamulussoftware/jamulus/discussions/1072#discussioncomment-410564](https://github.com/jamulussoftware/jamulus/discussions/1072#discussioncomment-410564)

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -53,7 +53,7 @@
 #include "global.h"
 #ifdef _WIN32
 # include <winsock2.h>
-# include <ws2ipdef.h>
+# include <ws2tcpip.h>
 # include <windows.h>
 # include <mmsystem.h>
 #elif defined ( __APPLE__ ) || defined ( __MACOSX )


pljones (approved):

hoffie (commented):
> Wrong include file in /jamulus/src/util.h
> Within the /jamulus/src/util.h file:
> '# include <ws2ipdef.h>'
> Should be replaced with:
> '# include <ws2tcpip.h>'
> "The Ws2def.h, Ws2ipdef.h, and Wsipv6ok.h header files should never be used directly."
> This bug fix was tested ok on Windows in this thread:
> [#1072 (reply in thread)](https://github.com/jamulussoftware/jamulus/discussions/1072#discussioncomment-410564)

Thanks for your PR! I think it is a valid finding that `ws2ipdef.h` should not be used directly. I assume your reference is from here: https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options

I'm not entirely sure about replacing it with `ws2tcpip.h`. Do you know why `ws2tcpip.h` would be the right one? When checking the documented functions for that file and searching for them in the Jamulus source, I can't find any (`grep -P $(echo $(curl -s https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/ | sed -rne 's/.*<a href.*ws2tcpip.*>([^<]+)<.*/\1/p') | tr ' ' '|') src/* -r`). I cannot find anything from the original `ws2ipdef.h` either (`grep -P $(echo $(curl -s https://docs.microsoft.com/en-us/windows/win32/api/ws2ipdef/ | sed -rne 's/.*<a href.*ws2ipdef.*>([^<]+)<.*/\1/p') | tr ' ' '|') src/* -r`).
I did the same check for `#define`s based on inofficial winsdk-10 sources and came to the same conclusion (`grep -P $(echo $(curl -s https://raw.githubusercontent.com/tpn/winsdk-10/master/Include/10.0.14393.0/um/WS2tcpip.h | grep -P '^#define' | awk '{print $2}') | tr ' ' '|') src/*`). Found no easy way to check for typedefs though. I assume this include is either not needed at all or it is needed for some side effect.

Git history makes me think that it was included for `ParseNetworkAddress`. This function is still used in Jamulus code. It comes from [iphlpapi.h](https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-parsenetworkstring). So I wonder if that would be the better replacement?

Can someone with deeper Windows knowledge enlighten me? :)

```
commit 9398d80bfb177dd617f93c1af06f6d74917858ee
Author: Volker Fischer <corrados@users.noreply.github.com>
Date:   Tue Jul 9 09:16:38 2019 +0200

    added a missing header for Windows

commit 11938e960b65dee85c5daff5b007b6bef81115aa
Author: Volker Fischer <corrados@users.noreply.github.com>
Date:   Tue Jul 9 08:52:38 2019 +0200

    use #pragma once

commit 96777cbab25fb5faceab57428c6ce5e85f014f87
Author: Volker Fischer <corrados@users.noreply.github.com>
Date:   Mon Jul 8 17:02:05 2019 +0200

    add support for IPv6 for ParseNetworkAddress

```

DavidSavinkoff:
The link you showed gives the explanation at the end of the page in the Remarks section.
[https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options](https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options)
Which states:
"Some of the IPPROTO_IP socket options are defined in the Ws2ipdef.h header file which is automatically included by the Ws2tcpip.h header file."

DavidSavinkoff:
@hoffie 
> Git history makes me think that it was included for `ParseNetworkAddress`. This function is still used in Jamulus code. It comes from [iphlpapi.h.](https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-parsenetworkstring) So I wonder if that would be the better replacement?

If you scroll down to the Remarks section of [iphlpapi.h.](https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-parsenetworkstring) you will see:
"In order to use the ParseNetworkString function and the NET_ADDRESS_INFO structure, the Winsock2.h and Ws2tcpip.h header files must be included before the Iphlpapi.h header file."

hoffie:
> If you scroll down to the Remarks section of [iphlpapi.h.](https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-parsenetworkstring) you will see:
> "In order to use the ParseNetworkString function and the NET_ADDRESS_INFO structure, the Winsock2.h and Ws2tcpip.h header files must be included before the Iphlpapi.h header file."

Thanks, that's what I missed! Still wondering if `lphlpapi.h` should then be included explicitly?

DavidSavinkoff:
I could not find iphlpapi.h in the current jamulus source code.
`find . -type f -print0 | xargs -0 grep -Ii phlpapi`
If jamulus compiles without warnings about something undefined, possibly iphlpapi.h is not required.
ParseNetworkAddress must be defined somewhere else.
To test; header files can be commented out to see what fails compilation.

hoffie:
Umm, I just notice that I made a mistake here:
> Git history makes me think that it was included for ParseNetworkAddress. This function is still used in Jamulus code. It comes from iphlpapi.h. So I wonder if that would be the better replacement?

`ParseNetworkAddress` is a Jamulus-defined function and I confused it with `ParseNetworkString` which is a Windows-API-defined function where the pointers regarding `iphlpapi.h` apply.

> If jamulus compiles without warnings about something undefined, possibly iphlpapi.h is not required.
> ParseNetworkAddress must be defined somewhere else.

Yep, that's a useful idea. `iphlpapi.h` should not be relevant (this was the result of my above mentioned mistake), but checking if `ws2ipdef.h` can be removed is a good idea, I guess.
I'm trying this in my fork with https://github.com/hoffie/jamulus/commit/fec6a37bb3b6480cce01c8ff00435a640ee327b4. Let's see what the autobuild says (don't have Windows to test myself): https://github.com/hoffie/jamulus/actions/runs/608812122

hoffie:
> Let's see what the autobuild says (don't have Windows to test myself): https://github.com/hoffie/jamulus/actions/runs/608812122

The autobuild without `ws2ipdef.h` or `ws2tcpip.h` ran successfully.

To summarize:
- The include was added in 9398d80bfb177dd617f93c1af06f6d74917858ee. The commit message does not contain a reason or linked any issue/PR.
- We don't find any function or define which would come from this file as per docs in any Jamulus code.
- The build works without this include.
- I scanned all git commits since 9398d80bfb177dd617f93c1af06f6d74917858ee for changes to network-related code within WIN32 ifdefs and have found nothing which would indicate that this include was added for code, which has been removed meanwhile and the include being forgotten.

My conclusion: Either the `ip2ipdef.h` include was unnecessary in the first place (unlikely, otherwise Volker wouldn't have added it) or it is required for older Windows versions.

I'm not sure how to proceed:
1. Accept the PR as-is because it fixes an issue (according to docs) and build on the current CI Windows version still works. We don't know what happens on earlier Windows versions.
2. Don't accept the PR as we don't understand why the include was there in the first place.
3. [Modify](https://github.com/hoffie/jamulus/commit/fec6a37bb3b6480cce01c8ff00435a640ee327b4) the PR to just drop the apparently unused include. We don't know what this causes for earlier Windows versions.

I don't feel comfortable deciding this, so I'm asking @softins and maybe @pljones for help again. :)

@DavidSavinkoff Thanks again for this PR and sorry for making this such a long discussion.

DavidSavinkoff:
@hoffie I am going to submit a PR for the TOS/DSCP IP header Quality of Service byte (for all platforms with only 2 lines of code). I will need ws2tcpip.h, or must to #ifdef out Windows because the IP_TOS define is used. This is how I discovered this bug in the first place.
Note that Windows ultimately overwrites the TOS/DSCP byte with 0x00 since Windows Vista, which nullifies the effort. Nevertheless, if you use Winsock2.h (as jamulus does) you might as well add ws2tcpip.h because ws2def.h, ws2ipdef.h, and wsipv6ok.h will come along for the ride. (I think it is thought that since Jamulus is a UDP only application, that including a header with tcp in the name doesn't make sense, but ws2tcpip.h is not exclusively tcp)

DavidSavinkoff:
I've made the more recent (DSCP Quality of Service #1310) PR which includes the changes in this (Update util.h) PR because #1310 required (Update util.h) to work on Windows. (I was expecting to make 2 sequential PRs, but that plan would prevent testing #1310 until after #1145 is merged).
In short:
If DSCP Quality of Service #1310 is accepted, then Update util.h #1145 is not required.

DavidSavinkoff:
- I have explained (a bug fix as the first comment in this thread) and researched enough about ws2tcpip.h in the rest of the thread #1145 to cover windows versions going back to Platform SDK released for Windows Server 2003 and Windows XP, where you would expect IA32.
- Furthermore, the Windows artifacts Autobuild to .exe for #1145 and #1310 (I don't have windows to test with)
- dcorson-ticino-com was able to build and test an executable on his computer. see #1072 (ws2tcpip.h is required)
- These topics are fully over-discussed in #1145 #1310 #1072
- #1310 requires ws2tcpip.h
- ws2tcpip.h is certainly the correct header file.
- Maybe Qt5 or some new library would cause IA32 incompatibility with old Windows versions.

ann0see:
Closed this since the other pr was merged

PR #1147, opened 2021-02-28
NSIS: Increase height of ASIO driver explanation label

SeeLook:
- also move down button under the label apparently

But wait for artifacts - I will check the installer then, take a picture...
Then we will see is it worthy of something.

### DIFF
--- a/windows/installer.nsi
+++ b/windows/installer.nsi
@@ -347,9 +347,9 @@ Function ASIOCheckInstalled
             Abort
         ${Endif}
 
-        ${NSD_CreateLabel} 0 0 100% 12u "$(ASIO_DRIVER_EXPLAIN)"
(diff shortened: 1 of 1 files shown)


ann0see:
Just stopped the CI since it took too long

ann0see:
@SeeLook you can now take a picture

ann0see:
For me it doesn't seem as if it fixed this issue.

SeeLook:
Jamulus-installer-en
Jaumulus-installer

Total failure....
But checking installer by artifacts of pull request is wild-goose chase,
so please take this burden out of my head.
As I said, I don't have build environment for Windows and it is not that easy to get one for me.

ann0see:
Ok. I'll look into it ;-).

PR #1148, opened 2021-02-28
NSIS: Increase height of ASIO driver explanation label

ann0see:
See: #1147

Some translations need more space in the new installer. 

The artifacts still need to be verified.

### DIFF
--- a/windows/installer.nsi
+++ b/windows/installer.nsi
@@ -347,9 +347,9 @@ Function ASIOCheckInstalled
             Abort
         ${Endif}
 
-        ${NSD_CreateLabel} 0 0 100% 12u "$(ASIO_DRIVER_EXPLAIN)"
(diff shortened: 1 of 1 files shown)


softins:
I wanted to review this, so I downloaded the Windows asset from "Checks" and installed it, but didn't have any pop-ups referring to ASIO. Would that be because I already have ASIO4all installed?

softins:
I'll try it on my wife's PC, which doesn't have Jamulus or ASIO at present.

ann0see:
Yes. That's because the ASIO registry key already exists on your PC.

softins (approved):
Jamulus-install

Seems to be fine.

ann0see:
Great!

hoffie (approved):
Looks good, tested using wine. Similar results to what @softins reported.

Issue #567, opened 2020-09-02
Suggestion of adding 'positional audio' of instruments or choir sections


kezzy1966:
This is a great feature from the 'low latency' Mumble audio protocol developed for gamers to communicate while 'in game' and moving around as a platoon shooting the enemy. The player spoke from a position relative to your position in the game. 

Wondered if you guys could consider the idea of having this concept in Jamulus. With stereo headphones on,  one could hear the sound of the lead guitar coming more from the left and the bass guitar more from the right, with drums in front and the singer behind you for example. It could be an amazing immersive experience.  

Likewise if choirs could have all the 'tenors' in one audio space, Bass in another etc. 

Not sure how you would manage 'positions' but I'll that to the cleaver ones to work out. Mumble is open source so you could see how they do it.

I refer to 'Positional Audio' here: 
https://wiki.mumble.info/wiki/Features

ann0see:
Do you mean panning? It should already be available if you enable mono-in/stereo out and have a modern jamulus server running v

kezzy1966:
Ann0see, well 'pan' is half of it. In Mumble you also have in front and behind you. 
So rather than an up/down slider for 'pan', you would might have a cross or an 8 point star to position your audio.

ann0see:
I‘m quite sure you can move people behind you too. But do you mean that you can also move sb. towards/away from you?

corrados:
There is a Git branch available which implements a delay pan (https://github.com/corrados/jamulus/issues/332). I have not tried it myself but Detlef Hennings (http://www.climate-one.de) did some tests with it for the intention to be used with choirs. It should have a similar effect as far as I understood.

Implementing a positional audio is usually done using an [HRTF](https://en.wikipedia.org/wiki/Head-related_transfer_function) which means that both stereo signals are filtered. This filtering would have to be calculated in the Jamulus server which increases the processor load. Also, since Mumble is intended for voice only, Jamulus is intended for musical instruments. I am pretty sure that the HRTF filtering will change the sound of the instrument significantly which most musicians would not want.

kezzy1966:
Hi Corrados,
I do not agree. I found that Mumble and its rebirth as Mumla (Swedish for Mumble as the Swedish team took over development) has outstanding audio if you turn the audio quality up. I think it could easily handle instrument frequencies.

corrados:
> outstanding audio if you turn the audio quality up. I think it could easily handle instrument frequencies

[...]

Issue #332, opened 2020-06-05
Support delay panning in the server


corrados:
Instead of a volume panning, a delay panning can be applied in the server.
There should be a server command line argument which switches between volume and delay panning.
Initial code was already supplied by Detlef Hennings and can be found on a branch:
https://github.com/corrados/jamulus/commit/d5459802f022604b837f2679116bd7fd69418b1d

corrados:
There is an updated code available on the following branch: https://github.com/corrados/jamulus/tree/feature_hennings_delay_pan2

ann0see:
Closed via: https://github.com/jamulussoftware/jamulus/pull/1151

PR #1151, opened 2021-02-28
Delay pan

Hk1020:
See #567. This reestablished and aligned #332 with current development.


src/main.cpp
```diff
@@ -381,6 +382,29 @@ int main ( int argc, char** argv )
...
+                               "--nodelayedpan" ) )
+        {
+            bDelayPan = false;
+            qInfo() << "- starting without delayed panning";
+            CommandLineOptions << "--nodelayedpan";
+            continue;
+        }
+
+        // Enable delayed panning on startup ----------------------------------------
+        if ( GetFlagArgument ( argv,
```
ann0see, line 398: Why do we need two cli flags? Wouldn't one be enough?

src/server.cpp
```diff
@@ -1211,17 +1234,62 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
...
+                            // right channel
+                            iRpan = i - iPanDelR;
+                            if ( iRpan < 0 )
+                            {                                            // get from second
+                                iRpan = iRpan + iServerFrameSizeSamples; //
+                                vecfIntermProcBuf[k + 1] += vecsData2[iRpan];
+                            }
+                            else
+                                vecfIntermProcBuf[k + 1] += vecsData[iRpan];
+                            ;
```
ann0see, line 1258: What's this for?

ann0see (commented):
Can't say too much here, but have a look at my comments. Also please verify the style.

src/main.cpp
Hk1020: Yes one would be enough. I realized this too after I made the PR. Two options would be useful should we change the default in a later release. We’d still need to decide on the default. I’d like _on_ as default in which case you need the `—nodelaypan` option. With the other default you want the `-P/—delayedpan` option.

src/server.cpp
Hk1020: What do you mean? The spurious semicolon? That can go of course.

Hk1020 (commented):
I don’t know what I did here with _review _

DominikSchaller:
Hello, I'm in contact with Detlef Hennings who wrote the original code. He improved his code since June 2020 and can provide an updated version in a couple days. It's up to you if you want to merge this now and then do a new pull request when we get the revised version, or if you wait for this update.

I am using this delay pan feature for a couple weeks now with my chorus, and it makes a real difference! I'm glad it is finally finding its way into the main branch.

src/server.cpp
ann0see: Yes.

src/main.cpp
ann0see: If switching it on means more processing power, I would not enable it by default (there is and was a lot going on with optimizing server performance and adding something which needs more power is counter productive). You can always switch it on of course.

ann0see:
> or if you wait for this update.

I'd wait for the update then. No need to rush this.

src/main.cpp
DominikSchaller: I remember that Volker mentioned that this delay pan might not be a better experience for everybody, for example bands. So I also wouldn't enable it per default.

Hk1020: Maybe someone with a band could try it?

src/server.cpp
Hk1020: How would I have to fix this? Push another commit?

And are the lone else's without braces ok?

DominikSchaller:
@Hk1020 This is the latest version of the code of Detlef which he just sent me. I have no idea of the content, but I think you should be able to detect any changes he made and determine if they are good. I used the compiled version of the code with my chorus for a couple weeks now, we had no bugs, the server was running stable, and the sound was good.
[Jamulus_360_delpan_modified-sources_009f.zip](https://github.com/jamulussoftware/jamulus/files/6065211/Jamulus_360_delpan_modified-sources_009f.zip)

src/server.cpp
ann0see: I would use a force push. Concerning the else I don't think that's the style used overall. But since we're getting improved code, I'd wait for this before working on more here.

Hk1020:
@DominikSchaller Excellent, I‘ll have a look later, thank you (and Detlef Hennings).

ann0see:
Can we close this and re-open it once the new code is available?

DominikSchaller:
> Can we close this and open a new PR once the new code is available?

The new code is available, I posted it here today.

Hk1020:
> 
> 
> Can we close this and open a new PR once the new code is available?

I am looking at the new version right now and compare it to this one.

Hk1020:
I'd like to keep it here.

ann0see:
> The new code is already available, I posted it here last night.

Ok. Thanks. Just saw it (again).

Hk1020:
I'm done with analyzing the code. There is no functional difference in the actual panning code except one cut/paste error I introduced. Everything else is just handling of how to turn it on - the new version has fewer options - and a lot of commented code. I'll prepare a new version soon.

Hk1020:
I've updated the PR. Where is a good place to discuss the code? Here or in an issue (#567)?

ann0see:
Probably here. You can start reviewing it again

src/server.cpp
```diff
@@ -1191,12 +1204,22 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
...
         for ( j = 0; j < iNumClients; j++ )
         {
             // get a reference to the audio data and gain/pan of the current client
             const CVector<int16_t>& vecsData = vecvecsData[j];
+            const CVector<int16_t>& vecsData2 = vecvecsData2[j];
+
             const float             fGain    = vecvecfGains[iChanCnt][j];
-            const float             fPan     = vecvecfPannings[iChanCnt][j];
+            const float             fPan     = bDelayPan ? 0.5f : vecvecfPannings[iChanCnt][j];
+            const int iPanDel = lround( (float)(2 * maxPanDelay - 2) * (vecvecfPannings[iChanCnt][j] - 0.5f) );
```
ann0see, line 1220: I'm not sure about the style here.

```suggestion
            const int iPanDel = lround( (float)( 2 * maxPanDelay - 2 ) * ( vecvecfPannings[iChanCnt][j] - 0.5f ) );
```

Does this line calculate the delay between left and right? Sorry if this is obvious ;-).

src/server.cpp
```diff
@@ -1211,17 +1234,61 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
...
+                        {
+                            // pan address shift
+                            // left channel
+                            iLpan = i - iPanDelL;
+                            if ( iLpan < 0 )
+                            {                                            // get from second
+                                iLpan = iLpan + iServerFrameSizeSamples; //
+                                vecfIntermProcBuf[k] += vecsData2[iLpan];
+                            }
+                            else
```
ann0see, line 1247: I think this doesn't match the style (the else)

src/server.cpp
```diff
@@ -357,6 +359,7 @@ CServer::CServer ( const int          iNewMaxNumChan,
 
         // we always use stereo audio buffers (which is the worst case)
         vecvecsData[i].Init ( 2 /* stereo */ * DOUBLE_SYSTEM_FRAME_SIZE_SAMPLES /* worst case buffer size */ );
+        vecvecsData2[i].Init ( 2 /* stereo */ * DOUBLE_SYSTEM_FRAME_SIZE_SAMPLES /* worst case buffer size */ );
```
ann0see, line 362: Is this also needed if delayed panning is off? If not, do we need to initialize it (does it have any performance impact)? (also applies to line 345).

klingstar: Detlef Hennings is working on being able to operate the delay panning in different modes. This is about the distribution of the singers in the virtual space. 

Therefore it would be good to keep the parameter '-P panmode' free for this and similar developments and to use for delay panning only  --DelayPan (or similar) for delay panning.  

Furthermore, delay panning is so important for singers that it should be possible to activate it via a checkbox.

ann0see: > Furthermore, delay panning is so important for singers that it should be possible to activate it via a checkbox.

I think this is already part of this PR: https://github.com/jamulussoftware/jamulus/blob/0c6b3fbfc261b1b078bbd43ed939107deb07132e/src/serverdlgbase.ui

ann0see:
I'll run this code on my server and will look for feedback from my choir. I hope that it doesn't require more processing power.

Hk1020:
> 
> 
> Probably here. You can start reviewing it again

Ok. I first wanted to say that this code is not verbatim of what we got in the zip file from Detlef Hennings. I have another branch locally where I traced everything. Do you somehow want it?

This here is based on what Volker did in #332.  The only real change is the cli option. The rest is adapting it to master. I compared it to Detlef Hennings' version, which looks quite a bit og work in progress. I compared everything and verified that they are functionally equivalent. There is certainly potential to streamline the code.

src/server.cpp
Hk1020: I am not sure. There are more places where you could save on temporary variables. So far I haven't really understood what's been done here.

Hk1020:
> 
> 
> I'll run this code on my server and will look for feedback from my choir. 

So will I.

>I hope that it doesn't require more processing power.

From a cursory check via htop on my Pi4 I couldn't see any difference. But I've also been wondering about this optimization in line 1230 (don't know how to directly link here):

`            if ( ( fGainL == 1.0f ) && ( fGainR == 1.0f ) )
            {`

This has been there before and leads to a lot of code duplication, especially now. Only to save one multiplication in the end. This if branch is dead and broken in the zip file version.

> 
> Edit: before this gets merged, the commits should be attributed to Detlef Hennings.

Do you mean in some comment or somewhere else?

src/server.cpp
Hk1020: > 
> 
> > Furthermore, delay panning is so important for singers that it should be possible to activate it via a checkbox.
> 
> I think this is already part of this PR: https://github.com/jamulussoftware/jamulus/blob/0c6b3fbfc261b1b078bbd43ed939107deb07132e/src/serverdlgbase.ui

Right now this is a server setting. I'd also prefer this to be a client setting but it isn't yet. If it is we wouldn't need the command line option.

Hk1020:
Since I've never done a PR before am I supposed to do something with these comments? What does this `Resolve comments` do?

src/server.cpp
ann0see: Hmm. Before merging this, we (or at least one/two of the main developers) have to be exactly sure what each line does (and comment it). If anything goes wrong it should be possible to debug it.

src/server.cpp
ann0see: This would probably need a new protocol message.

ann0see:
> Edit: before this gets merged, the commits should be attributed to Detlef Hennings.
Do you mean in some comment or somewhere else?

I‘d do it in the commit message: https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

> am I supposed to do something with these comments?

No. Just leave them open.

src/server.cpp
klingstar: Possibly a misunderstanding: I did not mean that each client should have to start DelayPan individually. But the person who starts the server should have a checkbox to do that.

Hk1020: > 
> 
> Possibly a misunderstanding: I did not mean that each client should have to start DelayPan individually. But the person who starts the server should have a checkbox to do that.

This checkbox is already there. At least in this version. But why not make this a client setting?

src/server.cpp
```diff
@@ -1191,12 +1204,22 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
...
             // get a reference to the audio data and gain/pan of the current client
             const CVector<int16_t>& vecsData = vecvecsData[j];
+            const CVector<int16_t>& vecsData2 = vecvecsData2[j];
+
             const float             fGain    = vecvecfGains[iChanCnt][j];
-            const float             fPan     = vecvecfPannings[iChanCnt][j];
+            const float             fPan     = bDelayPan ? 0.5f : vecvecfPannings[iChanCnt][j];
+            const int iPanDel = lround( (float)(2 * maxPanDelay - 2) * (vecvecfPannings[iChanCnt][j] - 0.5f) );
+            const int iPanDelL = ( iPanDel > 0 ) ? iPanDel : 0;
+            const int iPanDelR = ( iPanDel < 0 ) ? -iPanDel : 0;
```
Hk1020, line 1223: We now have with `bDelayPan = true`:

```
    const float             fGain    = vecvecfGains[iChanCnt][j];
    const float             fPan     = bDelayPan ? 0.5f : vecvecfPannings[iChanCnt][j];
```
`fPan = 0.5`

```
    const int iPanDel = lround( (float)(2 * maxPanDelay - 2) * (vecvecfPannings[iChanCnt][j] - 0.5f) );

    const int iPanDelL = ( iPanDel > 0 ) ? iPanDel : 0;
    const int iPanDelR = ( iPanDel < 0 ) ? -iPanDel : 0;

    const float fGainL = MathUtils::GetLeftPan ( fPan, false ) * fGain;
```
`GetLeftPan` is 1 for `fPan = 0.5`, i.e, `fGainL = fGain`

```
    const float fGainR = MathUtils::GetRightPan ( fPan, false ) * fGain;

```
Same here, `fGainR = fGain`.

So we could do away with `fGainL` and `fGainR` completely and just use `fGain`. Or what am I missing?

If I am right the rest of the code could shrink a lot.

src/server.h
```diff
@@ -21,6 +21,9 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
  *
 \******************************************************************************/
+//******************************************************************************************
```
pljones, line 24: This goes in the Changelog, not the source code.

softins:
I'll review this and try it out soon.

src/server.cpp
Hk1020: Having investigated the code more `iPanDel` is the position and the delay happens later by adding `iServerFrameSizeSamples` to one of the two channels. Correct me if I’m wrong. 

We should also check how `maxPanDelay` is set.

src/server.cpp
klingstar: Thought about it once more: it is a good idea after all to make this a client setting. Thanks for asking!

DetlefHennings:
Hello Everybody!

I'm the author of the original delay panning code for Jamulus, and I would like to mention, that this code has been tested since June 2020 and is in use by different vocal groups, choirs and and also by acoustic instrumentalists since then. 
The delay pan servers were always stable and not a single bug was reported during that time. So it may be a good idea to leave the functional part of the code unchanged, even if it does not exactly comply with the coding rules of Jamulus. Please have in mind the old principle of a good design: FORM follows FUNCTION (not the other way). 
Thank you !

PS 
In the client list on the server GUI I added a column 'channels'. Mainly in larger groups like choirs this is helpful to detect if any clients use their Jamuli in mono mode and cannot take advantage of the delay panning.  
For quick 'not connected' debugging during a session it may also be useful to show the server port number on the GUI. 

PPS 
I'm visiting github only occasionally. In case of questions contacting me via mail may be faster (address in my latest source files)

ann0see:
@DetlefHennings Thank you for the code and your answer here. Just had a choir rehearsal this evening and it does make a big difference! I couldn't see that that the server load was too high, but I don't have any reference. 

> In the client list on the server GUI I added a column 'channels'

That's a good idea but could probably wait if we want to implement it here.

Hk1020:
@DetlefHennings  thanks from me too. Are you still working on an extended version as you hint at in your e-book?

We too just had a rehearsal. I first asked everyone to switch to mono-in/stereo-out, Almost everyone complained they had bad or distorted sound. I then had them switch to stereo which produced good sound. But many people didn't like it while I found it quite good. I must add, though, that many people have bad sound equipment and I think were on mono to begin with. So the whole thing was a completely new experience for them. The other strange part was that allegedly the pan knobs disappeared in the clients for certain people. I couldn't investigate further as they got impatient and wanted to sing.

For me sound is much louder on mono-in/stereo-out while mono and stereo seem to be equal.

> 
> > In the client list on the server GUI I added a column 'channels'
> 
> That's a good idea but could probably wait if we want to implement it here.

I could add it if desired.

ann0see:
> Almost everyone complained they had bad or distorted sound.

We were all using Interfaces and only one complained. But I doubt that's due to this code.

Yes, delay panning sounds different (for me, it didn't seem to move the sound behind/in front of me but rather around my head above/below. But I think that I need to verify that that's the case again.

DetlefHennings:
A few more remarks:

- yes, delay panning is louder, if not compensated: A studio standard is -3dB in the center position of volume panning. In delay panning the level is 0dB at any position on both channels. So uncompensated delay panning can be expected to be 3dB louder in avarage. This can be compensated by a constant gain factor of 0.707 (-3dB).    

- when many clients are contributing to the sound like in a choir there will also be a loudness increase in the mix, that can be statistically expected to follow  root(num of clients), as the client signals are not correlated. This effect could also be compensated by a master gain factor of 1/root(num of clients). This compensation could start above about 10 clients, at lower numbers statistics is a too bad approximation. 

- delay panning is only a rough approximation to human spacial hearing. The main effect is the externalization, the sources are heard outside of ones head. Delay panning cannot distinguish between front and back, sometimes sources are heard behind, sometimes they are elevated. A more precise modelling is possible, but requires considerably more computing power and also more programming effort.  

- in our sessions we repeatedly observed increasing distortion at higher levels in the higher frequency range. This may be due to the OPUS codec, which is a nonlinear element in the digitaltransmission and thus can produce distortion below the clipping level. In consequence we decided to reduce our microphone levels to stay normally in the green level meter range. 

- we found that some headsets produce horrible sound, others are acceptable. We also fons that some low cost small diaphragm condenser microphones are really good for this purpose. For some vocalists an adjustable direct monitoring path in the audio interface is a big help to hear their own voices.  

- finally: yes, there is a working 'luxury' server version as I noted in my e-booklet. This version positiones the clients on a circle and all pans are calculated from the geometry. This version is working well, but it still makes use of my lab equipment and it is controlled manually in a way, that is far too complicated for distribution.    
The good news: I'm currently working on a semi-automatic version that is much easier to operate, and that may become public in the next few months.

klingstar:
> 
> 
> > Almost everyone complained they had bad or distorted sound.
> 
> We were all using Interfaces and only one complained. But I doubt that's due to this code.
> 
> Yes, delay panning sounds different (for me, it didn't seem to move the sound behind/in front of me but rather around my head above/below. But I think that I need to verify that that's the case again.)

klingstar:
Hearing other singers from above or behind, is a logical function of our brain. Your ears hear people sing, but your eyes don't see them. So your brain comes to the conclusion, that the other singers must be out of your sight: behind, above, below you. If you close your eyes when singing, the mismatch disappears, and your brain will remember the situation of other singers before or beside you.

Hk1020:
I tried this with a bunch of regional choir leaders here the other day. They were thrilled by delay pan.

Hk1020:
Ok, I‘ll change the single line if/else. 

> Yes. Some code style changes (mainly the else) and a description what the code does (maybe some more comments).
> 
> This is still pending since we need some more performance tests (for me it looks good) and we'll need to wait until 3.7.0 is out.

I was hoping to get it out sooner. 

I‘d also like to remove the block starting on line 1230 in the new version. I doubt it has any effect on performance and in the latest version that Detlef Hennings sent us here it is made inactive. He reports his version works fine.

DominikSchaller:
> Yes. Some code style changes (mainly the else) and a description what the code does (maybe some more comments).
> 
> This is still pending since we need some more performance tests (for me it looks good) and we'll need to wait until 3.7.0 is out.

FWIW We also use the version of Detlef weekly with my chorus for the last couple weeks with no negative effects.

Hk1020:
> > Yes. Some code style changes (mainly the else) and a description what the code does (maybe some more comments).
> > This is still pending since we need some more performance tests (for me it looks good) and we'll need to wait until 3.7.0 is out.
> 
> FWIW We also use the version of Detlef weekly with my chorus for the last couple weeks with no negative effects.

Is that the version from the zip file or the stuff I posted here?

Hk1020:
Ok, I will update the code later. Style and remove that if branch.

DetlefHennings:
- Dominik is right, the binaries he uses base on the same code as is in my latest zip. 
- in my code the gain=1 extra is switched off, because I assume that in real sessions gains are usually set to values <1.

ChangeLog
```diff
@@ -4,6 +4,8 @@
 
 ### 3.6.2dev <- NOTE: the release version number will be 3.7.0 ###
 
+- Pan-Delay code added by Detlef Hennings, www.eclim.de,   2020-06
```
pljones, line 7: Note that it'll actually be after 3.7.0 now but thanks.

src/serverdlgbase.ui
```diff
@@ -44,6 +44,16 @@
...
       </property>
      </column>
+     <column>
+      <property name="text">
+       <string>Channels</string>
+      </property>
+     </column>
+     <column>
+      <property name="text">
+       <string>   </string>
```
pljones, line 54: Could do with _something_ here :)

src/server.cpp
```diff
@@ -1191,60 +1204,112 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
...
+
         for ( j = 0; j < iNumClients; j++ )
         {
             // get a reference to the audio data and gain/pan of the current client
             const CVector<int16_t>& vecsData = vecvecsData[j];
+            const CVector<int16_t>& vecsData2 = vecvecsData2[j];
+
             const float             fGain    = vecvecfGains[iChanCnt][j];
-            const float             fPan     = vecvecfPannings[iChanCnt][j];
+            const float             fPan     = bDelayPan ? 0.5f : vecvecfPannings[iChanCnt][j];
```
pljones, line 1219: Does `fPan` get used?

Hk1020: Yes, lines 1226/1227, see next diff

src/server.cpp
pljones: There would need to be:
- a message from the server saying "I support delay pan if you want it"
- a message from the client saying "I would like delay pan"
- a client method to send the message
- a clientUI mechanism to request the feature _only if the server supports it_ (which would need to take server version and the specific message into account).  (I'm not sure if this should be in settings, as it's a user preference, or a on the server mixer, as the choice might vary depending on the jam session.)

I also think maybe get the server support finished and released "as is" before adding the client side.  I definitely think it's worth having, though.

Hk1020: I‘d also like to get this done. The sooner the better. People would benefit immediately. The client part can come later and this should go out right away (3.7.0). But it is not on me to decide. I too can’t imagine anyone wanting to go back to volume panning. 

I‘d been entertaining the idea to provide a client setting in a later PR. The grounds are laid and there isn’t much to add to the server logic except the protocol bit. Mixing is per client already so we only need to make `bDelPan` a vector, similar to `fPan`.

softins: The problem is, because we want to be certain about the quality and fully understand operation of the code that goes into a release, every additional feature causes further delay in the release of 3.7.0, which already is later than we would have liked. That is not to express any doubt about the code quality or its usefulness, it just takes time.

It will definitely be fully reviewed for inclusion in 3.7.1, and hopefully that release will not be too far into the future.

src/main.cpp
```diff
@@ -837,6 +850,7 @@ QString UsageArguments ( char **argv )
         "  -m, --htmlstatus      enable HTML status file, set file name\n"
         "  -o, --serverinfo      infos of this server in the format:\n"
         "                        [name];[city];[country as QLocale ID]\n"
+        "  -P, --delayedpan      start with delayed panning enabled\n"
```
hoffie, line 853:
```suggestion
        "  -P, --delayedpanning  start with delayed panning enabled\n"
```

DominikSchaller: This should be called "delaypan" or "delaypanning". I don't know where the term "delayed panning" comes from, but it is not accurate, because the panning is not delayed. The panning uses delays to achieve its function. Might be pedantic, but it shouldn't give the impression that this effect is delayed (and increases latency) in any way.

hoffie: Would be fine for me, I just suggested the change because the help text currently does not match the implementation. It has to be changed either wa. :)

DetlefHennings: The English term is 'delay panning' and the short form 'pan' is often used instead of 'panning'. So both 'delay panning' or  'delay pan' would be OK.

Hk1020: @hoffie where do you see this doesn’t match the implementation?

hoffie: The help text says `--delayedpan` (the line I commented on) and the implementation parses `--delayedpanning` ([line 389](https://github.com/jamulussoftware/jamulus/pull/1151/files#diff-34d21af3c614ea3cee120df276c9c4ae95053830d7f1d3deaf009a4625409ad2R389))

src/server.h
```diff
@@ -270,6 +274,9 @@ class CServer :
     void SetAutoRunMinimized ( const bool NAuRuMin ) { bAutoRunMinimized = NAuRuMin; }
     bool GetAutoRunMinimized() { return bAutoRunMinimized; }
 
+    int GetClientNumAudioChannels ( const int iChanNum )
+        { return vecChannels[iChanNum].GetNumAudioChannels(); }
+
```
hoffie, line 279: I think there are stray spaces here (which Github doesn't show?).

src/server.cpp
```diff
@@ -1191,60 +1204,112 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
...
+            const int iPanDelL = ( iPanDel > 0 ) ? iPanDel : 0;
+            const int iPanDelR = ( iPanDel < 0 ) ? -iPanDel : 0;
 
             // calculate combined gain/pan for each stereo channel where we define
             // the panning that center equals full gain for both channels
             const float fGainL = MathUtils::GetLeftPan ( fPan, false ) * fGain;
             const float fGainR = MathUtils::GetRightPan ( fPan, false ) * fGain;
 
-            // if channel gain is 1, avoid multiplication for speed optimization
-            if ( ( fGainL == 1.0f ) && ( fGainR == 1.0f ) )
```
hoffie, line 1207: This code seems to get lost and was annotated to be a speed optimization. I can't tell how important it was. Saving some multiplications sounds really minor (if the cost is two additional equality checks), but as this is part of the performance-critical code path we may want to double-check.

hoffie:
I have tried this with my choir and this was a huge improvement for us. Thanks to @DetlefHennings and @Hk1020 for working on this. I'd really welcome if we can get this into 3.7.1.

I wasn't able to have a complete look at the diff yet, but it certainly seems to be doable. :)

hoffie:
This still lacks some final reviews. Meanwhile, could you rebase/solve the Changelog conflict @Hk1020? Thanks!

src/main.cpp
Hk1020: Ok, I've just fixed that and use delaypan/delay panning throughout.

I am still wondering, though, if we shouldn't make this default to on. As Detlef Hennings wrote essentially nobody would go back to volume panning once they heard delay panning and I very much agree. Especially choirs will benefit from this, so this will be a lot of people, and from my experience with helping them they struggle a lot with setting up their own server. I think most won't touch anything again as they are glad the thing works at all. They would all miss out on this if we ask them to set another option. On the other hand, those who will be annoyed by the new sound will figure out how to go back if we provide the option for it.

Hk1020:
> 
> 
> Meanwhile, could you rebase/solve the Changelog conflict @Hk1020? Thanks!

Done.

src/main.cpp
DetlefHennings: Could you please rename the option to either '--delaypan' or '--delaypanning', but please do not use '--delayedpan' and '--delayedpannning', which are no correct terms.   

My suggestion to make delpan public is to give the information, if delay or volume panning is used, in the public server lists. This could be done by adding a (d) for delay panning and a (v) for volume panning behind the musicians count/capacity, e.g.  18/30 (d)  or  5/8 (v).

Hk1020: `--delaypan` is what I set it to.

I was rather thinking of private server installations, which is what a lot of choirs use. If we make the default _on_ people would just have to update their server and get it. If we also have to make them update their service installation they will be lost.

Announcement in the server list is beyond of what I did here.

DetlefHennings: --delaypan  is fine (I didn't see, that it was already changed). 

Delay pan as default, that's what I would also prefer. Maybe to some users it will be a (positive) surprise then ...

pljones (approved):

ann0see:
@gilgongo I think this needs to be documented on the changes branch in the documentation. @DetlefHennings can you give a short description for the documentation so we can add it here: https://jamulus.io/wiki/Command-Line-Options

DetlefHennings:
The short description can be: 
--delaypan       Switches the server to 'delay panning' mode instead of the default 'volume panning' mode. 

A bit more detailed description (elsewhere) can be: 
When using a Jamulus server in 'delay panning' mode, the musicians listening under their head­phones can hear each other outside of their heads like in a physical meeting. The heard directions to the others can be con­trol­led by the respective PAN knobs. To achieve this result, all client Jamuli should be set to 'mono in / stereo out'.

Hk1020:
Although this has now been merged - thank you - it seems the conversation above https://github.com/jamulussoftware/jamulus/pull/1151#discussion_r599048572 about defaults got lost. I’d still like to raise it again.

softins:
> Although this has now been merged - thank you - it seems the conversation above [#1151 (comment)](https://github.com/jamulussoftware/jamulus/pull/1151#discussion_r599048572) about defaults got lost. I’d still like to raise it again.

Ok, see below

> Ok, I've just fixed that and use delaypan/delay panning throughout.
> 
> I am still wondering, though, if we shouldn't make this default to on. As Detlef Hennings wrote essentially nobody would go back to volume panning once they heard delay panning and I very much agree. Especially choirs will benefit from this, so this will be a lot of people, and from my experience with helping them they struggle a lot with setting up their own server. I think most won't touch anything again as they are glad the thing works at all. They would all miss out on this if we ask them to set another option. On the other hand, those who will be annoyed by the new sound will figure out how to go back if we provide the option for it.

I understand your motivation for this, but my view is that we shouldn’t make enabled by default a new behaviour that most people have not tried out and experienced. No problem with highlighting it and promoting the benefits.

If there is a lot of positive feedback from people who have tried it, you can be sure that word will spread, and we could consider making it the default in 3.8

npostavs:
Speaking as someone fairly ignorant of the details of audio tech, I'd be curious as to what the differences are. I've briefly tried this out, but I couldn't notice anything obvious (although having it controlled by command line parameter makes it more difficult to compare side-by-side).

Hk1020:
> Speaking as someone fairly ignorant of the details of audio tech, I'd be curious as to what the differences are. I've briefly tried this out, but I couldn't notice anything obvious (although having it controlled by command line parameter makes it more difficult to compare side-by-side).

Actually, if you start the server with the GUI there is a checkbox to turn it on or off on the fly. It makes the difference really obvious. But not with mono!

npostavs:
Oh, I was passing the option to the client instead, no wonder I didn't notice any difference. Oops! (maybe passing it to the client should give a warning though, instead of saying `- starting with delay panning`).

In searching for explanations/examples about it I found [this youtube video](https://www.youtube.com/watch?v=UQOkSF8auFc) which talks about "the Haas effect". Is that the same as "delay panning"?

softins:
> Oh, I was passing the option to the client instead, no wonder I didn't notice any difference. Oops! (maybe passing it to the client should give a warning though, instead of saying `- starting with delay panning`).

There are actually a lot of server-only options that do not generate a warning when given to a client, although they are then ignored. It would be good to tidy that up a bit!

Hk1020:
> > Oh, I was passing the option to the client instead, no wonder I didn't notice any difference. Oops! (maybe passing it to the client should give a warning though, instead of saying `- starting with delay panning`).
> 
> There are actually a lot of server-only options that do not generate a warning when given to a client, although they are then ignored. It would be good to tidy that up a bit!

From what I could see there isn’t any support in the current command line parsing for further dependency checking. Has to be done manually.

DetlefHennings:
> Oh, I was passing the option to the client instead, no wonder I didn't notice any difference. Oops! (maybe passing it to the client should give a warning though, instead of saying `- starting with delay panning`).
> 
> In searching for explanations/examples about it I found [this youtube video](https://www.youtube.com/watch?v=UQOkSF8auFc) which talks about "the Haas effect". Is that the same as "delay panning"?

Delay panning, when used with headphones, simply 'simulates' the runtime difference between the two ears depending on the horizontal angle of sound incidence (it's not the Haas effect). This relation to human hearing makes it sound so natural. BTW, the effekt is lost, when a delay pan is set exactly to center.

Hk1020:
> > In searching for explanations/examples about it I found [this youtube video](https://www.youtube.com/watch?v=UQOkSF8auFc) which talks about "the Haas effect". Is that the same as "delay panning"?
> 
> Delay panning, when used with headphones, simply 'simulates' the runtime difference between the two ears depending on the horizontal angle of sound incidence (it's not the Haas effect). This relation to human hearing makes it sound so natural. BTW, the effekt is lost, when a delay pan is set exactly to center.

Now you got me curious. How is this not the Haas effect (which I learned about just now)?

softins:
> > > In searching for explanations/examples about it I found [this youtube video](https://www.youtube.com/watch?v=UQOkSF8auFc) which talks about "the Haas effect". Is that the same as "delay panning"?
> > 
> > 
> > Delay panning, when used with headphones, simply 'simulates' the runtime difference between the two ears depending on the horizontal angle of sound incidence (it's not the Haas effect). This relation to human hearing makes it sound so natural. BTW, the effekt is lost, when a delay pan is set exactly to center.
> 
> Now you got me curious. How is this not the Haas effect (which I learned about just now)?

I just watched the video. Really interesting, and a good demonstration of delay panning. Although more than a couple of ms difference is more than the ears would naturally encounter, because sound travels through air at approximately 34cm per ms.

I then read about the Haas effect at https://en.wikipedia.org/wiki/Precedence_effect. That seems to be talking about something slightly different: hearing a sound with both ears, followed by an identical sound, again with both ears, a very short delay later (a few ms). Something like a very quick echo. It doesn't appear to be talking why a single sound's perceived direction is because of the difference in arrival time at each ear.

So it seems the guy making the video was mistaken in calling it the Haas effect.

DominikSchaller:
Thanks for merging this feature into master! This stale branch can be deleted now: https://github.com/jamulussoftware/jamulus/tree/feature_hennings_delay_pan2

ann0see:
I think it's removed now

DetlefHennings:
### Next generation choir server

It seems that delay panning is going to be delayed (!) by introducing more and more issues into v.3.7.1, or now v.3.8.0. But meanwhile the next generation server for choirs and other acoustic music ensembles is under test. 

This server arranges all clients in a circle geometry and all (delay) pans are calculated from this geometry and are set automatically. This avoids controlling a large number of pans. Individual client positions on the circle are set either by numeric input or by a pseudo random algorithm. 

For testing Debian and Ubuntu beta binaries and a short explanation are available under [this link](http://eclim.de/jam/Jamulus_CircPosE.htm).

pljones:
We're in planning for release now :).

jstallkamp:
Circle mode sounds awesome. It'd be great if geometry could also be controlled "by group" (without manually assigning numbers to participants' names). But that might be the/a next step, after the feature has been shown to work.

PR #1155, opened 2021-03-01
Improve version update detection.

softins:
Send version requests on startup to two different central servers.
It doesn't matter if they both reply, so no need to wait and retry.
Only compare if server is running a released version (no suffix).

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -861,9 +861,6 @@ void CClientDlg::OnCLVersionAndOSReceived ( CHostAddress           ,
     int serverSuffixIndex;
     QVersionNumber serverVersion = QVersionNumber::fromString ( strVersion, &serverSuffixIndex );
 
-    qDebug() << "My version: " << myVersion << " suffix: " << QString( VERSION ).mid( mySuffixIndex );
-    qDebug() << "Server version: " << serverVersion << " suffix: " << strVersion.mid( serverSuffixIndex );
-
     // only compare if the server version has no suffix (such as dev or beta)
     if ( strVersion.size() == serverSuffixIndex && QVersionNumber::compare ( serverVersion, myVersion ) > 0 )
     {
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -573,7 +573,15 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // chance and the update check is not time-critical at all)
     CHostAddress CentServerHostAddress;
 
-    if ( NetworkUtil().ParseNetworkAddress ( DEFAULT_SERVER_ADDRESS, CentServerHostAddress ) )
+    // Send the request to two servers for redundancy if either or both of them
+    // has a higher release version number, the reply will trigger the notification.
+
+    if ( NetworkUtil().ParseNetworkAddress ( UPDATECHECK1_ADDRESS, CentServerHostAddress ) )
+    {
+        pClient->CreateCLServerListReqVerAndOSMes ( CentServerHostAddress );
+    }
+
+    if ( NetworkUtil().ParseNetworkAddress ( UPDATECHECK2_ADDRESS, CentServerHostAddress ) )
     {
         pClient->CreateCLServerListReqVerAndOSMes ( CentServerHostAddress );
     }
@@ -847,7 +855,17 @@ void CClientDlg::OnCLVersionAndOSReceived ( CHostAddress           ,
 {
     // update check
 #if ( QT_VERSION >= QT_VERSION_CHECK(5, 6, 0) ) && !defined ( DISABLE_VERSION_CHECK )
-    if ( QVersionNumber::compare ( QVersionNumber::fromString ( strVersion ), QVersionNumber::fromString ( VERSION ) ) > 0 )
+    int mySuffixIndex;
+    QVersionNumber myVersion = QVersionNumber::fromString ( VERSION, &mySuffixIndex );
+
+    int serverSuffixIndex;
+    QVersionNumber serverVersion = QVersionNumber::fromString ( strVersion, &serverSuffixIndex );
+
+    qDebug() << "My version: " << myVersion << " suffix: " << QString( VERSION ).mid( mySuffixIndex );
+    qDebug() << "Server version: " << serverVersion << " suffix: " << strVersion.mid( serverSuffixIndex );
+
+    // only compare if the server version has no suffix (such as dev or beta)
+    if ( strVersion.size() == serverSuffixIndex && QVersionNumber::compare ( serverVersion, myVersion ) > 0 )
     {
         // show the label and hide it after one minute again
         lblUpdateCheck->show();
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -567,23 +567,23 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
         chbLocalMute->setCheckState ( Qt::Checked );
     }
 
-    // query the central server version number needed for update check (note
+    // query the update server version number needed for update check (note
     // that the connection less message respond may not make it back but that
     // is not critical since the next time Jamulus is started we have another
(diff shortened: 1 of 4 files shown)


hoffie (approved):
Thanks for doing this! Looks got besides spacing nits. :)

Is it worth a Changelog entry?

I tested the following:
- The new code does not currently trigger an update notification -- as expected.
- The new code does trigger an update notification when pointing update1.jamulus.io to a server which pretends to run 3.6.3 or 3.7.0 (via /etc/hosts) -- as expected.
- The new code not does trigger an update notification when pointing update1.jamulus.io to a server which pretends to run 3.6.3beta1(via /etc/hosts) -- as expected.

> Only compare if server is running a released version (no suffix).

I think this is a good approach. However, Volker currently runs c1bc4274 on central1. This would be no issue for 3.7.0, but we will have to make sure that neither Volker nor @pljones run anything other than final releases on these servers. Would we need to document that somewhere? If so, where?

softins:
> Is it worth a Changelog entry?

Probably. I haven't got into the habit of adding those yet, as previously, Volker would add them all when getting a release ready.

> I tested the following:
> 
> * The new code does not currently trigger an update notification -- as expected.
> * The new code does trigger an update notification when pointing update1.jamulus.io to a server which pretends to run 3.6.3 or 3.7.0 (via /etc/hosts) -- as expected.
> * The new code not does trigger an update notification when pointing update1.jamulus.io to a server which pretends to run 3.6.3beta1(via /etc/hosts) -- as expected.

Great. My tests were very similar.

> > Only compare if server is running a released version (no suffix).
> 
> I think this is a good approach. However, Volker currently runs [c1bc427](https://github.com/jamulussoftware/jamulus/commit/c1bc427414e6713026e38a5408958289f81b71d4) on central1. This would be no issue for 3.7.0, but we will have to make sure that neither Volker nor @pljones run anything other than final releases on these servers. Would we need to document that somewhere? If so, where?

Once 3.7.0 is out, the use of suffixes ceases to be a problem, in general (but see next para). The issue we are currently avoiding is putting up a 3.7.0beta server, which 3.6.2 clients would see and show the red message. Older clients don't matter, because there is already a newer version available than they are. Once the current release is 3.7.0, it doesn't matter to older clients if they see 3.7.1beta and put up the message, because it is still true.

With the redundancy, we just need if possible to make sure that one or other of the servers is running a release (at the moment neither is). So yes, Volker and @pljones need to know to check the other one is on a release before trying out a new beta or dev build.

hoffie:
>With the redundancy, we just need if possible to make sure that one or other of the servers is running a release (at the moment neither is). So yes, Volker and @pljones need to know to check the other one is on a release before trying out a new beta or dev build.

Yes, that's what I meant.

Updating the current reference server to a beta and using pre 3.7.0 clients would risk a false positive (update available, despite none being there) until 3.7.0 is released.
Updating both new reference servers to a beta and using 3.7.0 clients would risk false negatives (no update available, despite there being one).

src/clientdlg.cpp
```diff
@@ -847,7 +855,17 @@ void CClientDlg::OnCLVersionAndOSReceived ( CHostAddress           ,
...
 #if ( QT_VERSION >= QT_VERSION_CHECK(5, 6, 0) ) && !defined ( DISABLE_VERSION_CHECK )
-    if ( QVersionNumber::compare ( QVersionNumber::fromString ( strVersion ), QVersionNumber::fromString ( VERSION ) ) > 0 )
+    int mySuffixIndex;
+    QVersionNumber myVersion = QVersionNumber::fromString ( VERSION, &mySuffixIndex );
+
+    int serverSuffixIndex;
+    QVersionNumber serverVersion = QVersionNumber::fromString ( strVersion, &serverSuffixIndex );
+
+    qDebug() << "My version: " << myVersion << " suffix: " << QString( VERSION ).mid( mySuffixIndex );
+    qDebug() << "Server version: " << serverVersion << " suffix: " << strVersion.mid( serverSuffixIndex );
```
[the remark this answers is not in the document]
softins: I suppose. I'm not a fan of that style. But looking at other code, it seems like the convention is space before `(` if there are arguments, but not before `()` ?

I was in two minds about leaving the debugs in or taking them out.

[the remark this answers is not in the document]
softins: I think on reflection, I'm probably inclined to take out the debugs, but leave them in the commit history (no squash merge!). Would you agree?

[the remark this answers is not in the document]
hoffie: > I suppose. I'm not a fan of that style.

Neither am I. But I think we should be consistent.

> But looking at other code, it seems like the convention is space before `(` if there are arguments, but not before `()` ?

Yes, I think so.

> I was in two minds about leaving the debugs in or taking them out.

If we had some `--debug` or `--verbose` flag which controlled if `qDebug()` messages are shown or not, I'd say: keep them.
With the current default log level, they do sound a bit verbose, in my opinion, so my tendency would be to remove them.

softins:
> Updating the current reference server to a beta and using pre 3.7.0 clients would risk a false positive (update available, despite none being there) until 3.7.0 is released.

That's right, any betas of 3.7.0 need to be built as 3.6.2dev. We can't build as 3.7.0betaN.

src/clientdlg.cpp
```diff
@@ -573,7 +573,15 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // chance and the update check is not time-critical at all)
     CHostAddress CentServerHostAddress;
 
-    if ( NetworkUtil().ParseNetworkAddress ( DEFAULT_SERVER_ADDRESS, CentServerHostAddress ) )
+    // Send the request to two servers for redundancy if either or both of them
+    // has a higher release version number, the reply will trigger the notification.
+
+    if ( NetworkUtil().ParseNetworkAddress ( UPDATECHECK1_ADDRESS, CentServerHostAddress ) )
```
ann0see, line 579: Any reason why you don't use an or (||) here?

```
if ( NetworkUtil().ParseNetworkAddress ( UPDATECHECK1_ADDRESS, CentServerHostAddress ) || NetworkUtil().ParseNetworkAddress ( UPDATECHECK2_ADDRESS, CentServerHostAddress )  )

Comment (hoffie) [reply src/clientdlg.cpp:579]: I thought about this, too, but this is not just a check for the return value of `ParseNetworkAddress()`, but the call populates `CentServerHostAddress`. If this was done in one go, the second call would overwrite the first one.
Of course, one could split this into CentServerHostAddress1 and CentServerHostAddress2 as well, but I guess the current code saves some bytes of memory. :)

Comment (softins) [reply src/clientdlg.cpp:579]: And we actually want to send requests to both servers. There’s no point waiting for the reply from one before sending the request to the other, and that would be a lot more complicated to do.

Comment (softins) [reply src/clientdlg.cpp:579]: And each test-and-send is independent of the other. They just happen to share the same temporary variable.

Comment (ann0see) [reply src/clientdlg.cpp:579]: Ok. Thanks for the explanation.

File: src/clientdlg.cpp [hunk 3]
```diff
@@ -573,7 +573,15 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // chance and the update check is not time-critical at all)
     CHostAddress CentServerHostAddress;
```
pljones, line 574: I'd have changed this variable name...

pljones (approved):

pljones:
> The issue we are currently avoiding is putting up a 3.7.0beta server, which 3.6.2 clients (or local servers) would see and show the red message.

Which only applies to Volker's server.  Anything before 3.7.0 won't be looking at update2.

src/clientdlg.cpp
softins: Done. Now `UpdateServerHostAddress`.

ann0see:
Ready to be merged?

pljones (approved):

softins:
> Ready to be merged?

Yes, I just want to adjust my last commit, and then I will merge it.

Issue #1161, opened 2021-03-02
Check for duplicated accelerator keys in translations


softins:
Just making sure this activity is tracked for the release. I'm on the case.

softins:
Quite a few conflicts to resolve: https://gist.github.com/softins/d0587ae7e1ea44e9e0eeb352bc0eb8a0

softins:
I'll leave this open for now as a reminder to revisit it when we receive pt_PT and sk_SK translations.

Snayler:
> I'll leave this open for now as a reminder to revisit it when we receive pt_PT and sk_SK translations.

Did you mean pt_BR? I checked all the duplicates on pt_PT and they don't seem to interfere with each other, as they don't overlap in the different menus or on the main window.

Snayler:
> I'll leave this open for now as a reminder to revisit it when we receive pt_PT and sk_SK translations.

Just created a pull request with the updated translation. I'm using a notebook where I'm writting all the accelerator keys I'm using to make sure they don't overlap on the same menu or in the main window, so there should be no problems there.

softins:
pt_PT checked and merged. Thank you very much @Snayler!

hoffie:
@softins Would that script be something for the `tools/` folder?

softins:
Certainly, if it would be useful. It is not automated, it just lists potential clashes for human checking. It isn't possible, for example to distinguish between the same character used in entries down different menus.

softins:
Now that we have no more translation needed for 3.7.0, I'll move this issue to 3.7.1 as a reminder for the next release.

hoffie:
A checklist item for doing this check for each release has now been added to #1594. The 3.8.0 preparation Issue (#1595) has been updated as well. So this ticket could be closed?

ann0see:
Closing since it's part of the preparation issue.

PR #1165, opened 2021-03-03
Avoid conflicting accelerator keys.

softins:
Related to #1161, but might need to be revisited for pt_PT and sk_SK when we have received complete translations.


softins:
Just to clarify: I ran the GUIs in each language to verify which duplicates were actual conflicts and identify available keys.

pljones (approved):

hoffie (approved):
@softins Thanks a lot for doing this! This sounds like a lot of manual work, even when script-assisted.

I verified that this PR only changes accelator keys and leaves everything else untouched:
`diff -u0 <(git show | grep -P '^\+ ' | sed -re 's/&amp;//; s/&//; s/^\+//' | sort) <(git show | grep -P '^- ' | sed -re 's/&amp;//; s/&//; s/^-//' | sort)`

I have not verified uniqueness of the keys again. It can only get better :)

PR #1168, opened 2021-03-03
Update pt_PT translation

Snayler:


hoffie (approved):
Thank you very much for your work! Looks good to me.

- [x] Translator listed in the `src/util.cpp`https://github.com/jamulussoftware/jamulus/blob/466c76e3007c2a6ae5cef2a43669764cf7c1964b/src/util.cpp#L498
- [x] No untranslated strings (`grep unfinished -5 src/res/translation/translation_$TRANSLTION*.ts`)
- [x] Only a single `.ts` file checked in (+.qm which is known to be not a problem)

Snayler:
>     * [x]  Only a single `.ts` file checked in (+.qm which is known to be not a problem)

So no need for the .qm file next file I do the translations?

hoffie:
> So no need for the .qm file next file I do the translations?

As far as I understand: Yes, there's no need for you to check that in all required files will be generated as part of the release process. :)
It's no issue either though.

softins (approved):
Looks good to me. I couldn't find any conflicting accelerator keys.

Issue #909, opened 2021-01-28
Mute myself and mute own channel leads to sound coming back to headphones


ann0see:
@mulyaj discovered a bug:

Steps to reproduce:

1. Click the "mute myself" button
2. Click the "Mute" button on your own channel in the mix

Actual behaviour:

You hear sound coming back from the server (it's local monitoring)

Expected behaviour

No sound coming back from the server

DominikSchaller:
The sound is actually not coming back from the server, it's direct monitoring and I believe it's a feature not a bug.

ann0see:
Ah. Interesting. But that shouldn't be the case if you're muted in your own mix.

DominikSchaller:
I believe the mixer is always referring to the sound coming from the server and as you are not sending anything the mute button has nothing to mute. I wonder what the use case would be in which you are playing but nobody (not even you) should here it.

DavidSavinkoff:
I'm using Janulus 3.6.2 on linux which works correctly (not like you mention), and as expected. Please give more details.

ann0see:
Yes. Can verify that it's not coming from the server. 

> I believe the mixer is always referring to the sound coming from the server and as you are not sending anything the mute button has nothing to mute.

I think Me and mulyaj just fell through an issue @gilgongo mentioned (mixing board analogy).

I'm using 3.6.2git (not sure which commit I used to compile it).

chrisrimple:
I've just tested this on 3.6.2, on a server for which I have 200ms Overall Delay...

1. MUTE MYSELF = false and MUTE = false.  I DO hear myself returned from the server.
2. MUTE MYSELF = false and MUTE = true.  I DO NOT hear myself returned from the server.
3. MUTE MYSELF = true and MUTE = false.  I DO NOT hear myself returned from the server.  I DO hear myself locally.
4. MUTE MYSELF = true and MUTE = true.  I DO NOT hear myself returned from the server.  I DO NOT hear myself locally.  In other words, I hear nothing.

I believe that all above is "as expected".

ann0see:
Ok. Just trying to compile the latest version from master. Will reverify if this "solves" the issue.

ann0see:
I could now also reproduce the other behaviour: Click on mute and then on mute myself.

pljones:
Does the order of doing it matter?  i.e. If you had "Mute myself" on when you clicked mixer mute, is it the same effect as if you had mixer mute on and clicked "Mute myself"?

[...]

Issue #797, opened 2020-12-27
Highlight choir singer that is creating feedback


kezzy1966:
Please could you consider introducing a new feature to  some how to put a mark a choir singer's slider who is creating feedback or their audio is distorted.

I know this is a lot to ask but Jamulus is now being used by choirs over 20-30+ singers. Some singers are not able to self diagnose there audio. It would be very handy if someone could identify them causing a problem and can flag up a mark,  '?' or 'x' that can appear on their slider. Others could then drop that slider.

I am very aware Jamulus was not initially designed for the large choir market and what it can do is quite amazing.

chrisrimple:
Seems like this could be similar to how Mute works.  Maybe there's a "Mark" button for every fader, and anyone clicking it causes all clients (on that server) to show that fader's name in bright red (or similar).  "Mark" settings are NOT retained between sessions and not saved in the INI file (unlike Mute and Solo).  Disconnecting from a server or quitting Jamulus should clear all "Mark" settings.

ann0see:
The question is how you detect feedback. Just by clipping? We could maybe check if a user is clipping all the time and then highlight the slider/name?

kezzy1966:
Detection of feedback is normally by luck, trial and error. Especially on big choir sessions. Perhaps wait for a quiet bit and see what activity line/fader is still showing some audio. Then push the slider up of that person. It can be difficult thought. The point is not everyone in a choir can do that but someone not singing / tech support can do and flag it up.

Perhaps have a 'Britain's got talent' style 3x's and out ! 
...joking of course ;-)

ann0see:
Hmm. But how can Jamulus (Software) do this or would you like a flag named "clipping" set by somebody?

corrados:
Jamulus supports a clipping indicator already. If a musician has a loud feedback, the clipping indicator should be triggered and it should be easy to identify the channel.

kezzy1966:
Hi Corrados, 

I'm not talking about clipping as that should be immediately fairly obvious by seeing their vu meter line shooting up. 

I more concerned with someone causing 'bubbling' and general distortion to their audio. Perhaps due to their 'frames' not set correctly causing xruns or jitter buffers being set too low. 
It is not immediately obvious to pin point the individual - even on a small group.   

Still a great system though.

ann0see:
Ah. So it's not feedback in terms of "microphone records sound of speaker" but "this person has audio issues". I think we had a discussion about that somewhere #762

[...]

PR #1179, opened 2021-03-05
Basic audio feedback detection

JohannesBrx:
This draft PR is based on discussion #1048 and provides a very basic audio feedback protection. Feedback is detected for an interval of (currently) 3 seconds after connecting to a server and utilizes the input meter level. In case feedback is detected, the following happens:

- activate `mute myself` (protect other users)
- mute the user's channel (protect user)
- show a message box with an explanation (see below)

```
Audio feedback detected.
We muted your channel and activated 'Mute Myself'. Please
solve the feedback issue first and unmute yourself afterwards.
Did you plug-in earphones?
```

While this basically works, there are several things to be discussed:

- Shall we use more sophisticated, yet performant algorithms to detect feedback? (e.g., [this](https://dsp.stackexchange.com/a/3086))
- Feedback only occurs when there is an initial signal. When the user is silent after connecting and starts speaking several seconds later, feedback protection might already have been deactivated. Possible solution: Wait for first audio signal above threshold and check next seconds for audio feedback.
- How shall feedback protection be deactivated -- in the settings, via command line or directly via the message box?

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -1146,7 +1146,7 @@ void CClientDlg::OnSoundDeviceChanged ( QString strError )
         TimerCheckAudioDeviceOk.start ( CHECK_AUDIO_DEV_OK_TIME_MS );
     }
 
-    if ( TimerDetectFeedback.isActive() )
+    if ( pSettings->bEnableFeedbackDetection && TimerDetectFeedback.isActive() )
     {
         TimerDetectFeedback.start ( DETECT_FEEDBACK_TIME_MS );
         bDetectFeedback = true;
@@ -1205,10 +1205,13 @@ void CClientDlg::Connect ( const QString& strSelectedAddress,
         TimerBuffersLED.start         ( BUFFER_LED_UPDATE_TIME_MS );
         TimerPing.start               ( PING_UPDATE_TIME_MS );
         TimerCheckAudioDeviceOk.start ( CHECK_AUDIO_DEV_OK_TIME_MS ); // is single shot timer
-        TimerDetectFeedback.start     ( DETECT_FEEDBACK_TIME_MS ); // single shot timer
 
-        // enable audio feedback detection
-        bDetectFeedback = true;
+        // audio feedback detection
+        if ( pSettings->bEnableFeedbackDetection )
+        {
+            TimerDetectFeedback.start     ( DETECT_FEEDBACK_TIME_MS ); // single shot timer
+            bDetectFeedback = true;
+        }
     }
 }
 
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -1032,16 +1032,35 @@ void CClientDlg::OnTimerSigMet()
     lbrInputLevelL->SetValue ( pClient->GetLevelForMeterdBLeft() );
     lbrInputLevelR->SetValue ( pClient->GetLevelForMeterdBRight() );
 
-    if ( bDetectFeedback && pClient->GetLevelForMeterdBLeft() > 7.5 )
+    if ( bDetectFeedback &&
+        ( pClient->GetLevelForMeterdBLeft() > NUM_STEPS_LED_BAR - 0.5  ||
+        pClient->GetLevelForMeterdBRight() > NUM_STEPS_LED_BAR - 0.5 ) )
     {
         // mute locally and mute channel
         chbLocalMute->setCheckState ( Qt::Checked );
         MainMixerBoard->MuteMyChannel();
 
-        QMessageBox::warning ( this, APP_NAME, tr ( "Audio feedback detected.\n"
+        // show message box about feedback issue
+        QCheckBox* chb = new QCheckBox ( tr ( "Enable feedback detection" ) );
+        chb->setCheckState ( pSettings->bEnableFeedbackDetection ?
+            Qt::Checked : Qt::Unchecked );
+        QMessageBox msgbox;
+        msgbox.setText ( tr ( "Audio feedback detected.\n\n"
             "We muted your channel and activated 'Mute Myself'. Please "
             "solve the feedback issue first and unmute yourself afterwards. "
             "Did you plug-in earphones?") );
+        msgbox.setIcon ( QMessageBox::Icon::Warning );
+        msgbox.addButton ( QMessageBox::Ok );
+        msgbox.setDefaultButton ( QMessageBox::Ok );
+        msgbox.setCheckBox ( chb );
+
+        QObject::connect( chb, &QCheckBox::stateChanged, [this](int state)
+            {
+                ClientSettingsDlg.SetEnableFeedbackDetection ( state == Qt::Checked );
+            }
+        );
+
+        msgbox.exec();
     }
 }
 
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -212,6 +212,7 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     // setup timers
     TimerCheckAudioDeviceOk.setSingleShot ( true ); // only check once after connection
+    TimerDetectFeedback.setSingleShot ( true );
 
 
     // Connect on startup ------------------------------------------------------
@@ -393,6 +394,9 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     QObject::connect ( &TimerCheckAudioDeviceOk, &QTimer::timeout,
         this, &CClientDlg::OnTimerCheckAudioDeviceOk );
+
+    QObject::connect ( &TimerDetectFeedback, &QTimer::timeout,
+        this, &CClientDlg::OnTimerDetectFeedback );
 
     QObject::connect ( sldAudioReverb, &QSlider::valueChanged,
         this, &CClientDlg::OnAudioReverbValueChanged );
@@ -1027,6 +1031,18 @@ void CClientDlg::OnTimerSigMet()
     // show current level
     lbrInputLevelL->SetValue ( pClient->GetLevelForMeterdBLeft() );
     lbrInputLevelR->SetValue ( pClient->GetLevelForMeterdBRight() );
+
+    if ( bDetectFeedback && pClient->GetLevelForMeterdBLeft() > 7.5 )
+    {
+        // mute locally and mute channel
+        chbLocalMute->setCheckState ( Qt::Checked );
+        MainMixerBoard->MuteMyChannel();
+
+        QMessageBox::warning ( this, APP_NAME, tr ( "Audio feedback detected.\n"
+            "We muted your channel and activated 'Mute Myself'. Please "
+            "solve the feedback issue first and unmute yourself afterwards. "
+            "Did you plug-in earphones?") );
+    }
 }
 
 void CClientDlg::OnTimerBuffersLED()
@@ -1105,6 +1121,11 @@ void CClientDlg::OnTimerCheckAudioDeviceOk()
     }
 }
 
+void CClientDlg::OnTimerDetectFeedback()
+{
+    bDetectFeedback = false;
+}
+
 void CClientDlg::OnSoundDeviceChanged ( QString strError )
 {
     if ( !strError.isEmpty() )
@@ -1123,6 +1144,12 @@ void CClientDlg::OnSoundDeviceChanged ( QString strError )
     if ( TimerCheckAudioDeviceOk.isActive() )
     {
         TimerCheckAudioDeviceOk.start ( CHECK_AUDIO_DEV_OK_TIME_MS );
+    }
+
+    if ( TimerDetectFeedback.isActive() )
+    {
+        TimerDetectFeedback.start ( DETECT_FEEDBACK_TIME_MS );
+        bDetectFeedback = true;
     }
 
     // update the settings dialog
@@ -1178,6 +1205,10 @@ void CClientDlg::Connect ( const QString& strSelectedAddress,
         TimerBuffersLED.start         ( BUFFER_LED_UPDATE_TIME_MS );
         TimerPing.start               ( PING_UPDATE_TIME_MS );
         TimerCheckAudioDeviceOk.start ( CHECK_AUDIO_DEV_OK_TIME_MS ); // is single shot timer
+        TimerDetectFeedback.start     ( DETECT_FEEDBACK_TIME_MS ); // single shot timer
+
+        // enable audio feedback detection
+        bDetectFeedback = true;
     }
 }
 
@@ -1212,6 +1243,8 @@ void CClientDlg::Disconnect()
     TimerBuffersLED.stop();
     TimerPing.stop();
     TimerCheckAudioDeviceOk.stop();
+    TimerDetectFeedback.stop();
+    bDetectFeedback = false;
 
 
 // TODO is this still required???
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -1045,7 +1045,7 @@ void CClientDlg::OnTimerSigMet()
         chb->setCheckState ( pSettings->bEnableFeedbackDetection ?
             Qt::Checked : Qt::Unchecked );
         QMessageBox msgbox;
-        msgbox.setText ( tr ( "Audio feedback detected.\n\n"
+        msgbox.setText ( tr ( "Audio feedback or loud signal detected.\n\n"
             "We muted your channel and activated 'Mute Myself'. Please "
             "solve the feedback issue first and unmute yourself afterwards. "
             "Did you plug-in earphones?") );
@@ -1228,7 +1228,7 @@ void CClientDlg::Connect ( const QString& strSelectedAddress,
         // audio feedback detection
         if ( pSettings->bEnableFeedbackDetection )
         {
-            TimerDetectFeedback.start     ( DETECT_FEEDBACK_TIME_MS ); // single shot timer
+            TimerDetectFeedback.start ( DETECT_FEEDBACK_TIME_MS ); // single shot timer
             bDetectFeedback = true;
         }
     }
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -1046,9 +1046,8 @@ void CClientDlg::OnTimerSigMet()
             Qt::Checked : Qt::Unchecked );
         QMessageBox msgbox;
         msgbox.setText ( tr ( "Audio feedback or loud signal detected.\n\n"
-            "We muted your channel and activated 'Mute Myself'. Please "
-            "solve the feedback issue first and unmute yourself afterwards. "
-            "Did you plug-in earphones?") );
+            "We muted your channel and activated 'Mute Myself'. Please solve "
+            "the feedback issue first and unmute yourself afterwards." ) );
         msgbox.setIcon ( QMessageBox::Icon::Warning );
         msgbox.addButton ( QMessageBox::Ok );
         msgbox.setDefaultButton ( QMessageBox::Ok );
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -1054,11 +1054,8 @@ void CClientDlg::OnTimerSigMet()
         msgbox.setDefaultButton ( QMessageBox::Ok );
         msgbox.setCheckBox ( chb );
 
-        QObject::connect( chb, &QCheckBox::stateChanged, [this](int state)
-            {
-                ClientSettingsDlg.SetEnableFeedbackDetection ( state == Qt::Checked );
-            }
-        );
+        QObject::connect ( chb, &QCheckBox::stateChanged,
+            this, &CClientDlg::OnFeedbackDetectionChanged );
 
         msgbox.exec();
     }
--- a/src/clientsettingsdlgbase.ui
+++ b/src/clientsettingsdlgbase.ui
@@ -1135,6 +1135,33 @@
                </property>
               </spacer>
              </item>
+             <item>
+              <widget class="QLabel" name="lblDetectFeedback">
+               <property name="text">
+                <string>Feedback Protection</string>
+               </property>
+              </widget>
+             </item>
+             <item>
+              <widget class="QCheckBox" name="chbDetectFeedback">
+               <property name="text">
+                <string>Enable</string>
+               </property>
+              </widget>
+             </item>
+             <item>
+              <spacer name="verticalSpacer_15">
+               <property name="orientation">
+                <enum>Qt::Vertical</enum>
+               </property>
+               <property name="sizeHint" stdset="0">
+                <size>
+                 <width>20</width>
+                 <height>40</height>
+                </size>
+               </property>
+              </spacer>
+             </item>
             </layout>
            </item>
            <item>
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,9 @@
 - Automatic channel fader adjustment simplifies mixer setup by using the channel level meters (#1071).
   (contributed by @JohannesBrx)
 
+- Added basic audio feedback detection (#1179).
+  (contributed by @JohannesBrx)
+
 - The mute button is now also shown in mono mode (#1074)
   (contributed by @npostavs)
 
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1726,3 +1726,8 @@ void CAudioMixerBoard::SetChannelLevels ( const CVector<uint16_t>& vecChannelLev
         }
     }
 }
+
+void CAudioMixerBoard::MuteMyChannel()
+{
+    SetFaderIsMute ( iMyChannelID, true );
+}
--- a/src/audiomixerboard.h
+++ b/src/audiomixerboard.h
@@ -201,6 +201,7 @@ public:
     void            SetPanIsSupported();
     void            SetRemoteFaderIsMute ( const int iChannelIdx, const bool bIsMute );
     void            SetMyChannelID ( const int iChannelIdx ) { iMyChannelID = iChannelIdx; }
+    int             GetMyChannelID() const { return iMyChannelID; }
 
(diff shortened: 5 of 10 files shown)


ann0see:
Thanks for this PR! I'll test it right away! 

> How shall feedback protection be deactivated -- in the settings, via command line or directly via the message box?

I think settings & a button to settings in the message would be good. Especially once a new UI (as DonC) proposes comes up, it should be easy to add it in the audio tab. Not sure about a CLI argument?

> Feedback only occurs when there is an initial signal. When the user is silent after connecting and starts speaking several seconds later, feedback protection might already have been deactivated.

I'd not really consider this case. I'd say that nearly always the noise is loud enough?

src/clientdlg.cpp
```diff
@@ -1094,6 +1098,18 @@ void CClientDlg::OnTimerSigMet()
...
     lbrInputLevelL->SetValue ( pClient->GetLevelForMeterdBLeft() );
     lbrInputLevelR->SetValue ( pClient->GetLevelForMeterdBRight() );
+
+    if ( bDetectFeedback && pClient->GetLevelForMeterdBLeft() > 7.5 )
+    {
+        // mute locally and mute channel
+        chbLocalMute->setCheckState ( Qt::Checked );
+        MainMixerBoard->MuteMyChannel();
+
+        QMessageBox::warning ( this, APP_NAME, tr ( "Audio feedback detected.\n"
```
ann0see, line 1108: Open discussion here: Shouldn't the MUTED label be used instead of a real warning message?

ann0see:
I've stopped the CI to get the artifacts. I'll restart it soon.

ann0see:
Just tested it and I think a message like that is clear. I'd add a checkbox/button there which disables feedback detection. It might be a bit sensitive. Also it doesn't always seem to activate the mute button in the own mix (especially after disabling it once the feedback detection already ran before and you un-muted yourself.)

JohannesBrx:
> Just tested it and I think a message like that is clear. I'd add a checkbox/button there which disables feedback detection. It might be a bit sensitive. 

I think I'll take the checkbox.

> Also it doesn't always seem to activate the mute button in the own mix (especially after disabling it once the feedback detection already ran before and you un-muted your own slider.)

Currently, feedback detection only runs for 3 seconds after a new connection. If feedback occurs later, it will not be detected.

gene96817:
* How is the feedback detection implemented? 
* In the fullest context, feedback and echo cancellation are the same phenomena. The gain in the echo path could create runaway feedback (from positive gain) or reverb (negative gain). It might make sense to use some open source echo cancellation code for our feedback detect logic. (Just don't put it in line with our normal sound processing.)
* A channel could start out stable until the total volume (from the session mix) in the echo path gets loud enough to trigger the runaway feedback.

JohannesBrx:
> How is the feedback detection implemented?

The current implementation is very basic and just checks the input level for high amplitudes. However, this might already catch many cases of feedback.

> In the fullest context, feedback and echo cancellation are the same phenomena. The gain in the echo path could create runaway feedback (from positive gain) or reverb (negative gain). It might make sense to use some open source echo cancellation code for our feedback detect logic. 

Do you already have open source code in mind where to look at?

> (Just don't put it in line with our normal sound processing.)

If the code is based on processing audio samples, which alternative do we have?

> A channel could start out stable until the total volume (from the session mix) in the echo path gets loud enough to trigger the runaway feedback.

This is what I meant in my initial message. We would probably have to activate the feedback detection for a longer time to catch these cases.

gene96817:
@JohannesBrx 

> Do you already have open source code in mind where to look at?
Sorry I have not been following codec development closely. However I see that echo cancellation has become a common feature with voice codecs. We could use any that seems to be lightweight.

> If the code is based on processing audio samples, which alternative do we have?
Think of the code as a monitoring tap. Just take the audio stream to feed into this code and the only output is the control decision. Then this code does not insert any delay, though the decision does have a delay from the triggering event. 

> We would probably have to activate the feedback detection for a longer time to catch these cases.
The implication is you need the monitoring to always be running. As the mean volume of the full mix increases, the feedback condition could be triggered. We don'/t need the monitoring to be literally continuous. It is possible the monitoring is scanning (round-robin) all the channels or feedback. This kind of scanning would add minimal computing load.

pljones:
A lot of the suggested improvements are valid but an initial implementation is always open to future adaptation.  Getting something now is actually more useful that getting nothing now so that something with more extensive features is available later.  In waiting we may even find hurdles that do not currently block delivery that delay things further.

So I'd rather this change was not diverted to far from its initial goals.  Keep it simple.

src/clientdlg.cpp
JohannesBrx: I suspect that unexperienced user will never discover the *cause* of muting without an explicit warning message.

ann0see: Probably true. I'll leave this open for other thoughts. By testing it I'd say that the current behavior is better.

JohannesBrx:
For me, there some open questions we could discuss:

1) Shall we mention in the warning message that the cause of the detected feedback could also be a signal being too loud?
(Remember that the current implementation would also trigger on loud signals in the first 3 seconds.) Something like "Audio feedback or **loud/clipped signal** detected. We muted..." (see initial comment).

2) Shall the feedback detection end 3 seconds after *connecting* or 3 seconds after the *first audio signal* above a certain threshold? 
In my experiments I was able to provoke a situation where I was silent for several seconds and triggered the audio feedback with the first spoken word. On the other hand, the latter method would probably be a problem when the user waits for a  loud entry with, e.g., an electric guitar. I tend to stop detection 3 seconds after *connecting*, because it is a more comprehensible behavior.

3) Shall feedback detection be reactivated for 3 seconds after un-muting or not?
I tend to *not* reactivate it since the user is aware of the mute button and may push it again to stop feedback.

4) Shall the feedback detection only be triggered when the signal is clipped or already at high levels?
I am not completely sure whether audio feedback always produces clipping. In every case I would like to use the results of the `SignalLevelMeter` of `CClient`, such that there is no additional audio processing required. (Not sure whether I would get clipping results here.)

ann0see:
> Shall we mention in the warning message that the cause of the detected feedback could also be a signal being too loud?
(Remember that the current implementation would also trigger on loud signals in the first 3 seconds.) Something like "Audio feedback or loud/clipped signal detected. We muted..." (see initial comment).

I'd say yes since that's what you check.

>  2. [...] On the other hand, the latter method would probably be a problem when the user waits for a loud entry with, e.g., an electric guitar. I tend to stop detection 3 seconds after connecting, because it is a more comprehensible behavior.

Agree.

 > 3. [...] I tend to not reactivate it since the user is aware of the mute button and may push it again to stop feedback.

Yes. I'd find it very confusing if feedback detection started right after a un-mute.

4. The easiest way would be how you implemeted it. I'd say that clipping always occurs, but I don't know it for sure.

DavidSavinkoff:
Instability (screechin' feedback) is possible when the ouput is louder than the input because you didn't attenuate (or invert) the output signal before sending it back to the input.

With this in mind, I propose that the 'mute myself' be activated:
- if the user's signal going to the server is >= +6dB (that is 6 decibels higher than nominal. Jamulus Must(is required to) declare an absolute 0dB level that can be achieved by attenuating a respective slide-pot)
- if the user's signal going to the server is >= +12dB than the setting on the respective slide-pot. ++1 ++2

++1 Since feedback gets louder within the linear range of Jamulus, the 'mute myself' button will get tripped.
++2 Note that digital clipping must be more than +12dB greater than the estabilished 0dB level.

pljones:
@ DavidSavinkoff - we don't know the level in dB.  We have "clipping" and below.

DavidSavinkoff:
Since dB is relative, clipping could be defined as how much headroom you think should be above a certain level. The certain level is defined as 0dB, then the headroom can be expressed in dB.
Setting the 0dB level as; 6dB less than clipping is reasonable. Note that 6dB is twice the voltage of 0dB.

using base 10 logarithms:
dB = 20log(Vout/Vin) :: negative dB is attenuation (potentiometer)
Vout/Vin = gain :: attenuation is where gain < 1 (potentiometer)
antilog(dB/20) = gain

JohannesBrx:
@DavidSavinkoff Do you think your threshold approach would also work without a calibration of the overall sound system? Wouldn't the (externally regulated) volume of the speakers have to be taken into account? If I turn down the external volume, the feedback level will be modified.

DavidSavinkoff:
@JohannesBrx The threshold is designed as relative so it is not an absolute theshold. The only requirement for this method to work is that you have enough attenuation (by muting or using the slide-pot) to stop the feedback.
What happens with this method, is feedback can build up to either the threshold of clipping or +12dB above the slide-pot setting (whichever is less) before the Mute is activated. Feedback will usually build up quickly and be muted quickly. If your PA speakers are turned up louder, the feedback will be louder but Jamulus will mute the feedback before Jamulus goes into digital clipping. If a microphone gain is increased, the feedback will not increase louder at the PA speakers.
Note that there are 2 places where feedback is muted. One is to catch all, the other straddles the slide-pot so that the feedback volume is relative to the slide-pot setting.

JohannesBrx:
@DavidSavinkoff 
> What happens with this method, is feedback can build up to either the threshold of clipping or +12dB above the slide-pot setting (whichever is less) before the Mute is activated.

So when I put my own channel fader (or *slide-pot*) in the mixer to -20dB you would trigger the feedback detection at -8dB microphone input level, right?

Is the +12dB margin (factor 4) physically founded or just a heuristic setting?

What is the aim of using the +12dB relative threshold? That we would catch the feedback earlier (before clipping)?

What happens when I have turned down my channel fader for some reason to, e.g., -30dB. When I speak a bit louder into the microphone -- let's say the level goes beyond -10dB -- would the channel already be muted due to the feedback detection?

DavidSavinkoff:
> @DavidSavinkoff
> 
> > What happens with this method, is feedback can build up to either the threshold of clipping or +12dB above the slide-pot setting (whichever is less) before the Mute is activated.
> 
> So when I put my own channel fader (or _slide-pot_) in the mixer to -20dB you would trigger the feedback detection at -8dB microphone input level, right?
> 
Correct. This allows 12dB of headroom above the slide-pot setting (since the slide-pot is already attenuating)
> Is the +12dB margin (factor 4) physically founded or just a heuristic setting?
> 
You may have to experiment (+20dB ?). You don't want nuisance muting on the peaks here.
> What is the aim of using the +12dB relative threshold? That we would catch the feedback earlier (before clipping)?
> 
The muting is actually triggered by a signal that is too large, and this stops the feedback.
This is an opportunity to stop the feedback even sooner without compromise.
Note that the feedback is guaranteed by its increasing nature to go all the way to clipping.
> What happens when I have turned down my channel fader for some reason to, e.g., -30dB. When I speak a bit louder into the microphone -- let's say the level goes beyond -10dB -- would the channel already be muted due to the feedback detection?
> 
Yes. This is why I suggested that this should be more than 6dB.
An example: If you set your microphone to record your singing voice, you must consider how much louder you may sing. I simply mentioned +12dB, but maybe you can go +20dB louder.

pljones:
> the other straddles the slide-pot so that the feedback volume is relative to the slide-pot setting.

So if my input isn't feeding my output (I've muted my server mix channel), having other people's output feeding back to my input can't cause a feedback loop... but it will still be annoying for everyone else on the server (the "echo effect").

> The only requirement for this method to work is that you have enough attenuation (by muting or using the slide-pot) to stop the feedback.

And the best attenuation is "Mute myself", as it stops the "echo effect", too.

DavidSavinkoff:
> > the other straddles the slide-pot so that the feedback volume is relative to the slide-pot setting.
> 
> So if I've muted my server mix channel, having other people's output feeding back to my input can't cause a feedback loop... but it will still be annoying for everyone else on the server (the "echo effect").
> 
When I wrote this, I was thinking only of using the 'Mute myself'. However, there are probably more possibilities depending on what you do or mute.
> > The only requirement for this method to work is that you have enough attenuation (by muting or using the slide-pot) to stop the feedback.
> 
> And the best attenuation is "Mute myself", as it stops the "echo effect", too.

JohannesBrx:
The other day in a choir rehearsal: Someone joins in and gets louder and louder ... mute! Luckily he had a custom Jamulus version with the audio feedback detection :-) The reason was that he forgotten to plug in his earphones...

ann0see:
I think once 3.7.0 is out, this one is near to ready?

gilgongo:
@JohannesBrx is https://github.com/jamulussoftware/jamulus/issues/797 needed now? If the client can handle its own feedback, it feels hard to know whether messaging that to others would be useful or not.

pljones:
> 
> 
> @JohannesBrx is #797 needed now? If the client can handle its own feedback, it feels hard to know whether messaging that to others would be useful or not.

Also note that if we implement the feature that was requested to notify others on local muting being active (i.e. "has self muted", in addition to the existing "has you muted" icon), this might be enough, too.

JohannesBrx:
> I think once 3.7.0 is out, this one is near to ready?

@ann0see I would say so. The principle is very simple but I think it will work in practice to avoid most of the feedback cases.

> @JohannesBrx is #797 needed now? If the client can handle its own feedback, it feels hard to know whether messaging that to others would be useful or not.

@gilgongo I think the message (see below) shown to the user after feedback has been detected is self-explanatory enough. If the audio setup is wrong, it will be difficult to get help via a Jamulus audio conversation anyway. (What shall the others do when the person causing the feedback problem cannot speak due to audio setup problems?)
```
Audio feedback or loud signal detected.
We muted your channel and activated 'Mute Myself'. Please
solve the feedback issue first and unmute yourself afterwards.
Did you plug-in earphones?
```

gene96817:
>  If the client can handle its own feedback, it feels hard to know whether messaging that to others would be useful or not.

We are providing a solution for "near-end" feedback. This can be a pure client issue. 

> We muted your channel and activated 'Mute Myself'.

The client could say "Jamulus has activated "Mute Myself" to prevent sending feedback to the session." Then rest of the advisory is helfpul.

ann0see:
@JohannesBrx coould you please check what happens if you unmute your own slider, disconnect and afterwards trigger feedback detection? I’m not quite sure, but I‘m quite sure there’s a bug not muting the own slider somewhere.

JohannesBrx:
> @JohannesBrx coould you please check what happens if you unmute your own slider, disconnect and afterwards trigger feedback detection? I‘m quite sure there’s a bug not muting the own slider somewhere.

@ann0see I tried to reproduce this several times but my own slider has always correctly been muted. I have tested multiple variants using a local server and a remote one. In addition to that, I checked with and without _mute myself_ activated.

When feedback has been detected, the following method is called. I currently have no idea what could go wrong here.
```
void CAudioMixerBoard::MuteMyChannel()
{
    SetFaderIsMute ( iMyChannelID, true );
}
```

> you unmute your own slider, disconnect and afterwards trigger feedback detection?

Just to get sure: Did you _connect_ again before triggering feedback detection? Detection is only active for 3 seconds after connecting.

ann0see:
Yes, I did disconnect. The feedback detection did fire but I wasn't muted. I found the bug on macOS Big Sur.

JohannesBrx:
> Yes, I did disconnect. The feedback detection did fire but I wasn't muted. I found the bug on macOS Big Sur.

Strange. I guess your channel's `Mute` checkbox is also still unchecked, right? (Theoretically it would be possible that the checkbox is checked but the signal for muting the channel is not sent to the server.)

There's a comment in the source which says that changing the checkbox state automatically emits the signal to the server for muting the channel: 

```
void CChannelFader::SetFaderIsMute ( const bool bIsMute )
{
    // changing the state automatically emits the signal, too
    pcbMute->setChecked ( bIsMute );
}
```

ann0see:
Hmm. Maybe it's https://github.com/jamulussoftware/jamulus/issues/909

pljones:
> Maybe. But it could also be #909

Unless there's an error routing the Qt UI signal to the appropriate client slot, the code paths are totally separate.

For "Mute Myself", the client sends silence and the server has no idea anything happened.

For channel "Mute", the client sends a signal to the server and the server skips including the muted channel (whichever it is) in the mix returned.

ann0see:
Maybe this needs some more testing on another OS. I'll try it on debian 10.

ann0see:
Bug to reproduce:

1. Connect to a server while Mute myself is enabled
2. Start feedback detection
3. Hear yourself

JohannesBrx:
> Ok. Did some tests (which produced feedback) and unfortunately it didn't detect the feedback  3s are probably not enough? Maybe an other approach (exponential volume increase) would detect feedback more reliably?

Could you describe the chronological sequence of the feedback you produced? (Or even share an audio sample?) Did the feedback evolve right from the beginning of the connection and how fast did it develop (e.g, until clipping)?

ann0see:
Strange. After trying to reproduce it (full gain of headphones and microphone) the feedback detection fired directly.
Yesterday, I started Jamulus, put my headphones quite close to my microphone and waited for feedback. Yes, it didn't start directly, but quite fast. I'd say that this way of implementing feedback detection does prevent some cases, but not all. Merging this would be a good step forward since I don't think it harms/annoy anyone (you can always disable it)?

ann0see:
Just requesters review from two other main devs to hear their opinion if this can be merged as is.

src/clientdlg.cpp
```diff
@@ -1102,6 +1106,37 @@ void CClientDlg::OnTimerSigMet()
...
+        msgbox.setText ( tr ( "Audio feedback or loud signal detected.\n\n"
+            "We muted your channel and activated 'Mute Myself'. Please "
+            "solve the feedback issue first and unmute yourself afterwards. "
+            "Did you plug-in earphones?") );
+        msgbox.setIcon ( QMessageBox::Icon::Warning );
+        msgbox.addButton ( QMessageBox::Ok );
+        msgbox.setDefaultButton ( QMessageBox::Ok );
+        msgbox.setCheckBox ( chb );
+
+        QObject::connect( chb, &QCheckBox::stateChanged, [this](int state)
```
pljones, line 1132: Please stick to the old-style connects, even if it means more code, for now, as you've done at line 439.

src/clientsettingsdlgbase.ui
```diff
@@ -482,6 +489,25 @@
           <item>
            <widget class="CLanguageComboBox" name="cbxLanguage"/>
           </item>
+          <item>
```
pljones, line 492: You need to add a tabstop (around line 724).

JohannesBrx: I have rebased my code, now including #1415. In the new tabbed settings, a tabstop seems not to be required any more.

src/clientdlg.cpp
JohannesBrx: Done :)

JohannesBrx:
@pljones @ann0see 
After rebasing, I noticed the new tabbed settings (#1415) and rearranged the checkbox for feedback protection. I have now placed it under under `Input Boost` in the `Advanced Setup` panel.
image

src/clientdlg.cpp
```diff
@@ -1031,6 +1035,34 @@ void CClientDlg::OnTimerSigMet()
...
+
+        // show message box about feedback issue
+        QCheckBox* chb = new QCheckBox ( tr ( "Enable feedback detection" ) );
+        chb->setCheckState ( pSettings->bEnableFeedbackDetection ?
+            Qt::Checked : Qt::Unchecked );
+        QMessageBox msgbox;
+        msgbox.setText ( tr ( "Audio feedback or loud signal detected.\n\n"
+            "We muted your channel and activated 'Mute Myself'. Please "
+            "solve the feedback issue first and unmute yourself afterwards. "
+            "Did you plug-in earphones?") );
```
hoffie, line 1055: As the feedback detection only triggers upon connect, should we advise users to disconnect and try again here?

hoffie (commented):
This looks really good to me. Thanks for working on this!

I have done two tests:
1. Connect a Jamulus client to a server and feed audio via mplayer into it. Connect a second client (built from this PR) which has deliberate feedback via qjackctl. The feedback is properly detected.
2. Connect a Jamulus client (built from this PR) with deliberate feedback to an empty server and create some initial audio. The feedback detection does not trigger reliably. I suspect that this is due to the fact that it takes some time (probably more than the 3s) to reach a high level.

Not sure if (2) is supposed to be covered. I think it would be helpful and it seems like it may get better if we increased it to 5000ms?

src/clientdlg.cpp
JohannesBrx: I would not count on unexperienced users disconnecting and connecting again. Although it would achieve the goal, it would not be particularly intuitive.

Initially I thought about re-triggering the feedback detection interval after un-muting -- that would be a reliable technical solution.

JohannesBrx:
> I have done two tests:
> 
> 1. Connect a Jamulus client to a server and feed audio via mplayer into it. Connect a second client (built from this PR) which has deliberate feedback via qjackctl. The feedback is properly detected.
> 2. Connect a Jamulus client (built from this PR) with deliberate feedback to an empty server and create some initial audio. The feedback detection does not trigger reliably. I suspect that this is due to the fact that it takes some time (probably more than the 3s) to reach a high level.

Does the *initial audio* in (2) start directly after connecting to the server?

I understand the difference between your tests as follows: In (1), the first client using mplayer provokes the feedback for the second client. In (2), the only available client should provoke the feedback itself.

> Not sure if (2) is supposed to be covered. I think it would be helpful and it seems like it may get better if we increased it to 5000ms?

You could even wait 10 seconds before speaking anything in the microphone and possibly provoke feedback. I would rather regards this as an artificial case.

I thought about waiting for the first distinctive audio signal before starting the feedback detection interval. However, this would probably make things more complicated and less intuitive without catching much more feedback cases.

src/clientdlg.cpp
hoffie: I'm not saying we have to do it at all, I was just thinking that adjusting the text might direct users in a good direction without any more technical effort.
Re-triggering the feedback detection after un-mutung would also work for me but would need additional logic (even if not much).

hoffie:
> Does the initial audio in (2) start directly after connecting to the server?

More or less. This was a manual test, so I've attempted to make a noise directly after connecting, but I can't reliably say how much time it took (probably still <1s).

> I understand the difference between your tests as follows: In (1), the first client using mplayer provokes the feedback for the second client. In (2), the only available client should provoke the feedback itself.

Yes, right.

> You could even wait 10 seconds before speaking anything in the microphone and possibly provoke feedback. I would rather regards this as an artificial case.

Hrm, right...

> I thought about waiting for the first distinctive audio signal before starting the feedback detection interval. However, this would probably make things more complicated and less intuitive without catching much more feedback cases.

I agree. So maybe we can merge this as-is and extend it later if it turns out that this would further improve this feature?

hoffie (approved):

JohannesBrx:
> > I thought about waiting for the first distinctive audio signal before starting the feedback detection interval. However, this would probably make things more complicated and less intuitive without catching much more feedback cases.
> 
> I agree. So maybe we can merge this as-is and extend it later if it turns out that this would further improve this feature?

Yes, that would definitely be an option.

src/clientsettingsdlgbase.ui
```diff
@@ -1083,6 +1083,33 @@
                </property>
               </spacer>
              </item>
+             <item>
+              <widget class="QLabel" name="lblDetectFeedback">
```
pljones, line 1087: There needs to be "What's this?" text for the label and the control along with accessibility text for the control.

hoffie:
This also needs a rebase, btw.

hoffie:
@JohannesBrx It would be cool to get this into 3.8.0. Do you have some time for the rebase and the accessibility updates (https://github.com/jamulussoftware/jamulus/pull/1179#discussion_r618578843) by chance?

That would be really cool as we're planning a feature freeze for tomorrow to be able to start with translations.

JohannesBrx:
> @JohannesBrx It would be cool to get this into 3.8.0. Do you have some time for the rebase and the accessibility updates ([#1179 (comment)](https://github.com/jamulussoftware/jamulus/pull/1179#discussion_r618578843)) by chance?

Yes, I think I will manage it tomorrow at the latest.

Do you think I should investigate re-activating the feedback detection when the user disables mute? (not sure how many changes this would imply)

hoffie:
> Yes, I think I will manage it tomorrow at the latest.

Great! :)

> Do you think I should investigate re-activating the feedback detection when the user disables mute? (not sure how many changes this would imply)

I'd suggest doing that in a separate PR, if we think it can help. This PR is almost done, has already been reviewed and I think we shouldn't risk introducing new questions which might prevent it from going into 3.8.0.

In my own experience, we had people with feedback issues so this PR would have certainly helped. I think feedback issues are way less common when the initial setup has been completed. Maybe the best idea would be to create a discussion thread some weeks after the 3.8.0 to ask people (especially those with larger groups) if and how feedback is still an issue?

JohannesBrx:
@hoffie @pljones I have now rebased and added the accessibility texts. I still have to add the release notes, which I will add later this day.

src/clientdlg.cpp
```diff
@@ -1027,6 +1031,34 @@ void CClientDlg::OnTimerSigMet()
...
```
pljones, line 1051: > Did you plug-in earphones?

I'm not sure this would be helpful.  Certainly with ASIO4ALL, it could have all outputs enabled so plugging in _head_phones wouldn't prevent feedback from built-in speakers, necessarily.  I've had to tell people to take off their headphones and play so that they believe that their speakers are still sounding...

pljones: In fact, I'd rather it pointed to the "Hints and Tips" page on the web site and _that_ actually had a "methods for resolving feedback" section.

JohannesBrx: Ok, I can follow this reasoning and have removed the hint. Once the website provides a section about resolving feedback, we could add a corresponding link.

pljones (approved):

PR #1194, opened 2021-03-07
windows/sound.cpp: Explicitly set driverInfo to 0

npostavs:
https://github.com/jamulussoftware/jamulus/pull/870#issuecomment-791970524

> I noticed when looking at the Jamulus' current ASIO code is that it doesn't initialize `driverInfo` before passing it to `ASIOInit`

From what I can tell looking at it in the debugger, it gets semi-accidentally set to zero by virtue of being allocated on fresh memory. It would be better to set it explicitly; IMO this could go in to the release.

### DIFF
--- a/windows/sound.cpp
+++ b/windows/sound.cpp
@@ -62,6 +62,11 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName,
 
     loadAsioDriver ( cDriverNames[iDriverIdx] );
 
+    // According to the docs, driverInfo.asioVersion and driverInfo.sysRef
+    // should be set, but we haven't being doing that and it seems to work
+    // okay...
+    memset ( &driverInfo, 0, sizeof driverInfo );
+
     if ( ASIOInit ( &driverInfo ) != ASE_OK )
     {
         // clean up and return error string


windows/sound.cpp
```diff
@@ -520,6 +520,11 @@ CSound::CSound ( void           (*fpNewCallback) ( CVector<int16_t>& psData, voi
     // init pointer to our sound object
     pSound = this;
 
+    // According to the docs, driverInfo.asioVersion and driverInfo.sysRef
```
pljones, line 523: If we're going to zero the whole structure, `LoadAndInitializeDriver` should be doing this just before
```
if ( ASIOInit ( &driverInfo ) != ASE_OK )
```
rather than here.

npostavs: Okay. (I thought it would be a smaller change to zero it once, but I guess it shouldn't make a difference)

pljones: Does it work moved..? :)

pljones (approved):

windows/sound.cpp
npostavs: Works for me...

pljones (approved):

softins (approved):

ann0see:
@softins @npostavs I think we should add you as contributor to the contributors list.

PR #1198, opened 2021-03-07
Update it.nsi

dzpex:

PR #1203, opened 2021-03-08
Tools: Improve Wininstaller translation checker

hoffie:
The checker will now detect if the macros are commented out.

https://github.com/jamulussoftware/jamulus/pull/1198#pullrequestreview-605873460

### DIFF
--- a/tools/check-wininstaller-translations.sh
+++ b/tools/check-wininstaller-translations.sh
@@ -11,7 +11,7 @@ for LANGUAGE_FILE in ${LANGUAGE_FILES}; do
     echo
     echo "* ${LANGUAGE_FILE}"
     echo -n "  - Checking language file is included in ${INSTALLERLNG}... "
-    if grep -qF $(tr '/' '\\' <<<"${LANGUAGE_FILE}") "${BASE_DIR}/${INSTALLERLNG}"; then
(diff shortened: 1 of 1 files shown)


ann0see (approved):
Great! Thanks!

tools/check-wininstaller-translations.sh
```diff
@@ -11,7 +11,7 @@ for LANGUAGE_FILE in ${LANGUAGE_FILES}; do
     echo
     echo "* ${LANGUAGE_FILE}"
     echo -n "  - Checking language file is included in ${INSTALLERLNG}... "
-    if grep -qF $(tr '/' '\\' <<<"${LANGUAGE_FILE}") "${BASE_DIR}/${INSTALLERLNG}"; then
+    if grep -q '^!include "\${ROOT_PATH}\\'$(tr '/' '\\' <<<"${LANGUAGE_FILE}" | sed -re 's|\\|\0\0|g')'"' "${BASE_DIR}/${INSTALLERLNG}"; then
```
softins, line 14: The effect is the same, but I would just use `$(sed -e 's|/|\\\\|g' <<<"${LANGUAGE_FILE}")`

softins (approved):

tools/check-wininstaller-translations.sh
softins: It's no big deal. Only because it would achieve with one command (`/` > sed > `\\`) what is currently done with two (`/` > tr > `\` > sed > `\\`), because tr isn't able to add or double characters.

hoffie: Can't believe how I've missed that, even after your comment. Thanks for explaining again, you're completely right. I will update this as part of an upcoming tools PR.

Issue #1191, opened 2021-03-07
Unexpected reset of mixer levels


hoffie:
**Describe the bug**

In a rehearsal with about 20 users which my client had seen before, I experienced unexpected resets of fader levels and mute state in my mix. I **think** this always happened when someone (re)connected. It always affected the same users and I suspect that mixer/mute state got restored to what was saved from a previous session.

**To Reproduce**

- Run a Jamulus server (should not matter if 3.6.2 or master)
- Connect Client A, B (based on current master code)
- In Client A's mix, mute B.
- Close Client A (to ensure that mixer levels are saved)
- Re-connect Client A
- In Client A's mix, unmute B.
- Connect Client C
- In Client A: Observe, that Client B gets set to muted again (to be verified)

The same happens to mixer levels.

**Expected behavior**

In Client A, B's mute state should not change unexpectedly.

**Operating system**

Arch Linux

**Version of Jamulus**

Current master (e7389e3479871828bb2bea6a5c1c5b74da61d8be) for both server and at least the problematic client.

**Additional context**

#1009 seems to be the culprit. My assumption there was that it would always be safe to load stored mixer settings. I suspect that this is not the case and that mixer settings are only stored on specific events (client exit?). If this is the case,
- we could revert #1009 and solely rely on #1010 to keep #955 fixed or
- we could ensure that mixer state is stored more regularly (e.g. any time something changes or timer-based).

I'd prefer the second option. I think it might even fix two other edge cases:
- If a Jamulus client exits unexpectedly, all changed mixer state might not get stored (even in 3.6.2).
- ~~If a Jamulus client sees a new user for the first time and the fader levels are adjusted, then they might not be stored immediately. If that new user disconnects and re-connects later, I suspect they will start with initial levels again.~~ Update: Fader settings are stored when a fader slot becomes unused: https://github.com/jamulussoftware/jamulus/blob/22d34ed989800ff3b9e8eddd7ec80bc9c3b6c218/src/audiomixerboard.cpp#L1219

#1010 may also trigger this issue, but I'd argue that this is not that bad. Without #1010, mixer state for all clients is wrong. With #1010, mixer state may still not be fully up-to-date, but it will at least be close to being up-to-date (I suspect mixer state will be the same as during the initial start of each Jamulus client).

- [x] Reproduce with master-level server and master-level, clean client code and adapt issue title.
- [ ] Decide on a fix (I'd prefer optimization of the fader storage logic)

[...]

PR #1010, opened 2021-02-17
Server: Send empty channel list on connect

hoffie:
This ensures that clients will reset their audio mixer board state upon reconnect. Otherwise, gain levels may not be sent reliably to the server.

Fixes #955.

Note: This is an alternative approach to #1009. Both PRs are compatible and can be merged together.

PR #1210, opened 2021-03-08
Revert "Client: Always update gain levels"

hoffie:
This reverts PR #1009 (f116811a4e9748722a2f1b8178ebfe68737b57d0) as it has unintended side effects (#1191). The Changelog is also updated accordingly.

Basically, that PR assumed that applying stored fader levels would always be a safe thing to do.
This is not the case, as changes in fader levels are currently not stored immediately, but rather in specific events only (e.g. disconnects).

PR #1009 was initially supposed to help with #955 as a client-side fix. With the revert, we still have the server-side fix from #1010.

It is planned to re-apply #1009 once fader-level storage has been fully understood and extended to cover all relevant places (#1191).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -49,8 +49,7 @@
   logic (#633, #1092), suggested by atsampson, coded by hoffie
 
 - bug fix: properly restore mixer levels after reconnecting to a restarted
-  server. updating the server or the client is sufficient to fix this. coded
-  by hoffie (#955, #1009, #1010)
+  server. This is a server-side fix. coded by hoffie (#955, #1010)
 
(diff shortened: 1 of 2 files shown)


softins (approved):
I guess the ChangeLog reference should also remove the reference to #1009, but that can be done separately if needed.

hoffie:
> I guess the ChangeLog reference should also remove the reference to #1009, but that can be done separately if needed.

Thanks for the pointer. Fixed.

There is always something left which can be missed... :)

ann0see (approved):
I'm happy to approve this since it seems to just revert @hoffie s commit.

Merging is up to so. else

PR #1222, opened 2021-03-10
Client: Add support for input gain control

hoffie:
This adds a new client-only UI setting to apply a gain value on all audio input.

This solves the problem that some (internal?) microphones cannot be made to output louder signals using operating system mechanisms (Windows sound settings). Multiple people in my choir use builds with this code included.

This approach solves the problem as close to the source and as simple as possible. I would have preferred to permit extending the fader range in the mixer board, but this would have been a way more intrusive change (introduction of new protocol messages, etc.).

Related to: https://github.com/jamulussoftware/jamulus/discussions/1030

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -51,6 +51,7 @@ CClient::CClient ( const quint16  iPortNumber,
     iAudioInFader                    ( AUD_FADER_IN_MIDDLE ),
     bReverbOnLeftChan                ( false ),
     iReverbLevel                     ( 0 ),
+    iInputBoost                      ( 1 ),
     iSndCrdPrefFrameSizeFactor       ( FRAME_SIZE_FACTOR_DEFAULT ),
     iSndCrdFrameSizeFactor           ( FRAME_SIZE_FACTOR_DEFAULT ),
     bSndCrdConversionBufferRequired  ( false ),
@@ -1053,6 +1054,16 @@ void CClient::ProcessAudioDataIntern ( CVector<int16_t>& vecsStereoSndCrd )
 
 
     // Transmit signal ---------------------------------------------------------
+
+    if ( iInputBoost != 1 ) {
+        // apply a general gain boost to all audio input:
+        for ( i = 0, j = 0; i < iMonoBlockSizeSam; i++, j += 2 )
+        {
+            vecsStereoSndCrd[j + 1] = static_cast<int16_t> ( iInputBoost * vecsStereoSndCrd[j + 1] );
+            vecsStereoSndCrd[j]     = static_cast<int16_t> ( iInputBoost * vecsStereoSndCrd[j] );
+        }
+    }
+
     // update stereo signal level meter (not needed in headless mode)
 #ifndef HEADLESS
     SignalLevelMeter.Update ( vecsStereoSndCrd,
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -253,6 +253,24 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
     edtNewClientLevel->setWhatsThis ( strNewClientLevel );
     edtNewClientLevel->setAccessibleName ( tr ( "New client level edit box" ) );
 
+    // input boost
+    QString strInputBoost = "<b>" + tr ( "Input Boost" ) + ":</b> " +
+        tr ( "This setting allows you to increase your input signal level "
+        "by factors up to 10 (+20dB)."
+        "If your sound is too quiet, first try to increase the level by "
+        "getting closer to the microphone, adjusting your sound equipment "
+        "or increasing levels in your operating system's input settings. "
+        "Only if this fails, set a factor here. "
+        "If your sound is too loud, sounds distorted and is clipping, this "
+        "option will not help. Do not use it. The distortion will still be "
+        "there. Instead, decrease your input level by getting farther away "
+        "from your microphone, adjusting your sound equipment "
+        "or by decreasing your operating system's input settings."
+        );
+    lblInputBoost->setWhatsThis ( strInputBoost );
+    cbxInputBoost->setWhatsThis ( strInputBoost );
+    cbxInputBoost->setAccessibleName ( tr ( "Input Boost combo box" ) );
+
     // custom central server address
     QString strCentrServAddr = "<b>" + tr ( "Custom Central Server Address" ) + ":</b> " +
         tr ( "Leave this blank unless you need to enter the address of a central "
@@ -348,6 +366,15 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
     // update new client fader level edit box
     edtNewClientLevel->setText ( QString::number ( pSettings->iNewClientFaderLevel ) );
 
+    // Input Boost combo box
+    cbxInputBoost->clear();
+    cbxInputBoost->addItem ( tr ( "None" ) );
+    for ( int i = 2; i <= 10; i++ ) {
+        cbxInputBoost->addItem ( QString( "%1x" ).arg( i ) );
+    }
+    // factor is 1-based while index is 0-based:
+    cbxInputBoost->setCurrentIndex ( pSettings->iInputBoost - 1 );
+
     // update enable small network buffers check box
     chbEnableOPUS64->setCheckState ( pClient->GetEnableOPUS64() ? Qt::Checked : Qt::Unchecked );
 
@@ -426,6 +453,9 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
 
     QObject::connect ( cbxLanguage, &CLanguageComboBox::LanguageChanged,
         this, &CClientSettingsDlg::OnLanguageChanged );
+
+    QObject::connect ( cbxInputBoost, static_cast<void (QComboBox::*) ( int )> ( &QComboBox::activated ),
+        this, &CClientSettingsDlg::OnInputBoostChanged );
 
     // buttons
     QObject::connect ( butDriverSetup, &QPushButton::clicked,
@@ -766,4 +796,11 @@ void CClientSettingsDlg::UpdateCustomCentralServerComboBox()
             cbxCentralServerAddress->addItem ( pSettings->vstrCentralServerAddress[iLEIdx], iLEIdx );
         }
     }
+}
+
+void CClientSettingsDlg::OnInputBoostChanged()
+{
+    // index is zero-based while boost factor must be 1-based:
+    pSettings->iInputBoost = cbxInputBoost->currentIndex() + 1;
+    pClient->SetInputBoost ( pSettings->iInputBoost );
 }
--- a/src/clientsettingsdlg.h
+++ b/src/clientsettingsdlg.h
@@ -97,6 +97,7 @@ public slots:
     void OnEnableOPUS64StateChanged ( int value );
     void OnCentralServerAddressEditingFinished();
     void OnNewClientLevelEditingFinished() { pSettings->iNewClientFaderLevel = edtNewClientLevel->text().toInt(); }
+    void OnInputBoostChanged();
     void OnSndCrdBufferDelayButtonGroupClicked ( QAbstractButton* button );
     void OnSoundcardActivated ( int iSndDevIdx );
     void OnLInChanActivated ( int iChanIdx );
--- a/src/clientsettingsdlgbase.ui
+++ b/src/clientsettingsdlgbase.ui
@@ -422,6 +422,13 @@
            </widget>
           </item>
           <item>
+           <widget class="QLabel" name="lblInputBoost">
+            <property name="text">
+             <string>Input Boost</string>
+            </property>
+           </widget>
+          </item>
+          <item>
            <widget class="QLabel" name="lblNewClientLevel">
             <property name="text">
              <string>New Client Level</string>
@@ -451,6 +458,9 @@
           </item>
           <item>
            <widget class="QComboBox" name="cbxAudioQuality"/>
+          </item>
+          <item>
+           <widget class="QComboBox" name="cbxInputBoost"/>
           </item>
           <item>
            <layout class="QHBoxLayout" name="horizontalLayout_4">
@@ -682,6 +692,7 @@
   <tabstop>sldNetBufServer</tabstop>
   <tabstop>cbxAudioChannels</tabstop>
   <tabstop>cbxAudioQuality</tabstop>
+  <tabstop>cbxInputBoost</tabstop>
   <tabstop>edtNewClientLevel</tabstop>
   <tabstop>cbxSkin</tabstop>
   <tabstop>cbxLanguage</tabstop>
--- a/src/client.h
+++ b/src/client.h
@@ -243,6 +243,8 @@ public:
     void SetRemoteChanPan ( const int iId, const float fPan )
         { Channel.SetRemoteChanPan ( iId, fPan ); }
 
+    void SetInputBoost ( const int iNewBoost ) { iInputBoost = iNewBoost; }
+
     void SetRemoteInfo() { Channel.SetRemoteInfo ( ChannelInfo ); }
 
     void CreateChatTextMes ( const QString& strChatText )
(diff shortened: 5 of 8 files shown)


src/clientsettingsdlgbase.ui
```diff
@@ -455,7 +462,7 @@
           <item>
            <layout class="QHBoxLayout" name="horizontalLayout_4">
             <item>
-             <widget class="QLineEdit" name="edtNewClientLevel"/>
```
ann0see, line 458: Not sure, but is this deletion intended?

hoffie: I think it should be correct, it might just look strange in the diff:

- Technically, I add a new QLineEdit for edtInputGain.
- To a diff algorithm it looks like I'm changing the existing edtNewClientLevel (the line you marked) to edtInputGain (the line below) and add a new QLineEdit for the old, existing edtNewClientLevel (see line 477).

ann0see: Ok. Seems you're right!

I looked into the file and it's ok there.

ann0see:
Ok. Just tested the artifact.

An immediate thought I had after installing it: why isn't it a slider? But probably that's ok if it's a one time setting.

hoffie:
> An immediate thought I had after installing it: why isn't it a slider? But probably that's ok if it's a one time setting.

That's a deliberate choice. This is supposed to be a one-time setting. Changing this repeatedly or on-the-fly would be really annoying for other participants who would have to constantly update the person's level in their mix.

DavidSavinkoff:
Internal computer microphones will pick up extraneous noises and are basically junk. Jamulus cannot make up for hardware shortcomings. Get a better microphone ... I did!
Moreover, amplifying a digital signal requires that a stream of data has to be tapped into and every sample must have a multiply operation done ... This decreases the resolution of the data which increases quantization noise (as if that matters, but that's what you get for your trouble). Now everybody who doesn't have a crappy microphone has to put up with more bloat and less preformance for the sake of someone who insists on using an internal microphone which can only be improved marginally by adding gain and noise gates (throwing good money after bad). I'm also blown away that 7 files need to be touched for this change.
Code added to Jamulus for bad microphones should be added to a special build.
Maybe there should be a beginners version of Jamulus.

gene96817:
Also, a cheap gamer headset with a mic is much better than the internal microphone (that picks up the fan noise).

hoffie:
Thanks for your comments, @DavidSavinkoff!

> Internal computer microphones will pick up extraneous noises and are basically junk.

I agree.

> Jamulus cannot make up for hardware shortcomings.

It can improve things in some cases, at least.

> Get a better microphone ... I did!

I did. Not everyone can or wants to affort that.

> Moreover, amplifying a digital signal requires that a stream of data has to be tapped into and every sample must have a multiply operation done ... This decreases the resolution of the data which increases quantization noise (as if that matters, but that's what you get for your trouble). Now everybody who doesn't have a crappy microphone has to put up with more bloat and less preformance

Yes, the relevant change is in the audio processing path. I would not call a multiply operation bloat or a relevant performance degration unless proven. Might make sense to run a benchmark though.

> for the sake of someone who insists on using an internal microphone which can only be improved marginally by adding gain

I don't agree here. Adding gain decides about being able to use Jamulus or not in such cases.

> . I'm also blown away that 7 files need to be touched for this change.

This is due to the organization of UI code. Adding anything with a UI setting is going to touch these files.

> Code added to Jamulus for bad microphones should be added to a special build.

If it is decided not to accept this PR for main stream Jamulus, then this will continue to be my workaround: Distributing a special build to the users who need it. I'm doing this already. I have opened this PR because the code has been developed, has proven to work and has helped making Jamulus usable for those persons.

A previous version of your comment suggested guarding such new code by `ifdef`s. This wouldn't help as the users affected by such problems would certainly not build their own Jamulus versions.

> Maybe there should be a beginners version of Jamulus.

I see you have already opened a discussion for that, thanks!
https://github.com/jamulussoftware/jamulus/discussions/1227

Please also note that this is still a Pull **Request**, so it is not decided yet whether to have this in Jamulus or not.

pljones:
This has been discussed before and my stance hasn't changed: input audio processing happens outside Jamulus.

ann0see:
Although in general I agree that audio processing should take place on OS level if possible, we also need to keep usability in mind. This feature seems simple and I _do_ see value with it.

gene96817:
For uses having trouble with their internal microphone, do we consider a set of earbuds with a microphone from their smartphone a good step? It does seem everyone has a smartphone and, by extension, earbuds and microphone too. This could be a quick (temporary) fix for the internal microphone noise.

npostavs:
I notice that I need to click outside of the text field in order for the newly entered setting to take effect. I think it would be useful to mention this in the help text. Alternatively, maybe there a different signal than `QLineEdit::editingFinished` that would work better?

> a microphone from their smartphone a good step?

I'd be surprised if this gave reasonable latency. Has anyone tried it?

gene96817:
> a microphone from their smartphone a good step?
I'd be surprised if this gave reasonable latency. Has anyone tried it?

I wasn't clear. Many people have a wired set of earbuds with microphone for their smartphone. We can hate the reposnse curve of that microphone. Seems to me that would be better than the builtin notebook microphone. That won't introduce any latency issues; its wired.

npostavs:
> I wasn't clear. Many people have a wired set of earbuds with microphone for their smartphone.

Oh I see. It's not common in my experience so I didn't think of that. I don't know whether it would generally produce better results, but if someone has access to multiple microphones they may as well try them all to see which one sounds best.

> Internal computer microphones will pick up extraneous noises and are basically junk. Jamulus cannot make up for hardware shortcomings.

This is sometimes true, but other times not. I play with 3 other people using internal sound cards and builtin/not great microphones. For 2 of them, there is quite a bit of background noise coming from their microphones, and this feature would not help them. But the 3rd person's built-in microphone actually sounds perfectly fine, it's just kind of quiet (I could hear them almost okay with both the Jamulus slider and audio interface volume set to max, but the others had trouble). I got them to test out this build yesterday, and it worked quite well. Thanks for posting it @hoffie!

> This has been discussed before and my stance hasn't changed: input audio processing happens outside Jamulus.

As far as I can tell, when using ASIO4ALL, there is no way to adjust microphone gain. The standard OS audio settings don't affect the sound level. So while this position seems logical, I don't think it corresponds to the reality.

hoffie:
Thanks for the feedback so far. Although I'm not sure if this has a chance of being accepted, people do use it and I have therefore fixed two bugs in 39cafd823e95df526a6761ce806c249b5b2aab90:
- As @npostavs pointed out, a gain change was not applied immediately. I have changed this by using a `textChanged` event instead (and ensuring the number is valid).
- The actual input gain was not restored properly upon Jamulus restart (but confusingly, the settings field was). This is now fixed as well.

gilgongo:
My two cents is that while I agree with @pljones, I think this issue has become a "will of the people" thing. If we oppose it, we will forever more be explaining why. So I'm OK not spending the energy on that, and instead accepting what is in reality a pretty minor UI change.

src/client.cpp
```diff
@@ -1068,6 +1069,13 @@ void CClient::ProcessAudioDataIntern ( CVector<int16_t>& vecsStereoSndCrd )
                               static_cast<float> ( iReverbLevel ) / AUD_REVERB_MAX / 4 );
     }
 
+    // apply a general gain to all audio input:
```
pljones, line 1072: Don't enter this block unless it's needed.  It's got a loop, so it's a fair saving for a simple `if`.

src/clientsettingsdlg.cpp
```diff
@@ -253,6 +253,24 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
     edtNewClientLevel->setWhatsThis ( strNewClientLevel );
     edtNewClientLevel->setAccessibleName ( tr ( "New client level edit box" ) );
 
+    // input gain
+    QString strInputGain = "<b>" + tr ( "Input Gain" ) + ":</b>" +
+        tr ( "This setting allows to increase your input signal level. "
```
pljones, line 258: "This setting allows to" -> "This setting allows _you_ to".

src/clientsettingsdlg.cpp
```diff
@@ -253,6 +253,24 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
...
+        "If your sound is too quiet, first try to increase the level by "
+        "getting closer to the microphone, adjusting your sound equipment "
+        "or increasing levels in your operating system's input settings. "
+        "Only if this fails, set a value above 100% here. "
+        "If your sound is too loud, sounds distorted and is clipping, this "
+        "option will not help. Do not use it. The distortion will still be "
+        "there. Instead, decrease your input level by getting farther away "
+        "from your microphone, adjusting your sound equipment "
+        "or by decreasing your operating system's input settings.<br />"
+        "Valid range: 100% (default) - 1000% (ten-time increase)"
```
pljones, line 268: Can you add
`"<br/> + "(In decibels: 0dB to +20dB)"`
(for those who think in decibels).

src/clientsettingsdlg.h
```diff
@@ -97,6 +97,14 @@ public slots:
     void OnEnableOPUS64StateChanged ( int value );
     void OnCentralServerAddressEditingFinished();
     void OnNewClientLevelEditingFinished() { pSettings->iNewClientFaderLevel = edtNewClientLevel->text().toInt(); }
+    void OnInputGainTextChanged() {
```
pljones, line 100: This is a bit long for the `.h` file...

hoffie: Moved it to the `.cpp` file and fixed a style issue as well.

src/clientsettingsdlg.cpp
hoffie: Thanks, done.

src/client.cpp
hoffie: Thanks, done.

pljones (approved):

softins:
Just fetched this to build it and try it, and noticed it said it was building "3.7.0 (release)".

It looks like this was branched from 03bc4c1109e4a44db57fda57232c0bbb6ce9fa4a (tag r3_7_0). We ought never to branch from a tagged release. In this case the branch point should be the 3.7.0dev commit immediately afterwards fabe41cfa85d6bb1f21c063b39483b05ddc08d08 or later.

softins (commented):
The code as it is looks fine, and does what it was intended to, but I have some comments and suggestions:
1. I noticed that changing the value did not affect the input level meter. It would probably help the user determine the optimum setting if the input level meter was driven from the scaled input.
2. I imagine the user does not need a very fine resolution in the setting, and I wonder if it would be better to have a drop-down offering multiples of 100% (or multiples of 2dB?)
3. For people who have a microphone in one channel and an instrument in the other, might they want to set the input gain independently for each channel?

softins:
In fact, since 100% is the lowest accepted value, maybe it should be called Input Boost, rather than Input Gain? That might make its purpose clearer?

hoffie:
> We ought never to branch from a tagged release. In this case the branch point should be the 3.7.0dev commit immediately afterwards [fabe41c](https://github.com/jamulussoftware/jamulus/commit/fabe41cfa85d6bb1f21c063b39483b05ddc08d08) or later.

Uhm, you are right. This shouldn't have happened. I've rebased on master now. Sorry for that.

Thanks for your other suggestions, I'll look at them later.

hoffie:
>     1. I noticed that changing the value did not affect the input level meter. It would probably help the user determine the optimum setting if the input level meter was driven from the scaled input.

This sounds like a good idea. I have adapted the code by moving the gain application right before the input level meter update.

>     2. I imagine the user does not need a very fine resolution in the setting, and I wonder if it would be better to have a drop-down offering multiples of 100% (or multiples of 2dB?)

- [ ] Yes, makes sense. Will do that tomorrow.

>     3. For people who have a microphone in one channel and an instrument in the other, might they want to set the input gain independently for each channel?

Hrm, good point. This would require more UI space and more logic though. I wonder if we should just wait if someone comes and requests this feature? It might not be that common to have a bad, low-gain microphone in addition to having an instrument attached (probably via some interface)? I'd tend to not add it now to keep it as simple as possible.

> In fact, since 100% is the lowest accepted value, maybe it should be called Input Boost, rather than Input Gain? That might make its purpose clearer?

Good idea, I have renamed all variables, functions, comments and labels accordingly.

DavidSavinkoff (commented):
In:
src/client.cpp
The following code is inefficient:
    if ( iInputBoost != 100 ) {
        // apply a general gain boost to all audio input:
        for ( i = 0, j = 0; i < iMonoBlockSizeSam; i++, j += 2 )
        {
            vecsStereoSndCrd[j + 1] = static_cast<int16_t> ( iInputBoost * vecsStereoSndCrd[j + 1] / 100);
            vecsStereoSndCrd[j]     = static_cast<int16_t> ( iInputBoost * vecsStereoSndCrd[j] / 100);
        }
Fix before for loop:
Boost=iInputBoost*0.01;

Then use Boost instead of iInputBoost, and get rid of the /100

hoffie:
> I imagine the user does not need a very fine resolution in the setting, and I wonder if it would be better to have a drop-down offering multiples of 100% (or multiples of 2dB?)

I have turned this into a combo box with factors ranging from "None" for no boost to 10x. What do you think?

>  The following code is inefficient:

Thanks for the pointer. I have incorporated this into the latest state of this PR. Due to the change from percentages to factors, the division by 100 is eliminated completely.

DavidSavinkoff (approved):
Audio processing code in src/client.cpp is efficient and minimally invasive.Good.

gene96817:
>  (or multiples of 2dB?)

The smallest useful increment is 3dB. However, most people cannot hear that difference. If a coarse adjustment is enough then 10dB is a common increment.

softins (approved):
There are a couple of unneeded .qm file updates here too, but it doesn't matter, as they will be updated when rebuilding anyway.
Code looks good, and the feature operates as intended.

ann0see:
Just a small question (didn't test the new code here): Why don't we support lowering the input gain too?

hoffie:
> Just a small question (didn't test the new code here): Why don't we support lowering the input gain too?

I don't see a use case for this. Reasoning:
- If a client is loud, but not clipping, that this is optimal because it uses the full range to express details of the audio. Other clients can simply turn the fader down in their mix.
- If a client is too loud and is clipping, reducing gain will not help. While this technically removes clipping, it does not remove the distortion introduced by it ("Garbage in, garbage out"). Even worse, it will probably prevent Jamulus' clipping indicator from working at all.

I also tried explaining this in the `What's this` text.

Let me know if you think otherwise. :)

ann0see:
> If a client is too loud and is clipping, reducing gain will not help. While this technically removes clipping, it does not remove the distortion introduced by it ("Garbage in, garbage out"). Even worse, it will probably prevent Jamulus' clipping indicator from working at all.

I usually have problems with people clipping not being to quiet. Therefore I tell them to reduce their input gain on OS level (if they don't have an interface). The question is where clipping occurs. If it happens before it comes into Jamulus, sure it wouldn't help. If it occurs in Jamulus, adding this feature does help.

hoffie:
As discussed with @ann0see via chat, we can merge this as-is for now and look if permitting lowering of input gain should be implemented.

DetlefHennings:
In a daily build of the currrent 3.7.1 state I noticed the implementation of the microphone boost. Certainly this is a very useful feature whenever the hardware microphone gain is insufficient. But the current implementation using equispaced numeric gain factors may not be adequate to human loudness perception. To match better with human perception gain factors should be spaced by equal factors, or even better, replaced by equispaced dB steps. These stops could either be  
0dB, +5dB, +10dB, +15dB, +20dB or, at a higher resolution, 
0dB, +2.5dB, +5dB, +7.5dB, +10dB, +12.5dB , +15dB, +17.5dB, +20dB 
The suggested steps result in an unchanged boost range, as 0dB and +20dB correspond to factors of 1 and 10 respectively.

softins:
For a setting that will be adjusted frequently in use, I would agree with you. But when I reviewed and tested this PR, with a low-level input, I did find a noticeable change with each boost factor. This feature is designed for set-and-forget use, and typically the user will just choose the setting appropriate to their input device and leave it there.

DavidSavinkoff:
There are already too many Gain levels... +6dB=2  +12dB=4  +20dB=10 are abundantly generous.
Attenuation is accomplished by right shifting a signed integer: -6dB=1/2  -12dB=1/4 -18dB=1/8
This hack (feature) will be abused and create more problems than it solves. Get a proper setup.

PR #1240, opened 2021-03-13
Tools: Add get_release_contributors.py, simplify check-wininstaller-translations.sh

hoffie:
* Tools: Add get_release_contributors.py
  This generates a list of Github user names who contributed to a specific release. This list is used in release announcements.
* Tools: Simplify check-wininstaller-translations.sh

### DIFF
--- a/tools/check-wininstaller-translations.sh
+++ b/tools/check-wininstaller-translations.sh
@@ -11,7 +11,7 @@ for LANGUAGE_FILE in ${LANGUAGE_FILES}; do
     echo
     echo "* ${LANGUAGE_FILE}"
     echo -n "  - Checking language file is included in ${INSTALLERLNG}... "
-    if grep -q '^!include "\${ROOT_PATH}\\'$(tr '/' '\\' <<<"${LANGUAGE_FILE}" | sed -re 's|\\|\0\0|g')'"' "${BASE_DIR}/${INSTALLERLNG}"; then
+    if grep -q '^!include "\${ROOT_PATH}\\'$(sed -re 's|/|\\\\|g' <<<"${LANGUAGE_FILE}")'"' "${BASE_DIR}/${INSTALLERLNG}"; then
         echo "ok"
     else
         echo "ERROR"
--- a/tools/get_release_contributors.py
+++ b/tools/get_release_contributors.py
@@ -1,0 +1,177 @@
+#!/usr/bin/env python3
+"""
+Generates a list of Github user names who contributed to a specific release.
+
+May need a Github Personal access token to avoid hitting rate limits.
+
+Usage:
+./tools/get_release_contributors.py --from release --to translate3_7_0 --github-token YOUR_TOKEN --repo /path/to/your/jamuluswebsite
+./tools/get_release_contributors.py --from r3_6_2 --to r3_7_0rc1 --github-token YOUR_TOKEN --repo /path/to/your/jamulus
+
+"""
+import requests
+import argparse
+import logging
+import os
+import subprocess
+import yaml
+
+logger = logging.getLogger('')
+
+
+class UnexpectedGithubStatus(RuntimeError):
+    """
+    Exception which is raised when an API request fails.
+    """
+    pass
+
+
+# List of user names which should be ignored (such as bots):
+ignore_list = ['github-actions[bot]']
+
+
+class Authors:
+    """
+    The Authors class provides methods to get the Github login of each
+    commit in an efficient fashion.
+    It uses the Github API and caches name+email-to-login associations
+    to avoid excessive API usage.
+    """
+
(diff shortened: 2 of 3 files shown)


tools/check-wininstaller-translations.sh
```diff
@@ -11,7 +11,7 @@ for LANGUAGE_FILE in ${LANGUAGE_FILES}; do
     echo
     echo "* ${LANGUAGE_FILE}"
     echo -n "  - Checking language file is included in ${INSTALLERLNG}... "
-    if grep -q '^!include "\${ROOT_PATH}\\'$(tr '/' '\\' <<<"${LANGUAGE_FILE}" | sed -re 's|\\|\0\0|g')'"' "${BASE_DIR}/${INSTALLERLNG}"; then
+    if grep -q '^!include "\${ROOT_PATH}\\'$(sed -re 's|/|\\\\|g' <<<"${LANGUAGE_FILE}")'"' "${BASE_DIR}/${INSTALLERLNG}"; then
```
npostavs, line 14: I see how getting rid of the "backslash to nul character" transformation makes sense, but it looks like replacing `tr` with `sed` does the opposite of simplifying.

hoffie: > but it looks like replacing tr with sed does the opposite of simplifying.

I agree that it still doesn't look nice, but also consider that it replaces one `tr` + non-nice `sed` with just a single (still) non-nice `sed`. @softins initially proposed this and I think it does simplify, but if there's agreement that it does not, then it's fine for me as well.

This was initially proposed here: https://github.com/jamulussoftware/jamulus/pull/1203/files#diff-0307ece3ec605715410994dc14469df63db22548ac539792e241b7f561d6a918R14


This is really just a helper script for the maintainers and nothing in the Jamulus core. Maybe we should even move these into their own repo?

softins: I don't believe that is a null character. It looks to me like \0 in that context means "what was matched", the same as & would. I can't see how it would work otherwise.

I meant to reply earlier. My suggestion was only from the point of view that it seemed more efficient to do `/` > sed > `\\` than doing `/` > tr > `\` > sed > `\\`

But it's only a micro-optimisation.

hoffie: > I don't believe that is a null character. It looks to me like \0 in that context means "what was matched", the same as & would. I can't see how it would work otherwise.

Yes, that was the intention and that's how it works.

npostavs: > I don't believe that is a null character. It looks to me like \0 in that context means "what was matched", the same as & would.

Oh, you're right. Never mind then.

tools/get_release_contributors.py
```diff
@@ -0,0 +1,134 @@
...
+
+def find_contributors(git_log_selector, from_, to):
+    contributors = set()
+    commits = subprocess.check_output(['git', 'log', '--format=format:%H %an <%ae>', '%s..%s' % (from_, to), '--'] + git_log_selector)
+    commits = commits.decode('utf-8')
+    for commit in commits.split('\n'):
+        hash, author_key = commit.split(' ', 1)
+        login = authors.get_login(author_key, hash)
+        contributors.add(login)
+    return sorted(contributors)
```
softins, line 121: I'm not very familiar with python, but have tried running this script. Is it possible to sort the list of usernames case-insensitively? Currently, `@SeeLook` and `@Snayler` come before `@ann0see`.

softins (approved):

ann0see (commented):
Not sure about python, but can you please add some comments describing what it does? 

From scanning it, it seems to use the GH API?

hoffie:
> Not sure about python

Yeah, I did not ask beforehand. My thoughts were: "This is repeated, manual work, lets automate it", "If I automate it for me for 3.7.0, I should probably share what I have done so that others can use it in the future as well". This started out as a bash "oneliner", but quickly became too complex, which is why I resorted to python.

> but can you please add some comments describing what it does?

A high-level comment is at the [top](https://github.com/jamulussoftware/jamulus/pull/1240/files#diff-323986ba9bba39f6d392bf151262523d7310442d42a4ce1882deff3f83a20ea5R3), but I can add some additional comments about the specific steps.

tools/get_release_contributors.py
hoffie: Done, will update the announcement as well.

hoffie:
> Not sure about python, but can you please add some comments describing what it does?
> 
> From scanning it, it seems to use the GH API?

Meanwhile, I have added explaining docstrings in relevant places and have extended the `--help` text. Is this sufficient?

ann0see (approved):
Yes. I'd that it's ok like that.

PR #1247, opened 2021-03-14
Add tool to aid manual checking for duplicated accelerator keys.

softins:
This is the perl script used to find potential conflicts for manual checking. Used when preparing #1161.

It should be run in the directory `src/res/translation`

### DIFF
--- a/tools/checkkeys.pl
+++ b/tools/checkkeys.pl
@@ -1,0 +1,76 @@
+#!/usr/bin/perl
+
+use open qw(:std :utf8);
+use XML::Simple qw(:strict);
+use Data::Dumper;
+
+my %keys;
+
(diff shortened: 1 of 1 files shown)


pljones (approved):
Approving without reading as it's not core code and my PERL days are a fuzzy distant memory.

hoffie (approved):
Approving as it works and is helpful. :)

Still some ideas for you to decide:
- Should we rather name this check_accelerator_keys.pl or something? `checkkeys.pl` is ambiguous, might also refer to cryptographic keys. :)
- It would be helpful if the script errored out if run from the wrong directory (it just fails with almost empty output currently) and/or if the comment from your PR was a commen in the header of the script as well?

PR #1260, opened 2021-03-14
Copy ASIO4ALL settings from old version

### LINKED ISSUE #1244: Installing new 64 bit Jamulus loses ASIO4ALL settings (& installer strongly recommends to remove the old version)
**Describe the bug**

The latest Jamulus builds for Windows are 64 bit, the previous versions were 32 bit. I think ASIO4ALL maintains customized settings per-exe file path and the install location is different for 32 vs 64 bit programs. This means the previously working settings for the old 32 bit version are not carried over to the new 64 bit version. Furthermore, the installer strongly recommends removing the older 32 bit version (this makes sense, because otherwise you end up with 2 sets of shortcuts, and things get kind of confusing).

**To Reproduce**

Have Jamulus working with ASIO4ALL on the old 32 bit version. Install a new 64 bit version, and click yes when installer strongly recommends removing old 32 bit version. Realize that ASIO4ALL is using the default settings now, and you don't remember what the good settings were. Reinstall the old Jamulus version to get the old settings, take a screenshot and get copy over the working settings to the 64 bit version.

I spent about an hour and a half talking someone through this over the phone yesterday.

**Expected behavior**

I don't think it's feasible for Jamulus to avoid this outright, but at the point where the installer recommends removing the old 32 bit Jamulus, it would be good to have a warning about it.

**Operating system**

This applies to Windows only.

**Jamulus version**

3.7.0, 64 bit builds.

ann0see:
Fixes #1244

### DIFF
--- a/windows/NSISCopyRegistryKey/registry.nsh
+++ b/windows/NSISCopyRegistryKey/registry.nsh
@@ -1,0 +1,166 @@
+!ifndef REGISTRY_NSH
+!define REGISTRY_NSH
+!include "${NSISDIR}\Examples\System\system.nsh"
+
+!define HKEY_CLASSES_ROOT_ENUM           0x80000000
+!define HKEY_CURRENT_USER_ENUM           0x80000001
+!define HKEY_LOCAL_MACHINE_ENUM          0x80000002
+!define HKEY_USERS_ENUM                  0x80000003
+!define HKEY_PERFORMANCE_DATA_ENUM       0x80000004
+!define HKEY_CURRENT_CONFIG_ENUM         0x80000005
+!define HKEY_DYN_DATA_ENUM               0x80000006
+
+!define KEY_ALL_ACCESS          0x0002003F
+
+!define RegCreateKeyEx 'advapi32::RegCreateKeyEx(i, t, i, t, i, i, i, *i, *i) i'
+!define RegOpenKeyEx 'advapi32::RegOpenKeyEx(i, t, i, i, *i) i'
+!define RegCloseKey  'advapi32::RegCloseKey(i) i'
+!define SHCopyKey 'shlwapi::SHCopyKey(i, t, i, i) i'
+
+####################################################################################################
+!macro SET_HKEY_ENUM HKEY_NAME HKEY_ENUM LABEL
+    StrCmp ${HKEY_NAME} HKCR 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_CLASSES_ROOT_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKEY_CLASSES_ROOT 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_CLASSES_ROOT_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKCU 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_CURRENT_USER_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKEY_CURRENT_USER 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_CURRENT_USER_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKLM 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_LOCAL_MACHINE_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKEY_LOCAL_MACHINE 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_LOCAL_MACHINE_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKU 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_USERS_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKEY_USERS 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_USERS_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKPD 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_PERFORMANCE_DATA_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKEY_PERFORMANCE_DATA 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_PERFORMANCE_DATA_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKCC 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_CURRENT_CONFIG_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKEY_CURRENT_CONFIG 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_CURRENT_CONFIG_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKDD 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_DYN_DATA_ENUM}
+    GoTo ${LABEL}
+    StrCmp ${HKEY_NAME} HKEY_DYN_DATA 0 +3
+    StrCpy ${HKEY_ENUM} ${HKEY_DYN_DATA_ENUM}
+${LABEL}:
+!macroend
+####################################################################################################
+!macro COPY_REGISTRY_KEY_BASE TYPE
+!ifndef ${TYPE}COPY_REGISTRY_KEY_DEFINE
+!define ${TYPE}COPY_REGISTRY_KEY_DEFINE
+Function ${TYPE}CopyRegistryKey
+    Exch $1 ;Target subkey
+    Exch
+    Exch $2 ;Target key name
+    Exch
+    Exch 2
+    Exch $3 ;Source subkey
+    Exch
+    Exch 2
+    Exch 3
+    Exch $4 ;Source key name
+    Push $5 ;Source key handle
+    Push $6 ;Target key handle
+    Push $7 ;return value
+    Push $8 ;HKEY enum for target key
+    Push $9 ;HKEY enum for source key
+
+!insertmacro SET_HKEY_ENUM $2 $8 next
+!insertmacro SET_HKEY_ENUM $4 $9 next2
+
+     SetPluginUnload alwaysoff
+
+     StrCpy $5 0
+     System::Call '${RegOpenKeyEx}(i r9, t r3, 0, ${KEY_ALL_ACCESS}, .r5) .r7'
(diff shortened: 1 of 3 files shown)


windows/installer.nsi
```diff
@@ -212,6 +218,30 @@ Section "Install_64Bit" INST_64
                 MessageBox MB_OKCANCEL|MB_ICONEXCLAMATION "$(OLD_VER_REMOVE_FAILED)" /sd IDCANCEL IDOK continueinstall
                 goto quit
             ${EndIf}
+
+            ; ask to move old ASIO4ALL registry configuration
+
+            ; check if old ASIO4ALL keys exit
+
+            ClearErrors
+            EnumRegKey $0 HKCU "SOFTWARE\ASIO4ALL v2 by Wuschel\8A9E7A56" 0
```
ann0see, line 227: Check if the key of the old version of A4A exists for the currently logged in user (not really a good way to check that. The currently logged in user might not have started Jamulus + A4A yet)

windows/installer.nsi
```diff
@@ -212,6 +218,30 @@ Section "Install_64Bit" INST_64
...
+
+            IfErrors 0 continueinstall
+                MessageBox MB_YESNO|MB_ICONEXCLAMATION "Do you want to copy ASIO4ALL settings from the previous version if possible?" /sd IDYES IDNO continueinstall
```
ann0see, line 230: If we find this key, we'll ask if he wants to copy the settings (this message doesn't come up yet.)

windows/installer.nsi
```diff
@@ -212,6 +218,30 @@ Section "Install_64Bit" INST_64
...
+
+                    IntOp $0 0 + 0
+                    EnumStart:
```
ann0see, line 233: For each user, replace the old with the new key contents

windows/installer.nsi
```diff
@@ -212,6 +218,30 @@ Section "Install_64Bit" INST_64
                 MessageBox MB_OKCANCEL|MB_ICONEXCLAMATION "$(OLD_VER_REMOVE_FAILED)" /sd IDCANCEL IDOK continueinstall
                 goto quit
             ${EndIf}
```
ann0see, line 220: This code will only be executed if we find jamulus in the 32 bit prog. files folder and the current install is 64 bit.

ann0see (commented):
Not yet ready; added some explanations

ann0see:
Feel free to edit the code.

windows/installer.nsi
softins: Should we do that? It feels wrong to update other users and not just HKCU.

ann0see: On my PC I have two users: 
1. Admin (user which I use to install Jamulus)
2. User (user which I use to use Jamulus)
If we don't  copy all users' settings, this issue would only be fixed for Admin but not for User.

ann0see:
The latest commit makes it very simple. And seems work. We're not checking for anything, so probably it would also overwrite some other things.

windows/installer.nsi
softins: But if Admin has never run Jamulus, it won't find the ASIO key in HKCU? Or if it does, the settings may well not be the same as were set the last time the User user Jamulus? Does the check for existing settings also need to scan the registry for users then?

The installer presumably has admin privileges, because I've never switched to admin to install Jamulus.

ann0see: > But if Admin has never run Jamulus, it won't find the ASIO key in HKCU? 

On my PC at least, Admin did run Jamulus. 

> Does the check for existing settings also need to scan the registry for users then?

Yes. This makes this code a lot more complicated.

Furthermore something like that would have to be re-implemented (but it's not that easy unfortunately):

foreach (user in users) {
    if (user.reg.oldkey.exists() && !user.reg.newkey.exists()) {
        user.reg.newkey.setValue(oldkey.getValue());
    }
}

hoffie: This pseudo-code sounds how it should work. Too bad that NSIS seems to be so complicated. :(

If NSIS implementation is the problem, would there be other workarounds? E.g. replacing the Jamulus shortcut by one pointing to a wrapper script which performs the per-user migration (maybe in Powershell?) and launches Jamulus.exe afterwards?
Not sure what side effects we would have to check for then or how it would look like to the user.

ann0see: NSIS should support that. We could use https://nsis.sourceforge.io/mediawiki/index.php?title=Registry_plug-in&oldid=16847 but I don't really want to include another plug-in. We already have too many in the installer.

ann0see:
Tests so far with new code: 

1. Install new version on top of old version from non admin account with admin privileges from admin account: If I click "yes" in the my message, the settings get transfered. Note: the new registry key doesn't exist yet.
2. Remove new version and install old version --> old settings are saved for the old version
3. Install new version on top of old version after changes have been made in the new version. Press no to the copy message on the message in the installer --> Previously set settings from the new version get saved
4. Remove new version, install old version (without removing registry key), afterwards install new version. New key doesn't get touched. As expected. I'll need to move the message.
Still todo: 

1. Check what happens for the admin user (at least the new registry key seems to be created)

windows/deploy_windows.ps1
```diff
@@ -284,19 +284,19 @@ Function Build-NSProcess
...
         Remove-Item -Path "$WindowsPath\nsProcess\Release\" -Force -Recurse
         $OriginalEnv | % { Set-Item "Env:$($_.Name)" $_.Value }
     }
 }
 
-Clean-Build-Environment
-Install-Dependencies
-Build-App-Variants -QtInstallPath $QtInstallPath
-Build-NSProcess -QtInstallPath $QtInstallPath
+#Clean-Build-Environment
```
ann0see, line 298: This is used for debugging only!!!

windows/installer.nsi
ann0see: Ok. I think I got it to work now.

windows/installer.nsi
```diff
@@ -212,6 +218,34 @@ Section "Install_64Bit" INST_64
...
+
+            ; ask to move old ASIO4ALL registry configuration
+
+            MessageBox MB_YESNO|MB_ICONEXCLAMATION "If your're using ASIO4ALL: Do you want to copy ASIO4ALL settings from the previous version if possible? We will do this for every user." /sd IDYES IDNO continueinstall
```
ann0see, line 224: Do we need this message?

hoffie: I can't think of any reason why users would not want it.
As a user, I'd expect Jamulus+ASIO4ALL to work the same after the update than before. This logic tries to do that.
The logic is supposed to be non-destructive, i.e. it should not overwrite any configs (we should be really sure about that).

So my tendency would be to not add a (possibly confusing) message. Would also have the benefit of not needing another translators ping...

ann0see: Ok. Just did a force push without this change. I think now we'll just need to do some tests.

ann0see:
Still one edge case: What should happen if an user now sets a different installation path? Do we need to handle this?

hoffie:
> Still one edge case: What should happen if an user now sets a different installation path? Do we need to handle this?

That would have always been a problem, wouldn't it?

I think we could add a message on the page where the target path is chosen and warn about it ("If you change the installation path from previous installations, you may have to re-configure ASIO4ALL."). However, as this is not a regression as far as I understand) and because it would benefit from being translated, I'd suggest to postpone that part for 3.7.1?

ann0see:
> That would have always been a problem, wouldn't it?

I think so.

softins (approved):
Appears to work ok on my W7 64-bit:
1. Installed 3.6.2
2. Made sure key 7A49ECC9 did not exist and 8A9E7A56 did exist.
3. Installed artifact from this PR. Was warned about removing the 32-bit version, which I accepted.
4. Checked registry and confirmed everything in 8A9E7A56 had been copied to 7A49ECC9.
5. Only one user on this PC had ever used Jamulus, so couldn't check for others.

ann0see:
Logged in as admin: 
1. Removed new key in registry; installed the version from the artifact --> Previous settings are restored (since the new key gets created)

I think this is the most important check we need to make.

windows/installer.nsi
```diff
@@ -212,6 +218,32 @@ Section "Install_64Bit" INST_64
...
+            EnumStart:
+                EnumRegKey $R1 HKEY_USERS "" $0 ; foreach user
+                IntOp $0 $0 + 1
+                StrCmp $R1 ".DEFAULT" EnumStart
+                StrCmp $R1 "" EnumEnd
+
+                ; check if new key already exists. If this is the case, we'll not continue
+                ClearErrors
+                EnumRegKey $1 HKU "$R1\SOFTWARE\ASIO4ALL v2 by Wuschel\7A49ECC9" 0
+                IfErrors 0 EnumStart ; if the above line gives an error, it can not find the key --> We'll continue
```
hoffie, line 234: Nit:
```suggestion
                IfErrors 0 EnumStart ; if the above line gives an error, it cannot find the key --> We'll continue
```

windows/installer.nsi
```diff
@@ -212,6 +218,32 @@ Section "Install_64Bit" INST_64
...
+
+                ; check if old key already exists. If this is true, we'll continue and move the content of the old one to the new one.
```
hoffie, line 236:
```suggestion
                ; check if old key exists. If this is true, we'll continue and move the content of the old one to the new one.
```

windows/installer.nsi
```diff
@@ -212,6 +218,32 @@ Section "Install_64Bit" INST_64
...
+                ClearErrors
+                EnumRegKey $1 HKU "$R1\SOFTWARE\ASIO4ALL v2 by Wuschel\8A9E7A56" 0
+                IfErrors EnumStart 0 ; if the above line gives an error, it can not find the key --> skip this user
```
hoffie, line 239:
```suggestion
                IfErrors EnumStart 0 ; if the above line gives an error, it cannot find the key --> skip this user
```

hoffie (approved):
I can't really comment on the NSH and NSI code as I have zero experience with it. I can't test on Windows either, sadly.

However, I'm confident that we should have this change and logic-wise it looks sounds.

I'd be happy if yet another person could test the installer, either explicitly or through another rc.

windows/NSISCopyRegistryKey/registry.nsh
```diff
@@ -0,0 +1,166 @@
+!ifndef REGISTRY_NSH
```
pljones, line 1: This code has no "watermark" -- it needs some declaration of authorship, licence, etc.

ann0see: It should also be under the same license as nsProcess. Maybe we should just add it here?

ann0see: Done now. I hope this comment is ok.

ann0see:
Before merging this, I'd like to have feedback from @gilgongo

gilgongo:
I can install it on my machine? It has old settings from a previous install. Where do I download it?

ann0see:
I think the windows compilation process isn't finished yet, so you'll need to wait a bit. Afterwards you can find the binary on the checks page of this PR.

ann0see:
I've uploaded the binary here: https://github.com/jamulussoftware/jamulus/releases/tag/untagged-964d405309c4e6a354c3

pljones (approved):

ann0see:
I've merged this now since I think it works as expected.

PR #1267, opened 2021-03-15
Fix changing GUI style.

### LINKED ISSUE #1238: Switching between skins causes mixing board to stay white
Seems as if I can trace down a bug to https://github.com/jamulussoftware/jamulus/commit/f5a58a0d2db7f38d8f6cdc5243b276062c9b2b8b

On Linux: Debian 10 Gnome after this commit the mixing board seems not to change back to fancy, normal, compact mode if I switch it during runtime.

See
https://github.com/jamulussoftware/jamulus/discussions/1217#discussioncomment-464967

Steps to reproduce:

Now I could reproduce it like that:

1. Start Jamulus with fancy mode enabled
1. Switch to normal mode while settings is open
1. Switch back to fancy mode (maybe you'll need to switch back and forth between the different modes).

Note: I'm not connected to a server.

I seen the only one who can reproduce this bug

softins:
The stylesheet on the main mixer board must be removed before applying
a global stylesheet to the GUI, and then reapplied afterwards.

Fixes #1238

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1102,11 +1102,6 @@ void CAudioMixerBoard::UpdateTitle()
     if ( eRecorderState == RS_RECORDING )
     {
         strTitlePrefix = "[" + tr ( "RECORDING ACTIVE" ) + "] ";
-        setStyleSheet ( AM_RECORDING_STYLE );
-    }
-    else
-    {
-        setStyleSheet ( "" );
     }
 
     setTitle ( strTitlePrefix + tr ( "Personal Mix at: " ) + strServerName );
--- a/src/audiomixerboard.h
+++ b/src/audiomixerboard.h
@@ -184,12 +184,6 @@ protected:
 template<>
 class CAudioMixerBoardSlots<0> {};
 
-#define AM_RECORDING_STYLE "QGroupBox::title { subcontrol-origin: margin; \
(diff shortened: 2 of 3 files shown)


ann0see:
Thank you very much! I'll try it this afternoon/evening. After this and the ASIO4ALL copy thing #1260 is fixed, I'd tag the release!

dcorson-ticino-com:
Thanks for finding this Tony !
Is the SetMixerBoardDeco routine still needed or can it be removed ?
Under Win10 your correction works without SetMixerBoardDeco.

softins:
> Is the SetMixerBoardDeco routine still needed or can it be removed ?

Yes, it's still needed, to reapply the recording indication after changing GUI. Otherwise it stops being red.

dcorson-ticino-com (approved):

ann0see (approved):
Just tested it on my machine, and it fixes the bug. Thanks again!

softins:
There is a small issue on my pi with the title positioning. Just looking at it.

softins:
> There is a small issue on my pi with the title positioning. Just looking at it.

The issue is identical in 3.6.2, so I don't think it is worth delaying the release for. It only occurs in Normal/Compact style.

dcorson-ticino-com (approved):

pljones (approved):

PR #1301, opened 2021-03-18
Exit properly in client mode with nosound

### LINKED ISSUE #1291: Jamulus built with CONFIG+="headless nosound" abort()s upon Ctrl+C
**Describe the bug**

Running Jamulus built with `CONFIG+="headless nosound"` dumps core when pressing Ctrl+C.

**To Reproduce**

- Build with `CONFIG+="headless nosound"`
- `./Jamulus`
- Ctrl+C
- Observe crash.

**Expected behavior**

Clean exit

**Screenshots**

```
$ ./Jamulus
No GUI support compiled. Running in headless mode.
 *** Jamulus, Version 3.7.0rc2dev-913e4cc8
 *** Internet Jam Session Software
 *** Released under the GNU General Public License (GPL)
^CQThread: Destroyed while thread is still running
Aborted (core dumped)
```
```
Stack trace of thread 58464:
#0  0x00007f539b803ef5 raise (libc.so.6 + 0x3cef5)
#1  0x00007f539b7ed862 abort (libc.so.6 + 0x26862)
#2  0x00007f539bd629ac _ZNK14QMessageLogger5fatalEPKcz (libQt5Core.so.5 + 0x909ac)
#3  0x00007f539bd6329a n/a (libQt5Core.so.5 + 0x9129a)
#4  0x000055ea4274d0b7 n/a (/home/christian/Projects/jamulus/Jamulus + 0x210b7)
#5  0x00007f539c21a550 _ZTV7QThread (libQt5Core.so.5 + 0x548550)
```

**Operating system**

`Linux wuechoo 5.11.1-arch1-1 #1 SMP PREEMPT Tue, 23 Feb 2021 14:05:30 +0000 x86_64 GNU/Linux
`

**Version of Jamulus**

3.7.0rc2

**Additional context**

- I don't think this is a recent regression.
- I don't think it's too relevant because it happens on intended exit anyway.
- We should still look into it.
- I don't have an idea yet, so feel free to take over and assign yourself.

hoffie:
In builds with CONFIG+="nosound" the only supported mode is server mode.
Error out properly instead of crashing.

Fixes #1291

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -340,7 +340,10 @@ win32 {
     DEFINES += HAVE_STDINT_H
 
     # only include jack support if CONFIG nosound is not set
-    !contains(CONFIG, "nosound") {
+    contains(CONFIG, "nosound") {
+        message(Restricting build to server-only due to CONFIG+=nosound.)
+        DEFINES += SERVER_ONLY
+    } else {
         message(Jack Audio Interface Enabled.)
 
         contains(CONFIG, "raspijamulus") {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -565,6 +565,14 @@ int main ( int argc, char** argv )
     Q_UNUSED ( bMuteStream )           // avoid compiler warnings
 #endif
 
+#ifdef SERVER_ONLY
+    if ( bIsClient )
+    {
+        qCritical() << "Only --server mode is supported in this build with nosound.";
+        exit ( 1 );
+    }
+#endif
+
     // the inifile is not supported for the headless server mode
     if ( !bIsClient && !bUseGUI && !strIniFileName.isEmpty() )
     {


pljones (approved):

softins (approved):
Haven’t tested, but looks good.

hoffie:
Converting to Draft again as I suspect that this might break Windows as it seems that `WITH_SOUND` is only used with jack? Needs confirming.

softins:
You're right. The `nosound` config option is only effective in Unix/Linux mode. Other platforms are always build with sound support (need to check Mac server build to be certain).

I think the easiest thing is to add to `Jamulus.pro` in the `nosound` section:
```
CONFIG += SERVER_ONLY
```
And then use `#ifdef SERVER_ONLY` to include your code.

hoffie:
Yes, thanks. The `nosound` handling is split all over Jamulus.pro, so I chose to at this logic only once in a single place. This means if someone is building on Windows (for whatever reason) with `CONFIG+=nosound`  this would be a change in behavior despite being not required. Previously, this would have been a no-op.

Ready for another review now.

Jamulus.pro
```diff
@@ -1078,4 +1078,8 @@ contains(CONFIG, "disable_version_check") {
     DEFINES += DISABLE_VERSION_CHECK
 }
 
+contains(CONFIG, "nosound") {
+    message(Restricting build to server-only due to CONFIG+=nosound.)
+    DEFINES += SERVER_ONLY
+}
```
softins, line 1084: Since the `nosound` option is only otherwise used within `else:unix`, I would move this section up to line 343 or 356, and make one of them an `else`. Then it doesn't affect any other platforms.

hoffie: > (For clarity, it is WITH_SOUND that is split all over Jamulus.pro, not the nosound config option.)

Yes, this is where I got confused. Moved it.

softins (approved):
That's great, thanks. Test build on my Pi works fine.

pljones (approved):

PR #1303, opened 2021-03-18
Generate .qm files during build

### LINKED ISSUE #1111: Remove QM files from source tree; generate during build process
Translators are providing translations through translation source files (TS / ```.ts```):
```bash
src/res/translation/translation_*.ts
```

Prior to release, these need to be turned into (QM / ```.qm```) files and shipped with the app. Currently, this is done manually. QM files are in a compact binary format that is used by the localized application.

These QM files are currently sitting in the source tree:
```bash
src/res/translation/translation_*.qm
```

Suggested action:

- QM files (binary blobs!) have no business residing in the source tree. And the ones that are currently in there are not consistently updated anyway. They also lead to confusion (e.g. see #1099). To be removed.
- These QM files should be generated as part of the build process.

hoffie:
This removes .qm files from git, adds them to .gitignore and changes Jamulus.pro to generate the files dynamically during build in a way which hopefully even works on ancient Qt versions.

This makes use of custom compiler support in qmake which seems to be
available (at least) as of Qt 5.5.
https://doc.qt.io/archives/qt-5.5/qmake-advanced-usage.html#adding-compilers

Tested on Linux, needs validation on Mac/Windows.

Fixes #1111


softins (changes requested):
This needs a bit more investigation before being included. I went round the same loop a month or so ago, without success. The build gave warnings that the .qm resources could not be found, as they were mentioned in the resource file. And if I built anyway, I think the executable did not include the other languages.

Currently, building from source does not have lrelease as a dependency, and with this approach it will do. That’s ok, but a change in requirements.

pljones:
> Currently, building from source does not have lrelease as a dependency, and with this approach it will do. That’s ok, but a change in requirements.

Like I said last time, I think most Qt developers -- and that should be _anyone_ building from source -- would have `lrelease`.

hoffie:
I'm closing this for now as I'm not sure whether it's the right approach to tackle this anymore.

PR #1310, opened 2021-03-20
DSCP Quality of Service

DavidSavinkoff:
This pull request is expected to:
- build on Windows and Linux and the others also
- adds Quaility of Service to the command line: eg Jamulus -Q 0 :: Jamulus -s -Q 104 :: Jamulus --help

It is time for a PR after much discussion see #1072


dakhubgit:
Since the actual DSCP value and the numeric value for what was once TOS differ, maybe it would make sense to provide a number of symbolic names (like AF31) here?  Picking a sensible value will be tricky either way.

src/socket.cpp
```diff
@@ -40,6 +40,9 @@ void CSocket::Init ( const quint16 iPortNumber )
 
     // create the UDP socket
     UdpSocket = socket ( AF_INET, SOCK_DGRAM, 0 );
+    //
+    const char tos = (const char) iQosNumber;  // Quality of Service
+    setsockopt ( UdpSocket, IPPROTO_IP, IP_TOS, &tos, sizeof(tos) );
```
dakhubgit, line 45: This will prioritise all traffic on all sockets, even the purely informational traffic between server directory (central server) and server.  For an approach that at least tries to confine the prioritisation to sockets where urgent traffic happens (though not distinguishing on a per-message basis which would, for efficiency reasons, require using a raw socket where changing prioritisation on a per-message basis does not come at the cost of a system call), see https://github.com/dakhubgit/jamulus/commit/d206159e60caf6d21d3b1821248a7f9a1d9cff15

DavidSavinkoff: This is currently how Jamulus does things... in bulk as UDP. Note that audio is > 99.9% of the traffic. Chat messages and server lists may, some day, be separated. Note that iQosNumber can be intercepted before it gets here. Thus, this code already is heading the way you want... but that is not for me.

dakhubgit: Did you look at the commit I specified?  The whole point was that the IP_TOS for a CSocket was only switched for those sockets that had at least one CHighPrioSocket open on them.  Why would a directory server be contacted via a CHighPrioSocket?

DavidSavinkoff: @dakhubgit : Why would a directory server be contacted via a CHighPrioSocket?
Thats the way Jamulus was before I wandered past.

pljones: @dakhubgit All the Any Genre directory servers are all processing audio priority network traffic from multiple clients - they're normal servers providing a real-time jam platform, _as well_.  All traffic uses the same sockets currently, only differentiated at the application protocol layer.

(The genre-specific directory servers only have a single client connection allowed and I run them at lower CPU priority than my jam server.)

src/main.cpp
```diff
@@ -321,6 +322,24 @@ int main ( int argc, char** argv )
...
+                                  argv,
+                                  i,
+                                  "-Q",
+                                  "--qos",
+                                  0,
+                                  255,
+                                  rDbleArgument ) )
+        {
+            iQosNumber            = static_cast<quint16> ( rDbleArgument );
+            qInfo() << qUtf8Printable( QString( "- selected tos value: %1" )
```
pljones, line 336: Users aren't going to like "qos" for the parameter name and "tos" for the description of what it is when it's printed... :)

src/main.cpp
```diff
@@ -823,6 +844,7 @@ QString UsageArguments ( char **argv )
         "                        supported for headless server mode)\n"
         "  -n, --nogui           disable GUI\n"
         "  -p, --port            set your local port number\n"
+        "  -Q, --qos             set QoS: DF=0 AF31=104 CS4=128default EF=184\n"
```
pljones, line 847: Can I ask for the help string over two lines with the first saying `request network quality of service` and the second giving the presets?

DavidSavinkoff: - Help string is now two lines, where the first line is a description and the second line is a usage.
- This option is for those few that want to select a different QoS for their network.

pljones (approved):
I'm going to approve (there's a single space I'd maybe have liked added before "default" in the help message).

DavidSavinkoff:
> I'm going to approve (there's a single space I'd maybe have liked added before "default" in the help message).

I left the 'space' out because 'default' looks ambiguous when it is on its own.
-       (e.g. --qos 0) DF=0 AF31=104 CS4=128 default EF=184
-       (e.g. --qos 0) DF=0 AF31=104 CS4=128-default EF=184
-       (e.g. --qos 0) DF=0 AF31=104 CS4=128default EF=184
I can change this to whatever you think is best.

pljones:
Yeah, that's why I thought it best left out... it would then have needed commas, which would have become clumsier.

src/main.cpp
```diff
@@ -823,6 +844,8 @@ QString UsageArguments ( char **argv )
         "                        supported for headless server mode)\n"
         "  -n, --nogui           disable GUI\n"
         "  -p, --port            set your local port number\n"
+        "  -Q, --qos             request quality of service in the IP header\n"
+        "                        (e.g. --qos 0) DF=0 AF31=104 CS4=128default EF=184\n"
```
hoffie, line 848: What about denoting the default like this?
```suggestion
        "                        common values: DF=0 AF31=104 CS4=128 EF=184\n"
        "                        default: --qos 128\n"
```

This makes it more verbose, of course, but I really had a hard time figuring out what "128default" is about, possibly because I do know QoS and TOS, but I certainly don't know the different classifications offhand. When reading "DF" in the network context, I initially thought about the "Don't fragment" bit...

hoffie (commented):
Thanks for your PR! Looks good to me. I have added a comment about a compiler warning I'm seeing.

Pinging @gilgongo as I think this should be documented?

src/socket.cpp
```diff
@@ -40,6 +40,9 @@ void CSocket::Init ( const quint16 iPortNumber )
 
     // create the UDP socket
     UdpSocket = socket ( AF_INET, SOCK_DGRAM, 0 );
+    //
+    const char tos = (const char) iQosNumber;  // Quality of Service
```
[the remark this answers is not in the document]
DavidSavinkoff: Thanks. I'll remove the 'const' because, after researching, it looks like the right thing to do.

src/main.cpp
DavidSavinkoff: My intention was to show that you can set --qos to 0, and that 0 corresponds to DF, and that there are more values you can use, and that you don't need to set --qos to 128 because that is the default value.

Instead of 'common values:' I would say suggested values, because there are 256 possible values where most are not used (you need a table).
Showing "default: --qos 128" may suggest that you need to set the default.

The reason the command line is used; is for those that are determined to not use CS4 (value 128), otherwise, this function can be ignored.

It is difficult to explain things in two lines, however, there is no harm in experimenting and researching to discover what can be done. I can add extra documentation elsewhere.
Discussion at #1072 has all of the documentation, but maybe, it should be diststilled and put on the wiki. Windows users will have to follow the directions in a picture on #1072 because Windows overrides this function to DF=0 (best effort priority) for all internet traffic.

I would prefer to leave the Jamulus --help for qos the way they are, and put the rest on the wiki.

ann0see:
Any updates here?

hoffie (approved):
Looks good to me now. Thanks!

@DavidSavinkoff
- Can you add yourself to [the contributors list](https://github.com/jamulussoftware/jamulus/blob/664fd9edfe4aee7db2951e8907675c5f885b0164/src/util.cpp#L490) as part of this PR? (I'm not merging the PR yet for this reason)
- As discussed, it would be good to have distilled docs about this. Can you add a PR against the website repo for this? Be sure to base your PR on the `changes` branch there.
  - Add the short description to [the command line flag overview](https://github.com/jamulussoftware/jamuluswebsite/edit/changes/wiki/en/en-Command-Line-Options.md)
  - From there, link to a longer explanation (distilled from the issue comments). I think the best place for this would be [Tips and Tricks](https://github.com/jamulussoftware/jamuluswebsite/edit/changes/wiki/en/en-Tips-Tricks-More.md) (@gilgongo or @ann0see may want to correct me if they know a better place).

Thanks again!

DavidSavinkoff:
When I added my name to the contributers list, util.cpp in my fork was older that the jamulussoftware:master.
I hope you know what to do.

PR #1325, opened 2021-03-23
Don't clean up ASIO stuff if it wasn't started

### LINKED ISSUE #872: Jamulus process doesn't get killed if you select a device with wrong sample rate (ASIO4ALL)
I think some if not all read cross errors in ASIO4ALL are related to an application competing with Jamulus. This lead me to the thought that Jamulus itself might be the problem sometimes. Inexperienced users will probably try Jamulus with ASIO4ALL by trial and error. 

Now my bug report:

Steps to reproduce:
0. Make sure only ASIO4ALL is installed on your PC
1. Select a device which doesn't support 48kHz in ASIO4ALL (I chose an output device) and the devices which work
2. Wait for the "No usable ASIO driver found... do you want to open the driver settings" warning
2. Click on yes and unselect the wrong device
3. Click on the warning which asks you if you configured your device correctly and click on yes
4. See the warning "Jamulus could not be started because of audio interface issues" (this shouldn't come up if you chose the correct devices!)
5. Click on quit
6. Open the Task manager (this will show you that a Jamulus process (at least as background process) is still running)
7. Open Jamulus. Sometimes (not always) ASIO4ALL now shows a red cross which disallows you to select the correct input/output device. (I think you have to unselect the working device to experience this issue).

To solve this problem, you have to find the non working Jamulus process in the Windows Taskmanager and kill it.

npostavs:
Possibly fixes #872. I can't reproduce that problem with ASIO4ALL itself, but I was getting segfaults in `ASIOStop()` and `ASIODisposeBuffers()` when I forced Jamulus to pretend the only available device sample rate is 44.1kHz.

### DIFF
--- a/windows/sound.cpp
+++ b/windows/sound.cpp
@@ -109,8 +109,15 @@ strCurDevName = cDriverNames[iDriverIdx];
 void CSound::UnloadCurrentDriver()
 {
     // clean up ASIO stuff
-    ASIOStop();
-    ASIODisposeBuffers();
+    if ( bRun )
+    {
+        Stop();
+    }
+    if ( bufferInfos[0].buffers[0] )
+    {
+        ASIODisposeBuffers();
+        bufferInfos[0].buffers[0] = NULL;
+    }
     ASIOExit();
     asioDrivers->removeCurrentDriver();
 }
@@ -524,6 +531,10 @@ CSound::CSound ( void           (*fpNewCallback) ( CVector<int16_t>& psData, voi
 
     // init pointer to our sound object
     pSound = this;
+
+    // We assume NULL'd pointers in this structure indicate that buffers are not
+    // allocated yet (see UnloadCurrentDriver).
+    memset ( bufferInfos, 0, sizeof bufferInfos );
 
     // get available ASIO driver names in system
     for ( i = 0; i < MAX_NUMBER_SOUND_CARDS; i++ )


ann0see:
Thank you! It does seem to fix the bug you mentioned. However I'll be doing some more tests here.

ann0see:
Ok. Can confirm this does fix the mentioned bug.

ann0see (approved):

windows/sound.cpp
```diff
@@ -109,8 +109,15 @@ strCurDevName = cDriverNames[iDriverIdx];
 void CSound::UnloadCurrentDriver()
 {
     // clean up ASIO stuff
-    ASIOStop();
-    ASIODisposeBuffers();
+    if ( bRun )
+    {
+        this->Stop();
+    }
+    if ( bufferInfos[0].buffers[0] )
```
pljones, line 116: What happens here if `bufferInfos[0]` is unset?  (Can it get here in that situation?)

npostavs: I clear all of `bufferInfos` in the constructor to make sure that it has a defined value.

softins: This is ok, because it is addressing part of the struct, not following a pointer
(it's not `bufferInfos[0]->buffers[0]`)

windows/sound.cpp
```diff
@@ -109,8 +109,15 @@ strCurDevName = cDriverNames[iDriverIdx];
 void CSound::UnloadCurrentDriver()
 {
     // clean up ASIO stuff
-    ASIOStop();
-    ASIODisposeBuffers();
+    if ( bRun )
+    {
+        this->Stop();
```
softins, line 114: This can just be `Stop();` can't it? (no need for the redundant `this->`)

npostavs: Oh right.

softins (approved):

Issue #1346, opened 2021-03-25
Selected inputs reset on connection to Server with v3.7.0


pgScorpio:
**Describe the bug**

Selected input channels always reset to the first two channels as soon as you connect to a server.

**To Reproduce**

Using ASIO Link Pro, with more than 2 channels, as device for Jamulus.
Now in Jamulus settings select, channels lets say. LinkIn3,LinkIn4 and LinkOut3, LinkOut4.
Now Connect to a server. 
As soon as the connection is established the selected channels revert to LinkIn1,LinkIn2 and LinkOut1, LinkOut2.
(After that you can select LinkIn3,LinkIn4 and LinkOut3, LinkOut4 again and it will work as expected.)

**Expected behavior**

No changing of selected channels when connecting to a server.

**Screenshots**

**Operating system**

Windows 10

Strangely enough it reproduces consequently on my PC:
Dutch Windows 10 Pro, build 19041
AMD Athlon II X2 215, 6GB RAM, 512GB SSD

But it doesn't reproduce on my laptop:
English Windows 10 Home, build 19042
AMD A4-5000, 8GB RAM, 1TB HDD

**Version of Jamulus**

jamulus v3.7.0 downloaded via Jamulus.io (https://github.com/jamulussoftware/jamulus/releases/download/r3_7_0/jamulus_3.7.0_win.exe)

**Additional context**

A lot of choirs use ASIO Link Pro to connect the audio between Jamulus and Zoom. (See my JamZoom-ALP.pdf document in the Jamulus Choral Community on Facebook.)
As far as I've heard until now about 60% to 70% of these ALP/Jamulus installs seem to show this problem with v3.7.0
Reverting back to Jamulus 3.6.2 always solves the problem.

ann0see:
Probably this is related to a bug fix which was done to solve a bug in ASIO4ALL. #796 

Channel mapping should only be reset if device properties change, not if you connect to a server (and I thought this would be the case). Not yet sure how we can solve this issue since the previous approach was even worse (and might have resulted in a crash of Jamulus if you change inputs/outputs in ASIO4ALL).

ann0see:
Maybe, not totally sure, we could re-implement the removed if and check if the number of channels changed. This might solve both bugs.

ann0see:
@pgScorpio once the compilation in my PR #1347  is finished, could you please check if the changes I made solved your issue?

pgScorpio:
[...]

Issue #796, opened 2020-12-26
Loud noise if Input is disabled in ASIO4ALL while still selected in Jamulus


ann0see:
If you select an input in ASIO4ALL, set it in Jamulus and later disable it in ASIO4All (and obviously can't in Jamulus since the in/output dialog is hidden) this results in loud noise on the unavailable input. This might be related to some buffers not being cleared.

Steps to reproduce:

1. Select an (internal) soundcard with multiple inputs: e.g. enable your integrated mic(s) and your headphone mic.
2. Set the left channel in Jamulus to your internal left mic and set your right channel to the mic of your headphones
3. Disable one of the microphone inputs in ASIO4ALL --> the input will no longer be available in Jamulus and you will hear a terrible loud noise on the input you disabled.
4. Enable the input: the sound disappears. 

Expected behaviour:

Debatable.

1. Give a warning that the inputs changed and tell the user to select the correct inputs/outputs in Jamulus;
2.  Optional: if now there's only one available, automatically select it (or ask the user if he wants to set this)

corrados:
I cannot reproduce the loud noise but when I enable/disable inputs/outputs in ASIO4All, the Jamulus software freezes and then crashes. So I think it makes sense to first try to fix the issue with the crash.

ann0see:
> the Jamulus software freezes and then crashes

oh. That’s even worse. Luckily it didn’t for me.

nico-dr:
My expierence is that this is related to exclusive access (the windows default) to the audio hardware 

I think you can workaround these kind of freeze by disabling the exclusive access in the windows settings.

![image](https://user-images.githubusercontent.com/76488290/103209935-8ed5b880-4904-11eb-84c6-d64bc527fd8a.png)

ann0see:
Do you know if there are any drawbacks once you disable exclusive access?

corrados:
> I cannot reproduce the loud noise but when I enable/disable inputs/outputs in ASIO4All, the Jamulus software freezes and then crashes.

I just debugged the issue. The problem is that the function ASIOCreateBuffers() in CSound::Init() suddenly blocks and does not return. When this happens, I then run into the MutexDriverReinit.lock() of CClient::OnSndCrdReinitRequest with the GUI thread and the GUI gets stuck.

[...]

PR #1347, opened 2021-03-25
Reset channel mapping if number of in/out changed

ann0see:
This will no longer reset channel mapping on every device change.

Fixes: https://github.com/jamulussoftware/jamulus/issues/1346 (at least I hope so)

With this https://github.com/jamulussoftware/jamulus/issues/796 does not occur

### DIFF
--- a/windows/sound.cpp
+++ b/windows/sound.cpp
@@ -60,13 +60,10 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName,
         return tr ( "The current selected audio device is no longer present in the system." );
     }
 
-    // Save number of channels from last driver (if we're not on the first run)
+    // Save number of channels from last driver
     // Need to save these (but not the driver name) as CheckDeviceCapabilities() overwrites them
-    if ( lNumInChanPrev != INVALID_INDEX && lNumOutChanPrev != INVALID_INDEX )
-    {
-        lNumInChanPrev = lNumInChan;
-        lNumOutChanPrev = lNumOutChan;
-    }
+    long lNumInChanPrev = lNumInChan;
+    long lNumOutChanPrev = lNumOutChan;
 
     loadAsioDriver ( cDriverNames[iDriverIdx] );
 
@@ -98,10 +95,6 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName,
 
             // mapping to the defaults (first two available channels)
             ResetChannelMapping();
-
-            // save the new number of in/out channels
-            lNumInChanPrev = lNumInChan;
-            lNumOutChanPrev = lNumOutChan;
 
             // store ID of selected driver if initialization was successful
             strCurDevName = cDriverNames[iDriverIdx];
--- a/windows/sound.cpp
+++ b/windows/sound.cpp
@@ -60,6 +60,14 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName,
         return tr ( "The current selected audio device is no longer present in the system." );
     }
 
+    // Save number of channels from last driver (if we're not on the first run)
+    // Need to save these (but not the driver name) as CheckDeviceCapabilities() overwrites them
+    if ( lNumInChanPrev != INVALID_INDEX && lNumOutChanPrev != INVALID_INDEX )
+    {
+        lNumInChanPrev = lNumInChan;
+        lNumOutChanPrev = lNumOutChan;
+    }
+
     loadAsioDriver ( cDriverNames[iDriverIdx] );
 
     // According to the docs, driverInfo.asioVersion and driverInfo.sysRef
@@ -80,15 +88,24 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName,
     // check if device is capable
     if ( strStat.isEmpty() )
     {
-// TODO: In order to fix https://github.com/jamulussoftware/jamulus/issues/796 we reset the channel mapping on every property change. This is not ideal. 
-	    
-// the device has changed, per definition we reset the channel
-// mapping to the defaults (first two available channels)
-ResetChannelMapping();
+        // Reset channel mapping if the sound card name has changed or the number of channels has changed
+        if ( ( strCurDevName.compare ( strDriverNames[iDriverIdx] ) != 0 ) ||
+             ( lNumInChanPrev != lNumInChan ) ||
+             ( lNumOutChanPrev != lNumOutChan ) )
+        {
+            // In order to fix https://github.com/jamulussoftware/jamulus/issues/796
+            // this code runs after a change in the ASIO driver (not when changing the ASIO input selection.)
 
-// store ID of selected driver if initialization was successful
-strCurDevName = cDriverNames[iDriverIdx];
+            // mapping to the defaults (first two available channels)
+            ResetChannelMapping();
 
+            // save the new number of in/out channels
+            lNumInChanPrev = lNumInChan;
+            lNumOutChanPrev = lNumOutChan;
+
+            // store ID of selected driver if initialization was successful
+            strCurDevName = cDriverNames[iDriverIdx];
+        }
     }
     else
     {
@@ -334,7 +351,7 @@ _exit(1);
         }
         else
         {
-            // ASIO SDK 2.2: "Notes: When minimum and maximum buffer size are 
+            // ASIO SDK 2.2: "Notes: When minimum and maximum buffer size are
             // equal, the preferred buffer size has to be the same value as
             // well; granularity should be 0 in this case."
             if ( HWBufferInfo.lMinSize == HWBufferInfo.lMaxSize )
--- a/windows/sound.cpp
+++ b/windows/sound.cpp
@@ -80,7 +80,7 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName,
     }
 
     // check device capabilities if it fulfills our requirements
-    const QString strStat = CheckDeviceCapabilities();
+    const QString strStat = CheckDeviceCapabilities(); // also sets lNumInChan and lNumOutChan
 
     // check if device is capable
     if ( strStat.isEmpty() )
--- a/windows/sound.h
+++ b/windows/sound.h
@@ -53,7 +53,7 @@ public:
              const QString& strMIDISetup,
              const bool     ,
              const QString& );
-    
+
     virtual ~CSound() { UnloadCurrentDriver(); }
 
     virtual int  Init ( const int iNewPrefMonoBufferSize );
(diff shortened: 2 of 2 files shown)


windows/sound.h
```diff
@@ -93,8 +93,10 @@ class CSound : public CSoundBase
     int              iASIOBufferSizeStereo;
 
     long             lNumInChan;
+    long             lNumInChanPrev = INVALID_INDEX; // number of channels of previous sound card/driver state
```
ann0see, line 96: Probably not optimal code wise?

windows/sound.cpp
```diff
@@ -60,6 +60,13 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName,
         return tr ( "The current selected audio device is no longer present in the system." );
     }
 
+    // save number of channels from last driver (if they are set); this is done to check if we need to reset channel mapping
```
ann0see, line 63: This could also be moved directly above CheckDeviceCapabilities() probably.

ann0see:
Thus bug (and some others which are not documented in the Changelog) might even warrant a new release...

windows/sound.cpp
```diff
@@ -80,15 +87,26 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName,
...
+             ( strCurDevName.compare ( strDriverNames[iDriverIdx] ) != 0 ) )
+        {
+            // TODO: In order to fix https://github.com/jamulussoftware/jamulus/issues/796
+            // we reset the channel mapping if the number of channels has changed or the device
+            // name has changed. This is not ideal.
+
+            // the device has changed, per definition we reset the channel
+            // mapping to the defaults (first two available channels)
+            ResetChannelMapping();
+            // save the new number of in/out channels
```
hoffie, line 102: Empty line should probably be before this comment and not after?

hoffie (approved):
Logic-wise, this sounds sane to me. I cannot test this though.

windows/sound.cpp
ann0see: Yes.

windows/sound.cpp
```diff
@@ -80,15 +87,26 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName,
...
-	    
-// the device has changed, per definition we reset the channel
-// mapping to the defaults (first two available channels)
-ResetChannelMapping();
-
-// store ID of selected driver if initialization was successful
-strCurDevName = cDriverNames[iDriverIdx];
-
+        // Reset channel mapping if the number of channels has changed or the sound card name has changed
+        if ( ( lNumInChanPrev != lNumInChan ) ||
```
pljones, line 91: I'd make the logic clearer:
- If _(_ the driver name changed _)_ _OR_ _(_ the indexes changed _)_

(and swap the comment around -- soundcard changing takes precedence over inputs).

windows/sound.cpp
```diff
@@ -80,15 +87,26 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName,
...
+             ( lNumOutChanPrev != lNumOutChan ) ||
+             ( strCurDevName.compare ( strDriverNames[iDriverIdx] ) != 0 ) )
+        {
+            // TODO: In order to fix https://github.com/jamulussoftware/jamulus/issues/796
+            // we reset the channel mapping if the number of channels has changed or the device
+            // name has changed. This is not ideal.
+
+            // the device has changed, per definition we reset the channel
```
pljones, line 99: I'd add a clarification here: this code runs after a change in the ASIO driver -- not when changing the ASIO input selection.  Hence, if I switch from "ASIO4ALL with four inputs" to "Echo AudioFire4 with four inputs", it will run -- I'll then have to switch to inputs 3&4 on the AudioFire, if that's what I want, as `ResetChannelMapping();` will revert from any previous settings I might have had.

(Maybe we need an array of settings per soundcard to get this working pleasantly...)

windows/sound.cpp
pljones, line 63: The `if` statement doesn't implement this comment.  The `if` says "save the number of channels if I haven't already saved the number of channels before".  What is the actual intent here?

ann0see: The intent here was to check if we're not on the first start. That's right.

Channel mapping might be reset on the first start if we don't do this (and that might be the bug we want to solve). INVALID_INDEX will never be the correct number of channels? 

I think the comment was added at the wrong position and might be outdated from one of my first tests.

pljones: Just writing down my thoughts...

So at launch, they're INVALID_INDEX.  `strCurDevName` appears to be playing a similar role to the `lNum*ChanPrev` pair here - hold the "old" value until we've established we're okay, then check what happened (gets initialised in both Windows CSound and base CSoundBase).

If 
- we acquire the driver successfully the first time through (first `CheckDeviceCapabilities();`) - `strStat.isEmpty()` is true
  and (
  - the driver name changed (which is should do on success the first time in)
    or
  - the number of ins or outs changed (which they should do on success the first time in)

  )

then we call `ResetChannelMapping();` and store the values (ins, outs, driver name).

Next time we acquire "some driver" (maybe the same one, maybe different), we save the "old" values num chan values here before they're overwritten by `CheckDeviceCapabilities();`.  That's different from the driver name because it doesn't get overwritten.

So then, we can check whether any value _changed_ and only call `ResetChannelMapping();` if it did.

----

Can you make the comment say "need to save these (but not the driver name), as `CheckDeviceCapabilities();` overwrites them" or words to that effect.  (The reason can be seen in the later code, I don't think it needs to be commented here.  Mentioning the driver name will make the association with the later `if` clear enough.)

I'm still not at all sure the `if` is needed.  Maybe `lNumInChan` and `lNumOutChan` should get set to `INVALID_INDEX` in ` windows/sound.h` instead?

ann0see: I think you're right with this analysis.

A few questions we might want to ask:

What happens and what should happen on the first invocation of LoadAndInitializeDriver()? Do we need to reset channel mapping or not? Did reloading channel mapping here result in the mentioned bug?


We set the channel mapping here for the first time, I think:

https://github.com/jamulussoftware/jamulus/blob/b398f7aaf8909ffbcb13c1970d0ae2960939efed/windows/sound.cpp#L590

What should the value of lNum*ChanPrev be on the first start?

pljones: If `lNum*Chan` are `INVALID_INDEX` first time through, then leaving the `if` out would mean `lNum*ChanPrev` get set to `INVALID_INDEX` here (regardless of what they were initialised to, so could default to 0.

If the driver is successfully acquired, `lNum*Chan` get the new numbers.  As that means a change happened, `ResetChannelMapping();` gets called (still first time through).

We then _don't need to store_ `lNum*Chan` inside the `ResetChannelMapping();` `if` block -- as they'll get stored here next time through.  ...  Assuming this logic is working of course...

Next time through, we store `lNum*Chan` here.  Then, if the driver name and `lNum*Chan` don't change, `ResetChannelMapping();` isn't called, otherwise it is (and the driver name is updated).

With this logic, `lNum*ChanPrev` seem to be _local variables_ in usage, too -- they don't need to persist across calls, just across the acquiring of the driver capabilities.

ann0see: > If lNum*Chan are INVALID_INDEX first time through, then leaving the if out would mean lNum*ChanPrev get set to INVALID_INDEX here (regardless of what they were initialised to, so could default to 0.

Yes.

> If the driver is successfully acquired, lNum*Chan get the new numbers. As that means a change happened, ResetChannelMapping(); gets called (still first time through).

Are you sure that this is what we want? Currently we don't do that.

ann0see: Hmm. Maybe was a bit confused... The comment should be clear now.

windows/sound.cpp
ann0see: > (Maybe we need an array of settings per soundcard to get this working pleasantly...)

Maybe in future

windows/sound.cpp
ann0see: @pljones will check your logic but not now.

ann0see: Ok. At least my bug doesn't seem to occur with this logic @pgScorpio @JoeSouthern

Could you please check if this binary works for you: https://github.com/jamulussoftware/jamulus/suites/2375362007/artifacts/50403146

windows/sound.cpp
```diff
@@ -60,6 +60,13 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName,
         return tr ( "The current selected audio device is no longer present in the system." );
     }
 
+    // Save number of channels from last driver
+    // Need to save these (but not the driver name) as CheckDeviceCapabilities() overwrites them
+
```
pljones, line 65: Sorry, tiny one: I'd drop the blank lines between the comment and what it's commenting (it's "nearer" the block before it as is).  Also, then, only the one line after the declares.

ann0see: Ok. Please re check

hoffie (approved):
I was a bit confused about the logic around `lNumInChan` as it looked like the value would not be modified anywhere between initializing `lNumInChanPrev` and checking for `lNumInChanPrev != lNumInChan`.
However, it seems like these `lNumInChan`and `lNumOutChan` are updated as part of `CheckDeviceCapabilities()`, which seems a bit surprising to me (assuming that a Check* function would not modify an instance member).
Maybe a comment in next to the call to `CheckDeviceCapabilities()` would clarify it, maybe it is just me.

This is nothing new however, it's just that reviewing the PR made it more obvious to me.

pljones:
> However, it seems like these `lNumInChan`and `lNumOutChan` are updated as part of `CheckDeviceCapabilities()`, which seems a bit surprising to me (assuming that a Check* function would not modify an instance member).
> Maybe a comment in next to the call to `CheckDeviceCapabilities()` would clarify it, maybe it is just me.

Nope, I had to expand it all to find it.  And yes, maybe a comment just to mention it for now.  That method name is poor, though.  I'm not sure, from the name, what it's scope is.

ann0see:
Ok. Added a brief comment

pljones (approved):

windows/sound.cpp
pgScorpio: > Ok. At least my bug doesn't seem to occur with this logic @pgScorpio @JoeSouthern
> 
> Could you please check if this binary works for you: https://github.com/jamulussoftware/jamulus/suites/2375362007/artifacts/50403146

Finally got some time to do a quick test and this version seems OK too, At least concerning the channel mapping reset.
I still have some comments on other issues though:
Window title:  Server info is moved to the end of the title (Great!), but this only takes effect when using hostname, without using hostname on the commandline the server info is still in front of "Jamulus".

PR #1360, opened 2021-03-26
Improve licence acceptance prompt dialog

pljones:
At the moment, it's not clear what control has focus when the licence acceptance dialog is initially displayed.  This puts focus on the only editable control - the checkbox.  (You don't want it on either button.)

I'm less certain about the other change - automatically moving focus when enabling "Accept".  You can still "just hit enter" as a keyboard user -- I'm just lazy and prefer to get away with "space", "space".

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -539,7 +539,9 @@ CLicenceDlg::CLicenceDlg ( QWidget* parent ) : CBaseDlg ( parent )
 
     QVBoxLayout*  pLayout    = new QVBoxLayout ( this );
     QHBoxLayout*  pSubLayout = new QHBoxLayout;
-    QLabel*       lblLicence = new QLabel ( tr ( "This server requires you accept conditions before you can join. Please read these in the chat window." ), this );
+    QLabel*       lblLicence = new QLabel ( tr ( "This server requires you accept conditions before you can join." ) +
+                                            "<br/><br/>" +
+                                            tr ( "Please read these in the chat window." ), this );
     QCheckBox*    chbAgree   = new QCheckBox ( tr ( "I have read the conditions and &agree." ), this );
     butAccept                = new QPushButton ( tr ( "Accept" ), this );
     QPushButton*  butDecline = new QPushButton ( tr ( "Decline" ), this );
@@ -554,6 +556,9 @@ CLicenceDlg::CLicenceDlg ( QWidget* parent ) : CBaseDlg ( parent )
     // set some properties
     butAccept->setEnabled ( false );
     butAccept->setDefault ( true );
+
+    QObject::connect ( this, &CLicenceDlg::show,
+        this, &CLicenceDlg::OnShow );
 
     QObject::connect ( chbAgree, &QCheckBox::stateChanged,
         this, &CLicenceDlg::OnAgreeStateChanged );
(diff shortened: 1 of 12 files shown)


src/util.cpp
```diff
@@ -538,8 +538,10 @@ CLicenceDlg::CLicenceDlg ( QWidget* parent ) : CBaseDlg ( parent )
 
     QVBoxLayout*  pLayout    = new QVBoxLayout ( this );
     QHBoxLayout*  pSubLayout = new QHBoxLayout;
-    QLabel*       lblLicence = new QLabel ( tr ( "This server requires you accept conditions before you can join. Please read these in the chat window." ), this );
-    QCheckBox*    chbAgree   = new QCheckBox ( tr ( "I have read the conditions and &agree." ), this );
+    QLabel*       lblLicence = new QLabel ( tr ( "This server requires you accept conditions before you can join." ) +
+                                            "<br/><br/>" +
+                                            tr ( "Please read these in the chat window." ), this );
```
softins, line 543: This splits the message in two, and will need the translations to be updated similarly. This kind of change could probably be done without understanding the translated language.

src/util.h
```diff
@@ -415,10 +415,12 @@ class CLicenceDlg : public CBaseDlg
     CLicenceDlg ( QWidget* parent = nullptr );
 
 protected:
+    QCheckBox*   chbAgree;
```
softins, line 418: Trying to understand why `chbAgree` has been moved into the class rather than being local to the constructor, as it was previously?

src/util.cpp
pljones: It's more a matter of understanding the translation system...

src/util.h
pljones: Because I _had_ been using it outside the constructor...  I thought I'd reverted all that change.

pljones:
OK, I've _manually_ updated the translation source files... but I didn't want to try to do the `.qm` files.  I'd like to get them done, too, really.  @softins ... what do I need to do..? :)

hoffie (approved):
Code looks good to me, translations seem to have been updated.

When connecting to a server which requires a license to be accepted, the dialog will be focused (even with 3.6.2). When pressing the space key, the checkbox gets checked (even with 3.6.2). I somehow would have expected that this was different before the PR (I'm testing on Arch/sway).

When pressing space again, the Accept button gets pressed (only with this PR). Both previous and new behavior would be ok for me.

pljones:
On Windows, focus appears to start "nowhere".  Tab and backtab both move focus to "Decline".  From there, either tab or backtab gets to the checkbox.  Here's the old flow:
image
(nothing focussed)
image
(tab once)
image
(tab twice)

pljones:
Mmm, trying it on Ubuntu... yeah, seems to default to the first control in the old code.  And even with the new code, it doesn't clearly highlight the focused field like on Windows.

pljones:
I've opened https://github.com/pljones/jamulus/pull/8 with a couple of extra tweaks to make Escape functional again where it should be - hopefully without breaking the Android fix from @nefarius2001.  I'll raise the PR to jamulussoftware/jamulus once this one's merged.

pljones:
Strange.  When I try it now on Windows r3.7.0, it seems focus is in the right place.  I'm going to close this...

PR #1370, opened 2021-03-27
Client name added to dialog window titles  #1352

dcorson-ticino-com:
routine to make title string in util
4 calls in clientdlg, one for each dialog when it is shown:
Connect, Chat, Musician Profile and Settings

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -49,7 +49,6 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 {
     setupUi ( this );
 
-
     // Add help text to controls -----------------------------------------------
     // input level meter
     QString strInpLevH = "<b>" + tr ( "Input Level Meter" ) + ":</b> " + tr ( "This shows "
@@ -1004,6 +1003,7 @@ void CClientDlg::ShowConnectionSetupDialog()
     // show connect dialog
     bConnectDlgWasShown = true;
     ConnectDlg.show();
+    ConnectDlg.setWindowTitle ( MakeClientNameTitle( tr ( "Connect" ) , pClient->strClientName ) );
 
     // make sure dialog is upfront and has focus
     ConnectDlg.raise();
@@ -1014,6 +1014,7 @@ void CClientDlg::ShowMusicianProfileDialog()
 {
     // show musician profile dialog
     MusicianProfileDlg.show();
+    MusicianProfileDlg.setWindowTitle ( MakeClientNameTitle( tr ( "Musician Profile" ) , pClient->strClientName ) );
 
     // make sure dialog is upfront and has focus
     MusicianProfileDlg.raise();
@@ -1024,6 +1025,7 @@ void CClientDlg::ShowGeneralSettings()
 {
     // open general settings dialog
     ClientSettingsDlg.show();
+    ClientSettingsDlg.setWindowTitle ( MakeClientNameTitle( tr ( "Settings" ) , pClient->strClientName ) );
 
     // make sure dialog is upfront and has focus
     ClientSettingsDlg.raise();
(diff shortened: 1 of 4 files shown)


src/clientdlg.cpp
```diff
@@ -1001,6 +1000,7 @@ void CClientDlg::ShowConnectionSetupDialog()
     // show connect dialog
     bConnectDlgWasShown = true;
     ConnectDlg.show();
+    ConnectDlg.setWindowTitle ( MakeClientNameTitle( tr ( "Connect" ) , pClient->strClientName ) );
```
pljones, line 1003: Sorry, spaces after the method name and before the `(`, too, please, in `MakeClientNameTitle(` - throughout.

src/util.cpp
```diff
@@ -1581,3 +1580,13 @@ QString GetVersionAndNameStr ( const bool bWithHtml )
 
     return strVersionText;
 }
+
+QString MakeClientNameTitle ( QString win, QString client )
```
pljones, line 1584: Hm.  I thought we had client name first.  But we don't.  It's "window name clientname" when the client name is specified.  The change to the mixer window was to move the appname to the end -- but _not_ if clientname has been used.  Which is a bit weird but... this is following the same pattern, so fine :).

src/clientdlg.cpp
dcorson-ticino-com: Fine with me, will do, I was just orienting on the style in the clientdlg constructor.

src/clientdlg.cpp
```diff
@@ -1001,6 +1000,7 @@ void CClientDlg::ShowConnectionSetupDialog()
     // show connect dialog
     bConnectDlgWasShown = true;
     ConnectDlg.show();
+    ConnectDlg.setWindowTitle( MakeClientNameTitle( tr ( "Connect" ) , pClient->strClientName ) );
```
pljones, line 1003: > Sorry, spaces after the method name and before the `(`, too, please

You've removed the one that was before the `(`.  It needs a space both sides.  (Yes, I phrased it a bit clumsily now I read it back.)

src/clientdlg.cpp
```diff
@@ -1000,46 +999,50 @@ void CClientDlg::ShowConnectionSetupDialog()
...
     bConnectDlgWasShown = true;
-    ConnectDlg.show();
+    ConnectDlg.show ();
+    ConnectDlg.setWindowTitle ( MakeClientNameTitle ( tr ( "Connect" ) , pClient->strClientName ) );
 
     // make sure dialog is upfront and has focus
-    ConnectDlg.raise();
-    ConnectDlg.activateWindow();
+    ConnectDlg.raise ();
+    ConnectDlg.activateWindow ();
```
softins, line 1007: I know the formatting style can be confusing! But when submitting a PR, there is usually no need to reformat lines that you haven't changed.

In this case, the "house style" seems to be _no_ space before an empty argument list `()`, but a space before and after an opening `(` when there are arguments to the function.

ann0see:
Any updates here?

softins (approved):
Compiled and tested ok, and rebased on master to resolve conflicts.

src/clientdlg.cpp
```diff
@@ -948,7 +947,7 @@ void CClientDlg::SetMyWindowTitle ( const int iNumClients )
     {
         // if --clientname is used, the APP_NAME must be the very first word in
         // the title, otherwise some user scripts do not work anymore, see #789
-        strWinTitle += QString ( APP_NAME ) + " " + pClient->strClientName + " ";
+        strWinTitle += QString ( APP_NAME ) + " - " + pClient->strClientName + " ";
```
pljones, line 950: This block (947 to 981) all seems a little clumsy to me and I have to re-read it several times to get the gist.  I think it probably wants a refactor at some point.  Not this change, though.

pljones (approved):

Issue #1338, opened 2021-03-24
Do not quit Jamulus if no valid audio device was found


ann0see:
**Has this feature been discussed and generally agreed?**
No. But a I assume many people are aware of it

**Describe the solution you'd like**

If you select devices with a wrong sample rate in e.g. ASIO4ALL or misconfigure devices, Jamulus currently handles this as fatal error (See soundbase.cpp SetDev). Users of ASIO4ALL might select a device which doesn't support 48kHz. This results in a warning and a fatal error. Jamulus will then ask you to close the app, even if the driver settings are open. Also you can't even open Jamulus if something is wrong with your audio devices since it quits with an error message. #305 fixed some parts of this issue, but in my opinion the current handling is not user friendly. I'd expect:

1. A clear error message that the current device is not supported in this state which doesn't disappear unless you fixed the issue.
2. A clear hint how to fix this issue (but sometimes that's not easy)
3. Jamulus not being quit. This confuses a lot of users. In my opinion it's not an error which should result in Jamulus being quit. Maybe we could even allow the user to change basic settings or connect to a server with chat only?

pgScorpio:
Same goes for automatically selecting another device if the configured device is not available (i.e. forgot to plug in.)
This should only give a warning. Do not change any settings automatically, since this confuses a lot of users.

Issue #882, opened 2021-01-23
Local-Loopback mode for audio driver test


nefarius2001:
When setting up Jamulus for someone new, choosing the right audio settings is the most crucial step (which interface, is my headset good enough, etc)

A loopback mode would make this independant from any network-related stuff. I can imagine this two ways:
1) separate button: check it and local-loopback is enabled, uncheck and it is gone. Server-connection overwrites this. Checkbox and a level-regulator are available in settings dialog.
2) make this default disconnected behaviour: having a loopback active in disconnected state makes the transition to "connected" more continuous, since my own sound was already there. A "client-stream" dialog is shown in the main window just as if you were connected to a server with only you as client. 

(This would also solve issues with running Discord in parallel to jamulus, where in "disconnected" state the Discord-Sound is unavailable, after connecting to a Jamulus-Server the Discord-Sound is available again. Inconvenient for remote tech-support)

This goes in line with the request to have Sound-Level display in disconnected state, but takes it further to also outputting the local input to the local output for system-setup before joining sessions & servers,

nefarius2001:
Related to #712

also I see a benefit in telling local and network related buffet underruns apart.

Please state if such feature is actively discouraged, or just not interesting but will be accepted if implemented by someone.

WolfganP:
I really see value in this feature request, as a local loop (ie jamulus client getting audio from the audio device, generate the packets, loopback and decode them and play on the audio interface) will help new users to make sure everything local is working (and avoid audio test with local audio device loopback, which doesn't guarantee is picked up by jamulus later and/or going to the server).
And it will certainly provide a more seamless transition to going online.
Good one!

dcorson-ticino-com:
My experience getting people up-and-running is that the problem is getting the input volume set correctly, not using a server to get the loop-back. There are enough empty servers or use the server you are planning on joining.
Of course this would be a nice feature, but I'm not sure it will really help getting people set up correctly.

DavidSavinkoff:
Note that it is normal for a local client to connect to the loopback interface (localhost 127.0.0.1) when you run a local server (do not use the server's external ip-address because it causes unnecessary ip traffic in this case).
This is already done with:  Jamulus client :: Connect :: Server Address > 127.0.0.1
Problem Solved.

nefarius2001:
> Problem Solved.

[...]

Issue #727, opened 2020-11-16
MacOS: If an audio device is no longer available, show an error rather than switching to Default


wareynolds96:
Hi There,
Hope you are doing well. I was just wondering if anyone else has come across this?
We are using for our input channel mapping:
L: Analog 5
R: Analog 6
We are using for our output channel mapping:
L: Analog 3
R: Analog 4
We are using this via totalmix as well so we can route back audio between two computers and have multiple inputs.
The issue we are having is that when we are in the middle of a session Jamulus changes the input and output channel mapping to 1 & 2 which completely kills any signal being able to be passed through.
I am not sure if this is a new thing or a bug related to version 3.6.0. but could someone help us out? Is there a workaround or is it something we can get fixed quickly or is there something we are missing?
We are open to any ideas of help and if we are wrong that is great to know. Just want to get it sorted out really.
Thanks so much for your time!
Will

wareynolds96:
@corrados just posting this here.

Thanks

W

WolfganP:
This repository is rather for developers. You may be getting better help on the users forums https://sourceforge.net/p/llcon/discussion/ and https://www.facebook.com/groups/619274602254947/

corrados:
@WolfganP There is already a discussion going on in the forums: https://sourceforge.net/p/llcon/discussion/software/thread/1e69217ce0

corrados:
> Jamulus changes the input and output channel mapping to 1 & 2 which completely kills any signal being able to be passed through

Jamulus only resets the channel mapping when a new audio device is selected.

> I am not sure if this is a new thing or a bug related to version 3.6.0

There were no changes to the audio channel mapping done in 3.6.0 so I do not think this is a specific problem of that version.

From the forum discussion on Sourceforge:
> What I have seen is that if Jamulus can't access the selected interface it picks 1&2 of the first interface it finds.
Could it be that your interface is having intermittent dropouts ?

Can you please answer this question?

wareynolds96:
Hey Volker,

Thanks so much for this info. To answer your question no we are not having
dropouts at all. We are trying to use this in conjunction with Logic Pro X
and Pro Tools 2020.11. It seems to happen intermittently when the DAW is
open. Is there any way to lock the I/O or save the I/O so that it never
changes?

The thing that is a bit confusing is when you click on the box of the
Jamulus I/O it looks as if it is still on 3&4 but has actually gone back to
1&2.

Hope that helps a little.

Thanks so much for your time

Will





[...]

PR #1378, opened 2021-03-28
Don't insist that selected audio device is working

npostavs:
**EDIT by maintainers:** In order to move this forward, we'd need someone with macOS background who can help debugging

```
That is, don't immediately try to select a different working device if
the current one fails to work.
```

This PR is intended to make the discussion in #1338 a bit more concrete. I think this basically handles points (1) and (3) from there. I didn't make any attempt at (2); none of the messages are different.

Here's a screenshot of what happens when I've selected a device that's turned off.

uninitialized-device

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -1163,10 +1163,7 @@ void CClientDlg::OnSoundDeviceChanged ( QString strError )
     if ( !strError.isEmpty() )
     {
         // the sound device setup has a problem, disconnect any active connection
-        if ( pClient->IsRunning() )
-        {
-            Disconnect();
-        }
+        Disconnect();
 
         // show the error message of the device setup
         QMessageBox::critical ( this, APP_NAME, strError, tr ( "Ok" ), nullptr );
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -1212,10 +1212,11 @@ void CClientDlg::Connect ( const QString& strSelectedAddress, const QString& str
         catch ( const CGenErr& generr )
         {
             // show error message and return the function
-            QMessageBox msgBox ( QMessageBox::Critical, APP_NAME, generr.GetErrorText(),
-                                 QMessageBox::NoButton, this );
+            QMessageBox  msgBox ( QMessageBox::Critical, APP_NAME, generr.GetErrorText(), QMessageBox::NoButton, this );
             QPushButton* openSettingsButton = msgBox.addButton ( tr ( "Open Settings" ), QMessageBox::ActionRole );
-            QPushButton* autoSelectButton = msgBox.addButton ( tr ( "Auto Select Device" ), QMessageBox::ActionRole );
+            // See also CClientSettingsDlg::UpdateSoundDeviceChannelSelectionFrame()
+            QPushButton* autoSelectButton =
+                ( pClient->GetSndCrdDevNames().count() > 1 ) ? msgBox.addButton ( tr ( "Auto Select Device" ), QMessageBox::ActionRole ) : nullptr;
             QPushButton* closeButton = msgBox.addButton ( tr ( "Close" ), QMessageBox::AcceptRole );
 
             msgBox.setTextFormat ( Qt::RichText ); // Some error messages are HTML formatted
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -1212,7 +1212,25 @@ void CClientDlg::Connect ( const QString& strSelectedAddress, const QString& str
         catch ( const CGenErr& generr )
         {
             // show error message and return the function
-            QMessageBox::critical ( this, APP_NAME, generr.GetErrorText(), "Close", nullptr );
+            QMessageBox msgBox ( QMessageBox::Critical, APP_NAME, generr.GetErrorText(),
+                                 QMessageBox::NoButton, this );
+            QPushButton* openSettingsButton = msgBox.addButton ( tr ( "Open Settings" ), QMessageBox::ActionRole );
+            QPushButton* autoSelectButton = msgBox.addButton ( tr ( "Auto Select Device" ), QMessageBox::ActionRole );
+            QPushButton* closeButton = msgBox.addButton ( tr ( "Close" ), QMessageBox::AcceptRole );
+
+            msgBox.setTextFormat ( Qt::RichText ); // Some error messages are HTML formatted
+            msgBox.setDefaultButton ( openSettingsButton );
+            msgBox.setEscapeButton ( closeButton );
+
+            msgBox.exec();
+            if ( msgBox.clickedButton() == openSettingsButton )
+            {
+                ShowGeneralSettings ( SETTING_TAB_AUDIONET );
+            }
+            else if ( msgBox.clickedButton() == autoSelectButton )
+            {
+                ClientSettingsDlg.OnTryLoadAnyDriverClicked();
+            }
             return;
         }
 
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -263,6 +263,8 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     lblGlobalInfoLabel->setStyleSheet ( ".QLabel { background: red; }" );
     lblGlobalInfoLabel->hide();
 
+    SetErrorMessage ( pNSetP->strLoadErrors );
+
     // prepare update check info label (invisible by default)
     lblUpdateCheck->setOpenExternalLinks ( true ); // enables opening a web browser if one clicks on a html link
     lblUpdateCheck->setText ( "<font color=\"red\"><b>" + APP_UPGRADE_AVAILABLE_MSG_TEXT.arg ( APP_NAME ).arg ( VERSION ) + "</b></font>" );
@@ -1041,6 +1043,12 @@ void CClientDlg::OnLocalMuteStateChanged ( int value )
     }
 }
 
+void CClientDlg::SetErrorMessage ( const QString& message )
+{
+    lblGlobalErrorMessage->setText ( message );
+    lblGlobalErrorMessage->setVisible ( !message.isEmpty() );
+}
+
 void CClientDlg::OnTimerSigMet()
 {
     // show current level
@@ -1175,6 +1183,8 @@ void CClientDlg::OnSoundDeviceChanged ( QString strError )
         TimerDetectFeedback.start ( DETECT_FEEDBACK_TIME_MS );
         bDetectFeedback = true;
     }
+
+    SetErrorMessage ( strError );
 
     // update the settings dialog
     ClientSettingsDlg.SetDeviceErrors ( strError );
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -1177,6 +1177,7 @@ void CClientDlg::OnSoundDeviceChanged ( QString strError )
     }
 
     // update the settings dialog
+    ClientSettingsDlg.SetDeviceErrors ( strError );
     ClientSettingsDlg.UpdateSoundDeviceChannelSelectionFrame();
 }
 
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -695,6 +695,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     QObject::connect ( butDriverSetup, &QPushButton::clicked, this, &CClientSettingsDlg::OnDriverSetupClicked );
 #endif
     QObject::connect ( butTryLoadAnyDriver, &QPushButton::clicked, this, &CClientSettingsDlg::OnTryLoadAnyDriverClicked );
+    QObject::connect ( butDriverReset, &QPushButton::clicked, this, &CClientSettingsDlg::OnDriverResetClicked );
 
     // misc
     // sliders
@@ -910,6 +911,8 @@ void CClientSettingsDlg::SetEnableFeedbackDetection ( bool enable )
 #if defined( _WIN32 ) && !defined( WITH_JACK )
 void CClientSettingsDlg::OnDriverSetupClicked() { pClient->OpenSndCrdDriverSetup(); }
 #endif
+
+void CClientSettingsDlg::OnDriverResetClicked() { pClient->SetSndCrdDev ( pClient->GetSndCrdDev() ); }
 
 void CClientSettingsDlg::OnNetBufValueChanged ( int value )
 {
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -844,7 +844,9 @@ void CClientSettingsDlg::UpdateSoundDeviceChannelSelectionFrame()
         cbxSoundcard->addItem ( strDevName );
     }
 
-    cbxSoundcard->setCurrentText ( pClient->GetSndCrdDev() );
+    const QString& sSndCrdName = pClient->GetSndCrdDev();
+    cbxSoundcard->setCurrentText ( sSndCrdName );
+    butDriverSetup->setEnabled ( !sSndCrdName.isEmpty() );
 
     // update input/output channel selection
 #if defined( _WIN32 ) || defined( __APPLE__ ) || defined( __MACOSX )
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -846,6 +846,9 @@ void CClientSettingsDlg::UpdateSoundDeviceChannelSelectionFrame()
     {
         cbxSoundcard->addItem ( strDevName );
     }
+    // Selecting devices automatically only makes sense if there is more than
+    // one to choose from.
+    butTryLoadAnyDriver->setVisible ( slSndCrdDevNames.count() > 1 );
 
     const QString& sSndCrdName = pClient->GetSndCrdDev();
     cbxSoundcard->setCurrentText ( sSndCrdName );
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -145,6 +145,9 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                TOOLTIP_COM_END_TEXT );
 #    endif
 
+    lblErrors->setText ( pSettings->strLoadErrors );
+    lblErrors->setWordWrap ( true );
+
     // sound card input/output channel mapping
     QString strSndCrdChanMapp = "<b>" + tr ( "Sound Card Channel Mapping" ) + ":</b> " +
                                 tr ( "If the selected sound card device offers more than one "
@@ -784,6 +787,8 @@ QString CClientSettingsDlg::GenSndCrdBufferDelayString ( const int iFrameSize, c
     return QString().setNum ( static_cast<double> ( iFrameSize ) * 2 * 1000 / SYSTEM_SAMPLE_RATE_HZ, 'f', 2 ) + " ms (" +
            QString().setNum ( iFrameSize ) + strAddText + ")";
 }
+
+void CClientSettingsDlg::SetDeviceErrors ( const QString& strError ) { lblErrors->setText ( strError ); }
 
 void CClientSettingsDlg::UpdateSoundCardFrame()
 {
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -264,6 +264,8 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     butDriverSetup->setToolTip ( strSndCardDriverSetupTT );
 #endif
 
+    butDriverReset->setIcon ( butDriverReset->style()->standardIcon ( QStyle::SP_BrowserReload ) );
+
     // fancy skin
     lblSkin->setWhatsThis ( "<b>" + tr ( "Skin" ) + ":</b> " + tr ( "Select the skin to be used for the main window." ) );
 
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -694,6 +694,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     // Driver Setup button is only available for Windows when JACK is not used
     QObject::connect ( butDriverSetup, &QPushButton::clicked, this, &CClientSettingsDlg::OnDriverSetupClicked );
 #endif
+    QObject::connect ( butTryLoadAnyDriver, &QPushButton::clicked, this, &CClientSettingsDlg::OnTryLoadAnyDriverClicked );
 
     // misc
     // sliders
@@ -918,6 +919,14 @@ void CClientSettingsDlg::OnNetBufServerValueChanged ( int value )
 {
     pClient->SetServerSockBufNumFrames ( value );
     UpdateJitterBufferFrame();
+}
+
+void CClientSettingsDlg::OnTryLoadAnyDriverClicked()
+{
+    pClient->TryLoadAnyDev();
+
+    UpdateSoundDeviceChannelSelectionFrame();
+    UpdateDisplay();
 }
 
 void CClientSettingsDlg::OnSoundcardActivated ( int iSndDevIdx )
--- a/src/soundbase.cpp
+++ b/src/soundbase.cpp
@@ -102,7 +102,7 @@ QString CSoundBase::SetDev ( const QString strDevName )
     return LoadAndInitializeDriver ( strDevName, false );
 }
 
-QVector<QString> CSoundBase::LoadAndInitializeFirstValidDriver ( const bool bOpenDriverSetup )
+QString CSoundBase::LoadAndInitializeFirstValidDriver ( const bool bOpenDriverSetup )
 {
     QVector<QString> vsErrorList;
 
@@ -134,7 +134,24 @@ QVector<QString> CSoundBase::LoadAndInitializeFirstValidDriver ( const bool bOpe
         iDriverCnt++;
     }
 
-    return vsErrorList;
+    if ( !vsErrorList.isEmpty() )
+    {
+        // create error message with all details
+        QString sErrorMessage = "<b>" + tr ( "No usable " ) + strSystemDriverTechniqueName + tr ( " audio device (driver) found." ) + "</b><br><br>" +
+                                tr ( "In the following there is a list of all available drivers "
+                                     "with the associated error message:" ) +
+                                "<ul>";
+
+        for ( int i = 0; i < lNumDevs; i++ )
+        {
+            sErrorMessage += "<li><b>" + GetDeviceName ( i ) + "</b>: " + vsErrorList[i] + "</li>";
+        }
+        sErrorMessage += "</ul>";
+
+        return sErrorMessage;
+    }
+
+    return "";
 }
 
 /******************************************************************************\
--- a/src/soundbase.cpp
+++ b/src/soundbase.cpp
@@ -97,8 +97,14 @@ QString CSoundBase::SetDev ( const QString strDevName )
 {
     QMutexLocker locker ( &MutexDevProperties );
 
-    strCurDevName = strDevName; // Update current device, even if loading fails.
+    if ( !strCurDevName.isEmpty() )
+    {
+        // a device was already been initialized and is used, first clean up
+        // driver
+        UnloadCurrentDriver();
+    }
 
+    strCurDevName = strDevName; // Update current device, even if loading fails.
     return LoadAndInitializeDriver ( strDevName, false );
 }
 
@@ -106,32 +112,26 @@ QString CSoundBase::LoadAndInitializeFirstValidDriver ( const bool bOpenDriverSe
 {
     QVector<QString> vsErrorList;
 
-    // load and initialize first valid ASIO driver
-    bool bValidDriverDetected = false;
-    int  iDriverCnt           = 0;
-
     // try all available drivers in the system ("lNumDevs" devices)
-    while ( !bValidDriverDetected && ( iDriverCnt < lNumDevs ) )
+    for ( int iDriverCnt = 0; iDriverCnt < lNumDevs; iDriverCnt++ )
     {
+        if ( !strCurDevName.isEmpty() )
+        {
+            UnloadCurrentDriver();
+        }
+
         // try to load and initialize current driver, store error message
-        const QString strCurError = LoadAndInitializeDriver ( GetDeviceName ( iDriverCnt ), bOpenDriverSetup );
+        strCurDevName             = GetDeviceName ( iDriverCnt );
+        const QString strCurError = LoadAndInitializeDriver ( strCurDevName, bOpenDriverSetup );
 
         vsErrorList.append ( strCurError );
 
         if ( strCurError.isEmpty() )
         {
-            // initialization was successful
-            bValidDriverDetected = true;
-
-            // store ID of selected driver
-            strCurDevName = GetDeviceName ( iDriverCnt );
-
             // empty error list shows that init was successful
             vsErrorList.clear();
+            break;
         }
-
-        // try next driver
-        iDriverCnt++;
     }
 
     if ( !vsErrorList.isEmpty() )
@@ -148,6 +148,7 @@ QString CSoundBase::LoadAndInitializeFirstValidDriver ( const bool bOpenDriverSe
         }
         sErrorMessage += "</ul>";
 
+        strCurDevName = ""; // clear device selection.
         return sErrorMessage;
     }
 
--- a/src/soundbase.cpp
+++ b/src/soundbase.cpp
@@ -76,10 +76,18 @@ QStringList CSoundBase::GetDevNames()
 
     QStringList slDevNames;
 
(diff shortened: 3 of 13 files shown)


ann0see:
Interesting approach. Nevertheless removing all the error message might be counter productive: "While configuring ASIO4ALL: Why don't I hear sound anymore? Without ASIO Driver: Why do I get an error message if I click connect?" Users would probably not look at settings on a first run.

Would it be possible to show the error messages, initialise a disconnect and ask the user if he wants to quit Jamulus or open Settings to change the device (settings)?

src/soundbase.cpp
```diff
@@ -94,114 +102,9 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
-                "check the audio device setting." );
-        }
-
-        // try to load and initialize any valid driver
-        QVector<QString> vsErrorList = LoadAndInitializeFirstValidDriver();
-
-        if ( !vsErrorList.isEmpty() )
-        {
-            // create error message with all details
-            QString sErrorMessage = "<b>" + tr ( "No usable " ) +
```
ann0see, line 174: I think this error should still be very prominent

src/soundbase.cpp
```diff
@@ -94,114 +102,9 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
-        {
-            strReturn = tr ( "The previously selected audio device "
-                "is no longer available or the audio driver properties have changed to a state which "
-                "is incompatible with this software. We now try to find a valid audio device. This new "
-                "audio device might cause audio feedback. So, before connecting to a server, please "
-                "check the audio device setting." );
-        }
-
-        // try to load and initialize any valid driver
-        QVector<QString> vsErrorList = LoadAndInitializeFirstValidDriver();
```
ann0see, line 169: Can we ask the user if he wants to do this?

ann0see:
Just tested what happens if you connect to a server with an invalid ASIO Driver. Jamulus crashes. This shouldn't happen. Can we give an error message if the user tries to connect with invalid drivers?

ann0see (changes requested):
Yep. Also on macOS I now get an error and no sound.

pljones:
> Just tested what happens if you connect to a server with an invalid ASIO Driver. Jamulus crashes. This shouldn't happen. Can we give an error message if the user tries to connect with invalid drivers?

It shouldn't try to connect -- connection shouldn't be possible without a working sound driver.  In fact, the mixer screen isn't much use: the user should be in the settings screen until they've got it sorted, I think.

> Also on Linux this now results in Jamulus not connecting to JACK. I assume this is a much bigger topic :-(.

I've always avoided looking at the soundcard code... :)

gilgongo:
Is this PR also related to https://github.com/jamulussoftware/jamulus/issues/727 ?

pljones:
> Is this PR also related to #727 ?

The whole way Jamulus handles the audio hardware is a bit problematic, really.  It _requires_ there to be a functional audio device, even it it's not using it.  Whilst the "require" is in the common code, the "provide" is platform-dependent.  So each of the "it doesn't work if..." scenarios has a different solution ... if you look at them one way.

Looked at another - rewrite how Jamulus acquires the device - they may have the same solution...

npostavs:
How about showing the error in the main window as well? (I've reused the mute message label, not sure if it's better to have two separate labels and swap between them instead).

> Just tested what happens if you connect to a server with an invalid ASIO Driver. Jamulus crashes. This shouldn't happen.

Hmm, doesn't happen for me. I get an error message after a few seconds of "TRYING TO CONNECT".

image

src/soundbase.cpp
npostavs: I've added a button in the settings window to do this.

ann0see:
I think we shouldn't abuse the muted message. 

Doing some tests now.

ann0see:
Ok. The crash only occurs if we only have one device (ASIO4ALL) which is wrongly configured and I connect to a server

npostavs:
> Ok. The crash only occurs if we only have one device (ASIO4ALL) which is wrongly configured and I connect to a server

Hmm, I'm not sure how to reproduce that on my machine unfortunately.

npostavs:
When changing to use a separate label I bumped into several cases where I got a segfault. Actually, I'm a bit confused how I didn't see them before. Anyway those should be fixed now, and I also disabled the "Connect" button when audio isn't working.

ann0see:
Thanks!

> Anyway those should be fixed now, and I also disabled the "Connect" button when audio isn't working.

I think we should not disable it (since you can just open the connect dialog via "view"). Can you g give a warning (You can't connect because your audio device isn't working: <Error> Do you want to open Jamulus's settings or close Jamulus?) right before it actually starts the connection process? This then catches all cases (and probably CLI only clients too.

npostavs:
>  think we should not disable it (since you can just open the connect dialog via "view").

Ah right. Throwing an exception seems to handle both cases. I'm not sure about CLI clients though. Is it possible to change devices without a GUI?

ann0see:
> Is it possible to change devices without a GUI?

Not sure. I think it's not possible on Windows at least. But I don't know honestly.

ann0see:
Ok. Just tested it. Unfortunately double clicking a server still tries to connect. Maybe you should add it in the Sound class? Not looked into the code yet, but it should happen in CSound or in Init()? 

So all the code used to be in CSoundBase::SetDev which used to throw an exception in line 200 if something went wrong: https://github.com/jamulussoftware/jamulus/pull/1378/files#diff-3826707a9164c8d7018040fddfb703fe2a72cb7777a9f1866c551a3866efa237L200

What if we catch this exception, initiate a disconnect (or disable audio) and throw a warning message which asks if you want to open settings?

Also the error message "previous sound device no longer present in system" is not really an error. We might give a warning message and still initiate an automatic device selection.

I'm really afraid of this PR since it seems to change complex core functionality.

npostavs:
> Ok. Just tested it. Unfortunately double clicking a server still tries to connect.

I can't reproduce that.

> Also the error message "previous sound device no longer present in system" is not really an error. We might give a warning message and still initiate an automatic device selection.

I disagree very strongly with this. Automatically switching to a different device because the previous happens to not be working at the moment causes a lot of confusion to people I've played with.

ann0see:
> I disagree very strongly with this. Automatically switching to a different device because the previous happens to not be working at the moment causes a lot of confusion to people I've played with.

Understand your point. However what about a deliberate error message with a prominent "Yes, change my audio device" or "Open settings to select another audio device" button?




>   Ok. Just tested it. Unfortunately double clicking a server still tries to connect.
>> I can't reproduce that.

Strange. At I tried it on Windows 10 with ASIO4ALL only and it did happen. Can you try it in a VM?

npostavs:
>  what about a deliberate error message with a prominent "Yes, change my audio device" or "Open settings to select another audio device" button?

Yeah, I think if the user makes the choice it should be okay.

> Strange. At I tried it on Windows 10 with ASIO4ALL only and it did happen. Can you try it in a VM?

Um, not easily. But since `CConnectDlg::OnServerListItemDoubleClicked` does nothing but call `OnConnectClicked`, I can't see how double clicking could behave any different. I'm on Windows 10 as well (though not using ASIO4ALL since I haven't been able to get it to disable). Just to be clear, you're not getting anything like this:

image

ann0see:
No. If I connect to a server I input into the "Server" field, it does show this message. If I connect to a server by double clicking, Jamulus tries to connect but then crashes after a few seconds.

ann0see:
> since I haven't been able to get it to disable

On VirtualBox the virtual sound card only supports 44.1 kHz. Therefore it’s a good way to test.

npostavs:
Um, I can't figure it out. ASIO4ALL shows no devices, but doesn't report any errors!? Are there some audio particular settings you need for VirtualBox?

image

ann0see:
I think you‘ll need to select the Null Audio Driver in VirtualBox

npostavs:
> I think you‘ll need to select the Null Audio Driver in VirtualBox

Ah, that did it yes. But I'm still not able to reproduce the problem. Double clicking the server doesn't let me connect.

image

ann0see:
Hmm. I could reproduce it once (changed sound device settings to an invalid configuration after I started Jamulus). However if you change ASIO4ALL settings while you are connected to a server and they are invalid, Jamulus crashes.

npostavs:
> However if you change ASIO4ALL settings while you are connected to a server and they are invalid, 

The problem is I don't know how to do this. Actually, I can't even seem to bring up the ASIO4ALL settings when VirtualBox is using the Null Audio Driver. (also, Windows Update is pegging the virtual CPU at 100% which may be interfering, not sure)

ann0see:
Can you route your interface via USB to Virtual Box?

npostavs:
Oh, I managed to do it by connecting with VoiceMeeter enabled, and then closing VoiceMeeter while connected to a Jamulus server and re-selecting the VoiceMeeter device in Jamulus.

I think the main problem was that I dropped the `UnloadCurrentDriver` call in `CSoundBase::SetDev`. Should be fixed now.

ann0see:
Hmm. Just tested it again and it still crashes if I change to an unsupported device while being connected to a server ;-(. 

My configuration is as follows:

* Only available ASIO driver ASIO4ALL
* Bluetooth speaker which only supports 44.1kHz
* Yamaha keyboard which only supports 44.1kHz

Unfortunately, I couldn't connect my interface to VirtualBox... Do you have a device which doesn't support 48kHz? Headset, Bluetooth loudspeaker, ...?

npostavs:
I borrowed a webcam that only supports 32 kHz and managed to reproduce the problem! I'll also note that I had to enable the "Hardware Buffers" option in ASIO4ALL (some devices are in "WaveRT mode" and don't support this option at all), otherwise it doesn't report any errors back to Jamulus.

The issue was calling `Sound.Start()` after reloading the sound driver fails. It should be fixed now.

I did manage to get Jamulus in a funny state where the GUI looks as if it's still connected to the server, but in fact isn't though. I guess it should just be a matter of explicitly calling `Disconnect` in those places.

pljones:
I still think any changes to the sound settings shouldn't be possible whilst connected.  I don't think it makes sense, logically.

npostavs:
> I still think any changes to the sound settings shouldn't be possible whilst connected. I don't think it makes sense, logically.

There isn't any way to stop the user from interacting with the ASIO control panel while connected. Users can behave illogically.

pljones:
> 
> 
> > I still think any changes to the sound settings shouldn't be possible whilst connected. I don't think it makes sense, logically.
> 
> There isn't any way to stop the user from interacting with the ASIO control panel while connected. Users can behave illogically.

Yes there is.  You disconnect them.  It's very simple.  The button sends a signal.  Route that to somewhere that calls `Disconnect()`.

npostavs:
> Yes there is. You disconnect them. It's very simple. The buttons sends a signal. Route that to somewhere that calls `Disconnect()`.

That button isn't the only way to open the control panel. You can also access it from the system tray. Or you can leave the control panel open while you connect.

Furthermore, even if that was possible, I think it's a terrible idea. Forcing the user to go through a disconnect, reconfigure, and reconnect dance to listen for the effect of every little change? Sound configuration is already difficult, this would make it just about intolerable.

pljones:
Any method that sends a signal that causes the dialogue to open can be rerouted to disconnect first.

Like I say, I think it makes no sense to have the network connection active if the audio system is being adjusted.

npostavs:
> Any method that sends a signal that causes the dialogue to open can be rerouted to disconnect first.

Neither of those methods send a signal to Jamulus. The system tray is outside of Jamulus. And obviously **not closing** the panel doesn't send anything anywhere.

> Like I say, I think it makes no sense to have the network connection active if the audio system is being adjusted. (People ought to disconnect when plugging in and unplugging guitars and microphones, etc., too but there's nothing we can do about that.)

I still say it's a terrible idea to prevent to user from getting immediate feedback from adjustments. But if you're only talking about the network connection, then I guess if we implemented local-loopback (#882/#1252) so that the user can still listen while adjusting it would be okay.

pljones:
The aim is to prevent one user who is having problems affecting 149 other users that user is affecting.  (That's the current limit; it could go up to 254 pretty easily - but, well CPU and network load...)

If you mean the ASIO4ALL tooltray icon, agreed - nothing can be done about that.  I was confused with the server tooltray icon.  I don't use ASIO4ALL, so I don't see its icon, of course.

Basically, if audio stops working, we should prevent the user affecting anyone else whilst they sort it out.  The feedback prevention fix is another example of this.  Something similar could be done, I suppose - rather than disconnecting, switch on Mute Myself.

npostavs:
> The aim is to prevent one user who is having problems affecting 149 others on the same server.

I'm coming from the point of view of playing in much smaller groups, where we generally don't start until everyone is setup properly. So blocking communication when someone is having sound troubles just makes things more difficult (now I have to try calling them on the phone or something).

> If you mean the ASIO4ALL tooltray icon, agreed - nothing can be done about that. I was confused with the server tooltray icon.

Ah, yeah, I meant the ASIO4ALL or other ASIO icons (e.g., the Motu ASIO driver control panel is also accessible by tooltray icon).

>  rather than disconnecting, switch on Mute Myself.

Hmm, since Mute Myself also activates local loopback, I guess that might be okay, although potentially quite confusing.

pljones:
By the way, for comparison, on Linux, it's pretty insistent that the audio device remains working:
https://github.com/jamulussoftware/jamulus/blob/master/linux/sound.cpp#L221-L226
The code throws a CGenErr -- terminating the client completely.  This could be triggered, for example, by using QJackCtl to change to an incompatible or unavailable device, I guess (I've not actually tried this).

npostavs:
> I guess it should just be a matter of explicitly calling `Disconnect`

Actually, I just had to remove a conditional around an existing `Disconnect` call. I also fixed another segfault triggered by trying to open the control panel after auto selection failed (meaning no device is selected).

npostavs:
I've added a button to reset the device, so that the user can let Jamulus know when a device changes from non-working to working (technically you could already do this by reselecting the device from the combobox, but that is confusing).

I consider this PR complete at this point (although I still plan to test on Linux), so I'm removing the draft status.

ann0see:
Ok. Windows seems to work now (with a small bug):

If I only have one device (ASIO4ALL) which isn't configured correctly, the ASIO-Device-Settings dialog doesn't open. I first need to click on "Reset driver".
***

UI:


I think the "Reset driver" button should be next to the device selection and be called "Reload or Reset". Another option would be a reload button (like in a browser). Adding too many buttons gets confusing.

Also the auto select device might be placed somewhere less prominent. Not sure how we should handle that. But that's something for @mulyaj and @gilgongo (UX-Experts).

Another thing is the message if you try to connect to a server while having a non-working config: I think it should show the cause of the error (not only "you can't connect because your config is wrong"). What about a button with which you can open Jamulus's settings.

npostavs:
> If I only have one device (ASIO4ALL) which isn't configured correctly, the ASIO-Device-Settings dialog doesn't open. I first need to click on "Reset driver".

Ah, the ASIO backend was unloading the driver on error. Fixed now.

> I think the "Reset driver" button should be next to the device selection and be called "Reload or Reset".

Adding synonyms to the button label seems like it would just make things more confusing.

> Another option would be a reload button (like in a browser).

You mean an icon instead of text?

> Another thing is the message if you try to connect to a server while having a non-working config: I think it should show the cause of the error

Added that.

> What about a button with which you can open Jamulus's settings.

Hmm, is that more useful than the existing button in the main window with which you can open Jamulu's settings?

ann0see:
> You mean an icon instead of text?

Yes. Having too many buttons right above each other might look confusing.

> Hmm, is that more useful than the existing button in the main window with which you can open Jamulu's settings?

Yes. If there's an error I think a person would like to have a "one click" solution right in the error message (UX).

npostavs:
Okay, it looks like this:

image
image

ann0see:
Yes. That looks great. Will do some tests with the new code (especially on Windows). 

I think the UI experts should have a look at the configuration here (@mulyaj @gilgongo) since this is touching a critical part of the audio setup process (which confuses a lot of people).

ann0see:
Ok. Another bug: 

1. Connect to a server
2. Change ASIO4ALL settings to an incompatible state
3. Try to reconnect
4. Get an error message which just says "you can't connect because your current audio device configuration isn't working:" without a reason

npostavs:
Oh, apparently assigning a boolean value to a `QString` compiles!? Fixed.

> Also another request: After disabling the wrong device in ASIO4ALL, the driver doesn't seem to get reloaded automatically. I think it should do so.

Currently Jamulus doesn't get messages from the driver once it goes into an error state. Probably I could change that, but I want to keep the changes limited, as this PR is already getting a bit long.

> "Auto Select device". I think this button should either not show up if only one device is present

Yeah, makes sense. Done.

> Sorry for all these bug reports and feature requests ;-(. This area seems to be quite difficult and error prone.

Contrariwise, thank you for all these bug reports and requests! It would be hard to get everything working without your thorough testing.

mulyaj:
> image


Here you could clarify the action to the user better. Maybe add a line break adding,

 "A compatible device is available:
 _Name of Device_
 Would you like to select it?"

Close - Select Device - Open Settings

src/clientdlg.cpp
```diff
@@ -1159,7 +1159,25 @@ void CClientDlg::Connect ( const QString& strSelectedAddress, const QString& str
         catch ( const CGenErr& generr )
```
pljones, line 1159: `linux/sound.cpp`, `android/sound.cpp`, `ios/sound.mm`, `src/soundbase.cpp`, `mac/sound.cpp`, `windows/sound.cpp` - these all throw CGenErr from multiple places.

`src/socket.cpp` throws the error, too - and it's caught here.

src/clientsettingsdlg.cpp
```diff
@@ -244,6 +244,8 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     butDriverSetup->setAccessibleName ( tr ( "ASIO Device Settings push button" ) );
```
pljones, line 244: Are the changes to `src/clientsettingsdlg.cpp` and `src/clientsettingsdlgbase.ui` relevant to this change?

ann0see:
Yep. Your last commit fixed this error. 

> Currently Jamulus doesn't get messages from the driver once it goes into an error state. Probably I could change that, but I want to keep the changes limited, as this PR is already getting a bit long.


Hmm. Yes that's probably out of scope then (but also highly confusing probably worth a new issue/pr after this one gets merged). 

There's also another issue with ASIO4ALL which I can't reproduce yet. If I connect to a server, it doesn't show the mixing board.

npostavs:
> "A compatible device is available:
> *Name of Device*

The main problem is that we don't know ahead of time whether there is a compatible device available (or what its name is). There is a reasonable chance that the correct thing to do is actually stay with the current device, but adjust its configuration (not possible from within Jamulus).

src/clientdlg.cpp
npostavs: Hmm, you think we should add a subtype of error to distinguish the audio error case?

src/clientsettingsdlg.cpp
npostavs: By "this change" you mean the PR as a whole? Yes, since we aren't quitting Jamulus when the sound card fails, we need to provide the user with more actions to deal with that state.

mulyaj:
> The main problem is that we don't know ahead of time whether there is a compatible device available (or what its name is).

Ok I see — what would you say would be a good description of what auto-select does? 

I'm not too familiar with the technical side of Jamulus so that was my best guess, and I'm certain others might be unsure as well too.

npostavs:
> Ok I see — what would you say would be a good description of what auto-select does?

You can imagine what the user would have to do if there were no auto-select function: Choose each device from the dropdown menu in turn, and stop when you hit one that doesn't throw an error. The auto-select button just automates the tedious part.

----

Thinking about it some more, I suppose we *could* have auto-select run speculatively without the user asking for it, to be able to present a working device choice. I don't want to add that to this PR because it's already getting kind of long, like I mentioned above.

mulyaj:
Here are two dialogs I've thought up. One with _find_ vs. _automatically select_.

If @npostavs can confirm, auto-select might not necessarily find a compatible device, so using _find_ in the dialog can give a user an expectation of the possibility of failure. 

"Other devices are available. 
Would you like Jamulus to automatically select a compatible device? 
Alternatively, you can manually select your desired device in the settings menu."

Close - Auto-Select - Settings

"Other devices are available.
Would you like Jamulus to find a compatible device?
Alternatively, you can manually select your desired device in the settings menu."

Close - Find - Settings

To add, is there any pop-up dialog if nothing compatible is found through auto-select?

pljones:
"search for" rather than "automatically select" or "find" might be better still - that _definitely_ implies the chance of failure ("find" being the successful outcome of a "search" operation).

On successfully finding a device, should its configuration be shown so the user has a clue what's happening?  If not, then "search for and select" might be appropriate.

ann0see:
Tests I made and can think of currently. 

- [ ] Testing on macOS: remove device while jamming, select non supported device, select unsupported sample rate in AUDIO MIDI Setup, anything else?
—-

- [ ] Testing on Windows: Similar to macOS; however more tests with 
- [ ] 1 driver (ASIO4ALL only)
  - [ ] Check errors with one supported device while connected/not connected to server and one non 48 kHz device enabled. Does everything work as expected?
  - [ ] ???
 - [ ] Two drivers
   - [ ] Same as above with each driver
   - [ ] also select another driver while making incompatible changes to A4A and select ASIO4ALL during a session with a working different device
   - [ ] ???

ann0see:
Did a test on macOS Big Sur and it seems as if I can't even connect to a server (hangs on trying to connect and then gives back that the audio interface doesn't work). Do you have access to macOS?

npostavs:
> Do you have access to macOS?

No, I don't.

ann0see:
That's a pitty. I'm not sure if I can help finding the bug on macOS. Maybe someone else with a mac can help out?

ann0see:
Missing here:

- [ ] Debugging macOS: Why is the sound-system broken?
- [ ] More in depth tests with ASIO4ALL on Windows 10

ann0see:
@npostavs @ngocdh @emlynmac probably moving this forward is something for you. 

Benefits:

* Better UX if there are audio issues
* Possibly laying foundations for PortAudio

Issues:
* Modifies critical code
* Needs extensive testing (and maybe even a review from more than 2 developers with C++ knowledge).

npostavs:
I rebased on latest master, but I'm afraid I still have no idea what to do about the macOS failures.

pljones:
> I rebased on latest master, but I'm afraid I still have no idea what to do about the macOS failures.

Whatever you did seems to have fixed it.  Sometimes they just need rerunning.

npostavs:
I didn't mean automated CI failures, but rather the problems that ann0see found (e.g., https://github.com/jamulussoftware/jamulus/pull/1378#issuecomment-849847817).

gilgongo:
I've got access to a Mac at the moment but not a build environment on it. Is there a binary I can test?

pljones:
https://github.com/jamulussoftware/jamulus/actions/runs/1419274793
Those are the assets from the build, I think.  Somewhere around there, anyway.

gilgongo:
Thanks - I can confirm the issues that @ann0see observes though :-(  I'm also on Big Sur.

src/soundbase.cpp
```diff
@@ -76,159 +76,83 @@ QStringList CSoundBase::GetDevNames()
 
     QStringList slDevNames;
 
+    bool currentInList = false;
+
     // put all device names in the string list
     for ( int iDev = 0; iDev < lNumDevs; iDev++ )
     {
+        currentInList |= ( strDriverNames[iDev] == strCurDevName );
```
softins, line 84: Strictly speaking, this should be `||=`, not `|=`.

softins:
The conflicts stem from the recent PR #2215 which only shows the ASIO setup button if we are on Windows AND we are not using Jack.

On what platforms do the additional buttons need to be displayed and handled? Just Windows, or all? This probably needs clarifying before trying to debug any macOS issues.

ann0see:
I‘m quite sure the ASIO Device Setup button is only shown on Windows with ASIO.

softins:
> I‘m quite sure the ASIO Device Setup button is only shown on Windows with ASIO.

Well yes. What I don't know is whether the additional two buttons this PR introduces are only needed on Windows, or on all platforms. I suspect maybe at least on Mac too. I haven't yet studied the intended functionality.

npostavs:
Yes, the additional buttons apply across all platforms, conflicts should be resolved now.

ann0see:
I assume the macOS code works fundamentally different to windows and linux. @ngocdh do you know anything about how the macOS code handles audio device failures?

pgScorpio:
I just tested [https://github.com/jamulussoftware/jamulus/actions/runs/1700022274](https://github.com/jamulussoftware/jamulus/actions/runs/1700022274) and indeed it doesn't change device when the device is not working, but pressing the "ASIO Device settings" button does NOT open the ASIO settings either.... So still no way to correct the problem.
Also after "Auto Select device" and then changing back to the faulty device the settings window still shows the previous drivers channel mapping.

P.S. To quickly check a faulty device I use the KoordASIO driver with a mono input device selected.

P.S.2 Why Jamulus does not accept a mono input seems a bit strange to me too, since with more than 2 inputs (also strange why not with 2 inputs) I can select the same input for both channels.

npostavs:
> P.S. To quickly check a faulty device I use the KoordASIO driver with a mono input device selected.

I don't have such a device, so I can't reproduce this easily. Are you getting the `The selected audio device is incompatible since it doesn't support 1 in/out channels. Please select another device or configuration.` message?

pgScorpio:
> I don't have such a device

KoordASIO is a simple replacement for ASIO4ALL, you can download it via the link I provided. And almost any USB headset will have a mono input.



> Are you getting the `The selected audio device is incompatible since it doesn't support 2 in/out channels. Please select another device or configuration.` message?

No I get "Couldn't initialise the audio driver. Check if your audio hardware is plugged in and verify your driver settings"

I can generate the `The selected audio device is incompatible" message too, using another device in KoordASIO, but in that case I CAN open the ASIO Settings.

pgScorpio:
> EDIT: forcing the `lNumInChan` variable in...

That's no option for the average user though....

npostavs:
> No I get "Couldn't initialise the audio driver. Check if your audio hardware is plugged in and verify your driver settings"

Ah, I see where that's from. I think avoiding the `removeCurrentDriver` call should fix this, a 2nd case of https://github.com/jamulussoftware/jamulus/pull/1378#issuecomment-846593084. Can you try out the new executable at https://github.com/jamulussoftware/jamulus/actions/runs/1705718177

It's not clear to me what causes that error though. Seems to be unrelated to Jamulus requiring a stereo device (that check comes later). Is it possible that KoordASIO requires stereo?

> > EDIT: forcing the `lNumInChan` variable in...
> 
> That's no option for the average user though....

Yeah, I was just trying to see if I could reproduce your problems on my end. Obviously a user wouldn't be using a debugger.

pgScorpio:
> try out the new executable

i just tested jamulus_3.8.1dev-3838fdd_win.exe with mixed results.

> It's not clear to me what causes that error though

Neither it is to me now, since it doesn't seem to be mic related at all (the mic. seems to have 2ch after all, my mistake) and I don't get the "Couldn't initialise the audio driver..." anymore, (Did the check itself change too?) but with the same KoordASIO config as I got this message before I now can connect to a server, mic working, but I get no audio on the headphones (though I hear some faint white noise after connecting, so the device seems to be initialized.). 
Note: The device is configured OK in windows and working with ASIO4ALL (2.15Beta1), so this might be a KoordASIO bug as well.

Second thing I noticed is that any changed ASIO settings don't take effect until the device is re-selected (You get the same error again when trying to connect after changing the ASIO settings). So maybe the driver should be re-initialized before connecting?

And another question: Is it right that the Äuto Select Device" always selects the first working device in the list? Not the last used working device, and even when the current device is working ok? 
Also I found myself accidentally hitting this button several times, maybe it would be better to have an "auto-select" option in the device drop down list? I think this would be more logical, and it also doesn't take up extra space in the UI.

npostavs:
> Neither it is to me now, since it doesn't seem to be mic related at all (the mic. seems to have 2ch after all, my mistake) and I don't get the "Couldn't initialise the audio driver..." anymore, (Did the check itself change too?)

No change in Jamulus' checks. Is it possible there's another application competing for the mic/speakers and interfering? Are you in shared or exclusive mode?

> I now can connect to a server, mic working, but I get no audio on the headphones (though I hear some faint white noise after connecting, so the device seems to be initialized.).

By "mic working" you can see the levels go up when you make noise then? Even though you don't get any echo back?

> Second thing I noticed is that any changed ASIO settings don't take effect until the device is re-selected (You get the same error again when trying to connect after changing the ASIO settings). So maybe the driver should be re-initialized before connecting?

I think this is a KoordASIO problem, it doesn't send the `kAsioResetRequest` message after changing settings.

> And another question: Is it right that the "Auto Select Device" always selects the first working device in the list? Not the last used working device, and even when the current device is working ok?
> Also I found myself accidentally hitting this button several times, maybe it would be better to have an "auto-select" option in the device drop down list? I think this would be more logical, and it also doesn't take up extra space in the UI.

Oh, I didn't consider the accidental button press, that is a good point. I'll try and see if I can make a drop down option work instead.

pgScorpio:
> Are you in shared or exclusive mode?

Tested both. And I can force a "Your sound card is not working correctly..." in exclusive mode when making sure the device is already in use. But that message doesn't appear until a few seconds after "trying to connect".

Yesterday I always immediately got the "Couldn't initialize the audio driver..." with this configuration regardless exclusive mode and in use or not. So I really don't know what changed this behavior.

> By "mic working" you can see the levels go up when you make noise then?

Yes, and I even can hear the audio on another client.... But no audio from the server on this client with this particular config.

(P.S. Maybe the change in behavior is because in the mean time I rebooted my computer?, since I re-installed jamulus_3.8.1dev-f2b59e8_win.exe and I still can't reproduce the "Couldn't initialise the audio driver", while yesterday it was very consistent. But also no return audio with this configuration on 3.8.1dev-f2b59e8 too, while the same ASIO config works in reaper.)

> I think this is a KoordASIO problem, it doesn't send the `kAsioResetRequest` message after changing settings.

Might be. But isn't it always wise to reset before connecting, just to make sure the latest settings are loaded?

> Oh, I didn't consider the accidental button press, that is a good point. I'll try and see if I can make a drop down option work instead.

That would be great! Shouldn't be too hard to implement the drop down option.... ;=)

ann0see:
I‘m really sorry to say that, but I‘m afraid we need to move that to a feature branch once more?

pgScorpio:
> I‘m really sorry to say that, but I‘m afraid we need to move that to a feature branch once more? (Of course unless we find someone who helps out debugging the macOS related issues)

I'm already diving into the mac code, and I can see why there are problems. This is since the mac code has a totally different approach (channel selection via the device selector) and so quite a different implementation. (Exactly what I wanted to prevent with my ["audio redesign"](https://github.com/jamulussoftware/jamulus/pull/2407) PR.)


> I‘d suggest to open/link the appropriate issues though.

Please explain what what you mean here...

pgScorpio:
> There are issues which could be solved with this PR. They should be found and linked to this PR.

I you know more sound relater issues, please let me know. I will solve them. I probably solved a lot (even still unknown bugs) already.

> Great to hear you found issues on the macOS part. 

And not only in the macOS part. ;=)

I'm quite far with the CSoundBase redesign and the Windows implementation....
I will post an update in the ["audio redesign"](https://github.com/jamulussoftware/jamulus/pull/2407) PR soon.

ann0see:
Great thanks. Yes, sometimes changing in/output devices in ASIO4ALL freezes Jamulus, if ASIO4ALL is not the only driver you can't change it (that's what you mentioned), failed audio drivers don't reload,...

I hope your changes get accepted, but maybe you need to give micro pr's instead (one for each OS). Especially the ASIO part is tricky - as you know.

pgScorpio:
> Great thanks. Yes, sometimes changing in/output devices in ASIO4ALL freezes Jamulus,

Freezing when changing settings while connected seems to be solved! (At least on windows.)

Though I encountered another "freeze" problem that seems to occur only when closing Jamulus while using Jack for Windows.
No idea what is happening here, but it then sometimes freezes in Windows itself while trying to play the sound of a messagebox but I can't find out why, since when this happens the debugger doesn't show any Jamulus code in the call stack, anymore, just Windows code itself).

> if ASIO4ALL is not the only driver you can't change it (that's what you mentioned), failed audio drivers don't reload,...

?? What I mentioned was that you can't correct a wrong driver configuration when the driver has no own "offline settings" (not just ASIO4ALL)
But with the redesign you will always be able to to open ASIO device settings when a device has incorrect configuration....

> I hope your changes get accepted,

So do I ! There are a lot of improvements.... (I will post the most important ones in a separate post.)

> but maybe you need to give micro pr's instead (one for each OS). Especially the ASIO part is tricky - as you know.

But since it is a (major) change in CSoundBase this is not possible... We will have to merge them all at once.

So I could use some advice here...
Should I close this PR for now, while working on it on my own branch ?

In that case I can check in my modifications one by one, without affecting master. But anyone could check my personal branch to test some things and see the progress.

Once all OS's are working with the new structure we can decide when and how to merge (new PR or re-opening PR?).

ann0see:
@ngocdh @pgScorpio I will close this due to the sound redesign. But please do not remove the branch. I think it would be great if you both exchanged ideas!

PR #1401, opened 2021-03-30
Windows: Delay detaching from console

### LINKED ISSUE #1281: Jamulus -h does not terminate correctly in Windows cmd console
**Describe the bug**
When 'jamulus -h' is issued in a cmd console window, the help text is displayed **below** the next command prompt.
I.e. first the next prompt appears, and only then the current help text is displayed.
The 2nd line in **Screenshots** below shows the problem: "C:\Program Files (x86)\Jamulus>" must not be appended by "Usage: jamulus [option] [optional argument]". "C:\Program Files (x86)\Jamulus>" shall be displayed **BELOW** the help text. Currently you need to enter Ctrl+C or return to clean up and get the next (empty) command prompt.
Same problem for jamulus -v

**To Reproduce**
- open a cmd console (e.g. press WINDOWS+R, enter cmd, press return)
- enter cd "C:\Program Files (x86)\Jamulus" including the double quotes which will display the following prompt
    C:\Program Files (x86)\Jamulus>
- enter jamulus -h <return>

**Expected behavior**
The help text shall be displayed, followed by an empty new line prompt, e.g.
C:\Program Files (x86)\Jamulus>jamulus -h
Usage: jamulus [option] [optional argument]
:
Example: jamulus -s --inifile myinifile.ini
C:\Program Files (x86)\Jamulus>

**Screenshots**
C:\Program Files (x86)\Jamulus>jamulus -h
C:\Program Files (x86)\Jamulus>Usage: jamulus [option] [optional argument]
Recognized options:
  -h, -?, --help        display this help text and exit
  :
Example: jamulus -s --inifile myinifile.ini

**Operating system**
Windows 8.1 and above

**Version of Jamulus**
3.6.2

**Additional context**
I think the output of tsConsole needs to be flushed prior to exiting the main program.

hoffie:
This moves the call to attach to the parent process' console from the start of the main routine to a place after command line parsing.

The goal is to improve command-line-only invocations of Jamulus for non-headless builds, e.g. `Jamulus -h` or `Jamulus -n -s`.

It also modifies the logic to disable the detaching not only for headless builds but also for non-headless builds with --nogui invocations.

Hopefully fixes #1281.

Disclaimer: I cannot test this at all. I don't even know if it compiles (CI will show).

@drummer1154  Can you please do a test with the artifacts once the build has finished?

- [ ] Is #1281 fixed?
- [ ] Do startup messages still appear as they used to? https://github.com/jamulussoftware/jamulus/issues/1281#issuecomment-803160782

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -89,13 +89,6 @@ int main ( int argc, char** argv )
     QString      strWelcomeMessage           = "";
     QString      strClientName               = "";
 
-#if !defined(HEADLESS) && defined(_WIN32)
-    if (AttachConsole(ATTACH_PARENT_PROCESS)) {
-        freopen("CONOUT$", "w", stdout);
-        freopen("CONOUT$", "w", stderr);
-    }
-#endif
-
     // QT docu: argv()[0] is the program name, argv()[1] is the first
     // argument and argv()[argc()-1] is the last argument.
     // Start with first argument, therefore "i = 1"
@@ -625,6 +618,12 @@ int main ( int argc, char** argv )
         iPortNumber += 10; // increment by 10
     }
 
+#if defined(_WIN32)
+    if (bUseGUI && AttachConsole(ATTACH_PARENT_PROCESS)) {
+        freopen("CONOUT$", "w", stdout);
+        freopen("CONOUT$", "w", stderr);
+    }
+#endif
 
     // Application/GUI setup ---------------------------------------------------
     // Application object


hoffie:
Direct link to the artifact: https://github.com/jamulussoftware/jamulus/suites/2382936172/artifacts/50615670

npostavs:
Seems to be worse: I get no output at all when passing `-h` and `--version`.

hoffie:
> Seems to be worse: I get no output at all when passing `-h` and `--version`.

Thanks for testing. So @pljones was right in the first place. :)
We'll have to find some other way then.

Issue #1394, opened 2021-03-30
Start Jamulus two times on two sound cards


newlaurent62:
Hello,

I'm on Linux (UbuntuStudio)

I created two presets for two sound cards in qjackctl : U22 and XR18.

With the commands 

`qjackctl -n U22 & qjackctl -n XR18 &`

I can start Jack two times with different parameters by deactivating the Service D-Bus.

But Jamulus stuck at finding the jack server. (There is no DEFAULT Jack server)

Could it be possible to add a command line parameter "Jack Server Name", to select the Jack server to use ?

BR,
Laurent

PR #1406, opened 2021-03-31
allow using specific jack server instance

djfun:
Add a new command line parameter "--jackservername <servername>", which will
be used to connect to a jack server instance with the specified name.
If the parameter is missing, connect to the default jack server.

See #1394 and #1397 

So far, I have only tested building for linux.

### DIFF
--- a/android/sound.cpp
+++ b/android/sound.cpp
@@ -33,6 +33,7 @@ CSound::CSound ( void           (*fpNewProcessCallback) ( CVector<short>& psData
                  void*          arg,
                  const QString& strMIDISetup,
                  const bool     ,
+                 const QString&,
                  const QString& ) :
     CSoundBase ( "Oboe", fpNewProcessCallback, arg, strMIDISetup )
 {
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -28,12 +28,16 @@
 
 #ifdef WITH_SOUND
 void CSound::OpenJack ( const bool  bNoAutoJackConnect,
-                        const char* jackClientName )
+                        const char* jackClientName,
+                        const char* jackServerName )
 {
     jack_status_t JackStatus;
 
     // try to become a client of the JACK server
-    pJackClient = jack_client_open ( jackClientName, JackNullOption, &JackStatus );
+    if (strlen(jackServerName) > 0)
+        pJackClient = jack_client_open ( jackClientName, JackServerName, &JackStatus, jackServerName );
+    else
+        pJackClient = jack_client_open ( jackClientName, JackNullOption, &JackStatus );
 
     if ( pJackClient == nullptr )
     {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -88,6 +88,7 @@ int main ( int argc, char** argv )
     QString      strServerListFilter         = "";
     QString      strWelcomeMessage           = "";
     QString      strClientName               = "";
+    QString      strJackServerName           = "";
 
 #if !defined(HEADLESS) && defined(_WIN32)
     if (AttachConsole(ATTACH_PARENT_PROCESS)) {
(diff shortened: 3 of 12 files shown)


android/sound.cpp
```diff
@@ -33,6 +33,7 @@ CSound::CSound ( void           (*fpNewProcessCallback) ( CVector<short>& psData
                  void*          arg,
                  const QString& strMIDISetup,
                  const bool     ,
+                 const QString&,
```
hoffie, line 36: It's suboptimal that we have to add implementation-specific flags such as this jack option to all implementations (Windows, Android, Mac), but I can't think of any sane alternatives. Also, it follows the existing logic for jack-specific options (`bNoAutoJackConnect`), so I guess it's ok.

src/main.cpp
```diff
@@ -532,6 +533,22 @@ int main ( int argc, char** argv )
         }
 
 
+        // Jack server name ----------------------------------------------------
```
hoffie, line 536: Could we only add this logic if we have jack support at all (`#ifdef WITH_SOUND`, it seems. #1408)?

(I know that the existing code for jack options does not lead by good example in this regard, we should try to be better nevertheless :))

src/main.cpp
```diff
@@ -875,6 +893,7 @@ QString UsageArguments ( char **argv )
         "      --mutemyown       mute me in my personal mix (headless only)\n"
         "  -c, --connect         connect to given server address on startup\n"
         "  -j, --nojackconnect   disable auto Jack connections\n"
+        "      --jackservername  use specific Jack server instance\n"
```
hoffie, line 896: Should also be guarded by `#ifdef WITH_SOUND`, IMO.

hoffie (commented):
Thanks for the PR! Looks good in general. I have added some small comments.

src/main.cpp
djfun: I updated my commit

hoffie (approved):
Thanks!
Could you also add yourself to the contributors' list in https://github.com/jamulussoftware/jamulus/blob/ad12ba0543e21792b6760d6b0e1df860b3e61741/src/util.cpp#L489?

hoffie:
Thinking some more about it, would this do the same as the `JACK_DEFAULT_SERVER` environment variable? If so, does the latter work for Jamulus? If so, I'm not sure if we should have an additional command line flag.

djfun:
> Thinking some more about it, would this do the same as the `JACK_DEFAULT_SERVER` environment variable? If so, does the latter work for Jamulus? If so, I'm not sure if we should have an additional command line flag.

:-/ yes it works, at least in my test setup. It should probably just be mentioned somewhere for people to find it.

pljones:
Ideally... there'd be a UI to accompany this.  Somewhere in the Device selector, in Settings:
image
(Or list the available Jack instances in the "Device" box?  That sort of fits the meaning, I guess.)

softins (dismissed):
I've built this on a RPi, and it appears to start a jack server with the appropriate name as given, or to connect to the running `jackdbus` if the name is not specified or is given as `default`.

softins:
> I've built this on a RPi, and it appears to start a jack server with the appropriate name as given, or to connect to the running `jackdbus` if the name is not specified or is given as `default`.

Having reviewed the discussion more closely, I have just confirmed that both these commands give exactly the same effect:
```
./Jamulus --jackservername fred
```
```
JACK_DEFAULT_SERVER=fred ./Jamulus
```
So probably better to document this somewhere rather than add a command option.

linux/sound.cpp
```diff
@@ -28,12 +28,16 @@
...
-                        const char* jackClientName )
+                        const char* jackClientName,
+                        const char* jackServerName )
 {
     jack_status_t JackStatus;
 
     // try to become a client of the JACK server
-    pJackClient = jack_client_open ( jackClientName, JackNullOption, &JackStatus );
+    if (strlen(jackServerName) > 0)
+        pJackClient = jack_client_open ( jackClientName, JackServerName, &JackStatus, jackServerName );
```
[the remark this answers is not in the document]
pljones: And it needs `{` and `} adding to make the statement a block in the standard code style.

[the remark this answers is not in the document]
djfun: yes, sorry, I am too used to the linux kernel coding style.

pljones:
> Yes, I was about to write the same. I don't think we should add features (especially those which need UI logic und space) without documented need by users? :)

If a feature is needed (i.e. we need to choose a Jack server instance), then it should have a UI by default, not require a command line parameter.

softins:
Closing this as it's not needed after all, since the environment variable provides equivalent functionality for those who need it.

PR #1407, opened 2021-03-31
Change references to "central server"

pljones:
See https://github.com/jamulussoftware/jamuluswebsite/pull/403 for more.

### DIFF
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -300,13 +300,13 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
     cbxInputBoost->setAccessibleName ( tr ( "Input Boost combo box" ) );
 
     // custom directory server address
-    QString strCentrServAddr = "<b>" + tr ( "Custom Central Server Address" ) + ":</b> " +
-        tr ( "Leave this blank unless you need to enter the address of a central "
+    QString strCentrServAddr = "<b>" + tr ( "Custom Directory Server Address" ) + ":</b> " +
+        tr ( "Leave this blank unless you need to enter the address of a directory "
         "server other than the default." );
 
     lblCentralServerAddress->setWhatsThis ( strCentrServAddr );
     cbxCentralServerAddress->setWhatsThis ( strCentrServAddr );
-    cbxCentralServerAddress->setAccessibleName ( tr ( "Central server address combo box" ) );
+    cbxCentralServerAddress->setAccessibleName ( tr ( "Directory server address combo box" ) );
 
     // current connection status parameter
     QString strConnStats = "<b>" + tr (  "Current Connection Status "
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -299,7 +299,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
     cbxInputBoost->setWhatsThis ( strInputBoost );
     cbxInputBoost->setAccessibleName ( tr ( "Input Boost combo box" ) );
 
-    // custom central server address
+    // custom directory server address
     QString strCentrServAddr = "<b>" + tr ( "Custom Central Server Address" ) + ":</b> " +
         tr ( "Leave this blank unless you need to enter the address of a central "
         "server other than the default." );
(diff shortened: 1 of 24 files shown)


src/main.cpp
```diff
@@ -399,6 +399,21 @@ int main ( int argc, char** argv )
...
+                                 strArgument ) )
+        {
+            strCentralServer = strArgument;
+            qInfo() << qUtf8Printable( QString( "- directory server: %1" )
+                .arg( strCentralServer ) );
+            CommandLineOptions << "--directoryserver";
+            continue;
+        }
+
+        // Central server ** D E P R E C A T E D ** ----------------------------
```
hoffie, line 412: Should we have a `// TODO: Remove with Jamulus 4` so that we can have a quick way to grep for things to remove the next time we are allowed to do compatibility breakage?

hoffie (commented):
Does the below command line option check also need updating?

https://github.com/jamulussoftware/jamulus/blob/ad12ba0543e21792b6760d6b0e1df860b3e61741/src/settings.cpp#L835

I guess we don't want to go down the road of changing the name in other places, especially in the config...?

softins (commented):
My first reaction is: is there a real problem that this change is attempting to solve? The terminology of Central Server is well embedded in the Jamulus world, and all over the source code.

If the consensus is that this change is indeed needed (personally, I'm not convinced), then it needs to affect all user-visible instances of the terminology. Locations I found are (line numbers from current `master`):
* `clientdlg.cpp` line 768
* `clientsettingsdlg.cpp` lines 257, 258, 263
* `main.cpp` lines 610, 871. Also add deprecation message when logging the option.
* `serverdlg.cpp` lines 64, 73, 77, 78, 83, 86
* `settings.cpp` line 835
* `util.h` line 673

pljones:
> Does the below command line option check also need updating?
> https://github.com/jamulussoftware/jamulus/blob/ad12ba0543e21792b6760d6b0e1df860b3e61741/src/settings.cpp#L835

Yes, needs both...  

> 
> I guess we don't want to go down the road of changing the name in other places, especially in the config...?

Mmmm... Not the config.  But I'll get it into the translatable bits...  I think it's worth doing.

src/main.cpp
pljones: Mm, if we're going to do that, I'd suggest a separate issue raised to identify and tag the parts to go.  There's a lot already there.  But yes, a standardised tag would be good.

pljones:
> My first reaction is: is there a real problem that this change is attempting to solve? The terminology of Central Server is well embedded in the Jamulus world, and all over the source code.

I guess that depends on jamulussoftware/jamuluswebsite#403 - this is just making sure things are aligned.

src/clientsettingsdlg.cpp
```diff
@@ -253,14 +253,14 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
     edtNewClientLevel->setWhatsThis ( strNewClientLevel );
     edtNewClientLevel->setAccessibleName ( tr ( "New client level edit box" ) );
 
-    // custom central server address
-    QString strCentrServAddr = "<b>" + tr ( "Custom Central Server Address" ) + ":</b> " +
-        tr ( "Leave this blank unless you need to enter the address of a central "
+    // custom directory server address
+    QString strCentrServAddr = "<b>" + tr ( "Custom Directory Server Address" ) + ":</b> " +
```
ann0see, line 257: Shouldn't we also change variable names?

pljones: I wasn't going to do variable names in this change - just the user visible strings, to match the documentation change.

ann0see: Ok. Makes sense.

ann0see:
Can we merge this once the conflicts have been resolved?

pljones:
I think I resolved the conflicts okay...

ann0see:
@pljones I think this should also go into the next release.

ann0see (approved):

ann0see:
Resolved the conflicts. Probably a rebase would be cleaner.

hoffie (approved):
Looks good to me. I'll add a note to announce that to translators and in the release announcements.

Issue #80, opened 2020-04-16
Store Jack connections


corrados:
_This issue has been migrated from Sourceforge.  Created: 2015-03-13
Creator: Volker Fischer_

Right now we always make auto connections to the default inputs and outputs of the system. It would be better if the user changes the connections to store the new connections and if they are available at Jamulus startup then reconnect them. If not, then use the system default as is done right now.

corrados:
This Issue is a very old one. It was created by myself a long time ago.

My opinion regarding Jack auto connecting has changed since then. I prefer to do the connections outside of Jamulus. We support the essential mode that the system input/outputs are connected automatically which is very convenient for most use cases. And we also support a command line argument to set the Jack connections manually.

I prefer to do the connections in a script which calls Jamulus myself. Storing connections has the big disadvantage that if the other application is not running before Jamulus is started, the auto connection is not possible. If you are using a script, you can much better control what is actually connected and not.

Therefore I will close this Issue now.

WolfganP:
As I'm having issues launching jamulus as client in a headless raspberry (no X11, other GUI system nor dbus installed, loopback sound interface to allow lab testing of multi-threading), this issue caught my attention.
No matter how I configure jackd and launch, jamulus doesn't detect it and triggers the launch of a new jack server (obviously oblivious to the params I passed on my own jack server running already)

Looking at the examples at https://github.com/corrados/jamulus/wiki/Tips,-Tricks-&-More#jamulus-client-linux-start-script and https://github.com/corrados/jamulus/blob/master/distributions/raspijamulus.sh none of them seems to use any cmd line switch to avoid the auto jack config (I expected to see -j or --nojackconnect on the scripts but none is used).

Any tip?

corrados:
That is interesting. I thought only one Jack server can exist on a system. Why do you want to start your own Jack server before starting Jamulus? Why not use that auto-started new server?

WolfganP:
For my setup, it's a matter of selecting the right card interface with preconfig parameters in my headless system:

```
$ sudo JACK_NO_AUDIO_RESERVATION=1 jackd --no-realtime --timeout 2000 -d alsa --device hw:1,1 --softmode --rate 48000 --period 256 --nperiods 2 &
[1] 1842
jackdmp 1.9.12
Copyright 2001-2005 Paul Davis and others.
Copyright 2004-2016 Grame.
Copyright 2016-2017 Filipe Coelho.
[...]

PR #1417, opened 2021-04-01
Delay pan update

softins:
This only performs the floating-point delay calculation if delay pan is enabled.

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1212,6 +1212,7 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
 
         const int maxPanDelay = MAX_DELAY_PANNING_SAMPLES;
 
+        int iPanDelL, iPanDelR, iPanDel;
         int iLpan, iRpan, iPan;
 
         for ( j = 0; j < iNumClients; j++ )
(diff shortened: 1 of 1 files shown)


src/server.cpp
```diff
@@ -1212,6 +1212,7 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt,
 
         const int maxPanDelay = MAX_DELAY_PANNING_SAMPLES;
```
hoffie, line 1213: Do we know why this is necessary? Maybe for readability?
Might be another opportunity to check if the compiler optimizes this out or if it would be useful to just use the define in the code below.

hoffie (approved):
Loogs good to me. :)

pljones (approved):

pljones:
One client, no pan, delaypan server:
image
18 clients, no pan:
image
I'm seeing the CPU usuage bouncing around a lot, mostly 2.8 - just caught a low at 2.7 here, but it goes up to 3.5 at times and down 2.1 (might have seen one lower).

18 clients, hard pan:
image
I saw the mix threads drop down quite low here.

pljones:
The server is an Intel(R) Core(TM) i5-8600 CPU @ 3.10GHz box (all mine... doing quite a lot).  Kernel is 5.4.0-72-lowlatency #80-Ubuntu SMP PREEMPT.  Jamulus is Version 3.7.0dev-3c36aa98.

dcorson-ticino-com:
That is great !
On my ARM server I see about 9% increase in processor load.  
I guess the Intel has really good floating point support in comparison.

PR #1427, opened 2021-04-02
Server: Update statusfile on exit

### LINKED ISSUE #1423: HTML status file contains wrong information after terminating the Jamulus server
**Describe the bug**
The HTML status file created by the command line option "-m myStatusPathname.dat" contains wrong information after the server was terminated by receiving SIGTERM. It wrongly contains the last status **prior to server termination**, either that no client is connected or the list of previously connected clients.

The file contents suggests that the Jamulus server is still running where in fact it is not.

**To Reproduce**
Please note that multiple status file lines are reproduced below w/o markup in one line.
- Start Jamulus as server with command line options
o -s
o -d
o -m myStatusPathname.dat
- Read myStatusPathname.dat => "No client connected" => OK
- Have client "MyClient" connect to the Jamulus server
- Read myStatusPathname.dat => "MyClient" => OK
- Send SIGTERM to the Jamulus server which then terminates
- Read myStatusPathname.dat => "MyClient" => NOK

**Expected behavior**
Prior to terminating itself, the server shall write "Jamulus server terminated." into myStatusPathname.dat.

**Screenshots**
n/a

**Operating system**
GCP n1-standard-1, debian-10-buster-v20210217, Jamulus built from source
git clone https://github.com/corrados/jamulus.git
cd jamulus
git fetch
git checkout latest
=> HEAD is now at 337a484 r3_6_2
qmake "CONFIG+=nosound" Jamulus.pro && make clean && make

**Version of Jamulus**
3.6.2

**Additional context**
Same behavior in Windows when quitting the server via Ctrl+Q or clicking Window - Exit.

hoffie:
If the server terminates, empty the status file to avoid signalling that there are still clients connected.

@drummer1154 Can you test this?

Fixes #1423

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -722,6 +722,11 @@ void CServer::OnAboutToQuit()
     {
         UnregisterSlaveServer();
     }
+
+    if ( bWriteStatusHTMLFile )
+    {
+        WriteHTMLServerQuit();
+    }
 }
 
 void CServer::OnHandledSignal ( int sigNum )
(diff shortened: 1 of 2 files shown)


pljones (approved):
...although I'd have put the new method after the existing one... :)

Also note: there was a plan to remove the status file, once the server had a "better" means of having its status queried.  (As I argued against removing it, I'm not arguing against adding this...)

hoffie:
> ...although I'd have put the new method after the existing one... :)

Ok, changed.

> Also note: there was a plan to remove the status file, once the server had a "better" means of having its status queried. (As I argued against removing it, I'm not arguing against adding this...)

Yes, I think it would be better to have such things in a structured API or file, not in an HTML file...
However, this should not prevent us from improving the existing mechanism to be more correct. :)

softins (approved):

PR #1437, opened 2021-04-02
Replace non-ASCII characters in recording filenames.

### LINKED ISSUE #1424: Replace non-ASCII characters in recording file names
As discussed in https://github.com/jamulussoftware/jamulus/discussions/1384 

Jamlus supports unicode characters in profile names (as it should), and the OS will also support them in file names too. However, Audacity (and possibly other apps) don't support unicode characters in file names. So if a profile name contains such a character, the server operator cannot import the recording file(s) due to the unicode character being present.

A suggested fix for this is to replace any non-ASCII characters with for example a hyphen when Jamulus writes the recording file using the client's profile name (so the LOF file will use the corrected filenames).

softins:
Replaces high-order Latin characters with their nearest ASCII equivalent for readability.
Other non-ASCII chars are replaced with '_'.

Fixes #1424

### DIFF
--- a/src/recorder/jamrecorder.cpp
+++ b/src/recorder/jamrecorder.cpp
@@ -122,7 +122,8 @@ QString CJamClient::TranslateChars (const QString& input) const
 
     for (auto &c : r)
     {
-        c = (c >= 128) ? charmap[c - 128] : (c >= 32) ? c : '_';
+        unsigned char uc = c;
+        c = (uc >= 128) ? charmap[uc - 128] : (uc >= 32) ? c : '_';
     }
 
     return QString::fromLatin1(r);
--- a/src/recorder/jamrecorder.cpp
+++ b/src/recorder/jamrecorder.cpp
@@ -102,6 +102,32 @@ void CJamClient::Disconnect()
     wavFile = nullptr;
 }
 
+/**
+ * @brief CJamClient::TranslateChars Replace non-ASCII chars with nearest equivalent, if any
+ */
+QString CJamClient::TranslateChars (const QString& input) const
+{
+    static const char charmap[128] = {
+        '_', '_', ',', 'f', '"', '_', '_', '_', '^', '%', 'S', '<', 'E', '_', 'Z', '_',
+        '_', '\'', '\'', '"', '"', '_', '-', '-', '~', '_', 's', '>', 'e', '_', 'z', 'Y',
+        ' ', '_', '_', '_', '_', 'Y', '|', '_', '_', '_', 'a', '"', '-', '_', '_', '_',
+        '_', '_', '_', '_', '_', 'u', '_', '.', ',', '_', 'o', '"', '_', '_', '_', '_',
+        'A', 'A', 'A', 'A', 'A', 'A', 'E', 'C', 'E', 'E', 'E', 'E', 'I', 'I', 'I', 'I',
+        'D', 'N', 'O', 'O', 'O', 'O', 'O', 'x', 'O', 'U', 'U', 'U', 'U', 'Y', 'P', 'S',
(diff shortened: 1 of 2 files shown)


src/recorder/jamrecorder.cpp
```diff
@@ -102,6 +102,32 @@ void CJamClient::Disconnect()
...
+        'D', 'N', 'O', 'O', 'O', 'O', 'O', 'x', 'O', 'U', 'U', 'U', 'U', 'Y', 'P', 'S',
+        'a', 'a', 'a', 'a', 'a', 'a', 'e', 'c', 'e', 'e', 'e', 'e', 'i', 'i', 'i', 'i',
+        'd', 'n', 'o', 'o', 'o', 'o', 'o', '_', 'o', 'u', 'u', 'u', 'u', 'y', 'p', 'y'
+    };
+
+    QByteArray r = input.toLatin1();
+
+    for (auto &c : r)
+    {
+        c = (c >= 128) ? charmap[c - 128] : (c >= 32) ? c : '_';
```
hoffie, line 125: Getting this with gcc 10.2.0:

```
src/recorder/jamrecorder.cpp: In member function 'QString recorder::CJamClient::TranslateChars(const QString&) const':
src/recorder/jamrecorder.cpp:125:16: warning: comparison is always false due to limited range of data type [-Wtype-limits]
  125 |         c = (c >= 128) ? charmap[c - 128] : (c >= 32) ? c : '_';
      |              ~~^~~~~~
```

In fact, German 'Ä' is replaced by '_' but I guess it was supposed to be replaced by 'A' in your logic?

softins: Ok, this must be a portability thing. On your system `auto &c` must be getting typed as `signed char`, whereas on mine it gets typed as `unsigned char`. That would be why on your system it never uses the array.

I could try it with `unsigned char &c` instead, but that might end up as a type mismatch. Or maybe just cast `c` to unsigned within the loop body.

No time on Sat to look at it, will do Sun or Mon. I see the Mac build failed too, so will have to see why.

softins: I’ve pushed a change that hopefully might be portable. Let’s see.

pljones:
If this is going to be done, it might as well remove all the nasties, rather than adding some -- specifically the ones that then get replaced: `-`, `.`, `:`, `/`, `\`, ` ` (space).  However, it's best to avoid `,`, `~`, `%`, `|`, `^`, `<`, `>`, `'`, and `"`, really, as they're difficult on certain platforms (I should probably have stripped them, too).

src/recorder/jamrecorder.cpp
hoffie: No more warnings for me and `Ä` is proplery replaced with `A`. Looks good.

hoffie (approved):
Should probably be squash-merged?

softins:
> If this is going to be done, it might as well remove all the nasties, rather than adding some -- specifically the ones that then get replaced: `-`, `.`, `:`, `/`, `\`, ` ` (space). However, it's best to avoid `,`, `~`, `%`, `|`, `^`, `<`, `>`, `'`, and `"`, really, as they're difficult on certain platforms (I should probably have stripped them, too).

Well the function of `TranslateChars()` is to replace high-order Latin1 characters with their nearest equivalents, if any, not to sanitise for platform filename conventions. That is the job of the Regex replace further down the expression that constructs the filename.

softins:
Unless you would prefer that `TranslateChars()` had a complete 256-byte table and also took over the job of sanitising, instead of the `.replace(QRegExp`? @pljones

pljones:
> Unless you would prefer that `TranslateChars()` had a complete 256-byte table and also took over the job of sanitising, instead of the `.replace(QRegExp`? @pljones

If it's only going to be used in the one place, it might as well do it -- it'll be cheaper than the QRegExp, too, I expect.  Yeah...

pljones (approved):

PR #1438, opened 2021-04-03
add hint for JACK_DEFAULT_SERVER

djfun:
See discussion in #1397

See jack2 source code for comparison:
https://github.com/jackaudio/jack2/blob/a9c362060ba4182e3ea61a8d142d45da9ef63762/common/varargs.h#L44

### DIFF
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -31,6 +31,13 @@ void CSound::OpenJack ( const bool  bNoAutoJackConnect,
                         const char* jackClientName )
 {
     jack_status_t JackStatus;
+    const char *serverName;
+
+    if ( ( serverName = getenv( "JACK_DEFAULT_SERVER" ) ) == NULL ) {
+        serverName = "default";
+    }
+    qInfo() << qUtf8Printable( QString( "Connecting to jack \"%1\" instance (use the JACK_DEFAULT_SERVER environment variable to change this)." )
+        .arg( serverName ) );
 
     // try to become a client of the JACK server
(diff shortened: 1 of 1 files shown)


linux/sound.cpp
```diff
@@ -31,6 +31,13 @@ void CSound::OpenJack ( const bool  bNoAutoJackConnect,
                         const char* jackClientName )
 {
     jack_status_t JackStatus;
+    const char *serverName;
+
+    if ((serverName = getenv("JACK_DEFAULT_SERVER")) == NULL) {
```
hoffie, line 36: With the rather spacing-centric coding style of Jamulus, it should probably look like this :)

```suggestion
    if ( ( serverName = getenv( "JACK_DEFAULT_SERVER" ) ) == NULL ) {
```

hoffie (commented):
Seems good, logic-wise. Works properly.

It currently looks like this:
```
- connect on startup to address: localhost
- initialization file name: /tmp/x
Connecting to jack "default" instance (use the JACK_DEFAULT_SERVER environment variable to change this).
```

Should it also have a leading "- " similar to the other, status messages?

pljones:
> Should it also have a leading "- " similar to the other, status messages?

The "-" is leading on command line option outputs.  This is emitted from the linux sound code on start up.  So, not necessarily.  (I base this on how the jam recorder works.)

linux/sound.cpp
djfun: I added the changes and force-pushed a new version.

hoffie (approved):
Thanks!

pljones (approved):

Issue #1421, opened 2021-04-01
Add option to delete stored servers from combo lists


pgScorpio:
**Has this feature been discussed and generally agreed?**

Issue has been mentioned multiple times in discussion groups and on Jamulus facebook groups, but afaik not on github.
The given solution is always to edit the ini file (or even to delete it!), but I don't think this is very user friendly.

**Describe the solution you'd like**

 
Add an option to delete servers from the "Server" and "Custom Central Server" lists.
i.e. right-click server and select "Remove from list"

**Describe alternatives that have been considered**

Until now users have to edit the ini file to clean up their server lists.

DominikSchaller:
What is the use case? Why would someone need to delete a server? If it's not used, it would just move to the end of the list.

pgScorpio:
> What is the use case? Why would someone need to delete a server? If it's not used, it would just move to the end of the list.

For one "How do I delete a server" is a question I regularly get on several communities. Often because they accidentally entered a wrong server address and regularly, again accidentally, keep clicking this wrong server. And sometimes they just want to delete the server because the server no longer exists (also logical).

Also accidentally clicking the wrong server happens regularly with people that have participated in an event, and therefore sometimes end up in another event where they are not supposed to be (if this was a one time event, they should be able to delete the server after the event, right?)

Especially if most of the server addresses are IP numbers a mistake is easily made.
Unfortunately accidentally clicking the wrong server brings it back to the top of the list, increasing the chance of making the same mistake again.

In all of these cases it would be very appreciated if they just could delete that server address in a simple way.

gilgongo:
Just to be clear, you are taking about the combo boxes that store a list of previously selected addresses, is that right?

![combo](https://user-images.githubusercontent.com/4561747/113397107-6c848880-9394-11eb-97f0-e653e73e626a.png)


![image](https://user-images.githubusercontent.com/4561747/113396895-20d1df00-9394-11eb-91cf-372c881f65e6.png)

pgScorpio:
> Just to be clear, you are talking about the combo boxes that store a list of previously selected addresses, is that right?
> 
> ![combo](https://user-images.githubusercontent.com/4561747/113397107-6c848880-9394-11eb-97f0-e653e73e626a.png)
> 
> ![image](https://user-images.githubusercontent.com/4561747/113396895-20d1df00-9394-11eb-91cf-372c881f65e6.png)
> 
[...]

PR #1453, opened 2021-04-06
Line added to Combobox context menu to delete server from the list, issue #1421

dcorson-ticino-com:
ServerAddr combobox and CentralServer combobox
- select server to delete
- right click in combobox edit line
- choose Delete Server

### DIFF
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -399,10 +399,6 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
     // add "Delete Server" to cbxCentralServerAddress context menu
     cbxCentralServerAddress->setContextMenuPolicy ( Qt::CustomContextMenu ) ;
     m_deleteAction = new QAction ( tr ( "Delete Server" ), this );
-    connect ( m_deleteAction, SIGNAL ( triggered() ),
-              this, SLOT ( deleteServer() ));
-    connect ( cbxCentralServerAddress, SIGNAL ( customContextMenuRequested ( QPoint ) ),
-              this, SLOT ( contextMenuRequest() ) );
 
     // Connections -------------------------------------------------------------
     // timers
@@ -473,6 +469,12 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
         static_cast<void (QButtonGroup::*) ( QAbstractButton* )> ( &QButtonGroup::buttonClicked ),
         this, &CClientSettingsDlg::OnSndCrdBufferDelayButtonGroupClicked );
 
+    // for combobox context menu
+    QObject::connect  ( m_deleteAction, &QAction::triggered ,
+              this, &CClientSettingsDlg::deleteServer );
+
+    QObject::connect ( cbxCentralServerAddress, &CClientSettingsDlg::customContextMenuRequested,
+              this, &CClientSettingsDlg::contextMenuRequest );
 
     // Timers ------------------------------------------------------------------
     // start timer for status bar
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -816,12 +816,8 @@ void CClientSettingsDlg::OnInputBoostChanged()
 
 void CClientSettingsDlg::contextMenuRequest()
 {
-    // add line to context menu, is dynamic, made at each request
-    QMenu* menu = cbxCentralServerAddress->lineEdit()->createStandardContextMenu();
-    menu->addSeparator();
-    menu->addAction ( qaDeleteAction );
-    menu->popup ( QCursor::pos() );
-    connect ( menu, SIGNAL ( aboutToHide() ), menu, SLOT ( deleteLater() ) );
+    // add line to context menu
+    gfContextMenuRequest( cbxCentralServerAddress, qaDeleteAction );
 }
 
 void CClientSettingsDlg::deleteServer ()
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -825,7 +825,7 @@ void CClientSettingsDlg::deleteServer ()
     // delete selected line from cbxCentralServerAddress
     cbxCentralServerAddress->removeItem ( cbxCentralServerAddress->currentIndex() );
 
-    // Update IP addresse list
+    // Update IP address list
     int iCount = qMax ( cbxCentralServerAddress->count(), MAX_NUM_SERVER_ADDR_ITEMS );
     int iIdx;
     for ( iIdx = 0; iIdx < iCount; iIdx++ )
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -396,6 +396,13 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
 
     UpdateSoundCardFrame();
 
+    // add "Delete Server" to cbxCentralServerAddress context menu
+    cbxCentralServerAddress->setContextMenuPolicy ( Qt::CustomContextMenu ) ;
+    m_deleteAction = new QAction ( tr ( "Delete Server" ), this );
+    connect ( m_deleteAction, SIGNAL ( triggered() ),
+              this, SLOT ( deleteServer() ));
+    connect ( cbxCentralServerAddress, SIGNAL ( customContextMenuRequested ( QPoint ) ),
+              this, SLOT ( contextMenuRequest() ) );
 
     // Connections -------------------------------------------------------------
     // timers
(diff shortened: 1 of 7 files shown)


src/clientsettingsdlg.cpp
```diff
@@ -396,6 +396,13 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
 
     UpdateSoundCardFrame();
 
+    // add "Delete Server" to cbxCentralServerAddress context menu
+    cbxCentralServerAddress->setContextMenuPolicy ( Qt::CustomContextMenu ) ;
+    m_deleteAction = new QAction ( tr ( "Delete Server" ), this );
+    connect ( m_deleteAction, SIGNAL ( triggered() ),
```
pljones, line 402: See lines 407 onward.  Could you move your connections to that section and follow the same code style, please.  (I know it's Qt4 style...)

dcorson-ticino-com: Done, also in connectdlg, any other comments?
I'll push all at once.

hoffie (commented):
Some initial feedback (mostly coding style).
I've also noticed that:
- I was able to make Jamulus segfault after repeatedly adding and removing entries. I was not able to reproduce this though.
- The UI becomes laggy (opening the dropdown or typing) when adding entries with more than 10 (?) characters. This is reproducible and does not happen with 3.6.2.

I have not checked yet whether

dcorson-ticino-com:
Mmm... I have been trying now 15 minutes to get it to slow down or do anything else untoward, to no avail.
As well with entries with more than 10 char and more than 10 servers in the list.
Any other clues what might be happening, I can't reproduce up 'til now.
I'm on Win10.

gilgongo:
Can you post a screenshot of what this looks like in the UI?

dcorson-ticino-com:
Sure...
Note that I just added the separator and the last line, the rest was already present in 3.7.0.  I was unable to find where the rest is defined, but thought it could be useful especially on tablets so decided not to dig further and left it.

DeleteServer

hoffie:
> Mmm... I have been trying now 15 minutes to get it to slow down or do anything else untoward, to no avail.
> As well with entries with more than 10 char and more than 10 servers in the list.
> Any other clues what might be happening, I can't reproduce up 'til now.
> I'm on Win10.

Now I can't reproduce it anymore either. :(
Sorry for the noise, I have no idea what happened.

I was adding entries by typing 0<Enter>1<Enter>2<Enter>... into the field. As soon as I had an entry with read 01234567890, the cursor started moving really slowly and the opening the dropdown was slow as well. When I deleted all entries and directly added one with 01234567890, the issue appeared as well. I then ran the same test with 3.6.2 and did not see it. Now I don't see it with this PR either.

src/clientsettingsdlg.cpp
```diff
@@ -804,3 +813,31 @@ void CClientSettingsDlg::OnInputBoostChanged()
     pSettings->iInputBoost = cbxInputBoost->currentIndex() + 1;
     pClient->SetInputBoost ( pSettings->iInputBoost );
 }
+
+void CClientSettingsDlg::contextMenuRequest()
+{
+    // add line to context menu, is dynamic, made at each request
+    QMenu* menu = cbxCentralServerAddress->lineEdit()->createStandardContextMenu();
```
pljones, line 820: This seems like unnecessary overhead.  Define the menu once, on start up, with the extra item in it, all connected up.  Then here, just call it's popup action.

src/connectdlg.cpp
```diff
@@ -923,3 +933,31 @@ void CConnectDlg::DeleteAllListViewItemChilds ( QTreeWidgetItem* pItem )
         delete pCurChildItem;
     }
 }
+
+void CConnectDlg::contextMenuRequest()
+{
+    // add line to context menu, is dynamic, made at each request
+    QMenu* menu = cbxServerAddr->lineEdit()->createStandardContextMenu();
```
pljones, line 940: As in `src/clientsettingsdlg.cpp`.  Perhaps define the menu in `src/util.cpp`, or somewhere common to both client and server, and use the same one.

src/clientsettingsdlg.cpp
pljones: Maybe keep the "connect it up" separate, so the client and server can share the menu.

src/connectdlg.cpp
```diff
@@ -923,3 +933,27 @@ void CConnectDlg::DeleteAllListViewItemChilds ( QTreeWidgetItem* pItem )
         delete pCurChildItem;
     }
 }
+
+void CConnectDlg::contextMenuRequest()
+{
+    // add line to context menu
+gfContextMenuRequest( cbxServerAddr, qaDeleteAction );
```
pljones, line 940: Missing indent.

src/clientsettingsdlg.cpp
```diff
@@ -804,3 +813,27 @@ void CClientSettingsDlg::OnInputBoostChanged()
     pSettings->iInputBoost = cbxInputBoost->currentIndex() + 1;
     pClient->SetInputBoost ( pSettings->iInputBoost );
 }
+
+void CClientSettingsDlg::contextMenuRequest()
```
pljones, line 817: Lots of missing indents.

src/clientsettingsdlg.cpp
```diff
@@ -804,3 +813,27 @@ void CClientSettingsDlg::OnInputBoostChanged()
...
+    // add line to context menu
+    gfContextMenuRequest( cbxCentralServerAddress, qaDeleteAction );
+}
+
+void CClientSettingsDlg::deleteServer ()
+{
+    // delete selected line from cbxCentralServerAddress
+    cbxCentralServerAddress->removeItem ( cbxCentralServerAddress->currentIndex() );
+
+    // Update IP addresse list
```
pljones, line 828: `addresse` -> `address`?

src/connectdlg.cpp
```diff
@@ -923,3 +933,27 @@ void CConnectDlg::DeleteAllListViewItemChilds ( QTreeWidgetItem* pItem )
...
+}
+
+void CConnectDlg::deleteServer()
```
pljones, line 943: Could you pass the array reference to this method, so it can be reused, as otherwise there's quite a bit of duplicated logic.

src/util.h
```diff
@@ -101,6 +101,9 @@ inline int CalcBitRateBitsPerSecFromCodedBytes ( const int iCeltNumCodedBytes,
 
 QString GetVersionAndNameStr ( const bool bWithHtml = true );
 QString MakeClientNameTitle ( QString win, QString client );
+#ifndef HEADLESS
```
pljones, line 104: Similarly, the `GUI Utilities` section starts at line 371, so this should be added inside the existing `ifndef`, around line 497.

pljones:
Not for this PR -- it might be nice to add the client Custom Central Server Address functionality to the server, too.

dcorson-ticino-com:
>Could you pass the array reference to this method, so it can be reused, as otherwise there's quite a bit of duplicated logic.

I did the changes except this one.  I had tried for the last commit to do this, but have arrived at the limits of my C++.
I could not find a way to pass the vector of strings that would allow either indexing or incrementing in the function, so I left it as it worked.

pljones:
> 
> 
> > Could you pass the array reference to this method, so it can be reused, as otherwise there's quite a bit of duplicated logic.
> 
> I did the changes except this one. I had tried for the last commit to do this, but have arrived at the limits of my C++.
> I could not find a way to pass a pointer to the vector of strings that would allow either indexing or incrementing in the function, so I left it as it worked.

That's okay - I wasn't sure myself how to do it...  If someone does get to adding this to the server, it might be worth spending more time on. I'm approving as is now.

pljones (dismissed):
Thanks!

gilgongo:
I think a context menu for removing is the wrong decision, since it's literally the only Jamulus function we have on right click. So if somebody finds it, they will be forever right-clicking around the app wondering if there are other hidden functions. That's wrong and best practice with context menus is to have all (or a constant subset) of functions available - or none. Better in this case to have simply had a delete icon (with explanatory tool tip) next to the current entry in the input field I think.

dcorson-ticino-com:
Note that all the editable fields in Jamulus already have a right click context menu, that is nothing new.
And adding a button to each line in a list for a function that will most never be used adds clutter and confusion on the screen.

hoffie:
> Note that all the editable fields in Jamulus already have a right click context menu, that is nothing new.

I agree.

@dcorson-ticino-com Looks like this will need rebase.

hoffie (dismissed):
Approving, but not merging yet due to @gilgongo's comment (+ conflicts). Maybe @ann0see can comment as well?

gilgongo:
> Note that all the editable fields in Jamulus already have a right click context menu, that is nothing new.

That's not my objection though. It's that this is the first time a _Jamulus function_ has been added to that menu. This means that users who see it will think (logically, if in error) that there may be other functions hidden behind right-clicks. It's a text book case of when not to use context menus, which are for when all objects have contexts which Jamulus doesn't. I posted a video of how an alternative UI would work for this, but I can't remember where that was now.

ann0see:
Not really a strong opinion on this. I think we should better delay it and think of an alternative UI. The right click also isn’t obvious in any means.

gilgongo:
@ann0see I found the demo of an alternative UI:


https://user-images.githubusercontent.com/4561747/115992158-1f2abe00-a5c4-11eb-952c-2f91f162479a.mp4

ann0see:
Yes. This would then also work on mobile.
Adding a "Do you want to delete <IP> from your server connection list?" question would be good to make clear what it does and prevent wrongly deleted entries.

gilgongo:
> Adding a "Do you want to delete from your server connection list?" question would be good to make clear what it does and prevent wrongly deleted entries.

We need to be very careful with "Are you sure?" messages, particularly on operations that are recoverable (if they do it in error they can probably just enter the address again, even if they have to look it up elsewhere). The reason is that applications that have "OK/cancel" dialogues in them can habituate users into clicking "OK" all the time, which is fine _most_ of the time until it isn't. The _preferred_ method is [undo](https://alistapart.com/article/neveruseawarning/), but I don't this this is important enough to need that.

pljones:
Having OK'd the code, I think @gilgongo is right about the UI/UX here.  Note that on the server, there's a similar issue where the recording directory needs to be unset.  That could have been an right-click entry but it isn't.  It's done like this:
image
(Note that the actual indicator is just some weird glyph I found - at some point a better "clear" icon might be nice...)

ann0see:
Looks as if this PR is stale. I'll close it and probably another approach should be used. If it's still valid, please re-open this PR

Issue #1284, opened 2021-03-17
Jamulus server: Wrong initial recording status wording in log


drummer1154:
**Describe the bug**
Start a Jamulus server with command line option '--norecord'.
The log will reflect this as "- recording will not be enabled" which is completely misleading because by sending SIGUSR2, recording **will** be enabled, reflected in the log by "Recording state enabled".
It is only an issue on line 373 in main.cpp which shall be changed to contain "- Recording state: disabled"

**To Reproduce**
see above

**Expected behavior**
If a Jamulus server - otherwise ready for recording - is invoked with command line option '--norecord', the log shall reflect this by "- Recording state: disabled", else "- Recording state: enabled"

**Screenshots**
see above

**Operating system**
any

**Version of Jamulus**
3.6.2

**Additional context**
- I did not quickly find a location to contain the default text "- Recording state: enabled".
- the upper/lower case of the log texts shall be consolidated.

hoffie:
Thanks for your report!

> I did not quickly find a location to contain the default text "- Recording state: enabled".

From a quick glance, it looks like it comes from here:

https://github.com/jamulussoftware/jamulus/blob/976ad09cf0806842800cce4b6dee175065e9c5d2/src/recorder/jamcontroller.cpp#L56

Do you want to send a PR for this?

drummer1154:
@hoffie Please see https://github.com/jamulussoftware/jamulus/issues/1283#issuecomment-802433582

drummer1154:
I think a "recording control" finite state machine (FSM) can have these states:
o **Disabled** - this is the initial state. If there was no command line argument "-R" or the specified directory is not writable, then this state remains assumed for the current lifetime of the application. Reception of SIGUSR1 or SIGUSR2 has no effect in this state.
o **Enabled** - this is assumed if -R is followed by a writable directory. Reception of SIGUSR2 advances the state to **Active** if at least one client is connected or to **Armed** otherwise.
o **Armed** - in this state the recording function waits for a client to connect; if connected, the state advances to **Active**.
[...]

PR #1463, opened 2021-04-07
Recorder: Don't crash on failures / improve logging

### LINKED ISSUE #1289: Jamulus server (Linux) crashes when recorder cannot create recording session subdirectory
**Describe the bug**
- Run a Jamulus server as a service on Linux with jamulus.service containing
 ProtectHome=false
 User=jamulus_usr
 Group=nogroup
 :
 ExecStart=/usr/local/bin/Jamulus -R /home/vm_owner/Rec --norecord

- To toggle the recording state, issue "sudo /bin/systemctl kill -s SIGUSR2 jamulus"

- This will execute "QDir().mkpath(sessionDir.absolutePath())" in jamrecorder.cpp

- Most likely because there is an access right problem, the server crashes and the log contains
 -- Logs begin at Tue 2021-03-16 09:37:11 UTC. --
 Mar 16 13:27:47 vm-name jamulus[441]: Recording state enabled
 Mar 16 13:27:47 vm-name jamulus[441]: Qt has caught an exception thrown from an event handler. Throwing
 Mar 16 13:27:47 vm-name jamulus[441]: exceptions from an event handler is not supported in Qt.
 Mar 16 13:27:47 vm-name jamulus[441]: You must not let any exception whatsoever propagate through Qt code.
 Mar 16 13:27:47 vm-name jamulus[441]: If that is not possible, in Qt 5 you must at least reimplement
 Mar 16 13:27:47 vm-name jamulus[441]: QCoreApplication::notify() and catch all exceptions there.
 Mar 16 13:27:47 vm-name jamulus[441]: terminate called after throwing an instance of 'std::runtime_error'
 Mar 16 13:27:47 vm-name jamulus[441]:   what():  /home/vm_owner/Rec/Jam-20210316-132747781 does not exist but could not be created
 Mar 16 13:27:47 vm-name systemd[1]: jamulus.service: Main process exited, code=killed, status=6/ABRT
 Mar 16 13:27:47 vm-name systemd[1]: jamulus.service: Failed with result 'signal'.
 Mar 16 13:28:17 vm-name systemd[1]: jamulus.service: Service RestartSec=30s expired, scheduling restart.
 Mar 16 13:28:17 vm-name systemd[1]: jamulus.service: Scheduled restart job, restart counter is at 1.
 Mar 16 13:28:17 vm-name systemd[1]: Stopped Jamulus-Server.
 Mar 16 13:28:17 vm-name systemd[1]: Started Jamulus-Server.
 Mar 16 13:28:17 vm-name jamulus[2777]: - server mode chosen
 Mar 16 13:28:17 vm-name jamulus[2777]: - disconnect all clients on quit
 Mar 16 13:28:17 vm-name jamulus[2777]: - no GUI mode chosen
 Mar 16 13:28:17 vm-name jamulus[2777]: - maximum number of channels: 16
 Mar 16 13:28:17 vm-name jamulus[2777]: - using 64 samples frame size mode
 Mar 16 13:28:17 vm-name jamulus[2777]: - using multithreading
 Mar 16 13:28:17 vm-name jamulus[2777]: - welcome message: This is vm-name in FFM ...
 Mar 16 13:28:17 vm-name jamulus[2777]: - server info: vm-name;FFM;82
 Mar 16 13:28:17 vm-name jamulus[2777]: - recording directory name: /home/vm_owner/Rec
 Mar 16 13:28:17 vm-name jamulus[2777]: - recording will not be enabled
 Mar 16 13:28:18 vm-name jamulus[2777]: Recording state disabled
 Mar 16 13:28:18 vm-name jamulus[2777]:  *** Jamulus, Version 3.6.2
 Mar 16 13:28:18 vm-name jamulus[2777]:  *** Internet Jam Session Software
 Mar 16 13:28:18 vm-name jamulus[2777]:  *** Released under the GNU General Public License (GPL)

- All connected users are disconnected.

**To Reproduce**
See above

**Expected behavior**
If the sub-directory for the recording session cannot be created, the recording state shall revert back to disabled and the server shall continue to run with the currently connected users remaining connected.

**Screenshots**
See above

**Operating system**
GCP n1-standard-1, debian-10-buster-v20210217, Jamulus built from source
git clone https://github.com/corrados/jamulus.git
cd jamulus
git fetch
git checkout latest
=> HEAD is now at 337a4840 r3_6_2
qmake "CONFIG+=nosound" Jamulus.pro && make clean && make

**Version of Jamulus**
3.6.2

**Additional context**
See also #1284

hoffie:
This PR contains three recorder fixes.
They are individual commits which should be kept seperate as they do different things. Please don't squash-merge. :)

**Crash fixes**
Failures to create required directories or files for recording had been handled via exceptions already. However, these exceptions were not caught anywhere. This made them end up in Qt's default exception handler which crashed the whole application.

This commit adds exception handling, logging and propagation for the recorder failure state back to the jam controller. This fixes Jamulus crashes due to lack of permission, lack of disk space or other file system errors.

*(Please bear with me, this is my first attempt at exception handling/signals in C++/Qt)*

Fixes #1289

@drummer1154 Can you test this?

**Dead code removal**
This PR also removes dead code (the `SessionDirToReaper` function which is not called anywhere as far as I can see).

**Log improvements**
This PR also improves two log messages:
- `--norecord` logging on startup (#1284)
- Wording in an error message


pljones:
#1 The crash of the application is _intentional_.  Please restore it.

#2 The "unused" method is there because I use it when testing changes.  Please restore it.

#3 The "start up" logging isn't logging a command line option (i.e. as in main.cpp), it's logging a state change.  Please put it back how it was.

hoffie:
>   1. The crash of the application is _intentional_.

Ok, but what is the benefit of crashing? I think the current behavior is confusing for two reasons:
- This is a hard crash. The console output is full with Qt-internal messages and the actual error message is hidden in-between.
- Why do we have to crash (or exit) in the first place? It's not like a failure of the recorder leads Jamulus' core functionality into some kind of unrecoverable state. I'd certainly prefer to handle this gracefully and the two bug reports (+ a recent comment by one of the WorldJam people) seem to show that I'm not the only one.

>   2. The "unused" method is there because I use it when testing changes.  Please restore it.

Ok, I'll add a comment about that.

>   3. The "start up" logging isn't logging a command line option (i.e. as in main.cpp), it's logging a state change.

I assume you are referring to [this comment](https://github.com/jamulussoftware/jamulus/pull/1463/commits/118f6aa0b46cbe1c0c10c62427dd8e368438637a), right? I have not changed the format of this message. I only added two words to help clarify that `--norecord` does not disable recording completely, but rather disables automatic start of recording.

ann0see:
> The crash of the application is intentional. Please restore it.

But probably not during runtime? I'd be very upset if the server just crashed because there's no space left on the HDD. I'd check if there are issues on first start and exit if there's an error. If this happens during runtime, recording should end, the error should be logged but the server should still run to not interrupt running sessions.

hoffie:
> I'd check if there are issues on first start and exit if there's an error.

Just to be clear: There are two places which perform checks and error handling:
- The first place is [JamRecorder::Init](https://github.com/jamulussoftware/jamulus/blob/c7abaa7fd0b4e7efcf0b1c7c90b46fac3062e553/src/recorder/jamrecorder.cpp#L370). This checks that the given recording path is a writable directory (or can be created). This uses regular error handling and shows "nice" error messages. It runs on startup, even with `--norecord`, but it is not fatal (i.e. it doesn't crash or exit Jamulus). We could make these checks exit Jamulus if necessary.
- The second place is what this PR touches. It's about checks or error handling which happens when starting the actual recording (which may be triggered by signals). Those checks literally crashed if they triggered and this PR attempts to fix them.

So, if the intention is to exit Jamulus when errors are detected on startup, I agree that this can have benefits and I can work on changing that behavior. Note however that this was not the case before this PR and this is not touched by this PR (yet).

pljones:
The aim is to crash _as soon as possible_ so that the problem can be resolved.

hoffie:
> The aim is to crash _as soon as possible_ so that the problem can be resolved.

Do you suggest that for the `--norecord`, i.e. signal-triggered recording case or just for the record on startup case? I agree with the latter, but I think the first should be handled differently.

pljones:
If I'm recording, I want to know _immediately_ there's a problem, so I can fix it rather than lose what was about to be recorded.  The most obvious way to know something is wrong is to have the server stop working.

If there's a way that consistently writes _captured output_ stating the error (including Windows), then I'm happy for that to be used as an alternative to throwing an unhandled exception.  At the time, we couldn't find one.  (The exception _should_ end up in the Windows log -- whereas normal server output vanishes without trace.)

softins:
> If I'm recording, I want to know _immediately_ there's a problem, so I can fix it rather than lose what was about to be recorded. The most obvious way to know something is wrong is to have the server stop working.

I wonder if it might be good for the server to generate a chat message in bold red to say it could not start recording. Then everyone will see it, as the chat window opens automatically when a chat message is sent. At least people would know what the problem is then, rather than just wondering why the server has stopped.

ann0see:
The chat message would need to be sent to every newly connected user too. Not just once

hoffie:
> The chat message would need to be sent to every newly connected user too. Not just once

Why wouldn't the existing recording indicator be sufficient for that though?

ann0see:
> Why wouldn't the existing recording indicator be sufficient for that though?

We'd need to change the wording then?

hoffie:
@softins You started to add the chat-based notification yesterday and noticed that messages were created once per client. This issue also occured with this PR and I think I've fixed this now.

A summary of what should happen (IMO) and what currently happens with this PR when a post-start recording fails:
- [x] The operator of a headless server should be notified once about the problem.
- [ ] The operator of a UI server should be notified once about the problem
  - [ ] Recording state should go back to "stopped". This doesn't work yet.
  - [ ] An error message should appear (?). This doesn't happen yet.
- [ ] Users should be notified.
  - [ ] Recording state must be broadcast as "disabled" again. This doesn't work yet.
  - [ ] If there is additional need to notify users, a single Chat message can be sent (I'm not yet sure if this is required though?)

As multiple items are still unchecked, I'm converting this PR back to a draft.

hoffie:
A quick update:
- @softins has improved the failure handling in a way to ensure that clients receive the proper "no recording" notification in their UI
- I've added a message box for failures upon on-demand recording start
- I've added logic to sync the recorder checkbox state with the actual state in order to provide a way to re-enable recordings after fixing permissions/cleanup up the disk

hoffie:
> Checkbox behavior is still wrong when no client is connected and one wants to enable recording

This should now also be fixed, so this PR is ready for review.

> If there is additional need to notify users, a single Chat message can be sent (I'm not yet sure if this is required though?)

I'm not sure if this must be done as part of this PR and if it must be done at all. Personally, I don't think so.

src/recorder/jamcontroller.h
```diff
@@ -73,4 +75,7 @@ class CJamController : public QObject
 
 }
 
+// this must be included AFTER the above definition of class CJamController
```
pljones, line 78: Why?  (i.e. explain it in the comment)

src/recorder/jamcontroller.cpp
```diff
@@ -26,7 +26,8 @@
 
 using namespace recorder;
 
-CJamController::CJamController() :
+CJamController::CJamController( CServer* server ) :
```
pljones, line 29: Tiny point: the naming would be either `pServer` (pointer to Server Class) or `pNServer` (if you want to avoid the initialiser having the same name as the member - not that it matters - for "new" pointer to ...).

src/recorder/jamrecorder.cpp
```diff
@@ -160,7 +160,7 @@ CJamSession::CJamSession(QDir recordBaseDir) :
 
     if (!fi.exists() && !QDir().mkpath(sessionDir.absolutePath()))
     {
-        throw std::runtime_error( (sessionDir.absolutePath() + " does not exist but could not be created").toStdString() );
+        throw std::runtime_error( (sessionDir.absolutePath() + " does not exist and could not be created").toStdString() );
```
pljones, line 163: Maybe throw/catch `CGenErr`, as it's there...  Not many examples of it being used.

src/recorder/jamrecorder.cpp
```diff
@@ -396,15 +396,30 @@ void CJamRecorder::Start() {
...
+            currentSession = nullptr;
+            error = err.what();
+        }
     }
     ChIdMutex.unlock();
 
-    emit RecordingSessionStarted ( currentSession->SessionDir().path() );
+    if ( currentSession )
+    {
+        emit RecordingSessionStarted ( currentSession->SessionDir().path() );
```
pljones, line 419: Put these inside the `try {}` rather than after the `catch {}`.

src/recorder/jamrecorder.cpp
```diff
@@ -396,15 +396,30 @@ void CJamRecorder::Start() {
...
+        return;
```
pljones, line 420: I'd rather `else` emit the error, than `return` here, style-wise.

src/recorder/jamrecorder.cpp
```diff
@@ -585,6 +602,12 @@ void CJamRecorder::OnFrame(const int iChID, const QString name, const CHostAddre
         Start();
     }
 
+    // Start() may have failed, so check again:
+    if ( !isRecording )
+    {
+        return;
```
pljones, line 608: This `return` is okay.

src/recorder/jamrecorder.h
```diff
@@ -178,6 +178,7 @@ class CJamRecorder : public QObject
 
 signals:
     void RecordingSessionStarted ( QString sessionDir );
+    void RecordingFailed ( QString error );
```
pljones, line 181: This would become `CGenErr error`.

src/serverdlg.cpp
```diff
@@ -600,6 +600,12 @@ void CServerDlg::OnServerStopped()
 void CServerDlg::OnStopRecorder()
 {
     UpdateRecorderStatus ( QString::null );
+    if ( pServer->GetRecorderErrMsg() != QString::null )
```
pljones, line 603: This is a bit ugly.  The recorder should have an `inError` state, exposed by the server as `GetRecorderInErrorState()` boolean.  I don't like checking for null strings as a state indicator...

pljones:
If we're going to improve it, I'd take the approach of making it work properly.  My original approach didn't...  (C++ seemed very strange, Qt even more so, and the Jamulus code even more!)

Now that `CJamController` exists, it needs to take control.  It needs to track the state of the recorder(s).  If recording fails, it needs to expose that state and the reason for failure and emit the failure indicator.  It should handle failure not only when starting to record but when writing any frame (as that's when the disk could fill).

The current method of sticking the error message in the filename is horrible...  The server UI should get told by a signal something went wrong during recording (possibly right at the start) and the details should be available to retrieve via the server.

As part of testing this I'll want to see:
* Evidence that failure is logged on Windows (to the system event log for monitoring), MacOS (I've no idea where usually gets monitored...) and Linux (journald, syslog, or whatever).
* A list of scenarios that were tested, what happened for each on the old code, what happens now (with the outputs in both cases - for clients, servers and logs)
* Bonus points for anything else you can think of to keep me happy...

src/recorder/jamcontroller.h
softins: I've just added a clarifying comment.

src/serverdlg.cpp
hoffie: I'll take a look at that.

hoffie:
> Now that CJamController exists, it needs to take control. It needs to track the state of the recorder(s). If recording fails, it needs to expose that state and the reason for failure and emit the failure indicator. It should handle failure not only when starting to record but when writing any frame (as that's when the disk could fill).
> 
> The current method of sticking the error message in the filename is horrible... The server UI should get told by a signal something went wrong during recording (possibly right at the start) and the details should be available to retrieve via the server.

I agree, but I'm a bit hesitant to do it in this PR. I started the PR as an attempt o "fix the bleeding". I think it would be good to get the most relevant fixes merged rather sooner than later (with the hope to get them into 3.8.0). We could have another issue for improving the other places that you've mentioned.

> Evidence that failure is logged on Windows (to the system event log for monitoring), MacOS (I've no idea where usually gets monitored...) and Linux (journald, syslog, or whatever).

I suspect that the failure will not be logged to the Windows event log. As far as I can see, that would [require explicit API calls](https://stackoverflow.com/questions/37035958/log-to-event-viewer-on-windows-with-c/37036187#37036187). If we want this kind of integration, then we should probably do it globally as part of a [QtMessageHandler](https://doc.qt.io/qt-5/qtglobal.html#qInstallMessageHandler) for all messages above a specific thresholding (e.g. all warnings). I'd argue that this should not be part of this PR.
No idea regarding MacOS.
On Linux, this depends on the supervisor. If started by systemd, journald should have it. If started manually, it will probably be console-only.

@softins FYI I've rebased and integrated your comment adjustment to your initial commit.

softins:
> @softins FYI I've rebased and integrated your comment adjustment to your initial commit.

Thanks. I've held off actually doing a review until Peter is happy. I'm happy in general.

pljones:
> I've held off actually doing a review until Peter is happy.

You could be waiting a good while... ;)

> I'm happy in general.

(There's an answer for that, too ...!)

I'll take a look now.

src/main.cpp
```diff
@@ -377,7 +377,7 @@ int main ( int argc, char** argv )
                                "--norecord" ) )
         {
             bDisableRecording = true;
-            qInfo() << "- recording will not be enabled";
+            qInfo() << "- recording will not be enabled on startup";
```
pljones, line 380: Might even be better to say "recording will not take place until enabled", which is what it means.  I don't like the "on startup" -- it's not really about "on startup" - it's until further action.

src/recorder/jamrecorder.cpp
pljones: Yeah...  I hadn't seen CGenErr - there aren't many, saw my first one a few weeks ago!

src/recorder/jamrecorder.cpp
pljones: I don't feel here it's an early return.  It's
```
void method() {
 // always
 if ( a ) {
  // something if a
 } else {
  // something else
 }
}
```
There's nothing after the end of the if/else before the method returns all by itself.  We're basically saying "either emit this or emit that" and idiomatically that's an if/else.

src/recorder/jamcontroller.cpp
```diff
@@ -169,6 +174,19 @@ void CJamController::SetRecordingDir ( QString newRecordingDir,
...
+{
+    if ( !bEnableRecording )
+    {
+        // Recording has already been stopped, possibly
+        // by a previous OnRecordingFailed call from another client/thread.
+        return;
+    }
+    strRecorderErrMsg = error;
+    qWarning() << "Could not start recording:" << error;
+    pServer->SetEnableRecording ( false );
```
pljones, line 187: Might be worth a comment saying what we're doing here.  Does this just stop the _current_ recording or turn off the recorder until it's turned back on?

src/recorder/jamrecorder.cpp
hoffie: > There's nothing after the end of the if/else before the method returns all by itself. We're basically saying "either emit this or emit that, done" and idiomatically that's an if/else.

I agree. It has little benefits right now. However, if it is ever extended, it would keep the code path simpler. This sounds like YAGNI, but at the same time we said in the referenced ticket, that it is hard to re-structure existing code that way without risk of breakage, so... if we want to get better there, we might want to start early.

Nevertheless, I've adjusted the code like you've suggested now (with the next push).

pljones: The question is whether it was the right place for the return then.  If it was "return on failure", it should have been after the other emit.  It was more likely that the if block with the return (as the "positive" side) would be the one to grow.  This way, should any code get added to the positive side, then the if can be easily inverted and a return added.

Perhaps that's the better option now:
```
void method() {
 if ( !itWorked() ) {
  emit itFailed();
  return;
 }
 emit itDidNotFail();
 doMoreStuffMaybeOneDay();
}
```

src/recorder/jamrecorder.h
hoffie: As discussed, I'm not changing that for now as I'm having troubles with exactly that declaration (maybe due to lack of such a constructor?). The alternative would be passing by reference, which I'm not sure is a safe thing to do from catch blocks and to pass on via signals.

src/recorder/jamrecorder.cpp
hoffie: Yes, I think what you've suggested is the best version. I have adapted the code accordingly.

src/main.cpp
hoffie: Done.

src/recorder/jamrecorder.cpp
hoffie: Done.

src/serverdlg.cpp
hoffie: I had a quick look. The `jamcontroller.*` part seemed manageable, but as the error messages are coming from `CJamRecorder::Init()`, this would have required refactoring as well (probably to use exceptions and catch them in the caller...?). As you say, there is already existing code working this way.

To avoid another round of reviews for that, I'd like to keep this out of this PR. Maybe we should have a new issue to track that?

src/recorder/jamcontroller.cpp
hoffie: Added a comment.

src/recorder/jamcontroller.cpp
hoffie: Changed to `pNServer` in all three relevant places.

hoffie:
Thanks for the reviews. Hopefully all reviews are addressed now.

As for testing: The most important change is the reworked error handling in 589ff6163311af9b6ee5655fbac80a9cae57823e.

This can be tested as follows:
- `Jamulus -s -n -R /tmp/foo` (with /tmp/foo not existing yet or being a writable directory)
- `rmdir /tmp/foo && touch /tmp/foo`
- Connect a client

Behavior before this PR:
Jamulus server crashes. Client gets disconnected.

Behavior after this PR:
Jamulus server keeps running, an error gets logged. Client stays connected, recording state stays disabled.

```
2021-05-01 00:09:10, 127.0.0.1, connected (1)
Could not start recording: "/tmp/foo/Jam-20210430-220910823 does not exist and could not be created"
Recording state: disabled
```

When running a GUI server, the same message is logged. In addition, an error dialog is shown and the "Activate recording" checkbox is unchecked (to provide a way to re-enable later).
recorder-error

With this PR, the error handling is also confirmed to work when triggered via signals. This is also true when `--norecord` was passed initially.
With this PR, it is also confirmed that recording can still be started properly and that recording indicators work.

src/recorder/jamcontroller.h
```diff
@@ -73,4 +75,9 @@ class CJamController : public QObject
 
 }
 
+// This must be included AFTER the above definition of class CJamController,
+// because server.h defines CServer, which contains an instance of CJamController,
+// and therefore needs to know the its size.
+#include "../server.h"
```
pljones, line 81: Oh.  Are you sure it can't just forward-declare the class?
```
...
class CJamController;
...
class CSomethingElse
{
  ...
  CJamController JamController;
  ...
}
```

src/recorder/jamrecorder.cpp
```diff
@@ -396,14 +396,30 @@ void CJamRecorder::Start() {
...
+        {
+            currentSession = nullptr;
+            error = err.GetErrorText();
+        }
     }
     ChIdMutex.unlock();
 
+    if ( !currentSession )
+    {
+        emit RecordingFailed ( error );
```
pljones, line 419: Ah, sorry...  If this is _only_ emitted on start of recording, I'd like it called `RecordingFailedToStart`.  That way, if ever there's a failure during recording, it can be named - i.e. signalled - separately (as handling _may_ need to be different).

pljones: Only worry about this if the change to the `#include "../server.h"` is relevant.

src/recorder/jamcontroller.h
npostavs: AFAIK, that could only work if `JamController` was a pointer rather than a full object.

src/recorder/jamcontroller.cpp
```diff
@@ -169,6 +174,22 @@ void CJamController::SetRecordingDir ( QString newRecordingDir,
...
 
+void CJamController::OnRecordingFailed ( QString error )
+{
+    if ( !bEnableRecording )
+    {
+        // Recording has already been stopped, possibly
+        // by a previous OnRecordingFailed call from another client/thread.
+        return;
+    }
+    strRecorderErrMsg = error;
```
pljones, line 185: We're single-threaded here, right?  So no need for mutex protection?

src/recorder/jamcontroller.h
pljones: Ahh, right.  All the code used to be part of CServer.  Cut'n'paste job to create CJamController (with some replace-all help) -- but I didn't try to clean up to that extent when Volker ask it to leave :).  Perhaps it should have been a pointer.  Not for now.

pljones (approved):
Can't see anything wrong now...

src/recorder/jamcontroller.cpp
hoffie: Both `bEnableRecording` and `strRecorderErrMsg` access is similar to other recorder code, so I think it should be fine.

As far as I understand multithreading in Jamulus there are two relevant places:
- Mixing may be multithreaded (`-T`). This shouldn't affect any of those variables.
- The recorder uses its own thread. `pJamRecorder` lives in that thread, but is not able to touch these variables (which are only used in `pJamController` which should live on the main thread).

softins (approved):
All looks good to me - thanks!

PR #1483, opened 2021-04-10
Issue templates: Guide users to Discussions for Support/Ideas

hoffie:
Example: https://github.com/hoffie/jamulus/issues/new/choose (last two entries only)

It would be even better to have these two entries first, but I haven't found a way to do that.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,8 @@
+contact_links:
+  - name: Support Request or Question
+    url: https://github.com/jamulussoftware/jamulus/discussions/new
+    about: Please ask questions by opening a new Discussion.
+
+  - name: Idea/Feature Request
+    url: https://github.com/jamulussoftware/jamulus/discussions/new
+    about: Please open a Discussion if you have ideas for new features or functions.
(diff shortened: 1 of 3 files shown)


gilgongo:
Nice! 

Coincidentally, in your admin doc draft, for the issues template suggested we say, _"Please note that feature requests need to link to the Github Discussion that gave rise them. Bugs must be clearly defined and (preferably) reproducible."_ 

I'm petty sure that GH will see sense and allow repos to restrict issue posts to team members only though, once Discussions is out of beta.

gilgongo (approved):

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,8 @@
+contact_links:
+  - name: Support Request or Question
+    url: https://github.com/jamulussoftware/jamulus/discussions/new
+    about: Please ask questions by opening a new Discussion.
+
+  - name: Idea
```
ann0see, line 6:
```suggestion
  - name: Idea/Feature Request 
```


I think this makes it a bit clearer that we don't want feature requests in the issues list since they'll probably need discussion

ann0see:
Great work!

ann0see (approved):
Thanks this looks great!

PR #1485, opened 2021-04-10
Jamulus -h correction

gilgongo:
Put -p and -Q in correct place.

PR #1486, opened 2021-04-10
--help: Fix QoS help string order

hoffie:


gilgongo (approved):

hoffie:
As @softins pointed out, #1485 should probably be reverted, so this PR is redundant as well.

PR #1500, opened 2021-04-12
Use aqtinstall instead of brew

softins:
CodeQL for Mac is now failing with `brew`. This PR uses the same `aqtinstall` as the artifact build scripts, and makes CodeQL succeed again.


softins:
Looks like the problem with `brew` was temporary, but I still think the changes in this PR are an improvement, and maybe a step towards combining the artifact and CodeQL builds for Mac.

pljones (approved):
Thank you!

hoffie (approved):
I know little about brew or aqt, but CI is green, logs look sane and synchronizating CodeQL and artifact build sounds like a good idea.

PR #1514, opened 2021-04-14
Use transparent pixel for slider groove.

### LINKED ISSUE #1513: QFSFileEngine::open: No file name specified
**Describe the bug**
When Jamulus client is run from the command line, the warning message `QFSFileEngine::open: No file name specified` gets displayed at times in the program output. It happens every time a new fader gets displayed.

**To Reproduce**
Start Jamulus from the command line: `Jamulus` or `./Jamulus`, etc. Then connect to a server. The message will be displayed for each displayed channel.

**Expected behavior**
No warning message.

**Operating system**
Observed on Linux, running on Raspberry Pi or Intel PC.

**Version of Jamulus**
All versions to 3.7.0

**Additional context**
The cause is this line in `src/audiomixerboard.cpp`:
```
"QSlider::groove { image:          url();"
```
`url()` should not have a missing parameter.

softins:
Removing the `url()` entry completely didn't work - it caused the
slider itself to be displayed wrongly. So instead, a transparent
pixel png is used.

Fixes #1513


hoffie (approved):
Looks good to me.
Does not affect the Normal style (as expected).

Jamulus.pro
```diff
@@ -745,6 +745,7 @@ DISTFILES += ChangeLog \
     src/res/fronticon.png \
     src/res/fronticonserver.png \
     src/res/mixerboardbackground.png \
+    src/res/1x1.png \
```
pljones, line 748: It's not just _any_ 1x1 pixel, is it.  Can it have a more meaningful name, please?

softins: Yeh, valid point. But since it could be used in multiple ways, I'll just call it `transparent1x1.png`.

pljones (approved):
Good enough... I suppose... :)

PR #1516, opened 2021-04-14
Jamulus.pro: Support CONFIG+=debug

hoffie:
Previously, there was -- to my knowledge -- no way to create non-release builds without editing Jamulus.pro. This commit only adds CONFIG+=release when CONFIG does not contain "debug".

Per the [docs](https://doc.qt.io/qt-5/qmake-project-files.html#general-configuration):
> The project can be built in release mode or debug mode, or both. If debug and release are both specified, the last one takes effect.

----

A second commit in this PR fixes some whitespace in Jamulus.pro.

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -8,11 +8,11 @@ contains(CONFIG, "noupcasename") {
 
 # allow detailed version info for intermediate builds (#475)
 contains(VERSION, .*dev.*) {
-    exists(".git/config"){
+    exists(".git/config") {
         GIT_DESCRIPTION=$$system(git describe --match=xxxxxxxxxxxxxxxxxxxx --always --abbrev --dirty) # the match should never match
         VERSION = "$$VERSION"-$$GIT_DESCRIPTION
         message("building version \"$$VERSION\" (intermediate in git repository)")
-    }else{
+    } else {
         VERSION = "$$VERSION"-nogit
         message("building version \"$$VERSION\" (intermediate without git repository)")
     }
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -22,8 +22,7 @@ contains(VERSION, .*dev.*) {
 
 CONFIG += qt \
     thread \
-    lrelease \
-    release
+    lrelease
 
 QT += network \
     xml \


npostavs:
> Previously, there was -- to my knowledge -- no way to create non-release builds without editing Jamulus.pro.

It may not be applicable to msvc, but with msys/mingw I got the `debug_and_release` behaviour described in https://doc.qt.io/qt-5/qmake-variable-reference.html#config

> When you use the `debug_and_release` option (which is the default under Windows), the project will be processed three times: one time to produce a "meta" Makefile, and two more times to produce a Makefile.Debug and a Makefile.Release.

hoffie:
> It may not be applicable to msvc, but with msys/mingw I got the `debug_and_release` behaviour described in https://doc.qt.io/qt-5/qmake-variable-reference.html#config
> 
> > When you use the `debug_and_release` option (which is the default under Windows), the project will be processed three times: one time to produce a "meta" Makefile, and two more times to produce a Makefile.Debug and a Makefile.Release.

Thanks for the pointer. I had seen this option but did not consider that it may work (despite `debug` not working).

I've tried **`qmake Jamulus.pro CONFIG+=debug_and_release && make -f Makefile.Debug`** and this worked for me.

That's not as short as what I've proposed, but on the other hand this is something for really technical users and/or developers, so I think this is fine.

I'm closing this PR as there is already a way to create debug builds without editing the project file.

npostavs:
> `make -f Makefile.Debug`

There's a pseudo-target such that `make debug` also works.

By the way, even though it's *possible* to get a debug build with the current code, I think your PR makes sense.

hoffie:
> There's a pseudo-target such that `make debug` also works.

Ah, even better :)

> By the way, even though it's _possible_ to get a debug build with the current code, I think your PR makes sense.

Hrm, why? It would only make the call slightly shorter, but would introduce additional code in `Jamulus.pro`, so I'm not sure if it's worth the tradeoff?

npostavs:
Having `CONFIG+=debug` silently not work is confusing. I'm a developer/technical user but I hadn't used qt before Jamulus; it definitely confused me. I'm still unsure how to get a debug build using qtcreator (see https://github.com/jamulussoftware/jamulus/pull/870#issuecomment-766461546 and https://github.com/jamulussoftware/jamulus/pull/870#issuecomment-767675567). I don't know if this is related to that, but the behaviour introduced by this PR makes things more sane,

But actually, is the 
```
!contains(CONFIG, "debug") {
    CONFIG += release
}
```
part needed? I seem to get the same behaviour without it. That is, when neither `debug` nor `release` is added to `CONFIG`, the resulting `Makefile` uses release options (`-O2` etc) in `CXXFLAGS` & `CFLAGS` (this is on my Linux server, as opposed to the mingw behaviour I described earlier).

softins:
> That is, when neither `debug` nor `release` is added to `CONFIG`, the resulting `Makefile` uses release options (`-O2` etc) in `CXXFLAGS` & `CFLAGS` (this is on my Linux server, as opposed to the mingw behaviour I described earlier).

That is what I have just observed too, so I agree with you, that just removing `release` from `Jamulus.pro` would not change default behaviour, but would allow someone to do `qmake 'CONFIG += debug' Jamulus.pro` if they wanted to.

softins:
@hoffie based on the above comments, could you update this just to remove `release` from `Jamulus.pro`? I think it is a worthwhile improvement. Don't forget also to remove the trailing `\` from `lrelease`.

hoffie:
> @hoffie based on the above comments, could you update this just to remove `release` from `Jamulus.pro`? I think it is a worthwhile improvement. Don't forget also to remove the trailing `\` from `lrelease`.

Yes, this sounds like a good idea. Thanks @softins and @npostavs :)
Did that now.

pljones (approved):

softins (approved):

Issue #1515, opened 2021-04-14
Jamulus server 3.7.0 terminates when started as systemd service when sending SIGUSR2 to toggle recording


helgeerbe:
**Describe the bug**
After updating to version 3.7.0 jamulus server is terminating by SIGTERM (15) after recieving SIGUSR2(12) to toggle recording.

````
Apr 14 12:02:38 jamulus jamulus[26951]: 2021-04-14 12:02:38, 79.202.58.202, connected (1)
Apr 14 12:02:46 jamulus sudo[26959]:    herbe : TTY=pts/1 ; PWD=/etc/systemd/system ; USER=root ; COMMAND=/usr/bin/systemctl kill -s SIGUSR2 jamulus
Apr 14 12:02:46 jamulus sudo[26959]: pam_unix(sudo:session): session opened for user root by herbe(uid=0)
Apr 14 12:02:46 jamulus jamulus[26951]: OnHandledSignal: 12
Apr 14 12:02:46 jamulus jamulus[26951]: Recording state: enabled
Apr 14 12:02:46 jamulus sudo[26959]: pam_unix(sudo:session): session closed for user root
Apr 14 12:02:46 jamulus jamulus[26951]: OnHandledSignal: 15
Apr 14 12:02:46 jamulus jamulus[26951]: Session RPP: "/home/herbe/recordings/Jam-20210414-120246148/Jam-20210414-120246148.rpp"
Apr 14 12:02:46 jamulus jamulus[26951]: Session LOF: "/home/herbe/recordings/Jam-20210414-120246148/Jam-20210414-120246148.lof"
Apr 14 12:02:46 jamulus jamulus[26951]: 2021-04-14 12:02:46,, server idling -------------------------------------
Apr 14 12:02:46 jamulus systemd[1]: jamulus.service: Succeeded.
Apr 14 12:04:22 jamulus sudo[26961]:    herbe : TTY=pts/1 ; PWD=/etc/systemd/system ; USER=root ; COMMAND=/usr/bin/systemctl restart jamulus.service
Apr 14 12:04:22 jamulus sudo[26961]: pam_unix(sudo:session): session opened for user root by herbe(uid=0)
Apr 14 12:04:22 jamulus systemd[1]: Started Jamulus-Server.
Apr 14 12:04:22 jamulus sudo[26961]: pam_unix(sudo:session): session closed for user root
Apr 14 12:04:22 jamulus jamulus[26965]: - using 64 samples frame size mode
[...]

PR #1517, opened 2021-04-14
Remove sh -c 

gilgongo:
As per https://github.com/jamulussoftware/jamulus/issues/1515#issuecomment-819753767

### DIFF
--- a/distributions/jamulus-server.service
+++ b/distributions/jamulus-server.service
@@ -8,7 +8,7 @@ Type=simple
 Restart=always
 RestartSec=1
 User=jamulus
-ExecStart=/bin/sh -c '/usr/bin/jamulus -s -n -l /var/log/jamulus -e jamulus.fischvolk.de -g -o "$(uname -n);;"'
+ExecStart=/usr/bin/jamulus -s -n -l /var/log/jamulus -e jamulus.fischvolk.de -g -o "$(uname -n);;"
 
 [Install]
 WantedBy=multi-user.target
--- a/distributions/jamulus-server.service
+++ b/distributions/jamulus-server.service
@@ -8,7 +8,7 @@ Type=simple
 Restart=always
 RestartSec=1
 User=jamulus
-ExecStart=/usr/bin/jamulus -s -n -l /var/log/jamulus -e jamulus.fischvolk.de -g -o "$(uname -n);;"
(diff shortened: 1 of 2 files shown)


distributions/jamulus-server.service
```diff
@@ -8,7 +8,7 @@ Type=simple
 Restart=always
 RestartSec=1
 User=jamulus
-ExecStart=/bin/sh -c '/usr/bin/jamulus -s -n -l /var/log/jamulus -e jamulus.fischvolk.de -g -o "$(uname -n);;"'
+ExecStart=/usr/bin/jamulus -s -n -l /var/log/jamulus -e anygenre1.jamulus.io:22124 -g -o "$(uname -n);;"
```
hoffie, line 11: I suspect `$(uname -n)` will not expand properly without the `sh -c` wrap. This would likely need the `exec` as @softins suggested (haven't tried myself):

```suggestion
ExecStart=sh -c 'exec /usr/bin/jamulus -s -n -l /var/log/jamulus -e anygenre1.jamulus.io:22124 -g -o "$(uname -n);;"'
```

helgeerbe (approved):
This looks good for me. And will be consistent with the wiki documentation.

distributions/jamulus-server.service
gilgongo: I would expect it to be OK since it's not a `sh` only thing. Putting `sh -c` back here will cause the problem with SIGUSR2, no?

helgeerbe: I wasn't aware of the second file. @hoffie suggested change will work. I already tested it with SIGUSR2.

softins (changes requested):
I feel strongly that the correct solution is to retain the `sh -c` and add the `exec`.

distributions/jamulus-server.service
npostavs: If I read https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Specifiers correctly

    ExecStart=/usr/bin/jamulus -s -n -l /var/log/jamulus -e anygenre1.jamulus.io:22124 -g -o %H

would work.

gilgongo:
@softins I see you're going to do that so I'll close this.

PR #1536, opened 2021-04-18
Windows servers: Use Qt::PreciseTimer for maximal precision

npostavs:
I'm not sure if this makes a real difference, but I don't see any reason not to use the best precision available here.

```
According to https://doc.qt.io/qt-5/qtimer.html#timerType-prop

    The default value for this property is Qt::CoarseTimer.

According to https://doc.qt.io/qt-5/qt.html#TimerType-enum

    On Windows, Qt will use Windows's Multimedia timer facility (if
    available) for Qt::PreciseTimer and normal Windows timers for
    Qt::CoarseTimer
```

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -58,6 +58,8 @@ CHighPrecisionTimer::CHighPrecisionTimer ( const bool bNewUseDoubleSystemFrameSi
     veciTimeOutIntervals[1] = 1;
     veciTimeOutIntervals[2] = 0;
 
+    Timer.setTimerType ( Qt::PreciseTimer );
+
     // connect timer timeout signal
     QObject::connect ( &Timer, &QTimer::timeout,
         this, &CHighPrecisionTimer::OnTimer );


DavidSavinkoff:
> I'm not sure if this makes a real difference, but I don't see any reason not to use the best precision available here.
> 
> ```
> According to https://doc.qt.io/qt-5/qtimer.html#timerType-prop
> 
>     The default value for this property is Qt::CoarseTimer.
> 
> According to https://doc.qt.io/qt-5/qt.html#TimerType-enum
> 
>     On Windows, Qt will use Windows's Multimedia timer facility (if
>     available) for Qt::PreciseTimer and normal Windows timers for
>     Qt::CoarseTimer
> ```

According to https://doc.qt.io/qt-5/qt.html#TimerType-enum

On UNIX (including Linux, macOS, and iOS), Qt will keep millisecond accuracy for Qt::PreciseTimer. **For Qt::CoarseTimer**, the interval will be adjusted up to 5% to align the timer with other timers that are expected to fire at or around the same time. **The objective is to make most timers wake up at the same time, thereby reducing CPU wakeups and power consumption.**
- Non-Windows may be disadvantaged by #1526. Maybe #1526 should be ifdef'd for Windows only.

npostavs:
@DavidSavinkoff The code affected by this PR is already inside `#ifdef _WIN32` anyway. 

Other platforms use nanosecond sleep functions. Which probably cause even more CPU wakeups/power consumption than the 1ms Qt timer. But I'm not sure why you think trading power increased consumption for better precision is a **dis**advantage. Rather, it looks like the correct choice to me.

DavidSavinkoff:
> The code affected by this PR is already inside #ifdef _WIN32 anyway.

Good, I didn't know that.

> But I'm not sure why you think trading power increased consumption for better precision is a disadvantage.

Some CPUs may get bogged down because they don't have enough brute force.

The timer interrupts are separate from the data interrupts. There is no substitution that will be as good as using the data interrupts (per frame). Increasing the timer interrupts is a brute force punishment. Thus, it is better to have an objective to synchronize the interrupts or data if you can. This Qt wakeup and power consumption thing may also service the data quicker too (at least it is efficient).

DavidSavinkoff:
Data interrupts will normally be serviced with a driver or kernel module in a general purpose OS. An ethernet card will signal (interrupt) the OS when it needs more data to send, or has received a block of data. The sound card will do the same for 64 128 ... sized buffers. In Jamulus, the server doesn't deal with the sound card, but it must manage blocks of data from many clients. I'm not even sure the Jamulus server needs sophisticated timers because the clients generate the sound data; as long as the server keeps up with the clients, it is doing its job. The server cannot run ahead of all of the clients because it uses client data (which arrives at regular intervals).

DavidSavinkoff (approved):

npostavs:
> I'm not even sure the Jamulus server needs sophisticated timers because the clients generate the sound data; as long as the server keeps up with the clients, it is doing its job.

You need a timer, otherwise the server could wait an unbounded amount of time for a dropped packet. And generally, I don't think the server can rely on data-driven interrupts because there are no guarantees on when (and in what order) the data will arrive.

hoffie (approved):
Logic-wise, this seems correct and useful to me. I don't have a Windows machine to test.

@npostavs Have you confirmed that this does not make things worse for some unexpected reason?

npostavs:
Well I ran the server locally on my Windows box, connected a single client and it sounded okay. But I haven't thoroughly tested to make sure nothing gets worse in any situation.

pljones (approved):
This could help a lot for Windows users, I think.  Thanks!

PR #1542, opened 2021-04-19
Move network and stats to the right

pljones:
It can look like this:
image

### DIFF
--- a/src/clientsettingsdlgbase.ui
+++ b/src/clientsettingsdlgbase.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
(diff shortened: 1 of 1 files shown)


hoffie (approved):
This looks like a very good improvement. Thanks!

Pinging @ngocdh @dcorson-ticino-com who also had UI-related changes recently.

dcorson-ticino-com:
I wonder how this looks on a tablet, as that was the original complaint.
On the other hand any multi-window solution will probably run into trouble there.
I see no harm in this.

softins (approved):
Looks good to me. I'll go ahead and merge.

PR #1549, opened 2021-04-21
Multi level mixing

pljones:
Extend the earlier "one or two" rows for the mixer panel with "one to eight" rows.  Note that the sort order may work out differently.  I'd like feedback on that.

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -833,7 +833,7 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent ) :
     bNoFaderVisible      ( true ),
     iMyChannelID         ( INVALID_INDEX ),
     iRunningNewClientCnt ( 0 ),
-    iNumMixerPanelRows   ( 1 ),
+    iNumMixerPanelRows   ( 1 ), // pSettings->iNumMixerPanelRows is not yet available
     strServerName        ( "" ),
     eRecorderState       ( RS_UNDEFINED ),
     eChSortType          ( ST_NO_SORT )
@@ -1071,10 +1071,6 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
     // sort the channels according to the first of the pair
     std::stable_sort ( PairList.begin(), PairList.end() );
 
-    // calculate the number of the faders in the first row by distribute
-    // the faders equally in the available number of rows
-    const int iNumFadersFirstRow = ( iNumVisibleFaders + 1 ) / iNumMixerPanelRows;
-
     // add channels to the layout in the new order, note that it is not required to remove
     // the widget from the layout first but it is moved to the new position automatically
     int iVisibleFaderCnt = 0;
@@ -1088,8 +1084,8 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
             // per definition: the fader order is colum-first/row-second (note that
             // the value in iNumFadersFirstRow defines how many rows we will get)
             pMainLayout->addWidget ( vecpChanFader[iCurFaderID]->GetMainWidget(),
-                    iVisibleFaderCnt / iNumFadersFirstRow,
-                    iVisibleFaderCnt % iNumFadersFirstRow );
+                    iVisibleFaderCnt % iNumMixerPanelRows,
+                    iVisibleFaderCnt / iNumMixerPanelRows );
 
             iVisibleFaderCnt++;
         }
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1082,7 +1082,7 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
         if ( vecpChanFader[iCurFaderID]->IsVisible() )
         {
             // per definition: the fader order is colum-first/row-second (note that
-            // the value in iNumFadersFirstRow defines how many rows we will get)
+            // the value in iNumFadersFirstRows defines how many rows we will get)
             pMainLayout->addWidget ( vecpChanFader[iCurFaderID]->GetMainWidget(),
                     iVisibleFaderCnt % iNumMixerPanelRows,
                     iVisibleFaderCnt / iNumMixerPanelRows );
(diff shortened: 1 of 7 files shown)


src/clientsettingsdlg.cpp
```diff
@@ -334,6 +334,11 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
         "you will not have much fun using the " ) + APP_NAME +
         tr ( " software." ) + TOOLTIP_COM_END_TEXT );
 
+    QString wtNumMixerPanelRows = "<b>" + tr ( "Number of Mixer Panel Rows" ) + ":</b> " +
```
hoffie, line 337: Convention for all other What's this string variables seems to be to use str*.

```suggestion
    QString strNumMixerPanelRows = "<b>" + tr ( "Number of Mixer Panel Rows" ) + ":</b> " +
```

hoffie (changes requested):
I see that this may help for large groups, yeah. :+1: 

I've tested this locally and the binary is crashing for me on Linux as soon as I start it in GUI mode, even with a fresh ini file.

I suspect it's related to `pSettings` initialization order?

```
Thread 1 "Jamulus" received signal SIGSEGV, Segmentation fault.
CAudioMixerBoard::CAudioMixerBoard (this=0x555555b8ded0, parent=0x555555a31340) at src/audiomixerboard.cpp:836
836	    iNumMixerPanelRows   ( pSettings->iNumMixerPanelRows ),
(gdb) bt
#0  CAudioMixerBoard::CAudioMixerBoard (this=0x555555b8ded0, parent=0x555555a31340) at src/audiomixerboard.cpp:836
#1  0x000055555567b53d in Ui_CClientDlgBase::setupUi (this=0x7ffffffab610, CClientDlgBase=0x7ffffffab5e0)
    at ./ui_clientdlgbase.h:383
#2  0x0000555555670bf6 in CClientDlg::CClientDlg (this=0x7ffffffab5e0, pNCliP=0x7ffffffac010, pNSetP=0x7ffffffab390, 
    strConnOnStartupAddress=..., strMIDISetup=..., bNewShowComplRegConnList=false, bShowAnalyzerConsole=false, 
    bMuteStream=false, parent=0x0) at src/clientdlg.cpp:49
#3  0x00005555555a883a in main (argc=1, argv=0x7fffffffe1a8) at src/main.cpp:743
(gdb) print pSettings
$1 = (CClientSettings *) 0x0
```

src/clientsettingsdlg.cpp
pljones: Ah, I've no idea what I was thinking..!

src/audiomixerboard.cpp
```diff
@@ -833,7 +833,7 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent ) :
     bNoFaderVisible      ( true ),
     iMyChannelID         ( INVALID_INDEX ),
     iRunningNewClientCnt ( 0 ),
-    iNumMixerPanelRows   ( 1 ),
+    iNumMixerPanelRows   ( 1 ), // pSettings->iNumMixerPanelRows is not yet available
```
pljones, line 836: Amended.

src/clientdlg.cpp
```diff
@@ -164,6 +164,7 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     // set the settings pointer to the mixer board (must be done early)
     MainMixerBoard->SetSettingsPointer ( pSettings );
+    MainMixerBoard->SetNumMixerPanelRows ( pSettings->iNumMixerPanelRows );
```
pljones, line 167: Done here instead.

src/audiomixerboard.cpp
```diff
@@ -1088,8 +1084,8 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
             // per definition: the fader order is colum-first/row-second (note that
             // the value in iNumFadersFirstRow defines how many rows we will get)
```
hoffie, line 1085:
```suggestion
            // the value in iNumFadersFirstRows defines how many rows we will get)
```

hoffie (commented):
Ok, starts up properly now and seems to work as expected.

It might be due to my screen size, but with 3 rows I already get vertical scroll bars. I guess other people may use different screen aspect ratios or rotations.

I'm wondering where the setting is best placed. Yes, it's certainly an *Advanced* option, but on the other hand it's also UI-related where there is a specific area in the *User* tab. I might have rather expected it in the latter place. Maybe let's see what others think.

pljones:
The "User" tab is details _about_ the user, not about the user interface, so it shouldn't go there.

pljones:
Hm.
```
Build assets for MacOS (codeQL)
GitHub Actions has encountered an internal error when running your job.
```

pljones:
Rebasing picks up some changes to the autobuild...

softins:
> Rebasing picks up some changes to the autobuild...
> 
> ...and that appears to be what's breaking the build. The `make-the-basic-settings-less-tall` was working and now fails, having rebased.

Hmm, I merged #1500 21 hours ago, but that had passed its checks before. I see the Mac builds are currently showing as "Cancelled"

pljones:
And that's just as odd, because they were not... Hmmm...

softins:
> And that's just as odd, because they were not... Hmmm... (And yes, it certainly worked at least once after your change went in.)

And in fact, my PR only changed the CodeQL build to match what was already in Artifacts. And now they are both failing. Perhaps it's some temporary problem in the build system?

pljones:
Possibly - I saw an Android build fail earlier and that's working again now.  I'll give it a kick...

pljones:
Ah, the MacOS asset build now ran okay.

hoffie:
> The "User" tab is details _about_ the user, not about the user interface, so it shouldn't go there.

Well, but other options which are not about the user but rather about the user interface are also placed in this tab (like Fancy vs. Normal mode).

hoffie (approved):

softins:
Looking at the `git log --graph` for this PR, it appears to incorporate the changes from #1542 as well. Is that correct? If so, then I guess it's important for 1542 to get merged first?

softins (approved):
Built and run ok on my Pi. I already merged #1542

PR #1551, opened 2021-04-22
Increase number of groups

pljones:
Feedback on choice of colours would be appreciated.  I needed to increase the contrast so that it was easier to see the different styles.

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -63,13 +63,15 @@ CChannelFader::CChannelFader ( QWidget* pNW ) :
 
     // define the popup menu for the group checkbox
     pGroupPopupMenu = new QMenu ( "", pcbGroup );
-    pGroupPopupMenu->addAction ( tr ( "&No grouping" ), this, SLOT ( OnGroupMenuGrpNone() ) );
-    pGroupPopupMenu->addAction ( tr ( "Assign to group" ) + " &1", this, SLOT ( OnGroupMenuGrp1() ) );
-    pGroupPopupMenu->addAction ( tr ( "Assign to group" ) + " &2", this, SLOT ( OnGroupMenuGrp2() ) );
-    pGroupPopupMenu->addAction ( tr ( "Assign to group" ) + " &3", this, SLOT ( OnGroupMenuGrp3() ) );
-    pGroupPopupMenu->addAction ( tr ( "Assign to group" ) + " &4", this, SLOT ( OnGroupMenuGrp4() ) );
-#if ( MAX_NUM_FADER_GROUPS != 4 )
-# error "MAX_NUM_FADER_GROUPS must be set to 4, see implementation in CChannelFader()"
+    pGroupPopupMenu->addAction ( tr ( "&No grouping" ),
+        this, [=] { OnGroupMenuGrp ( INVALID_INDEX ); } );
+    for ( int iGrp = 0 ; iGrp < MAX_NUM_FADER_GROUPS ; iGrp++ )
+    {
+        pGroupPopupMenu->addAction ( tr ( "Assign to group" ) + ( QString ( " &%1" ) .arg( iGrp + 1 ) ),
+            this, [=] { OnGroupMenuGrp ( iGrp ); } );
+    }
+#if ( MAX_NUM_FADER_GROUPS != 8 )
+# error "MAX_NUM_FADER_GROUPS must be set to 8, see implementation in CChannelFader()"
 #endif
 
     // setup channel level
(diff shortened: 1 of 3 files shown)


src/audiomixerboard.cpp
```diff
@@ -63,13 +63,15 @@ CChannelFader::CChannelFader ( QWidget* pNW ) :
...
-    pGroupPopupMenu->addAction ( tr ( "Assign to group" ) + " &2", this, SLOT ( OnGroupMenuGrp2() ) );
-    pGroupPopupMenu->addAction ( tr ( "Assign to group" ) + " &3", this, SLOT ( OnGroupMenuGrp3() ) );
-    pGroupPopupMenu->addAction ( tr ( "Assign to group" ) + " &4", this, SLOT ( OnGroupMenuGrp4() ) );
-#if ( MAX_NUM_FADER_GROUPS != 4 )
-# error "MAX_NUM_FADER_GROUPS must be set to 4, see implementation in CChannelFader()"
+    pGroupPopupMenu->addAction ( tr ( "&No grouping" ),
+        this, [=] { OnGroupMenuGrp ( INVALID_INDEX ); } );
+    for ( int iGrp = 0 ; iGrp < MAX_NUM_FADER_GROUPS ; iGrp++ )
+    {
+        pGroupPopupMenu->addAction ( tr ( "Assign to group" ) + ( QString ( " &%1" ) .arg( iGrp + 1 ) ),
```
pljones, line 70: Just a note.  The new implementation is pretty flexible apart from here.  The `" &%`"` would need to become a bit more horrible and only have the `&` if the `iGrp` was 1-9, as above that it would duplicate itself, so eight felt like a better limit for now.

`QString ( ( iGrp < 8 ? " &"%1 : "%1"  ) .arg( iGrp + 1 )`

src/audiomixerboard.cpp
```diff
@@ -291,36 +293,67 @@ void CChannelFader::SetDisplayPans ( const bool eNDP )
...
+        {
+        case 0:
+            strBorderStyle = "solid";
+            break;
+
+        case 1:
+            strBorderStyle = "dashed";
+            break;
+
+        case 2:
```
pljones, line 340: 2 and 3 currently not used.  These allow for up to 16 groups.  Beyond that, good contrasting styling becomes more problematic.  And probably harder to cope with mentally.  "Five plus or minus two" applies to many things - so eight groups should be more than enough.

src/audiomixerboard.h
```diff
@@ -135,11 +135,7 @@ public slots:
     void OnMuteStateChanged ( int value );
     void OnGroupStateChanged ( int );
 
-    void OnGroupMenuGrpNone() { SetGroupID ( INVALID_INDEX ); }
-    void OnGroupMenuGrp1()    { SetGroupID ( 0 ); }
-    void OnGroupMenuGrp2()    { SetGroupID ( 1 ); }
-    void OnGroupMenuGrp3()    { SetGroupID ( 2 ); }
-    void OnGroupMenuGrp4()    { SetGroupID ( 3 ); }
+    void OnGroupMenuGrp ( int iGrp ) { SetGroupID ( iGrp ); }
```
pljones, line 138: I'm not sure this is needed.  All it's doing is handing off the caught signal to the method.  Nothing outside the class uses this slot, so the `addAction` on the menu could directly call `SetGroupID`.

src/audiomixerboard.cpp
```diff
@@ -63,13 +63,15 @@ CChannelFader::CChannelFader ( QWidget* pNW ) :
...
-# error "MAX_NUM_FADER_GROUPS must be set to 4, see implementation in CChannelFader()"
+    pGroupPopupMenu->addAction ( tr ( "&No grouping" ),
+        this, [=] { OnGroupMenuGrp ( INVALID_INDEX ); } );
+    for ( int iGrp = 0 ; iGrp < MAX_NUM_FADER_GROUPS ; iGrp++ )
+    {
+        pGroupPopupMenu->addAction ( tr ( "Assign to group" ) + ( QString ( " &%1" ) .arg( iGrp + 1 ) ),
+            this, [=] { OnGroupMenuGrp ( iGrp ); } );
+    }
+#if ( MAX_NUM_FADER_GROUPS != 8 )
+# error "MAX_NUM_FADER_GROUPS must be set to 8, see implementation in CChannelFader()"
```
softins, line 74: Is this necessary? The implementation looks like it works with any value of `MAX_NUM_FADER_GROUPS` up to 8. So suggest to only reject values > 8.

softins (approved):
Apart from my one minor comment, I've built and tried this, and it looks fine.

softins:
Oh, and the colours look fine to me, but I'm not colour-blind.

src/audiomixerboard.cpp
pljones: Well, it was `!=` before, I just changed the number :).  But yes, it could say `>`.

hoffie (approved):
Looks good to me, both code-wise and visually. Colors look good in normal mode, too.

Grouping colors are a bit hard to distinguish in fancy mode, but this is independent of this PR. Maybe a thicker border would help, but I'm not sure what other layout consequences this could have. Maybe something for future consideration.

pljones:
I went for something easy to read in compact mode, as I think that's most likely to be in use with large groups where the grouping feature comes into play.  I didn't actually check fancy mode.  It's possible, of course, to use different colours dependent upon mode.

PR #1552, opened 2021-04-22
Favorites functionality

dcorson-ticino-com:
PR Favorites

This PR implements a Favorite functionality as a tab in the Connect window.
The idea being able to use favorites even when directories are down, the server names, IP, max users and genre are saved.

The favorites list now has 16 members, the most recently used is at the top, the others pushed down.  Not used servers will fall off the end of the list.
To add a server to the list, connect to the server using the connect window Directory tab (unchanged behaviour) and then click the Add to Favorites button on the clientdlg.  This way any server can be added, be it private or public, using IP address or URL.  Of course if the connection does not use the services of a directory there is no max users and no genre.  Adding a server that is already on the list again does no harm.  The Add to Favorites button is disabled when not connected to a server.

The last used server in the favorites list is automagically selected when the list is built so that if you use only one server all you need to do is click on connect. 

The changes are almost all in connectdlg.  Settings adds the saving of the new values.  Clientdlg has the new Add to Favorites button.

There is a problem with this code that I have not been able to solve. Certain servers sometimes do not respond to the ping requests.  It is usually the same servers not responding and very strangely if you go to the Directory tab and directory of the server, wait a bit, and then go back to the Favorites the server responds.  When the server is not responding it is also not possible to connect to it from the favorites tab, from the servers tab, however, works.
I need help finding this, for the moment I have no idea what is happening.

FAV01

Here we see one server that is not responding, frosch
FAV03

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -96,15 +96,6 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     butConnect->setAccessibleName ( tr ( "Connect and disconnect toggle button" ) );
 
-
-    // add to favorites button
-    butAddtoFavorites->setWhatsThis ( "<b>" + tr ( "Add to Favorites Button" ) + ":</b> " +
-        tr ( "Click this button to add the currently connected server to the Favorites list"
-        "of the Connect window.  This button is disabled when not connected to a server." ) );
-
-    butAddtoFavorites->setAccessibleName (
-        tr ( "Add to Favorites button" ) );
-
     // reverberation level
     QString strAudReverb = "<b>" + tr ( "Reverb effect" ) + ":</b> " +
                            tr ( "Reverb can be applied to one local mono audio channel or to both "
@@ -348,12 +339,20 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     pEditMenu->addAction ( tr ( "Auto-Adjust all &Faders" ), this, SLOT ( OnAutoAdjustAllFaderLevels() ), QKeySequence ( Qt::CTRL + Qt::Key_F ) );
 
+    // Add to Favorites "button" -----------------------------------------------
+    paFAVAction = new QAction ( tr ( "Add to &Favorites" ) );
+    QObject::connect ( paFAVAction, SIGNAL( triggered() ), &ConnectDlg, SLOT( OnAddtoFavorites() ) );
+    paFAVAction->setEnabled ( false );
+
     // Main menu bar -----------------------------------------------------------
     QMenuBar* pMenu = new QMenuBar ( this );
 
     pMenu->addMenu ( pFileMenu );
     pMenu->addMenu ( pViewMenu );
     pMenu->addMenu ( pEditMenu );
+    pMenu->addMenu ( "|" );
+    pMenu->addAction ( paFAVAction );
+    pMenu->addMenu ( "|" );
     pMenu->addMenu ( new CHelpMenu ( true, this ) );
 
     // Now tell the layout about the menu
@@ -397,9 +396,6 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // Connections -------------------------------------------------------------
     // push buttons
     QObject::connect ( butConnect, &QPushButton::clicked, this, &CClientDlg::OnConnectDisconBut );
-
-    QObject::connect ( butAddtoFavorites, &QPushButton::clicked,
-        &ConnectDlg, &CConnectDlg::OnAddtoFavorites );
 
     // check boxes
     QObject::connect ( chbSettings, &QCheckBox::stateChanged, this, &CClientDlg::OnSettingsStateChanged );
@@ -509,12 +505,8 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
                        this,
                        &CClientDlg::OnCreateCLServerListReqConnClientsListMes );
 
-<<<<<<< HEAD
-    QObject::connect ( &ConnectDlg, &CConnectDlg::accepted, this, &CClientDlg::OnConnectDlgAccepted );
-=======
     // disable Add to Favorites button by default
-    butAddtoFavorites->setEnabled(false);
->>>>>>> 2eaf6151 (Favorites functionality)
+    paFAVAction->setEnabled( false );
 
     // Initializations which have to be done after the signals are connected ---
     // start timer for status bar
@@ -713,14 +705,10 @@ void CClientDlg::OnConnectDisconBut()
     if ( pClient->IsRunning() )
     {
         Disconnect();
-<<<<<<< HEAD
-        SetMixerBoardDeco ( RS_UNDEFINED, pClient->GetGUIDesign() );
-=======
         SetMixerBoardDeco( RS_UNDEFINED, pClient->GetGUIDesign() );
 
         // disable Add to Favorites button
-        butAddtoFavorites->setEnabled(false);
->>>>>>> 2eaf6151 (Favorites functionality)
+        paFAVAction->setEnabled( false );
     }
     else
     {
@@ -1191,7 +1179,7 @@ void CClientDlg::Connect ( const QString& strSelectedAddress, const QString& str
         MainMixerBoard->SetServerName ( strMixerBoardLabel );
 
         // enable Add to Favorites button
-        butAddtoFavorites->setEnabled(true);
+        paFAVAction->setEnabled( true );
 
         // start timer for level meter bar and ping time measurement
         TimerSigMet.start ( LEVELMETER_UPDATE_TIME_MS );
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -339,10 +339,17 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     pEditMenu->addAction ( tr ( "Auto-Adjust all &Faders" ), this, SLOT ( OnAutoAdjustAllFaderLevels() ), QKeySequence ( Qt::CTRL + Qt::Key_F ) );
 
-    // Add to Favorites "button" -----------------------------------------------
-    paFAVAction = new QAction ( tr ( "Add to &Favorites" ) );
+    // Favorites menu  --------------------------------------------------------------
+    paFAVAction = new QAction ( tr ( "&Add to Favorites" ) );
     QObject::connect ( paFAVAction, SIGNAL ( triggered() ), &ConnectDlg, SLOT ( OnAddtoFavorites() ) );
     paFAVAction->setEnabled ( false );
+    paFAVAction->setShortcut ( Qt::CTRL + Qt::Key_A );
+
+    QMenu* pFavoritesMenu = new QMenu ( tr ( "Fav&orites" ), this );
+
+    pFavoritesMenu->addAction ( paFAVAction );
+
+    pFavoritesMenu->addAction ( tr ( "Sho&w Favorites..." ), this, SLOT ( OnOpenFavoritesConnectionDialog() ), QKeySequence ( Qt::CTRL + Qt::Key_W ) );
 
     // Main menu bar -----------------------------------------------------------
     QMenuBar* pMenu = new QMenuBar ( this );
@@ -350,9 +357,7 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     pMenu->addMenu ( pFileMenu );
     pMenu->addMenu ( pViewMenu );
     pMenu->addMenu ( pEditMenu );
-    pMenu->addMenu ( "|" );
-    pMenu->addAction ( paFAVAction );
-    pMenu->addMenu ( "|" );
+    pMenu->addMenu ( pFavoritesMenu );
     pMenu->addMenu ( new CHelpMenu ( true, this ) );
 
     // Now tell the layout about the menu
@@ -911,6 +916,14 @@ void CClientDlg::SetMyWindowTitle ( const int iNumClients )
     }
 #    endif
 #endif
+}
+
+void CClientDlg::ShowFavoritesConnectionDialog()
+{
+    // set to Favorites Tab
+    pSettings->bFavoriteWasShownConnect = true;
+    // show connect dialog
+    ShowConnectionSetupDialog();
 }
 
 void CClientDlg::ShowConnectionSetupDialog()
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -96,6 +96,15 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     butConnect->setAccessibleName ( tr ( "Connect and disconnect toggle button" ) );
 
+
+    // add to favorites button
+    butAddtoFavorites->setWhatsThis ( "<b>" + tr ( "Add to Favorites Button" ) + ":</b> " +
+        tr ( "Click this button to add the currently connected server to the Favorites list"
+        "of the Connect window.  This button is disabled when not connected to a server." ) );
+
+    butAddtoFavorites->setAccessibleName (
+        tr ( "Add to Favorites button" ) );
+
     // reverberation level
     QString strAudReverb = "<b>" + tr ( "Reverb effect" ) + ":</b> " +
                            tr ( "Reverb can be applied to one local mono audio channel or to both "
@@ -389,6 +398,9 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // push buttons
     QObject::connect ( butConnect, &QPushButton::clicked, this, &CClientDlg::OnConnectDisconBut );
 
+    QObject::connect ( butAddtoFavorites, &QPushButton::clicked,
+        &ConnectDlg, &CConnectDlg::OnAddtoFavorites );
+
     // check boxes
     QObject::connect ( chbSettings, &QCheckBox::stateChanged, this, &CClientDlg::OnSettingsStateChanged );
 
@@ -497,7 +509,12 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
                        this,
                        &CClientDlg::OnCreateCLServerListReqConnClientsListMes );
 
+<<<<<<< HEAD
     QObject::connect ( &ConnectDlg, &CConnectDlg::accepted, this, &CClientDlg::OnConnectDlgAccepted );
+=======
+    // disable Add to Favorites button by default
+    butAddtoFavorites->setEnabled(false);
+>>>>>>> 2eaf6151 (Favorites functionality)
 
     // Initializations which have to be done after the signals are connected ---
     // start timer for status bar
@@ -696,7 +713,14 @@ void CClientDlg::OnConnectDisconBut()
     if ( pClient->IsRunning() )
     {
         Disconnect();
+<<<<<<< HEAD
         SetMixerBoardDeco ( RS_UNDEFINED, pClient->GetGUIDesign() );
+=======
+        SetMixerBoardDeco( RS_UNDEFINED, pClient->GetGUIDesign() );
+
+        // disable Add to Favorites button
+        butAddtoFavorites->setEnabled(false);
+>>>>>>> 2eaf6151 (Favorites functionality)
     }
     else
     {
@@ -1165,6 +1189,9 @@ void CClientDlg::Connect ( const QString& strSelectedAddress, const QString& str
 
         // set server name in audio mixer group box title
         MainMixerBoard->SetServerName ( strMixerBoardLabel );
+
+        // enable Add to Favorites button
+        butAddtoFavorites->setEnabled(true);
 
         // start timer for level meter bar and ping time measurement
         TimerSigMet.start ( LEVELMETER_UPDATE_TIME_MS );
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -505,6 +505,9 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
                        this,
                        &CClientDlg::OnCreateCLServerListReqConnClientsListMes );
 
+    QObject::connect ( &ConnectDlg, &CConnectDlg::accepted,
+        this, &CClientDlg::OnConnectDlgAccepted );
+
     // disable Add to Favorites button by default
     paFAVAction->setEnabled( false );
 
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -349,7 +349,10 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     pFavoritesMenu->addAction ( paFAVAction );
 
-    pFavoritesMenu->addAction ( tr ( "Sho&w Favorites..." ), this, SLOT ( OnOpenFavoritesConnectionDialog() ), QKeySequence ( Qt::CTRL + Qt::Key_W ) );
+    pFavoritesMenu->addAction ( tr ( "Sho&w Favorites..." ),
+                                this,
+                                SLOT ( OnOpenFavoritesConnectionDialog() ),
+                                QKeySequence ( Qt::CTRL + Qt::Key_W ) );
 
     // Main menu bar -----------------------------------------------------------
     QMenuBar* pMenu = new QMenuBar ( this );
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -341,7 +341,7 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     // Add to Favorites "button" -----------------------------------------------
     paFAVAction = new QAction ( tr ( "Add to &Favorites" ) );
-    QObject::connect ( paFAVAction, SIGNAL( triggered() ), &ConnectDlg, SLOT( OnAddtoFavorites() ) );
+    QObject::connect ( paFAVAction, SIGNAL ( triggered() ), &ConnectDlg, SLOT ( OnAddtoFavorites() ) );
     paFAVAction->setEnabled ( false );
 
     // Main menu bar -----------------------------------------------------------
@@ -505,11 +505,10 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
                        this,
                        &CClientDlg::OnCreateCLServerListReqConnClientsListMes );
 
-    QObject::connect ( &ConnectDlg, &CConnectDlg::accepted,
-        this, &CClientDlg::OnConnectDlgAccepted );
+    QObject::connect ( &ConnectDlg, &CConnectDlg::accepted, this, &CClientDlg::OnConnectDlgAccepted );
 
     // disable Add to Favorites button by default
-    paFAVAction->setEnabled( false );
+    paFAVAction->setEnabled ( false );
 
     // Initializations which have to be done after the signals are connected ---
     // start timer for status bar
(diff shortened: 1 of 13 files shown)


ngocdh:
Thanks @dcorson-ticino-com. I have a request though: for mobile devices, the "Add to Favorites" button would make it worse for iOS, considering its already bad situation:

47A89825-4655-4124-B4C9-8C5D5C926B44

Would you be able to have a look at iOS and try to fix it? If not possible, maybe "Add to Favorites" could be implemented as a Menu item?

hoffie:
Thanks for working on this -- I think there have been some request for that.
1. Can you link the relevant issue(s)?
2. Build is failing (both locally for me and on CI)
3. I also think that the button might be better placed somewhere else. This is a feature which is not needed by every Jamulus user and for those who do want it it will probably only be needed a few times. So unless we come up with some better idea, it might be the best option to move it to the menu?

dcorson-ticino-com:
Hi ngocdh,
What kind of device is that?
It looks to me like our present UI is completely doomed on that.  
I do not really understand because I have seen Jamulus on a small Win10 tablet and it was no problem at all.
Is there no way to shrink the text etc. but keeping it readable on that device?

ngocdh:
@dcorson-ticino-com 
> What kind of device is that?
It's iPhone 12. It seems to me that the app was rendered in low DPI mode in iOS. 
I have to note that in my Android phone (Google pixel), the Mixer board is rendered perfectly. I couldn't find the reason for this difference.

> Is there no way to shrink the text etc. but keeping it readable on that device?
I have been trying to modify properties in the UI using QT Creator, but to no avail.

dcorson-ticino-com:
The Windows build appears to be broken.  
I see timeout errors, but have no idea why there should be timeouts when compiling and building.

dcorson-ticino-com:
Looking further today I see that the problem I mentioned has 2 manifestations.

- Servers with dynamic addresses.  When the IP:port changes I no longer find it.  When I select them on the directories tab and do an Add to Favorites I get a new entry as they have a new IP.  There are then 2 entries for the same server name and genre.  Doesn't harm, but looks stupid.  I guess I could check and remove the older entry if there are 2 entries with the same name and genre, but I don't see how this could be automated to check for new IP.  (if a server doesn't respond make a request at it's Directory to check and see if it has a new IP???)

- Servers that for some reason which I don't know are not responding to pings, although IP:port are correct.  I can also not connect to these servers from the Favorites tab.  From the Directories tab they are, however, working.  A complete mystery for me at the moment.

Any and all ideas how to get this working are welcome :)

BTW: I pushed a new version with the "Add to Favorites" action in the menu.  Doing the same for "Settings" we could eliminate another button on the left too.  I also hesitate to suggest the same for "My Profile".  Recently the biggest problems I have had with new users was getting their profile set.  If it is not immediately visible, it does not exist.

pljones:
You're building something that needs to work the same way a directory server does - basically, you're providing a "favourites directory server".  However, you've got the problem that, unlike an actual directory server, hosts aren't updating their details with you every 15 minutes, so your information is doomed to go stale (which is why servers are required to re-register every fifteen minutes).

I can only think of one solution right now.  First, never store an IP address.  Store the genre directory server host name (and port), plus the favourite server name as held in the directory server list (and port).  When a user selects a favourite, basically redirect to the directory server to do the look up.

I was hoping this would be a new "Favourites"/"Bookmarks" menu, with "Add" and "Manage" options, plus an "Add" on the existing directory server list.  That would avoid opening the directory list at all.

gene96817:
I agree with @pljones. Storing server names with IP:port will end up with some stale data. The directory server hides the fact that IP:port numbers change. The best you can do is store a list of Genre+server names. Then when the Genre+server is selected, you have to get the latest IP:port from the directory server. If the user always uses the same Genre, then the servers of this Genre will already be in their client. If the desired favorite is in a different Genre, you will need a mechanism to get the latest from the Directory server.

dcorson-ticino-com:
Without having thought about this for a long time, I think this should be possible without any problems.  Private servers remain a special case.

@pljones I am more interested in why you think a new menu item with the Favorites would be better than this tabbed connect window.  What would "manage" need other than delete?   How would one add private servers?  I could call up the present window with the combobox, but would it be simpler for the user and easier to find?
Why would opening a favorites menu be better than opening the server list (if I understand you correctly) ?
I can imagine that it might be good to be able to select a server for the favorites without connecting to it, but don't see any real advantage.  Lots of questions.

My simplest case would be to 1) connect to a server once, 2) Add to Favorites, 3) switch to Favorites and then to connect simply click connect for the rest of time.  How could that be made simpler?

softins:
This functionality seems like it bypasses the magic port-opening provided by the Directory Server when a client requests a server list (the Directory Server tells each server in its list to send an empty message to the requesting client). If a particular server is relying on this, rather than having its inbound port specifically open and forwarded, and a client then tries to connect to that server directly from its "Favourites" list, the connection will fail.

dcorson-ticino-com:
Thanks Tony, then the solution is clear.

pljones:
> switch to Favorites and then to connect simply click connect for the rest of time. How could that be made simpler?

Because the connect window doesn't stay open on connect, unlike the mixer window.  So a menu item on the mixer window is a shorter shortcut than having to open a window, switch tabs, then double-click the item.  It's open menu, select, done.  So lots and lots simple.

I'm also favouring making the Jamulus menu more in keeping with "standard" menus.  Having to go somewhere other than the main window to find favourites feels odd to me.  Why would I want to make favourites harder to find than having them directly on the main window?  The point generally is that they're "just there", not off somewhere else I have to navigate to first.

Like in a browser, there _is_ a "Bookmarks" window, but most people just use the menu -- because it's quick and easier, takes up no screen estate when not in use and gives you an instant response.

----

> What would "manage" need other than delete? 

A user might want to adjust the order.  Or mark an entry "never forget".  Maybe not in this initial version, though, indeed.  (Basically, I saw "manage" as opening your existing new tab.)

dcorson-ticino-com:
> It's open menu, select, done. So lots and lots simpler.

So you are suggesting that there is only one Favorite ?
Otherwise I fail to see how putting the Favorites in a menu opening a window makes less clicks than having it in the connect window.
Of course the connect window remembers its last state so if you always use favorites it opens with the favorites tab showing and the last used server selected.  Click connect and away you go.

I see you want to expand the use of the menus.  That is a possibility to add features and reduce clutter and until recently would have been my first thought.  But in the last 2 months I have had an onslaught of people who apparently don't know what menus are.  They are unable to set their profile because it is hidden in a menu.  Even with hand-holding by telephone it was a problem. I guess they only have a smartphone and even thought things are even more hidden there they are unable to use something that has an interface different from their iphone.  So my approach has been to use the space of the menu bar for "buttons", but reduce the menus as much as possible.  Of course this is all debatable, I don't claim to have the definitive answer.

pljones:
> So you are suggesting that there is only one Favorite ?

Opening a menu doesn't reveal only one thing to select.

dcorson-ticino-com:
Keeping original structure so that private servers and port forwarded
servers will work without any Directory server intervention.  Thus they
can be used when the Directory servers are striking (which continues to
happen, not alot, but all too often).

If no ping is received from a server a server list request is made to
it's Directory server (max. 1 request per Directory server).  Then pings
are requested again.
When a server list is seen from a Directory server the stored IP and max
users are checked and updated in the Favorites if necessary.
Requests are send 1 per OnTimerPing so as to keep down network traffic.
Servers are marked "offline" if after a server list request and
following ping requests there is still no ping received.
The entire sequence can be restarted by reopening the Favorites tab or
reopening the window.

The sort order is last used or added server at the top of the list
pushing down the previous servers.

Unsolved problem: if there are 2 servers with the same name on the same
Directory server the IP update may update the wrong one.

FAV05

pljones:
> Unsolved problem: if there are 2 servers with the same name on the same Directory server the IP update may update the wrong one.

Mmm...  The directory is unique on ( ip-address, port ) currently.  At worst, it can change every time a server refreshes - having gone stale at some point between.  I've seen this from one or two servers that I assume were having serious problems with their connection or set up.  It doesn't happen often.

So yes, having to refresh the favourites by name within directory could be an issue (as the directory could also have multiple stale entries for a host).  I don't think there's much you can do other than hope people avoid clashing names and problems don't occur for servers people are likely to have in favourites.  I'd say it wasn't particularly likely and not a blocking issue here.

dcorson-ticino-com:
Added Sorting by Directory Server to Favorites

Last used sort is unchanged, the last added or used server is in the top
row pushing down, top row selected when window opened.

For Directory sort the servers are sorted according to directory, within
the directory last used.  The default row selecton is the last row used.

Favorites Sort and Favorites last row added to settings.

src/clientsettingsdlg.cpp
```diff
@@ -36,8 +36,8 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     // IOS needs menu to close
     QMenuBar* pMenu  = new QMenuBar ( this );
     QAction*  action = pMenu->addAction ( tr ( "&Close" ) );
+    QMenuBar* pMenu = new QMenuBar ( this );
```
pljones, line 37: This line appears to duplicate line 37.

src/clientdlg.cpp
```diff
@@ -96,6 +96,15 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     butConnect->setAccessibleName ( tr ( "Connect and disconnect toggle button" ) );
 
+
+    // add to favorites button
+    butAddtoFavorites->setWhatsThis ( "<b>" + tr ( "Add to Favorites Button" ) + ":</b> " +
```
pljones: `Button` should be lower case "B".

src/clientdlg.cpp
```diff
@@ -497,7 +509,12 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
                        this,
                        &CClientDlg::OnCreateCLServerListReqConnClientsListMes );
 
+<<<<<<< HEAD
```
pljones: This conflict hasn't been resolved.

src/clientdlg.h
```diff
@@ -84,38 +84,42 @@ class CClientDlg : public CBaseDlg, private Ui_CClientDlgBase
...
-    QTimer         TimerStatus;
-    QTimer         TimerPing;
-    QTimer         TimerCheckAudioDeviceOk;
-    QTimer         TimerDetectFeedback;
-
-    virtual void closeEvent ( QCloseEvent* Event );
-    virtual void dragEnterEvent ( QDragEnterEvent* Event ) { ManageDragNDrop ( Event, true ); }
-    virtual void dropEvent ( QDropEvent* Event ) { ManageDragNDrop ( Event, false ); }
-    void         UpdateDisplay();
+    void               SetGUIDesign ( const EGUIDesign eNewDesign );
```
pljones, line 87: Appears to have lost the clang-format changes.

src/connectdlg.cpp
```diff
@@ -25,32 +25,48 @@
 #include "connectdlg.h"
```
pljones, line 25: Appears to have lost the clang-format changes.

src/connectdlg.cpp
```diff
@@ -678,19 +855,88 @@ void CConnectDlg::OnConnectClicked()
...
+
+    // if Favorites, check if servers aren't responding, send server request if so to open ports
+    if( pSettings->bFavoriteWasShownConnect == CONTAB_FAVORITES )
     {
-        CHostAddress CurServerAddress;
+        // iWasSent, bitfield to remember which directorys have been queried
+        int iWasSent = 0;
+
+        // pList->topLevelItem ( iIdx )->data ( 1 ) stores sequencer
+        //  val == 0, first timeout -> skip, check if no directory (privat server), set val to 4 to do no request
```
pljones, line 872: `privat` should be "private" throughout.

pljones:
We need to ensure all new code has clang-format applied.

This will need to be done be committers or either the CI will be rejecting the commit or we risk letting the CI make code changes to reformat as it wants it, which will mean it doesn't match what you wrote.  So it's more comfortable to fit in with what the formatter is going to do...

dcorson-ticino-com:
How do I apply the clang format?
Is there a definition somewhere of what it expects we should be writing?
What I saw of the new formatting seemed at times to not be very consistent, the Qt connects, for example sometimes on 3 lines, sometimes on 1, for no apparent reason.

dcorson-ticino-com:
I am worried.
Those lines in the diffs do not correspond with my source code at all.
For the first on, for example, I have this:

36    // IOS needs menu to close
37    QMenuBar* pMenu = new QMenuBar ( this );
38    QAction*  action = pMenu->addAction ( tr ( "&Close" ) );
39    connect ( action, SIGNAL ( triggered() ), this, SLOT ( close() ) );
40    // Now tell the layout about the menu
41    layout()->setMenuBar ( pMenu );


How can that be after a rebase and push ?

src/clientdlg.cpp
dcorson-ticino-com: This code should not be there. There is no longer a butAddtoFavorites.
What does the "Outdated" at the top mean ?

pljones: It means a later commit than my comment changed it.

dcorson-ticino-com:
Directory sort
FAV06

Last used sort
FAV07

softins:
I've pushed an update to restore some items that were accidentally lost during the recent rebases, and have built it successfully on my Pi.

I wasn't keen on having a top-level menu for "Add to Favorites" instead of a button. And I particularly didn't like the dummy menu items consisting just of `|`

dcorson-ticino-com:
Thanks Tony.
I am not keen about the menu thing either, but I had people complaining about the screen real estate used by a button and made that as a suggestion to use otherwise wasted real estate.
As you can see in the first picture at the top of the discussion here, I had it originally as a button under Connect. Maybe to prominent for something that won't be used often, but it seemed logical there.
What do you think ?

pljones:
I'd rather have a "Favourites" menu, just as a normal menu.  Then Add can have a shortcut.  It also makes the feature more extensible - e.g. with appropriate parts of  #1728.

dcorson-ticino-com:
Note that for WorldJam users the use is (as I imagine it):
- one time, connect to the servers you will be using and add them to the favorites, menu "Favorites>Add to Favorites".
- to connect either click Connect, Favorites tab or in the menu "Favorites>Show Favorites", then set sort to by Directory.  The WorldJam servers are now all together.
- Double click on the first server you will be joining.
- then to change to the next server do "Favorites>Show  Servers" and double click on the next server you want to join.  You do not need to explicitly disconnect and then connect to the next server.

pljones:
The idea is to have a "studio running order".  So you can set the order of connections and simply (one click) move to the next entry.

dcorson-ticino-com:
I guess then that the solution for that would be to add management to allow making an arbitrary list of the favorite servers with delete and reordering to make a list could be made and run down one at a time.
Personally I would put all the management functions including the click to next entry on the RMB.
As is now the list would be stored in the ini, no need for save/load.
The management of server referencing etc. would remain unchanged.
I will let the interested person do this.
I think the favorites already do all that is necessary and that the simple last used sort is for all non-WorldJam users the right way and should not be removed when WorldJam features are added.

pljones:
> As is now the list would be stored in the ini, no need for save/load.

The need for save/load is that you don't want to have to _necessarily_ have separate profiles for normal use and WorldJam use.  Being able to switch favourites lists on the fly is a very useful feature.

geotechfirst:
Hi! I would like to help. What is left to do? @softins @pljones @dcorson-ticino-com

ann0see:
I think it would be great to start a discussion on this topic first (saving servers as favourites). To me it seems to be a feature which is wanted but not simple I think, we should start a discussion on the broader topic first and not focus on the actual implementation yet. Maybe @dcorson-ticino-com and @geotechfirst can agree on a simple solution?

We could maybe add a "star" or pin functionality to a server on the public list. This would then move the pinned server(s) based on their genre lists at the top or add new servers based on their IP.

dcorson-ticino-com:
> We could maybe add a "star" or pin functionality to a server on the public list. This would then move the pinned server(s) based on their genre lists at the top or add new servers based on their IP.

For me the initiating factor for the Favorites list was the (ongoing) unreliability of the Directory servers.  This may be not obvious and well hidden, but for most servers the Favorites list allows connecting to public servers even if the Directory is down.  I can not do that functionality by simply sorting differently (i.e. pinning to the top) the list of servers from the Directory.

jp8:
> Any and all ideas how to get this working are welcome :)

It seems to me that if you don't want to discard favourites whose ip address has changed, then the jamulus servers will need to generate a keypair, and publish the public key to the directory server.

The favourites list would store, for each entry:
- ip
- dns name (initially empty, but can be entered manually by the user)
- server's directory server
- server's public key
- server's friendly name

The favourites list would check each entry against it's stored directory server, searching by public key, to see if the ip (or indeed the friendly name) have changed.  If it has updated info from the directory server, it updates the favourites entry.

The directory server would make sure it is dealing with the correct jamulus server by generating a random secret, encrypting it with the jamulus server's public key, and challenging the jamulus server to decrypt it.

If the directory server has no information about the jamulus server, the favourites list would mark it in grey as 'currently unavailable' with the user having the option to delete it from the list.

Private servers could also be stored in the favourites list, with similar identify verification done directly between jamulus client and jamulus server.  To help with changing ip addresses, the favourites list could give the user the option to provide a dns name which would be used in preference to the ip address.

dcorson-ticino-com:
>It seems to me that if you don't want to discard favourites whose ip address has changed

Please reference the latest push.  These problems have all been solved.  Note that servers register with their directory every 15 minutes so that is the longest time that possibly invalid info could be seen.

The sort by directory is for me unnecessary fluff that should be removed again.  It was just a response to the WorldJam questions.

jp8:
>  These problems have all been solved
Oh sorry as usual I am quoting the wrong thing.

>  Unsolved problem: if there are 2 servers with the same name on the same Directory server the IP update may update the wrong one.
That is the problem I was trying to think about.  Did you already find a way around it ?

dcorson-ticino-com:
>Oh sorry as usual I am quoting the wrong thing.

No problem, just makes it hard to understand :)

> Unsolved problem: if there are 2 servers with the same name on the same Directory server the IP update may update the wrong one.
>That is the problem I was trying to think about. Did you already find a way around it ?

No I haven't, but I think this is a negligible problem as: 1) I don't remember ever seeing two servers with the same name 2) I have only seen once a server changing their IP address.  Having both happening together will be a miracle.
I would suggest having directories reject registering a second server with an existing name, but am not sure how this info could be sent as an error message to the rejected server.
In any case I don't think it is a hindering reason for this PR.

ann0see:
> ) I don't remember ever seeing two servers with the same name

On Any Genre 1 there are 2 servers with a default (similar) name:

E9F24B12-203C-417E-9A62-3209AD48E85C

> I have only seen once a server changing their IP address. 

Dynamic IPs? Home hosted servers? Probably not *that* uncommon. 

Bring self hosted raspberry Pi servers following a default guide together —> the case you mentioned gets not impossible

dcorson-ticino-com:
>On Any Genre 1 there are 2 servers with a default (similar) name:

Here we have the worst case, defaults everywhere and no possibility for a user to know which server they are choosing.
I think the server should disallow registering duplicate names.
But, as I said, how do we give that info to the server and it's administrator?

jp8:
The issue that I see is :

If a user goes to a directory server, they can reasonably expect to see whatever jamulus servers are currently registered, even if server B is using the same name as server A currently uses or used in the past.

But if a user bookmarks a server A, they should never be sent to server B.  If server A disappears, and server B (with the same name) appears, I think the user would expect the bookmark to be invalid.  That's why I think a keypair is required, so that the directory server and client can positively identify a jamulus server (and match it up with the bookmark).

pljones:
Uniqueness is IP address + port number.  The directory server assumes any requests based on that pair are the same server.  Given that pair should be unique in the IPv4 UDP address space, that's what "unique" means.

jp8:
Right...  I prefaced it all with

> It seems to me that if you don't want to discard favourites whose ip address has changed, 

The options I see are either

1. if IP address has changed, invalidate the bookmark, or
2. use a keypair to allow positive identification of jamulus server, and update the bookmarked ip address

gene96817:
> On Any Genre 1 there are 2 servers with a default (similar) name:
Here we have the worst case, defaults everywhere and no possibility for a user to know which server they are choosing.
I think the directory should disallow registering duplicate names.
But, as I said, how do we give that info to the server and it's administrator?

1- consider the default server name should be the public IP address and not "yourServerName".

2- if two servers register with the same name XXX, then we can register them as XXX and XXX(2), etc.
However, there is a small problem of XXX being restarted with a new IP address. If the restart occurs before the shutdown of XXX is detected it will be registered as XXX(2). Eventually, XXX will disappear and the server will be XXX(2).  For simplicity, that is just the way it is. When "another" XXX tries to register it could be XXX or XXX(3), etc. Each registration of XXX gets the first available suffix. In this way, the problem is bounded and eventually fixes itself if there is only one server named XXX. If more than one server is using the name XXX, they will have to accept the suffix assignment or choose a unique name.

I suggest we don't try too hard to solve the problem create by default server names or for admins who choose a name that collides with another server. (If they really want a certain name, they can resort to copyrights, a system that exists outside of our Jamulus efforts.)

gene96817:
> That's why I think a keypair is required, so that the directory server and client can positively identify a jamulus server (and match it up with the bookmark).

A key pair certainly guarantees uniqueness of a server.  This scheme also needs a mechanism to reset and propagate the key pair. We can't assume that once a server creates a key pair, it will always be valid. Mistakes can be made that then require a new key pair.

SImilarly, IP+port cannot be assumed never to change. There are a range of normal events that would change the IP+port.

dcorson-ticino-com:
>1- consider the default server name should be the public IP address and not "yourServerName".

Looking at the code I see that the default server name is "", not "yourServerName".  Apparently someone has made a distribution setting that name.  :( changing the default name will not help.

gene96817:
> default server name is ""

Can this be fixed with the code that receives the command line -o argument? If -o is not present, then set the name to the IP address.

pljones:
> There are a range of normal events that would change the IP+port.

If the server changes IP or port, the directory considers it newly created.  So "no", it's not a "normal" event.

dcorson-ticino-com:
>Can this be fixed with the code that receives the command line -o argument? If -o is not present, then set the name to the IP address. (This fix sounds too obvious.)

I think something like that would work, but it won't help in the cases we are seeing as the server name is set somewhere to "yourServerName".

dcorson-ticino-com:
The problem for the favorites is that the real determining factor is the IP:Port and not the server name.  Not simple to solve.
And I don't really want to simply remove IP:Ports not seen as they could simply be offline of not responding at the moment for some reason.

pljones:
> And I don't really want to simply remove IP:Ports not seen as they could simply be offline or not responding at the moment for some reason.

Yes, not being registered doesn't mean the IP+port isn't valid, it just means it can't be connected to.  The thing is, it could now be registed on another directory...  So do you go hunting..?

jp8:
> server name is being set to "yourServerName", probably with a -o (not sure if there are other ways to do it).

Yes I think it comes from this doc:  https://jamulus.io/it/wiki/Choosing-a-Server-Type

jp8:
> I don't really want to simply remove IP:Ports not seen as they could simply be offline or not responding at the moment for some reason.

I imagine in the favourites list, these servers would turn inactive (grey).  They might turn active again, if they come alive in the future.  Or the user might have to manually clean them from the list.

jp8:
> , not being registered doesn't mean the IP+port isn't valid, it just means it can't be connected to. The thing is, it could now be registed on another directory... So do you go hunting..?

Am I right, that in the case where the server admin hasn't opened the udp port, you need to check through each directory server  anyway, just to know whether the IP+port can be connected to or not...

gene96817:
> Yes I think it comes from this doc: https://jamulus.io/it/wiki/Choosing-a-Server-Type

Oh nuts, the new user is misunderstanding the intent of "yourServerName".  Its a documentation bug.

gene96817:
> There are a range of normal events that would change the IP+port.
If the server changes IP or port, the directory considers it newly created. So "no", it's not a "normal" event.

That wasn't what I intended. It does not happen often, but it is considered normal for an ISP of residential service to occasionally change your gateway's public IP address. This is why some applications need a dynamic DNS service,

jp8:
t it is considered normal for an ISP of residential service to occasionally change your gateway's public IP address. This is why some applications need a dynamic DNS service,

I think small Amazon cloud VMs also get a new IP on restart.

dcorson-ticino-com:
> it is considered normal for an ISP of residential service to occasionally change your gateway's public IP address. This is why some applications need a dynamic DNS service

The present pull request has no problem with this... as long as the server names are unique.
Maybe the thing to do is enforce unique server names as Gene suggested, that would also work with the "yourServerName" problem.

Note that servers not responding are shown as offline  in the present PR.

jp8:
> You're building something that needs to work the same way a directory server does - basically, you're providing a "favourites directory server". 

What would you think about a UI like this:

Screenshot_2021-07-09_11-47-52

Though I think instead of 'Favourites' perhaps named something like 'My Connections' .

The 'My Connections' directory is local to the client, and would show recent connections, with a little star or heart beside each one in order to make it a favourite.  Favourites would go to the top of the list, and could be re-ordered by drag-drop.

The "server connect" combobox would only need to appear when "My connections" is selected.

ann0see:
> The "server connect" combobox would only need to appear when "My connections" is selected.


Hmm. IMHO this makes connecting to private servers (behind a normal home connection) more complicated.

jp8:
> an onslaught of people who apparently don't know what menus are. They are unable to set their profile because it is hidden in a menu.

I think problem is that it is on the wrong menu.  Users would expect the View menu to hold:
-  compact | normal | fancy 
- sort by instrument...
- sort by city ...
On a Settings menu they would expect to find
- my profile
- auto-adjust faders
- input boost

Anyway it's for a different discussion.

jp8:
> > The "server connect" combobox would only need to appear when "My connections" is selected.
> 
> Hmm. IMHO this makes connecting to private servers (behind a normal home connection) more complicated.

Perhaps, but... why should a user see the option to go to a private server when they've selected a Directory that has nothing to do with that private server ?  Private servers by definition are not associated with directories.  Maybe the option should be Directory: None instead of Directory: My Connections.

ann0see:
It’s one more step which is confusing. This comes with

> an onslaught of people who apparently don't know what menus are.

jp8:
> It’s one more step which is confusing. This comes with
> 
> > an onslaught of people who apparently don't know what menus are.

Yeah I think the menus are a separate issue though.

I'm not proposing that it is necessary to hide the custom server combobox - just that it would clean up the connect dialog and leave more remove for the server list, and that the combobox only really makes sense (in my view) if the user has *not* already selected a remote directory server.

If the user has selected a remote directory server, they have in fact already taken an extra step, away from the default "my connections" area where they can see their history, favourites, and enter a new server.

pljones:
What's the state here?  Can it be picked up?

I'd rather not totally lose this and #1728 (which I'd like to be done through this change, really -- once you've got one favourites list, why not have multiple lists....? "Normal use list", "Worldjam Studio Manager list"... etc)

ann0see:
@jp8 @dcorson-ticino-com this PR seems to be stale. As I mentioned somewhere else, I'm not totally behind this PR (but that's since I usually use only one server...)

dcorson-ticino-com:
As you said, stale.  As another solution was found to be better I am letting those with the better solution complete it.

ann0see:
Ok. So I'm closing this in terms of Peter's persistence flag and jp8's directories. Nevertheless, I think it's still worth having the code on your repo. So don't delete it ;-).

PR #1554, opened 2021-04-24
Changed Tab names, Audio/Network Setup, User Setup

dcorson-ticino-com:


ann0see:
Thanks. Not sure what happened here. There are a lot more changes than I expected. Did you tweak the GUI?

dcorson-ticino-com:
Yes, just the Pan slider on the Advanced tab.
I had made a temporary change just to see how it looked a while ago and apparently forgot to change it back in what got pushed at the time.  It is now clean again.

ann0see:
Yep. The pan slider also looks better now. Could you please also have a look at the code? I think there are some references to "Advanced". 

I know that's not related to your UI change, but "Measurements" doesn't look correctly aligned. Maybe one needs to move the "Enable small network buffers" into the "network buffers box". 

adv

ann0see (approved):

dcorson-ticino-com:
I did not change the formatting as under Win10 it looks fine and I think we need to keep space on the left side for the device selection on Macs

pljones:
I'd actually prefer to stick with "My Profile" for the tab name - it's what people get told to look for currently to set their details.  Changing it will cause unnecessary confusion, I think, even though it now also has the player's preference for language and skin.  It's easier to say "Those have moved to your profile" than to say "'My Profile' is now under 'User' and called 'Musician's Profile'".

pljones:
I think it's just the way the frame borders are rendered - they've got margins insetting them from the widgets outside.  But that's not changed, as @dcorson-ticino-com says.  (And I'm against having the network buffers checkbox near the audio buffer radio buttons.  I needs to be on the network side of the tab.)

dcorson-ticino-com:
Do we have any consensus concerning My Profile vs. User Setup ?
I have no strong feelings although my first thoughts were as plj said.

softins:
> Do we have any consensus concerning My Profile vs. User Setup ?
> I have no strong feelings although my first thoughts were as plj said.

I think I would prefer to keep the name “My Profile”, as it is definitely what people are used to.

gilgongo:
Yes, "My Profile" may be better as it's a bit of a convention, and might get mixed up with the forthcoming "Setup" page on the website (eg "Have a look at Setup" "On the website?" "No, in the app").

BTW there is always a debate about using "My" for these things. The gradual consensus over the years has been that "My..." may be confusing if users think it means the "the system/software" (so not theirs). And there's always the hilarious tech support conversation: "If you go to My Profile..." "_Your_ profile??" "No, I mean my .. uh, your profile that's called 'my'"

hoffie (approved):
Works for me, looks good.

@gilgongo 
> BTW there is always a debate about using "My" for these things. The gradual consensus over the years has been that "My..." may be confusing if users think it means the "the system/software" (so not theirs). And there's always the hilarious tech support conversation: "If you go to My Profile..." "Your profile??" "No, I mean my .. uh, your profile that's called 'my'"

Not sure if you meant to advise not yet merging this PR? Therefore I'm not merging yet. Please do if you think it is ok. :)

pljones (approved):

gilgongo:
@hoffie Sorry, it's just that thinking about it, we don't use "My..." anywhere else, so maybe we should just call it "Profile" unless we want to go into the great UX "My..." debate and have "My Settings" and stuff as well?

hoffie:
If this was a new option, I'd certainly argue in favor of just "Profile". Arguments were brought up to keep this as close to the previous term which would mean keeping "My Profile" although it's not that consistent with the other tabs.

Either works for me. :)

gilgongo:
Ok sure - it's not a big deal.

PR #1557, opened 2021-04-25
Fixes #1556 duplicate nl accelerator keys

### LINKED ISSUE #1556: Same ALT shortcut used for Menu and item on screen in Dutch language
A minor issue exists in the Dutch language.
The menu item for file is named: Bestand
And the ALT combination used for this is ALT + s

The Settings item (displayed on screen, not the menu) is named: Instellingen
And the ALT combindation used for this is ALT + s

Meaning both options are using the same combination.

**Expected behavior**
There should not be 2 items available at the same time using the same ALT combination.
In most applications Bestand will use the shortcut ALT + B, so that menu item should be updated.
In the view menu, there is also an option to get to the settings. This is using ALT + I. I would propose to use the ALT + I on the main screen as well for the Instellingen item.

**Screenshots**
jamulus-alt-keys

**Operating system**
Windows 10

**Version of Jamulus**
3.7.0

henkdegroot:
Fixes #1556 duplicate nl accelerator keys

### DIFF
--- a/src/res/translation/translation_nl_NL.ts
+++ b/src/res/translation/translation_nl_NL.ts
@@ -889,7 +889,7 @@
     <message>
         <location filename="../../clientdlg.cpp" line="247"/>
         <source>&amp;File</source>
-        <translation>Be&amp;stand</translation>
+        <translation>&amp;Bestand</translation>
     </message>
     <message>
         <location filename="../../clientdlg.cpp" line="262"/>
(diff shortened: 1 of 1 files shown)


pljones (approved):
@softins, does your magic script concur?

softins (approved):
This is fine, although we will have to go through the checking exercise again once all the new translations have been done (e.g. "Auto-Adjust all Faders").

PR #1560, opened 2021-04-25
Fix X/Y distribution on mixer board

pljones:
Five clients, one row:
image
Two rows:
image
Three rows (the window would open any taller for me):
image

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1071,6 +1071,11 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
     // sort the channels according to the first of the pair
     std::stable_sort ( PairList.begin(), PairList.end() );
 
+    // we want to distribute iNumVisibleFaders across the first row, then the next, etc
+    // up to iNumMixerPanelRows.  So row wants to start at 0 until we get to some number,
+    // then increase, where "some number" means we get no more than iNumMixerPanelRows.
+    const int iNumFadersFirstRow = 1 + ( iNumVisibleFaders / iNumMixerPanelRows );
+
     // add channels to the layout in the new order, note that it is not required to remove
     // the widget from the layout first but it is moved to the new position automatically
     int iVisibleFaderCnt = 0;
@@ -1081,11 +1086,11 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
 
         if ( vecpChanFader[iCurFaderID]->IsVisible() )
         {
-            // per definition: the fader order is colum-first/row-second (note that
-            // the value in iNumFadersFirstRows defines how many rows we will get)
+            // channels are added row-first, up to iNumFadersFirstRow, then onto
+            // the next row.
             pMainLayout->addWidget ( vecpChanFader[iCurFaderID]->GetMainWidget(),
-                    iVisibleFaderCnt % iNumMixerPanelRows,
-                    iVisibleFaderCnt / iNumMixerPanelRows );
+                    iVisibleFaderCnt / iNumFadersFirstRow,
+                    iVisibleFaderCnt % iNumFadersFirstRow );
 
             iVisibleFaderCnt++;
         }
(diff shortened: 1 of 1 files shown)


pljones:
Note: this was introduced by part of commit fe818ae7:
```patch
diff --git a/src/audiomixerboard.cpp b/src/audiomixerboard.cpp
index 315d2208..34f0875f 100755
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1071,10 +1071,6 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
     // sort the channels according to the first of the pair
     std::stable_sort ( PairList.begin(), PairList.end() );

-    // calculate the number of the faders in the first row by distribute
-    // the faders equally in the available number of rows
-    const int iNumFadersFirstRow = ( iNumVisibleFaders + 1 ) / iNumMixerPanelRows;
-
     // add channels to the layout in the new order, note that it is not required to remove
     // the widget from the layout first but it is moved to the new position automatically
     int iVisibleFaderCnt = 0;
@@ -1086,10 +1082,10 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
         if ( vecpChanFader[iCurFaderID]->IsVisible() )
         {
             // per definition: the fader order is colum-first/row-second (note that
-            // the value in iNumFadersFirstRow defines how many rows we will get)
+            // the value in iNumFadersFirstRows defines how many rows we will get)
             pMainLayout->addWidget ( vecpChanFader[iCurFaderID]->GetMainWidget(),
-                    iVisibleFaderCnt / iNumFadersFirstRow,
-                    iVisibleFaderCnt % iNumFadersFirstRow );
+                    iVisibleFaderCnt % iNumMixerPanelRows,
+                    iVisibleFaderCnt / iNumMixerPanelRows );

             iVisibleFaderCnt++;
         }

```

pljones:
The fix here basically reverts the original change but makes the comments clearer - and doesn't rely on "2 - 1 = 1" to get the maths right.  The patch against fe818ae is:
```patch
diff --git a/src/audiomixerboard.cpp b/src/audiomixerboard.cpp
index 315d2208..2744b6e7 100755
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1071,9 +1071,10 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
     // sort the channels according to the first of the pair
     std::stable_sort ( PairList.begin(), PairList.end() );

-    // calculate the number of the faders in the first row by distribute
-    // the faders equally in the available number of rows
-    const int iNumFadersFirstRow = ( iNumVisibleFaders + 1 ) / iNumMixerPanelRows;
+    // we want to distribute iNumVisibleFaders across the first row, then the next, etc
+    // up to iNumMixerPanelRows.  So row wants to start at 0 until we get to some number,
+    // then increase, where "some number" means we get no more than iNumMixerPanelRows.
+    const int iNumFadersFirstRow = ( iNumVisibleFaders + iNumMixerPanelRows - 1 ) / iNumMixerPanelRows;

     // add channels to the layout in the new order, note that it is not required to remove
     // the widget from the layout first but it is moved to the new position automatically
@@ -1085,8 +1086,8 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )

         if ( vecpChanFader[iCurFaderID]->IsVisible() )
         {
-            // per definition: the fader order is colum-first/row-second (note that
-            // the value in iNumFadersFirstRow defines how many rows we will get)
+            // channels are added row-first, up to iNumFadersFirstRow, then onto
+            // the next row.
             pMainLayout->addWidget ( vecpChanFader[iCurFaderID]->GetMainWidget(),
                     iVisibleFaderCnt / iNumFadersFirstRow,
                     iVisibleFaderCnt % iNumFadersFirstRow );
```

dcorson-ticino-com:
I must have misunderstood.
I thought you were making one row per group.

pljones:
> 
> 
> I must have misunderstood.
> I thought you were making one row per group.

No - I just used the groups to clarify the layout.  This is simply extending the existing 1 row / 2 row menu option to more rows.  But I broke the previous X/Y layout so entries were added column first, rather than row first.  This is the PR to fix that.

softins (approved):
Have tried it out, and looks great.

hoffie (approved):
Logic and UI look good. :)

Should probably be squash-merged.

pljones:
Yep, I'd meant to squash locally before updating the PR.

Issue #141, opened 2020-04-25
Bind to a specific interface/IP


marcodmb:
Maybe useful for Jamulus to bind to a specific interface (multipath routing with iproute2, ecc...). At the moment works only with the interface related to the default gateway.
Thank you and great job!

pljones:
If I remember correctly...

The server just listens on 0.0.0.0:<local port>.  So, no routing applicable there, really - if you want to limit connections, your firewall can do that easily enough.  Are you saying you want to limit the listen to a specific local IP (i.e. the address of eth0 or whatever)?

The client routes to the target IP <remote IP>:<remote port> using whatever routing rules set up, then uses the local IP and port it gets allocated to listen for replies.  You should be able to write routing for <remote IP> already that would ensure the right interface was used, right?

marcodmb:
No, is not a firewall issue. That's a load balancing setup.
My Linux server has 2 public IP addresses IP_A/subnet_A and IP_B/subnet_B.
Each public IP has a gateway for the respective subnet.
Via iproute2 I've set the proper source based routing rules:
Packets from IP_A -> GATEWAY_A and packets from IP_B -> GATEWAY_B.
The main routing table has GATEWAY_A as default (not GATEWAY_B), thus if i start a socket with any server on the internet the path is thru link A using IP_A and GATEWAY_A.

Then starting Jamulus I have:
`root@myserver:~# netstat -anp | grep Jamulus
udp        0      0 0.0.0.0:22124           0.0.0.0:*                           24720/Jamulus`
thus Jamulus listens on 0.0.0.0:22124.

If I run Jamulus setting IP_A:22124 as destination on client C just works. But if i run Jamulus setting IP_B:22124 as destination on client C **doesn't work**. UDP packets reach my server thru link_B but return thru link_A, not link_B!
Adding interface/IP binding option to Jamulus should fix this issue and permit load balancing configurations (to lower latecy for example, using multiple links and multiple Jamulus intances).

pljones:
It sounds like you're managing the set up outside the server wrongly, still, to me.  Either there is one service endpoint address - and that means one instance of Jamulus, as it's not sharing state - or there are two endpoint addresses, each with one discrete instance.

Client A connects to IP 1, that gets network load balanced once it hits your network - but you need to bring those multiple routes together before arriving at the Jamulus instance handling IP 1.

marcodmb:
The setup I need is:
`Server    IP_A   <-> Router A <-> Client A`
[...]

PR #1561, opened 2021-04-25
feature request #141: add server binding address

buv:
Tying a specific IP address to a jamulus server on a host with multiple IP addresses I find very useful. So I added a command line argument --serverbindip (followed by the IP address to bind to) and routed it through until it reaches socket.cpp.

Since I am not a C++ programmer, any feedback is welcome.


gilgongo:
Hi - thanks for this! I've linked this PR to a prior discussion in which one method was discussed to do this but I'm not sure what happened in the end.

pljones (approved):

softins (approved):
Nicely done - thank you! I've built and tested, and all seems ok.

softins (changes requested):
There is a Windows build issue that needs to be resolved before this can be merged. `arpa/inet.h` is not available under Windows.

softins:
@buv `arpa/inet.h` is not available under Windows. Please could you remove that `#include` and instead of using `inet_addr()`, revise your address binding in `CSocket::Init()` to use the same technique as is used in `CSocket::SendPacket()` to set the destination address when sending? Thank you!

pljones (changes requested):
As per @softins comment.

buv:
Changes are made.

softins (approved):
That's great, thank you!

pljones (approved):

pljones:
Thanks for the change!

gilgongo:
Just a minor thing, but seeing as this option doesn't have a short form, would it be too late to drop the "server" bit from it? We don't use (for example) `--serverlog` or `--serverdelaypan` for example, and just have it as `--bindip`. Just makes things a bit more compact on the CLI is all.

pljones:
`--serverpublicip` is there already.  Arguably, so is `--centralserver` :)...

gilgongo:
Oh OK. Well never mind then.

buv:
I tried to "orient" (euphemism for copy-and-paste) myself on the --serverpublicip parameter and did not want to pertubate the shortcut name space. "Shortcuts for all" is perhaps then another feature request? ;-)

hoffie:
@buv Do you want to add yourself to the contributors list?
https://github.com/jamulussoftware/jamulus/blob/0d593fc959a7bdca351d27fe10d59e775cc6228c/src/util.cpp#L492

This would have to be a separate PR as this one has already been merged.

buv:
Thank you very much for the invitation!

PR #1562, opened 2021-04-26
Dutch language update

### LINKED ISSUE #1558: Mix of formal and informal language used in Dutch text
**Describe the bug**
Currently the Dutch language is using a mix of formal and informal. I would like to update all to make them consistent, but to that I need to know what the preferred option is.

In some cases the text can be altered slightly to eliminate this, but unfortunately this cannot be done in all cases.

**Expected behavior**
Textstrings should be consistent.

**Additional context**
This is similar as to using du and Sie in German.

henkdegroot:
Fixes #1558

### DIFF
--- a/src/res/translation/translation_nl_NL.ts
+++ b/src/res/translation/translation_nl_NL.ts
@@ -209,7 +209,7 @@
     <message>
         <location filename="../../audiomixerboard.cpp" line="849"/>
         <source>When connected to a server, the controls here allow you to set your local mix without affecting what others hear from you. The title shows the server name and, when known, whether it is actively recording.</source>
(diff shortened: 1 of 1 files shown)


henkdegroot:
@jerogee  I have updated the 10 strings. Can you please review this PR as well?

gilgongo:
Looking at the linked issue, it seems this can be merged @jerogee ?

jerogee (approved):
All correct replacements/improvements.

gilgongo (approved):

henkdegroot:
@gilgongo  I have noticed that 1 updated string in the NL file has not been merged into the final master branch. Looks like being caused by an update performed to change "central server" into "directory server". The update from central to directory will have to be done later (at least I assume before it is decide to make a new release available). But should I do another PR in the meanwhile to get this single string updated?

It happens to be a line which includes the text "central server", but I want to prevent this is forgotten.
Or is there another method which can be used to merge this 1 string?

PR #1569, opened 2021-04-27
Shorten no ASIO error message

ann0see:
Multiple error messages in Jamulus are rather lengthy and probably not read by users. This changes one of the first messages a new and inexperienced user might see.

See: https://github.com/jamulussoftware/jamulus/discussions/1568#discussioncomment-665995

### DIFF
--- a/windows/sound.cpp
+++ b/windows/sound.cpp
@@ -563,7 +563,7 @@ CSound::CSound ( void           (*fpNewCallback) ( CVector<int16_t>& psData, voi
         throw CGenErr ( "<b>" + tr ( "No ASIO audio device driver found." ) + "</b><br><br>" +
             QString ( tr ( "Please install an ASIO driver before running %1. "
             "If you own a device with ASIO support, install its official ASIO driver. "
-            "If not, you'll need to use a driver like ASIO4ALL." ) ).arg ( APP_NAME ) );
+            "If not, you'll need to use a universal driver like ASIO4ALL." ) ).arg ( APP_NAME ) );
     }
     asioDrivers->removeCurrentDriver();
 
--- a/windows/sound.cpp
+++ b/windows/sound.cpp
@@ -560,13 +560,10 @@ CSound::CSound ( void           (*fpNewCallback) ( CVector<int16_t>& psData, voi
     // in case we do not have a driver available, throw error
     if ( lNumDevs == 0 )
     {
-        throw CGenErr ( "<b>" + tr ( "No ASIO audio device (driver) found." ) + "</b><br><br>" +
-            tr ( "The " ) + APP_NAME + tr ( " software requires the low latency audio "
-            "interface ASIO to work properly. This is not a standard "
-            "Windows audio interface and therefore a special audio driver is "
-            "required. Either your sound card has a native ASIO driver (which "
-            "is recommended) or you might want to use alternative drivers like "
-            "the ASIO4All driver." ) );
+        throw CGenErr ( "<b>" + tr ( "No ASIO audio device driver found." ) + "</b><br><br>" +
+            QString ( tr ( "Please install an ASIO driver before running %1. "
+            "If you own a device with ASIO support, install its official ASIO driver. "
+            "If not, you'll need to use a driver like ASIO4ALL." ) ).arg ( APP_NAME ) );
     }
     asioDrivers->removeCurrentDriver();
 
--- a/windows/sound.cpp
+++ b/windows/sound.cpp
@@ -563,7 +563,7 @@ CSound::CSound ( void           (*fpNewCallback) ( CVector<int16_t>& psData, voi
         throw CGenErr ( "<b>" + tr ( "No ASIO audio device driver found." ) + "</b><br><br>" +
             QString ( tr ( "Please install an ASIO driver before running %1. "
             "If you own a device with ASIO support, install its official ASIO driver. "
-            "If not, you'll need to use a universal driver like ASIO4ALL." ) ).arg ( APP_NAME ) );
+            "If not, you'll need to download and install a universal driver like ASIO4ALL." ) ).arg ( APP_NAME ) );
     }
     asioDrivers->removeCurrentDriver();
 


windows/sound.cpp
```diff
@@ -560,13 +560,10 @@ CSound::CSound ( void           (*fpNewCallback) ( CVector<int16_t>& psData, voi
...
-            tr ( "The " ) + APP_NAME + tr ( " software requires the low latency audio "
-            "interface ASIO to work properly. This is not a standard "
-            "Windows audio interface and therefore a special audio driver is "
-            "required. Either your sound card has a native ASIO driver (which "
-            "is recommended) or you might want to use alternative drivers like "
-            "the ASIO4All driver." ) );
+        throw CGenErr ( "<b>" + tr ( "No audio device with an ASIO driver detected" ) + "</b><br><br>" +
+            tr ( "To use " ) + APP_NAME + tr ( ", you need to install an ASIO driver first. "
+            "If you own a device with ASIO support, install its official ASIO driver. "
+            "If not, you'll need to use one like ASIO4ALL. " ) );
```
hoffie: Is the trailing space required?

```suggestion
            "If not, you'll need to use one like ASIO4ALL." ) );

```

ann0see: No. See the latest commit.

windows/sound.cpp
```diff
@@ -560,13 +560,10 @@ CSound::CSound ( void           (*fpNewCallback) ( CVector<int16_t>& psData, voi
...
     {
-        throw CGenErr ( "<b>" + tr ( "No ASIO audio device (driver) found." ) + "</b><br><br>" +
-            tr ( "The " ) + APP_NAME + tr ( " software requires the low latency audio "
-            "interface ASIO to work properly. This is not a standard "
-            "Windows audio interface and therefore a special audio driver is "
-            "required. Either your sound card has a native ASIO driver (which "
-            "is recommended) or you might want to use alternative drivers like "
-            "the ASIO4All driver." ) );
+        throw CGenErr ( "<b>" + tr ( "No audio device with ASIO driver detected" ) + "</b><br><br>" +
+            tr ( "To use " ) + APP_NAME + tr ( ", you need to install an ASIO driver first. "
```
hoffie, line 564: I wonder if this construct makes translations harder. Previously, it was just an article ("The"), now it's an infinitive construction. This works for German, but I'm not sure about other languages.
The following may be more flexible, but I'm not sure if it's better overall. Maybe someone else can comment as well. I think @pljones once mentioned this pattern.

`QString ( tr ( "The %1 software requires..." ) ).arg ( APP_NAME )`

(Hrm, I'm pretty sure I had already typed this once, but somehow it got lost!?)

windows/sound.cpp
```diff
@@ -560,13 +560,10 @@ CSound::CSound ( void           (*fpNewCallback) ( CVector<int16_t>& psData, voi
...
     if ( lNumDevs == 0 )
     {
-        throw CGenErr ( "<b>" + tr ( "No ASIO audio device (driver) found." ) + "</b><br><br>" +
-            tr ( "The " ) + APP_NAME + tr ( " software requires the low latency audio "
-            "interface ASIO to work properly. This is not a standard "
-            "Windows audio interface and therefore a special audio driver is "
-            "required. Either your sound card has a native ASIO driver (which "
-            "is recommended) or you might want to use alternative drivers like "
-            "the ASIO4All driver." ) );
+        throw CGenErr ( "<b>" + tr ( "No audio device with ASIO driver detected" ) + "</b><br><br>" +
```
pljones, line 563: I prefer the original `No ASIO audio device (driver) found.`, to be honest, for the first line.  (It's got a full stop, too :) .)

windows/sound.cpp
pljones, line 564: I think the following is more 
```
tr ( "You need to install an ASIO driver before running" ) + " " APP_NAME + ". "
```

pljones: Ah yes - this format does make it clearer for the translators.  I'd forgotten about it, too!  This way, they get more of the context, it's not so broken up.

windows/sound.cpp
ann0see: Yes, this is shorter but maybe not that clear for a new user who doesn’t know anything about ASIO. 

Tagging @mulyaj here who proposed this exact wording.

pljones: Whether they know about ASIO doesn't make one preferable to the other, as both describe the problem as being with something to do with ASIO.  The point is, it's not the _device_, it's the _driver.  Your phrasing changes the emphasis.  Indeed, the original (with the `()`) had it wrong, in my view.  But it had a full stop...

ann0see: My initial thought was "Install an ASIO driver to use Jamulus" which empathises the solution rather then the error. I think muljay didn’t like it because it doesn’t clearly explain the error.

mulyaj: > Your phrasing changes the emphasis

That's right, I agree with @pljones here.

windows/sound.cpp
mulyaj: Instead of 
``` You need to install an ASIO driver .... ```

Perhaps we can use
```Please install an ASIO driver ... ```


_You need to_, and moreso, _require_ can be read as demanding to the user.

Comment (pljones) [reply windows/sound.cpp:564]: True.  "Thou shalt!" might have been going too far, so I held back from that... :)

Comment (ann0see) [reply windows/sound.cpp]: Updated

File: windows/sound.cpp [hunk 4]
```diff
@@ -560,13 +560,10 @@ CSound::CSound ( void           (*fpNewCallback) ( CVector<int16_t>& psData, voi
...
-            tr ( "The " ) + APP_NAME + tr ( " software requires the low latency audio "
-            "interface ASIO to work properly. This is not a standard "
-            "Windows audio interface and therefore a special audio driver is "
-            "required. Either your sound card has a native ASIO driver (which "
-            "is recommended) or you might want to use alternative drivers like "
-            "the ASIO4All driver." ) );
+        throw CGenErr ( "<b>" + tr ( "No ASIO audio device driver found." ) + "</b><br><br>" +
+            QString ( tr ( "Please install an ASIO driver before running %1. "
+            "If you own a device with ASIO support, install its official ASIO driver. "
+            "If not, you'll need to use a driver like ASIO4ALL." ) ).arg ( APP_NAME ) );
```
pljones, line 566: Approving but I wonder if this should say "...a _universal_ driver like..."?

pljones (approved):

hoffie (approved):

windows/sound.cpp
ann0see: Yes. That's better.

ann0see: Done.

ann0see:
If everybody is ok with it, I'd merge this PR. The error could be more user friendly by adding a "more information" button to open jamulus.io but that's out of scope here (and not yet possible)

hoffie (approved):

mulyaj (approved):
Looks good to me!

PR #1572, opened 2021-04-28
Add buv to contributor list

buv:


hoffie (approved):
Thanks!
Commit history contains some commits more than necessary, but this should be solvable by sqash merging.

cc @gilgongo Please squash merge when you get around. :)

buv:
Thank you very much!

PR #1582, opened 2021-04-29
Server: Sync channel list right after connect

### LINKED ISSUE #1243: Improve client list synchronization or stop audio access during connect
**Describe the bug**

Client list updates are only sent when a ChanInfo protocol message has been received.
When some other event triggers an update (another client connects/disconnects/renames), state will be synchronized again.

**To Reproduce**

`channels.h:107`

**Expected behavior**

The list of participating (send/receive audio) clients should be transparent to all clients in all circumstances.

**Version of Jamulus**

3.7.0rc1

**Additional context**

I see two approaches for fixing this:

1. Make currently connecting clients transparent. This may lead to clients without a name showing up for some very short time (ideally). This might make it infeasable to continue running @softins's welcome message feature in Jamulus Explorer. :(
2. Do not accept audio from or send audio to unnamed clients.

@softins @pljones Thoughts?

hoffie:
Previously, the list of active channels was only distributed to clients when metadata such as names had changed. This allowed for clients to remain in a semi-connected state.

This change ensures that a complete channel list is sent as soon as a new client connects. Therefore, state is properly synced across all clients.

Fixes #1243.

Testing:
The difference can be observed when running client without `Protocol.CreateChanInfoMes`.
On a server without this PR, this client would be invisible to all previously connected clients.
On a server with this this PR, the client will be visible right away.

@softins Can you check how that works in combination with Jamulus Explorer's welcome message scraping? I fear it would cause a channel with an empty name to appear for a very short time. :(
With this approach, I could only think about delaying this synchronization step to avoid this. This would have the downside of not completely fixing this though.
A completely different approach is given in #1243 as Option 2.

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -665,6 +665,9 @@ void CServer::OnNewConnection ( int          iChID,
     // send recording state message on connection
     vecChannels[iChID].CreateRecorderStateMes ( JamController.GetRecorderState() );
 
+    // ensure that all other clients know about the new one
+    CreateAndSendChanListForAllConChannels();
+
     // reset the conversion buffers
     DoubleFrameSizeConvBufIn[iChID].Reset();
     DoubleFrameSizeConvBufOut[iChID].Reset();


softins:
> @softins Can you check how that works in combination with Jamulus Explorer's welcome message scraping? I fear it would cause a channel with an empty name to appear for a very short time. :(

OK, will do.

> With this approach, I could only think about delaying this synchronization step to avoid this. This would have the downside of not completely fixing this though.
> A completely different approach is given in #1243 as Option 2.

In fact, of the two options in #1243, I actually preferred option 2, but hadn't had the opportunity to investigate it.

softins:
I have just tried running this branch and registering it to `private.jamulus.io`. I connected to it from a GUI client. I then used Explorer to fetch the welcome message, and did indeed get a brief flash of the Explorer client in the GUI.

pljones:
> I'm not totally opposed to it though.

I am.  I'd rather have Explorer show that it's connected, snatched the welcome message and left without as much as a "Hi, how y'doin'?".

hoffie:
Another thought regarding keeping Explorer's welcome message scraping working: If we really want this PR (I'd be OK with not merging it if we have #1589), we could add a CLM_REQ_WELCOME_MESSAGE protocol message, which would allow Explorer to keep using the existing code for <=3.7.0 and use the CLM-based approach for >=3.8.0.
Adding these protocol extensions would add code again, so it would need consideration of course.

softins:
Personally, I don't think we need this PR if we have #1589.

hoffie:
Alright, I'm therefore closing this for now. Feel free to reopen if you feel otherwise.

PR #1588, opened 2021-04-30
Settings: Remember last used tab & move mixer rows setting to "My Profile"

dcorson-ticino-com:
If Settings window shown at program start, last tab is shown.
last state saved in ini file.
Some redundant code concerning tab change removed.

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -243,16 +243,19 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     QMenu* pViewMenu = new QMenu ( tr ( "&View" ), this );
 
     pViewMenu->addAction ( tr ( "&Connection Setup..." ), this,
-        SLOT ( OnOpenConnectionSetupDialog() ) );
+        SLOT ( OnOpenConnectionSetupDialog() ), QKeySequence ( Qt::CTRL + Qt::Key_C ) );
 
     pViewMenu->addAction ( tr ( "My &Profile..." ), this,
-        SLOT ( OnOpenMusicianProfileDialog() ) );
+        SLOT ( OnOpenUserProfileSettings() ), QKeySequence ( Qt::CTRL + Qt::Key_P ) );
 
     pViewMenu->addAction ( tr ( "C&hat..." ), this,
-        SLOT ( OnOpenChatDialog() ) );
+        SLOT ( OnOpenChatDialog() ), QKeySequence ( Qt::CTRL + Qt::Key_H ) );
 
-    pViewMenu->addAction ( tr ( "&Settings..." ), this,
-        SLOT ( OnOpenGeneralSettings() ) );
+    pViewMenu->addAction ( tr ( "Audio/Network &Settings..." ), this,
+        SLOT ( OnOpenAudioNetSettings() ), QKeySequence ( Qt::CTRL + Qt::Key_S ) );
+
+    pViewMenu->addAction ( tr ( "A&dvanced Settings..." ), this,
+        SLOT ( OnOpenAdvancedSettings() ), QKeySequence ( Qt::CTRL + Qt::Key_D ) );
 
     // optionally show analyzer console entry
     if ( bShowAnalyzerConsole )
@@ -862,12 +865,17 @@ void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
     SetMyWindowTitle ( iNewNumClients );
 }
 
-void CClientDlg::OnOpenGeneralSettings()
+void CClientDlg::OnOpenAudioNetSettings()
 {
     ShowGeneralSettings ( SETTING_TAB_AUDIONET );
 }
 
-void CClientDlg::OnOpenMusicianProfileDialog()
+void CClientDlg::OnOpenAdvancedSettings()
+{
+    ShowGeneralSettings ( SETTING_TAB_ADVANCED );
+}
+
+void CClientDlg::OnOpenUserProfileSettings()
 {
     ShowGeneralSettings ( SETTING_TAB_USER );
 }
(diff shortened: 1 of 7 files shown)


src/clientsettingsdlg.cpp
```diff
@@ -652,7 +652,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
     QObject::connect ( tabSettings, &QTabWidget::currentChanged,
         this, &CClientSettingsDlg::OnTabChanged );
 
-    tabSettings->setCurrentIndex ( iTabIdx );
+    tabSettings->setCurrentIndex ( pSettings->iSettingsTab ); //pSettings->iSettingsTab
```
pljones, line 655: I guess the comment was whilst copy/pasting?  It can go, right?

src/settings.cpp
```diff
@@ -533,6 +533,13 @@ if ( GetFlagIniSet ( IniXMLDocument, "client", "defcentservaddr", bValue ) )
         bWindowWasShownConnect = bValue;
     }
 
+    // selected Settings Tab
+    if ( GetNumericIniSet ( IniXMLDocument, "client", "settingstab",
```
pljones, line 537: clang-format will have line width somewhere around the 150 mark, so some lines are wrapped a bit narrow.

src/settings.cpp
```diff
@@ -533,6 +533,13 @@ if ( GetFlagIniSet ( IniXMLDocument, "client", "defcentservaddr", bValue ) )
...
+         0, 3, iValue ) )
```
hoffie, line 538: Really minor and haven't tested it, but if `0, 3` is the inclusive range of permitted values and  there are only three tabs, then I think this should rather be `0, 2` to permit indices 0, 1, and 2?

src/clientdlg.cpp
```diff
@@ -858,6 +858,16 @@ void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
...
 }
 
+void CClientDlg::OnOpenGeneralSettings()
+{
+    ShowGeneralSettings( SETTING_TAB_AUDIONET );
+}
+
+void CClientDlg::OnOpenMusicianProfileDialog()
+{
+    ShowGeneralSettings( SETTING_TAB_USER );
```
hoffie, line 868:
```suggestion
    ShowGeneralSettings ( SETTING_TAB_AUDIONET );
}

void CClientDlg::OnOpenMusicianProfileDialog()
{
    ShowGeneralSettings ( SETTING_TAB_USER );
```

src/clientsettingsdlg.cpp
dcorson-ticino-com: Oh, yes, of course

hoffie (dismissed):
Looks good to me, thanks for solving this so quickly! :)

Tested:
- First/Second/Third tab are properly restored on restart
- Works properly with empty inifile
- Opening settings still switches correctly to the right tab

hoffie (approved):
I just realized that this PR also includes the changes to move the mixer rows setting to My Profile. I have updated the PR title to make that clear.
Would be good, if you could also make the commit description clearer in this regard?

Nevertheless, this part works properly as well. :)

src/clientdlg.cpp
```diff
@@ -858,6 +858,16 @@ void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
     SetMyWindowTitle ( iNewNumClients );
 }
 
+void CClientDlg::OnOpenGeneralSettings()
```
pljones, line 861: Sorry to be a pain.  If this is `SETTING_TAB_AUDIONET`, can the slot be called `OnOpenAudioNetSettings`, please, for consistency.  Alternatively, change the name of the defined value.

pljones, line 861: I think the slot name needs to change to match what it does -- `OnOpenGeneralSettings` should be `OnOpenAudioNetSettings`.

src/clientdlg.cpp
```diff
@@ -858,6 +858,16 @@ void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
...
+{
+    ShowGeneralSettings ( SETTING_TAB_AUDIONET );
+}
+
+void CClientDlg::OnOpenMusicianProfileDialog()
```
pljones, line 866: Similarly, now there's no dedicated profile dialog, this should change, at least to end `...Settings`.

src/clientdlg.cpp
```diff
@@ -858,6 +858,16 @@ void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
...
+{
+    ShowGeneralSettings ( SETTING_TAB_USER );
+}
+
```
pljones, line 870: And part of me wants a View->Advanced settings menu option to get to that tab...

pljones:
None of my comments are blocking to get this into 3.8.0.

src/clientdlg.cpp
dcorson-ticino-com: For readability that is a plus.
I have been going on the minimal changes of existing, but that may not be the best way to go.
I will clean this up this afternoon.

pljones (approved):

pljones:
Approved.  Just a couple of notes.

----

I've not checked the ctrl key sequences for clashes.  I tend to look at something like [this](https://support.microsoft.com/en-us/windows/keyboard-shortcuts-in-apps-139014e7-177b-d1f3-eb2e-7298b2599a34) to get an idea on what's "normal" -- i.e. how hard to avoid things like the clipboard sequences.

----

I'm of the opinion we should start slowly moving towards the functor/lambda style connections.

Rather than:
```
    pViewMenu->addAction ( tr ( "A&dvanced Settings..." ), this,
        SLOT ( OnOpenAdvancedSettings() ), QKeySequence ( Qt::CTRL + Qt::Key_D ) );
```
and needing three methods:
```
void CClientDlg::OnOpenAdvancedSettings()
{
    ShowGeneralSettings ( SETTING_TAB_ADVANCED );
}
```
just do it in line:
```
    pViewMenu->addAction ( tr ( "A&dvanced Settings..." ),
        this, [=]{ ShowGeneralSettings ( SETTING_TAB_ADVANCED ); },
        QKeySequence ( Qt::CTRL + Qt::Key_D ) );
```

PR #1589, opened 2021-04-30
Only send audio to clients that have been identified.

### LINKED ISSUE #1243: Improve client list synchronization or stop audio access during connect

softins:
Fixes #1243.

### DIFF
--- a/src/channel.cpp
+++ b/src/channel.cpp
@@ -728,9 +728,11 @@ void CChannel::PrepAndSendPacket ( CHighPrioSocket*        pSocket,
                                    const CVector<uint8_t>& vecbyNPacket,
                                    const int               iNPacketLen )
 {
+    // From v3.8.0 onwards, a server will not send audio to a client until that client has sent channel info.
+    // This addresses #1243 but means that clients earlier than v3.3.0 (24 Feb 2013) will no longer be compatible.
     if ( bIsServer && !bIsIdentified )
     {
-        return;     // A server will not send audio to a client until that client has sent channel info
+        return;
     }
 
     QMutexLocker locker ( &MutexConvBuf );
--- a/src/channel.cpp
+++ b/src/channel.cpp
@@ -614,8 +614,8 @@ EPutDataStat CChannel::PutAudioData ( const CVector<uint8_t>& vecbyData,
                     eRet = PS_AUDIO_ERR;
                 }
 
-                // manage audio fade-in counter
-                if ( iFadeInCnt < iFadeInCntMax )
+                // manage audio fade-in counter, after channel is identified
+                if ( bIsIdentified && iFadeInCnt < iFadeInCntMax )
                 {
                     iFadeInCnt++;
                 }
--- a/src/channel.cpp
+++ b/src/channel.cpp
@@ -728,7 +728,7 @@ void CChannel::PrepAndSendPacket ( CHighPrioSocket*        pSocket,
                                    const CVector<uint8_t>& vecbyNPacket,
                                    const int               iNPacketLen )
 {
-    if (bIsServer && !bIsIdentified)
+    if ( bIsServer && !bIsIdentified )
     {
         return;     // A server will not send audio to a client until that client has sent channel info
     }
--- a/src/channel.cpp
+++ b/src/channel.cpp
@@ -361,7 +361,7 @@ float CChannel::GetPan ( const int iChanID )
 
 void CChannel::SetChanInfo ( const CChannelCoreInfo& NChanInf )
 {
-    bIsIdentified = true;
+    bIsIdentified = true;   // Indicate we have received channel info
 
     // apply value (if different from previous one)
     if ( ChannelInfo != NChanInf )
@@ -730,7 +730,7 @@ void CChannel::PrepAndSendPacket ( CHighPrioSocket*        pSocket,
 {
     if (bIsServer && !bIsIdentified)
     {
-        return;
+        return;     // A server will not send audio to a client until that client has sent channel info
     }
 
     QMutexLocker locker ( &MutexConvBuf );
--- a/src/channel.cpp
+++ b/src/channel.cpp
@@ -615,7 +615,7 @@ EPutDataStat CChannel::PutAudioData ( const CVector<uint8_t>& vecbyData,
                 }
 
                 // manage audio fade-in counter, after channel is identified
-                if ( bIsIdentified && iFadeInCnt < iFadeInCntMax )
+                if ( iFadeInCnt < iFadeInCntMax && bIsIdentified )
                 {
                     iFadeInCnt++;
                 }
--- a/src/channel.cpp
+++ b/src/channel.cpp
@@ -37,6 +37,7 @@ CChannel::CChannel ( const bool bNIsServer ) :
     iFadeInCntMax          ( FADE_IN_NUM_FRAMES_DBLE_FRAMESIZE ),
     bIsEnabled             ( false ),
     bIsServer              ( bNIsServer ),
+    bIsIdentified          ( false ),
     iAudioFrameSizeSamples ( DOUBLE_SYSTEM_FRAME_SIZE_SAMPLES ),
     SignalLevelMeter       ( false, 0.5 ) // server mode with mono out and faster smoothing
 {
@@ -360,6 +361,8 @@ float CChannel::GetPan ( const int iChanID )
 
 void CChannel::SetChanInfo ( const CChannelCoreInfo& NChanInf )
 {
+    bIsIdentified = true;
+
     // apply value (if different from previous one)
     if ( ChannelInfo != NChanInf )
     {
@@ -725,6 +728,11 @@ void CChannel::PrepAndSendPacket ( CHighPrioSocket*        pSocket,
                                    const CVector<uint8_t>& vecbyNPacket,
                                    const int               iNPacketLen )
 {
+    if (bIsServer && !bIsIdentified)
+    {
+        return;
+    }
+
     QMutexLocker locker ( &MutexConvBuf );
 
     // use conversion buffer to convert sound card block size in network
--- a/src/channel.h
+++ b/src/channel.h
@@ -98,7 +98,7 @@ public:
     bool GetAddress ( CHostAddress& RetAddr );
     const CHostAddress& GetAddress() const { return InetAddr; }
 
-    void ResetInfo() { ChannelInfo = CChannelCoreInfo(); } // reset does not emit a message
+    void ResetInfo() { bIsIdentified = false; ChannelInfo = CChannelCoreInfo(); } // reset does not emit a message
     QString GetName();
     void SetChanInfo ( const CChannelCoreInfo& NChanInf );
     CChannelCoreInfo& GetChanInfo() { return ChannelInfo; }
--- a/src/channel.h
+++ b/src/channel.h
@@ -228,6 +228,7 @@ protected:
 
     bool                    bIsEnabled;
     bool                    bIsServer;
+    bool                    bIsIdentified;
 
     int                     iNetwFrameSizeFact;
     int                     iNetwFrameSize;


softins:
This is an alternative to #1582, implementing option 2 from #1243 instead.

At the moment, it has been successfully compiled, but not yet tested.

src/channel.cpp
```diff
@@ -725,6 +728,11 @@ void CChannel::PrepAndSendPacket ( CHighPrioSocket*        pSocket,
                                    const CVector<uint8_t>& vecbyNPacket,
                                    const int               iNPacketLen )
 {
+    if (bIsServer && !bIsIdentified)
```
hoffie, line 731: Sadly, this is another two checks for each packet of each client, as far as I understand. I don't see a way to change this though and I can't judge what the performance impact is. I'd say that two comparisons should be cheap, but I lack knowledge about how much this is percentage-wise.

hoffie (commented):
Thanks! Code looks good.

This will probably prevent Jamulus clients before version 3.4.5 from working with newer servers completely as pointed out in https://github.com/jamulussoftware/jamulus/issues/1243#issuecomment-829637082.
I'm not saying that this is a problem though as those clients already had a vastly degraded user experience (lack of faders or at least fader names), the version is old and there have been no explicit compatibility guarantees for these earlier versions, as far as I understand.

The PR seems to do what the title promises. ;)
https://github.com/jamulussoftware/jamulus/issues/1243#issue-831131410's Option 2 would have also included stopping to accept audio from non-identified clients. Should this also be done? Should it be done as part of this PR?

For me, the undetected eaves-dropping case sounds more important and should be covered by this PR.

Not approving yet as I haven't tested either. :)

hoffie (changes requested):
Hrm, I suspect that a `bIsIdentified = false` is missing somewhere (`SetEnable()`?) as Channel instances are re-used, IIRC?

hoffie:
I have now tested this and it seems to work as expected:
- No audio for clients which did not send channel infos
- Proper audio for clients which behave correctly.

So, regarding correctness only the channel-reuse issue above remains.

softins:
> Hrm, I suspect that a `bIsIdentified = false` is missing somewhere (`SetEnable()`?) as Channel instances are re-used, IIRC?

This was one of the things I was planning to check today: the actual lifecycle of a channel instance.

src/channel.cpp
softins: I would think compared with all the other per-channel operations (codec, mixing, etc.), it's infinitesimal.

softins:
> This will probably prevent Jamulus clients before version 3.4.5 from working with newer servers completely as pointed out in [#1243 (comment)](https://github.com/jamulussoftware/jamulus/issues/1243#issuecomment-829637082).
> I'm not saying that this is a problem though as those clients already had a vastly degraded user experience (lack of faders or at least fader names), the version is old and there have been no explicit compatibility guarantees for these earlier versions, as far as I understand.

It might even be an advantage, so long as we pointed out the minimum compatible client version in the release notes. But I'll look to see if we can also easily be compatible with what previous client versions did (send channel name instead of channel info?).

> The PR seems to do what the title promises. ;)
> [#1243 (comment)](https://github.com/jamulussoftware/jamulus/issues/1243#issue-831131410) Option 2 would have also included stopping to accept audio from non-identified clients. Should this also be done? Should it be done as part of this PR?

Obviously, it is the reception of audio that creates the channel, but it would be worthwhile excluding an unidentified channel from the mix. I'll look at that.

pljones:
> the actual lifecycle of a channel instance.

As far as I remember, they're a pool (fixed size on server start).  Unused entries get reused lowest number first.  You can connect, stop your packet stream, connect again, repeat, and use up channels using (currently none or) the same user details.

pljones:
I still see absolutely no benefit here.  Anyone can put anything in.  We're no living in a police state that requires people never, ever to even think about withholding information about themselves.

softins:
> I still see absolutely no benefit here. Anyone can put anything in. We're no living in a police state that requires people never, ever to even think about withholding information about themselves.

This is really part of the conversation in #1243 itself, as to whether it is an issue that needs to be addressed. The present PR is one way to address it, if that is considered necessary. I'm not aware of any real-life situation that has arisen.

But people are still free not to set their profile information, and Jamulus will send empty info. I see no downside to wait until that has happened before starting to send audio.

> About the only "benefit" is to Jamulus Explorer.

Obviously, I'm biassed :) But I think some people find the "fetch welcome message" feature in Explorer useful, and would be disappointed if it were removed again. And I think (many) other people would find the flashing up of a transient client, from Explorer doing so, annoying and distracting. I am in both sets, so would hope to change your mind!

I also see some value in the approach taken in this PR. In an ideal world, Jamulus would behave like SIP, where a connection is established between client and server, and the audio characteristics negotiated, before either end starts sending any audio at all.

pljones:
I think the flashing client welcome fetch connect is yet another example of why we should have a server API... rather than a reason to do this.

softins:
> I think the flashing client welcome fetch connect is yet another example of why we should have a server API... rather than a reason to do this.

Maybe, but let's go for the easy wins first!

softins:
> > This will probably prevent Jamulus clients before version 3.4.5 from working with newer servers completely as pointed out in [#1243 (comment)](https://github.com/jamulussoftware/jamulus/issues/1243#issuecomment-829637082).
> > I'm not saying that this is a problem though as those clients already had a vastly degraded user experience (lack of faders or at least fader names), the version is old and there have been no explicit compatibility guarantees for these earlier versions, as far as I understand.
> 
> It might even be an advantage, so long as we pointed out the minimum compatible client version in the release notes. But I'll look to see if we can also easily be compatible with what previous client versions did (send channel name instead of channel info?).

I have now checked the history. This PR should in fact be compatible with all client versions from 3.3.0 onwards (24 Feb 2013), as that is the version which introduced Channel Infos.

softins:
> Hrm, I suspect that a `bIsIdentified = false` is missing somewhere (`SetEnable()`?) as Channel instances are re-used, IIRC?

Fixed in 908d356

pljones:
OK, if it's as old as 3.3.0, I'll have to back down.  It's worth updating the protocol to note that the ChanInfo is now (effectively) required.  We should also note it in the release notes.

softins:
> OK, if it's as old as 3.3.0, I'll have to back down. It's worth updating the protocol to note that the ChanInfo is now (effectively) required. We should also note it in the release notes.

I'll add a comment in the code to this effect too.

> If we're moving in this direction, I'd actually like to be able to drop all or selected protocol messages received from a client whilst they're in particular states - i.e. unidentified, not having OK'd the licence. (And licence first, of course. No point them identifying if they don't agree.)

This is worth exploring, but not before the upcoming release.

src/channel.cpp
```diff
@@ -611,8 +614,8 @@ EPutDataStat CChannel::PutAudioData ( const CVector<uint8_t>& vecbyData,
                     eRet = PS_AUDIO_ERR;
                 }
 
-                // manage audio fade-in counter
-                if ( iFadeInCnt < iFadeInCntMax )
+                // manage audio fade-in counter, after channel is identified
+                if ( bIsIdentified && iFadeInCnt < iFadeInCntMax )
```
hoffie, line 618: Turning the boolean evaluation order around should in theory save one instruction in the most common path?

```suggestion
                if ( iFadeInCnt < iFadeInCntMax && bIsIdentified )
```

- Whether fade in is still active has to be checked in all cases (as before)
- Whether the client is identified will only need to be checked if fade in is active

hoffie (approved):
Looks good to me, thanks! :)

src/channel.cpp
```diff
@@ -725,6 +728,13 @@ void CChannel::PrepAndSendPacket ( CHighPrioSocket*        pSocket,
                                    const CVector<uint8_t>& vecbyNPacket,
                                    const int               iNPacketLen )
 {
+    // From v3.8.0 onwards, a server will not send audio to a client until that client has sent channel info.
```
pljones, line 731: Could we send a block of silence?  Would it help?

src/channel.h
```diff
@@ -98,7 +98,7 @@ class CChannel : public QObject
     bool GetAddress ( CHostAddress& RetAddr );
     const CHostAddress& GetAddress() const { return InetAddr; }
 
-    void ResetInfo() { ChannelInfo = CChannelCoreInfo(); } // reset does not emit a message
+    void ResetInfo() { bIsIdentified = false; ChannelInfo = CChannelCoreInfo(); } // reset does not emit a message
```
pljones, line 101: Where is this called from?  This change makes it do more than its name declares.  It now resets info and stops audio.

src/channel.cpp
```diff
@@ -611,8 +614,8 @@ EPutDataStat CChannel::PutAudioData ( const CVector<uint8_t>& vecbyData,
                     eRet = PS_AUDIO_ERR;
                 }
 
-                // manage audio fade-in counter
-                if ( iFadeInCnt < iFadeInCntMax )
+                // manage audio fade-in counter, after channel is identified
+                if ( iFadeInCnt < iFadeInCntMax && bIsIdentified )
```
pljones, line 618: Why is this _not_ added to the `if` at line 600?

src/channel.cpp
```diff
@@ -725,6 +728,13 @@ void CChannel::PrepAndSendPacket ( CHighPrioSocket*        pSocket,
                                    const CVector<uint8_t>& vecbyNPacket,
                                    const int               iNPacketLen )
 {
+    // From v3.8.0 onwards, a server will not send audio to a client until that client has sent channel info.
+    // This addresses #1243 but means that clients earlier than v3.3.0 (24 Feb 2013) will no longer be compatible.
+    if ( bIsServer && !bIsIdentified )
```
pljones, line 733: Why not send silence?

src/channel.h
softins: No, strictly speaking, it _is_ just resetting the info, including the `bIsIdentified` flag, which is just conveniently indicating there is no info. The fact that audio stops is due to decisions elsewhere in the code.

It is called in the constructor for `CChannel` and also in `CServer::PutAudioData` when it is about to use or re-use a `CChannel`.

src/channel.cpp
softins: Because that would all require much more understanding and testing for any side-effects. This was just a simple way of keeping the channel gain at zero until the channel is identified.

We still need to be able to process a new connection a few lines further down.

src/channel.cpp
softins: I don't see any point in doing so. And it's only for a very short period until the client sends its info.

hoffie: The current code in this PR also has the slight side effect of increasing the required effort for amplification attacks a tiny bit (more understanding, more bytes). Sending silence would negate that and would require more code. I'd be happier with the send-nothing approach.

pljones: OK, good reason.  And if it's been tried out and seems to work, fair enough -- I was just being cautious about changes in behaviour that the client isn't expecting.

src/channel.cpp
pljones: OK

pljones (approved):

Issue #1316, opened 2021-03-20
Install the .desktop file and icon for Jamulus Server (optionally or not)


yurivict:
In the FreeBSD port I added such desktop item. The desktop file is the same except ```Name=Jamulus Server```, ```Comment=Jam Session Server```, ```Exec=jamulus -s```, ```Icon=jamulus-server```.

The icon file ```src/res/fronticonserver.png``` is installed as ```jamulus-server.png```.

This is a convenience feature to simplify starting the server.

You can make it optional with a qmake CONFIG option, or permanent.

jujudusud:
Hi Yuri, hi all,
I want to add a Jamulus server icon in Archlinux package too.
I noticed that Windows 10 installs two differents icons while my package for Arch only installs one (the blue).
I think this could be a good improvement if Il can simply have this icon in the .desktop.

ann0see:
Yes. That would be good. I think you could create a Pull Request. But no guarantee that it'll get merged although I'd support it.

jujudusud:
Is there anything new here?
@yurivict do you pull request the new desktop file ?

yurivict:
Sorry, I don't have spare cycles for this.

jujudusud:
I will do that if it's ok for you.
Just the time for to understand how to ... and it will be ok. :-)

ann0see:
@jujudusud I can assign you to this issue if you want

jujudusud:
@ann0see Ok !

ann0see:
Ok great. I'll unsubscribe from this issue now.

jujudusud:
The PR is here #1422

jujudusud:
Need help here to make the new jamulus-server.desktop.in work with the autobuild.
Anyone ?

PR #1605, opened 2021-05-03
adding server.desktop.in and new 2020 icons

jujudusud:
Please see : #1316 and also #1422


hoffie (commented):
For the record, the previous PR for this was #1422.

- Can you provide some context where this is used exactly? Our own packaging scripts don't use this, as far as I can tell, otherwise they would need to be updated, right?

- It would sound more logical to me to name the files similar to the existing, pixel-based logo (`jamulus.png`), so it would just be `jamulus.svg` and `jamulus-server.svg`. Or is there some reason which requires other names?

Note: I may be lacking history or context here why and why in this way this is useful, so please bear with me. :)

jujudusud:
As you know, I am co-maintainer of the Archlinux AUR package from Jamulus.
That said, the desktop file is used for all packages of all distributions that use Freedesktop. The Jamulus Debian package that is generated here also uses it.
I assume that the icons should be the same regardless of the platform used. So since the old icon was called Jamulus.png and the new one has a long name and svg extension, I just copied the original file and wait for the people who make the autobuild to modify and delete Jamulus.png.
I have no problem with using Jamulus.svg and Jamulus-server.svg instead of another name.
Please let me know what I have to do. ;-)

hoffie:
> As you know, I am co-maintainer of the Archlinux AUR package for Jamulus.

To be honest, I wasn't aware... Cool! :)

> That said, the desktop file is used for all packages of all distributions that use Freedesktop. The Jamulus Debian package that is generated here also uses it.

Yep.

> I just copied the original file and wait for the people who make the autobuild to modify and delete Jamulus.png

Ok. I think it would be ok to keep jamulus.png for now. Maybe @nefarius2001 can comment regarding autobuild.
A quick check shows that jamulus.png is referenced in Jamulus.pro, so this might be the place where the SVGs would have to be added?

>  I have no problem with using Jamulus.svg and Jamulus-server.svg instead of another name.

Ok. So if there is no requirement for the -2020 name, I'd suggest naming them similar to the existing file, e.g. jamulus.svg and jamulus-server.svg.

I think it would be a good idea to ensure that hey are used somewhere. I'm a bit wary of adding files which we don't use ourselves.

jujudusud:
The problem is that the deb package didn't installs server's icon for now and didn't provide the proper desktop menu. I think it's a lack. 
Using `server.desktop.in` file solve the issue. so we should merge and use it.

For the icons, I propose another thing :
- since the icons are present in /src/res, it is not necessary to copy in distribution directory,
- then I just have to suppress the to files
- Jamulus.pro have to be updated to use the new icons and to add the server one.

Issue closed -> PR properly merged...

What do you think @hoffie ?

nefarius2001:
> Ok. I think it would be ok to keep jamulus.png for now. Maybe @nefarius2001 can comment regarding autobuild.

Sorry, I have no deeper understanding of the "debuild" part.
server.desktop.in sounds like the ideal solution.

@jujudusud  can you please use "jamulus-server.desktop.in" and  "jamulus-client.desktop.in" (if there are no restricions on naming the files)?
I think "server.desktop" is very ambiguous! (and I prefere explicit suffixes for all variants over having implicit defaults)

jujudusud:
it is actually `jamulus-server.desktop.in` for the new one  and historically `jamulus.desktop.in`.
I have no problem with changing the name of the client file but, the Debian packaging have to be updated simultaneously.
Or the autobuild process will fail.
I can't update the Debian packaging files since I have nearly no knowledge in this process.
If anybody can do that, He will be able to change the name jamulus.desktop.in file to jamulus-client.desktop.in at the same time.
In the other hand,we need someone's help to make Jamulus.pro able to use the new icons. Debian package will follow.

jujudusud:
Just discover Yesterday evening that jamuluswebsite is using the same icon with the same filename... :-)

jujudusud:
Closing and going to #1672

PR #1619, opened 2021-05-03
Update pl_PL app translation for 3.8.0

SeeLook:


src/res/translation/translation_pl_PL.ts
```diff
@@ -1459,27 +1461,27 @@ We muted your channel and activated &apos;Mute Myself&apos;. Please solve the fe
...
         <location filename="../../clientsettingsdlg.cpp" line="304"/>
         <source>Leave this blank unless you need to enter the address of a directory server other than the default.</source>
-        <translation>Należy pozostawić to puste pole, chyba że konieczne jest wprowadzenie adresu centralnego serwera innego niż domyślny.</translation>
+        <translation>Należy pozostawić to puste pole, chyba że konieczne jest wprowadzenie adresu serwera innego niż domyślny.</translation>
     </message>
     <message>
         <location filename="../../clientsettingsdlg.cpp" line="309"/>
         <source>Directory server address combo box</source>
-        <translation>Lista wyboru serwera  centralnego</translation>
+        <translation>Lista wyboru serwera</translation>
```
hoffie, line 1484: As far as I understand, this lacks the translation for "Directory", is my understanding correct and is this by intention?

hoffie (commented):
Thanks!
Looks good, only a minor question about a detail.

- [x] App translations: Only a single `.ts` file checked in (+`.qm`, this is OK)
- [ ] ~~Installer translations: Passes `tools/check-wininstaller-translations.sh`~~

src/res/translation/translation_pl_PL.ts
SeeLook: It would not harm much if it stay like this, but let's have it tiptop :-)
Patch is coming...

PR #1620, opened 2021-05-03
Update IT translation

dzpex:


hoffie (approved):
Thanks, looks good to me. :)

- [x] App translations: Only a single `.ts` file checked in (+`.qm`, which is ok)
- [ ] ~Installer translations: Passes `tools/check-wininstaller-translations.sh`~

gilgongo (approved):

PR #1627, opened 2021-05-04
Improve directory server handling in GUI mode.

### LINKED ISSUE #1624: Server with GUI registers when it shouldn't
**Describe the bug**
If the "Custom Directory Server Address" field (formerly "Custom Central Server Address) is not empty and has focus, when it loses focus for any reason (tab, click back to Server Setup, or raise another app), the server registers with a Directory Server, _even if the "Make this server Public" option is unchecked_.

**To Reproduce**
1. In a GUI environment, run `./Jamulus -s -i test.ini` and observe the console output.
2. Ensure "Make My Server Public" is not checked.
3. Click between "Server Setup" and "Options", with "Custom Directory Server Address" empty. No registrations happen.
4. Enter a value into "Custom Directory Server Address", e.g. `private.jamulus.io`. Or if it is already non-empty, just click in the field to gain focus.
5. Click back to Server Setup. The window shows "Registered", and the console shows registration status updates, even though the public checkbox is still unchecked.
6. Clicking between Options and Server Setup will show repeated registrations in the console output.
7. In this state, the Genre setting can be changed, and the server will register with the new directory server.
8. If the Public checkbox is enabled and then disabled again, the server unregisters.

**Expected behavior**
If the Public checkbox is unchecked, the server should never register.

**Operating system**
Observed on Windows, Mac and Linux

**Version of Jamulus**
Observed on 3.6.0 and latest 3.8.0beta1. Probably in all versions since Central Servers were introduced.

softins:
Server was wrongly registering with a directory server every time the custom directory server field lost focus, even if the "Make public" checkbox was not enabled.

Fixes #1624

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -856,6 +856,9 @@ int main ( int argc, char** argv )
                 // only start application without using the GUI
                 qInfo() << qUtf8Printable( GetVersionAndNameStr ( false ) );
 
+                // enable server list if a directory server is defined
+                Server.SetServerListEnabled ( !strCentralServer.isEmpty() );
+
                 // update serverlist
                 Server.UpdateServerList();
 
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -202,7 +202,7 @@ void CServerListManager::SetCentralServerAddressType ( const ECSAddType eNCSAT )
 
     QMutexLocker locker ( &Mutex );
 
-    // no update the server type
+    // now update the server type
     eCentralServerAddressType = eNCSAT;
 
     // per definition: If we are in server mode and the directory server address
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -167,31 +167,52 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
 
 void CServerListManager::SetCentralServerAddress ( const QString sNCentServAddr )
 {
+    // if the address has not actually changed, do nothing
+    if ( sNCentServAddr == strCentralServerAddress ) {
+        return;
+    }
+
+    // if we are registered to a custom directory server, unregister before updating the name
+    if ( eCentralServerAddressType == AT_CUSTOM && GetSvrRegStatus() == SRS_REGISTERED ) {
+        SlaveServerUnregister();
+    }
+
     QMutexLocker locker ( &Mutex );
 
+    // now save the new name
     strCentralServerAddress = sNCentServAddr;
 
-    // per definition: If the directory server address is empty, the server list
-    // is disabled.
     // per definition: If we are in server mode and the directory server address
-    // is the localhost address, we are in directory server mode.
-    // For the directory server, the server list is always enabled.
-    if ( !strCentralServerAddress.isEmpty() )
-    {
-        bIsCentralServer =
-            (
-              ( !strCentralServerAddress.toLower().compare ( "localhost" ) ||
-                !strCentralServerAddress.compare ( "127.0.0.1" ) ) &&
-              ( eCentralServerAddressType == AT_CUSTOM )
-            );
+    // is the localhost address, and set to Custom, we are in directory server mode.
+    bIsCentralServer =
+        (
+         ( !strCentralServerAddress.toLower().compare ( "localhost" ) ||
+           !strCentralServerAddress.compare ( "127.0.0.1" ) ) &&
+         ( eCentralServerAddressType == AT_CUSTOM )
+        );
 
-        bEnabled = true;
+}
+
+void CServerListManager::SetCentralServerAddressType ( const ECSAddType eNCSAT )
+{
+    // if the type is changing, unregister before updating
+    if ( eNCSAT != eCentralServerAddressType && GetSvrRegStatus() == SRS_REGISTERED ) {
+        SlaveServerUnregister();
     }
-    else
-    {
-        bIsCentralServer = false;
-        bEnabled         = false;
-    }
+
+    QMutexLocker locker ( &Mutex );
+
+    // no update the server type
+    eCentralServerAddressType = eNCSAT;
+
+    // per definition: If we are in server mode and the directory server address
+    // is the localhost address, and set to Custom, we are in directory server mode.
+    bIsCentralServer =
+        (
+         ( !strCentralServerAddress.toLower().compare ( "localhost" ) ||
+           !strCentralServerAddress.compare ( "127.0.0.1" ) ) &&
+         ( eCentralServerAddressType == AT_CUSTOM )
+        );
 }
 
 void CServerListManager::Update()
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -168,12 +168,14 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
 void CServerListManager::SetCentralServerAddress ( const QString sNCentServAddr )
 {
     // if the address has not actually changed, do nothing
-    if ( sNCentServAddr == strCentralServerAddress ) {
+    if ( sNCentServAddr == strCentralServerAddress )
+    {
         return;
     }
 
     // if we are registered to a custom directory server, unregister before updating the name
-    if ( eCentralServerAddressType == AT_CUSTOM && GetSvrRegStatus() == SRS_REGISTERED ) {
+    if ( eCentralServerAddressType == AT_CUSTOM && GetSvrRegStatus() == SRS_REGISTERED )
+    {
         SlaveServerUnregister();
     }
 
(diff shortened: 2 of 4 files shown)


src/serverlist.cpp
```diff
@@ -167,31 +167,52 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
...
+void CServerListManager::SetCentralServerAddressType ( const ECSAddType eNCSAT )
+{
+    // if the type is changing, unregister before updating
+    if ( eNCSAT != eCentralServerAddressType && GetSvrRegStatus() == SRS_REGISTERED ) {
+        SlaveServerUnregister();
     }
+
+    QMutexLocker locker ( &Mutex );
+
+    // no update the server type
```
npostavs, line 205:
```suggestion
    // now update the server type
```

softins: Thanks, corrected

src/serverlist.cpp
```diff
@@ -167,31 +167,52 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
...
+    if ( eNCSAT != eCentralServerAddressType && GetSvrRegStatus() == SRS_REGISTERED ) {
+        SlaveServerUnregister();
     }
+
+    QMutexLocker locker ( &Mutex );
+
+    // now update the server type
+    eCentralServerAddressType = eNCSAT;
+
+    // per definition: If we are in server mode and the directory server address
```
pljones, line 208: Why did this - and why does it still - appear twice..?

src/serverlist.cpp
```diff
@@ -167,31 +167,52 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
 
 void CServerListManager::SetCentralServerAddress ( const QString sNCentServAddr )
 {
+    // if the address has not actually changed, do nothing
+    if ( sNCentServAddr == strCentralServerAddress ) {
+        return;
+    }
+
+    // if we are registered to a custom directory server, unregister before updating the name
```
pljones, line 175: Why does it matter that it's a custom server?

pljones:
There seems to be an awful lot happening here other than what the problem was.  I'll try to piece it together...

src/main.cpp
```diff
@@ -856,6 +856,9 @@ int main ( int argc, char** argv )
                 // only start application without using the GUI
                 qInfo() << qUtf8Printable( GetVersionAndNameStr ( false ) );
 
+                // enable server list if a directory server is defined
```
pljones, line 859: So the title says it's about GUI mode.  This change is explicitly non-GUI, even in the comment...

pljones:
You'll need to explain more what wasn't working and how this is fixing it, because I really can't understand it.

src/main.cpp
softins: The problem was evident in Gui mode, but the fix required the non-Gui item you flagged so that headless still worked.

The problem was that every time the central server input field lost focus and fired the “editing finished” event, the function to update the central server was called. The logic in that function was setting `bEnabled` irrespective of the checkbox, presumably a hangover from when there was no such checkbox.

Now, `bEnabled`is only ever controlled from:
1. The checkbox,
2. Reading the ini file,
3. The headless initialisation (now needed because it had been removed from the above function).

softins: Try the procedure listed in #1624 on 3.7.0 and you’ll see what I mean.

hoffie (approved):
Thanks!
The code looks good to me and this issue really sounds like it should be fixed as it may cause people to publish their servers to public directories by accident.

I have tested:
- that Jamulus 3.6.2 behaves as you describe and has the bug.
- that your change fixes the problem and no unexpected registrations occur.
- that your change keeps registrations working as they should.

I think this is fine to be merged soon as no translations are touched, as far as I can see.

softins:
The other thing I don’t like, but didn’t address as it would be more involved, is that you can’t set your List, Server Name, etc., until you have enabled registration, which means you may have to register temporarily with a directory you don’t want to.

hoffie:
> The other thing I don’t like, but didn’t address as it would be more involved, is that you can’t set your List, Server Name, etc., until you have enabled registration, which means you may have to register temporarily with a directory you don’t want to.

I agree. Might warrant a new Issue?

softins:
> > The other thing I don’t like, but didn’t address as it would be more involved, is that you can’t set your List, Server Name, etc., until you have enabled registration, which means you may have to register temporarily with a directory you don’t want to.
> 
> I agree. Might warrant a new Issue?

Yes, I'd like to get the present bug fixed asap anyway, and will raise a new issue for the above point.

@pljones - do you need more explanation before you can approve it? Ping me on Discord if so.

pljones:
It's more time...  Not had much this week (and my brain is awash with a new project at work leading me into the darker depths of the software there).

The checkbox was meant to switch from there being an address (i.e. you've registered with something), to there being no address (i.e. you're not registering).  If you change from registered to unregistered, or you change where you're registering, you should unregister first.

There _shouldn't be_ a check box.  The drop down should have "Unregistered" as the first entry and that should have the appropriate effect.  That gets rid of the label then, too.

One day we'll have a "custom directory server list maintenance" widget that can be shared by client and server, too...

But that's all too much for right now...

src/main.cpp
pljones: Anyway, I don't like comments that repeat code.  Say "why", not "what"...

src/serverlist.cpp
```diff
@@ -167,31 +167,52 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
...
+    if ( eCentralServerAddressType == AT_CUSTOM && GetSvrRegStatus() == SRS_REGISTERED ) {
```
pljones, line 176: Why only if it's custom?  Again, the comment says what.  Tell me why....

src/serverlist.cpp
```diff
@@ -167,31 +167,52 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
...
-            );
-
-        bEnabled = true;
-    }
-    else
-    {
-        bIsCentralServer = false;
-        bEnabled         = false;
+    // is the localhost address, and set to Custom, we are in directory server mode.
+    bIsCentralServer =
```
pljones, line 187: I want there to be a method that does this as it appears more than once...  Maybe not for now.

src/serverlist.cpp
```diff
@@ -167,31 +167,52 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
...
+    if ( eCentralServerAddressType == AT_CUSTOM && GetSvrRegStatus() == SRS_REGISTERED ) {
+        SlaveServerUnregister();
+    }
+
     QMutexLocker locker ( &Mutex );
 
+    // now save the new name
     strCentralServerAddress = sNCentServAddr;
 
-    // per definition: If the directory server address is empty, the server list
```
pljones, line 174: This is now done ... elsewhere ...

src/serverlist.cpp
```diff
@@ -167,31 +167,52 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
...
-              ( !strCentralServerAddress.toLower().compare ( "localhost" ) ||
-                !strCentralServerAddress.compare ( "127.0.0.1" ) ) &&
-              ( eCentralServerAddressType == AT_CUSTOM )
-            );
-
-        bEnabled = true;
-    }
-    else
-    {
-        bIsCentralServer = false;
```
pljones, line 192: ... and, presumably, so are these ...

src/serverlist.cpp
```diff
@@ -167,31 +167,52 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
 
 void CServerListManager::SetCentralServerAddress ( const QString sNCentServAddr )
 {
+    // if the address has not actually changed, do nothing
+    if ( sNCentServAddr == strCentralServerAddress ) {
```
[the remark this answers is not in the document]
softins: Although I've just noticed I should have put the `{` on a new line (a style I don't like, so have to remember to make myself do it!). I'll do that before merging.

src/serverlist.cpp
```diff
@@ -167,31 +167,52 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
...
+    bIsCentralServer =
+        (
+         ( !strCentralServerAddress.toLower().compare ( "localhost" ) ||
+           !strCentralServerAddress.compare ( "127.0.0.1" ) ) &&
+         ( eCentralServerAddressType == AT_CUSTOM )
+        );
+
+}
+
+void CServerListManager::SetCentralServerAddressType ( const ECSAddType eNCSAT )
```
pljones, line 196: Ideally this would do nothing more than send the appropriate address to `SetCentralServerAddress`.  We shouldn't really need `eCentralServerAddressType`.

pljones:
OK, the biggest problem I have is the old code, not the fix :).

I wouldn't do it like that.  I can't remember if it was Volker or me or a mixture of the two...  But what's there isn't great.  So if the fix works, I guess it'll have to do.

It still doesn't mean I like it (because it doesn't make me like the code...).

pljones (approved):

src/serverlist.cpp
pljones: It should be setting the actual address, rather than the enum value.  It should be what's doing the registration.  Picking one of the predefined addresses should call this method, really...

softins:
> It still doesn't mean I like it (because it doesn't make me like the code...).

Indeed, I agree with you. The functionality is currently split between layers. The `UpdateList` functions shouldn't be called by the GUI code that makes the changes. The change should just be fed straight down to `ServerListManager`, which itself decides when to update things. I would have loved to restructure it, but not within a beta release phase!

PR #1629, opened 2021-05-05
Change last instance of Central Server to Directory Server

softins:
The item in clientsettingsdlgbase.ui had been missed (reported on Facebook).

Also took the opportunity to change Genre to List in the server dialog,
for consistency with the client connect dialog.

Updated translation files, and marked the Custom Directory Server Address
item as unfinished, so it won't get forgotten.

### DIFF
--- a/src/clientsettingsdlgbase.ui
+++ b/src/clientsettingsdlgbase.ui
@@ -1070,7 +1070,7 @@
              <item>
               <widget class="QLabel" name="lblCentralServerAddress">
                <property name="text">
-                <string>Custom Central Server Address:</string>
+                <string>Custom Directory Server Address:</string>
                </property>
               </widget>
              </item>
--- a/src/serverdlgbase.ui
+++ b/src/serverdlgbase.ui
@@ -79,7 +79,7 @@
          <item>
           <widget class="QLabel" name="label">
            <property name="text">
-            <string>Genre</string>
+            <string>List</string>
            </property>
           </widget>
          </item>


pljones (approved):

hoffie (commented):
This looks good, but I wonder would be the best time to merge this:
- Merging now does not seem like a good idea because it risks getting merge conflicts on all the translation PRs. At least that's what I fear?
- Merging after translation PRs may mean manual rebasing again, but looks to be the more realstic option in my opinion?

softins:
> This looks good, but I wonder would be the best time to merge this:
> 
> * Merging now does not seem like a good idea because it risks getting merge conflicts on all the translation PRs. At least that's what I fear?
> * Merging after translation PRs may mean manual rebasing again, but looks to be the more realstic option in my opinion?

I’m happy to rebase and merge after the translations are finished. The only complication is that the strings “Custom Central Server Address” were changed manually to “Custom Directory Server Address” in the .ts files without changing the translation. That’s why I added the tag type=“unfinished” to that one. It should be possible to refer to other instances of “Directory Server” to get the translation.

henkdegroot:
@softins I just noticed this issue. I had already changed the "central server" to the equivalent in the Dutch language (even for this strings). Guessing it means that when this PR gets merged, I have to re-translate. Not a problem to do so....
However I wonder why this string the clientsettingsdlgbase contains a : at the end? None of the other in these dialogs have this. As you are updating this specific string, could this be an oppertunity to remove that : as well?

gilgongo:
Also on this subject, I notice the servers refer to themselves as "central server" in the server lists, but I guess that's not a big deal to change.

pljones:
> 
> 
> Also on this subject, I notice the servers refer to themselves as "central server" in the server lists, but I guess that's not a big deal to change.

That won't change until 3.8.0 final is released, of course.  Don't want to be confusing people yet.

softins:
Just updated this to do only the two `.ui` files, not the `.ts` files. They can be done with `lupdate` and then fixed up. Much simpler.

PR #1634, opened 2021-05-05
Try latest Qt for Mac

softins:
This is not for merging (yet), but to test out the artifact building for Mac with Qt 5.15.2.


pljones:
I thought we were using 5.9 for backwards compatibility, as we don't _need_ 5.15.2.

softins:
> I thought we were using 5.9 for backwards compatibility, as we don't _need_ 5.15.2.

Yes, that’s true. It was just to get an artifact built with the latest Qt5 to see if it did the same on Big Sur. I’m not sure how to manually kick off a build workflow.

pljones:
Go to the "Checks" tab, click "Re-run jobs".  The drop down gives you not what I'd call a choice...

PR #1635, opened 2021-05-05
Fix #1622 WhatsThis

### LINKED ISSUE #1622: What Is This option not available from My Profile settings
**Describe the bug**
The "What Is This" option is no longer available for the My Profile settings.
Previously the "My Profile" dialog included the ? in the menu bar to support this.
Right click on the any item (except the new mixer rows item) does not provide access to the "What Is This" details.

**To Reproduce**
Use r3.8.0 beta and open the "My Profile" tab
Right click any item in the musicians profile section, nothing happens.

**Expected behavior**
What Is This details are available and translated as well. So they should be accessible by the users.
Either by adding the ? option to the menu bar or by allowing the right click to show the "What Is This" popup or both :)

**Screenshots**
Screenshot 3.7.0 dialog:
musician-profile

Screenshot 3.8.0 beta dialog:
settings-3-8-0

**Operating system**
Windows 10 Version 20H2

**Version of Jamulus**
r3.8.0beta

dcorson-ticino-com:
As the WhatsThis connections in the clientsettingsdlg  to comboboxes and
edit fields were all not working they have been moved to the the
corresponding labels.


pljones:
Thank you!  I've been meaning to do this for ages.

src/clientsettingsdlg.cpp
```diff
@@ -228,11 +228,17 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient*         pNCliP,
     butDriverSetup->setToolTip ( strSndCardDriverSetupTT );
 
     // fancy skin
-    cbxSkin->setWhatsThis ( "<b>" + tr ( "Skin" ) + ":</b> " + tr (
+    lblSkin->setWhatsThis ( "<b>" + tr ( "Skin" ) + ":</b> " + tr (
         "Select the skin to be used for the main window." ) );
 
     cbxSkin->setAccessibleName ( tr ( "Skin combo box" ) );
 
+    // Interface Language
```
pljones, line 236: If this is new, it's going to need calling out to the translators, as the cut's been taken.

pljones (approved):

src/clientsettingsdlg.cpp
hoffie: > If this is new, it's going to need calling out to the translators, as the cut's been taken.

As far as I can see, this string does not appear in 3.8.0beta1 (even in other places), so it is new, yes.

I'd rather not restart translations for that. One option would be to merge this explicitly after all other translation PRs have been merged. This has the obvious downside that the new strings will not be translated for 3.8.0, but as it is only about a rather hidden text and hasn't been working before either, I think it would be acceptable to do so.

I'm therefore not approving yet.

dcorson-ticino-com: Yes, the Language whatsthis is new.
Other possibility is I take it out now, as it didn't exist before, and try to remember to put it back in for the next version.

softins: I agree with @hoffie - leave this in, and merge it after the translations. I also have a PR #1629 for the `.ts` files and can deal with them both at the same time.

dcorson-ticino-com:
Anyone know why it is not resolving?
I resolved exactly the conflicts it is showing once already.
?????????????????????

hoffie:
> Anyone know why it is not resolving?

Looks like there are no more conflicts now. :)

dcorson-ticino-com:
Ja, I just resolved it here a second time.  Waiting now for the build to finish so I can check it.
Not really how I would think to have my workflow.

softins (approved):

Issue #933, opened 2021-01-31
Move mute info icons to user label


Hk1020:
I've just been experimenting with three local clients to understand what the mute info icons above the faders actually do and I found it quite confusing. The icons appear if for some reason the person of that particular fader can't hear you because he soloed or muted something. It is out of my control. Everything else, however, affects and shows only what **I** hear.

I would find it much more logical to move this info icon into the user label area. There I see info about the remote user, like their name. What I see above is stuff that I do.

What do you think? There is a related issue #811 which would be affected by such a change.

I'd also like to see my name somewhere in the GUI like in the title bar or as `Joe's personal mix at: _server_` so I know which fader is mine.

pljones:
I agree that the current location is confusing.  The "speaker icon struck through" looks like I'm not including them in my mix.  An "ear icon struck through" next to the person's name would more clearly indicate that person isn't listening to me.

gilgongo:
We're currently limited to using Unicode characters for the icons I think. But the part of this ticket that related to the icon itself might also be relevant to https://github.com/jamulussoftware/jamulus/issues/899

pljones:
Accessibility should definitely be a consideration, too.  Hover text / alt-text on all icons is always beneficial.

I run using compact skin... so you'd end up with "smaller than small" text, probably less than readable.

gilgongo:
Yes, hence my doubts about the space. 

BTW I must admit I don't know much about screen readers and desktop apps. Do they parse alt text like they do with web browsers?

pljones:
> BTW I must admit I don't know much about screen readers and desktop apps. Do they parse alt text like they do with web browsers?

AFAIK, yes - I don't know if you'd need a Qt-specific one or if Qt marks up from the accessibility text with some common standard.  I just generally consider it good practice.

I guess it could be tested with at least the Windows screen reader.  I don't know on MacOS or Linux.

Hk1020:
My idea was to have something actually right in the user label itself, i.e., the white, red or green areas in your screenshot.

gilgongo:
Yes, although there's not currently room for much at all in there (and text would be clearer). 

There are a few things that might be accumulating that point toward a bit of a re-arrangement of the UI, I think. This ticket being one of them.

gilgongo:
Moving too discussion until we can get a firm spec on this.

PR #1640, opened 2021-05-05
Mute icon: Fix alignment by not hardcoding font height

### LINKED ISSUE #1312: Icon pushes the PAN control down
**Describe the bug**
Please see the third PAN control in the screenshot:
image

This isn't a big deal, just a small visual defect.

vimpostor:
Fixes #1312

**Before**: On the Left - Volume Sliders have mismatched height, Mute icon is slightly misaligned
**After**: On the Right - Volume Sliders and Mute icon are placed without weird misalignment
Screenshot_20210506_003727

Note that I tried some cleaner ways of fixing this misalignement - that is I tried not setting minimum height at all and using a more clever alignment. None of that worked though and I actually think there is a Qt bug when rendering Unicode symbols regarding the height of layouts.

Anyways, in the end I think this is the best method of solving the problem now.

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -86,9 +86,10 @@ CChannelFader::CChannelFader ( QWidget* pNW ) :
     pPan->setRange               ( 0, AUD_MIX_PAN_MAX );
     pPan->setValue               ( AUD_MIX_PAN_MAX / 2 );
     pPan->setNotchesVisible      ( true );
-    pInfoLabel->setMinimumHeight ( 15 ); // prevents jitter when muting/unmuting (#811)
-    pPanInfoGrid->addWidget      ( pPanLabel, 0, Qt::AlignLeft );
-    pPanInfoGrid->addWidget      ( pInfoLabel );
+    pInfoLabel->setMinimumHeight ( pPanLabel->height() ); // prevents jitter when muting/unmuting (#811)
+    pInfoLabel->setAlignment     ( Qt::AlignTop );
+    pPanInfoGrid->addWidget      ( pPanLabel, 0, Qt::AlignLeft | Qt::AlignTop );
+    pPanInfoGrid->addWidget      ( pInfoLabel, 0, Qt::AlignHCenter | Qt::AlignTop );
     pPanGrid->addLayout          ( pPanInfoGrid );
     pPanGrid->addWidget          ( pPan, 0, Qt::AlignHCenter );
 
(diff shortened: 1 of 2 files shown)


gilgongo:
Thanks for this! 

As a side issue, I've often wondered whether we could use an image and not a unicode character for the icon, since I would expect Qt to handle fonts with some kind of proportional leading (and they look different on different operating systems).

src/audiomixerboard.cpp
```diff
@@ -86,9 +86,10 @@ CChannelFader::CChannelFader ( QWidget* pNW ) :
     pPan->setRange               ( 0, AUD_MIX_PAN_MAX );
     pPan->setValue               ( AUD_MIX_PAN_MAX / 2 );
     pPan->setNotchesVisible      ( true );
-    pInfoLabel->setMinimumHeight ( 15 ); // prevents jitter when muting/unmuting (#811)
-    pPanInfoGrid->addWidget      ( pPanLabel, 0, Qt::AlignLeft );
-    pPanInfoGrid->addWidget      ( pInfoLabel );
+    pInfoLabel->setMinimumHeight ( pPanLabel->height() ); // prevents jitter when muting/unmuting (#811)
+    pInfoLabel->setAlignment( Qt::AlignTop );
```
hoffie, line 90: Can you also align this similar to the other calls?

hoffie (commented):
Thanks for this PR! :)
Can you confirm (maybe with a screenshot) that this also works properly for Mono setups without Pan widgets?

vimpostor:
Screenshot of working mono setup as requested:
Screenshot_20210506_233339

I have also manually tested it to be working with all combinations of {mono,stereo} x {compact, modern, standard theme}

hoffie (approved):
Thanks! Looks good to me.

Do you want to list yourself in the [contributors list](https://github.com/jamulussoftware/jamulus/blob/master/src/util.cpp#L493)?
You can do this as part of this PR.

vimpostor:
I am not sure, if such a small change warrants me being part of the contributors list, but I intend to submit more PRs to this project, so I might as well add myself now.
Done now in a separate commit. ;)

softins (approved):
Looks good, and I have tested on Linux.

vimpostor:
I agree that the general UI of this could be improved on. Perhaps a mouseover hover text that explains what the icon means, would already be better.

hoffie:
Maybe the [**previous discussion in #1251**](https://github.com/jamulussoftware/jamulus/discussions/1251) (generated from #933) should be revived so that a new PR can be created when there is consensus if and how to do this?

I suggest continuing thereas this is more likely to get some visibility than this closed PR :)

Issue #123, opened 2020-04-20
FR: Server logging improvement


pljones:
From #91
> On the very first registration attempt we could always output the result on the command line if "no GUI server" is run.
> @pljones can you please take care of it? It should be possible with just a few lines of code. But make sure it only outputs at the very first registration attempt and not the successive registrations every 15 minutes.

Yep.

corrados:
Well, I changed my mind. I want all outputs now (as it is implemented right now). So I see no need for improvement. Do you?

pljones:
I'm happy :)

PR #1655, opened 2021-05-08
Add action to select Xcode 11.7 for Qt5 compatibility on Mac

softins:
When testing a build for Mac with Qt 5.15.2, the following warning was received:
```
Project WARNING: Qt has only been tested with version 10.15 of the platform SDK, you're using 11.1.
Project WARNING: This is an unsupported configuration. You may experience build issues, and by using
Project WARNING: the 11.1 SDK you are opting in to new features that Qt has not been prepared for.
Project WARNING: Please downgrade the SDK you use to build your app to version 10.15, or configure
Project WARNING: with CONFIG+=sdk_no_version_check when running qmake to silence this warning.
```
This is because the default macOS build environment on Github is now using Xcode 12, which by default uses the 11.1 SDK.

Although Qt 5.9.9 does not give this warning, it seems likely that is because it just doesn't check, and it would seem reasonable to conclude that any potential incompatibility between Qt 5.15.2 and the 11.1 SDK would be even more the case with Qt 5.9.9.

This PR uses the action from https://github.com/marketplace/actions/setup-xcode-version to explicitly request Xcode 11.7, which by default uses the 10.15 SDK.

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -134,6 +134,9 @@ jobs:
 
     runs-on:                        ${{ matrix.config.building_on_os }}
     steps:
+
+      # For Qt5 on Mac, we need to ensure SDK 10.15 is used, and not SDK 11.x
+      # This is done by selecting Xcode 11.7 instead of the latest default of 12.x
       - name:                       Select Xcode version for Mac
         if:                         ${{ matrix.config.target_os == 'macos' }}
         uses:                       maxim-lobanov/setup-xcode@v1
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -134,6 +134,12 @@ jobs:
 
     runs-on:                        ${{ matrix.config.building_on_os }}
     steps:
+      - name:                       Select Xcode version for Mac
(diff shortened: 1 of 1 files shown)


pljones (approved):

.github/workflows/autobuild.yml
```diff
@@ -134,6 +134,12 @@ jobs:
 
     runs-on:                        ${{ matrix.config.building_on_os }}
     steps:
+      - name:                       Select Xcode version for Mac
+        if:                         ${{ matrix.config.target_os == 'macos' }}
+        uses:                       maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: '11.7'
```
hoffie, line 141: I'm usually not a fan to add comments in the style of "This is here because of issue #0123", but maybe just add a comment similar to the PR description to explain why the version is locked at all and why it is locked to exactly this version? This could help our future selves to figure out when it is safe to update. ;o

hoffie (approved):
Thanks! :)

.github/workflows/autobuild.yml
softins: Done

PR #1664, opened 2021-05-09
Swedish translation updates for 3.8.0.

genesisproject2020:


src/res/translation/translation_sv_SE.ts
```diff
@@ -856,12 +859,12 @@ We muted your channel and activated &apos;Mute Myself&apos;. Please solve the fe
...
         <location filename="../../clientdlg.cpp" line="254"/>
         <source>Audio/Network &amp;Settings...</source>
-        <translation type="unfinished"></translation>
+        <translation>Ljud och nätverks&amp;inställningar</translation>
     </message>
     <message>
         <location filename="../../clientdlg.cpp" line="257"/>
         <source>A&amp;dvanced Settings...</source>
-        <translation type="unfinished"></translation>
+        <translation>&amp;Avancerade inställningar</translation>
```
hoffie, line 867: Hrm, maybe keep the trailing `...` in those two places?

hoffie (approved):
Looks good to me -- thanks! :)

- [x] App translations: Only a single `.ts` file checked in (+`.qm`)
- [ ] ~Installer translations: Passes `tools/check-wininstaller-translations.sh`~

src/res/translation/translation_sv_SE.ts
gilgongo: Yes, the "..." is a convention used to indicate that a menu item or function will present more items to consider before completion of the operation. I don't know if we're consistent in that convention though :-)

gilgongo: Well, the convention is to use the "..." (elipsis) to indicate that the menu item isn't a command in its own right but leads to another command or dialogue. We are inconsistent though so need to fix it in the English really. 

image

PR #1127, opened 2021-02-26
add clang-format style definition

passing:
lint all code with defined style
prerequisite for #901

PR #1679, opened 2021-05-13
disable clang-format for code that is unaligned on purpose

passing:
as discussed in https://github.com/jamulussoftware/jamulus/pull/1127

### DIFF
--- a/src/protocol.cpp
+++ b/src/protocol.cpp
@@ -2879,7 +2879,9 @@ bool CProtocol::ParseMessageFrame ( const CVector<uint8_t>& vecbyData,
 
     // Extract actual data -----------------------------------------------------
 
+// clang-format off
 // TODO this memory allocation is done in the real time thread but should be
+// clang-format on
 //      done in the low priority protocol management thread
 
     vecbyMesBodyData.Init ( iLenBy );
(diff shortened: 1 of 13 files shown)


src/protocol.cpp
```diff
@@ -2879,7 +2879,9 @@ bool CProtocol::ParseMessageFrame ( const CVector<uint8_t>& vecbyData,
 
     // Extract actual data -----------------------------------------------------
 
+// clang-format off
 // TODO this memory allocation is done in the real time thread but should be
+// clang-format on
```
hoffie, line 2884: Could be moved that to the end of the comment block for better readability?

hoffie (commented):
Thanks, looks good to me! :)
One minor nit.

I did not spot any additional places via `grep '^//.*TODO' src/* -r5`, so it seems complete to me. Anything which might still have missed will hopefully show up in the formatting PR. :)



For context: As we agreed to keep such intentionally non-aligned comments it is important to disable clang-format in these places. Otherwise, clang-format would "fix" the alignment which would lead to a loss of information when the first complete clang-format run is performed (#1127).

src/protocol.cpp
passing: ah, missed that one... now it's fixed

hoffie (approved):
Thanks :)

passing:
closing as included in https://github.com/jamulussoftware/jamulus/pull/1127

PR #1685, opened 2021-05-15
Add menu item for About Qt

softins:
This is using a built-in feature of Qt and was inspired by the same function in MuseScore. In addition to the existing About Jamulus box, it provides another menu item for "About Qt", which could be useful when diagnosing issues:
image
That displays a message box built in to Qt, which gives information about the Qt version being used and its licensing:
image
The extra tweaks to the translation files can easily be done while preparing the RC version.


pljones (approved):

hoffie (approved):

pljones:
@softins 

> QMessageBox::aboutQt(this, QString("About Qt"));

1. Spacing!
2. Translation!

Issue #1628, opened 2021-05-05
ActiveTab heading unvisible at MAC 3.8.0beta


Sommi486:
**Describe the bug**
At the actual 3.8.0beta1 in settings there ist no heading for the active tab.
If you change tab it appers again and new chosen tab has no heading.

**Screenshots**
<img width="389" alt="Bildschirmfoto 2021-05-05 um 09 59 14" src="https://user-images.githubusercontent.com/42489734/117113606-fefcbb00-ad8a-11eb-9525-cb29caf12bd9.png">

<img width="399" alt="Bildschirmfoto 2021-05-05 um 10 18 26" src="https://user-images.githubusercontent.com/42489734/117113776-3f5c3900-ad8b-11eb-83d8-69b45e03be98.png">

**Operating system**
Big Sur 11.3

**Version of Jamulus**
3.8.0beta1

softins:
I suspect it is white text on a white background. Do you use dark theme or light theme? Or another (I only have Mojave, not Big Sur). Can you change themes and try again?

Sommi486:
I use dark theme.

Changing to light theme doesn't solve the bug.

softins:
This is what it looks like on my Mac running Mojave:
![image](https://user-images.githubusercontent.com/3224952/117132804-3e330800-ad9b-11eb-94ba-8110cb25a360.png)
What does a tabbed dialog from another app look like on your system?

softins:
Looks like this bug is not specific to Jamulus
https://forum.qt.io/topic/120907/macos-big-sur-and-qtabwidget-bug

Sommi486:
Here is how it looks on Catalina:
![Bildschirmfoto 2021-05-05 um 13 22 18](https://user-images.githubusercontent.com/42489734/117134391-d0d8a480-ada5-11eb-83e9-616b52f54dec.png)

This is how other Apps look on BigSur:
<img width="243" alt="Bildschirmfoto 2021-05-05 um 13 27 49" src="https://user-images.githubusercontent.com/42489734/117134441-e2ba4780-ada5-11eb-8669-eed1ac8bf2ef.png">

And this is how BigSur Settings look in dark & light mode:
<img width="413" alt="Bildschirmfoto 2021-05-05 um 13 24 05" src="https://user-images.githubusercontent.com/42489734/117134515-f6fe4480-ada5-11eb-9577-8d2316aca916.png">
[...]

PR #1687, opened 2021-05-15
Update Qt on Mac to 5.12.10, for better compatibility with Big Sur

softins:
Note that this makes the earliest version of macOS supported by the release builds
to be 10.12 (Sierra), and drops support for Yosemite and El Capitan.

### DIFF
--- a/autobuild/mac/artifacts/autobuild_mac_1_prepare.sh
+++ b/autobuild/mac/artifacts/autobuild_mac_1_prepare.sh
@@ -8,10 +8,10 @@
 
 echo "Install dependencies..."
 python3 -m pip install aqtinstall
-python3 -m aqt install --outputdir /usr/local/opt/qt 5.9.9 mac desktop
+python3 -m aqt install --outputdir /usr/local/opt/qt 5.12.10 mac desktop
 
 # add the qt binaries to the path
-export -p PATH=/usr/local/opt/qt/5.9.9/clang_64/bin:"${PATH}"
+export -p PATH=/usr/local/opt/qt/5.12.10/clang_64/bin:"${PATH}"
 echo "::set-env name=PATH::${PATH}"
 echo "the path is ${PATH}"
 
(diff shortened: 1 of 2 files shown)


hoffie (approved):

softins:
Waiting for feedback from someone on Facebook before I merge this.

hoffie:
> Waiting for feedback from someone on Facebook before I merge this.

If the feedback is positive, we should link this PR to #1628.

pljones (approved):

JohannesBrx:
> drops support for Yosemite and El Capitan

Unfortunately, I cannot install the official MacOS builds of Jamulus any more. At least I can still build Jamulus locally. 
(And yes, I know it's getting time for a new OSX version...)

hoffie:
@JohannesBrx Any specific reason why you require the old version?

JohannesBrx:
> Any specific reason why you require the old version?

Compatibility with legacy software... I assume that it will be broken with a newer MacOS version.

hoffie:
Understood.
If there is still comprehensible demand for this support, we could think about providing an additional Mac build for old Mac versions. This should be possible via some updates/refactoring of the autobuild scripts. I think we should only do this if there is enough demand, though. As far as I understand, those Mac vesions are unsupported by Apple. Using such an old Qt build for Jamulus may also keep known (Qt) security issues unfixed, so it needs proper considerations.

softins:
> > drops support for Yosemite and El Capitan
> 
> Unfortunately, I cannot install the official MacOS builds of Jamulus any more.

Interested to know what happens when you try. Does it just refuse to install? Or install and fail to run?

Which version of macOS are you running?

JohannesBrx:
> Interested to know what happens when you try. Does it just refuse to install? Or install and fail to run?
>
> Which version of macOS are you running?

I can still open the .dmg file and confirm the software license. But when I want to run Jamulus, I get the following error message, which stops execution after confirmation. As shown in the screenshot, I am using MacOS 10.11. I think I will try upgrading to High Sierra and see what happens to my legacy software...

image

ann0see:
@JohannesBrx I'm not sure if you know, but now we provide two builds of Jamulus (to also support older versions of macOS)

PR #1689, opened 2021-05-16
Update pt_PT translation file for 3.8.0

Snayler:
Sorry for the delay, here is the updated file


hoffie (approved):
Thanks! :)

- [x] App translations: Only a single `.ts` file checked in
- [x] Installer translations: Passes `tools/check-wininstaller-translations.sh`

softins (approved):

PR #1691, opened 2021-05-16
Muted Speaker Icon update

### LINKED ISSUE #1684: Empty space in mixerboard/Replace muted speaker character by a proper icon
**Describe the bug**
In the latest v3.8.0beta2 release, there is big empty space above the PAN dial.
Possible caused by the change to setting the height of the muted speaker icon.
Also note that on Win10 the muted speaker icon appears in standard colors and not in orange

**To Reproduce**
Ensure there is one client you cannot hear, this makes the muted speaker icon to appear

**Expected behavior**
There is no need for the extra space between the dial and the PAN/Speaker text.
Possible solution (as discussed in several other issues/discussions) is to replace the Muted Speaker by an actual graphic/picture instead of using a unicode character for this.

**Screenshots**
muted-speaker

**Operating system**
Win10

**Version of Jamulus**
v3.8.0beta2

**Additional context**
Also note that on Win10 the muted speaker icon appears in standard colors and not in orange, which according to the code should be the color used.
Is a muted speaker icon the best option for this?
Perhaps an ear icon with a cross or a headphone icon with a cross would better represent this function?

henkdegroot:
Fixes #1684


pljones (approved):

softins (changes requested):
Also need to add `src/res/mutediconorange.png` to `Jamulus.pro`, somewhere around line 752.

henkdegroot:
@softins , updated as requested

hoffie (approved):
Tested on Arch Linux. Looks good and is way better than before (as it was just an orange square for me, probably due to missing font glyphs).

softins (approved):
Looks good!

When revisiting this after the release, to change to the ear icon, we should clarify the What's This text: the icon showing just means the client cannot hear you. It might not be because they have you muted (which is what the text says), because it also shows if you turn your own fader right down.

DavidSavinkoff:
The 'muted speaker' in jamulus-3.8.0rc1looks good and works good. Please  keep it this way.

PR #1701, opened 2021-05-16
Added remaining items to change log

softins:
The item about Mac compatibility can be removed if third party testing reports that it doesn't correct #1628

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -53,9 +53,16 @@
   (contributed by @dcorson-ticino-com)
 
 - GUI: Translations have been updated
-  * Dutch, by @henkdegroot (#1562, #1556)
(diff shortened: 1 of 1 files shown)


hoffie (approved):
Thanks.

For the final version, I'd like to group this some more (i.e. all GUI stuff next to each other) and move internal stuff (clang-format) to the bottom.

I can do that within the coming days though, so no need to change it here, IMO. I'll also check if some of the entries should be merged (I think the muted speaker icon display had a previous post 3.7.0 entry as well, but haven't checked again).

softins:
> I assume you want to this to be part of 3.8.0rc1.

Yes, so people can see what's different since beta2. Happy for you to reorganise before final release.

Issue #1595, opened 2021-05-01
Prepare Release 3.8.0


hoffie:
**Target timeline**
Scheduled feature freeze / Start of translation process: 2021-05-03
Targeted translation completion date: 2021-05-15
Approximate release date: 2021-06-02
Current state: Code freeze (rc) 

**Checklist**
Note: Please discuss potential wishes for changes in processes in the generic #1594 PR for this (future) issue template so that we can keep this issue focused on 3.8.0.

- [x] Assign this issue to the release shepherd who is in charge of managing this checklist
- [x] Create a milestone and add all relevant issues
- [x] Ensure that all issues/PR targeted for this release are done by checking the Project board with an appropriate filter (e.g. [`milestone:"Release 3.8.0"`](https://github.com/orgs/jamulussoftware/projects/2?card_filter_query=milestone%3A"release+3.8.0")). Reminder main developers to review entries in *Waiting on team* state. De-tag unfinished Issues/PRs.
- [x] Declare a code freeze by updating this Issue and adding a comment. PRs can still be worked on and may get reviewed, but must not be merged unless agreed explicitly.
- [x] Check ./Jamulus -h output against the en-Command-Line-Options.md wiki page and update if necessary. (@gilgongo, @pljones)
- [x] Write the Changelog (@hoffie) #1600
- [x] Start Website translations
  - [x] Create new branch `translate3_y_z` based on branch `release`
  - [x] Squash-merge `changes` into `translate3_y_z`
  - [x] Ping translators by creating an Issue for each language. Reminders for the Issue content: Note the deadline. Translators should open PRs with their changes against the `translate3_y_z` branch (?),  use a good commit message ("Add German translations for 3.y.z") and link to their respective Issue.
- [x] Prepare App translations
  - [x] Generate `.ts` files in master via `lupdate`
  - [x] Get the list of translators from src/utils.cpp (some grep/sed would be nice; last one per language is the active one).
  - [x] 3.8.0-specific: Tell translators about the Central->Directory change
  - [x] Create a translation Issue for each language, note the deadline, assign to the designated translator, add the Project, add the Milestone. Translators do their work using linguist and submit PRs with their changes against master. 
- [x] [Tag a beta release](https://github.com/jamulussoftware/jamulus/blob/master/RELEASE-PROCESS.md#steps-for-a-specific-release) (@softins)
- [x] Announce the beta release on Github Discussions. Pin the thread.
- [x] Get feedback on the stability of the beta release
[...]

PR #1708, opened 2021-05-17
update translate fr for 3.8.0

jujudusud:
To be merged for 3.8.0 release.

### DIFF
--- a/src/res/translation/translation_fr_FR.ts
+++ b/src/res/translation/translation_fr_FR.ts
@@ -1367,17 +1367,17 @@ Nous avons coupé votre canal et activé &quot;Me silencer&quot;. Veuillez d&apo
     <message>
         <location filename="../../clientsettingsdlg.cpp" line="250"/>
         <source>Language</source>
-        <translation type="unfinished">Langue</translation>
+        <translation>Langue</translation>
     </message>
     <message>
         <location filename="../../clientsettingsdlg.cpp" line="250"/>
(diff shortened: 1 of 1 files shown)


softins (approved):
Thanks!

hoffie (approved):
The PR looks good, thanks.

@softins @pljones I'm unsure how to go forward with this though.
- We could do an rc2 quickly. Not sure if it's worth the overhead.
- We could merge and go on without another rc (and have some low risk of breakage -- might be mitigated if this PR's artifacts are tested properly). If we go with that route, should we ping other translators on Discord quickly?

Therefore, I'm not merging yet.

softins:
I don't think it needs another rc. I think the risk of merging this before the final release is very Low, and it's just adding translations for the few strings we were happy to leave untranslated.

But I don't think it's worth explicitly asking the other translators.

hoffie:
> I don't think it needs another rc. I think the risk of merging this before the final release is very low, so I would include it. It's just adding translations for the few strings we were happy to leave untranslated.

Ok. Will add that to the release checklist to clarify what kind of changes are still ok in rc phase.

> But I don't think it's worth explicitly asking the other translators. Maybe just give them the opportunity if they want to.

I agree. I have added a note (without explicit pinging) in https://github.com/jamulussoftware/jamulus/issues/1595#issuecomment-843645315.

jujudusud:
Thanks.

PR #1714, opened 2021-05-18
Update translation nl for 3.8.0

henkdegroot:
Please merge for 3.8.0 release

### DIFF
--- a/src/res/translation/translation_nl_NL.ts
+++ b/src/res/translation/translation_nl_NL.ts
@@ -239,7 +239,7 @@
         <location filename="../../audiomixerboard.cpp" line="248"/>
         <location filename="../../audiomixerboard.cpp" line="264"/>
         <source>Pan</source>
-        <translation>Balans</translation>
+        <translation>Bal.</translation>
     </message>
     <message>
         <location filename="../../audiomixerboard.cpp" line="48"/>
@@ -414,7 +414,7 @@
     <message>
         <location filename="../../audiomixerboard.cpp" line="233"/>
         <source>PAN</source>
-        <translation>BAL</translation>
+        <translation>BAL.</translation>
     </message>
     <message>
         <location filename="../../audiomixerboard.cpp" line="234"/>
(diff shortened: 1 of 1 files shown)


henkdegroot:
I noticed fr update was merged....so hoping you will accept this one as well.

ann0see (approved):
Looks good to me (just scanned it).

softins (approved):
Looks good, thanks!

src/res/translation/translation_nl_NL.ts
```diff
@@ -239,7 +239,7 @@
         <location filename="../../audiomixerboard.cpp" line="248"/>
         <location filename="../../audiomixerboard.cpp" line="264"/>
         <source>Pan</source>
-        <translation>Balans</translation>
+        <translation>Bal.</translation>
```
hoffie, line 242: I vaguely remember that this was changed the other way round recently. Maybe some agreement is needed to keep it one way. :)

src/res/translation/translation_nl_NL.ts
```diff
@@ -414,7 +414,7 @@
     <message>
         <location filename="../../audiomixerboard.cpp" line="233"/>
         <source>PAN</source>
-        <translation>BAL</translation>
+        <translation>BAL.</translation>
```
hoffie, line 417: Same as above.

hoffie (approved):

src/res/translation/translation_nl_NL.ts
henkdegroot: Correct, it was changed the other way around recently, but because of the length of the text it is causing shifting the screen when the muted speaker icon is displayed in compact skin mode. Therefore going back to the abbreviated text.

src/res/translation/translation_nl_NL.ts
henkdegroot: Correct here as well, I had second thoughts about this and the . (while it does not look nice in the GUI) should be there as it is an abbrevation and this is in-line with the text when showed in normal and compact skin.

ann0see:
Ready to merge this one?

henkdegroot:
@ann0see Believe it is ready. Not sure if @hoffie agrees as he had 2 questions. Not sure if he gets notifications from this.

Issue #1705, opened 2021-05-17
Build option "jackonwindows" not working with JACK2


henkdegroot:
**Describe the bug**
When compiling Jamulus with the "jackonwindows" option, the project file on Win32 is checking the availability of JACK and links the libraries required. However in this check, the old JACK1 folder location is used for windows. This results in a compiled version which only works with JACK1 and not with the more common/current version JACK2.

**To Reproduce**
Add the CONFIG += jackonwindows option to the Jamulus.pro file and compile the release.
This only succeeds when JACK1 is installed and fails when JACK2 is used.

**Expected behavior**
As JACK2 is the current standard, the Jamulus.pro file should be updated for the Win32 builds to use the JACK2 folder locations.

**Operating system**
Windows 10
JACK v1.9.11 64bit
JACK2 v1.9.18 64bit

**Version of Jamulus**
v3.8.0beta2

jujudusud:
JACK version 1.9.11 is also JACK2.

henkdegroot:
> JACK version 1.9.11 is also JACK2.

Interesting, I always assumed it wass jack because the name of the installer is JACK... and the folder it is installed into for windows is named Jack as well. While later versions have the installer named JACK2... and the folder it is installed into for windows is named Jack2 as well.

At least that explains why the compile was succesfull by only updating the references to the new libraries.
Still believe we should update the Jamulus.pro file to point to the correct (current) folders used by jack on windows.
Do you agree?

JACK v1.9.11 64bit: Installs into: C:\Program Files (x86)\JACK
JACK v1.9.18 64bit: Installs into: C:\Program Files\JACK2
JACK v1.9.18 32bit: Installs into: C:\Program Files (x86)\JACK2

jujudusud:
My opinion is as follows:
Jamulus should check if JACK is at the latest version (1.9.18)
If not advise the user to update.
;-)

henkdegroot:
> For this change of install directory, We should Ask upstream if it is a bug or not

Okay, you are sort of agreeing to this....as you state it should check the latest version is installed (not sure if that is possible though in the pro file) and as the latest version installs into a different folder, we should at update the win32 of the pro file.

@hoffie , as I know you have commented on #1678, can you please share your opinion/view on this?

My intend is to update the pro file and add two sections, one for 64 bit and one for 32 bit and make sure the path location check is correct and the INCLUDEPATH and LIBS is set to the correct folders and files.

[...]

PR #1718, opened 2021-05-18
Jamulus.pro jackonwindows support improvements

jujudusud:
#1705 
Work in progress -> Jamulus.pro for windows 10 64 bits + JACK on Windows build.
Need help to make the magic append.

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -94,17 +94,17 @@ win32 {
     # replace ASIO with jack if requested
     contains(CONFIG, "jackonwindows") {
         contains(QT_ARCH, "i386") {
-            !exist("C:/Program Files (x86)") {
-                message ("Cross compilation build")
+            !exists("C:/Program Files (x86)") {
+                message("Cross compilation build")
                 programfilesdir = "C:/Program Files (x86)"
-                !exists("$${"programfilesdir"}/JACK2/include/jack/jack.h") {
+                !exists("$${programfilesdir}/JACK2/include/jack/jack.h") {
                     message("Warning: jack.h was not found at the usual place, maybe JACK2 32 bits is not installed to cross compile.")
                 }
             }
-            !exist("C:/Program Files") {
+            !exists("C:/Program Files") {
                 message("Native i386 build")
                 programfilesdir = "C:/Program Files"
-                !exists("$${"programfilesdir"}/JACK2/include/jack/jack.h") {
+                !exists("$${programfilesdir}/JACK2/include/jack/jack.h") {
                     message("Warning: jack.h was not found at the usual place, maybe JACK2 is not installed.")
                 }
             }
@@ -113,7 +113,7 @@ win32 {
             message("Native x86_64 build")
             programfilesdir = "C:/Program Files"
             libjackname = "libjack64.lib"
-            !exists("$${"programfilesdir"}/JACK2/include/jack/jack.h") {
+            !exists("$${programfilesdir}/JACK2/include/jack/jack.h") {
                 message("Warning: jack.h was not found at the usual place, maybe JACK2 is not installed.")
             }
         }
@@ -125,8 +125,8 @@ win32 {
         DEFINES += WITH_JACK
         DEFINES += JACK_REPLACES_ASIO
         DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
-        INCLUDEPATH += "$${"programfilesdir"}/JACK2/include"
-        LIBS += "$${"programfilesdir"}/JACK2/lib/$${libjackname}"
+        INCLUDEPATH += "$${programfilesdir}/JACK2/include"
+        LIBS += "$${programfilesdir}/JACK2/lib/$${libjackname}"
     }
 
 } else:macx {
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -94,7 +94,7 @@ win32 {
     # replace ASIO with jack if requested
     contains(CONFIG, "jackonwindows") {
         contains(QT_ARCH, "i386") {
-            !exists("C:/Program Files (x86)") {
+            exists("C:/Program Files (x86)") {
                 message("Cross compilation build")
                 programfilesdir = "C:/Program Files (x86)"
             } else {
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -99,12 +99,12 @@ win32 {
         } else {
             message("Native x86_64 build")
             libjackname = libjack64.lib
-            !exists("C:/Program Files (86)/JACK2/include/jack/jack.h") {
+            !exists("C:/Program Files (x86)/JACK2/include/jack/jack.h") {
                 message("Warning: jack.h was not found at the usual place, install JACK2 32 bits to cross compile.")
             }
         }
           !exists("C:/Program Files/JACK2/include/jack/jack.h") {
-            message("Warning: jack.h was not found at the usual place, maybe JACK2 64 bits to compile")
+            message("Warning: jack.h was not found at the usual place, maybe JACK2 is not installed")
           }
 
             HEADERS -= windows/sound.h
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -95,65 +95,39 @@ win32 {
     contains(CONFIG, "jackonwindows") {
         message(Using Jack instead of ASIO.)
 
-        !exists("C:/Program Files (x86)/Jack/includes/jack/jack.h") {
-            message(Warning: jack.h was not found at the usual place, maybe jack is not installed)
+        contains(QT_ARCH, "x86") {
+		    !exists("C:/Program Files (x86)/JACK2/includes/jack/jack.h") {
+                message(Warning: jack.h was not found at the usual place, maybe jack2 32-bit is not installed)
+            }
+
+            HEADERS -= windows/sound.h
+            SOURCES -= windows/sound.cpp
+            HEADERS += linux/sound.h
+            SOURCES += linux/sound.cpp
+            DEFINES += WITH_JACK
+            DEFINES += JACK_REPLACES_ASIO
+            DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
+            INCLUDEPATH += "C:/Program Files (x86)/JACK2/include"
+            LIBS += "C:/Program Files (x86)/JACK2/lib/libjack.lib"
         }
 
-        HEADERS -= windows/sound.h
-        SOURCES -= windows/sound.cpp
-        HEADERS += linux/sound.h
-        SOURCES += linux/sound.cpp
-        DEFINES += WITH_JACK
-        DEFINES += JACK_REPLACES_ASIO
-        DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
-        INCLUDEPATH += "C:/Program Files (x86)/Jack/includes"
-        LIBS += "C:/Program Files (x86)/Jack/lib/libjack64.lib"
-    }
-} else:win64 {
-    DEFINES -= UNICODE # fixes issue with ASIO SDK (asiolist.cpp is not unicode compatible)
-    DEFINES += NOMINMAX # solves a compiler error in qdatetime.h (Qt5)
-    HEADERS += windows/sound.h
-    SOURCES += windows/sound.cpp \
-        windows/ASIOSDK2/common/asio.cpp \
-        windows/ASIOSDK2/host/asiodrivers.cpp \
-        windows/ASIOSDK2/host/pc/asiolist.cpp
-    RC_FILE = windows/mainicon.rc
-    INCLUDEPATH += windows/ASIOSDK2/common \
-        windows/ASIOSDK2/host \
-        windows/ASIOSDK2/host/pc
-    mingw* {
-        LIBS += -lole32 \
-            -luser32 \
-            -ladvapi32 \
-            -lwinmm \
-            -lws2_32
-    } else {
-        QMAKE_LFLAGS += /DYNAMICBASE:NO # fixes crash with libjack64.dll, see https://github.com/jamulussoftware/jamulus/issues/93
-        LIBS += ole32.lib \
-            user32.lib \
-            advapi32.lib \
-            winmm.lib \
-            ws2_32.lib
-    }
+        contains(QT_ARCH, "x86_64") {
+		    !exists("C:/Program Files/JACK2/include/jack/jack.h") {
+                message(Warning: jack.h was not found at the usual place, maybe jack2 64-bit is not installed)
+            }
 
-    # replace ASIO with jack if requested
-    contains(CONFIG, "jackonwindows") {
-        message(Using Jack instead of ASIO.)
-
-        !exists("C:/Program Files/JACK2/includes/jack/jack.h") {
-            message(Warning: jack.h was not found at the usual place, maybe jack is not installed)
+            HEADERS -= windows/sound.h
+            SOURCES -= windows/sound.cpp
+            HEADERS += linux/sound.h
+            SOURCES += linux/sound.cpp
+            DEFINES += WITH_JACK
+            DEFINES += JACK_REPLACES_ASIO
+            DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
+            INCLUDEPATH += "C:/Program Files/JACK2/include"
+            LIBS += "C:/Program Files/JACK2/lib/libjack64.lib"
         }
+    }
 
-        HEADERS -= windows/sound.h
-        SOURCES -= windows/sound.cpp
-        HEADERS += linux/sound.h
-        SOURCES += linux/sound.cpp
-        DEFINES += WITH_JACK
-        DEFINES += JACK_REPLACES_ASIO
-        DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
-        INCLUDEPATH += "C:/Program Files/JACK2/includes"
-        LIBS += "C:/Program Files/JACK2/lib/libjack64.lib"	
-	}
 } else:macx {
     contains(CONFIG, "server_bundle") {
         message(The generated application bundle will run a server instance.)
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -94,19 +94,19 @@ win32 {
     # replace ASIO with jack if requested
     contains(CONFIG, "jackonwindows") {
         contains(QT_ARCH, "i386") {
-        	message("Native i386 build")
-        	libjackname = libjack.lib
+            message("Native i386 build")
+            libjackname = libjack.lib
         } else {
-        	message("Native x86_64 build")
-		    libjackname = libjack64.lib
-		    !exists("C:/Program Files (86)/JACK2/include/jack/jack.h") {
-            	message("Warning: jack.h was not found at the usual place, install JACK2 32 bits to cross compile.")
-         	}	
+            message("Native x86_64 build")
+            libjackname = libjack64.lib
+            !exists("C:/Program Files (86)/JACK2/include/jack/jack.h") {
+                message("Warning: jack.h was not found at the usual place, install JACK2 32 bits to cross compile.")
+            }
         }
           !exists("C:/Program Files/JACK2/include/jack/jack.h") {
-         	 message("Warning: jack.h was not found at the usual place, maybe JACK2 64 bits to compile")
+            message("Warning: jack.h was not found at the usual place, maybe JACK2 64 bits to compile")
           }
-    
+
             HEADERS -= windows/sound.h
             SOURCES -= windows/sound.cpp
             HEADERS += linux/sound.h
@@ -116,7 +116,7 @@ win32 {
             DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
             INCLUDEPATH += "C:/Program Files/JACK2/include"
             LIBS += "C:/Program Files/JACK2/lib/$${libjackname}"
-	}
+    }
 
 } else:macx {
     contains(CONFIG, "server_bundle") {
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -97,25 +97,18 @@ win32 {
             !exists("C:/Program Files (x86)") {
                 message("Cross compilation build")
                 programfilesdir = "C:/Program Files (x86)"
-                !exists("$${programfilesdir}/JACK2/include/jack/jack.h") {
-                    message("Warning: jack.h was not found at the usual place, maybe JACK2 32 bits is not installed to cross compile.")
-                }
-            }
-            !exists("C:/Program Files") {
+            } else {
                 message("Native i386 build")
                 programfilesdir = "C:/Program Files"
-                !exists("$${programfilesdir}/JACK2/include/jack/jack.h") {
-                    message("Warning: jack.h was not found at the usual place, maybe JACK2 is not installed.")
-                }
             }
             libjackname = "libjack.lib"
         } else {
             message("Native x86_64 build")
             programfilesdir = "C:/Program Files"
             libjackname = "libjack64.lib"
-            !exists("$${programfilesdir}/JACK2/include/jack/jack.h") {
-                message("Warning: jack.h was not found at the usual place, maybe JACK2 is not installed.")
-            }
+        }
+        !exists("$${programfilesdir}/JACK2/include/jack/jack.h") {
+            message("Warning: jack.h was not found in the expected location ($${programfilesdir}). Ensure that the right JACK2 variant is installed (32bit vs. 64bit).")
         }
 
         HEADERS -= windows/sound.h
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -94,28 +94,39 @@ win32 {
     # replace ASIO with jack if requested
     contains(CONFIG, "jackonwindows") {
         contains(QT_ARCH, "i386") {
-            message("Native i386 build")
-            libjackname = libjack.lib
+            !exist("C:/Program Files (x86)") {
+                message ("Cross compilation build")
+                programfilesdir = "C:/Program Files (x86)"
+                !exists("$${"programfilesdir"}/JACK2/include/jack/jack.h") {
+                    message("Warning: jack.h was not found at the usual place, maybe JACK2 32 bits is not installed to cross compile.")
+                }
+            }
+            !exist("C:/Program Files") {
+                message("Native i386 build")
+                programfilesdir = "C:/Program Files"
+                !exists("$${"programfilesdir"}/JACK2/include/jack/jack.h") {
+                    message("Warning: jack.h was not found at the usual place, maybe JACK2 is not installed.")
+                }
+            }
+            libjackname = "libjack.lib"
         } else {
             message("Native x86_64 build")
-            libjackname = libjack64.lib
-            !exists("C:/Program Files (x86)/JACK2/include/jack/jack.h") {
-                message("Warning: jack.h was not found at the usual place, install JACK2 32 bits to cross compile.")
+            programfilesdir = "C:/Program Files"
+            libjackname = "libjack64.lib"
+            !exists("$${"programfilesdir"}/JACK2/include/jack/jack.h") {
+                message("Warning: jack.h was not found at the usual place, maybe JACK2 is not installed.")
             }
         }
-          !exists("C:/Program Files/JACK2/include/jack/jack.h") {
-            message("Warning: jack.h was not found at the usual place, maybe JACK2 is not installed")
-          }
 
-            HEADERS -= windows/sound.h
-            SOURCES -= windows/sound.cpp
-            HEADERS += linux/sound.h
-            SOURCES += linux/sound.cpp
-            DEFINES += WITH_JACK
-            DEFINES += JACK_REPLACES_ASIO
-            DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
-            INCLUDEPATH += "C:/Program Files/JACK2/include"
-            LIBS += "C:/Program Files/JACK2/lib/$${libjackname}"
+        HEADERS -= windows/sound.h
+        SOURCES -= windows/sound.cpp
+        HEADERS += linux/sound.h
+        SOURCES += linux/sound.cpp
+        DEFINES += WITH_JACK
+        DEFINES += JACK_REPLACES_ASIO
+        DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
+        INCLUDEPATH += "$${"programfilesdir"}/JACK2/include"
+        LIBS += "$${"programfilesdir"}/JACK2/lib/$${libjackname}"
     }
 
 } else:macx {
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -93,28 +93,19 @@ win32 {
 
     # replace ASIO with jack if requested
     contains(CONFIG, "jackonwindows") {
-        message(Using Jack instead of ASIO.)
-
-        contains(QT_ARCH, "x86") {
-		    !exists("C:/Program Files (x86)/JACK2/includes/jack/jack.h") {
-                message(Warning: jack.h was not found at the usual place, maybe jack2 32-bit is not installed)
-            }
-
-            HEADERS -= windows/sound.h
-            SOURCES -= windows/sound.cpp
-            HEADERS += linux/sound.h
-            SOURCES += linux/sound.cpp
-            DEFINES += WITH_JACK
-            DEFINES += JACK_REPLACES_ASIO
-            DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
-            INCLUDEPATH += "C:/Program Files (x86)/JACK2/include"
-            LIBS += "C:/Program Files (x86)/JACK2/lib/libjack.lib"
-        }
-
-        contains(QT_ARCH, "x86_64") {
-		    !exists("C:/Program Files/JACK2/include/jack/jack.h") {
-                message(Warning: jack.h was not found at the usual place, maybe jack2 64-bit is not installed)
-            }
+        contains(QT_ARCH, "i386") {
+        	message("Native i386 build")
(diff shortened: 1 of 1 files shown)


Jamulus.pro
```diff
@@ -109,6 +109,51 @@ win32 {
         INCLUDEPATH += "C:/Program Files (x86)/Jack/includes"
         LIBS += "C:/Program Files (x86)/Jack/lib/libjack64.lib"
     }
+} else:win64 {
```
hoffie, line 112: Is there an actual `win64` target?

Can't just use a `contains(QT_ARCH, x86_64)` conditional within the existing `win32` target? This should simplify the logic a lot and should save duplication if the sole intention is to change the jack path.

Jamulus.pro
```diff
@@ -109,6 +109,51 @@ win32 {
...
+            advapi32.lib \
+            winmm.lib \
+            ws2_32.lib
+    }
+
+    # replace ASIO with jack if requested
+    contains(CONFIG, "jackonwindows") {
+        message(Using Jack instead of ASIO.)
+
+        !exists("C:/Program Files/JACK2/includes/jack/jack.h") {
```
hoffie, line 143: If possible, it may make sense to define some JACK_BASE_PATH variable and derive all other paths from that?

Jamulus.pro
henkdegroot: @jujudusud , please note that the path for the jack include folder is now named include (without the s).
When I build the version locally with jackonwindows, I used the contains(QT_ARCH, x86_x64) which I believe saves duplicating/maintaining extra lines. And perhaps the message prompted that jack is not installed could change to "Warning: jack.h was not found at the usual place, maybe jack is not installed or it is not the latest version"

Thanks for doing this...I could have done it as well, but was waiting for feedback on the issue regarding the approach to take.

henkdegroot:
@jujudusud  here is the Jamulus.pro file I have used to build. Please note this contains the "jackonwindows" option in the CONFIG section, so only use the jack part in the win32 section please.
[Jamulus.pro.txt](https://github.com/jamulussoftware/jamulus/files/6506204/Jamulus.pro.txt)

Jamulus.pro
jujudusud: Now using `win32` target and `contains(QT_ARCH, x86_64)` conditional.
Thanks for reviewing and helping.

Jamulus.pro
npostavs: I think it would work like this:
```
        contains(QT_ARCH, "x86") {
            jack_dir = "C:/Program Files (x86)/Jack"
        } else {
            jack_dir = "C:/Program Files/Jack"
        }

        !exists($${jack_dir}/includes/jack/jack.h) {
            message(Warning: jack.h was not found at the usual place, maybe jack is not installed)
        }
```

```
        INCLUDEPATH += $${jack_dir}/includes"
        LIBS += $${jack_dir}/lib/libjack64.lib
```

jujudusud: do you think it is possible to use Windows environment variable instead of the hard coded one ?
Something like : `%ProgramFiles%` or the `ProgramFilesDir` key used in registry ?

npostavs: To access `%ProgramFiles(x86)%` we would need [`getenv`](https://doc.qt.io/qt-5/qmake-function-reference.html#getenv-variablename), which requires Qt 5.12 (might be okay, I think there's already a proposal to bump min Qt version for Mac).

```
        contains(QT_ARCH, "x86") {
            jack_dir = $$getenv("ProgramFiles(x86)")/Jack
        } else {
            jack_dir = $$getenv("ProgramFiles")/Jack
        }
```

henkdegroot: Autobuild script for Windows is using Qt 5.15.2 already as well, so it should not be a problem to use getenv.

Jamulus.pro
```diff
@@ -95,20 +95,39 @@ win32 {
     contains(CONFIG, "jackonwindows") {
         message(Using Jack instead of ASIO.)
 
-        !exists("C:/Program Files (x86)/Jack/includes/jack/jack.h") {
-            message(Warning: jack.h was not found at the usual place, maybe jack is not installed)
+        contains(QT_ARCH, "x86") {
```
pljones, line 98: If I install Windows 32bit, isn't "Program Files" where 32bit programs live?

I thought "Program Files (x86)" only existed on Windows 64bit systems (where "Program Files" is for 64bit programs and "Programs Files (x86)" for 32bit).

Jamulus.pro
```diff
@@ -95,20 +95,39 @@ win32 {
...
+		    !exists("C:/Program Files (x86)/JACK2/includes/jack/jack.h") {
+                message(Warning: jack.h was not found at the usual place, maybe jack2 32-bit is not installed)
+            }
+
+            HEADERS -= windows/sound.h
```
pljones, line 103: Get rid of this duplication.

Jamulus.pro
hoffie: > which requires Qt 5.12 (might be okay, I think there's already a proposal to bump min Qt version for Mac).

I think we only agreed to bump the Qt version for the official builds. We did not drop support for building with older versions of Qt, as far as I know.

Jamulus.pro
jujudusud: I don't know. If anyone can say something about that. 
(jujudusud is thinking of Microsoft just deprecated 32 bits windows 10...)
Anyway, If this is wrong I will change this line. Thanks for the review !

Jamulus.pro
jujudusud: To get rid of this duplication, I will also have to use a `lib-name` variable. This variable should return :

- libjack64.lib for x86_64,
- libjack.lib for x86.

So ... Will do when it's ok in my head.

Jamulus.pro
henkdegroot: > If I install Windows 32bit, isn't "Program Files" where 32bit programs live?
> 
> I thought "Program Files (x86)" only existed on Windows 64bit systems (where "Program Files" is for 64bit programs and "Programs Files (x86)" for 32bit).

That is correct, as long as you compile/build on the solution on a 64-bit system then this should not be an issue.

Jamulus.pro
henkdegroot: @jujudusud the line:  !exists("C:/Program Files/JACK2/includes/jack/jack.h")
is incorrect referring to /includes/ here...it should be /include/  

Reason I did not notice is because the "message" function is throwing an error.

 message(Warning: jack.h was not found at the usual place, maybe jack is not installed)

This is generating an error due to the , (comma) in the text. The error says that message requires one argument.
Either replace the , by another character or use "" around the text in the message.
Also as it is just a "message" it does not even stop the build process.

This "message" issue also appears in the "macx" section. Perhaps this can be adapted as well? Assuming the title of the PR should change in that case.

Jamulus.pro
pljones: If you're building Jamulus x64, then we want JACK x64 (which will be under "Program Files" on x64).  QT_ARCH won't be "x86".

The main reason anyone would be building x86 would be if they were building on x86 and want JACK x86 (which will be under "Program Files" on x86).  QT_ARCH will be "x86".

The only time to use "Program Files (x86)", then, is if cross-compiling from x64 to x86.  However, QT_ARCH alone doesn't tell you this, from what I can see?

pljones: So...

If QT_ARCH = x86
* If  "Program Files (x86)" exists, it's a cross-compile - set the install dir to "Program Files (x86)"
* Otherwise set the install dir to "Program Files"
* Set the x86-specific bits

Else
* Set the install dir to "Program Files"
* Set the x64-specific bits

Check the install is sane.

henkdegroot: @pljones when you build the instapper for Windows, it includes both the x86 and x86_x64 application in it. When installing this the appropiate version for your pc will be installen. So you need to be able to have both options at the same time.

Jamulus.pro
jujudusud: What about 'read_registry(tree, key[, flag])' instead of getenv ?
'read_registry(HKLM, ProgramFilesDir[, 64])'
Should return the good path for 64bits.

henkdegroot: My view is...why make it so complex? As JACK might even have been installed somewhere else....I would say it is a slight overkill to include all this. Not that is isn't nice to have it...

hoffie: > My view is...why make it so complex? As JACK might even have been installed somewhere else..

I agree. The compromise I'd aim for would be: Make sure the default works and provide a way to override for builds against different paths (`qmake JACK_INSTALL_DIR=/path/to/jack` -- maybe there are existing conventions what the variable should be named).

henkdegroot: > qmake JACK_INSTALL_DIR=/path/to/jack

That would also be beneficial in case we want to setup an automated build. Not sure how such script would install jack...but with the variable it could become an option to consider.

jujudusud: this is the use of read registry
`    # replace ASIO with jack if requested
    contains(CONFIG, "jackonwindows") {
        message(Using Jack instead of ASIO.)

        contains(QT_ARCH, "x86") {
        	ProgramFilesDir = read_registry(HKLM, ProgramFilesDir[, 32])
		    !exists("$${ProgramFilesDir}/JACK2/include/jack/jack.h") {
                message(Warning: "jack.h was not found at the usual place, maybe JACK2 is not installed")
            }

            HEADERS -= windows/sound.h
            SOURCES -= windows/sound.cpp
            HEADERS += linux/sound.h
            SOURCES += linux/sound.cpp
            DEFINES += WITH_JACK
            DEFINES += JACK_REPLACES_ASIO
            DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
            INCLUDEPATH += "$${ProgramFilesDir}/JACK2/include"
            LIBS += "$${ProgramFilesDir}/JACK2/lib/libjack.lib"
        }

        contains(QT_ARCH, "x86_64") {
		    ProgramFilesDir = read_registry(HKLM, ProgramFilesDir[, 64])
		    !exists("$${ProgramFilesDir}/JACK2/include/jack/jack.h") {
                message(Warning: "jack.h was not found at the usual place, maybe JACK2 64-bit is not installed")
            }

            HEADERS -= windows/sound.h
            SOURCES -= windows/sound.cpp
            HEADERS += linux/sound.h
            SOURCES += linux/sound.cpp
            DEFINES += WITH_JACK
            DEFINES += JACK_REPLACES_ASIO
            DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
            INCLUDEPATH += "$${ProgramFilesDir}/JACK2/include"
            LIBS += "$${ProgramFilesDir}/JACK2/lib/libjack64.lib"
        }`

Jamulus.pro
jujudusud: `        contains(QT_ARCH, "x86") {
        	ProgramFilesDir = read_registry(HKLM, ProgramFilesDir[, 32])
        	libjack-name = libjack.lib
		    !exists("$${ProgramFilesDir}/JACK2/include/jack/jack.h") {
                message(Warning: "jack.h was not found at the usual place, maybe JACK2 is not installed")
            }

            HEADERS -= windows/sound.h
            SOURCES -= windows/sound.cpp
            HEADERS += linux/sound.h
            SOURCES += linux/sound.cpp
            DEFINES += WITH_JACK
            DEFINES += JACK_REPLACES_ASIO
            DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
            INCLUDEPATH += "$${ProgramFilesDir}/JACK2/include"
            LIBS += "$${ProgramFilesDir}/JACK2/lib/$${libjack-name}"
        }

        contains(QT_ARCH, "x86_64") {
		    ProgramFilesDir = read_registry(HKLM, ProgramFilesDir[, 64])
		    libjack-name = libjack64.lib
		    !exists("$${ProgramFilesDir}/JACK2/include/jack/jack.h") {
                message(Warning: "jack.h was not found at the usual place, maybe JACK2 64-bit is not installed")
            }

            HEADERS -= windows/sound.h
            SOURCES -= windows/sound.cpp
            HEADERS += linux/sound.h
            SOURCES += linux/sound.cpp
            DEFINES += WITH_JACK
            DEFINES += JACK_REPLACES_ASIO
            DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
            INCLUDEPATH += "$${ProgramFilesDir}/JACK2/include"
            LIBS += "$${ProgramFilesDir}/JACK2/lib/$${libjack-name}"
        }`

Jamulus.pro
jujudusud: I don't how to code this into the Jamulus.pro file.

Jamulus.pro
jujudusud: ok ! think  it will be ok with 
c0a5110

Jamulus.pro
jujudusud: c0a5110

Jamulus.pro
jujudusud: Ok with :
c0a5110

pljones:
Maybe something like this would work - not tried it:
```
contains(CONFIG, "jackonwindows") {
  contains(QT_ARCH, "x86") {
    !exists("C:/Program Files (x86)/") {
      message(Native x86 build)
  	ProgramFilesDir = "C:/Program Files"
    } else {
      message(Cross-compile x86 build)
  	ProgramFilesDir = "C:/Program Files (x86)"
    }
    libjack-name = libjack.lib
  } else {
    libjack-name = libjack64.lib
  }
  !exists("$${ProgramFilesDir}/JACK2/include/jack/jack.h") {
    message("Warning: jack.h was not found at the usual place, maybe JACK2 is not installed.  Reverting to ASIO.")
	CONFIG -= "jackonwindows"
  } else {
    message(Using Jack instead of ASIO.)

    HEADERS -= windows/sound.h
    SOURCES -= windows/sound.cpp
    HEADERS += linux/sound.h
    SOURCES += linux/sound.cpp
    DEFINES += WITH_JACK
    DEFINES += JACK_REPLACES_ASIO
    DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
    INCLUDEPATH += "$${ProgramFilesDir}/JACK2/include"
    LIBS += "$${ProgramFilesDir}/JACK2/lib/$${libjack-name}"
  }
}
```

ChangeLog
```diff
@@ -38,8 +38,6 @@
   (contributed by @vimpostor)
 
 - The term "Central server" has been replaced with "Directory server" (#1407).
-  Note that the program will still accept the --centralserver option for backward
-  compatibility with existing system startup scripts.
```
hoffie, line 42: Hrm, might be an indicator for a partially bad merge/rebase?

Jamulus.pro
```diff
@@ -95,20 +95,28 @@ win32 {
...
+        	ProgramFilesDir = read_registry(HKLM, ProgramFilesDir[, 32])
+        	libjack-name = libjack.lib
+        } else {
+            ProgramFilesDir = read_registry(HKLM, ProgramFilesDir[, 64])
+		    libjack-name = libjack64.lib	
+          }
+          !exists("$${ProgramFilesDir}/JACK2/include/jack/jack.h") {
+            message("Warning: jack.h was not found at the usual place, maybe JACK2 is not installed")
+          }
+
```
hoffie, line 108: Indentation seems to be off here and further below.

src/res/translation/translation_fr_FR.ts
```diff
@@ -1367,17 +1367,17 @@ Nous avons coupé votre canal et activé &quot;Me silencer&quot;. Veuillez d&apo
     <message>
         <location filename="../../clientsettingsdlg.cpp" line="250"/>
         <source>Language</source>
-        <translation>Langue</translation>
+        <translation type="unfinished">Langue</translation>
```
hoffie, line 1370: The .ts files should not show up in this PR. The PR seems to revert those files to an earlier status.

Jamulus.pro
```diff
@@ -95,20 +95,28 @@ win32 {
     contains(CONFIG, "jackonwindows") {
         message(Using Jack instead of ASIO.)
 
-        !exists("C:/Program Files (x86)/Jack/includes/jack/jack.h") {
-            message(Warning: jack.h was not found at the usual place, maybe jack is not installed)
-        }
+        contains(QT_ARCH, "x86") {
```
hoffie, line 98: Has this been tested?
I'd assume that `contains(x86_64, x86)` and `contains(x86, x86)` will both be true?
[`equals()`](https://doc.qt.io/qt-5/qmake-test-function-reference.html#equals-variablename-value) looks more correct to me for this use case.

Jamulus.pro
```diff
@@ -95,20 +95,28 @@ win32 {
...
+        	ProgramFilesDir = read_registry(HKLM, ProgramFilesDir[, 32])
```
hoffie, line 99: As @henkdegroot mentioned, `read_registry` seems to be a [fairly recent Qt addition](https://doc.qt.io/qt-5/qmake-function-reference.html#read-registry-tree-key-flag). Adding this function call will likely break things for pre-5.12.1 Qt users.

Is this really necessary? It's a trade-off between compatibility with earlier Qt and covering even more cases. The existing code uses hardcoded paths as well and nobody seems to have complained, so I'd rather suggest to keep it simple and avoid introducing downsides by doing this?

hoffie (changes requested):
- Can you update the PR title to cover the fact that it re-works jack path handling?
- There seem to be git merge/rebase problems. If we squash-merge later, history within this PR should not be too important, so you could probably fix it like this:
  - `git fetch upstream` to ensure that you've got the latest master commits in your clone
  - `git checkout upstream/master Changelog src/res/translation/` to revert the modified files' content to what's currently in master
  - `git commit -m "Fix unintended master reverts" ,` to commit those reverts.
-  (Possible) break of compatibility with earlier Qt would be good to avoid.

Jamulus.pro
henkdegroot: > I'd assume that `contains(x86_64, x86)` and `contains(x86, x86)` will both be true?

Interesting spot! I have checked, and while it was working on my system when I compiled it...it sounds as if it was just working by luck.

When the build is performed on windows for 64-bit, the QT_ARCH is set to "x86_64".
When the build is performed on windows for 32-bit, the QT_ARCH is set to "i386"

So the jamulus.pro needs updating for this.

jujudusud: What about Windows 64 bits in 32 bits mode ? WOW64 let the user work in 32 bits mode. Is the QT_ARCH set to 'x86_64' in this case ?

That's why I wanna add a target first with win64 label... ;-)

Jamulus.pro
jujudusud: Ok.
On Windows, the user can use the latest QT développement environnement to build and I think it's better to say : Hey, please use the latest environnement.
We are talking about 1% of the user on Windows who will build there own Jamulus app...
I will hard code but don't like that.

src/res/translation/translation_fr_FR.ts
jujudusud: I think I made a mistake Yesterday.
I will solve that. Thanks

ChangeLog
jujudusud: My bas. Will solve.

jujudusud: Solved by 
5aa0168

Jamulus.pro
jujudusud: ok in
5aa0168

src/res/translation/translation_fr_FR.ts
jujudusud: solved 
5aa0168

Jamulus.pro
jujudusud: Think it's ok in the last commit:
5aa0168

Jamulus.pro
jujudusud: Hard coded in the last commit:
5aa0168

jujudusud:
Ready to be merged.

Jamulus.pro
```diff
@@ -93,22 +93,31 @@ win32 {
 
     # replace ASIO with jack if requested
     contains(CONFIG, "jackonwindows") {
-        message(Using Jack instead of ASIO.)
-
-        !exists("C:/Program Files (x86)/Jack/includes/jack/jack.h") {
-            message(Warning: jack.h was not found at the usual place, maybe jack is not installed)
+        contains(QT_ARCH, "i386") {
```
pljones, line 96: The indentation here is all over the place, making it very hard to read what's intended.  Please make sure only spaces are used (for example, there are leading tabs on many lines).

Jamulus.pro
```diff
@@ -93,22 +93,31 @@ win32 {
...
+        	message("Native i386 build")
```
pljones, line 97: Is this correct?  What if I'm running Windows 64-bit but a 32bit compile?  That's _not_ a native i386 build.

Jamulus.pro
```diff
@@ -93,22 +93,31 @@ win32 {
...
+        	libjackname = libjack.lib
```
pljones, line 98: Why is there no check for `jack.h` for 32bit?

Jamulus.pro
```diff
@@ -93,22 +93,31 @@ win32 {
...
+        } else {
+        	message("Native x86_64 build")
+		    libjackname = libjack64.lib
+		    !exists("C:/Program Files (86)/JACK2/include/jack/jack.h") {
+            	message("Warning: jack.h was not found at the usual place, install JACK2 32 bits to cross compile.")
```
pljones, line 103: The message above says this is a native compile.  Here, you're saying it's a cross-compile.  This seems confused.

Jamulus.pro
```diff
@@ -93,22 +93,31 @@ win32 {
...
+         	}	
         }
+          !exists("C:/Program Files/JACK2/include/jack/jack.h") {
```
pljones, line 106: This check appears to be happening regardless of whether on 32 or 64bit.

Jamulus.pro
jujudusud: I would say If you are running a Windows 64 bits, QT_ARCH will be set to `x86_64`.

Jamulus.pro
jujudusud: No, If `C:\Program Files (x86)` exist then you cross compile .

Jamulus.pro
jujudusud: Sorry, my bad.
087f835

Jamulus.pro
jujudusud: At line 106.

Jamulus.pro
jujudusud: That's right. `jack.h` is at the same place whether you are in 32 bits or 64 bits Windows OS.
64 bits is  - > `C:/Program Files/JACK2/include/jack/jack.h` + eventually if 32 bits is installed `C:/Program Files (86)/JACK2/include/jack/jack.h`
where 32 bits is "only" -> `C:/Program Files/JACK2/include/jack/jack.h`

Jamulus.pro
```diff
@@ -94,19 +94,19 @@ win32 {
...
-         	}	
+            message("Native x86_64 build")
+            libjackname = libjack64.lib
+            !exists("C:/Program Files (86)/JACK2/include/jack/jack.h") {
+                message("Warning: jack.h was not found at the usual place, install JACK2 32 bits to cross compile.")
+            }
         }
           !exists("C:/Program Files/JACK2/include/jack/jack.h") {
-         	 message("Warning: jack.h was not found at the usual place, maybe JACK2 64 bits to compile")
+            message("Warning: jack.h was not found at the usual place, maybe JACK2 64 bits to compile")
```
henkdegroot, line 107: This message is not correct. This shows for both cases (x86_64 and i386). But when you build on a 32-bit platform, you should install the 32-bit JACK version. Perhaps extend the text with this information? Something like "install JACK2 32/64 bits depending on your OS architecture to compile"
Also the text "maybe JACK2" would be more accurate if it is "install JACK2".

henkdegroot (changes requested):

Jamulus.pro
```diff
@@ -93,22 +93,31 @@ win32 {
...
-
-        !exists("C:/Program Files (x86)/Jack/includes/jack/jack.h") {
-            message(Warning: jack.h was not found at the usual place, maybe jack is not installed)
+        contains(QT_ARCH, "i386") {
+            message("Native i386 build")
+            libjackname = libjack.lib
+        } else {
+            message("Native x86_64 build")
+            libjackname = libjack64.lib
+            !exists("C:/Program Files (86)/JACK2/include/jack/jack.h") {
```
henkdegroot, line 102: "C:/Program Files (86)/JACK2/include/jack/jack.h" is missing an x, it should be

C:/Program Files (**x**86)/JACK2/include/jack/jack.h

jujudusud: Ok in 02db193

Jamulus.pro
jujudusud: Modified in 02db193

Jamulus.pro
```diff
@@ -116,7 +116,7 @@ win32 {
             DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
             INCLUDEPATH += "C:/Program Files/JACK2/include"
             LIBS += "C:/Program Files/JACK2/lib/$${libjackname}"
```
henkdegroot, line 118: @jujudusud , the LIBS value needs to set different when cross compiling (x86/i386 in a 64-bit Win OS). Current an error is generated when the lib is linked as it cannot find the **libjack.lib** in the folder "C:/Program Files/JACK2/lib/", as it is located in the "C:/Program Files (x86)/JACK2/lib/" folder.

jujudusud: That's why I said we have to had a win64 target...
Do you know how to fix that ?

Jamulus.pro
```diff
@@ -93,22 +93,31 @@ win32 {
...
+            SOURCES -= windows/sound.cpp
+            HEADERS += linux/sound.h
+            SOURCES += linux/sound.cpp
+            DEFINES += WITH_JACK
+            DEFINES += JACK_REPLACES_ASIO
+            DEFINES += _STDINT_H # supposed to solve compilation error in systemdeps.h
+            INCLUDEPATH += "C:/Program Files/JACK2/include"
+            LIBS += "C:/Program Files/JACK2/lib/$${libjackname}"
     }
+
```
hoffie, line 120: Just a quick note (haven't looked at the other recent updates yet): Indentation still seems to be off.

jujudusud: I think it is ok in 506876e4

Jamulus.pro
jujudusud: is it ok with the 506876e4 commit ?

Jamulus.pro
```diff
@@ -93,10 +93,29 @@ win32 {
...
+                message ("Cross compilation build")
+                programfilesdir = "C:/Program Files (x86)"
+                !exists("$${"programfilesdir"}/JACK2/include/jack/jack.h") {
+                    message("Warning: jack.h was not found at the usual place, maybe JACK2 32 bits is not installed to cross compile.")
+                }
+            }
+            !exist("C:/Program Files") {
+                message("Native i386 build")
+                programfilesdir = "C:/Program Files"
+                !exists("$${"programfilesdir"}/JACK2/include/jack/jack.h") {
```
hoffie, line 107: Should this rather be an `else` branch to the previous check? Otherwise, both checks would run, wouldn't they? I guess that would not be intended.

Jamulus.pro
```diff
@@ -93,10 +93,29 @@ win32 {
...
     contains(CONFIG, "jackonwindows") {
-        message(Using Jack instead of ASIO.)
-
-        !exists("C:/Program Files (x86)/Jack/includes/jack/jack.h") {
-            message(Warning: jack.h was not found at the usual place, maybe jack is not installed)
+        contains(QT_ARCH, "i386") {
+            !exist("C:/Program Files (x86)") {
+                message ("Cross compilation build")
+                programfilesdir = "C:/Program Files (x86)"
+                !exists("$${"programfilesdir"}/JACK2/include/jack/jack.h") {
```
hoffie, line 100: `$${"programfilesdir"}` -- this quoting looks surprising to me, but I don't claim to be a `qmake` expert. I would have expected `$${programfilesdir}` similar to `$${libjackname}` below. Is this necessary, if so, why?

This affects this line and some lines further down.

Jamulus.pro
```diff
@@ -93,10 +93,29 @@ win32 {
...
+                    message("Warning: jack.h was not found at the usual place, maybe JACK2 32 bits is not installed to cross compile.")
+                }
```
hoffie, line 102: It looks like the `exists` block is required in all three code paths. Also, it seems to be identical as variables are used (which is good :)). Maybe this can be condensed to a single check after the last block?

`message("Warning: jack.h was not found in the expected location ($${programfilesdir}). Ensure that the right JACK2 variant is installed (32bit vs. 64bit).")`

Jamulus.pro
```diff
@@ -93,10 +93,29 @@ win32 {
...
+            }
+            !exist("C:/Program Files") {
```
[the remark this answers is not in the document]
jujudusud: 86c02e43

Jamulus.pro
```diff
@@ -93,10 +93,29 @@ win32 {
 
     # replace ASIO with jack if requested
     contains(CONFIG, "jackonwindows") {
-        message(Using Jack instead of ASIO.)
-
-        !exists("C:/Program Files (x86)/Jack/includes/jack/jack.h") {
-            message(Warning: jack.h was not found at the usual place, maybe jack is not installed)
+        contains(QT_ARCH, "i386") {
+            !exist("C:/Program Files (x86)") {
+                message ("Cross compilation build")
```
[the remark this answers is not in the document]
jujudusud: 86c02e43

Jamulus.pro
jujudusud: 86c02e43

Jamulus.pro
jujudusud: If QT_ARCH is set to i386, it is a x86 native or a x64 cross compile for 32 bit. It should work like that. lets see wha append when @henkdegroot builds.

Jamulus.pro
jujudusud: The check is identical but the message really different. Don't know if it is better like that.

Jamulus.pro
```diff
@@ -94,28 +94,39 @@ win32 {
     # replace ASIO with jack if requested
     contains(CONFIG, "jackonwindows") {
         contains(QT_ARCH, "i386") {
-            message("Native i386 build")
-            libjackname = libjack.lib
+            !exist("C:/Program Files (x86)") {
+                message ("Cross compilation build")
+                programfilesdir = "C:/Program Files (x86)"
```
pljones, line 99: You're in a block that's tested for "not exists" on this directory.  I think you have the logic wrong.

Jamulus.pro
jujudusud: Adding an else at last.

Jamulus.pro
jujudusud: 9a5c412a

Jamulus.pro
jujudusud: Your right. my bad.
7f243127

henkdegroot (approved):

henkdegroot:
FYI, I tested these scenario's on my win10 64-bit machine:

For testing I updated the Jamulus.pro CONFIG section and added the "jackonwindows" option.
The below has been tested running the windows powershell script to build the application and generate the win installer.

No JACK2 installed: 
Validate both x86_64 and i386 report the correct message about missing JACK2.
Both options generate the message about missing JACK2.

Only JACK2 64-bit installed:
Validate x86_64 does not show missing JACK2 warning.
Validate i386 still generates the missing JACK2 warning.
Both working as intended.

JACK2 64 & 32 bit installed:
Validate x86_64 does not show missing JACK2 warning.
Validate i386 does not show missing JACK2 warning.
Both working as intended.

Validate build is successfull and Jamulus works with JACK2.
Working as intended, Jamulus can even connect to both JACK2 versions (32 and 64).

Well done @jujudusud

jujudusud:
Special thanks @henkdegroot for testing and also thanks a lot @hoffie and @pljones for your patience ;-)

I think it is now really ready to be merged as it does what it stands for. :-)

hoffie (approved):
Thanks! Code looks good to me now, thanks for the extensive testing @henkdegroot!

Although the changes look non-invasive and only touch the jack* logic, I'm not sure if we should risk pushing it to 3.8.0.

I have therefore added the 3.8.1 milestone.

pljones (approved):

jujudusud:
@ann0see this one should also ne merged if there is a RC2

ann0see:
This depends on what the other @jamulussoftware/maindevelopers say.

hoffie:
> @ann0see this one should also be merged if there is a RC2

I'd say: No.

While a lot has been merged post-rc1, all of it was either merged to avoid regressions being introduced by 3.8.0 (e.g. delay pan config persistence) or because it is clearly expected not to affect the final binaries (e.g. clang Github actions).

As said, this PR looks non-breaking, looks to be limited to jackonwindows and is therefore unlikely to affect official binaries. However, it still touches code which is run on every build, so I'd personally defer it to 3.8.1.

There's already 7 issues/PRs marked for 3.8.1, so I guess we could release a 3.8.1 really soonish after 3.8.0 (without running a *mandatory* translation round). So, that'd be my preference. If other developers think different, I'd be ok with trying to stuff it into 3.8.0 as well.

pljones:
I think 3.8.1 would be better, as we could also then automate the builds, with any luck.

Then have some fun getting them into a single Windows installer...

ann0see:
Merged. 

Further todo items as I can read from above:

- [ ] Improved Windows installer: check for JACK if it is a non ASIO build: The magic is here: https://github.com/jamulussoftware/jamulus/blob/d6acd63b98ce345c32c5ac87a99e3b4a532928f7/windows/installer.nsi#L368-L392
- [ ] Added automatic windows build for JACK

This checklist can be copied to an issue or other place

PR #1728, opened 2021-05-20
Feature connectionlist

geotechfirst:
Screen Shot 2021-05-20 at 17 42 20
The need:
- During WorldJam performances performers need to change servers multiple times. Using the currently available method this is not convenient.
- Sound engineers and other organisers need to move around a lot too.
- Most of these transitions are pre-defined and lists can be generated to provide a convenient path through the process.

The POC:
- This change adds a Menu item to the application where the user can select to load a server list *.jcl (Jamulus connection list)
- Format: "name,host:port"
- This list is provided in a new dialog box that allows to connect to the selected server via double click or using the relevant button. Also provides a connect to next button that takes the next server in the list and connects to that.
All previous connect/disconnect functionality stays un-affected.

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -400,6 +400,7 @@ FORMS_GUI = src/clientdlgbase.ui \
 HEADERS += src/buffer.h \
     src/channel.h \
     src/client.h \
+    src/connectionlistdlg.h \
     src/global.h \
     src/protocol.h \
     src/recorder/jamcontroller.h \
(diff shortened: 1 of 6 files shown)


pljones:
As it stands, I wouldn't want to accept this into the code base as it's not particularly usable.  There's no way to edit the list built in.

Doing so makes it very, very similar to #1552 - all you need to do is add
* save favourites
* load favourites
* next favourite (menu option)
* allow manual sorting of the list of favourites

geotechfirst:
I am happy to add the missing options to my change.
Would you support the PR if those items added?

pljones:
> Would you support the PR if those items added?

So long as it's based on #1552, rather than something independent.

geotechfirst:
Renamed to bookmarks and added load/save/add bookmarks.

ann0see:
Hi @geotechfirst I‘ve converted this to a draft for now. I think you should also have aloof at the other pull request.

ann0see:
@geotechfirst @dcorson-ticino-com I still think some people might like a favourites feature in Jamulus (maybe for public servers). As it doesn’t seem to be easily implementable if there’s still need I‘d suggest to open a new GitHub discussion and link both PRs. However I am closing this now since it seems to be stale

PR #1731, opened 2021-05-21
Add Single Click on Win OS Server System Tray Icon

### LINKED ISSUE #1722: Add Click action to Server System Tray Icon when using Windows OS
Not discussed yet, but happy to do here as it is not really complex.

Working on Windows with application in the systemtray, it is very common these days that a single (left) click on the tray icon opens a dialog or context menu. Current when running Jamulus with the server option, nothing happens with single click.
Double Click is implemented, but not intuitive for most people these days. Right Click is implemented and shows a context menu.
On Windows I would like to show the main server dialog when using the single (left) click on the tray icon.

**To consider**
On Linux, the single (left) click is showing the context menu.
Not sure about macOS.
Hence why the feature request is for Windows OS only.

henkdegroot:
Fixes #1722

### DIFF
--- a/src/serverdlg.cpp
+++ b/src/serverdlg.cpp
@@ -598,6 +598,14 @@ void CServerDlg::OnSysTrayActivated ( QSystemTrayIcon::ActivationReason ActReaso
     {
         ShowWindowInForeground();
     }
+
+    // on single click on the icon, show window in fore ground for windows only
+#ifdef _WIN32
+    if ( ActReason == QSystemTrayIcon::Trigger )
+    {
+        ShowWindowInForeground();
+    }
+#endif
 }
 
 void CServerDlg::OnCLVersionAndOSReceived ( CHostAddress, COSUtil::EOpSystemType, QString strVersion )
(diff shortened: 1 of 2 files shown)


src/serverdlg.cpp
```diff
@@ -598,6 +598,14 @@ void CServerDlg::OnSysTrayActivated ( QSystemTrayIcon::ActivationReason ActReaso
     {
         ShowWindowInForeground();
     }
+
```
pljones, line 601: I'd prefer
```
#ifdef _WIN32
if ( ActReason == QSystemTrayIcon::Trigger || ActReason == QSystemTrayIcon::DoubleClick )
#else
if ( ActReason == QSystemTrayIcon::DoubleClick )
#endif
```
over the duplication - that's partly why the code format keeps the `if` on its own line, I think.

henkdegroot: Thank you for the feedback, indeed makes sense to do so.  Will update, test and update the PR.

pljones (approved):

ann0see (approved):
Works as expected

hoffie:
CHANGELOG: GUI: Single clicking Jamulus' server system tray icon will now show the UI on Windows as well.

PR #1732, opened 2021-05-21
Improve UX in error messages (#1568)

ann0see:
This is the a start for a discussion!

I looked over some error messages especially concerning the sound system and changed some. It would be great if @gilgongo and @mulyaj looked over these messages from an UX point of view. Also I'd appreciate if @pljones or @softins looked at the content. 

Context:
https://github.com/jamulussoftware/jamulus/discussions/1568

Note: I need to keep https://github.com/jamulussoftware/jamuluswebsite/issues/339 in mind.

Missing before merge:

Final review

### DIFF
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -37,18 +37,15 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
         serverName = "default";
     }
     qInfo() << qUtf8Printable (
-        QString ( "Connecting to jack \"%1\" instance (use the JACK_DEFAULT_SERVER environment variable to change this)." ).arg ( serverName ) );
+        QString ( "Connecting to JACK \"%1\" instance (use the JACK_DEFAULT_SERVER environment variable to change this)." ) .arg ( serverName ) );
 
     // try to become a client of the JACK server
     pJackClient = jack_client_open ( jackClientName, JackNullOption, &JackStatus );
 
     if ( pJackClient == nullptr )
     {
-        throw CGenErr ( tr ( "The Jack server is not running. This software "
-                             "requires a Jack server to run. Normally if the Jack server is "
-                             "not running this software will automatically start the Jack server. "
-                             "It seems that this auto start has not worked. Try to start the Jack "
-                             "server manually." ) );
+        throw CGenErr ( tr ( "The JACK server couldn't be started automatically. "
+                             "Please start JACK manually and check for error messages." ) );
     }
 
     // tell the JACK server to call "process()" whenever
@@ -64,16 +61,9 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
     // check sample rate, if not correct, just fire error
     if ( jack_get_sample_rate ( pJackClient ) != SYSTEM_SAMPLE_RATE_HZ )
     {
-        throw CGenErr ( tr ( "The Jack server sample rate is different from "
-                             "the required one. The required sample rate is:" ) +
-                        " <b>" + QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ) + " Hz</b>. " +
-                        tr ( "You can "
-                             "use a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QJackCtl</a></i> "
-                             "to adjust the Jack server sample rate." ) +
-                        "<br>" +
-                        tr ( "Make sure to set the "
-                             "Frames/Period to a low value like " ) +
-                        QString().setNum ( DOUBLE_SYSTEM_FRAME_SIZE_SAMPLES ) + tr ( " to achieve a low delay." ) );
+        throw CGenErr ( QString ( tr ( "The JACK server isn't running at a sample rate of <b>%1 Hz</b>. Please use "
+                        "a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QJackCtl</a></i> to set the "
+                        "the JACK server sample rate to %1 Hz." ) ) .arg ( SYSTEM_SAMPLE_RATE_HZ ) );
     }
 
     // create four ports (two for input, two for output -> stereo)
@@ -87,7 +77,10 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
 
     if ( ( input_port_left == nullptr ) || ( input_port_right == nullptr ) || ( output_port_left == nullptr ) || ( output_port_right == nullptr ) )
     {
-        throw CGenErr ( tr ( "The Jack port registering failed." ) );
+        throw CGenErr ( QString ( tr ( "The JACK port registering failed. This is a probably an error with JACK. Please stop %1 and JACK. "
+                        "Afterwards check if another program at a sample rate of %2 Hz can connect to JACK." ) )
+                        .arg ( APP_NAME )
+                        .arg ( SYSTEM_SAMPLE_RATE_HZ ) );
     }
 
     // optional MIDI initialization
@@ -97,7 +90,9 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
 
         if ( input_port_midi == nullptr )
         {
-            throw CGenErr ( tr ( "The Jack port registering failed." ) );
+            throw CGenErr ( QString ( tr ( "The JACK port registering failed. This is a probably an error with JACK. Please stop %1 and JACK. "
+                            "Afterwards check if another MIDI program can connect to JACK." ) )
+                            .arg ( APP_NAME ) );
         }
     }
     else
@@ -108,7 +103,8 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
     // tell the JACK server that we are ready to roll
     if ( jack_activate ( pJackClient ) )
     {
-        throw CGenErr ( tr ( "Cannot activate the Jack client." ) );
+        throw CGenErr ( QString ( tr ( "Can't activate the JACK client. This is a probably an error with JACK. Please check the JACK output." ) )
+                        .arg ( APP_NAME ) );
     }
 
     if ( !bNoAutoJackConnect )
@@ -210,9 +206,10 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
     // without a Jack server, Jamulus makes no sense to run, throw an error message
     if ( bJackWasShutDown )
     {
-        throw CGenErr ( tr ( "The Jack server was shut down. This software "
-                             "requires a Jack server to run. Try to restart the software to "
-                             "solve the issue." ) );
+        throw CGenErr ( QString ( tr ( "The JACK server was shut down but %1 "
+                             "requires a JACK to run. Please restart %1 to "
+                             "start JACK again. If the error persists, please review "
+                             "the JACK output." ) ) .arg ( APP_NAME ) );
     }
 
     // get actual buffer size
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -206,7 +206,7 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
     // without a Jack server, Jamulus makes no sense to run, throw an error message
     if ( bJackWasShutDown )
     {
-        throw CGenErr ( QString ( tr ( "The JACK server was shut down but %1 "
+        throw CGenErr ( QString ( tr ( "The JACK server was shut down. %1 "
                              "requires a JACK to run. Please restart %1 to "
                              "start JACK again. If the error persists, please review "
                              "the JACK output." ) ) .arg ( APP_NAME ) );
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -44,7 +44,7 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
 
     if ( pJackClient == nullptr )
     {
-        throw CGenErr ( tr ( "The JACK server couldn't be started automatically. "
+        throw CGenErr ( tr ( "JACK couldn't be started automatically. "
                              "Please start JACK manually and check for error messages." ) );
     }
 
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -78,7 +78,7 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
 
     if ( ( input_port_left == nullptr ) || ( input_port_right == nullptr ) || ( output_port_left == nullptr ) || ( output_port_right == nullptr ) )
     {
-        throw CGenErr ( QString ( tr ( "The JACK port registering failed. This is a probably an error with JACK. Please stop %1 and JACK. "
+        throw CGenErr ( QString ( tr ( "The JACK port registration failed. This is probably an error with JACK. Please stop %1 and JACK. "
                                        "Afterwards check if another program at a sample rate of %2 Hz can connect to JACK." ) )
                             .arg ( APP_NAME )
                             .arg ( SYSTEM_SAMPLE_RATE_HZ ) );
@@ -91,8 +91,8 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
 
         if ( input_port_midi == nullptr )
         {
-            throw CGenErr ( QString ( tr ( "The JACK port registering failed. This is a probably an error with JACK. Please stop %1 and JACK. "
-                                           "Afterwards check if another MIDI program can connect to JACK." ) )
+            throw CGenErr ( QString ( tr ( "The JACK port registration failed. This is probably an error with JACK. Please stop %1 and JACK. "
+                                           "Afterwards, check if another MIDI program can connect to JACK." ) )
                                 .arg ( APP_NAME ) );
         }
     }
@@ -209,7 +209,7 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
     {
         throw CGenErr ( QString ( tr ( "The JACK server was shut down. %1 "
                                        "requires a JACK server to run. Please restart %1 to "
-                                       "start JACK again. If the error persists, please review "
+                                       "start JACK again. "
                                        "the JACK output." ) )
                             .arg ( APP_NAME ) );
     }
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -208,7 +208,7 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
     if ( bJackWasShutDown )
     {
         throw CGenErr ( QString ( tr ( "The JACK server was shut down. %1 "
-                                       "requires a JACK to run. Please restart %1 to "
+                                       "requires a JACK server to run. Please restart %1 to "
                                        "start JACK again. If the error persists, please review "
                                        "the JACK output." ) )
                             .arg ( APP_NAME ) );
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -61,9 +61,9 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
     // check sample rate, if not correct, just fire error
     if ( jack_get_sample_rate ( pJackClient ) != SYSTEM_SAMPLE_RATE_HZ )
     {
-        throw CGenErr ( QString ( tr ( "The JACK server isn't running at a sample rate of <b>%1 Hz</b>. Please use "
+        throw CGenErr ( QString ( tr ( "JACK isn't running at a sample rate of <b>%1 Hz</b>. Please use "
                                        "a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QjackCtl</a></i> to set the "
-                                       "the JACK server sample rate to %1 Hz." ) )
+                                       "the JACK sample rate to %1 Hz." ) )
                             .arg ( SYSTEM_SAMPLE_RATE_HZ ) );
     }
 
@@ -207,8 +207,8 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
     // without a Jack server, Jamulus makes no sense to run, throw an error message
     if ( bJackWasShutDown )
     {
-        throw CGenErr ( QString ( tr ( "The JACK server was shut down. %1 "
-                                       "requires a JACK server to run. Please restart %1 to "
+        throw CGenErr ( QString ( tr ( "JACK was shut down. %1 "
+                                       "requires JACK to run. Please restart %1 to "
                                        "start JACK again. "
                                        "the JACK output." ) )
                             .arg ( APP_NAME ) );
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -104,7 +104,7 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
     // tell the JACK server that we are ready to roll
     if ( jack_activate ( pJackClient ) )
     {
-        throw CGenErr ( QString ( tr ( "Can't activate the JACK client. This is a probably an error with JACK. Please check the JACK output." ) )
+        throw CGenErr ( QString ( tr ( "Can't activate the JACK client. This is probably an error with JACK. Please check the JACK output." ) )
                             .arg ( APP_NAME ) );
     }
 
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -37,7 +37,7 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
         serverName = "default";
     }
     qInfo() << qUtf8Printable (
-        QString ( "Connecting to JACK \"%1\" instance (use the JACK_DEFAULT_SERVER environment variable to change this)." ) .arg ( serverName ) );
+        QString ( "Connecting to JACK \"%1\" instance (use the JACK_DEFAULT_SERVER environment variable to change this)." ).arg ( serverName ) );
 
     // try to become a client of the JACK server
     pJackClient = jack_client_open ( jackClientName, JackNullOption, &JackStatus );
@@ -62,8 +62,9 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
     if ( jack_get_sample_rate ( pJackClient ) != SYSTEM_SAMPLE_RATE_HZ )
     {
         throw CGenErr ( QString ( tr ( "The JACK server isn't running at a sample rate of <b>%1 Hz</b>. Please use "
-                        "a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QJackCtl</a></i> to set the "
-                        "the JACK server sample rate to %1 Hz." ) ) .arg ( SYSTEM_SAMPLE_RATE_HZ ) );
+                                       "a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QJackCtl</a></i> to set the "
+                                       "the JACK server sample rate to %1 Hz." ) )
+                            .arg ( SYSTEM_SAMPLE_RATE_HZ ) );
     }
 
     // create four ports (two for input, two for output -> stereo)
@@ -78,9 +79,9 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
     if ( ( input_port_left == nullptr ) || ( input_port_right == nullptr ) || ( output_port_left == nullptr ) || ( output_port_right == nullptr ) )
     {
         throw CGenErr ( QString ( tr ( "The JACK port registering failed. This is a probably an error with JACK. Please stop %1 and JACK. "
-                        "Afterwards check if another program at a sample rate of %2 Hz can connect to JACK." ) )
-                        .arg ( APP_NAME )
-                        .arg ( SYSTEM_SAMPLE_RATE_HZ ) );
+                                       "Afterwards check if another program at a sample rate of %2 Hz can connect to JACK." ) )
+                            .arg ( APP_NAME )
+                            .arg ( SYSTEM_SAMPLE_RATE_HZ ) );
     }
 
     // optional MIDI initialization
@@ -91,8 +92,8 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
         if ( input_port_midi == nullptr )
         {
             throw CGenErr ( QString ( tr ( "The JACK port registering failed. This is a probably an error with JACK. Please stop %1 and JACK. "
-                            "Afterwards check if another MIDI program can connect to JACK." ) )
-                            .arg ( APP_NAME ) );
+                                           "Afterwards check if another MIDI program can connect to JACK." ) )
+                                .arg ( APP_NAME ) );
         }
     }
     else
@@ -104,7 +105,7 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
     if ( jack_activate ( pJackClient ) )
     {
         throw CGenErr ( QString ( tr ( "Can't activate the JACK client. This is a probably an error with JACK. Please check the JACK output." ) )
-                        .arg ( APP_NAME ) );
+                            .arg ( APP_NAME ) );
     }
 
     if ( !bNoAutoJackConnect )
@@ -207,9 +208,10 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
     if ( bJackWasShutDown )
     {
         throw CGenErr ( QString ( tr ( "The JACK server was shut down. %1 "
-                             "requires a JACK to run. Please restart %1 to "
-                             "start JACK again. If the error persists, please review "
-                             "the JACK output." ) ) .arg ( APP_NAME ) );
+                                       "requires a JACK to run. Please restart %1 to "
+                                       "start JACK again. If the error persists, please review "
+                                       "the JACK output." ) )
+                            .arg ( APP_NAME ) );
     }
 
     // get actual buffer size
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -62,7 +62,7 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
     if ( jack_get_sample_rate ( pJackClient ) != SYSTEM_SAMPLE_RATE_HZ )
     {
         throw CGenErr ( QString ( tr ( "The JACK server isn't running at a sample rate of <b>%1 Hz</b>. Please use "
-                                       "a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QJackCtl</a></i> to set the "
+                                       "a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QjackCtl</a></i> to set the "
                                        "the JACK server sample rate to %1 Hz." ) )
                             .arg ( SYSTEM_SAMPLE_RATE_HZ ) );
     }
--- a/mac/sound.cpp
+++ b/mac/sound.cpp
@@ -460,8 +460,8 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
          ( CurDevStreamFormat.mBitsPerChannel != 32 ) || ( !( CurDevStreamFormat.mFormatFlags & kAudioFormatFlagIsFloat ) ) ||
          ( !( CurDevStreamFormat.mFormatFlags & kAudioFormatFlagIsPacked ) ) )
     {
-        return tr ( "The stream format on the current input device isn't "
-                    "compatible with this software. Please select another device." );
+        return QString ( tr ( "The stream format on the current input device isn't "
+                    "compatible with this software. Please select another device." ) ) .arg ( APP_NAME );
     }
 
     // check the output
@@ -471,8 +471,8 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
          ( CurDevStreamFormat.mBitsPerChannel != 32 ) || ( !( CurDevStreamFormat.mFormatFlags & kAudioFormatFlagIsFloat ) ) ||
          ( !( CurDevStreamFormat.mFormatFlags & kAudioFormatFlagIsPacked ) ) )
     {
-        return tr ( "The stream format on the current output device isn't "
-                    "compatible with this software. Please select another device." );
+        return QString ( tr ( "The stream format on the current output device isn't "
+                    "compatible with %1. Please select another device." ) ) .arg ( APP_NAME );
     }
 
     // store the input and out number of channels for this device
--- a/mac/sound.cpp
+++ b/mac/sound.cpp
@@ -394,7 +394,7 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
             return QString ( tr ( "The sample rate on the current input device isn't %1 Hz and is therefore incompatible. "
                                   "Please select another device or try setting the sample rate to %1 Hz "
                                   "manually via Audio-MIDI-Setup (in Applications->Utilities)." ) )
-                                  .arg ( SYSTEM_SAMPLE_RATE_HZ );
+                .arg ( SYSTEM_SAMPLE_RATE_HZ );
         }
     }
 
@@ -413,9 +413,9 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
              noErr )
         {
             return QString ( tr ( "The sample rate on the current output device isn't %1 Hz and is therefore incompatible. "
-                                "Please select another device or try setting the sample rate to %1 Hz "
-                                "manually via Audio-MIDI-Setup (in Applications->Utilities)." ) )
-                                .arg ( SYSTEM_SAMPLE_RATE_HZ );
+                                  "Please select another device or try setting the sample rate to %1 Hz "
+                                  "manually via Audio-MIDI-Setup (in Applications->Utilities)." ) )
+                .arg ( SYSTEM_SAMPLE_RATE_HZ );
         }
     }
 
@@ -461,7 +461,8 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
          ( !( CurDevStreamFormat.mFormatFlags & kAudioFormatFlagIsPacked ) ) )
     {
         return QString ( tr ( "The stream format on the current input device isn't "
-                    "compatible with this software. Please select another device." ) ) .arg ( APP_NAME );
+                              "compatible with this software. Please select another device." ) )
+            .arg ( APP_NAME );
     }
 
     // check the output
@@ -472,7 +473,8 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
          ( !( CurDevStreamFormat.mFormatFlags & kAudioFormatFlagIsPacked ) ) )
     {
         return QString ( tr ( "The stream format on the current output device isn't "
-                    "compatible with %1. Please select another device." ) ) .arg ( APP_NAME );
+                              "compatible with %1. Please select another device." ) )
+            .arg ( APP_NAME );
     }
 
     // store the input and out number of channels for this device
--- a/mac/sound.cpp
+++ b/mac/sound.cpp
@@ -391,7 +391,7 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
         if ( AudioObjectSetPropertyData ( audioInputDevice[iDriverIdx], &stPropertyAddress, 0, NULL, sizeof ( Float64 ), &fSystemSampleRate ) !=
              noErr )
         {
-            return QString ( tr ( "The sample rate on the current input device isn't %1 Hz and therefore incompatible. "
+            return QString ( tr ( "The sample rate on the current input device isn't %1 Hz and is therefore incompatible. "
                                   "Please select another device or try setting the sample rate to %1 Hz "
                                   "manually via Audio-MIDI-Setup (in Applications->Utilities)." ) )
                                   .arg ( SYSTEM_SAMPLE_RATE_HZ );
@@ -412,7 +412,7 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
         if ( AudioObjectSetPropertyData ( audioOutputDevice[iDriverIdx], &stPropertyAddress, 0, NULL, sizeof ( Float64 ), &fSystemSampleRate ) !=
              noErr )
         {
-            return QString ( tr ( "The sample rate on the current output device isn't %1 Hz and therefore incompatible. "
+            return QString ( tr ( "The sample rate on the current output device isn't %1 Hz and is therefore incompatible. "
                                 "Please select another device or try setting the sample rate to %1 Hz "
                                 "manually via Audio-MIDI-Setup (in Applications->Utilities)." ) )
                                 .arg ( SYSTEM_SAMPLE_RATE_HZ );
--- a/mac/sound.cpp
+++ b/mac/sound.cpp
@@ -110,8 +110,8 @@ void CSound::GetAvailableInOutDevices()
 
     if ( AudioObjectGetPropertyData ( kAudioObjectSystemObject, &stPropertyAddress, 0, NULL, &iPropertySize, &audioInputDevice[lNumDevs] ) )
     {
-        throw CGenErr ( tr ( "CoreAudio input AudioHardwareGetProperty call failed. "
-                             "It seems that no sound card is available in the system." ) );
+        throw CGenErr ( tr ( "No sound card is available in your system. "
+                             "CoreAudio input AudioHardwareGetProperty call failed." ) );
     }
 
     iPropertySize               = sizeof ( AudioDeviceID );
@@ -119,8 +119,8 @@ void CSound::GetAvailableInOutDevices()
 
     if ( AudioObjectGetPropertyData ( kAudioObjectSystemObject, &stPropertyAddress, 0, NULL, &iPropertySize, &audioOutputDevice[lNumDevs] ) )
     {
-        throw CGenErr ( tr ( "CoreAudio output AudioHardwareGetProperty call failed. "
-                             "It seems that no sound card is available in the system." ) );
+        throw CGenErr ( tr ( "No sound card is available in the system. "
+                             "CoreAudio output AudioHardwareGetProperty call failed." ) );
     }
 
     lNumDevs++; // next device
@@ -281,7 +281,7 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName, bool )
     // if the selected driver was not found, return an error message
     if ( iDriverIdx == INVALID_INDEX )
     {
-        return tr ( "The current selected audio device is no longer present in the system." );
+        return tr ( "The currently selected audio device is no longer present. Please check your audio device." );
     }
 
     // check device capabilities if it fulfills our requirements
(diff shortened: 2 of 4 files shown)


mac/sound.cpp
```diff
@@ -391,11 +391,11 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
...
-            return QString ( tr ( "Current system audio input device sample "
-                                  "rate of %1 Hz is not supported. Please open the Audio-MIDI-Setup in "
-                                  "Applications->Utilities and try to set a sample rate of %2 Hz." ) )
-                .arg ( static_cast<int> ( inputSampleRate ) )
-                .arg ( SYSTEM_SAMPLE_RATE_HZ );
+            return QString ( tr ( "The selected input device is incompatible "
+                                  "since it does not support the a sample rate of %1 Hz. Please select another "
+                                  "input device or try to set the sample rate to %1 Hz "
+                                  "manually via Audio-MIDI-Setup (in Applications->Utilities)." ) )
+                                  .arg ( QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ) );
```
ann0see, line 398: Is QString().setNum() needed? Will check the CI.

linux/sound.cpp
```diff
@@ -87,7 +77,7 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
 
     if ( ( input_port_left == nullptr ) || ( input_port_right == nullptr ) || ( output_port_left == nullptr ) || ( output_port_right == nullptr ) )
     {
-        throw CGenErr ( tr ( "The Jack port registering failed." ) );
+        throw CGenErr ( tr ( "The JACK port registering failed." ) );
```
mulyaj, line 80: I'm not too familiar with this, but is there a reason for the failure?

linux/sound.cpp
```diff
@@ -210,9 +200,9 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
     // without a Jack server, Jamulus makes no sense to run, throw an error message
     if ( bJackWasShutDown )
     {
-        throw CGenErr ( tr ( "The Jack server was shut down. This software "
-                             "requires a Jack server to run. Try to restart the software to "
-                             "solve the issue." ) );
+        throw CGenErr ( tr ( "The JACK server was shut down but this software "
```
mulyaj, line 203: The "but" isn't necessary. I think the original first two sentences are fine.

Or, alternatively with a semicolon: The JACK server was shut down; this software requires a JACK server to run.

Also, is "this software" referencing Jamulus? If so, should we just say "Jamulus" instead? Since we're dealing with a lot of interacting hardware and software, I think saying Jamulus directly clears up any possible confusion a user may have.

gilgongo: BTW, when these messages come up, do they have some indication (in the window bar for example) that they are from Jamulus? If so, then that might make the writing of some messages (maybe not this one) a bit simpler.

gilgongo:
I should add the ritual UX invocation on all this: before sitting down to write an error message, consider whether the system can be changed such that the message isn't needed :-)

linux/sound.cpp
```diff
@@ -64,16 +61,9 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
...
-                             "the required one. The required sample rate is:" ) +
-                        " <b>" + QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ) + " Hz</b>. " +
-                        tr ( "You can "
-                             "use a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QJackCtl</a></i> "
-                             "to adjust the Jack server sample rate." ) +
-                        "<br>" +
-                        tr ( "Make sure to set the "
-                             "Frames/Period to a low value like " ) +
-                        QString().setNum ( DOUBLE_SYSTEM_FRAME_SIZE_SAMPLES ) + tr ( " to achieve a low delay." ) );
+        throw CGenErr ( QString ( tr ( "The JACK server doesn't run at a sample rate of <b>%1 Hz</b>. Please use "
```
gilgongo, line 64: Should this be "The JACK server isn't..."? Saying "doesn't" implies it never will.

linux/sound.cpp
```diff
@@ -108,7 +98,7 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
     // tell the JACK server that we are ready to roll
     if ( jack_activate ( pJackClient ) )
     {
-        throw CGenErr ( tr ( "Cannot activate the Jack client." ) );
+        throw CGenErr ( tr ( "Cannot activate the JACK client." ) );
```
gilgongo, line 101: Does the "client" here mean Jamulus? Can we recommend an action to the user?

mac/sound.cpp
```diff
@@ -103,24 +103,24 @@ void CSound::GetAvailableInOutDevices()
...
+    strDriverNames[lNumDevs] = "System Default In/Out Devices"; // TODO: why can't this be translated?
 
     iPropertySize               = sizeof ( AudioDeviceID );
     stPropertyAddress.mSelector = kAudioHardwarePropertyDefaultInputDevice;
 
     if ( AudioObjectGetPropertyData ( kAudioObjectSystemObject, &stPropertyAddress, 0, NULL, &iPropertySize, &audioInputDevice[lNumDevs] ) )
     {
-        throw CGenErr ( tr ( "CoreAudio input AudioHardwareGetProperty call failed. "
-                             "It seems that no sound card is available in the system." ) );
+        throw CGenErr ( tr ( "It seems that no sound card is available in your system. "
```
gilgongo, line 113: "It seems" is a bit strange in this context. Can we just say "No sound card is available...."?

mac/sound.cpp
```diff
@@ -281,7 +281,7 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName, bool )
     // if the selected driver was not found, return an error message
     if ( iDriverIdx == INVALID_INDEX )
     {
-        return tr ( "The current selected audio device is no longer present in the system." );
+        return tr ( "The current selected audio device is no longer present in the system. Please check your audio device." );
```
gilgongo, line 284:
```suggestion
        return tr ( "The currently selected audio device is no longer present. Please check your audio device." );
```

mac/sound.cpp
```diff
@@ -382,7 +382,7 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
 
     if ( AudioObjectGetPropertyData ( audioInputDevice[iDriverIdx], &stPropertyAddress, 0, NULL, &iPropertySize, &inputSampleRate ) )
     {
-        return QString ( tr ( "The audio input device is no longer available." ) );
+        return QString ( tr ( "The audio input device is no longer available. Please check if your input device is connected correctly." ) );
```
gilgongo, line 385:
```suggestion
        return QString ( tr ( "The audio device is no longer available. Please check if your device is connected correctly." ) );
```

src/soundbase.cpp
```diff
@@ -154,10 +154,10 @@ QString CSoundBase::SetDev ( const QString strDevName )
         if ( !strDevName.isEmpty() )
         {
             strReturn = tr ( "The previously selected audio device "
-                             "is no longer available or the audio driver properties have changed to a state which "
-                             "is incompatible with this software. We now try to find a valid audio device. This new "
-                             "audio device might cause audio feedback. So, before connecting to a server, please "
-                             "check the audio device setting." );
+                             "is no longer available or the audio driver properties have changed to an incompatible state."
+                             "We now try to find a valid audio device. "
```
gilgongo, line 158:
```suggestion
                             "We will try to find a valid audio device. "
```

src/soundbase.cpp
```diff
@@ -166,12 +166,9 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
         {
             // create error message with all details
-            QString sErrorMessage = "<b>" + tr ( "No usable " ) + strSystemDriverTechniqueName +
-                                    tr ( " audio device "
-                                         "(driver) found." ) +
+            QString sErrorMessage = "<b>" + QString ( tr ( "No usable %1 audio device found." ) ) .arg ( strSystemDriverTechniqueName ) +
                                     "</b><br><br>" +
-                                    tr ( "In the following there is a list of all available drivers "
-                                         "with the associated error message:" ) +
+                                    tr ( "Please have a look at this list of all available drivers with error messages:" ) +
```
gilgongo, line 171:
```suggestion
                                    tr ( These are all the available drivers with error messages:" ) +
```

src/soundbase.cpp
```diff
@@ -184,14 +181,14 @@ QString CSoundBase::SetDev ( const QString strDevName )
             // to be able to access the ASIO driver setup for changing, e.g., the sample rate, we
             // offer the user under Windows that we open the driver setups of all registered
             // ASIO drivers
-            sErrorMessage = sErrorMessage + "<br/>" + tr ( "Do you want to open the ASIO driver setups?" );
+            sErrorMessage = sErrorMessage + "<br/>" + tr ( "Do you want to open the ASIO driver setup to change driver settings?" );
```
gilgongo, line 184: Why is this confirmation here? Would there be some negative consequence of opening the setup?

src/soundbase.cpp
```diff
@@ -184,14 +181,14 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
 
             if ( QMessageBox::Yes == QMessageBox::information ( nullptr, APP_NAME, sErrorMessage, QMessageBox::Yes | QMessageBox::No ) )
             {
                 LoadAndInitializeFirstValidDriver ( true );
             }
 
-            sErrorMessage = APP_NAME + tr ( " could not be started because of audio interface issues." );
+            sErrorMessage = QString ( tr ( " %1 could not be started because of audio interface issues. Please restart %1 and check/reconfigure your audio settings." ) );
```
gilgongo, line 191:
```suggestion
            sErrorMessage = QString ( tr ( " Can't start %1. Please restart %1 and check/reconfigure your audio settings." ) );
```

windows/sound.cpp
```diff
@@ -73,7 +73,7 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName, bool bOpenDrive
     {
         // clean up and return error string
         asioDrivers->removeCurrentDriver();
-        return tr ( "The audio driver could not be initialized." );
+        return tr ( "The audio driver could not be initialized. Check if the audio device is plugged in and verify your driver settings." );
```
gilgongo, line 76:
```suggestion
        return tr ( "The audio driver could not be initialised. Check if the audio device is plugged in and verify your driver settings." );

```

mulyaj:
To add, I think contractions should be kept consistent. Currently, there's some variance for doesn't and couldn't.

In my opinion, I think we should use the contracted form to sound a bit less robotic.

linux/sound.cpp
ann0see: No. As far as I understand it's basically the audio API connection which Jamulus uses. For the user, I doubt it makes a difference. Not even sure when this error appears.

src/soundbase.cpp
ann0see: > Would there be some negative consequence of opening the setup

Probably not. However in some cases it won't help much since the device (not the driver) is somehow incompatible. The whole handling around this line is not perfect.

linux/sound.cpp
ann0see: Not sure code wise.

ann0see:
> In my opinion, I think we should use the contracted form to sound a bit less robotic.

Agree. This comes with the unification of style&tone probably?

src/soundbase.cpp
ann0see: Fixed.

mac/sound.cpp
ann0see: Done. Probably the writer thought that not having a sound card is impossible?

linux/sound.cpp
ann0see: Changed this one.

ann0see:
> In my opinion, I think we should use the contracted form to sound a bit less robotic.

Yes. Probably that's a good solution. Let's wait for others to comment.

Note to maintainers: Good candidate for squash and merge. Will maybe do it manually

mac/sound.cpp
```diff
@@ -103,24 +103,24 @@ void CSound::GetAvailableInOutDevices()
 
     // always add system default devices for input and output as first entry
     lNumDevs                 = 0;
-    strDriverNames[lNumDevs] = "System Default In/Out Devices";
+    strDriverNames[lNumDevs] = "System Default In/Out Devices"; // TODO: why can't this be translated?
```
ann0see, line 106: According to the language style guide shouldn't this be sound hardware?

```suggestion
    strDriverNames[lNumDevs] = "System default in/out hardware"; // TODO: why can't this be translated?
```

mac/sound.cpp
```diff
@@ -103,24 +103,24 @@ void CSound::GetAvailableInOutDevices()
...
 
     iPropertySize               = sizeof ( AudioDeviceID );
     stPropertyAddress.mSelector = kAudioHardwarePropertyDefaultInputDevice;
 
     if ( AudioObjectGetPropertyData ( kAudioObjectSystemObject, &stPropertyAddress, 0, NULL, &iPropertySize, &audioInputDevice[lNumDevs] ) )
     {
-        throw CGenErr ( tr ( "CoreAudio input AudioHardwareGetProperty call failed. "
-                             "It seems that no sound card is available in the system." ) );
+        throw CGenErr ( tr ( "No sound card is available in your system. "
```
ann0see, line 113:
```suggestion
        throw CGenErr ( tr ( "No sound hardware is available in your system. "
```

mac/sound.cpp
```diff
@@ -103,24 +103,24 @@ void CSound::GetAvailableInOutDevices()
...
     }
 
     iPropertySize               = sizeof ( AudioDeviceID );
     stPropertyAddress.mSelector = kAudioHardwarePropertyDefaultOutputDevice;
 
     if ( AudioObjectGetPropertyData ( kAudioObjectSystemObject, &stPropertyAddress, 0, NULL, &iPropertySize, &audioOutputDevice[lNumDevs] ) )
     {
-        throw CGenErr ( tr ( "CoreAudio output AudioHardwareGetProperty call failed. "
-                             "It seems that no sound card is available in the system." ) );
+        throw CGenErr ( tr ( "No sound card is available in the system. "
```
ann0see, line 122:
```suggestion
        throw CGenErr ( tr ( "No sound hardwaree is available in the system. "
```

mac/sound.cpp
```diff
@@ -281,7 +281,7 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName, bool )
     // if the selected driver was not found, return an error message
     if ( iDriverIdx == INVALID_INDEX )
     {
-        return tr ( "The current selected audio device is no longer present in the system." );
+        return tr ( "The currently selected audio device is no longer present. Please check your audio device." );
```
ann0see, line 284:
```suggestion
        return tr ( "The currently selected sound hardware is no longer present. Please check your audio device." );
```

mac/sound.cpp
```diff
@@ -463,7 +463,7 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
          ( !( CurDevStreamFormat.mFormatFlags & kAudioFormatFlagIsPacked ) ) )
     {
         return tr ( "The audio input stream format for this audio device is "
```
ann0see, line 465:
```suggestion
        return tr ( "The audio input stream format for this sound hardware is "
```

This suggestion sounds strange to me but would be needed if we follow the style guide.

src/soundbase.cpp
```diff
@@ -166,12 +166,9 @@ QString CSoundBase::SetDev ( const QString strDevName )
         if ( !vsErrorList.isEmpty() )
         {
             // create error message with all details
-            QString sErrorMessage = "<b>" + tr ( "No usable " ) + strSystemDriverTechniqueName +
-                                    tr ( " audio device "
-                                         "(driver) found." ) +
+            QString sErrorMessage = "<b>" + QString ( tr ( "No usable %1 audio device found." ) ) .arg ( strSystemDriverTechniqueName ) +
```
ann0see, line 169:
```suggestion
            QString sErrorMessage = "<b>" + QString ( tr ( "No usable %1 sound hardware found." ) ) .arg ( strSystemDriverTechniqueName ) +
```

"No usable CoreAudio sound hardware"/"No useable ASIO sound hardware" sounds strange too. Do we even need the API here?

src/soundbase.cpp
```diff
@@ -184,14 +181,14 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
-            sErrorMessage = sErrorMessage + "<br/>" + tr ( "Do you want to open the ASIO driver setups?" );
+            sErrorMessage = sErrorMessage + "<br/>" + tr ( "Do you want to open the ASIO driver setup to change driver settings?" );
 
             if ( QMessageBox::Yes == QMessageBox::information ( nullptr, APP_NAME, sErrorMessage, QMessageBox::Yes | QMessageBox::No ) )
             {
                 LoadAndInitializeFirstValidDriver ( true );
             }
 
-            sErrorMessage = APP_NAME + tr ( " could not be started because of audio interface issues." );
+            sErrorMessage = QString ( tr ( " Can't start %1. Please restart %1 and check/reconfigure your audio settings." ) );
```
ann0see, line 191:
```suggestion
            sErrorMessage = QString ( tr ( "Can't start %1. Please restart %1 and check/reconfigure your audio settings." ) );
```

windows/sound.cpp
```diff
@@ -54,7 +54,7 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName, bool bOpenDrive
     // if the selected driver was not found, return an error message
     if ( iDriverIdx == INVALID_INDEX )
     {
-        return tr ( "The current selected audio device is no longer present in the system." );
+        return tr ( "The selected audio device is no longer present in the system. Please check your audio device." );
```
ann0see, line 57:
```suggestion
        return tr ( "The selected sound hardware is no longer present in the system. Please check your audio device." );

```

windows/sound.cpp
```diff
@@ -143,9 +144,9 @@ QString CSound::CheckDeviceCapabilities()
     if ( ( CanSaRateReturn == ASE_NoClock ) || ( CanSaRateReturn == ASE_NotPresent ) )
     {
         // return error string
-        return tr ( "The audio device does not support the "
-                    "required sample rate. The required sample rate is: " ) +
-               QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ) + " Hz";
+        return QString ( tr ( "The selected audio device is incompatible "
```
ann0see, line 147:
```suggestion
        return QString ( tr ( "The selected sound hardware is incompatible "

```

windows/sound.cpp
```diff
@@ -154,15 +155,9 @@ QString CSound::CheckDeviceCapabilities()
...
-        return tr ( "The audio device does not support setting the required sampling "
-                    "rate. This error can happen if you have an audio interface like the "
-                    "Roland UA-25EX where you set the sample rate with a hardware switch "
-                    "on the audio device. If this is the case, please change the sample rate "
-                    "to " ) +
-               QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ) +
-               tr ( " Hz on the "
-                    "device and restart the " ) +
-               APP_NAME + tr ( " software." );
+        return QString ( tr ( "The current audio device configuration is incompatible "
```
ann0see, line 158:
```suggestion
        return QString ( tr ( "The sound hardware configuration is incompatible "

```

windows/sound.cpp
```diff
@@ -171,10 +166,8 @@ QString CSound::CheckDeviceCapabilities()
     if ( ( lNumInChan < NUM_IN_OUT_CHANNELS ) || ( lNumOutChan < NUM_IN_OUT_CHANNELS ) )
     {
         // return error string
-        return tr ( "The audio device does not support the "
-                    "required number of channels. The required number of channels "
-                    "for input and output is: " ) +
-               QString().setNum ( NUM_IN_OUT_CHANNELS );
+        return QString ( tr ( "The selected audio device is incompatible since it does not support "
```
ann0see, line 169:
```suggestion
        return QString ( tr ( "The selected sound hardware is incompatible since it does not support "

```

windows/sound.cpp
```diff
@@ -207,7 +200,8 @@ QString CSound::CheckDeviceCapabilities()
         if ( !CheckSampleTypeSupported ( channelInfosInput[i].type ) )
         {
             // return error string
-            return tr ( "Required audio sample format not available." );
+            return tr ( "The selected audio device is incompatible since "
```
ann0see, line 203:
```suggestion
            return tr ( "The selected sound hardware is incompatible since "

```

windows/sound.cpp
```diff
@@ -237,7 +231,8 @@ QString CSound::CheckDeviceCapabilities()
         if ( !CheckSampleTypeSupported ( channelInfosOutput[i].type ) )
         {
             // return error string
-            return tr ( "Required audio sample format not available." );
+            return tr ( "The selected audio device is incompatible since "
```
ann0see, line 234:
```suggestion
            return tr ( "The selected sound hardware is incompatible since "

```

ann0see (commented):
Went through it again...

windows/sound.cpp
```diff
@@ -143,9 +144,9 @@ QString CSound::CheckDeviceCapabilities()
     if ( ( CanSaRateReturn == ASE_NoClock ) || ( CanSaRateReturn == ASE_NotPresent ) )
     {
         // return error string
-        return tr ( "The audio device does not support the "
-                    "required sample rate. The required sample rate is: " ) +
-               QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ) + " Hz";
+        return QString ( tr ( "The selected audio device is incompatible "
+        "since it does not support a sample rate of %1 Hz. Please select another "
+        "device." ) ) .arg ( SYSTEM_SAMPLE_RATE_HZ );
```
ann0see, line 149: Do we need to convert SYSTEM_SAMPLE_RATE_HZ to a QString?

mac/sound.cpp
gilgongo: Ah yes - maybe :-) I added some definitions to the style guide in the hope that they would be possible in the actual text we have. So if the style guide rules make sense in this case, we should apply them.

linux/sound.cpp
mulyaj: Hm. That's tricky. 
At the moment, it's a dead-end error. Is there a general solution for this? Rebooting?

linux/sound.cpp
```diff
@@ -97,7 +87,7 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
 
         if ( input_port_midi == nullptr )
         {
-            throw CGenErr ( tr ( "The Jack port registering failed." ) );
+            throw CGenErr ( tr ( "The JACK port registering failed." ) );
```
mulyaj, line 90: Dead-end error here too.

mac/sound.cpp
gilgongo: Yes, the trouble with style guide vocabulary rules is that don't always fit the context. I'd say it would read a bit better if it was ".. for your sound hardware" perhaps.

linux/sound.cpp
pljones: It's "stop Jamulus, stop JACK, start JACK and see if another client can connect at 48KHz (e.g. Audacity)" or something like that.  But you're into diagnosing a JACK failure, really, rather than a Jamulus failure.

linux/sound.cpp
```diff
@@ -64,16 +61,9 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
...
-                        tr ( "You can "
-                             "use a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QJackCtl</a></i> "
-                             "to adjust the Jack server sample rate." ) +
-                        "<br>" +
-                        tr ( "Make sure to set the "
-                             "Frames/Period to a low value like " ) +
-                        QString().setNum ( DOUBLE_SYSTEM_FRAME_SIZE_SAMPLES ) + tr ( " to achieve a low delay." ) );
+        throw CGenErr ( QString ( tr ( "The JACK server isn't running at a sample rate of <b>%1 Hz</b>. Please use "
+                        "a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QJackCtl</a></i> to set the "
+                        "the JACK server sample rate to %1 Hz." ) ) .arg ( SYSTEM_SAMPLE_RATE_HZ ) );
```
pljones, line 66: If we have a general "Connecting Jamulus to JACK" page on the wiki, it might be worth linking that from each message, too.

linux/sound.cpp
pljones: "Check the JACK output".

linux/sound.cpp
pljones: "Check the JACK output".

linux/sound.cpp
```diff
@@ -210,9 +200,9 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
     // without a Jack server, Jamulus makes no sense to run, throw an error message
     if ( bJackWasShutDown )
     {
-        throw CGenErr ( tr ( "The Jack server was shut down. This software "
-                             "requires a Jack server to run. Try to restart the software to "
-                             "solve the issue." ) );
+        throw CGenErr ( QString ( tr ( "The JACK server was shut down but %1 "
```
pljones, line 203: "If the error persists, review the JACK output".

mac/sound.cpp
pljones: What is the string used for?  It appears to be being passed out of Jamulus to some other system.  How is that system using the string?  Is the value of the string important?

mac/sound.cpp
```diff
@@ -150,7 +150,7 @@ void CSound::GetAvailableInOutDevices()
             // in range
             if ( bIsInput_i && bIsOutput_j && ( lNumDevs < MAX_NUMBER_SOUND_CARDS ) )
             {
-                strDriverNames[lNumDevs] = "in: " + strDeviceName_i + "/out: " + strDeviceName_j;
+                strDriverNames[lNumDevs] = "in: " + strDeviceName_i + "/out: " + strDeviceName_j; //TODO: why can't this be translated?
```
pljones, line 153: Is it a UI string?  If not, do not translate.

pljones: See line 275, explicitly doing a string compare, for example.

mac/sound.cpp
pljones, line 398: I wouldn't say `does not support the a sample rate`... and then `try to set the sample rate` -- it sounds futile.

How about:
`The sample rate on the current input device is not %1.  Please select another device or set the same rate.`

mac/sound.cpp
```diff
@@ -413,11 +413,11 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
...
-            return QString ( tr ( "Current system audio output device sample "
-                                  "rate of %1 Hz is not supported. Please open the Audio-MIDI-Setup in "
-                                  "Applications->Utilities and try to set a sample rate of %2 Hz." ) )
-                .arg ( static_cast<int> ( outputSampleRate ) )
-                .arg ( SYSTEM_SAMPLE_RATE_HZ );
+            return QString ( tr ( "The selected output device is incompatible "
+                                  "since it does not support the a sample rate of %1 Hz. Please select another "
+                                  "output device or try to set the sample rate to %1 Hz "
+                                  "manually via Audio-MIDI-Setup (in Applications->Utilities)." ) )
+                                  .arg ( QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ) );
```
pljones, line 420: Same as above.

mac/sound.cpp
pljones: `The stream format on the current input device is not %1. Please select another device or reconfigure.`

I really don't like `hardware`.  From Jamulus's perspective, it's just the audio source or audio sink.  We have no idea what's on the end of the device driver.

gilgongo: We could maybe say in the guidelines that "device" is interchangeable with "hardware" if needed? The former is less newbie friendly, so would prefer "hardware" in most cases, but here perhaps "device" is more helpful maybe the OS refers to "devices" too?

windows/sound.cpp
gilgongo: So here, would we consider using device after all? As per https://github.com/jamulussoftware/jamulus/pull/1732#discussion_r637388198

mac/sound.cpp
pljones: In Windows, click the speaker icon in the tooltray to pop open the the volume, then click the "^" to pop up the selector list -- it says "Select playback device".  So Windows users will understand "recording device" (for an input) and "playback device" (for an output).

pljones: The overall control is "Open Sound settings", where "Output"->"Choose your output device" and "Input"->"Choose your input  device" live.

windows/sound.cpp
pljones: No, QString is rendering the args for you.

mac/sound.cpp
gilgongo: So we could modify what we have in the guide so that "device" would be OK along with related terms:

"Device" to refer to audio hardware seen by the OS. "No device selected", "Input device”, "Choose the device".

“Audio interface” to refer to external devices. “If your audio interface comes with a driver”, “Your audio interface may be set to ‘monitor’”.

"Sound card" to refer to built-in audio devices. "Your computer's sound card".

“Sound hardware” refers generically to both internal and external devices. “Your sound hardware will introduce some latency”, “Most sound hardware can be used”.

linux/sound.cpp
ann0see: Not sure?

mac/sound.cpp
ann0see: Yes. I think the problem here is that some devices don't support setting the sample rate automatically (--> in fact they are compatible with 48kHz) while others are set to a fixed sample rate which is not supported by Jamulus. The user should know:

1. His device might be incompatible --> he needs to use (=buy) another device
2. His device might be compatible but needs to be set to 48kHz manually

mac/sound.cpp
pljones: Line 275 says `if ( strDriverName.compare ( strDriverNames[i] ) == 0 )`, so leave `strDriverNames[lNumDevs]` alone.

mac/sound.cpp
pljones: Well, this is `mac/sound.cpp`, so it should follow Apple MacOS conventions, rather than Microsoft Windows conventions.  We should follow OS language conventions in OS-specific messages.

mac/sound.cpp
ann0see: I must admit that I haven't looked at the code yet.

src/soundbase.cpp
```diff
@@ -184,14 +181,14 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
-            sErrorMessage = sErrorMessage + "<br/>" + tr ( "Do you want to open the ASIO driver setups?" );
+            sErrorMessage = sErrorMessage + "<br/>" + tr ( "Do you want to open the ASIO driver setup to change driver settings?" );
 
             if ( QMessageBox::Yes == QMessageBox::information ( nullptr, APP_NAME, sErrorMessage, QMessageBox::Yes | QMessageBox::No ) )
             {
                 LoadAndInitializeFirstValidDriver ( true );
             }
 
-            sErrorMessage = APP_NAME + tr ( " could not be started because of audio interface issues." );
+            sErrorMessage = QString ( tr ( "Can't start %1. Please restart %1 and check/reconfigure your audio settings." ) );
```
pljones, line 191: Missing `.arg()`.

ann0see: Done.

src/soundbase.cpp
pljones: This error box is going to be copy/pasted by the user when they try to get help, hence all the error messages being listed.

Having the `strSystemDriverTechniqueName` will be useful to whoever reads the request.

(Does this ever appear?)

src/soundbase.cpp
pljones, line 184: Code consistency - use the `+=` operator to append to a QString.

pljones: The offer is only present for ASIO.

What's _far_ worse is line 191 (old and new) - they throw away this error, rather than letting CGenErr throw it, like for other drivers.

src/soundbase.cpp
```diff
@@ -184,14 +181,14 @@ QString CSoundBase::SetDev ( const QString strDevName )
             // to be able to access the ASIO driver setup for changing, e.g., the sample rate, we
             // offer the user under Windows that we open the driver setups of all registered
             // ASIO drivers
-            sErrorMessage = sErrorMessage + "<br/>" + tr ( "Do you want to open the ASIO driver setups?" );
+            sErrorMessage = sErrorMessage + "<br/>" + tr ( "Do you want to open the ASIO driver setup to change driver settings?" );
 
             if ( QMessageBox::Yes == QMessageBox::information ( nullptr, APP_NAME, sErrorMessage, QMessageBox::Yes | QMessageBox::No ) )
             {
                 LoadAndInitializeFirstValidDriver ( true );
```
pljones, line 188: What if this fixes the problem?  We still call `thow CGenErr`?

ann0see: I think I mentioned something like this somewhere else (talked about strStat as far as I know). 

Probably worth another issue & more investigation.

linux/sound.cpp
```diff
@@ -37,18 +37,15 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
         serverName = "default";
     }
     qInfo() << qUtf8Printable (
-        QString ( "Connecting to jack \"%1\" instance (use the JACK_DEFAULT_SERVER environment variable to change this)." ).arg ( serverName ) );
+        QString ( "Connecting to JACK \"%1\" instance (use the JACK_DEFAULT_SERVER environment variable to change this)." ).arg ( serverName ) );
```
pljones, line 40: We use APP_NAME for "Jamulus".  I think it would help translators to put "JACK" into a `#define` and use that throughout, too -- probably for a separate issue, though.

linux/sound.cpp
ann0see: Yes. This would probably be something for the community KB?

src/soundbase.cpp
ann0see: Yes, It does appear for users without proper sound hardware and only ASIO4ALL.

windows/sound.cpp
ann0see: Great. Thanks. There are some places where it is converted.

mac/sound.cpp
ann0see: Ok. So "stick to naming conventions your OS uses" should be added to the google doc probably?

src/soundbase.cpp
ann0see: Fixed

src/soundbase.cpp
pljones: If it only appears with ASIO4ALL, then it shouldn't be in soundbase.cpp, it should be in windows/sound.cpp.  Again, not for this change....

ann0see: Probably. I think there was an issue for this feature and probably Volker added this as a hotfix.

linux/sound.cpp
ann0see: @jujudusud since you’re working on JACK and the documentation, I think that’s something for you?

windows/sound.cpp
```diff
@@ -103,12 +104,12 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName, bool bOpenDrive
...
             QMessageBox::question ( nullptr,
                                     APP_NAME,
-                                    "Are you done with your ASIO driver settings of device " + GetDeviceName ( iDriverIdx ) + "?",
+                                    "Are you done with your ASIO driver settings of " + GetDeviceName ( iDriverIdx ) + "?",
                                     QMessageBox::Yes );
         }
 
         // driver cannot be used, clean up
-        asioDrivers->removeCurrentDriver();
+        asioDrivers->removeCurrentDriver(); // comment by ann0see: why do we clean up the driver even if we changed the driver settings?
```
ann0see, line 112: Worth a separate issue?

pljones:
Given that there are still unaddressed comments, it would be easier to have separate commits until just before the end.  Then rebase squashing locally and force push the result.

linux/sound.cpp
ann0see: JACK issue

Just checked this error and it shows that it's from Jamulus

ann0see:
Ok. Things to do:

- [ ] Agree on what to do with "audio device"/"Audio hardware"/...
- [ ] Open issues on the windows sound handling code discovered by @pljones 
- [ ] Do a final review

linux/sound.cpp
ann0see: I think this is resolved now.

linux/sound.cpp
```diff
@@ -210,9 +206,10 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
     // without a Jack server, Jamulus makes no sense to run, throw an error message
     if ( bJackWasShutDown )
     {
-        throw CGenErr ( tr ( "The Jack server was shut down. This software "
-                             "requires a Jack server to run. Try to restart the software to "
-                             "solve the issue." ) );
+        throw CGenErr ( QString ( tr ( "The JACK server was shut down but %1 "
```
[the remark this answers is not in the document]
ann0see: Thanks. Changed this one.

src/soundbase.cpp
```diff
@@ -154,10 +154,9 @@ QString CSoundBase::SetDev ( const QString strDevName )
         if ( !strDevName.isEmpty() )
         {
             strReturn = tr ( "The previously selected audio device "
-                             "is no longer available or the audio driver properties have changed to a state which "
-                             "is incompatible with this software. We now try to find a valid audio device. This new "
-                             "audio device might cause audio feedback. So, before connecting to a server, please "
-                             "check the audio device setting." );
+                             "is no longer available or the driver has changed to an incompatible state."
+                             "We'll try to find a valid audio device which might create feedback. "
```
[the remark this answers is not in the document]
ann0see: > It sounds like Jamulus is looking for a device that will create feedback (which I don't think is the intention :D).

Changed that ;-)

linux/sound.cpp
pljones: Did an issue get raised for this?

ann0see: Not yet

mac/sound.cpp
ann0see: @mulyaj to discuss this one today

linux/sound.cpp
ann0see: Done: https://github.com/jamulussoftware/jamulus/issues/1811

windows/sound.cpp
ann0see: See https://github.com/jamulussoftware/jamulus/issues/1812

src/soundbase.cpp
ann0see: See https://github.com/jamulussoftware/jamulus/issues/1812

mac/sound.cpp
ann0see: To be discussed. Only open question on this PR. Therefore waiting for a concensus.

mac/sound.cpp
```diff
@@ -462,8 +460,8 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
          ( CurDevStreamFormat.mBitsPerChannel != 32 ) || ( !( CurDevStreamFormat.mFormatFlags & kAudioFormatFlagIsFloat ) ) ||
          ( !( CurDevStreamFormat.mFormatFlags & kAudioFormatFlagIsPacked ) ) )
     {
-        return tr ( "The audio input stream format for this audio device is "
-                    "not compatible with this software." );
+        return tr ( "The stream format on the current input device isn't "
+                    "compatible with this software. Please select another device." );
```
ann0see, line 464:
```suggestion
        return QString ( tr ( "The stream format on the current input device isn't "
                    "compatible with this software. Please select another device." ) ) .arg ( APP_NAME );
```

mac/sound.cpp
```diff
@@ -473,8 +471,8 @@ QString CSound::CheckDeviceCapabilities ( const int iDriverIdx )
          ( CurDevStreamFormat.mBitsPerChannel != 32 ) || ( !( CurDevStreamFormat.mFormatFlags & kAudioFormatFlagIsFloat ) ) ||
          ( !( CurDevStreamFormat.mFormatFlags & kAudioFormatFlagIsPacked ) ) )
     {
-        return tr ( "The audio output stream format for this audio device is "
-                    "not compatible with this software." );
+        return tr ( "The stream format on the current output device isn't "
+                    "compatible with this software. Please select another device." );
```
ann0see, line 475:
```suggestion
        return QString ( tr ( "The stream format on the current output device isn't "
                    "compatible with %1. Please select another device." ) ) .arg ( APP_NAME );
```

windows/sound.cpp
ann0see: To be discussed. Only open question on this PR. Therefore waiting for a concensus.

windows/sound.cpp
ann0see, line 112:
```suggestion
        asioDrivers->removeCurrentDriver();
```

src/soundbase.cpp
ann0see: See https://github.com/jamulussoftware/jamulus/issues/1812

ann0see:
Ok. So there are still two missing items for this PR before it gets merged:

- [ ] Agree on what to do with "audio device"/"Audio hardware"/...
- [ ] Fix clang format

softins:
I don't have a strong view, but would prefer "audio device" over "sound hardware". And definitely not "sound card", as most audio devices are no longer "cards".

ann0see:
I‘m also not totally sure. Audio device sounds OK although the language guide (style and tone) I mentioned above wants a more specific (?) word. 

 @mulyaj and @gilgongo did it and are open to suggestions I suppose;-)

softins:
> I‘m also not totally sure. Audio device sounds OK although the language guide (style and tone) I mentioned above wants a more specific (?) word.
> 
> @mulyaj and @gilgongo did it and are open to suggestions I suppose;-)

Ah, ok. I hadn't read that doc (at least, not recently). I'm ok with "sound hardware" if we want to be generic to cover internal and external interfaces.

ann0see:
After thinking a bit more, I'd personally prefer staying with sound device since this includes:

1. internal devices
2. external devices
3. virtual devices (JACK, Blackhole,...)

while hardware is well, not software like JACK...

ann0see:
Can we merge this and focus on the audio device/unification of language in another PR?

pljones:
1 & 2 are all I could remember without going through all the closed off review comments again.

linux/sound.cpp
```diff
@@ -64,16 +61,10 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
...
-                        " <b>" + QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ) + " Hz</b>. " +
-                        tr ( "You can "
-                             "use a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QJackCtl</a></i> "
-                             "to adjust the Jack server sample rate." ) +
-                        "<br>" +
-                        tr ( "Make sure to set the "
-                             "Frames/Period to a low value like " ) +
-                        QString().setNum ( DOUBLE_SYSTEM_FRAME_SIZE_SAMPLES ) + tr ( " to achieve a low delay." ) );
+        throw CGenErr ( QString ( tr ( "The JACK server isn't running at a sample rate of <b>%1 Hz</b>. Please use "
+                                       "a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QJackCtl</a></i> to set the "
```
passing, line 65:
```suggestion
                                       "a tool like <i><a href=\"https://qjackctl.sourceforge.io\">QjackCtl</a></i> to set the "
```
(as on its website)

linux/sound.cpp
```diff
@@ -108,7 +104,8 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
     // tell the JACK server that we are ready to roll
     if ( jack_activate ( pJackClient ) )
     {
-        throw CGenErr ( tr ( "Cannot activate the Jack client." ) );
+        throw CGenErr ( QString ( tr ( "Can't activate the JACK client. This is a probably an error with JACK. Please check the JACK output." ) )
```
passing, line 107:
```suggestion
        throw CGenErr ( QString ( tr ( "Can't activate the JACK client. This is probably an error with JACK. Please check the JACK output." ) )
```

linux/sound.cpp
```diff
@@ -210,9 +207,11 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
     // without a Jack server, Jamulus makes no sense to run, throw an error message
     if ( bJackWasShutDown )
     {
-        throw CGenErr ( tr ( "The Jack server was shut down. This software "
-                             "requires a Jack server to run. Try to restart the software to "
-                             "solve the issue." ) );
+        throw CGenErr ( QString ( tr ( "The JACK server was shut down. %1 "
+                                       "requires a JACK to run. Please restart %1 to "
```
passing, line 211: "requires a JACK server" or "requires JACK" ?

ann0see: probably a JACK server?

passing:
```suggestion
                                       "requires a JACK server to run. Please restart %1 to "
```

ann0see:
@gilgongo could you please review the code so that we can squash and merge this?

linux/sound.cpp
```diff
@@ -37,18 +37,15 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
...
     pJackClient = jack_client_open ( jackClientName, JackNullOption, &JackStatus );
 
     if ( pJackClient == nullptr )
     {
-        throw CGenErr ( tr ( "The Jack server is not running. This software "
-                             "requires a Jack server to run. Normally if the Jack server is "
-                             "not running this software will automatically start the Jack server. "
-                             "It seems that this auto start has not worked. Try to start the Jack "
-                             "server manually." ) );
+        throw CGenErr ( tr ( "The JACK server couldn't be started automatically. "
```
gilgongo, line 47: We refer to "JACK server" and "JACK" on its own here. It it easier to just to refer to "JACK" and not add the word "server"? Or are they two different things (in which case this message is confusing!)

linux/sound.cpp
```diff
@@ -87,7 +78,10 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
 
     if ( ( input_port_left == nullptr ) || ( input_port_right == nullptr ) || ( output_port_left == nullptr ) || ( output_port_right == nullptr ) )
     {
-        throw CGenErr ( tr ( "The Jack port registering failed." ) );
+        throw CGenErr ( QString ( tr ( "The JACK port registering failed. This is a probably an error with JACK. Please stop %1 and JACK. "
```
gilgongo, line 81:
```suggestion
        throw CGenErr ( QString ( tr ( "The JACK port registration failed. This is probably an error with JACK. Please stop %1 and JACK. "
```

linux/sound.cpp
```diff
@@ -97,7 +91,9 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
 
         if ( input_port_midi == nullptr )
         {
-            throw CGenErr ( tr ( "The Jack port registering failed." ) );
+            throw CGenErr ( QString ( tr ( "The JACK port registering failed. This is a probably an error with JACK. Please stop %1 and JACK. "
```
gilgongo, line 94:
```suggestion
            throw CGenErr ( QString ( tr ( "The JACK port registration failed. This is probably an error with JACK. Please stop %1 and JACK. "
```

linux/sound.cpp
```diff
@@ -97,7 +91,9 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
...
+                                           "Afterwards check if another MIDI program can connect to JACK." ) )
```
gilgongo, line 95:
```suggestion
                                           "Afterwards, check if another MIDI program can connect to JACK." ) )
```

linux/sound.cpp
```diff
@@ -210,9 +207,11 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
     // without a Jack server, Jamulus makes no sense to run, throw an error message
     if ( bJackWasShutDown )
     {
-        throw CGenErr ( tr ( "The Jack server was shut down. This software "
-                             "requires a Jack server to run. Try to restart the software to "
-                             "solve the issue." ) );
+        throw CGenErr ( QString ( tr ( "The JACK server was shut down. %1 "
+                                       "requires a JACK server to run. Please restart %1 to "
+                                       "start JACK again. If the error persists, please review "
```
gilgongo, line 212: Given the fact that "review" is extremely vague, I'd just remove it. 

```suggestion
                                       "start JACK again. "
```

mac/sound.cpp
gilgongo: Shouldn't really use passive voice if we can help it. "Your system has no available sound hardware" would be the preferred grammar.

mac/sound.cpp
gilgongo: "The system has no available sond hardware"

src/soundbase.cpp
```diff
@@ -113,7 +113,7 @@ QString CSoundBase::SetDev ( const QString strDevName )
                 LoadAndInitializeDriver ( strCurDevName, false );
 
                 // store error return message
-                strReturn = QString ( tr ( "The selected audio device could not be used "
+                strReturn = QString ( tr ( "The selected audio device can't be used "
```
gilgongo, line 116:
```suggestion
                strReturn = QString ( tr ( "Can't use the selected audio device "
```

windows/sound.cpp
```diff
@@ -73,7 +73,7 @@ QString CSound::LoadAndInitializeDriver ( QString strDriverName, bool bOpenDrive
     {
         // clean up and return error string
         asioDrivers->removeCurrentDriver();
-        return tr ( "The audio driver could not be initialized." );
+        return tr ( "The audio driver couldn't be initialised. Check if the audio device is plugged in and verify your driver settings." );
```
gilgongo, line 76:
```suggestion
        return tr ( "Couldn't initialise the audio driver. Check if your audio hardware is plugged in and verify your driver settings." );

```

windows/sound.cpp
```diff
@@ -154,15 +155,10 @@ QString CSound::CheckDeviceCapabilities()
...
-                    "Roland UA-25EX where you set the sample rate with a hardware switch "
-                    "on the audio device. If this is the case, please change the sample rate "
-                    "to " ) +
-               QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ) +
-               tr ( " Hz on the "
-                    "device and restart the " ) +
-               APP_NAME + tr ( " software." );
+        return QString ( tr ( "The current audio device configuration is incompatible "
+                              "because the sample rate couldn't be set to %1 Hz. Please check for a hardware switch or "
+                              "driver setting to set the sample rate manually and restart the app." ) )
```
gilgongo, line 160: What does "the app" mean here? If it's Jamulus, we should say that.

gilgongo (changes requested):
Probably need to use more active case in the English here, for clarity.

linux/sound.cpp
ann0see: I think it’s just a technical description and probably there’s no need to talk about a server.

ann0see:
```suggestion
        throw CGenErr ( tr ( "JACK couldn't be started automatically. "
```

windows/sound.cpp
ann0see: Done.

gilgongo (approved):

ann0see:
CHANGELOG: Rewrote multiple error messages to improve UX.

Issue #1702, opened 2021-05-16
Integrate clang-format into checks, docs and tooling


hoffie:
With the initial clang-format definition and initial run with the config having been merged (#1127), the related issue (#901) was auto-closed, although its scope was greater than what the (huge) PR did. Therefore, I'm opening this follow-up issue for tracking the remaining parts for the actual integration.

**Has this feature been discussed and generally agreed?**

Yes, here: https://github.com/jamulussoftware/jamulus/issues/901#issuecomment-779437954
Some details still to be decided upon.

**Describe the solution you'd like**
 
- [x] Make a final decision about the type of Github action (shamelessly copied from @passing's https://github.com/jamulussoftware/jamulus/issues/901#issuecomment-779437954)
  1. Enforce style (auto-update PRs -- not sure if/how this is possible and feels like)
  2. Check style and block merge on invalid style (but I assume that admins can override in any case, needs to be checked).
  3. Check style, but don't block merge on invalid style.
- [x] Implement the Github action
- [x] Check if and what tools (Makefile?) or docs (CONTRIBUTING?) should be updated/extended for local clang checks

hoffie:
I'm adding my personal opinion as a comment:

1. Enforce style:
    - Advantage: Contributors do not need clang(-format), we don't have to remind them about applying it.
    - Advantage: We could use the latest clang-format (12) features
    - Disadvantage: I've never seen such an integration as part of a PR check. I don't know if and how it can be done and what consequences it would have (would contributors have to force-push when updating their branch?). Maybe @passing knows more about that?
    - If Enforcing only works as a Github Action (i.e. after merge), it would have the (huge) downside that code review will still happen on unfixed code, rendering some of the basic advantages useless
2. Check style and block merge on invalid style
    - Depending on how (1) works out in practice, this would be my preferred option (always assuming that admins can override in case of emergencies)
3. Check style, but don't block merge on invalid style.
    - Disadvantage: Contributors (or us) might get lax about it. I do not like this option and I've just included it for completeness.

@jamulussoftware/maindevelopers @passing What do you think?

pljones:
For me,
> Check style and block merge on invalid style

is really the best option.  Any check written as a standard CI action can be overridden, as far as I understand.

passing:
I'd prefer the "enforce style" option but only if we can clear out the disadvantages to a certain degree:

[...]

PR #1735, opened 2021-05-21
added github workflow for style check with clang-format

passing:
https://github.com/jamulussoftware/jamulus/issues/1702
Code Style only gets checked, not fixed.
Example PR: https://github.com/passing/jamulus/pull/5

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,20 @@
+name: Style Check
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  clang-format:
+    name: Clang-Format
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: DoozyX/clang-format-lint-action@v0.12
+      with:
+        source: '.'
+        exclude: './libs'
+        extensions: 'cpp,h'
+        clangFormatVersion: 10


pljones (approved):

dcorson-ticino-com:
Please explain how this is to be used.

pljones:
When you submit a PR, your PR will fail (like https://github.com/passing/jamulus/pull/5) if the clang lint says your code doesn't comply with the style guide.

dcorson-ticino-com:
And how do I know how to change the code so it will pass?
Or do I simply blindly do a clang-format?

passing:
@dcorson-ticino-com 
when you click on 'Details' right of the failing check (e.g. on the test PR https://github.com/passing/jamulus/pull/5), you see the diff needed to make the Style Check pass:
image
```
--- ./src/server.cpp	(original)
+++ ./src/server.cpp	(reformatted)
@@ -28,7 +28,7 @@
 #ifdef _WIN32
 CHighPrecisionTimer::CHighPrecisionTimer ( const bool bNewUseDoubleSystemFrameSize ) : bUseDoubleSystemFrameSize ( bNewUseDoubleSystemFrameSize )
 {
-        int bad_aligned;
+    int bad_aligned;
 
     // add some error checking, the high precision timer implementation only
     // supports 64 and 128 samples frame size at 48 kHz sampling rate
```

dcorson-ticino-com:
Ok, great.

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,20 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  clang-format:
+    name: Clang-Format
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: DoozyX/clang-format-lint-action@v0.12
```
hoffie, line 15: I'm rather new to Github Actions, but I think we should pin this third-party action to a specific sha1 hash. Did you review the code by this action, if not, could you do that?

(I'm aware that I'm asking for things which are not covered by existing actions either. I think it is important nevertheless and fixing it for existing workflows is on the agenda #1737).

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,20 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  clang-format:
+    name: Clang-Format
```
hoffie, line 11: Could you try to add a [`permission:` keyword](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token) with as little permissions as possible to run this action?

Currently, this is not necessary. If we switch to restrictive default permissions via #1737 though, we will need this. Maybe there are other users of this action where we could borrow the right set of permissions?

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,20 @@
+name: Style Check
```
hoffie, line 1:
```suggestion
name: Coding Style Check
```
?

.github/workflows/clang-format.yml
hoffie, line 11:
```suggestion
    name: clang-format
```
to match the name of the actual command being run?

hoffie (commented):
Thanks! :)

Additionally pinging @softins and @ann0see here for awareness.

hoffie:
I think this PR can be merged independent of the release (i.e. even during RC phase) as it does not touch any code of the final executables.

@dcorson-ticino-com 
> And how do I know how to change the code so it will pass?

I also think that we should add further documentation (CONTRIBUTING.md?) independent from this PR.

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  clang-format:
+    name: Clang-Format
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: DoozyX/clang-format-lint-action@v0.12
+      with:
+        source: '.'
+        exclude: './libs'
```
hoffie, line 18: We should also exclude `windows/nsProcess/` which is third-party code, I think. It seems like we forgot doing that when doing the first `clang-format` run.
@ann0see As you added that code -- what do you think? I think I'd prefer leaving it untouched. That would mean it should be ignored here and the formatting should be reverted there...?

ann0see: Yes. That’s 3rd party code. I don’t think modifying it does any harm but of course, we can exclude it.

.github/workflows/clang-format.yml
passing: the workflow name `name: Style Check` and the job name `name: Clang-Format` in the end are displayed in the list of checks like this.
image
the style is called "Clang-Format" (https://releases.llvm.org/10.0.0/tools/clang/docs/ClangFormatStyleOptions.html#), the toolset "ClangFormat" https://releases.llvm.org/10.0.0/tools/clang/docs/ClangFormat.html and the command "clang-format".
I don't mind changing it, but if the name (line 11) is the same as in line 10 you could actually remove it

.github/workflows/clang-format.yml
hoffie: I'm mainly worrying about if we ever had to update it. Diffs would look strange at first glance (although this could be fixed by formatting the new code as well).

.github/workflows/clang-format.yml
hoffie: > the workflow name `name: Style Check` and the job name `name: Clang-Format` in the end are displayed in the list of checks like this.

Yep, I'm aware of it and that's why suggested modifying it slightly. I think "Coding Style Check" may be a bit clearer, especially as there is work on Language Style Guides in the project as well?

> the style is called "Clang-Format" (https://releases.llvm.org/10.0.0/tools/clang/docs/ClangFormatStyleOptions.html#), the toolset "ClangFormat" https://releases.llvm.org/10.0.0/tools/clang/docs/ClangFormat.html and the command "clang-format".

Ok. I'm not fussed about it, we can also keep it like that.

Maybe someone else can comment as well.

softins:
> When you submit a PR, your PR will fail (like [passing#5](https://github.com/passing/jamulus/pull/5)) if the clang lint says your code doesn't comply with the style guide.

Is that doing you out of a job, Peter? ;)

.github/workflows/clang-format.yml
passing: I had a look at the project (https://github.com/DoozyX/clang-format-lint-action) and that's fine in general.
Only downside are the binaries that are included here, so you basically have to trust them: https://github.com/DoozyX/clang-format-lint-action/tree/master/clang-format

.github/workflows/clang-format.yml
pljones: Yes, we should revert it -- I thought I'd caught that at the time but there was so much happening...

passing: I have created a separate PR for that: https://github.com/jamulussoftware/jamulus/pull/1751

pljones:
> 
> 
> > When you submit a PR, your PR will fail (like [passing#5](https://github.com/passing/jamulus/pull/5)) if the clang lint says your code doesn't comply with the style guide.
> 
> Is that doing you out of a job, Peter? ;)

Hopefully :)

.github/workflows/clang-format.yml
pljones: Based on their Language Style Guide, we'd be after "Coding Style Check / Verify Clang-Format compliance" for the check title, so:
- workflow ` .github/workflows/coding-style-checks.yml`, `name: Coding Style Check`
- `jobs` `verify-clang-format`, `name: Verify Clang-Format compliance`

passing: >     * workflow ` .github/workflows/coding-style-checks.yml`, `name: Coding Style Check`
>     * `jobs` `verify-clang-format`, `name: Verify Clang-Format compliance`
I have changed the names accordingly

.github/workflows/clang-format.yml
passing: > I think we should pin this third-party action to a specific sha1 hash

Had a closer look at the documentation as well, and to be sure we need to use the hash, as the tag could be reused.
Changed the action accordingly.

.github/workflows/clang-format.yml
passing: I have restricted the action now to
```
    permissions:
      contents: read
```

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,20 @@
+name: Coding Style Check
```
pljones, line 1: I'd still rather the filename were `coding-style-check.yml`.

I'm thinking we might want to XML lint the UI files, for example.  That could live as a separate job in the same file.

pljones (approved):

.github/workflows/clang-format.yml
hoffie: Sounds good, thanks! :)

.github/workflows/clang-format.yml
hoffie: > Only downside are the binaries that are included here, so you basically have to trust them: https://github.com/DoozyX/clang-format-lint-action/tree/master/clang-format

That's really not optimal, but as this action should work with read-only permission only, it'd be OK, I think. The repo is public anyway, so even if there was harmful code, it would not gain any additional permissions by running there.

hoffie (approved):
Thanks!

As this does not touch any actual Jamulus code, I'd say this would be fine to merge at any time.
I'll merge in the coming days if noone objects. If someone else feels like it, go for it ;)

ann0see (approved):
Merging this one too.

ann0see:
I think we'll see what other contributors think about this. It's certainly a big improvement.

PR #1755, opened 2021-05-24
Update unit of measure display in settings dialog

### LINKED ISSUE #1753: Unit of measure displayed twice when ping exceeds 500ms
**Describe the bug**
When the ping time exceeds 500ms in an active session, the audio/network setting dialog tab shows the unit of measuer twice.

**To Reproduce**
Increase the ping time while in a session.

**Expected behavior**
Unit of measuer should only be displayed once.

**Screenshots**
exceed-500ms-ping

**Operating system**
win 10

**Version of Jamulus**
r3.8.0rc1

**Additional context**
Seems to be caused as the unit of measure is now a separate item

henkdegroot:
Unit of measure in settings dialog now displayed always (regardless the client is connected or not) to remove update to "fixed" text.
Unit of measure removed from error text when exceeding 500ms.

Fixes #1753

### DIFF
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -399,11 +399,8 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     ledNetw->SetType ( CMultiColorLED::MT_INDICATOR );
     ledOverallDelay->SetType ( CMultiColorLED::MT_INDICATOR );
     lblUpstreamValue->setText ( "---" );
-    lblUpstreamUnit->setText ( "" );
     lblPingTimeValue->setText ( "---" );
-    lblPingTimeUnit->setText ( "" );
     lblOverallDelayValue->setText ( "---" );
-    lblOverallDelayUnit->setText ( "" );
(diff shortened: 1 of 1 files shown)


src/clientsettingsdlg.cpp
```diff
@@ -1009,23 +1006,20 @@ void CClientSettingsDlg::SetPingTimeResult ( const int iPingTime, const int iOve
     // a certain value
     if ( iPingTime > 500 )
     {
-        const QString sErrorText = "<font color=\"red\"><b>&#62;500 ms</b></font>";
+        const QString sErrorText = "<font color=\"red\"><b>&#62;500</b></font>";
```
pljones, line 1009: Oops! :)  I missed that...

pljones (approved):

jujudusud (approved):

hoffie (approved):
Thanks! Given that this PR avoids introducing a new bug compared to 3.7.0, I'd be inclined to merge for 3.8.0 and I'm tagging it as such.

Opinions?

henkdegroot:
@ann0see  do we still need this PR? As the "ms" has been removed already as part of #1762  (at least as far as I can tell).

ann0see:
No. I don’t think so

pljones:
We can re-open if #1762 doesn't go ahead.

PR #1758, opened 2021-05-25
Try to improve ping status update performance on Mac

softins:
This still needs more testing, but I've raised it as a PR so that the test artifacts can be generated.


ann0see:
Should we close this now since we have an alternative approach?

softins:
Yes, closing this as no longer needed

PR #1759, opened 2021-05-26
Add reference to website in Help

henkdegroot:
Add reference to jamulus website when help text is displayed.
Also add linebreak to the start of the text so in windows cmd the text start at a new line.

See also #1660

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -731,19 +731,22 @@ int main ( int argc, char** argv )
 \******************************************************************************/
 QString UsageArguments ( char** argv )
 {
-    return "\nUsage: " + QString ( argv[0] ) +
-           " [option] [optional argument]\n"
-           "\nGeneral options:\n"
+    // clang-format off
+    return "\n"
+           "Usage: " + QString ( argv[0] ) + " [option] [optional argument]\n"
+           "\n"
+           "General options:\n"
            "  -h, -?, --help        display this help text and exit\n"
-           "  -i, --inifile         initialization file name (not\n"
-           "                        supported for headless server mode)\n"
+           "  -i, --inifile         initialization file name\n"
+           "                        (not supported for headless server mode)\n"
            "  -n, --nogui           disable GUI\n"
            "  -p, --port            set the local port number\n"
            "  -Q, --qos             set the QoS value. Default is 128. Disable with 0\n"
            "                        (see the Jamulus website to enable QoS on Windows)\n"
            "  -t, --notranslation   disable translation (use English language)\n"
            "  -v, --version         output version information and exit\n"
-           "\nServer only:\n"
+           "\n"
+           "Server only:\n"
            "  -d, --discononquit    disconnect all clients on quit\n"
            "  -e, --directoryserver address of the directory server with which to register\n"
            "                        (or 'localhost' to host a server list on this server)\n"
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -731,7 +731,7 @@ int main ( int argc, char** argv )
 \******************************************************************************/
 QString UsageArguments ( char** argv )
 {
-    return "Usage: " + QString ( argv[0] ) +
+    return "\nUsage: " + QString ( argv[0] ) +
```
ann0see, line 734: Why is the new line needed here?

ann0see (commented):
Thank you!

src/main.cpp
pljones, line 734: I don't like the adding of leading `\n` in various places.  Why here, at all, too?

ann0see: Had the same question, but @henkdegroot explains it here:

>  Also add linebreak to the start of the text so in windows cmd the text start at a new line.

src/main.cpp
```diff
@@ -731,7 +731,7 @@ int main ( int argc, char** argv )
...
            " [option] [optional argument]\n"
            "\nGeneral options:\n"
```
pljones, line 736: I'd rather the extra `\n`s were on their own lines - it's easier to see the resultant layout, that way.

src/main.cpp
```diff
@@ -776,7 +776,9 @@ QString UsageArguments ( char** argv )
            "      --ctrlmidich      MIDI controller channel to listen\n"
            "      --clientname      client name (window title and jack client name)\n"
            "\nExample: " +
-           QString ( argv[0] ) + " -s --inifile myinifile.ini\n";
+           QString ( argv[0] ) + " -s --inifile myinifile.ini\n"
+           "\nFor more information and localized help see:\n"
+           "\nhttps://jamulus.io/wiki/Command-Line-Options\n";
```
pljones, line 781: Why the extra blank line?

src/main.cpp
pljones: Qt should fix their Windows support so we're still attached to the cmd session...  (Or we display the usage in a window for MSWindows - that's a common approach, too.)

henkdegroot: > Or we display the usage in a window for MSWindows - that's a common approach, too.

Indeed a common approach, but won't resolve it when you use a Windows Server without Destop :)
In #1660 it is suggested to use QCommandLineParser...however don't think that will change the behaviour in this case as it states:
_On Windows, QCommandLineParser uses message boxes to display usage information and errors if no console window can be obtained._

src/main.cpp
henkdegroot: command-output

Because to me, that made sense....but happy to remove.

src/main.cpp
henkdegroot: While I add this, are you okay with updating the following lines:

Current Line:
```  -e, --directoryserver address of the directory server with which to register```
Proposed New Line:
```  -e, --directoryserver address of the directory server to register with```

Current Line:
```
  -i, --inifile         initialization file name (not
                        supported for headless server mode)
```

Proposed New Line:
```
  -i, --inifile         initialization file name
                        (not supported for headless server mode)
```

pljones: > -e, --directoryserver  Proposed New Line:

I prefer the correct English current version :).

> -i, --inifile Proposed New Line:

Nicer layout, yes.

pljones (approved):

hoffie (approved):
Looks good to me, thanks.
Tagging for 3.8.1 unless someone thinks otherwise.

Therefore, **not merging** yet.

hoffie:
CHANGELOG: The --help text now links to the website for further documentation.

PR #1752, opened 2021-05-23
Update left-side gui section

henkdegroot:
Update left-side gui section and add accessibility markers to Red and Yellow Fancy Buffer and Delay LEDs.

Fixes #1729

PR #1762, opened 2021-05-26
Move ping/delay/jitter measurements from settings to main window

dcorson-ticino-com:
The values are not updated on the cleintsettings tab.
I don't think this is too ugly, so it could be workable.
Please test to see if it no longer has memory loss.
The only problem I see is with the stream data rate which is now not
shown.  I could imagine to use a static variable to store the last value
and only update when changed, which will be almost never.

I will start now to clean this up with the accessibility stuff, etc.
where it should be.  Will not change the functioning.


softins:
You definitely need the stream rate displayed somewhere, and it makes sense to do that in the original place (just having the one line in the Measurements box), since it is only when changing audio quality or mono/stereo that it changes.

I have been running this PR under the same conditions as my earlier tests, and it does not ramp up the memory or CPU usage, so that is certainly an improvement. It also looks quite nice having the details on the screen like that.

softins:
I've also tested running with the mixer window not on top:

1. With the settings window on top instead, and
2. With a non-Jamulus window on top.

In both cases, the memory and CPU usage remained stable. So it looks like this is the way to go. Just display the audio stream rate in the Settings window as before, and move the ping time, and overall delay to the main window as you have done.

ann0see:
I'd say this is a good solution to the issue. (It might also close another issue I didn't look for yet)


I just noticed that the UI (mixing board) starts to move around while the Ding/Delay changes.

src/clientdlg.cpp
```diff
@@ -1035,7 +1035,7 @@ void CClientDlg::OnTimerBuffersLED()
 
     // update the buffer LED and the general settings dialog, too
     ledBuffers->SetLight ( eCurStatus );
-    ClientSettingsDlg.SetStatus ( eCurStatus );
+//    ClientSettingsDlg.SetStatus ( eCurStatus );
```
ann0see, line 1038:
```suggestion
```

src/clientdlg.cpp
```diff
@@ -1258,6 +1252,16 @@ void CClientDlg::UpdateDisplay()
...
     }
+    if ( !pClient->IsRunning() )
+    {
+        // clear text labels with client parameters
+        lblPingVal->setText ( "---" );
+        lblPingUnit->setText ( "" );
+        lblDelayVal->setText ( "---" );
+        lblDelayUnit->setText ( "" );
+//        lblUpstreamValue->setText ( "---" );
+//        lblUpstreamUnit->setText ( "" );
```
ann0see, line 1263:
```suggestion
```

ann0see:
Will you totally remove the Measurements box in settings or do you think we should still show the LEDs there?

softins:
> Will you totally remove the Measurements box in settings or do you think we should still show the LEDs there?

I don't think it needs to show the LEDs, but it does need to show the data rate, as that is not shown anywhere else, and doesn't change except when the user changes something.

ann0see:
Yes. But I think we could still show the LEDs only to at least have some stats in the settings window.

softins:
> Yes. But I think we could still show the LEDs only to at least have some stats in the settings window.

But the LEDs still get updated when their status changes, and it's frequent updates in the settings window that potentially cause the issue (even though we've only seen it with `setText()` so far). I would be in favour of removing them, since the numbers they relate to are no longer there, and those LEDs are duplicating what is already shown in the main window.

ann0see:
Yes. But they don't seem to produce the bug. Let's wait what others think.

dcorson-ticino-com:
I just pushed an update with corrections and cleaning up.
Personally I think the leds belong with the values, client or settings, one or the other, not both.
The data rate is back in the settings window.

softins (commented):
This is a great improvement. But a few comments:
1. Need to make the value fields long enough to contain the largest value (e.g. 999) without varying in length. And make the fields of fixed length so they don't make the right-hand separator jump. I think this is what @ann0see mentioned.
2. There are some translation updates, e.g. "buffers" to "jitter buffers". We might want to revert to "buffers" until after the next release?
3. I wasn't able to make the "> 500ms" message appear. Will try again.

dcorson-ticino-com:
This will have an effect on the translations in any case.  The data stream was part of the same whatsthis text as the ping etc.
As they are no longer together I had to change them.

src/clientdlgbase.ui
```diff
@@ -582,7 +593,7 @@
   <tabstop>rbtReverbSelR</tabstop>
  </tabstops>
  <resources>
-  <include location="resources.qrc"/>
+  <include location="../../jamulus3_8_0rc1/src/resources.qrc"/>
```
hoffie, line 596: Same

dcorson-ticino-com: I have no explanation how that got in there.  I checking to correct.

dcorson-ticino-com: Those lines do not exist in my source.
I do not know what to do.
I have pushed again with small changes to the UI to assure the display of >500 etc.
I hope these errors are now gone.

hoffie: >  I hope these errors are now gone.
Looks good!

(Not marking this as resolved yet as it contains a hint for @softins regarding the 500ms question above)

src/clientdlg.cpp
```diff
@@ -179,7 +179,35 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
...
+                           "<br>" +
+                           tr ( "Overall Delay is calculated from the current Ping Time and the "
+                                "delay introduced by the current buffer settings." ) ;
+
+    lblPing->setWhatsThis ( strConnStats );
+    lblPingVal->setWhatsThis ( strConnStats );
+    lblDelay->setWhatsThis ( strConnStats );
+    lblDelayVal->setWhatsThis ( strConnStats );
+    ledDelay->setWhatsThis ( strConnStats );
+    ledDelay->setToolTip ( tr ( "If this LED indicator turns red, "
```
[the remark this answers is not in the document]
henkdegroot: @dcorson-ticino-com would it make sense to update the Fancy LEDs as part of this change? In total 6 lines of codes you need to update and add two images (which you can take from my PR #1752)?
As all other items in that PR have been covered now by this update.

henkdegroot:
@dcorson-ticino-com would it make sense to update the Fancy LEDs as part of this change? In total 6 lines of codes you need to update and add two images (which you can take from my PR #1752)?
As all other items in that PR have been covered now by this update.

henkdegroot:
> This will have an effect on the translations in any case.

Will we do a translation update for this? Just noticed a "What is This?" dialog which now contains a mixture of English (not translated) and Dutch. Not too disturbing though....but just checking.

src/clientdlg.cpp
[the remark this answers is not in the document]
hoffie: > Not part of this PR, but we should use QString ( … %1 … ) .arg ( APP_NAME ); here

Yes, I agree. Just to re-iterate, this not only does not have to be part of this PR, it *should* not be part of this PR to avoid introducing even more translation movements.

hoffie (commented):
Haven't looked at all details yet, just some thoughts:

- Despite the technical necessity to avoid the memleak, I personally like the change as it'll allow me to keep the settings window closed.
- I'm wondering if we should hide all the labels in disconnected state in order to avoid cluttering the UI for new users which haven't even connected (maybe in a follow up PR?).
- As this PR "only" shifts UI elements around, no new translations should be necessary, right? Are there any manual .ts edits necessary to guarantee that? (Also wondering where the strangeness that @henkdegroot described comes from)
- I'm a bit wary if this PR may just shift the problem from the settings window to the main window. @softins's tests seem to show that this is not the case. @pianojoe's initial bug report contained information that the main window may have a problem as well, so we should be extra sure that we don't make things worse by this change. Input from @pianojoe would be highly appreciated :)

softins:
I'll look more and respond sometime tomorrow. I'm rather Jamulused-out this evening after 4 days solid.

dcorson-ticino-com:
>     * As this PR "only" shifts UI elements around, no new translations should be necessary, right?

As I wrote above:
This will have an effect on the translations in any case. The data stream was part of the same whatsthis text as the ping etc.
As they are no longer together I had to change them.

henkdegroot:
> * (Also wondering where the strangeness that @henkdegroot described comes from)

Let me double check....thing it is caused by the fact that the qm files are not automatic updating in the deploy windows powershell...building now with updated qm to check.

pljones:
Could we have screenshots of all three skin old and new layouts attached, please.

(Any UI change PR should have these if there's no referenced issue that has them, IMO.)

I still think my alternative of putting the labels below the tab control but still on the settings window might have been worth working with...

dcorson-ticino-com:
This is how it looks on Win10
Client01
Client02
Client03
Client04

henkdegroot:
Screenshots using Windows 10 for LED updates and exceeding >500 ms:

exceed-500ms-ping-normal

exceed-500ms-ping-fancy

fancy-leds-with-accessibility-markers

pljones:
All I'd do differently is get rid of the group box around the Audio Stream Rate and make that a simple label.  Other than that, looks good!

pljones:
One other thing -- when I moved the Audio/Network settings dialog around, some of the language translations were quite long, making the window grow wider.  Here there's very little space, so that could be an issue.  It would be wise to see if the appropriate translations can be put together and then tested.

src/clientdlg.cpp
```diff
@@ -154,8 +154,8 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     ledDelay->setAccessibleName ( tr ( "Delay status LED indicator" ) );
 
     // buffers LED
-    QString strLEDBuffers = "<b>" + tr ( "Buffers Status LED" ) + ":</b> " +
-                            tr ( "The buffers status LED shows the current audio/streaming "
+    QString strLEDBuffers = "<b>" + tr ( "Jitter Buffers Status LED" ) + ":</b> " +
```
pljones, line 157: Question: this LED was labelled "Local Jitter Buffer" in Settings.  Either it applies only to the local buffer (i.e. a locally detected over/underrun) or it doesn't.  This wording (including the original), is vaguer - implying it could be either the server (which doesn't tell us it's had a problem as far as I know) or local buffer.

As it's being reworded, it might be wise to make it clear now.  (I notice the Settings page didn't actually _have_ WhatsThis text, just the label.)

ann0see:
ConnectButton

@dcorson-ticino-com could you please check if this glitch (Connect button too small) also appears in other languages? Probably you'll need to set a flexible size for this button.

pljones:
Just thought... The audio stream rate is a consequence of the channels and quality.  Could move back under them...

src/clientdlg.cpp
dcorson-ticino-com: I don't see anything in the protocol so it must be local only.  I will reword that.  
I was a little thrown by why the "Buffers" is plural so thought it must mean both.
Will reword.

dcorson-ticino-com:
If I remember correctly the connect button has a fixed size, I will look again and change that.
But that must have always been like that, I didn't change anything there.

dcorson-ticino-com:
> Just thought... The audio stream rate is a consequence of the channels and quality. Could move back under them...

I am leaving it where it is just to balance left and right a little

henkdegroot:
@dcorson-ticino-com I noticed I can now make the settings dialog very narrow. Not sure if that is seen as being an issue or not.

About the connect button, think because the left side now expands due to different lenght of text, it became visible that the button does not expand with it. Think making it the same width as the left section would be good.

dcorson-ticino-com:
> @dcorson-ticino-com I noticed I can now make the settings dialog very narrow. Not sure if that is seen as being an issue or not.
> About the connect button, think because the left side now expands due to different lenght of text, it became visible that the button does not expand with it. Think making it the same width as the left section would be good.

I just pushed a correction for the connect button.  Please try it out.
Making the settings very narrow was a request a couple of weeks ago.

ann0see:
Yes. Looks much better now! Thank you!

henkdegroot:
Button looks good now indeed....I mean it now allows extreem narrow.....it was not like this the previous rc1 builds.
narrow-settings

ann0see:
Yes. But I think that's intended

hoffie (commented):
So, after having a closer look:

- UI-wise, I'm happy
- Code-wise, I'm mostly happy -- see minor nit wrt function naming (plus I only skimmed the .ui files)
- This PR does more than the title says. It not only moves the measurements, it also changes other aspects (labels/texts, resize behavior [?]). In general, this would not be a problem and I consider all of these changes to be good. However, given that this is (hopefully) our rescue against memory leaks in RC1 phase, I'm keen on keeping such changes limited.
  - The text changes invalidate translations (@softins hinted at earlier that as well). This raises new questions:
    1. Should we run another translation cycle? This will further delay the release. I'd really not do that.
    2. Should we accept that parts of the UI will be untranslated? It looks like this will "only" affect signal words (`Ping`, `Jitter`) which are likely to be translated verbatim anyway + What's this texts. So it may be acceptable to go this path.
    3. Should we revert the unnecessary updates here (or in a follow-up, 3.8.0-only PR)? Not sure if it's worth the effort...
  - This affects 7 places (according to a `grep '"unfinished"><' src/res/translation/translation_de_DE.ts -c` comparison between master and this PR). 
  - In any case, I'd suggest not trying to stuff further improvements into this PR, even if they are welcome in general. :)
  - @dcorson-ticino-com
    > This will have an effect on the translations in any case. The data stream was part of the same whatsthis text as the ping etc. As they are no longer together I had to change them.

    I'm not sure about this part. While I think they had been part of the same concatenated string, they had been split into individual translations in the past as well. So it may be possible to retain the translation?

@jamulussoftware/maindevelopers What do you think?

In any case, @softins will probably have the final word here after another test run with the current PR state which has the data-rate label re-added.

Thanks for working on this @dcorson-ticino-com! :)

src/clientdlg.cpp
hoffie: > As it's being reworded, it might be wise to make it clear now. (I notice the Settings page didn't actually have WhatsThis text, just the label.)

In general, I agree, but I'm questioning the "As it's being reworded" for the RC phase.

pljones: > I'm questioning the "As it's being reworded" for the RC phase.

Unfortunate as it is, if it's needed to unblock the MacOS issue, then we're stuck with the change.

src/main.cpp
```diff
@@ -538,7 +538,7 @@ int main ( int argc, char** argv )
     QCoreApplication* pApp = new QCoreApplication ( argc, argv );
 #else
 #    if defined( Q_OS_IOS )
-    bUseGUI        = true;
+    bUseGUI = true;
```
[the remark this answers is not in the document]
pljones: We comply with the version autobuild uses.

src/resources.qrc
```diff
@@ -1,35 +1,16 @@
...
-    <qresource prefix="/translations">
         <file alias="translation_pl">res/translation/translation_pl_PL.qm</file>
-    </qresource>
-    <qresource prefix="/translations">
         <file alias="translation_sk">res/translation/translation_sk_SK.qm</file>
-    </qresource>
-    <qresource prefix="/translations">
         <file alias="translation_sv">res/translation/translation_sv_SE.qm</file>
     </qresource>
-
```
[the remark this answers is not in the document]
pljones: I'm wondering if some newer version of Qt Creator than the advised "newest" (5.12.2) did this.  If so, please revert.

pljones:
Option 1: Release 3.8.0 with the known error and working translations, to be followed shortly by 3.8.1 with the fix (and potentially broken translations).

Option 2: Release 3.8.0 with the fix and broken translations.

Option 3: Delay 3.8.0, include the fix and update the translations.

Personally, I'd go with 1 or 2.  It depends on how the number of MacOS users compares with the number of translation users.

ann0see:
I wouldn’t go with 1. This bug is already fixed and we shouldn’t ship broken software. Not having a translation isn’t that bad having a bug is far worse.

pljones:
Shipping with known, documented defects is perfectly acceptable -- if the impact is small.  Like I say, if the number of people who'd be affected severely adversely by the broken translations was large (which I doubt, to be honest - the area affected is small) and the number of people affected by the bug was small, I'd go with the bug. Conversely, and I tend to believe more likely, if the number affected by the bug is greater, then fixing becomes the priority.  But if the number affected by the translations _is_ large, delay becomes an option.

hoffie:
> Option 1: Release 3.8.0 with the known error and working translations, to be followed shortly by 3.8.1 with the fix (and potentially broken translations).
> 
> Option 2: Release 3.8.0 with the fix and broken translations.
> 
> Option 3: Delay 3.8.0, include the fix and update the translations.

I'd cleraly prefer option 2.

My point was just that we could possibly limit the effect of broken translations even without doing another round of translations and delay. @softins did a very good job doing so in other places, but I suspect Tony's time budget is a bit taxed for this release... :)
Still, this is being extra-picky as the affected translations seem to only consist of signal words and What's this texts.

softins:
I have some time to devote to this tomorrow (Sat), and possibly a bit later this evening too. I would prefer option 2 or if we can manage it, 2.5 :)

src/resources.qrc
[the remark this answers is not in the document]
dcorson-ticino-com: Hmm ????
I have never touched that file and am using the same version of Qt since the end of last year.  Why it should suddenly decide to change something ????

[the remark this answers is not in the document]
henkdegroot: > I have never touched that file

You did:) .... when you added the two image files for the Fancy status LEDs with accessibility markers. Not sure if you did that manually or in Qt of course...

src/clientdlg.cpp
```diff
@@ -1258,6 +1285,14 @@ void CClientDlg::UpdateDisplay()
         chbChat->setChecked ( true );
         chbChat->blockSignals ( false );
     }
+    if ( !pClient->IsRunning() )
```
pljones, line 1288: Should be in `CClientDlg::Disconnect()`?

src/resources.qrc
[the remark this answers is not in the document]
pljones: You can also fix it like this:
```
git reset HEAD~
git add (all the other files)
git add -p src/resources.qrc
(n to everything except the two new png files)
git commit -m 'your commit message'
git checkout -- src/resources.qrc
```
(then, of course, test a local build before `git push --force`)

src/server.h
```diff
@@ -103,7 +103,7 @@ class CHighPrecisionTimer : public QThread
 protected:
     virtual void run();
 
-    bool     bRun;
```
pljones, line 106: Please revert this change.

src/main.cpp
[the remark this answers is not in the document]
pljones: But, as it appears to be the only change to main.cpp - please revert.

dcorson-ticino-com:
There is a problem with the Clang-format compliance.
I ran the directory through clang-format before committing, the autobuild style check says it is not OK, however.

src/clientdlg.cpp
dcorson-ticino-com: Good catch!
Copy paste is good, turning brain on is better :)

src/server.h
dcorson-ticino-com: I did not change this.
If anything changed it, it was clang-format.

softins:
@dcorson-ticino-com I see that this PR was raised from your `master` branch. You should never do that, for several reasons:

1. It will cause you later confusion in your master branch after the change has been merged, especially if it gets squashed, or modified by committers.
2. It prevents you from working on more than one feature at a time in a clean way.
3. It makes it more difficult for project maintainers to push a change to your PR if required.

When starting to work on a new feature or a bug fix, you should _always_:
- Get your master up to date with the upstream master:
  - `git checkout master`
  - `git fetch upstream`
  - `git rebase upstream/master`
  - `git push` (now your local and origin masters match the upstream master)
- Create a new branch for your changes:
  - `git checkout -b my-new-branch-name`
- Make all your changes on this branch. If you are also working on other unrelated changes, they can be in a different branch using the same technique, and you can switch between them easily.
- Commit your changes to the branch:
  - `git status` (make sure you are still on the expected branch)
  - `git add <files>...` (or instead just do `git commit -a` on the next line)
  - `git commit` (add the commit message)
- If this branch has never been pushed before, do `git push` and it will tell you the full command to use, which you can copy and paste. Or just type it in:
  - `git push --set-upstream origin my-new-branch-name`
- If you are ready to raise a PR from it, you can do that at Github. If you have more to do, just make your new changes, commit as above, and do a simple `git push` to update Github.
- If the PR has not yet been merged, you can push further commits to it as above. If it has been merged, any further changes should be in a new branch and a separate PR.
- If anyone else pushes a change to your branch (e.g. a project maintainer), you need to fetch those changes:
  - `git checkout my-new-branch-name`
  - `git pull`
- If, due to the passage of time, the upstream master gets significantly updated and you want to get your work-in-progress up to date, you need to rebase and force-push:
  - `git checkout master`
  - `git fetch upstream`
  - `git rebase upstream/master` (if you keep your own `master` clean of changes, this will just fast-forward)
  - `git push` (updates your origin master)
  - `git checkout my-new-branch-name` (the one you are working on)
  - `git rebase master` (get the latest upstream changes into your work-in-progress - you may need to handle conflicts)
  - Once all conflicts are resolved and the result compiles and tests ok:
  - `git push --force`

pljones:
`git rebase master` -- following two weeks of misery at work, I now _always_ recommend `rebase -i` over plain `rebase`, so you have some idea of what commits are going to be applied.  If they are _not_ only the ones from your branch, you're in for pain.

src/resources.qrc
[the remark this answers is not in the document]
softins: I don't think there is a problem with this change. It just built happily on my Pi, and also on the Github CI. It is also more consistent with the sections for other resources.

softins (approved):
I think this is good to merge now. I will check the built Mac artifact when it becomes available, and will merge then, once it has a second approval.

hoffie (approved):
Thanks for the updates, @dcorson-ticino-com and @softins!
Approving now, but maybe @pljones can also indicate if the PR is acceptable for him now?

pljones:
Is there another issue raised to cover reinstating the changed reverted?  I'd like that done before things are lost...

softins:
> Is there another issue raised to cover reinstating the changed reverted? I'd like that done before things are lost...

It's still in the commit history. I don't propose to squash merge. But I've raised an issue for 3.8.1 to keep it visible #1775 

> I'd like the unnecessary part of the change to `src/resources.qrc` reverted - i.e. only the two new lines added.

I don't think that's necessary, but in the interests of harmony will do it - although it will be 6 new lines if the format is kept consistent.

Issue #1738, opened 2021-05-22
emoji don't show up correctly


RobyDati:
**Describe the bug**
emoji don't show up correctly in Jamulus release 3.8.0rc1 with Mac Os Big Sur 11.3.1

**To Reproduce**
check emoji used in MyProfile name of Jamulus clients, or in the Server Name

**Expected behavior**
emoji show up correctly in previous Jamulus releases up to 3.8.0beta2 also with Mac Os Big Sur 11.3.1... Maybe it's compatibility with qt5?  Furthermore I tested Jamulus release 3.8.0rc1 with Mac Os Mojave and with Linux Ubuntu and the problem is not present there.

**Screenshots**
I don't have at this time: there are a box instead of any expected emoji whenever it is used in user MyProfile or in Server Name in the Connection setup or in the icons in the Mixer.

**Operating system**
Mac Os Big Sur 11.3.1   (it does not occur in my other Mac  Mojave 10.14.6 machine and on Linux Ubuntu)

**Version of Jamulus**
Jamulus release 3.8.0rc1   (it worked fine on previous versions, it looks related to qt5)

**Additional context**
I am not sure if this may help: https://bugreports.qt.io/browse/QTBUG-87689  https://github.com/rstudio/rstudio/issues/7897

hoffie:
Thanks for the report!
Could you please try 3.8.0beta1 as well? This would help pin-pointing the source of the issue as this was the last pre-release with the older Mac/Qt build environment (#1655).

cc @softins

hoffie:
I'm tagging this for 3.8.0 now so it doesn't get lost. We still have to investigate whether we can do something about it at all and whether it is important enough to delay 3.8.0 final for that.

RobyDati:
I tried again 3.8.0beta1 and emoji show up correctly with Mac OS Big Sur, as it also works OK 3.8.0beta2 with Big Sur. I experience the problem starting from Jamulus release 3.8.0rc1 with Mac Os Big Sur 11.3.1 (and also on Big Sur 11.2.3).

softins:
That will be when we switched Qt from 5.9 to 5.12 to address a different issue on Big Sur. :(

softins:
@RobyDati please could you try the Mac builds at https://github.com/softins/jamulus/actions/runs/880290268 and let us know which ones display correctly?

RobyDati:
[...]

PR #1768, opened 2021-05-27
Mac current and legacy

softins:
This PR updates the autobuild of artifacts to create two builds for Mac, using Qt 5.15.2 for the main Mac build and Qt 5.9.9 for a legacy Mac build.

Users of macOS 10.13 (High Sierra) onwards should use the main Mac build, and users of macOS 10.10 (Yosemite), 10.11 (El Capitan) and 10.12 (Sierra) should use the legacy Mac build.

This fixes both #1628 and #1738


hoffie:
Thanks! :) Tagging for 3.8.0 as it avoids a regression (#1738).

ann0see:
The build seems to be broken on macOS (new)

softins:
Hmm, ok. Must have overlooked something when omitting a build siffux. Will fix soon.

hoffie (approved):
Looks good to me. :)

The only thing I'm wondering is if and how we could or should make the difference clearer (thinking of the Github release entry). Would `legacy` be clear enough`? Adding all targeted, legacy code names to the file name would probably make it too long as well. So it might be best to keep `legacy` as this PR does.

softins:
> The only thing I'm wondering is if and how we could or should make the difference clearer (thinking of the Github release entry). Would `legacy` be clear enough? Adding all targeted, legacy code names to the file name would probably make it too long as well. So it might be best to keep `legacy` as this PR does.

I went through the same thought process, and came to the same conclusion.

I think we just need to make it clear in the release notes, based on my first comment above.

autobuild/mac/artifacts/autobuild_mac_3_copy_files.sh
```diff
@@ -23,7 +27,7 @@ echo ""
 
 echo ""
 echo ""
-artifact_deploy_filename=jamulus_${jamulus_buildversionstring}_mac.dmg
+artifact_deploy_filename=jamulus_${jamulus_buildversionstring}_mac${BUILD_SUFFIX}.dmg
```
ann0see, line 30: This also needs to be specified on the website (new variable and button on the macOS page).

ann0see (approved):

ann0see:
@JohannesBrx see this PR.

Issue #1737, opened 2021-05-22
Review Github Actions for Security


hoffie:
We are using Github Actions in several places:
- jamulus: Autobuild (including Releases and CodeQL)
- jamuluswebsite:
  - Jekyll
  - Merge between branches

We are not only using official Github-provided Actions there, but also multiple third-party actions (see below).
I am not seeing any use of the `permission:` keyword there, implying that they run with default permissions. This means that those actions have access to a GITHUB_TOKEN with read and write permission to the relevant repo, as far as I understand.

I have reviewed the following docs and articles:
https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
https://francoisbest.com/posts/2020/the-security-of-github-actions

My conclusion is that we should do the following:
- [ ] Review all referenced actions for correctness and safety and reference them by sha1 hash instead of branch or tag.
  - TODO: Do it for all actions or just third-party stuff? Doing only the latter might be a good compromise and might safe work.
  - Pinning to exact commits can be done by a single person. Review of the actual actions should probably split and assigned to thoese developers who introduced the action?
- [ ] Ensure that [Dependabot](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot) is active as it supports alerting us if we miss updates to those pinned actions.
- [ ] Set Github default action permissions to be restrictive (read-only) and update those workflows which need it to have write permission as well.
- [ ] Document the necessity to keep all of this in mind when reviewing PRs which touch these workflow files. I'm planning to include this in the Admin wiki page.

Note: I'm little worried about Github-official actions such as `actions/` or `github/` (we are trusting Github anyway!) or actions for other large open source projects with high reputation (`ruby/`), but I do worry about actions by third-party persons or orgs which we (or at least I?) don't know.

```
$ grep uses: jamulus*/.github/workflows/*.yml
jamulus/.github/workflows/autobuild.yml:           uses:                    actions/checkout@v2
jamulus/.github/workflows/autobuild.yml:           uses:                    dev-drprasad/delete-tag-and-release@v0.1.2
jamulus/.github/workflows/autobuild.yml:           uses:                    actions/create-release@v1
jamulus/.github/workflows/autobuild.yml:        uses:                       maxim-lobanov/setup-xcode@v1
[...]

PR #1778, opened 2021-05-29
Restrict permissions to Autobuild

ann0see:
Related to https://github.com/jamulussoftware/jamulus/issues/1737

Although not yet tested, I think we only need content permissions (which is quite a lot unfortunately) for the autobuild CI.

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -34,6 +34,8 @@ on:
 name:                               Auto-Build
 jobs:
   create_release:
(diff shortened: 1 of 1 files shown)


pljones:
Are you working directly on jamulussoftware/jamulus rather than your own repo?

If you were on your own repo, you might be able to test (i.e. by going through a release cycle on a branch - just changing the Jamulus.pro version should trigger it, right?).

ann0see:
Yes. I shouldn’t do that.

PR #1790, opened 2021-05-30
Update FR app translation for 3.8.0

jujudusud:
Adding more translation update in RC2 stage.

### DIFF
--- a/src/res/translation/translation_fr_FR.ts
+++ b/src/res/translation/translation_fr_FR.ts
@@ -790,27 +790,27 @@
     <message>
         <location filename="../../clientdlg.cpp" line="186"/>
         <source>Current Connection Status Parameter</source>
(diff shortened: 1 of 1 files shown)


hoffie:
Thanks. From a quick look at the diff, this "just" cleans up the .ts file (which is good!), but does not change any visible translations, right?
If this is the case, I suggest merging it post-3.8.0 to avoid all unnecessary changes?

hoffie (approved):
Just saw that one word changes: "Jitter". Let's wait what the others say wrt. right time to merge.

jujudusud:
It is the latest translations needed by the new arrangement of the main window (jitter added and removed from parameters).
If the modifications of the jitter is merged, then the translation have also to be merged. If not, the translation will not be complete.

hoffie:
Changed milestone to 3.8.0 as this PR avoids a visible missing translation.

ann0see (approved):
Merging this since I think it’s small

Issue #1791, opened 2021-05-31
Jamulus 3.7 hangs -eventually- on Mac Big Sur


RonCohen1:
**Describe the bug**
 Jamulus hangs -- pretty much every time -- after a while -- can be 5 minutes, can be an hour, on my Macbook Pro (2020, 2 GHz Intel quad-core I5) running Big Sur 11.2.3.  Not a problem with same Jamulus version on an older Mac running an older OS.

**To Reproduce**
Join a Jamulus session, then don't touch anything.  Eventually I will get a spinning ball and can't activate any Jamulus controls.   Sometimes Jamulus continues to function; at other times, the connection drops.    (does not happen with current Jamulus on an old Mac running High Sierra.) 

**Expected behavior**
 Expected behavior is a normal Jamulus session.

**Screenshots**

**Operating system**
 MacOS 11.2.3

**Version of Jamulus**
 Jamulus on this same machine seems to have buffer underruns/overflows -- sound gets glitchy: I can temporarily cure it by re-setting the buffer delay.  It seems to be when I don't do reset the buffer delay that the hang occurs.

hoffie:
Hi!
Thanks for the report. This sounds very similar to #1643, but I'd like to keep it open for now as the other issue mainly targeted a regression introduced in 3.8.0 (despite the initial report).

We suspect a general problem in the Qt event handling on Mac with recent Qt versions and yet have to find a way to work around it.

It would help if you could try the following:
- [3.8.0rc2 (regular Mac build)](https://github.com/jamulussoftware/jamulus/releases/download/r3_8_0rc2/jamulus_3.8.0rc2_mac.dmg) - It woud be best if this version did not show any issues for you as it is planned to become the 3.8.0 final release. I suspect that you *will* encounter the issue with this version as well, but this information would help a lot.
- [3.8.0rc2 (legacy Mac build)](https://github.com/jamulussoftware/jamulus/releases/download/r3_8_0rc2/jamulus_3.8.0rc2_mac_legacy.dmg) - I suspect that it will not show the issue, but may have design glitches in the settings window tabs (please confirm).
- [3.6.2](https://sourceforge.net/projects/llcon/files/Jamulus/3.6.2/Jamulus-3.6.2-installer-mac.dmg/download) - Unless you had already been running 3.6.2 or earlier without issues, it would be interesting to know if this version still works reliably.

gilgongo:
[...]

Issue #1643, opened 2021-05-06
CPU usage increases over time until the UI becomes unresponsive. [mac]


pianojoe:
On Mac, after two hours of rehearsal with ~40 participants, the UI becomes unresponsive. A click may take 20s to execute. `top` shows a processor usage of 103%. (As compared to 35% one hour earlier, with the same number of participants.)
The Mac mouse cursor changes into the "spinning beachball of death", as it is commonly reffered to.

CPU usage shouldn’t grow over time. 

I’m on macos 11.3.1 on a 2.9 GHz 6-Core Intel Core i9 MacBookPro (definitely not underpowered). I’m using a USB audio interface Focusrite Scarlett 4i4.

The behaviour was observed on Jamulus 3.7.0 (obtained as a .dmg from jamulus.io) AND on Jamulus 3.8.0beta1 (which I compiled myself).

hoffie:
Thanks for the report.

1. Just to confirm, you are talking about the Jamulus client, right?

2. Just a blind guess because it seems to be the reason for another recent Mac-related issue: Which version of Qt did you use on your local build? The official `.dmg` is built against rather old Qt (5.9) which seems to cause problems (UI glitches) on recent Mac versions.
Could you try this build and report back if this improves things for you?
https://github.com/jamulussoftware/jamulus/issues/1628#issuecomment-832859835

3. Does memory usage also increase?

pianojoe:
hoffie, 

1. Yes, I’m talking about the Jamulus Client.
2. Qt 5.15.2. This is the very same Qt version as your build.
3. Need to check. I’ll get back to you on this.

hoffie:
Ok. My only idea for getting close to the cause of the issue would be to try different working/non-working versions. Does 3.6.2 exhibit the same problem? If not, that would be something to start from. Are you familiar with git bisect? That way you could generate different builds and check if they are broken or not. This will likely require some iterations of building / verifying / building again, etc., but will lead to a single commit which is the cause for the change in behavior.

This requires effort, but I don't know where else to start. :(

softins:
Also, when it gets into this state, does clicking Disconnect and reconnecting give any improvement, or is it necessary to exit and restart the program?

MattMerz-Violin:
I have the same problem. Jamulus 3.7 becomes unresponsive after playing for some time and not touching any UI item. It then takes up to 20 seconds until the mouse click is recognized and the UI is reacting again. MacOS 11.3.1.

[...]

PR #1794, opened 2021-05-31
Mac autobuild: Update to Xcode 12

hoffie:
Let the autobuild default to latest xcode (= 12.x) again, but still ensure that we limit SDK support to 10.15 due to lack of Qt-wise support for newer versions.

This is a different approach to the already merged #1655. This change has a (low) chance of fixing recent Mac issues on Big Sur 
(#1643, #1791). At least, it makes us use the latest officially recommended setup.

This only touches the regular Mac build, not the legacy build.

Support matrix:
https://doc.qt.io/qt-5/macos.html

Variables:
https://doc.qt.io/qt-5/qmake-variable-reference.html#qmake-mac-sdk

Explanations:
https://doc.qt.io/qt-5/macos-deployment.html

This PR is just a blind test if it works like that at all (cannot try locally). If it does, it would be interesting to have affected users test this version.

cc @softins


hoffie:
Autobuild failed, `QMAKE_MAC_SDK=macosx10.15` doesn't seem to be respected while the other option (`QMAKE_MACOSX_DEPLOYMENT_TARGET=10.13`) is. Investigating.

softins:
I think what you have done is not enough. Xcode 12 only includes the 11.1 SDK by default. See my test run from 25 days ago at https://github.com/softins/jamulus/runs/2522272216 where I had added a line `xcodebuild -sdk -version` to the prepare scripts to see which SDKs were present. There is only the latest one SDK for each platform.

You would need to determine what commands are needed to load an additional older SDK before you can select it for building with. I started to look at that, but then side-stepped it by switching to Xcode 11.

hoffie:
> I think what you have done is not enough. Xcode 12 only includes the 11.1 SDK by default. See my test run from 25 days ago at https://github.com/softins/jamulus/runs/2522272216 where I had added a line `xcodebuild -sdk -version` to the prepare scripts to see which SDKs were present. There is only the latest one SDK for each platform.

Yep, thanks, I noticed that some minutes ago and tried Xcode 12 which still ships 10.15. Currently fails in python scripts, looking into this now.

> You would need to determine what commands are needed to load an additional older SDK before you can select it for building with. I started to look at that, but then side-stepped it by switching to Xcode 11.

Yeah, I think this was and is the simplest solution. It's just that I'm hoping that code generated by newer Xcode might magically make our leak bugs go away...

softins:
If you are pinning the Xcode version to 12.0 (I think 12.1 would work too), in order to get the 10.15 SDK as default, I suspect you don't need the `EXTRA_QMAKE_ARGS` any more?

softins:
https://en.wikipedia.org/wiki/Xcode#Xcode_11.x_-_12.x_(since_SwiftUI_framework)

hoffie:
> If you are pinning the Xcode version to 12.0 (I think 12.1 would work too)

Ah, right. Trying with 12.1.1 now.

> in order to get the 10.15 SDK as default, I suspect you don't need the `EXTRA_QMAKE_ARGS` any more?

Hrm, yep, can probably drop it. Will leave it in now.

Do you know a way to use older SDKs with latest Xcode? All I've found was inofficial binary dumps, which I'd rather not use due to legal and security questions...?

ann0see:
@emlynmac since you‘re a macOS specialist, could you have a look at this please?

emlynmac:
Is there a reason to use 12.1.1 instead of the current 12.4 Xcode?

ann0see:
I doubt there’s a reason. Probably 12.1.1 was the latest version as this PR was opened?

emlynmac:
Maybe let's get as far up the chain as we can then?

ann0see:
Yes, I think @dcorson-ticino-com worked on Qt 6 support, but I'm not sure how far he got.
So we should move this one up to the latest Xcode version.

.github/workflows/autobuild.yml
```diff
@@ -143,13 +145,13 @@ jobs:
...
-      # For Qt5 on Mac, we need to ensure SDK 10.15 is used, and not SDK 11.x
-      # This is done by selecting Xcode 11.7 instead of the latest default of 12.x
+      # For Qt5 on Mac, we need to ensure SDK 10.15 is used.
+      # Xcode 12.0 is the last version which ships SDK 10.15, therefore pin it:
       - name:                       Select Xcode version for Mac
         if:                         ${{ matrix.config.target_os == 'macos' }}
         uses:                       maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: '11.7'
+          xcode-version: '12.1.1'
```
ann0see, line 154:
```suggestion
          xcode-version: '12.4.0'
```

ann0see (changes requested):

ann0see:
Let's have a look at the CI

dcorson-ticino-com:
> 
> 
> Yes, I think @dcorson-ticino-com worked on Qt 6 support, but I'm not sure how far he got.
> So we should move this one up to the latest Xcode version.

Yes, I have a Qt6 branch that compiles and works fine as far as I can see.
As I don't know what Xcode is I don't think I changed it.

softins:
> Is there a reason to use 12.1.1 instead of the current 12.4 Xcode?

@emlynmac  Yes, see the comments further up (I've not been active over the weekend, so just picking up this conversation). Qt5 is only compatible with SDKs up to 10.15. If you use SDK 11, you get lots of warning messages from Qt about compatibility.

Xcode 12.1.1 is apparently the last one that comes with SDK 10.15, which is why we pinned that one. We should continue to do so, unless someone can come up with a way to use SDK 10.15 with later versions of Xcode.

ann0see:
@dcorson-ticino-com could you please open a PR on your work? Probably all the macOS issues are somehow related and Qt6 might solve them.

@emlynmac @softins is it worth to move this forward for the next release or not?

softins:
> @dcorson-ticino-com could you please open a PR on your work? Probably all the macOS issues are somehow related and Qt6 might solve them.

Certainly worth trying, but I'm not optimistic that Qt6 would be any better than Qt5 on Mac.

> @emlynmac @softins is it worth to move this forward for the next release or not?

Maybe for 3.9.0, but certainly not for 3.8.1.

dcorson-ticino-com:
> @dcorson-ticino-com could you please open a PR on your work? Probably all the macOS issues are somehow related and Qt6 might solve them.

Give me a couple of days, I'm no longer sure in what kind of state it is in.

ann0see:
Ok. De-tagged.

ann0see:
> Give me a couple of days, I'm no longer sure in what kind of state it is in.

@emlynmac @dcorson-ticino-com @pljones not quite sure what to do with this. I'd rather open an issue to discuss the macOS problems...

dcorson-ticino-com:
> not quite sure what to do with this. I'd rather open an issue to discuss the macOS problems...

Tony has gone forward with this, but having compilation problems on git.  I had it working OK under both QT5 and QT6 compiled locally.
As to the macOS problems I think it will be worth while to see how QT6 helps.

ann0see:
Ok. If we do remove the license note, we could get it to work.

ann0see:
https://github.com/ann0see/jamulus/runs/4685522957?check_suite_focus=true

hoffie:
@softins @ann0see @emlynmac What's the current state of this PR?
I think the original proposal (update to 12.1.1) still makes sense.

The initial PR did this, but it has been modified to update to 12.4, which is sub-optimal, I think:

This official page confirms that 12.1.x is (and probably will stay?) the last version to support SDK 10.15, which we need for Qt5 compatibility: https://developer.apple.com/support/xcode/
This inofficial page lists 12.1.1 as the most-recent 12.1.x release: https://xcodereleases.com/

As all the newly introduced logic wrt `EXTRA_QMAKE_ARGS` as its no longer needed, I'm closing this PR and will open a new one for the 12.1.1 update.

PR #1796, opened 2021-05-31
Mac build: Pin dmgbuild to 1.4.2 as 1.5.0 is broken

hoffie:
We are using `dmgbuild` in the Mac build process. Version 1.5.0 was released [some hours ago](https://pypi.org/project/dmgbuild/1.5.0/#history) and it breaks our build process as seen here:

https://github.com/hoffie/jamulus/runs/2709826916?check_suite_focus=true#step:6:1329 (that branch is a copy of `master`)

This PR pins the version to the version we've been using previously.
Unless `dmgbuild` is updated again on PyPI, this PR will have to be merged for the final Jamulus release as autobuild will fail otherwise. At the same time, it should be save to merge because it restores the state to what was used for rc2.

I have raised an upstream issue as well: https://github.com/al45tair/dmgbuild/issues/40

### DIFF
--- a/mac/deploy_mac.sh
+++ b/mac/deploy_mac.sh
@@ -44,7 +44,7 @@ build_installer_image()
 {
     # Install dmgbuild (for the current user), this is required to build the installer image
     python -m ensurepip --user --default-pip
-    python -m pip install --user dmgbuild
+    python -m pip install --user dmgbuild==1.4.2
     local dmgbuild_bin="$(python -c 'import site; print(site.USER_BASE)')/bin/dmgbuild"
 
     # Get Jamulus version


hoffie:
Heh, the upstream maintainer was really quick... 1.5.1 was released with a fix. I'm still wondering whether we should do the pinning for two reasons:
- The build process is now different between rc2 and the final (due another dmgbuild version being used).
- For the security reasons outlined in #1797

Therefore, I'm keeping the PR open for now.

hoffie:
> 1.5.1 was released with a fix

... and @softins noticed that the new version is also broken (https://github.com/al45tair/dmgbuild/issues/41).

I'd suggest merging this PR for now.

softins (approved):

pljones (approved):

PR #1798, opened 2021-05-31
honour JACK_DEFAULT_SERVER environment variable

jp8:
commit 2a798c626883a2651f56b47da87d9c54e03c228e checks the environment variable but then does nothing with it.

This pull request passes the server name through to the jack client library so it can look for the named jackd instance.

### DIFF
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -29,9 +29,9 @@
 #ifdef WITH_JACK
 void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientName )
 {
-    jack_status_t JackStatus;
+    jack_status_t  JackStatus;
     jack_options_t options = JackServerName;
-    const char*   serverName;
+    const char*    serverName;
 
     if ( ( serverName = getenv ( "JACK_DEFAULT_SERVER" ) ) == NULL )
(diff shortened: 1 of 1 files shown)


hoffie:
Hrm, have you seen the previous discussion around this?
https://github.com/jamulussoftware/jamulus/discussions/1397
https://github.com/jamulussoftware/jamulus/pull/1406

Basically, the assumption is that the enviuronment variable is interpreted by libjack without having to change anything within Jamulus. I just tested this again and it seems to work as expected.

So I'm not sure why we need this PR?

jp8:
No I didn't see the previous discussions.  When it didn't work (in my environment) I checked the code, saw the serverName was going nowhere, and added a line to pass it through.
Perhaps my initial test was flawed.

hoffie:
Ok. If you cannot get it to work without this PR, we should have a deeper look why and may need to merge this PR.
If it works without this PR when re-testing, I suggest closing it.

henkdegroot:
Just want to add, I have tested my "windows jack build" and this also works fine when setting the JACK_DEFAULT_SERVER value.

jp8:
I tested again, and it works without this PR.

PR #1799, opened 2021-05-31
Server list persistence for directory servers

pljones:
Currently, when restarting a directory server, it clears its list of registered servers.  This means clients consulting the directory get an empty server list returned (well, containing the directory server itself only).

In order to make maintenance of directory servers less annoying for client users, this feature adds persistence of the list of registered servers across restarts.

Persistence is optional.  It's enabled by specifying an _existing, readable_ filename as an additional entry in the server info (i.e. `name;city;country;persistencefile`).  Providing an empty file the first time is mandatory.

I have chosen not to add to the `--help` display, as this is only for directory server lists.  I needs adding to the appropriate Wiki page, however.

When specified, the file is read on start up and written on shutdown (if writeable).  Each line written to the file contains an entry from the server list (excluding the directory server itself).  These are read on start up, prepopulating the list.

Servers must still re-register regularly to maintain their entry in the list.

**NOTE:** a7bb996259ab8baa8af81a5cf02e326eba38ff97 is #1803, on which this change depends.

### DIFF
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -25,6 +25,72 @@
 #include "serverlist.h"
 
 /* Implementation *************************************************************/
+
+// --- CServerListEntry ---
+CServerListEntry CServerListEntry::parse ( QString strHAddr,
+                                           QString strLHAddr,
+                                           QString sName,
+                                           QString sCity,
+                                           QString strCountry,
+                                           QString strNumClients,
+                                           bool    isPermanent )
+{
+    CHostAddress haServerHostAddr;
+    NetworkUtil::ParseNetworkAddress ( strHAddr, haServerHostAddr );
+    if ( CHostAddress() == haServerHostAddr )
+    {
+        // do not proceed without server host address!
+        return CServerListEntry();
+    }
+
+    CHostAddress haServerLocalAddr;
+    NetworkUtil::ParseNetworkAddress ( strLHAddr, haServerLocalAddr );
+    if ( haServerLocalAddr.iPort == 0 )
+    {
+        haServerLocalAddr.iPort = haServerHostAddr.iPort;
+    }
+
+    // Capture parsing success of integers
+    bool ok;
+
+    QLocale::Country lcCountry = QLocale::AnyCountry;
+    int              iCountry  = strCountry.trimmed().toInt ( &ok );
+    if ( ok && iCountry >= 0 && iCountry <= QLocale::LastCountry )
+    {
+        lcCountry = static_cast<QLocale::Country> ( iCountry );
+    }
+
+    int iNumClients = strNumClients.trimmed().toInt ( &ok );
+    if ( !ok )
+    {
+        iNumClients = 10;
+    }
+
+    return CServerListEntry ( haServerHostAddr,
+                              haServerLocalAddr,
+                              CServerCoreInfo ( FromBase64ToString ( sName.trimmed().left ( MAX_LEN_SERVER_NAME ) ),
+                                                lcCountry,
+                                                FromBase64ToString ( sCity.trimmed().left ( MAX_LEN_SERVER_CITY ) ),
+                                                iNumClients,
+                                                isPermanent ) );
+}
+
+QString CServerListEntry::toCSV()
+{
+    QStringList sl;
+
+    sl.append ( this->HostAddr.toString() );
+    sl.append ( this->LHostAddr.toString() );
+    sl.append ( ToBase64 ( this->strName ) );
+    sl.append ( ToBase64 ( this->strCity ) );
+    sl.append ( QString::number ( this->eCountry ) );
+    sl.append ( QString::number ( this->iMaxNumClients ) );
+    sl.append ( QString::number ( this->bPermanentOnline ) );
+
+    return sl.join ( ";" );
+}
+
+// --- CServerListManager ---
 CServerListManager::CServerListManager ( const quint16  iNPortNum,
                                          const QString& sNCentServAddr,
                                          const QString& strServerInfo,
@@ -38,7 +104,7 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
     eSvrRegStatus ( SRS_UNREGISTERED ),
     iSvrRegRetries ( 0 )
 {
-    // set the directory server address
+    // set the directory server address (also bIsCentralServer)
     SetCentralServerAddress ( sNCentServAddr );
 
     // set the server internal address, including internal port number
@@ -69,20 +135,15 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
         iServInfoNumSplitItems = slServInfoSeparateParams.count();
     }
 
-    // per definition, the very first entry is this server and this entry will
-    // never be deleted
-    ServerList.clear();
-
     // Init server list entry (server info for this server) with defaults. Per
     // definition the client substitutes the IP address of the directory server
-    // itself for his server list. If we are the directory server, we assume that
-    // we have a permanent server.
+    // itself for his server list. If we are a directory server, we assume that
+    // we are a permanent server.
     CServerListEntry
-        ThisServerListEntry ( CHostAddress(), SlaveCurLocalHostAddress, "", QLocale::system().country(), "", iNumChannels, GetIsCentralServer() );
+        ThisServerListEntry ( CHostAddress(), SlaveCurLocalHostAddress, "", QLocale::system().country(), "", iNumChannels, bIsCentralServer );
 
     // parse the server info string according to definition:
-    // [this server name];[this server city]; ...
-    //    [this server country as QLocale ID]; ...
+    // [this server name];[this server city];[this server country as QLocale ID] (; optional permanent servers ... below)
     // per definition, we expect at least three parameters
     if ( iServInfoNumSplitItems >= 3 )
     {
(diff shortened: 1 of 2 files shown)


softins:
> Persistence is optional. It's enabled by specifying an _existing, readable_ filename as an additional entry in the server info (i.e. `name;city;country;persistencefile`). Providing an empty file the first time is mandatory.

This seems a slightly arcane way of specifying the persistence file. Why not just add a command-line option such as `--persistfile` to specify the filename?

softins:
And for that matter, what is wrong with having the server create an empty file if it is specified but does not already exist?

hoffie:
> > Persistence is optional. It's enabled by specifying an _existing, readable_ filename as an additional entry in the server info (i.e. `name;city;country;persistencefile`). Providing an empty file the first time is mandatory.
> 
> This seems a slightly arcane way of specifying the persistence file. Why not just add a command-line option such as `--persistfile` to specify the filename?

I agree, but if it is limited to directory servers, we should probably have that context in the flag name. `--directoryserver-persistfile` (not sure if we have any flags with dashes in the middle yet though, but it's getting hard to read otherwise).

src/serverlist.cpp
```diff
@@ -25,6 +25,71 @@
...
+
+    sl.append ( this->HostAddr.toString() );
+    sl.append ( this->LHostAddr.toString() );
+    sl.append ( this->strName );
+    sl.append ( this->strCity );
+    sl.append ( QString::number ( this->eCountry ) );
+    sl.append ( QString::number ( this->iMaxNumClients ) );
+    sl.append ( QString::number ( this->bPermanentOnline ) );
+
+    return sl.join ( ";" );
```
passing, line 89: a server can have a semicolon in its name or city (when it is started from the GUI) and would get sorted out when loading the file

pljones: Good points.  We should make `;` illegal in the GUI, in that case (and I can strip here), as it's illegal in `--serverinfo`.  That was overlooked when adding the GUI.  Same for newlines (which should never be allowed).

pljones:
> 
> 
> > Persistence is optional. It's enabled by specifying an _existing, readable_ filename as an additional entry in the server info (i.e. `name;city;country;persistencefile`). Providing an empty file the first time is mandatory.
> 
> This seems a slightly arcane way of specifying the persistence file. Why not just add a command-line option such as `--persistfile` to specify the filename?

Because it's a file containing server info.  It also avoids adding unnecessary code bloat.

pljones:
> And for that matter, what is wrong with having the server create an empty file if it is specified but does not already exist?

Because I expect anyone using the option to know what they're doing.  This way it can be validated.

passing:
I see a few disadvantages in extending `--serverinfo` to `name;city;country;persistencefile`

- In case some other `serverinfo` field that a server can provide should be added later (maybe its Genre, or other metadata), you will have to add that to the end again, but then the 4th parameter needs to be left empty as it is only relevant for the special case of running an own directory server.
- Another use case could be to run Jamulus as a Directory Server only. In that case, the sub-parameters 1-3 would need to be left empty.

I'd even rather suggest to split `--serverinfo` into 3 independent parameters - keeping the original for backwards compatibility - to not have to provide 3 values in the specific format & order `name;city;country`

pljones:
> In case some other serverinfo field that a server can provide should be added later ...

Then that case can be dealt with if it ever arises.

> Another use case could be to run Jamulus as a Directory Server only. In that case, the sub-parameters 1-3 would need to be left empty.

No, that's not true.  The directory server needs its own server info values as it's always in its own list.

----

> IMO it makes sense have separate Directory server only options.

Not part of this change.

passing:
> > Another use case could be to run Jamulus as a Directory Server only. In that case, the sub-parameters 1-3 would need to be left empty.
> 
> No, that's not true. The directory server needs its own server info values as it's always in its own list.

@pljones, that's how it is right now. But logically it doesn't make sense that a directory server needs to be a server it its own list.

My main idea is to not make the way to a more modular architecture harder.
I am just trying to help here with suggesting alternatives / trying to discuss them. But I don't have the impression that this is wanted.

pljones:
> @pljones, that's how it is right now. But logically it doesn't make sense that a directory server needs to be a server it its own list.

Yes, it does.  The server _has to_ register with itself.  As it can also be a jam server, it _should_ supply details.

pljones:
This is now running on Rock, Jazz, Classical/Folk and Choral/Barbershop genre servers.

I have to admit, having messed up configuration the first time I set it up, having a separate command line option might make it easier.

```
Server only:
...
  -e, --directoryserver address of the directory server with which to register
                        (or 'localhost' to host a server list on this server)
      --directoryfile   filename in which to store directory server list
                        across restarts
...
```

I'm also thinking splitting the Directory server only options out might be good, too - but not in this change.  KISS (hence #1803 and #1804).

src/serverlist.cpp
softins: A thought, after reviewing #1803 and #1804: those PRs only prevent the server itself creating names containing `\n` and `;`; they don't do anything to prevent older registering servers from sending them to the directory. The code below strips out `\n` but not `;`.

Also, I think there is some code duplicated between those PRs and this one? (e.g. `reIllegal`)

Volker avoided the problem of strange characters in names stored in the ini file by base64 encoding the names and writing in XML format. Maybe there is some code there that could be used here, instead of having to disallow characters?

ann0see:
> Yes, it does. The server has to register with itself ("localhost" and default port), otherwise it's not a directory server. 

Could you please clarify why this is needed? Yes, currently a directory server is a normal Jamulus server which also allows "slave servers" to register with it. However, the Jam server functionality is not needed. From my tests with @passing's python library, It is perfectly ok and possible to have a directory which is not shown on the list. This could also work with your directories which are only limited to one user (which is honestly just an implementation workaround).

softins:
I thought a while ago that it would be useful if a directory server could accept `-u 0` to prevent it adding itself to the list of registered servers, and from accepting any audio connections. I just never got around to implementing it.

hoffie:
> I thought a while ago that it would be useful if a directory server could accept `-u 0` to prevent it adding itself to the list of registered servers, and from accepting any audio connections. I just never got around to implementing it.

I agree and this was requested by users as well. We would still have to be careful because the first Directory server reply in the server list is handled specially in the client. @passing found [a way around that](https://github.com/jamulussoftware/jamulus/discussions/1525#discussioncomment-683645) though. Should probably create a new Issue for that topic if it is going to be implemented.

pljones:
> Should probably create a new Issue for that topic if it is going to be implemented.

It's not related to this work, so yes.

src/serverlist.cpp
```diff
@@ -25,6 +25,71 @@
...
+// --- CServerListEntry ---
+CServerListEntry CServerListEntry::parse ( QString strHAddr,
+                                           QString strLHAddr,
+                                           QString sName,
+                                           QString sCity,
+                                           QString strCountry,
+                                           QString strNumClients,
+                                           bool    isPermanent )
+{
+    bool ok;
```
ann0see, line 38: Could you please describe in a comment what this bool does?

src/serverlist.cpp
```diff
@@ -469,6 +519,7 @@ void CServerListManager::CentralServerQueryServerList ( const CHostAddress& Inet
                     // servers behind a NAT and dealing with external, public
                     // clients.
                     vecServerInfo[iIdx].HostAddr = ServerList[iIdx].LHostAddr;
+                    // ?? Shouldn't this send the ping, as below ??
```
ann0see, line 522: @hoffie ?

src/serverlist.cpp
```diff
@@ -490,6 +541,105 @@ void CServerListManager::CentralServerQueryServerList ( const CHostAddress& Inet
...
+    return INVALID_INDEX;
+}
+
+void CServerListManager::CentralServerLoadServerList ( const QString strServerList )
+{
+    QFile file ( strServerList );
+
+    if ( !file.open ( QIODevice::ReadWrite | QIODevice::Text ) )
+    {
+        qWarning() << qUtf8Printable ( QString ( "Could not open '%1' for read/write" ).arg ( ServerListFileName ) );
```
ann0see, line 565:
```suggestion
        qWarning() << qUtf8Printable ( QString ( "Could not open '%1' for read/write. Please check if Jamulus has read and write permissions and if your disk is full." ).arg ( ServerListFileName ) );
```

Goes with https://github.com/jamulussoftware/jamulus/discussions/1568

src/serverlist.cpp
```diff
@@ -490,6 +541,105 @@ void CServerListManager::CentralServerQueryServerList ( const CHostAddress& Inet
...
+    CHostAddress haServerHostAddr;
+
+    QTextStream in ( &file );
+    while ( !in.atEnd() )
+    {
+        QString     line   = in.readLine();
+        QStringList slLine = line.split ( ";" );
+        if ( slLine.count() != 7 )
+        {
+            qWarning() << qUtf8Printable ( QString ( "Could not parse '%1' successfully - bad line" ).arg ( line ) );
```
ann0see, line 582: As above. Can the user do anything? Fix the file, look for syntax errors,...?

src/serverlist.cpp
```diff
@@ -490,6 +541,105 @@ void CServerListManager::CentralServerQueryServerList ( const CHostAddress& Inet
...
+            continue;
+        }
+
+        CServerListEntry serverListEntry =
+            CServerListEntry::parse ( slLine[0], slLine[1], slLine[2], slLine[3], slLine[4], slLine[5], slLine[6].toInt() != 0 );
+
+        // We expect servers to have addresses...
+        if ( ( CHostAddress() == serverListEntry.HostAddr ) )
+        {
+            qWarning() << qUtf8Printable ( QString ( "Could not parse '%1' successfully - invalid host" ).arg ( line ) );
```
ann0see, line 600: As above. Probably not that critical...

src/serverlist.cpp
pljones: Undo whatever manual edit they did to break it ;)

src/serverlist.cpp
pljones: It's used by `String.toInt(&ok)` - checking for zero isn't good enough validation in some places.

src/serverlist.cpp
pljones: Not a bad idea.  Convert the values to base64, then join with `;`.  It would save having to have the other two PRs, too.

pljones: Just to note -- the INI file does _not_ use base64 encoding for server name and city:
```
826     // name
827     PutIniSetting ( IniXMLDocument, "server", "name", pServer->GetServerName() );
828
829     // city
830     PutIniSetting ( IniXMLDocument, "server", "city", pServer->GetServerCity() );
```
Again, that's probably an oversight.

pljones:
My directory servers now running this latest version (briefly having taken each down, reformatted the persistence file, then started up with the new version).

src/serverlist.cpp
ann0see: So it should say that.

pljones: Well, that's guessing and potentially wrong information.  I really, really don't like messages that try to guess what's wrong and suggest potentially incorrect things to do on the basis of too little information.  Either the user knows what they've done - and therefore knows without being told - or they don't and they can't do anything about it.  Putting annoying guidance in the message is simply that - annoying.

So no.

pljones:
> I have to admit, having messed up configuration the first time I set it up, having a separate command line option might make it easier.
> ...
> In fact, I'll leave it as is for now - I'll add the new command line option separately and re-organise the list at that time.

See https://github.com/pljones/jamulus/pull/9

I've left the server list options in with the other server options for now:
* `--directoryserver` applies to both, so it doesn't have an obvious home
* there's only `--directorylist` and `--listfilter` other than that
* `--serverpublicip` is supplied to a server that's registering, but so is `--serverinfo` - would they go under server or directory server?

pljones:
https://github.com/pljones/jamulus/pull/9 now running all my directory servers.

passing:
> I've left the server list options in with the other server options for now:
> 
>     * `--directoryserver` applies to both, so it doesn't have an obvious home
>     * there's only `--directorylist` and `--listfilter` other than that
>     * `--serverpublicip` is supplied to a server that's registering, but so is `--serverinfo` - would they go under server or directory server?

It would make sense to me to keep `--directoryserver`, `--serverpublicip` and `--serverinfo` under "Server only". So it would be just 2 "Directory Server only" options at the moment, which would make the server options list a bit shorter at least. But I also don't see a problem in keeping it as it is.

src/serverlist.cpp
hoffie: Hrm, so this is about making the `--serverpublicip` Directory-side logic more consistent with the regular code path regarding NAT traversal, right?

I agree that a ping should be send to be as close to the normal path.

On the other hand, this feature is only used by people who already have the capability to set up proper port forwardings (otherwise they couldn't run a public directory server behind NAT). Therefore, I'd argue that it might not be necessary. As this hasn't been reported as a bug or missing feature either, I'd rather not do something right now. If anyone disagrees, please say so and I can open a PR.

PR #1803, opened 2021-06-01
New lines are illegal in --serverinfo values

pljones:
Required for #1799.

### DIFF
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -36,7 +36,8 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
     strMinServerVersion ( "" ),              // disable version check with empty version
     pConnLessProtocol ( pNConLProt ),
     eSvrRegStatus ( SRS_UNREGISTERED ),
-    iSvrRegRetries ( 0 )
+    iSvrRegRetries ( 0 ),
+    reIllegal ( "\n" )
 {
     // set the directory server address
     SetCentralServerAddress ( sNCentServAddr );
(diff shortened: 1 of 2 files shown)


src/serverlist.cpp
```diff
@@ -87,10 +88,10 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
     if ( iServInfoNumSplitItems >= 3 )
     {
         // [this server name]
-        ThisServerListEntry.strName = slServInfoSeparateParams[0].left ( MAX_LEN_SERVER_NAME );
+        ThisServerListEntry.strName = slServInfoSeparateParams[0].remove ( reIllegal ).left ( MAX_LEN_SERVER_NAME );
```
ann0see, line 91: Just removing \n automatically without warning the user might cause confusion. I'd give a warning on stderr and strip the new line afterwards.

pljones: It can go on the Wiki.  It's quite difficult to get a new line into the string in the first place - I doubt people are going to use it.

ann0see: Hmm yes that’s probably more common. Can you open an issue on the website repo to document this change?

pljones: Done: https://github.com/jamulussoftware/jamuluswebsite/issues/514

ann0see: Thanks!

gilgongo:
BTW I've added this to the new Server Admin Manual. PR-ing shortly.

pljones:
Going with a different approach here, so this is no longer needed.

PR #1805, opened 2021-06-01
COMPILING.md - android instructions

ngocdh:


ann0see (approved):
Thank you!

COMPILING.md
```diff
@@ -145,3 +145,15 @@ Will build the file and make it available in `./Release/Jamulus.app`
 * Run `/path/to/qt/5.15.2/ios/bin/qmake -spec macx-xcode Jamulus.pro`
 * Open the generated .xcodeproject in XCode, check the Signing & Capabilities and configure a team
 * To run on a iOS device, build and run on the device, you'll have first trust your developer profile in device's Settings.
+
+## Android
```
gilgongo, line 149: I don't know anything about Android development, but to refer explicitly to UI elements "the Projects tab" and menus, as well as `git clone` and `cd jamulus` seems a bit unnecessary/risky to me. Can we not bring it into line with the brevity of the other platforms?

When I originally did this doc, I removed a bunch of stuff I assumed experienced engineers/power users would see as noise, and we would be lumbered with checking in future...

COMPILING.md
```diff
@@ -145,3 +145,15 @@ Will build the file and make it available in `./Release/Jamulus.app`
...
+* Install Qt, including the Android support from the Qt installer.
+* Configure the Android settings in the Qt Creator (Tools -> Options -> Devices -> Android). Solve all warnings/errors in that dialog (set up paths to SDKs and NDKs and select the target for android, etc.). Tick on the "Automatically create kits for Android Toolk chains").
+* Make sure the Android Kit is detected on Tools -> Options -> Kits -> Kits (example: Android Qt 5.15.2 Clang Multi-Abi). It should also show in the Qt Versions tab (example: Qt 5.15.2 for Android).
+* Clone Jamulus and its submodules:
+`git clone https://github.com/jamulussoftware/jamulus.git`
+`cd jamulus`
+`git submodule update --init`
+* Open Jamulus.pro in Qt Creator
+* In the Projects tab -> Build & Run It should appear the Android Qt, might be disabled. Enabled it if it is not enabled.
+* Then you can connect your device and build or run the project.
```
gilgongo, line 159: Given the preceding detailed instructions on how to set up QT, is how to "connect your device and build or run the project" obvious? But see my earlier general comment.

ngocdh: Thanks @gilgongo for the feedback. Updated accordingly (and some bug fixes).

gilgongo: I'm worried that these instructions are going to mean we'll need to maintain them too much, and they seem to be (sort of) aimed at somebody who's never used QT before. This isn't the document for that. 

Instead, is there anything here an experienced person who has compiled and run an Android QT app would not know? `git submodule update --init` after cloning the repo perhaps? Also, would we not need to list the necessary SDKs and NDKs?

Again, see the style of the instructions for other platforms in this document for the sort of thing I mean. The guide for iOS does not attempt to teach people how to use Qt.

ngocdh: It's really based on my personal experience with both, iOS and Android. I didn't have experience with either Android or iOS.  
For iOS, it was kind of easy, I didn't have to deal with Qt Creator, just needed the Qt binaries to generate XCode project, and then I was good to go.
For Android though, Qt Creator and all the configurations seemed a lot more complicated.

Of course for an experienced person, I suppose they wouldn't need all these detailed instructions. I'll remove some more now.

gilgongo: I don't know anything about Android really, but in general for this doc (and for various reasons), we need to just have the minimum we think is needed by somebody who has done this kind of thing before (as we do for the other platforms). We don't yet support Android, so I hope anyone expecting to use it after compiling will understand that too.

ngocdh: @gilgongo should be good now :)

gilgongo: Excellent - thanks!

gilgongo (approved):

Issue #148, opened 2020-04-26
Honour own fader and Mute button in “Mute Myself”


softins:
“Mute Myself” is a great feature, but when selected, I hear myself at full volume. My fader doesn’t control the volume, and the Mute button on my channel is also ignored.

It would be great if the local mix of my channel in that mode could take account of my own fader and Mute button.

elliotclee:
What I found was that when I connected to a server, I could hear myself even with "Mute Myself" on, "Mute" on my channel, and the slider all the way down.

When I disconnect from a server I didn't hear myself at all, so it is something to do with Jamulus or some weird feature of my sound setup.

This is on Windows with ASIO4ALL. I don't know if there is something about my setup that is doing software monitoring.

While doing this, I also discovered that the "Mute" and "Solo" buttons can be set at the same time. Should be mutually exclusive...

pljones:
Audio only gets to Jamulus audio outs when it's receiving audio from a server.

"Mute myself" means "Bypass the server".  It stops anyone else on the server hearing you, muting you.

The input fader then lets you adjust your level.

Ideally you should never need to use the mixer fader for your own level, only the input fader.

softins:
What input fader? Jamulus doesn’t have one, does it?

The point is, if for reasons of balance I have my own fader at a lower level than 100%, then if I click on Mute Myself, what I hear of myself suddenly changes to full volume. It should stay the same. My signal should be multiplied by the fader position on my channel when mixing directly to my output. And while doing that it would be sensible and complete to obey the Mute button on my channel too.

Once more familiar with the code, I could submit a PR for it, but I’m sure Volker would know immediately how to do it.

corrados:
It is not trivial. The client does not know which fader is his. You could compare the fader name with yours but if another has given his fader the same name, you have ambiguous results.

Why do you not have your fader on maximum? I am right now working on improving the sound card audio input stream selection. In the future you will be able to change the input gain of each input stream of your sound card. Then you can do the attenuation there instead of the server mixer.

softins:
> It is not trivial. The client does not know which fader is his. You could compare the fader name with yours but if another has given his fader the same name, you have ambiguous results.

OK, I understand.

> Why do you not have your fader on maximum?

Often I do, but if in a group where some of the participants have a lower signal level, I might set myself lower so I can hear them better. After all, the faders must be there for a purpose, not just to leave on maximum!

pljones:
[...]

PR #1809, opened 2021-06-02
Own fader to first position

ngocdh:
This pull request set own fader always to 1st position, regardless of the sorting setting.
See discussion #1646

In case we want to have the choice to set or unset the option, then UI changes must be involved.

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1070,9 +1070,15 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
     // create a pair list of lower strings and fader ID for each channel
     QList<QPair<QString, int>> PairList;
     int                        iNumVisibleFaders = 0;
+    int                        iMyFader          = -1;
 
     for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
     {
+        if ( vecpChanFader[i]->GetIsMyOwnFader() )
+        {
+            iMyFader = i;
+        }
+
         if ( eChSortType == ST_BY_NAME )
         {
             PairList << QPair<QString, int> ( vecpChanFader[i]->GetReceivedName().toLower(), i );
@@ -1116,6 +1122,19 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
 
     // sort the channels according to the first of the pair
     std::stable_sort ( PairList.begin(), PairList.end() );
+
+    // move my fader to first position
+    if ( pSettings->bOwnFaderFirst )
+    {
+        for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
+        {
+            if ( iMyFader == PairList[i].second )
+            {
+                PairList.move ( i, 0 );
+                break;
+            }
+        }
+    }
 
     // we want to distribute iNumVisibleFaders across the first row, then the next, etc
     // up to iNumMixerPanelRows.  So row wants to start at 0 until we get to some number,
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -308,6 +308,12 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // View menu  --------------------------------------------------------------
     QMenu* pViewMenu = new QMenu ( tr ( "&View" ), this );
 
+    // own fader first option: works from server version 3.5.5 which supports sending client ID back to client
+    QAction* OwnFaderFirstAction =
+        pViewMenu->addAction ( tr ( "O&wn Fader First" ), this, SLOT ( OnOwnFaderFirst() ), QKeySequence ( Qt::CTRL + Qt::Key_W ) );
+
+    pViewMenu->addSeparator();
+
     QAction* NoSortAction =
         pViewMenu->addAction ( tr ( "N&o User Sorting" ), this, SLOT ( OnNoSortChannels() ), QKeySequence ( Qt::CTRL + Qt::Key_O ) );
 
@@ -324,6 +330,9 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     QAction* ByCityAction =
         pViewMenu->addAction ( tr ( "Sort Users by &City" ), this, SLOT ( OnSortChannelsByCity() ), QKeySequence ( Qt::CTRL + Qt::Key_T ) );
+
+    OwnFaderFirstAction->setCheckable ( true );
+    OwnFaderFirstAction->setChecked ( pSettings->bOwnFaderFirst );
 
     // the sorting menu entries shall be checkable and exclusive
     QActionGroup* SortActionGroup = new QActionGroup ( this );
--- a/src/settings.h
+++ b/src/settings.h
@@ -133,6 +133,7 @@ public:
         bWindowWasShownSettings ( false ),
         bWindowWasShownChat ( false ),
         bWindowWasShownConnect ( false ),
+        bOwnFaderFirst ( false ),
         pClient ( pNCliP )
     {
         SetFileName ( sNFiName, DEFAULT_INI_FILE_NAME );
@@ -167,6 +168,7 @@ public:
     bool       bWindowWasShownSettings;
     bool       bWindowWasShownChat;
     bool       bWindowWasShownConnect;
+    bool       bOwnFaderFirst;
 
 protected:
     // No CommandLineOptions used when reading Client inifile
--- a/src/audiomixerboard.h
+++ b/src/audiomixerboard.h
@@ -84,6 +84,7 @@ public:
     int    GetRunningNewClientCnt() { return iRunningNewClientCnt; }
     void   SetChannelLevel ( const uint16_t iLevel );
     void   SetIsMyOwnFader() { bIsMyOwnFader = true; }
+    bool   GetIsMyOwnFader() { return bIsMyOwnFader; }
     void   UpdateSoloState ( const bool bNewOtherSoloState );
 
 protected:
--- a/src/clientdlg.h
+++ b/src/clientdlg.h
@@ -159,6 +159,11 @@ public slots:
     void OnOpenAdvancedSettings();
     void OnOpenChatDialog() { ShowChatWindow(); }
     void OnOpenAnalyzerConsole() { ShowAnalyzerConsole(); }
+    void OnOwnFaderFirst()
+    {
+        pSettings->bOwnFaderFirst = !pSettings->bOwnFaderFirst;
+        MainMixerBoard->SetFaderSorting ( pSettings->eChannelSortType );
+    }
     void OnNoSortChannels() { MainMixerBoard->SetFaderSorting ( ST_NO_SORT ); }
     void OnSortChannelsByName() { MainMixerBoard->SetFaderSorting ( ST_BY_NAME ); }
     void OnSortChannelsByInstrument() { MainMixerBoard->SetFaderSorting ( ST_BY_INSTRUMENT ); }
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -270,6 +270,12 @@ void CClientSettings::ReadSettingsFromXML ( const QDomDocument& IniXMLDocument, 
         eChannelSortType = static_cast<EChSortType> ( iValue );
     }
 
+    // own fader first sorting
+    if ( GetFlagIniSet ( IniXMLDocument, "client", "ownfaderfirst", bValue ) )
+    {
+        bOwnFaderFirst = bValue;
+    }
+
     // number of mixer panel rows
     if ( GetNumericIniSet ( IniXMLDocument, "client", "numrowsmixpan", 1, 8, iValue ) )
     {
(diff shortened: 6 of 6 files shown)


ann0see:
Thanks for this. I think it should be possible to enable this in the UI to not confuse existing users. Tagging @gilgongo and @mulyaj here to voice their opinion on where to put it. I‘m still unsure where to put it since I fear that the new tabbed settings UI opens up cluttering the settings dialog with too many features.

ngocdh:
> Thanks for this. I think it should be possible to enable this in the UI to not confuse existing users. Tagging @gilgongo and @mulyaj here to voice their opinion on where to put it. I‘m still unsure where to put it since I fear that the new tabbed settings UI opens up cluttering the settings dialog with too many features.

A check box in the Edit menu, close to Sorting options

dcorson-ticino-com:
Thanks @ngocdh, this was on my "to be done soon" list.
I agree that enabling/disabling is most sensibly in the edit menu with the other sorting choices.

ngocdh:
> Thanks @ngocdh, this was on my "to be done soon" list.

Cool thanks @dcorson-ticino-com. When you get time, could you implement the menu? It seems CClientSettings (to store/load the option) and CClientDlg (the Menu item and action) are involved.

ngocdh:
So I got around to implement that  Menu action :)

Note though: the feature will not work if connected to old servers, just like any other feature relying on `OnClientIDReceived`, because it seems old servers don't send client id back to each client?

softins:
> Note though: the feature will not work if connected to old servers, just like any other feature relying on `OnClientIDReceived`, because it seems old servers don't send client id back to each client?

If that's the case, could you determine which version server is the first to support that, and note it in a code comment and commit message? And mention it here too. Thanks!

ngocdh:
Found: 3.5.5. 
Ref: https://github.com/jamulussoftware/jamulus/issues/148#issuecomment-634208150

src/clientdlg.cpp
```diff
@@ -304,6 +304,12 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // Edit menu  --------------------------------------------------------------
     QMenu* pEditMenu = new QMenu ( tr ( "&Edit" ), this );
 
+    // own fader first option: works from server version 3.5.5 which supports sending client ID back to client
```
ngocdh, line 307: Feature doesn't work if connected to server prior to version 3.5.5.

pljones:
Just want a status update on this.  Is anything outstanding?  (I don't appear to have complained about anything, so I guess I better read the code :) ).

pljones:
This will need some refactoring follow @jp8's menu changes for 3.8.1.

jp8:
Easy enough to move this setting to the View menu instead of the Edit menu.

But there's still a lot to think about, especially around the multi-row, sort-by-group views.

pljones:
> But there's still a lot to think about, especially around the multi-row, sort-by-group views.

For an initial implementation, a "first means first" by whatever sort order is in use approach should be taken.  Keep it simple is part of the Jamulus creed.

pljones:
@ngocdh are you still able to pick this up?

The client creates the channel for itself as the very first channel.  It is _always_ channel zero.  So finding it is "easy".  Then you just apply the sort to the later channels in the list.

I had a thought on the GUI.

Following the menu reorganisation, I'd really like to get the check boxes for "Settings" and "Chat" removed, so that they're controlled through the menus (with shortcuts available).

That would give space for a check box for "Keep me first in the mixer" or "Mixer 1st" as it needs to be very short.
image

gilgongo:
Is "Keep me first in the mixer" that important to be given such prominence? Seem a little odd given other settings we have.

jp8:
My view was this should be the default when"sort by arrival time" is selected.  And "sort by arrival time" still needs a better name.

But not when "sort by instrument" or "sort by name", etc. are selected.  That would be allowing the users to select two conflicting settings.

And no need for an individual checkbox for it.

pljones:
> My view was this should be the default when"sort by arrival time" is selected. And "sort by arrival time" still needs a better name.
> 
> But not when "sort by instrument" or "sort by name", etc. are selected. That would be allowing the users to select two conflicting settings.
> 
> And no need for an individual checkbox for it.

First, I'm never a fan of changing existing behaviour -- people have set expectations and shouldn't actively have to adjust a setting to restore to that behaviour.

I can't see why it shouldn't be a choice in all sort modes.  It's fairly clear: put me first, always, then sort everyone else per whatever the sort rule is.  So it's not exactly conflicting.

And there would need to be a way for the user to choose to have the behaviour or not.  Maybe it should be a checkable option on the view menu.  (Maybe I'd prefer that and clear even more check boxes off the mixer...)

gilgongo:
@jp8 Yes, that seems sensible. 

As to the wording, I would think that most people would expect the default to be by arrival/joining time left to right, so we could just call it "default" and explain in the docs for the curious?

gilgongo:
> put me first, always, then sort everyone else per whatever the sort rule is. So it's not exactly conflicting.

One for the settings panel I think, not the main screen though.

jp8:
> First, I'm never a fan of changing existing behaviour -- people have set expectations and shouldn't actively have to adjust a setting to restore to that behaviour.

You are certainly right.  But just to be clear, I wasn't suggesting they would be able to adjust a setting to restore the old way.  My suggestion would be to modify the 'Default" sort order so that the user's own channel is at the left (with no setting anywhere for the user to turn that off), and leave all the other sort orders unchanged.

My reasoning is that people who use the "Default" sort order just want to have the faders remain in the same place, and not jump around when musicians join.  Putting own fader at the left doesn't (according to my theory) inconvenience them.

jp8:
> > put me first, always, then sort everyone else per whatever the sort rule is. So it's not exactly conflicting.
> 
> One for the settings panel I think, not the main screen though.

The reason i disagree with this is because if the view menu says "sort by name", and the user selects that, but the channels are not sorted by name....  it's a non-intuitive user interface.

pljones:
> 
> 
> > > put me first, always, then sort everyone else per whatever the sort rule is. So it's not exactly conflicting.
> > 
> > 
> > One for the settings panel I think, not the main screen though.
> 
> The reason i disagree with this is because if the view menu says "sort by name", and the user selects that, but the channels are not sorted by name.... it's a non-intuitive user interface.

Not if there's a check item on the same menu saying "Own channel first", surely?

jp8:
> Not if there's a check item on the same menu saying "Own channel first", surely?

Right - if the setting exists I think it needs to be a checkbox on the View menu, right below the selectable views (like you first suggested).

But I am not convinced that users of the 'Default' sort order need an option to have their own fader somewhere other than the left position.

And I am not convinced that users who have sorted by name, instrument, etc, need an option to make an exception for their own fader.  In my view, it should *always* go to the correct position for the chosen sort order.

jp8:
> so you're essentially saying "I don't want this feature"?

No I didn't mean to come across as not wanting the feature.  But I would like the Jamulus interface to be clear and consistent for the users, so I am trying to think carefully about it.

pljones:
I just don't like arbitary decisions made on behalf of the user.

jp8:
Yeah I just want to think carefully, get a sensible default, and only add choices in the UI if they are really necessary.

The original requirement was "own fader should be easy to find".... maybe we should have discussed highlighting it visually but leaving it in position.

DavidSavinkoff:
If there are several sort criteria, they should be prioritised by the user
in the user settings.
eg. sort A third, sort B first, sort C none, sort D Nth sort E none

No sort check box should be on the main menu.
eg. the DIR command in MSDOS can be configured perfectly in
autoexec.bat or config.sys to perfect the directory listing.
(it appears modern operating systems have lost this ability)

ngocdh:
Team, this feature is actually simple: if user wants their fader first on the mixer, then check the option, regardless of other sorting options.
I don't think there should be a checkbox on the mixer for this though.

softins:
> Following the menu reorganisation, I'd really like to get the check boxes for "Settings" and "Chat" removed, so that they're controlled through the menus (with shortcuts available).

Not sure I like this idea. I don't think I have ever opened the Settings and Chat windows via the menu bar, always with the buttons in the mixer.

pljones:
> 
> 
> > Following the menu reorganisation, I'd really like to get the check boxes for "Settings" and "Chat" removed, so that they're controlled through the menus (with shortcuts available).
> 
> Not sure I like this idea. I don't think I have ever opened the Settings and Chat windows via the menu bar, always with the buttons in the mixer.

That's because before the menu reorganisation, finding the menu items was hard and now you're used to it the way it was easiest.

softins:
I like this option the way it is, and have just been testing it. The PR was branched from 3.8.0rc2, and I have just squashed and rebased it to current master. @ngocdh is it ok if I push the rebased version to your branch?

ngocdh:
Sure Tony

softins (approved):
Rebased on master and tested.

pljones (approved):
Looks okay.

src/clientdlg.cpp
```diff
@@ -308,6 +308,12 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // View menu  --------------------------------------------------------------
     QMenu* pViewMenu = new QMenu ( tr ( "&View" ), this );
 
+    // own fader first option: works from server version 3.5.5 which supports sending client ID back to client
+    QAction* OwnFaderFirstAction =
+        pViewMenu->addAction ( tr ( "O&wn Fader First" ), this, SLOT ( OnOwnFaderFirst() ), QKeySequence ( Qt::CTRL + Qt::Key_W ) );
```
ann0see, line 313: @gilgongo Not sure if we could improve the wording here?
Own fader first is not 100% percise in my opinion.
Own fader at first position?

src/settings.h
```diff
@@ -133,6 +133,7 @@ class CClientSettings : public CSettings
         bWindowWasShownSettings ( false ),
         bWindowWasShownChat ( false ),
         bWindowWasShownConnect ( false ),
+        bOwnFaderFirst ( false ),
```
ann0see, line 136: Why not enable it by default?

src/audiomixerboard.cpp
```diff
@@ -1117,6 +1123,19 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
...
+    if ( pSettings->bOwnFaderFirst )
+    {
+        for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
+        {
+            if ( iMyFader == PairList[i].second )
+            {
+                PairList.move ( i, 0 );
+                break;
+            }
+        }
```
ann0see, line 1136: Not quite happy with the code here bcs. of the break; but it's valid...

src/settings.h
ann0see: ok. got it...

src/audiomixerboard.cpp
softins: Why not happy? It seems a right use of `break;` to me: once the correct entry has been found, there is no reason to continue with the loop. I feel the correct use of `break`, `continue` and early `return` contribute greatly to the readability of code and a reduction in unnecessary nesting and conditions.

ann0see: Yeah, probably it’s preference 😀

src/clientdlg.cpp
gilgongo: Not sure - "Show own fader first" perhaps (I don't think the initial capitals are in style BTW)? At least, I don't think "first" is ambiguous so that it needs "position" as well:  "They came first in the race" vs "They came in first position in the race". "To put your own interests first" vs "To put your own interests in first position". "Position" feels redundant.

jp8: Talking of being 100% precise...

I am still of the opinion that having the user's own fader at the left is good in the default (unsorted) sort, but I don't see any reason for the user to have the option to turn it on and off.

If I play violin, and I sort by instrument, I would expect my fader to be right there with the other violins.

If I am in London and I sort by city, I would expect my fader to be right there with the other London people.

I also haven't seen any convincing argument of why a user would need to both:
- sort by instrument, and also
- have their own fader appear at the left instead of with the others of the same instrument

Users don't have that option now.  But is there somebody that really wants to do exactly that ?  And if so, is that need strong enough to justify an extra setting ?  Especially one that (in my view, at least) conflicts with the wording of the existing sort options ...?

I personally would really struggle to come up with 100% precise accurate wording for all the sort options that would no longer be strictly correct.

ngocdh: I agree with every point you made here.

src/audiomixerboard.cpp
```diff
@@ -1070,9 +1070,15 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
     // create a pair list of lower strings and fader ID for each channel
     QList<QPair<QString, int>> PairList;
     int                        iNumVisibleFaders = 0;
+    int                        iMyFader          = -1;
 
     for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
```
ann0see, line 1075: @ngocdh 
There might be issues with the code here (see #2738) can you please look at this again?

pljones: (And note, again, I'd prefer if all this "state" information was held by the client - the UI should be responsible for no more than displaying that information as held.)

PR #1810, opened 2021-06-02
Add two missing translations for DE

softins:


hoffie (approved):

src/res/translation/translation_de_DE.ts
```diff
@@ -1396,12 +1396,12 @@ Wir haben Deinen Kanal stummgeschaltet und die Funktion &apos;Stummschalten&apos
...
         <location filename="../../clientsettingsdlg.cpp" line="248"/>
         <source>Select the language to be used for the user interface.</source>
-        <translation type="unfinished"></translation>
+        <translation type="unfinished">Wähle die Sprache für die Benutzeroberfläche aus.</translation>
     </message>
     <message>
         <location filename="../../clientsettingsdlg.cpp" line="250"/>
         <source>Language combo box</source>
-        <translation type="unfinished"></translation>
+        <translation type="unfinished">Sprache Combo-Box</translation>
```
ann0see, line 1404:
```suggestion
        <translation type="unfinished">Sprachen Combo-Box</translation>
```

softins: Sure. Is that for plural or for a different case? (The English was Language, not Languages)

ann0see: Yes. That's plural. After thinking a bit more, I think this would be better:

```suggestion
        <translation type="unfinished">Sprach Combo-Box</translation>
```

softins: The only similar constructions I could find were:
- Eingangs-Verstärkungs Combo box
- Verzeichnisserveradresse Combo-Box (which I changed from Zentralserveradresse)
- Oberfläche Combo-Box

ann0see: I'd leave it like that for the moment although I'm not really happy with it. @rolamos @ewarning should look at it and decide on what sounds more natural.

hoffie: > All three don't sound natural, maybe @hoffie can clarify what is better here.

It's a can of worms which should not be opened as part of this PR, IMO. :)

@softins What you listed can all be interpreted to be correct in some way, but it's not consistent. :)

- One could argue in favor just copying the original label "Sprache" and adding "Combo Box"
- One could argue in favor of proper translation "Sprach-Combo-Box" or "Combo-Box für die Sprachauswahl"
- There are probably more options... :)

ann0see: Ok. So we agree to merge this PR and let the two translators decide later

ann0see (approved):

rolamos:
Do you still need anything from me here ?

softins:
> Do you still need anything from me here ?

We’re good for the release, thanks, but would be interested in your opinion about the best translation of “Language combo box”, as discussed above.

PR #1829, opened 2021-06-04
Add autobuild for Windows with JACK

henkdegroot:
As requested in #1750 

Autobuild scripts for windows now have an BuildOption variable to allow the creation of the "jackonwindows" build.

Edit by @ann0see 

# Current status:

Almost finished; waiting on second review of maindev


ann0see (changes requested):
I think we should try not to have that much duplication here. Probably we could just extend the scripts we already have

autobuild/windows/autobuild_windowsinstaller_jack_1_prepare.ps1
```diff
@@ -0,0 +1,27 @@
...
+pip install aqtinstall
+echo "Get Qt 64 bit..."
+# intermediate solution if the main server is down: append e.g. " -b https://mirrors.ocf.berkeley.edu/qt/" to the "aqt"-line below
+aqt install --outputdir C:\Qt 5.15.2 windows desktop win64_msvc2019_64
+
+echo "Get Qt 32 bit..."
+# intermediate solution if the main server is down: append e.g. " -b https://mirrors.ocf.berkeley.edu/qt/" to the "aqt"-line below
+aqt install --outputdir C:\Qt 5.15.2 windows desktop win32_msvc2019
+
+echo "Install JACK2 64-bit..."
```
ann0see, line 21: Should be easy to do avoid duplication with an if

windows/deploy_windows_jack.ps1
```diff
@@ -0,0 +1,303 @@
...
+    {
+        if ($_ -Match "^VERSION *= *(.*)$")
+        {
+            $AppVersion = $Matches[1]
+            break
+        }
+    }
+
+    Invoke-Native-Command -Command "$WindowsPath\NSIS\makensis" `
+        -Arguments ("/v4", "/DAPP_NAME=$AppName", "/DAPP_VERSION=$AppVersion", `
```
ann0see, line 270: You should be able to pass an argument to the nsis system here

henkdegroot: @ann0see  Do you mean also modify the installer.nsi to handle both? Or can that be separate?
Keeping the installer_jack.nsi sounds a lot cleaner to me.

ann0see: If we have two installers that also means two places to maintain. The we must weigh how much "bloat" would be in one installer if we use one for both.

henkdegroot: My guess is that it will double the size when using a single installeer and the build process needs to altered for this as the "temp" deploy folders x86 and x86_64 will need to be different. Also I thought the sort of agreement was already to keep them seperate as you don't want to confuse the majority of the users. JACK adds a lot more complexticity for standard users.

jujudusud: I don't think it is a good Idea to have a single installer for now.
Because, for now, The use case is:
- You dont want to use JACK on Windows -> Install regular Jamulus and enjoy
- You want to use JACK on Windows -> you have to use a custom build of Jamulus (linux sound c++ build on windows).

ann0see: Ok. But if that’s the case, we could go for a separate file. Waiting for others to comment

pljones: Most PC JACK users a very techy, in my view.  (I'd say the same for most 32bit PC users, too, with some bit of kit they just can't leave behind.)

In fact, I'd fall on the side that offering JACK to everyone on Windows would cause more problems than it solves.  I'd be pretty firmly on the side of a separate installer.

ann0see: Ok. So then I think we agreed on a delegate installer. Probably giving a hint that this installs the JACK version would be good (same as the ASIO warning but without checks for now).

src/serverdlg.cpp
```diff
@@ -593,13 +593,13 @@ void CServerDlg::OnClearRecordingDirClicked()
 
 void CServerDlg::OnSysTrayActivated ( QSystemTrayIcon::ActivationReason ActReason )
 {
-    #ifdef _WIN32
+#ifdef _WIN32
```
ann0see, line 596: Attention on this one ;-)

ann0see (commented):
This looks much better! Thank you

src/serverdlg.cpp
henkdegroot: Hm....noticed this came in my PR after merging master into it. Guess I should somehow be able to "remove" this from the PR.

henkdegroot:
> This looks much better!

Will fix the standard build shortly, waiting for the jack build to complete to see if additaion fixes are needed.

src/serverdlg.cpp
henkdegroot: Is there a method I can use to remove this from my branch? I.e. using git revert perhaps?

windows/deploy_windows_jack.ps1
henkdegroot: > hint that this installs the JACK version

Excellent suggestion I would say, will look into updating the installer_jack.nsi and a this dialog.

src/serverdlg.cpp
ann0see: git rebase probably. But I‘m not an expert with that unfortunately.

windows/deploy_windows_jack.ps1
henkdegroot: Adding this dialog means adding text for this. The dialog will not have the button to link to the jamulus.io site for the installation of JACK but just text to indicate it needs JACK.

At the same time, this allows the additional check for "Is JACK installed?" option?
The check is very similar as with ASIO. Shall I just add this as well?

src/serverdlg.cpp
henkdegroot: It got included when I did the rebase to master....going to test/try...worst case I'll have to create a new PR.

windows/deploy_windows_jack.ps1
ann0see: > At the same time, this allows the additional check for "Is JACK installed?" option?

Yes. You could just check for jack.exe, I suppose?

henkdegroot: For ASIO the existence of a registry is checked, JACK has a similar registry key (when it is installed). So same logic can be used. Will add this to the installer for jack.

ann0see: Ah ok. That's great!

ann0see:
Just tested the JACK version on Windows and it seems to work quite well (at least with ASIO4ALL).

windows/deploy_windows.ps1
```diff
@@ -256,6 +257,19 @@ function Build-App-Variants
 # Build Windows installer
 Function Build-Installer
 {
+    param(
```
ann0see, line 260:
```suggestion
    param (
```

Style

windows/deploy_windows.ps1
```diff
@@ -256,6 +257,19 @@ function Build-App-Variants
...
+        [string] $BuildOption
+    )
+
+    if ($BuildOption -Eq "jackonwindows")
```
ann0see, line 264: Maybe good to add a comment here why we use a separate installer here.

windows/installer_jack.nsi
```diff
@@ -0,0 +1,443 @@
+; Jamulus NSIS Installer with Modern User Interface
+
```
ann0see, line 2:
```suggestion
; JACK on Windows installer

```

windows/installer_jack.nsi
```diff
@@ -0,0 +1,443 @@
...
+                MessageBox MB_OKCANCEL|MB_ICONEXCLAMATION "$(OLD_VER_REMOVE_FAILED)" /sd IDCANCEL IDOK continueinstall
+                goto quit
+            ${EndIf}
+
+            goto continueinstall
+
+        quit:
+            Abort
+
+    continueinstall:
```
ann0see, line 221: This could be removed in theory (since we could see this Jamulus Version as something totally different)

ann0see (commented):
Looks quite good! Thank you. Not yet approving since I'm waiting for others to test this.

henkdegroot:
@ann0see thank you for all valuable comments and thorough reviews you are doing! Going to update later

windows/deploy_windows.ps1
henkdegroot: Only 1 other "param" in the scripts is following this coding style (with the space), all others user param(.
There are also two using Param instead of param. 
Would it be better to update all to use the style "Param (" ?  Or "param ("?  
not sure what the preference would be for the of the first capital...if you let me know I can amend all and align them.

ann0see: Hmm. So I didn’t check that. Shame on me ;-). Personally I would prefer param( but that’s not the jamulus C++ style.

windows/installer_jack.nsi
henkdegroot: I don't mind removing this, but could this still help in cases you currently have an old version with ASIO and you now decide to install the JACK version (because it has been made available now)?

jujudusud: I think we should provide uninstalling any versions if possible.
Previous, ASIO, wrong install...

windows/deploy_windows.ps1
henkdegroot: Okay will make them all matching and use param( for this powershell script....

windows/deploy_windows.ps1
henkdegroot: Will be done in next commit

windows/installer_jack.nsi
henkdegroot: Will include in my next commit

windows/installer_jack.nsi
ann0see: Hmm. Probably true. But I'm in favor of removing it some day (it's just added logic we won't need in future). Yes, it does break backwards-compatibility, so maybe we'll need to wait a bit more.

ann0see:
I've updated the PR description, and I think we're ready to merge this if some more tests are made! Thank you very much. Probably after this one was merged, we're should tag a nightly (since having JACK on Windows is certainly helpful)

.github/workflows/autobuild.yml
```diff
@@ -140,6 +140,14 @@ jobs:
              cmd3_postbuild:        powershell .\autobuild\windows\autobuild_windowsinstaller_3_copy_files.ps1
              uses_codeql:           true
 
+           - config_name:           Windows JACK (artifact+codeQL)
+             target_os:             windows
+             building_on_os:        windows-latest
+             cmd1_prebuild:         powershell .\autobuild\windows\autobuild_windowsinstaller_1_prepare.ps1 -BuildOption jackonwindows
+             cmd2_build:            powershell .\autobuild\windows\autobuild_windowsinstaller_2_build.ps1 -BuildOption jackonwindows
+             cmd3_postbuild:        powershell .\autobuild\windows\autobuild_windowsinstaller_3_copy_files.ps1 -BuildOption jackonwindows
+             uses_codeql:           true
```
ann0see, line 149: I think there’s no need to run codeql here

henkdegroot: Wasn't sure how what the preferred way was to handle this. Removed in my last commit.

henkdegroot:
> we're should tag a nightly

That would be really good to have indeed.

henkdegroot:
@jujudusud , would you be able to run a JACK test with the latest installer?

henkdegroot:
Screenshots for the two dialogs which show when the installer is executed without having JACK installed:

Dialog 1:
jack-dialog-message

Dialog 2 (will be shown when you click next in dialog 1):
jack-warning-message

_NOTE: In the second dialog the buttons will be shown in the OS language....which is Dutch in my case :)_

ann0see:
... but we couldn't find one on your PC. I think you should say:

This Jamulus version needs the JACK Audio Connection Kit to work, but it doesn't seem to be installed on your PC. You should install JACK for Windows first. Do you still want to continue with the installation of Jamulus without installing JACK first?

src/res/translation/wininstaller/de.nsi
```diff
@@ -22,10 +22,10 @@ LangString ASIO_DRIVER_HEADER ${LANG_GERMAN} \
      "ASIO Treiber"
 
 LangString ASIO_DRIVER_SUB ${LANG_GERMAN} \
-     "Jamulus benötigt einen ASIO Treiber"
+     "${APP_NAME} benötigt einen ASIO Treiber"
```
ann0see, line 25: Should have been a separate PR ;-)

ann0see (commented):
Todo: 

- [ ] Final tests for ASIO and JACK build
- [ ] In deph review (I don't think there's much left)
- [ ] Proposal for ChangeLog
- [ ] Merge

Anything else?

jujudusud:
> @jujudusud , would you be able to run a JACK test with the latest installer?

Yes, I will have a look tonight.
Could you please tell the link were the .exe is available ?
Don't where are the temporary assets. :-(

hoffie:
> Could you please tell the link were the .exe is available ?
> Don't know where are the temporary assets. :-(

https://github.com/jamulussoftware/jamulus/suites/2923337005/artifacts/65615938

"Show all checks" in this PR, then Details, then Artifacts in the top right corner.

henkdegroot:
> Anything else?

Not sure how the remain updates should be handled. In my view the following needs updating (at some point in time) as well:

Readme.md - change Windows (ASIO) to Windows (ASIO or JACK)?
Installer text updates for all languages - most likely needs a task somewhere to ensure translator are made aware about changes in the installer res lang file.
Add some reference on the "Installation for Windows" page to hint this is available.

src/res/translation/wininstaller/de.nsi
henkdegroot: Do you want me to undo and create a seperate PR for that? DE is not the only language in which I have updated that.

ann0see: No. It’s fine like that. Just for future notice it would be better to have one PR per change.

ann0see:
> Readme.md - change Windows (ASIO) to Windows (ASIO or JACK)?

Yes. This could be included in this PR

>  Installer text updates for all languages - most likely needs a task somewhere to ensure translator are made aware about changes in the installer res lang file.

Open an issue to track this?

>  Add some reference on the "Installation for Windows" page to hint this is available.

--> website repo:

JACK documentation
or similar?

windows/installer_jack.nsi
```diff
@@ -0,0 +1,444 @@
+; Jamulus NSIS Installer with Modern User Interface
+; Installer is build with support for JACK on Windows
```
ann0see, line 2:
```suggestion
; Installer for Jamulus with support for JACK on Windows
```
Grammar seems to be strange otherwise (build vs. built --> avoid it ;-))

ann0see (approved):

ann0see:
Works well on Win 10 (64Bit) with 64Bit JACK. I think all combinations were already tested with in the other PR.

henkdegroot:
> Yes. This could be included in this PR

Done

> --> website repo:

Looks good to me



> Open an issue to track this?

Done. I created #1844 for this.

jujudusud:
It's also ok here W10 64 bits.

pljones:
From @ann0see above:

----
Todo: 

- [ ] Final tests for ASIO and JACK build
- [ ] In deph review (I don't think there's much left)
- [ ] Proposal for ChangeLog
- [ ] Merge

Anything else?

----

How are things progressing?

ann0see:
Documentation is missing (at least I think it is) and the tests did succeed as I mentioned above.

jujudusud:
I think we should talk about the documentation in the website repository.

jujudusud (approved):

autobuild/windows/autobuild_windowsinstaller_2_build.ps1
```diff
@@ -27,4 +28,11 @@ if (("$jamulus_project_path" -eq $null) -or ("$jamulus_project_path" -eq "")) {
 
 echo "Build installer..."
 # Build the installer
-powershell "$jamulus_project_path\windows\deploy_windows.ps1" "C:\Qt\5.15.2"
+if ($BuildOption -Eq "jackonwindows")
```
pljones, line 31: `-Ne """` - that way, whatever option is passed in gets passed through.

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
```diff
@@ -33,11 +34,24 @@ if (($jamulus_buildversionstring -eq $null) -or ($jamulus_buildversionstring -eq
 
 # Rename the file
 echo "rename"
-$artifact_deploy_filename = "jamulus_${Env:jamulus_buildversionstring}_win.exe"
-echo "rename deploy file to $artifact_deploy_filename"
-cp "$jamulus_project_path\deploy\Jamulus*installer-win.exe" "$jamulus_project_path\deploy\$artifact_deploy_filename"
-
+if ($BuildOption -Eq "jackonwindows")
```
pljones, line 37: If `BuildOption` were `jack`, rather than `jackonwindows`, you could set the deploy filename like
`..._${BuildOption}_win.exe`  And then the `cp` below would be `installer-${BuildOption}-win.exe`...

henkdegroot: I know, but that is not the buildoption name. There also is a jackonmac option, so believe it will be best to keep it like this to avoid confusion.

autobuild/windows/autobuild_windowsinstaller_2_build.ps1
henkdegroot: Sounds good....to me. Will update the PR.

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
pljones: Ah, right.  Thought it was probably something like that.  Maybe one set up at the top to create the insert value (empty by default, then set to "jack" if jack, and then the insert value is inserted appropriately if not empty)?  I'm thinking if we have PortAudio as well, that could then just be one more value that is recognised once by the script, without needing to modify everywhere all over again.

henkdegroot: Okay, I would propose the following in the autobuild-copy-files.ps1 file:

Add the beginning: When a BuildOption is provided and this option is "jackonwindows", set the BuildOption to "jack"

Now there will be two options to copy/rename the final articfact name:

1. No BuildOption set, handle as currently and don't alter the installer name.
2. When BuildOption is set, include the BuildOption in the installer name as part of the "artifact_deploy_filename"

This should allow the flexibility to create different output artifact names, depending on the option specified.

Please let me know if you agree with this approach, so I can update the script and PR.

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
```diff
@@ -33,11 +34,24 @@ if (($jamulus_buildversionstring -eq $null) -or ($jamulus_buildversionstring -eq
...
+{
+    $artifact_deploy_filename = "jamulus_${Env:jamulus_buildversionstring}_jack_win.exe"
+}
+else
+{
+    $artifact_deploy_filename = "jamulus_${Env:jamulus_buildversionstring}_win.exe"
+}
 
+echo "rename deploy file to $artifact_deploy_filename"
+if ($BuildOption -Eq "jackonwindows")
```
pljones, line 47: This should be "not empty string".

windows/deploy_windows.ps1
pljones: I'd like the "installer suffix" passed in.  If it's not empty, set the name to "installer_${InstallerSuffix}.nsi" otherwise "installer.nsi".  That's because the portaudio installer will also be separate and should have a different name.

windows/deploy_windows.ps1
```diff
@@ -256,6 +257,22 @@ function Build-App-Variants
...
 {
+    param(
+        [string] $BuildOption
+    )
+
+    # Building the installer with JACK support on Windows is using a separate nsi script which creates a separate installer.
+    # The installer with JACK on Windows is made available for more users who know JACK, ASIO will be kept as the mainstream version.
+    # When not building with JACK support, the installer is created with ASIO support.
+    if ($BuildOption -Eq "jackonwindows")
+    {
```
pljones, line 268: I'd like the "installer suffix" passed in.  If it's not empty, set the name to "installer_${InstallerSuffix}.nsi" otherwise "installer.nsi".  That's because the portaudio installer will also be separate and should have a different name.

windows/installer_jack.nsi
```diff
@@ -0,0 +1,444 @@
+; Jamulus NSIS Installer with Modern User Interface
+; Installer for Jamulus with support for JACK on Windows
```
pljones, line 2: How much of this is identical to the standard installer?  Can those parts be factored out and !included here?

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
pljones: I've had a quick look through for anything else and added comments.  But the idea is that _any_ part that can be factored out and used the same way _should be_ factored out.  So it's not just the deploy artifact name: there are a few other bits.

windows/installer_jack.nsi
henkdegroot: Main difference are the custom ASIO installed check dialogs, which have been replaced by custom JACK installed check dialogs. I believe we should be able to pass on the buildoption into this script as well and use conditions in the script to enabled/disable some parts. The reason why a separate file was used was mainly to keep it simple/easy and clear.

pljones: I'm worried where things that _should_ be common are separated and that commonality is no longer apparent.

I don't mind having things separate if it makes the behaviour easier to achive but I'd like the commonality clear - preferably by having it stored in a common file.

ann0see: Yes, I also [wondered why we need two files](https://github.com/jamulussoftware/jamulus/pull/1829#discussion_r645861615). If possible, we should unify them.

henkdegroot: I have checked now and it is actually not that complex to use 1 nsi installer file. So will update to use just 1 installer.nsi.
The installer.nsi will include a buildoption value, so based on that different actions can be performed in the installer.

windows/installer.nsi
```diff
@@ -365,45 +379,81 @@ Function createdesktopshortcut
...
+
+        JACKExists:
+
+    FunctionEnd
+
+    Function ExitJACKInstalled
+        ClearErrors
+        EnumRegKey $0 HKLM "SOFTWARE\JACK" 0
+        IfErrors 0 SkipMessage
+            MessageBox MB_YESNO|MB_ICONEXCLAMATION "$(JACK_EXIT_NO_DRIVER)" /sd IDNO IDYES SkipMessage
```
ann0see, line 411: I hope SkipMessage jumps to line 413 not 453. It doesn't matter here anyway.

What about an !include which only gets included during compile time depending on which version we're building?

windows/deploy_windows.ps1
```diff
@@ -282,10 +270,21 @@ Function Build-Installer
...
-    Invoke-Native-Command -Command "$WindowsPath\NSIS\makensis" `
-        -Arguments ("/v4", "/DAPP_NAME=$AppName", "/DAPP_VERSION=$AppVersion", `
-        "/DROOT_PATH=$RootPath", "/DWINDOWS_PATH=$WindowsPath", "/DDEPLOY_PATH=$DeployPath", `
-        "$WindowsPath\$InstallerNSI")
+    if ($BuildOption -ne "")
+    {
+        Invoke-Native-Command -Command "$WindowsPath\NSIS\makensis" `
+            -Arguments ("/v4", "/DAPP_NAME=$AppName", "/DAPP_VERSION=$AppVersion", `
+            "/DROOT_PATH=$RootPath", "/DWINDOWS_PATH=$WindowsPath", "/DDEPLOY_PATH=$DeployPath", `
+            "/DBUILD_OPTION=$BuildOption", `
```
ann0see, line 278: Can't we pass an empty BUILD_OPTION? This would avoid the if here.

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
```diff
@@ -27,31 +27,30 @@ if (($jamulus_buildversionstring -eq $null) -or ($jamulus_buildversionstring -eq
     $jamulus_buildversionstring = "NoVersion"
 }
 
+# Amend BuildOption when needed, only needed when the name build option in the artifact needs to be different
+if ($BuildOption -Eq "jackonwindows")
```
ann0see, line 31: Should be a switch if possible in PS. We might add more APIs like PortAudio in future

windows/installer.nsi
```diff
@@ -365,45 +379,81 @@ Function createdesktopshortcut
...
+            Pop $Label
+            ${NSD_CreateButton} 0 21u 100% 15u "$(ASIO_DRIVER_MORE_INFO)"
+            Pop $Button
+            ${NSD_OnClick} $Button OpenASIOHelpPage
+
+            nsDialogs::Show
+
+        ASIOExists:
+
+    FunctionEnd
```
ann0see, line 441: Does NSIS allow returning bools from functions? If so, it would be great to have something like 

if (apiRegKeyExists()) {
...
} else {
  throw error message.
}

pljones: You have "then" and "else" targets, from what I remember.

pljones: https://nsis.sourceforge.io/Macro_vs_Function#Hybrid

Something like
* Push the return value onto the stack
* Pop it in the caller
* If test to switch between the targets

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
henkdegroot: BuildOption is now use when renaming the generated installer. The installer name remains generic until the artifact is deploy/uploaded.

windows/deploy_windows.ps1
henkdegroot: No longer using a separate installer script.

windows/installer.nsi
henkdegroot: The page and functions are not "seen" by the code, they are removed/ignored in the compile step. For this reason I also had to make sure variables that were not used for the "jackonwindows" option were not defined, as nsis would complain about being there and not used.

windows/deploy_windows.ps1
henkdegroot: I tried that, but when the command is invoked, it generates a script error. So we probable would end up having another "if" somewhere else to cover that.

windows/installer.nsi
henkdegroot: I would say this does not make it clearer to me, just adds extra complexity to the code.

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
henkdegroot: Updated in latest commit to use switch command

henkdegroot:
@pljones  @ann0see  
Currently the autobuild copy ps1 script includes the BUILD_OPTION like this:

Standard version (ASIO):
jamulus_3.8.0dev-5df8fca_win.exe

JACK version:
jamulus_3.8.0dev-5df8fca_jack_win.exe

I notice the legacy mac artifact has the "legacy" text after the word mac.

Should I update the windows copy script to append the BUILD_OPTION?
I,e. jamulus_3.8.0dev-5df8fca_win_jack.exe

pljones:
> I notice the legacy mac artifact has the "legacy" text after the word mac.

Hm.  Only because it got there first :) -- but yes.

autobuild/windows/autobuild_windowsinstaller_2_build.ps1
```diff
@@ -27,4 +28,11 @@ if (("$jamulus_project_path" -eq $null) -or ("$jamulus_project_path" -eq "")) {
...
 # Build the installer
-powershell "$jamulus_project_path\windows\deploy_windows.ps1" "C:\Qt\5.15.2"
+if ($BuildOption -ne "")
+{
+    powershell "$jamulus_project_path\windows\deploy_windows.ps1" "C:\Qt\5.15.2" -BuildOption $BuildOption
+}
+else
+{
+    powershell "$jamulus_project_path\windows\deploy_windows.ps1" "C:\Qt\5.15.2"
+}
```
hoffie, line 38: Not sure about Windows/PowerShell behavior, but if passing an empty parameter is an option (maybe needs quoting?) the logic could be simplified?

```suggestion
powershell "$jamulus_project_path\windows\deploy_windows.ps1" "C:\Qt\5.15.2" -BuildOption "$BuildOption"
```

windows/deploy_windows.ps1
hoffie: Hm, maybe there's still a way to avoid duplicating this rather long `Invoke-Native-Command` call?

windows/installer.nsi
```diff
@@ -40,7 +40,10 @@ BrandingText "${APP_NAME}. Make music online. With friends. For free."
 
 ; Add support for copying registry keys
 
-!insertmacro COPY_REGISTRY_KEY
+!if ${BUILD_OPTION} != "jackonwindows"
+    ;Only insert macro when not building for JACK
```
hoffie, line 44:
```suggestion
    ; This macro is only needed when using ASIO4ALL
```

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
```diff
@@ -26,20 +27,32 @@ if (($jamulus_buildversionstring -eq $null) -or ($jamulus_buildversionstring -eq
     $jamulus_buildversionstring = "NoVersion"
 }
 
+# Amend BuildOption when needed, only needed when the name build option in the artifact needs to be different
+switch ($BuildOption)
+{
+    "jackonwindows"   {$BuildOption = "jack"; break}
+}
+
```
hoffie, line 35:
```suggestion
$artifact_suffix = ""
# Certain build options lead to very different builds and must therefore use another artifact name:
switch ($BuildOption)
{
    "jackonwindows"   {$artifact_suffix = "_jack"; break}
}

```

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
```diff
@@ -26,20 +27,32 @@ if (($jamulus_buildversionstring -eq $null) -or ($jamulus_buildversionstring -eq
...
-$artifact_deploy_filename = "jamulus_${Env:jamulus_buildversionstring}_win.exe"
+if ($BuildOption -ne "")
+{
+    $artifact_deploy_filename = "jamulus_${Env:jamulus_buildversionstring}_win_${BuildOption}.exe"
+}
+else
+{
+    $artifact_deploy_filename = "jamulus_${Env:jamulus_buildversionstring}_win.exe"
+}
+
```
hoffie, line 51: If the suggestion from above is applied, this could be simplified like that?

```suggestion
$artifact_deploy_filename = "jamulus_${Env:jamulus_buildversionstring}_win${artifact_suffix}.exe"
```

hoffie (commented):
Thanks! This looks good to me (code-wise), although I cannot test. Some minor nits/ideas for simplification.

Leaving final approval to a developer with Windows access. :)

autobuild/windows/autobuild_windowsinstaller_2_build.ps1
henkdegroot: Passing an empty parameter is not a big issue, however specify an option without a value (even empty) seems to be an issue.
To keep it consistent in the scripts and allowing a "build" without specifying an option, I have set it up like this.

windows/installer.nsi
henkdegroot: Understand the idea, but I used this text to be inline with the != option. So it will use the macro for ASIO4ALL, but also for other buildoptions they may be added in the future.

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
henkdegroot: I believe that would mean you always have to add your buildoption to the switch statement, even if you do not have to change the buildoption name. I.e. an option portaudio would not need to be changed and can be used a suffix but with the above logic you would need to add this option to the switch statement to ensure the _ is added. Otherwise, if no buildoption is specified you get an _ + "" as the suffix.

pljones: @henkdegroot, please see @hoffie's comment on lines 30-34 and note the default value of `artifact_suffix`.

windows/deploy_windows.ps1
pljones: I guess you could put the `-Arguments` into a list, apart from `/DBUILD_OPTION`, then - if `$BuildOption` is set, insert an extra argument into the list in the last but one position, then pass th list of arguments to `-Arguments`.

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
henkdegroot: This means that you always need to add every buildoption you want to use in the output name of the artifact in the switch statement. Not that I mind that is needed, but to me that would be extra changes that would not be required when using the current proposed solution.

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
henkdegroot: @pljones @hoffie , for me using the "artifact_suffix" is not really adding any value and will force you to always set it, even if you want to use the same as the "buildoption". Is that really what is desired? The only reason why the switch is used, is because the text "jackonwindows" is not a nice one to use in the output name of the artifact. If another option would be used, for instance "portaudio", then there is no need to change the name, but when using the artifact_suffix it is required to add this to the switch statement.

henkdegroot: @hoffie 
> Hm, maybe there's still a way to avoid duplicating this rather long Invoke-Native-Command call?

Not sure, I believe every option will require some kind of "if" solution. To me the current option seems to be clear when you look at the code and what it is doing.

autobuild/windows/autobuild_windowsinstaller_2_build.ps1
hoffie: Hrm, too bad. :(
Then it will probably have to stay like this.

windows/installer.nsi
hoffie: > So it will use the macro for ASIO4ALL, but also for other buildoptions they may be added in the future.

Hrm, yes, I think this is what would happen. However, the intention of the code is to limit the macro inclusion to builds which support ASIO/ASIO4ALL, isn't it?
And sadly we cannot check this explicitly with the current code, which is why we resort to the negative check?

I'm usually in favor of adding comments about the (non-obvious) intention of code rather than comments which repeat (and maybe elaborate on) what the code already says.

I wonder if we should do both here?

```suggestion
    ; This macro is only needed when using ASIO4ALL.
    ; Technically, we cannot detect this easily.
    ; Therefore, we only insert the macro when not building for JACK.
```

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
hoffie: Hrm. I guess it's about different approaches:
- The initially proposed approach basically passes through buildoptions (would it handle multiple build options, btw?) and just special-cases jackonwindows by tuning the resulting name, right?
- My suggestion would make this tuning explicit and a requirement (to be correct). Build options other than jackonwindows would yield to binaries with the generic name, which may be a downside as well.

So, maybe a solution would be:
- Stick to the initial proposal and make sure that multiple build options would be handled correctly (or at least that the code would fail explicitly if it can't handle it easily).
- Think about renaming "jackonwindows" (and "jackonmac") to just "jack", which would make this special-casing redundant. This should be considered carefully and would probably be a follow-up PR.

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
hoffie: Agreed, see other comment thread about jackonwindows regarding the remaing points.

autobuild/windows/autobuild_windowsinstaller_3_copy_files.ps1
henkdegroot: > would it handle multiple build options, btw?

Currently not considered to handle multiple buildoptions in 1. Hence why the name BuildOption is not plural.
The idea was that when you use a BuildOption, it always results in a separate installer output. This way the criteria for the buildoption can be made specific to the buildoption (i.e. some may need additional components to be installed, others may just need to skip some items, etc.)

> (or at least that the code would fail explicitly if it can't handle it easily)
I did not test, but believe the build will currently fail if you would specify more then one item.

> Think about renaming "jackonwindows" (and "jackonmac") to just "jack"
Sounds like a good next step and would remove the need to renaming the buildoption in the artifact.

ann0see:
@pljones ready to merge?

pljones (approved):

pljones:
Needs a rebase and squash?

ann0see:
I think so. And maybe even a "only build on release" option in autobuild. But that’s out of scope

ann0see:
Probably only the README file has conflicts. That doesn’t seem to be too difficult to solve

PR #1838, opened 2021-06-06
Do not reset fMuteOutStreamGain when reinit sound parameters

ngocdh:
See discussion #909.

Expected behaviour, as described here https://github.com/jamulussoftware/jamulus/issues/909#issuecomment-769383616

```
MUTE MYSELF = false and MUTE = false. I DO hear myself returned from the server.
MUTE MYSELF = false and MUTE = true. I DO NOT hear myself returned from the server.
MUTE MYSELF = true and MUTE = false. I DO NOT hear myself returned from the server. I DO hear myself locally.
MUTE MYSELF = true and MUTE = true. I DO NOT hear myself returned from the server. I DO NOT hear myself locally. In other words, I hear nothing.
```

However, when MUTE MYSELF = true and MUTE = true, change any sound parameter (buffer size, channels, quality, etc.), you will hear yourself locally, meaning expected behaviour is no longer respected. 

This is because of line 930: `fMuteOutStreamGain = 1.0f;` in `CClient::Init()`, called by every parameter change. 

This line is actually not necessary, because CClient::CClient() already does it.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -927,7 +927,7 @@ void CClient::Init()
     vecZeros.Init ( iStereoBlockSizeSam, 0 );
     vecsStereoSndCrdMuteStream.Init ( iStereoBlockSizeSam );
 
-    fMuteOutStreamGain = 1.0f;
+    //fMuteOutStreamGain = 1.0f;
 
     opus_custom_encoder_ctl ( CurOpusEncoder,
                               OPUS_SET_BITRATE ( CalcBitRateBitsPerSecFromCodedBytes ( iCeltNumCodedBytes, iOPUSFrameSizeSamples ) ) );
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -927,8 +927,6 @@ void CClient::Init()
     vecZeros.Init ( iStereoBlockSizeSam, 0 );
     vecsStereoSndCrdMuteStream.Init ( iStereoBlockSizeSam );
 
-    // fMuteOutStreamGain = 1.0f;
-
     opus_custom_encoder_ctl ( CurOpusEncoder,
                               OPUS_SET_BITRATE ( CalcBitRateBitsPerSecFromCodedBytes ( iCeltNumCodedBytes, iOPUSFrameSizeSamples ) ) );
 
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -927,7 +927,7 @@ void CClient::Init()
     vecZeros.Init ( iStereoBlockSizeSam, 0 );
     vecsStereoSndCrdMuteStream.Init ( iStereoBlockSizeSam );
 
-    //fMuteOutStreamGain = 1.0f;
+    // fMuteOutStreamGain = 1.0f;
 
     opus_custom_encoder_ctl ( CurOpusEncoder,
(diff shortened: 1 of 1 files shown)


src/client.cpp
```diff
@@ -927,7 +927,7 @@ void CClient::Init()
     vecZeros.Init ( iStereoBlockSizeSam, 0 );
     vecsStereoSndCrdMuteStream.Init ( iStereoBlockSizeSam );
 
-    fMuteOutStreamGain = 1.0f;
+    // fMuteOutStreamGain = 1.0f;
```
hoffie, line 930: If this PR is merged, the line should be removed completely (not just commented). git has all the history already.
Should probably condense the two blank lines to one then.

hoffie (commented):
Thanks!

For reference:
- This PR does not attempt to change the discussed behavior of Mute myself + Mute in itself, but rather tries to make the existing behavior (local monitoring) more consistent, as far as I undertand. The linked issue is rather lengthy and is about the whole topic. The relevant comment is https://github.com/jamulussoftware/jamulus/issues/909#issuecomment-804332596.
- The initial commit which introduced the line in question was a49cb6408 (9171caa41 converted doubles to floats).

I have a hard time judging if the removal of this line could have other unintended side effects in behavior. :(

ngocdh:
> This PR does not attempt to change the discussed behavior of Mute myself + Mute in itself, but rather tries to make the existing behavior (local monitoring) more consistent, as far as I undertand. 

Yes correct.

> I have a hard time judging if the removal of this line could have other unintended side effects in behavior. :(

I've been testing this for a few hours. Haven't seen unintended side effects.

ann0see:
The actual question is why this line was added and if we need a conditional. Does the mentioned commit contain any information?

softins:
> The actual question is why this line was added and if we need a conditional. Does the mentioned commit contain any information?

I've just looked at the files in question, and I agree with @ngocdh that the re-init of `fMuteOutStreamGain` in `CClient::Init()` is not necessary. The initialisation of it in the constructor for `CClient` should be enough to ensure the value is always in range before the user moves their slider, and then it should always be allowed to remember the value it was set to.

ngocdh:
Because it's in `CClient::Init()`, so maybe they thought to initialize `dMuteOutStreamGain` there (renamed to `fMuteOutStreamGain` later).

However, this variable must always match the actual own fader value, thus line 338 (in `CClient::SetRemoteChanGain()`). And when there's a parameter change, `CClient::Init()` is called again resetting `fMuteOutStreamGain` back to 1.0, while the own fader value stays the same. This is when the inconsistency happens.

We don't need that line in `CClient::Init()`, because it's already initialized in `CClient::CClient()`.

softins (approved):

ann0see:
> Because it's in CClient::Init(), so maybe they thought to initialize dMuteOutStreamGain there (renamed to fMuteOutStreamGain later).

Yes, this could be an explanation.

ann0see (approved):
Will approve and merge it.

hoffie:
Thanks!

CHANGELOG: Improve mute behavior after sound parameter adjustments.

PR #1859, opened 2021-06-11
multiline/monospaced chat messages

passing:
https://github.com/jamulussoftware/jamulus/discussions/1021

add multiline input mode in the chat dialog (disabled when client starts)
apply monospaced format on multiline chat messages in server

### DIFF
--- a/src/chatdlg.cpp
+++ b/src/chatdlg.cpp
@@ -35,7 +35,7 @@ CChatDlg::CChatDlg ( QWidget* parent ) : CBaseDlg ( parent, Qt::Window ) // use 
 
     txvChatWindow->setAccessibleName ( tr ( "Chat history" ) );
 
-    // input message text
+    // single-line input message text
     edtLocalInputText->setWhatsThis ( "<b>" + tr ( "Input Message Text" ) + ":</b> " +
                                       tr ( "Enter the chat message text in the edit box and press enter to send the "
                                            "message to the server which distributes the message to all connected "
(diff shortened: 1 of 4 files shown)


passing:
Tested on Linux.
Please verify it is fine on Windows and macOS

image

ann0see:
This needs changes on the client and server side? How do we ensure that users don’t get confused why it "doesn’t work" because a server is not up to date?

passing:
> This needs changes on the client and server side? How do we ensure that users don’t get confused why it "doesn’t work" because a server is not up to date?

On an old server, the multiline messages with be merged into a single line
see: https://github.com/jamulussoftware/jamulus/discussions/1021#discussioncomment-835951
Do you have an idea how to handle this / or accept it?

ann0see:
Either the client gives a warning (as placeholder text (?)) or we leave it like that.

passing:
I could add something like this to the placeholder text:
image
WDYT?

pljones:
The client _knows_ the server version.  It  shouldn't be operating in this mode unless the server is offering the feature.  It should revert to the existing behaviour.  It should only offer this if requested, too (so a settings entry is needed).  The server should also need configuring to offer it, perhaps - not sure on whether that's really needed, though.

passing:
@pljones The chat window will start with the existing behavior:
image
The new input mode can then be selected via the menu or by keyboard shortcut (Ctrl+M):
image

> The client _knows_ the server version. It shouldn't be operating in this mode unless the server is offering the feature.
> The server should also need configuring to offer it, perhaps - not sure on whether that's really needed, though.

Note that this doesn't really change what a client is sending to the server. You can actually already enter multiline messages in the single-line input field by just copying&pasting them from somewhere. The 2nd input mode just makes this easier.
A multiline message sent to an old server is getting merged into a single line, just like it is already happening when you paste a multiline message into the input field.
Because of that, I would prefer to keep it simple and to not add an extra server option for this.

Btw. I had described how I wanted to implement this and asked for feedback here https://github.com/jamulussoftware/jamulus/discussions/1021#discussioncomment-835951 . You probably just missed that - but not having got any objections I proceeded.
I'd like to contribute to the project more often when the time allows, but would really favor to be able to agree on a feature before starting to implement.

pljones:
Sorry, missed the link in the top post of the PR, yes!

Having the option on the chat window is good, the short cut is even better.  I'd still like the client to check the server version and only "make it easy" when it's going to actually work.  I'm not particularly fussed about the server side option, to be honest, so long as it's not opening things up.  Work was done to cut down what text was allowed through the server at one point.

passing:
> I'd still like the client to check the server version and only "make it easy" when it's going to actually work

I'm not sure about this. Right now the chat dialog even allows for sending a message even you are not connected to a server (it will just disappear then). So disallowing multiline messages on older servers will need quite some extra code / additional complexity, especially when you want to react to switching from a new server to an old server.

hoffie:
1. Permitting the feature even if the server does not handle it in the best way, may be confusing or lead to the conclusion that the feature is broken
2. Enabling/disabling the option depending on the server version may be equally confusing when switching servers and seeing it work / stop working.

Given that both options have downsides, I'd go for the technically simpler option 1 of offering this feature unconditionally (and maybe outlining the requirement in the changelog).

I'd rather not make this server-configurable unless there are specific reasons to do that.

(Haven't looked at the code in detail yet)

pljones:
I can't see anything on the server side that would need to be configurable - it maintains the safety, just adds the `<pre>` tags if it spots the `\n` character.

Having broken things not get improved because they're already broken isn't a good argument, in my view.  It generally means people don't make things better.

pljones:
Just to add

> Enabling/disabling the option depending on the server version may be equally confusing when switching servers and seeing it work / stop working.

What, more confusing that having no indication that it's not going to work until after you've spent time formatting the input, only to lose your effort when the server doesn't add the tags?

passing:
Closing as this needs more alignment: https://github.com/jamulussoftware/jamulus/discussions/1021#discussioncomment-900153

pljones:
Thanks for the effort so far and the explanation in the discussion of where we've got to, @passing.

PR #1512, opened 2021-04-13
iOS sound support

ngocdh:
iOS now has stereo sound!

>I had to modify opus lib because of "Compile sources as" "Objective C++" option when compiling for iOS.
-> not anymore thanks to @jeroenvv

Note that for the UI to work nicely (able to quit the settings window for example), #1450 is required.

TODO: Bluetooth headset not working correctly (maybe because of different sample rate). -> or is it, because Android Bluetooth is working fine?

Update: I also added a feature for mobile devices (iOS and Android) in this PR: possibility to switch to built-in mic instead of the external device chosen automatically by the OS, and back. It's helpful when I'm jamming on mobile without external soundcard and want to talk when I'm not playing. More info here: https://github.com/jamulussoftware/jamulus/issues/1512#issuecomment-824713252_

PR #1865, opened 2021-06-12
iOS sound support 2

ngocdh:
I messed #1512 up. So this is a brand new branch for iOS sound support. I hope this is not creating much trouble for the team.

~~TODO: clang-format~~


ngocdh:
closed because of some unwanted files (jamulus_qml_plugin_import.cpp and Release-iphoneos/)

ngocdh:
Reopening in case team wants to merge this instead of #1512

android/sound.cpp
```diff
@@ -345,3 +345,36 @@ void CSound::Stats::log() const
              << "frames_in: " << frames_in << ",frames_out: " << frames_out << ",frames_filled_out: " << frames_filled_out
              << ",in_callback_calls: " << in_callback_calls << ",out_callback_calls: " << out_callback_calls << ",ring_overrun: " << ring_overrun;
 }
+
```
pljones, line 348: Why is this android file changing in an iOS change?

ios/sound.h
```diff
@@ -2,7 +2,7 @@
  * Copyright (c) 2004-2020
  *
  * Author(s):
- *  ann0see based on code from Volker Fischer
+ *  ann0see and ngocdh based on code from Volker Fischer
```
pljones, line 5: and the copyright date could do with updating.

ios/sound.mm
```diff
@@ -2,7 +2,7 @@
  * Copyright (c) 2004-2020
  *
  * Author(s):
- *  ann0see based on code from Volker Fischer
+ *  ann0see and ngocdh based on code from Volker Fischer
```
pljones, line 5: and the date.

ios/sound.mm
```diff
@@ -25,208 +25,295 @@
...
-  //CSoundBase::Init ( iCoreAudioBufferSizeMono );
-
-  // set internal buffer size value and calculate stereo buffer size
-  //iCoreAudioBufferSizeStereo = 2 * iCoreAudioBufferSizeMono;
-
-  // create memory for intermediate audio buffer
-  //vecsTmpAudioSndCrdStereo.Init ( iCoreAudioBufferSizeStereo );
-
-  return iCoreAudioBufferSizeMono;
+void checkStatus(int status){
```
pljones, line 32: Why are these global?

Also style is wrong here.  I think clang-format is ignoring the .mm files, which it should treat like .cpp files.

src/chatdlg.cpp
```diff
@@ -65,6 +65,9 @@ CChatDlg::CChatDlg ( QWidget* parent ) : CBaseDlg ( parent, Qt::Window ) // use
     connect ( action, SIGNAL ( triggered() ), this, SLOT ( close() ) );
 #endif
 
+#if defined( ANDROID ) || defined( Q_OS_ANDROID )
```
pljones, line 68: Again, this change is for iOS, so why this android change?

src/clientdlg.cpp
```diff
@@ -375,6 +375,17 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     pMenu->addMenu ( pEditMenu );
     pMenu->addMenu ( new CHelpMenu ( true, this ) );
 
+#if defined( Q_OS_IOS ) or defined( Q_OS_ANDROID ) or defined( ANDROID )
```
pljones, line 378: I'd suggest this whole change - switching input device - be split out from the main pull request.  It will make more sense as a separate change to have both iOS and android support then.  Here the android bit looks lost.

pljones:
OK, now I can see what's going on, I'd like the introduction of the new feature of selecting input device split into a separate change, dependent upon this one.

There's a lot happening here already.  Having that in as an "different thing" just makes this PR too big, in my opinion.

Also, there are too many commits to simply merge this - we'd have to squash and that's a bit messy.  Or someone would have to tidy it up a bit.

ngocdh:
Thank you @pljones for the thorough review. I’ll check the .mm file style.

About the input selection feature, it would be complicated for me to split, now that the iOS PR is not yet merged.

And yeah I’d love to have my commit tidied up.

ann0see:
> About the input selection feature, it would be complicated for me to split, now that the iOS PR is not yet merged.

Yes, but the other problem is we can't merge this as is (https://github.com/jamulussoftware/jamulus/pull/1865#issuecomment-860204403). Can you have a look at related commits and describe which features this PR exactly adds?

ngocdh:
Ok so here comes some explanation for this PR:
- My 1st commit (https://github.com/jamulussoftware/jamulus/pull/1865/commits/d9abe62124b7d6074504f216ca57b37693a14674) was for iOS sound support, meaning iOS sound fully worked in this commit
- While testing on my iPhone, I felt the need to implement a feature: "switching between intenal/external microphones/devices". However, the feature depends on the first unmerged commit, so I couldn't just create another unrelated branch. I had to continue to develop on the same branch. 2 next commits did this task.
- While that feature worked great on iPhone, I thought maybe Android could benefit the same, so I wrote the Android part. That's when things got messy: I could have written the Android part in a separated branch, but since the same feature was written for iOS, I figured maybe I could just add code in the same branch.
I also didn't have a proper dev environment for Android, so I wrote directly on github.com and wait for the auto-build to finish to test the apk (yeah I know you can laugh at me). The following long list of commits were because of this. (Actually it's a bit more complicated, this feature is not yet fully complete in Android: it's working, but you have to know how to make it work, i.e. it's not a few clicks away. For it to be complete: java code, and thus a bridge must be involved).
- Then some more commits to reduce the launch time in iOS, some to fix a socket crash in socket.cpp, specific to iOS environment, some to fix clang style.

The key here is: that additional feature could only have been implemented in a separate PR if the iOS sound support code had already been merged.

ann0see:
Ok. Thanks for this explanation. This means that we can probably reduce the number of commits dramatically:

Commit 1: Add iOS sound support 

This should also include the code which fixes the crash, formatting and the removal of the libraries to reduce the launch time.

Commit 2: Allow automatic switching between internal/external audio devices (iOS)
You can branch off an unmerged branch by using `git checkout -b myNewFeature` while being on the branch with the changes you made.

Commit 3: Add initial auto switching of audio device to Android

You'd squash all the related commits into one or even better use `fixup` in an interactive rebase.

ngocdh:
I'm not sure what to do.
I tried `git rebase -i` to squash commits related to ios sound only and already got conflicts...

The best thing I can do now, I think is to create a new branch from jamulus master, manually applied changes from this branch, and then remove all android related changes. That way I have a new iOS-only PR. Let's just say the feature comes with the sound support in iOS. Android can come later.

ngocdh:
Closing this as I've created #1875 for iOS, and a clean commit history.
Will create other PRs for Android.

ann0see:
unrelated: @hoffie I just ran your changelog-helper script and it listed this PR as being added to the next release?

ann0see:
I'd just add CHANGELOG: SKIP here

hoffie:
> unrelated: @hoffie I just ran your changelog-helper script and it listed this PR as being added to the next release?

Yes, as it was referenced in fac5d9d1e2a1321fdd53fda350a71f1d91f66b8c (which will be part of 3.9.0).
Meta: Maybe we should work on this in a new Issue if that's considered a generic problem.

ann0see:
Yes. It’s at least counter intuitive to have a closed PR in the changelog. In on the other hand I think feature branch merges are not in the changelog.

PR #1867, opened 2021-06-13
Persistent server lists using --directoryfile

pljones:
# Short description of changes

This change makes the server lists supplied by directory servers more reliable.

# Context: Fixes an issue?

A number of users have complained server lists are not available reliably.  When a directory server is restarted, it could take up to 15 minutes for all the servers registered with it to re-register.  With this change in place, the directory should be fully populated on start up.

# Does this change need documentation? What needs to be documented and how?

The help output has been added.

The Server Administrators' Guide will need updating to cover this option, when it's needed and how to use it.

It relies on the server administrator specifying a file in which to store the list across restarts.

# Status of this pull request

Replacement for https://github.com/jamulussoftware/jamulus/pull/1799 and https://github.com/pljones/jamulus/pull/9.  Makes #1803 and #1804 redundant.  All of those now closed.

This is running on all my directory servers (i.e. the genre-specific ones, plus my additional ones).

I'll also be using the Windows Client code (just to confirm no adverse side effects) from 19th June onwards.

## What is missing until this pull request can be merged?
Review of the latest version of the code.

## Checklist
- [x] I verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)


src/serverlist.cpp
```diff
@@ -93,43 +155,59 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
...
-        if ( !slServInfoSeparateParams[2].isEmpty() && ( iCountry >= 0 ) && ( iCountry <= QLocale::LastCountry ) )
+        bool      ok;
+        const int iCountry = slServInfoSeparateParams[2].toInt ( &ok );
+        if ( ok && iCountry >= 0 && iCountry <= QLocale::LastCountry )
         {
             ThisServerListEntry.eCountry = static_cast<QLocale::Country> ( iCountry );
         }
     }
 
+    // per definition, the very first entry is this server and this entry will
```
ann0see, line 166: ... is this server (maybe better: this directory server?) Why is this comment here, not below ServerList.clear()? Without having looked at what ServerList.clear() does, I assume it clears the server list and ServerList.append ( ThisServerListEntry ) adds the directory server entry?

src/serverlist.cpp
```diff
@@ -172,8 +252,9 @@ void CServerListManager::SetCentralServerAddress ( const QString sNCentServAddr
 
     // per definition: If we are in server mode and the directory server address
     // is the localhost address, and set to Custom, we are in directory server mode.
-    bIsCentralServer = ( ( !strCentralServerAddress.toLower().compare ( "localhost" ) || !strCentralServerAddress.compare ( "127.0.0.1" ) ) &&
-                         ( eCentralServerAddressType == AT_CUSTOM ) );
+    bIsCentralServer =
+        ( ( !strCentralServerAddress.compare ( "localhost", Qt::CaseInsensitive ) || !strCentralServerAddress.compare ( "127.0.0.1" ) ) &&
+          ( eCentralServerAddressType == AT_CUSTOM ) );
```
ann0see, line 257: Shouldn't this be added as a method somewhere? I mean, it's also repeated in line 275

src/serverlist.cpp
```diff
@@ -490,6 +542,107 @@ void CServerListManager::CentralServerQueryServerList ( const CHostAddress& Inet
...
+    {
+        return;
+    }
+
+    QFile file ( ServerListFileName );
+
+    if ( !file.open ( QIODevice::WriteOnly | QIODevice::Truncate | QIODevice::Text ) )
+    {
+        // Not a useable file
+        qWarning() << QString ( tr ( "Could not write to '%1'" ) ).arg ( ServerListFileName );
```
ann0see, line 627: Suggestion as above: ... Please check that Jamulus has permission (and that there is free space).

src/serverlist.cpp
```diff
@@ -490,6 +542,107 @@ void CServerListManager::CentralServerQueryServerList ( const CHostAddress& Inet
...
+}
+
+void CServerListManager::CentralServerLoadServerList ( const QString strServerList )
+{
+    QFile file ( strServerList );
+
+    if ( !file.open ( QIODevice::ReadWrite | QIODevice::Text ) )
+    {
+        qWarning() << qUtf8Printable (
+            QString ( "Could not open '%1' for read/write.  Please check that Jamulus has permission (and that there is free space)." )
```
ann0see, line 567: Isn't there one space too much?

src/main.cpp
```diff
@@ -732,59 +753,62 @@ int main ( int argc, char** argv )
...
-           "      --clientname      client name (window title and JACK client name)\n"
            "\n"
-           "Example: " + QString ( argv[0] ) + " -s --inifile myinifile.ini\n"
+           "Example: %1 -s --inifile myinifile.ini\n"
            "\n"
            "For more information and localized help see:\n"
-           "\n"
-           "https://jamulus.io/wiki/Command-Line-Options\n";
+           "https://jamulus.io/wiki/Command-Line-Options\n"
+    ).arg(argv[0]);
```
ann0see, line 828: Styling?

src/main.cpp
```diff
@@ -732,59 +753,62 @@ int main ( int argc, char** argv )
...
            "  -M, --mutestream      starts the application in muted state\n"
            "      --mutemyown       mute me in my personal mix (headless only)\n"
-           "  -c, --connect         connect to given server address on startup\n"
-           "  -j, --nojackconnect   disable auto JACK connections\n"
+           "      --clientname      client name (window title and jack client name)\n"
            "      --ctrlmidich      MIDI controller channel to listen\n"
-           "      --clientname      client name (window title and JACK client name)\n"
            "\n"
-           "Example: " + QString ( argv[0] ) + " -s --inifile myinifile.ini\n"
+           "Example: %1 -s --inifile myinifile.ini\n"
```
ann0see, line 824: Yeah. Much better ;-).

src/main.cpp
```diff
@@ -498,23 +509,33 @@ int main ( int argc, char** argv )
...
+        {
+            qWarning() << "Server Bind IP will only take effect when running a server.";
+        }
+        else
         {
-            qWarning() << "Server Bind IP is invalid. Only plain IP addresses are supported.";
+            QHostAddress InetAddr;
+            if ( !InetAddr.setAddress ( strServerBindIP ) )
+            {
+                qWarning() << "Server Bind IP is invalid. Only plain IP addresses are supported.";
```
ann0see, line 552: Only plain IP-addresses are supported. --> Please check that the IP you entered is valid.

src/main.cpp
```diff
@@ -498,23 +509,33 @@ int main ( int argc, char** argv )
...
         }
     }
 
     if ( !strServerBindIP.isEmpty() )
     {
-        QHostAddress InetAddr;
-        if ( !InetAddr.setAddress ( strServerBindIP ) )
+        if ( bIsClient )
+        {
+            qWarning() << "Server Bind IP will only take effect when running a server.";
```
ann0see, line 545: ... running a server. You're running a client.

src/main.cpp
```diff
@@ -498,23 +509,33 @@ int main ( int argc, char** argv )
...
+            qWarning() << "Server Public IP will only take effect when registering a server with a directory server.";
         }
-        if ( strCentralServer.isEmpty() || bIsClient )
+        else
         {
-            qWarning() << "Server Public IP will only take effect when registering a server with a directory server.";
+            QHostAddress InetAddr;
+            if ( !InetAddr.setAddress ( strServerPublicIP ) )
+            {
+                qWarning() << "Server Public IP is invalid. Only plain IP addresses are supported.";
```
ann0see, line 536: See below.

src/server.cpp
```diff
@@ -239,7 +240,14 @@ CServer::CServer ( const int          iNewMaxNumChan,
...
     HighPrecisionTimer ( bNUseDoubleSystemFrameSize ),
-    ServerListManager ( iPortNumber, strCentralServer, strServerInfo, strServerPublicIP, strServerListFilter, iNewMaxNumChan, &ConnLessProtocol ),
+    ServerListManager ( iPortNumber,
+                        strCentralServer,
+                        strServerListFileName,
+                        strServerInfo,
+                        strServerPublicIP,
+                        strServerListFilter,
+                        iNewMaxNumChan,
+                        &ConnLessProtocol ),
```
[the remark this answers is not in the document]
pljones: Green build, so yes.

src/serverlist.cpp
```diff
@@ -25,8 +25,75 @@
...
+
+    sl.append ( this->HostAddr.toString() );
+    sl.append ( this->LHostAddr.toString() );
+    sl.append ( ToBase64 ( this->strName ) );
+    sl.append ( ToBase64 ( this->strCity ) );
+    sl.append ( QString::number ( this->eCountry ) );
+    sl.append ( QString::number ( this->iMaxNumClients ) );
+    sl.append ( QString::number ( this->bPermanentOnline ) );
+
+    return sl.join ( ";" );
```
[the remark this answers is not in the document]
pljones: I can't see it does.  This is columnular, multi-record data.  It doesn't lend itself to XML.  It's not a structured document.

src/serverlist.cpp
```diff
@@ -25,8 +25,75 @@
...
+                                                iNumClients,
+                                                isPermanent ) );
+}
+
+QString CServerListEntry::toCSV()
+{
+    QStringList sl;
+
+    sl.append ( this->HostAddr.toString() );
+    sl.append ( this->LHostAddr.toString() );
```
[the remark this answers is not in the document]
pljones: `HostAddr` and `LHostAddr` are structured data, not strings.  The `.toString()` on the class renders to `IP:port`.

src/serverlist.cpp
```diff
@@ -69,20 +136,15 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
...
-
     // Init server list entry (server info for this server) with defaults. Per
     // definition the client substitutes the IP address of the directory server
-    // itself for his server list. If we are the directory server, we assume that
-    // we have a permanent server.
+    // itself for his server list. If we are a directory server, we assume that
+    // we are a permanent server.
     CServerListEntry
-        ThisServerListEntry ( CHostAddress(), SlaveCurLocalHostAddress, "", QLocale::system().country(), "", iNumChannels, GetIsCentralServer() );
+        ThisServerListEntry ( CHostAddress(), SlaveCurLocalHostAddress, "", QLocale::system().country(), "", iNumChannels, bIsCentralServer );
```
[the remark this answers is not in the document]
pljones: Yep.  There's no point calling a method to get a property of the class the getter method is exposing to other classes.  We're not external to ourselves - it introduces an extra method call unnecessarily.

src/serverlist.cpp
pljones: Yes.

src/serverlist.cpp
pljones: No, too annoying

src/serverlist.cpp
pljones: I'm old.  Full stops get two spaces.  I can remove the annoying suggestion as well as the spaces.

src/main.cpp
pljones: ?

src/main.cpp
pljones: That removes information.

src/main.cpp
pljones: ?

src/main.cpp
pljones: See above

src/server.cpp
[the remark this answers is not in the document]
ann0see: I think I remember that this was changed by running clang-format in other places? But not part of this PR then.

src/main.cpp
ann0see: Shouldn’t this be .arg ( argv[0] ); with spaces?

src/main.cpp
ann0see: We should give the user more information and not let him think too much. Therefore, we should state that this option is not supported since he‘s running a client. Maybe he forgot adding `-s` or something else. See https://github.com/jamulussoftware/jamulus/discussions/1568 and https://uxplanet.org/how-to-write-good-error-messages-858e4551cd4

src/serverlist.cpp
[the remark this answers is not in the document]
pljones: They're not related in any way.

src/main.cpp
pljones: Yes, it should - and clang-format should have rejected it.  Hm....

src/main.cpp
pljones: It's not part of this change, though.  I'm just moving stuff around.

pljones: The message tells the user what the problem is.  I refuse to write messages that take a rough guess as to why the user has caused themselves the problem if there's any chance it could be wrong.  That the program isn't running as a server is true.

They might have remove processing from the code to handle "-s" for all I know and still passed "-s".  No point me trying to guess, is there?

src/main.cpp
pljones: Ah, it's the line before the "clang-format on"..!  Oops.

src/server.cpp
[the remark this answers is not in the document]
pljones: I think I may have just exceeded it's "all on one line" rule with the extra parameter.

pljones:
It must be something on this branch. #1867 is working, for example.

ann0see:
Todo: 

- [ ] Some more tests
- [ ] In deph review (I don't feel comfortable doing that; however, I trust you that it should all be ok)

pljones:
> 
> 
> Todo:
> 
>     * [ ]  Some more tests

Agreed - I've run this the way I expect it to work.  I've not tried to break it at all.

>     * [ ]  In deph review (I don't feel comfortable doing that; however, I trust you that it should all be ok)

Also agreed.  I know @softins has looked at the directory server code; it also looks like @passing and @hoffie have touched it:
```
peter@fs-peter:~/git/Jamulus-wip$ git blame src/serverlist.cpp | grep -v '(Volker Fischer   \|(Peter L' | sed -e 's/^[^(]*(//' -e 's/  *20.*) .*$//' | sort -u
```

passing:
With https://github.com/jamulussoftware/jamulus/commit/2bd16542a5ac3d08ad318a124781bab1195a65b6 I have touched all the code :smile:

pljones:
> 
> 
> With [2bd1654](https://github.com/jamulussoftware/jamulus/commit/2bd16542a5ac3d08ad318a124781bab1195a65b6) I have touched all the code 😄

Ah yes, true...

ann0see:
`Jamulus -h` gives back `--directoryfile` not --directorylist --> change cli option or pull request title depending on which name we want to use. 

No error message is given if we run a non directory server.

If we don't give a file name but a directory, the server does not exit and a warning message is not quite obvious. Shouldn't the "Could not open ..." etc. messages be a bit more prominent (e.g. by appending "WARNING:")

Only after gracefully stopping a directory the file gets written which means that we don't handle a crash. Is this intended?
 
All in all I think it's on a good way to get merged.

pljones:
Thanks for the negative tests!

> No error message is given if we run a non directory server.

Well, a warning message - it'll carry on starting up at that point.

> If we don't give a file name but a directory, the server does not exit and a warning message is not quite obvious. Shouldn't the "Could not open ..." etc. messages be a bit more prominent (e.g. by appending "WARNING:")

Mm, yes... I need to say "either it doesn't exist or it's a file".  I could do that in `src/main.cpp`, too.  That still doesn't give reliable results -- you could pass a name that doesn't exist that can't be created, as you point out: it's still got to be something that can be created.

Overall, I'm tempted to going back to requiring an _existing, readable, writeable file_ to be passed: no sudden surprises that way.

pljones:
Hm.  Made worse by the server ini file.  I can use a server ini file to set my server list to "custom" "localhost" and run as a directory server.

On start up, as I don't supply `-serverlist`, validation can't check -- because the ini file hasn't been loaded -- whether or not a combination of options is actually valid.

Like I've said elsewhere, we need to refactor start up further -- parse the command line options, load the ini file (either default or specified) and then override ini file from command line.  Currently, it's the other way around (ini file overrides command line), which feels totally wrong.  Once all of that's been done, then validate the settings work together -- you (should) know what got to the ini file was valid, so it's only going to be command line options causing the problem.  As it stands, I can't know whether I'm running as a directory server when parsing the command line options.  Or even if I'm registering.

pljones:
Well, after several attempts, I've come up with something that does the best it can.

If we're a headless server, there _is no inifile_ (regardless of passing the argument), so the command line specifies everything there is to know.  Otherwise, certain checks cannot be made as the inifile might change things.

I might have missed some bits, though.

src/main.cpp
pljones: It's meant to imply you entered a hostname, rather than an IP address.  I don't think it cares much whether the IP is _valid_, either.

pljones:
And, of course, whilst I was tidying up, I did some more changes...

src/main.cpp
```diff
@@ -490,53 +546,179 @@ int main ( int argc, char** argv )
...
+                    }
+                    else if ( !serverListFileInfo.isReadable() || !serverListFileInfo.isWritable() )
+                    {
+                        qWarning() << qUtf8Printable (
+                            QString ( "Server list file %1 must be readable and writeable.  Please check the permissions." )
+                                .arg ( strServerListFileName ) );
+                        strServerListFileName = "";
+                    }
+                }
+
```
ann0see, line 672: Hmm. It's a lot of new code in main.cpp then.

pljones: Yep - you suggested that `--directoryfile` should only be allowed when running as a directory server.  That meant lots of new code.  As I was doing it for the one option, it felt necessary to be consistent, so I did it for as many I as I could.

ann0see: Yes. I know. Probably it's a good move then anyway. So the next steps would be another few tests and another person who reviews this.

pljones: Currently CClientSettings and CServerSettings need, respectively, to be passed an instance of CClient and CServer.  Which means you have to create the Client or Server _before loading the inifile_.  The class in question configures itself according to its defaults or command line option overrides.  Then, if there's an inifile (i.e. not headless server), a whole load of other options get applied, changing how the class is configured.

This is why you _cannot_ know whether a server is a directory server if it's not headless when
- validating the command line options
- when creating the server instance

It's not until during inifile loading that you know.  That's after the directoryfile parameter has been set in the server.  Too late now!

All of this is confusing -- it leaves me totally confused every time I have to think about it, anyway.

ann0see:
Compiling on Linux gives a warning: 

```
In file included from src/serverlist.cpp:25:
src/serverlist.h: In constructor ‘CServerListManager::CServerListManager(quint16, const QString&, const QString&, const QString&, const QString&, const QString&, int, CProtocol*)’:
src/serverlist.h:213:16: warning: ‘CServerListManager::pConnLessProtocol’ will be initialized after [-Wreorder]
  213 |     CProtocol* pConnLessProtocol;
      |                ^~~~~~~~~~~~~~~~~
src/serverlist.h:203:19: warning:   ‘ESvrRegStatus CServerListManager::eSvrRegStatus’ [-Wreorder]
  203 |     ESvrRegStatus eSvrRegStatus;
      |                   ^~~~~~~~~~~~~
src/serverlist.cpp:94:1: warning:   when initialized here [-Wreorder]
   94 | CServerListManager::CServerListManager ( const quint16  iNPortNum,
      | ^~~~~~~~~~~~~~~~~~
```

Probably you'll need to move methods around.

src/main.cpp
ann0see: Thanks for the explanation. 

This seems to explain why my tests with the GUI didn't do what I thought they would do: I set up a GUI directory via CLI which resulted in a message saying it is a directory, but I couldn't register with it or have the register my server box checked. 

There must be a reason why Volker did it like it is at the moment?

pljones: I think it was very much "that's the way it's always worked, why change it?".  Which is valid if you're trying to reduce the amount of change being made to something you have time to manage, which he was.

ann0see:
Just re-ran the auto build again since Android seemed to be broken. I think this PR would be ready then, however, I'd like to wait for others to test it too.

pljones:
> Compiling on Linux gives a warning:

Good spot - CodeQL should have objected more vigorously to the new warning, too.

ann0see:
Missing: @ann0see test again.

ann0see (approved):
Works as expected; however the cli handling is - as you mentioned - not perfect. But that's not part of this PR.

pljones:
OK, I've just split #1896 out from this - this still has that commit in, as I'm depending on it locally.

softins (commented):
This appears to include all the changes that are separately in #1896. I've approved the latter; maybe that needs to be merged first and this PR then rebased again?

ann0see:
No. I think if this PR is merged, the other one will close automatically?

ann0see:
@softins this PR is about something else (persistent server lists for directories)?

softins:
> @softins this PR is about something else (persistent server lists for directories)?

Yes, but it had been branched off the commit for #1896 rather than directly off master. Now that #1896 has been merged, this PR only contains code relevant to the intended improvement.

On first read through it looks good. I'll try it out later and approve after that.

softins (approved):
Just tested this out on my Pi and it appears to work fine. Approving. Thank you Peter!

PR #1873, opened 2021-06-13
Enable Mnemonics Keys for MacOS

### LINKED ISSUE #1726: Mac: Enable accelerator key functionality
Just found this in the [Qt docs](https://doc.qt.io/qt-5/qshortcut.html#details):

> On Mac, shortcuts are disabled by default. Call qt_set_sequence_auto_mnemonic() to enable them. However, because mnemonic shortcuts do not fit in with Aqua's guidelines, Qt will not show the shortcut character underlined.

We should probably consider activating them on Mac, too?
cc @softins @ann0see

_Originally posted by @hoffie in https://github.com/jamulussoftware/jamulus/discussions/1693#discussioncomment-755670_

henkdegroot:
Fixes #1726

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -39,12 +39,19 @@
 #endif
 #if defined( Q_OS_MACX )
 #    include "mac/activity.h"
+extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
 #endif
 
 // Implementation **************************************************************
 
 int main ( int argc, char** argv )
 {
+
+#if defined( Q_OS_MACX )
+    // Mnemonic keys are default disabled in Qt for MacOS. The following function enables them.
+    // Qt will not show these with underline characters in the GUI on MacOS.
+    qt_set_sequence_auto_mnemonic ( true );
+#endif
 
     QString        strArgument;
     double         rDbleArgument;


src/main.cpp
```diff
@@ -39,13 +39,20 @@
...
 #endif
 
 // Implementation **************************************************************
 
 int main ( int argc, char** argv )
 {
 
+    // mnemonic keys are disabled by default on MacOS the following function enables them
+    // by "mac" design they will not be shown with underline characters in the GUI
+#if defined( Q_OS_MACX )
```
pljones, line 52: Could you move the comment inside the `#if`.

Could you also start new sentences in appropriate places to make the comment easier to read.

henkdegroot: Thanks for the comments, I have made the updates. Hope the text reads better now.

pljones: You've still not split it into sentences.  There are three separate ones but no full stops.

henkdegroot: For me there were just two sentences. I have updated the text.

pljones: > mnemonic keys are disabled by default on MacOS the following function enables them

This reads in multiple ways and hence needs punctuation

> mnemonic keys are disabled by default
> on MacOS the following function enables them

> mnemonic keys are disabled by default on MacOS
> the following function enables them

Also, as the line is fairly long, it's not clear whether the second line is part of the same sentance as the first due to the lack of punctuation.  Overall, it hampers reading and understanding.

henkdegroot: @pljones I have revised the text now. Are you checking the latest?

pljones: Apparently not...  Sorry!

pljones (approved):

ann0see (approved):
Tested client on Big Sur. Works as expected. --> merge

ann0see:
CHANGELOG: Keyboard shortcuts now also work on macOS

ann0see:
@hoffie I think this PR was wrongly added to the nightly changelog...

hoffie:
Probably due to 0764536fab6b84c69fafb0b4a7e865cd088de346. I'll open a separate issue to avoid that.

PR #1856, opened 2021-06-10
Add macOS signing changes 

emlynmac:
In order to give Mac users a better install experience, I've modified the build script to have an optional signing step, using:
`./mac/deploy_mac.sh -s <cert_name>`

https://github.com/jamulussoftware/jamulus/discussions/1851 for more details.

For now, I'll build the signed version manually otherwise we need to figure out a _secure_ means to store the certificate / private key that I use to sign it.

PR #1874, opened 2021-06-13
Add Apple AppStore license note

ann0see:
See also: https://github.com/nextcloud/ios/blob/master/COPYING.iOS
#1856 

Jamulus is licensed under the GPL. Currently there is ongoing work on getting Jamulus on the app store. However, there might be incompatibilities between the Apple TOS and the GPL. 

This text should be modified and added to the main repo to ensure that we will not sue anyone (especially, but not only @emlynmac) who publishes the code. 

I'd be more than happy if any contributor who doesn't want his code to be published on the app store remove his code. 

Tagging
@jamulussoftware/maindevelopers here.

### DIFF
--- a/APPLEAPPSTORE.LICENSE.WAIVER
+++ b/APPLEAPPSTORE.LICENSE.WAIVER
@@ -1,0 +1,19 @@
+The inclusion of Jamulus in the App Store is the only practical way of allowing its use on Apple mobile devices. Although Apple will assert the GPL as the "Custom EULA” for Jamulus, their T&Cs impose some extra conditions. This act of addition when distributing the software is not allowed under the GPL.
+
+However, Apple's terms apply to those who obtain Jamulus from the App Store. It does not affect the copyright holders' assertion of the GPL, or otherwise affect the Jamulus project:
+
+https://www.apple.com/legal/internet-services/itunes/us/terms.html
+
+Moreover, there are no provisions in the Apple terms and conditions that we identify as being unacceptable to the copyright holders, or to the Jamulus project overall. Anyone using App Store services would already be accepting Apple’s terms for their use of any service they are obtaining. This includes being unable to re-distribute the software.
+
+The copyright holders do not want a legal conflict to prevent the distribution of Jamulus on the Apple App Store. Therefore, and in consideration of the above points, we the copyright holders in Jamulus hereby waive any objections to the fact that Apple would be violating the GPL in making Jamulus available on the App Store.
+
+Apple reserves the right to change its terms. If at any time they do that in a way that we find unacceptable, we will remove Jamulus from the App Store.
+
+This waiver applies solely to the specific case of obtaining Jamulus from the Apple App Store. All other distribution methods must fully comply with the terms of the GPL.
+
+Dissent
+
+Any Jamulus contributor who does not wish their code to be published via the App Store must ensure their code is not so published. This can be achieved for example by removing their code from the Jamulus source, or by using technical measures (e.g. #ifdef) to ensure the code in question will not be compiled and included for App Store releases.
+
+[ends]


LICENSE.APPLEAPPSTORE
```diff
@@ -0,0 +1,10 @@
+Jamulus contributors and maintainers are aware that the GPL and Apple App Store terms
+of service might conflict. Specific rights the GPL guarantees might not be possible 
+to be granted by AppStore distribution. Copyright holders of the source code do not 
+want this legal conflict to prevent the distribution of Jamulus on the Apple App Store.
+
+Every contributor to the Jamulus source code and copyright holder shall not sue license 
+violations soely based on conflicts between the GPL and the App Store guidelines which prevent 
+AppStore releases. Every  contributor or copyright holder who doesn't wish his code to be published
```
pljones, line 8: Spacing.

pljones:
I have a bad feeling about this...

It looks to me that that statement effectively violates and voids the GPL2 licence that has been used up until now.

ann0see:
That’s the problem we’re trying to solve: how can we allow Jamulus on the Apple App Store but still ensure the GPL rights (in practical terms). It’s always possible to build the macOS version from source and have all the rights from this version.

emlynmac:
Is there a licence that will work for both publishing to App Store and for keeping open source changes back?
I see we don't actually include a licence file in the repository at the moment - we probably should?

For discussion of this, I found:
https://stackoverflow.com/questions/459833/which-open-source-licenses-are-compatible-with-the-apples-iphone-and-its-offici 

The Apache 2 licence seems to be the recommendation there, but I wasn't around at the inception of jamulus and why it was GPLv2 rather than Apache back then. 

If we cannot have a licence that will work with the App Store, then we cannot publish something for iOS.

hoffie:
> Is there a licence that will work for both publishing to App Store and for keeping open source changes back?

I don't think we can change the license at will at all. This would likely need an OK by **all** contributors which might be close to impossible.
Regarding the direction of (possible) license changes, there were different discussions with the intent of protecting Jamulus as an open source project even more (e.g. AGPLv3, which is more restrictive than GPL), and not lifting restrictions.

I understand that having a less restrictive license would be beneficial for the app store topic, but I doubt it is easily possible or even desirable for the other reasons.

I'd hope for something like the Nextcloud to work, although I don't quite understand all legal consequences. Maybe @gilgongo's friendly lawyer can help? :)

> I see we don't actually include a licence file in the repository at the moment - we probably should?

We do -- the typical name (at least when using the GPL?) is [COPYING](https://github.com/jamulussoftware/jamulus/blob/master/COPYING) though.

emlynmac:
> We do -- the typical name (at least when using the GPL?) is [COPYING](https://github.com/jamulussoftware/jamulus/blob/master/COPYING) though.

Ah ha. missed that... Thanks. I'm used to seeing that in a LICENSE file

emlynmac:
> I don't think we can change the license at will at all. This would likely need an OK by **all** contributors which might be close to impossible.
> Regarding the direction of (possible) license changes, there were different discussions with the intent of protecting Jamulus as an open source project even more (e.g. AGPLv3, which is more restrictive than GPL), and not lifting restrictions.

So then the question becomes, are we all comfortable in putting in a statement akin to what https://github.com/nextcloud/ios have done? Specifically, their wording is: [here](https://github.com/nextcloud/ios/blob/master/COPYING.iOS)

This should be for both macOS and iOS app stores, so maybe the file would be a `.apple_app_store` suffix?

ann0see:
> Specifically, their wording is: here

I know, but I didn’t want to copy it. The text above is at least partly based on the nextcloud file

emlynmac:
> > Specifically, their wording is: here
> 
> I know, but I didn’t want to copy it. The text above is at least partly based on the nextcloud file

I see a lot more words on theirs - something I'm told legal folks like ;D 
Do we have any lawyers that could look at it?

emlynmac:
So I ran into another GPL software on the Mac app store today:
https://macos.telegram.org, which calls out their GPLv2 licence and is kept [here](https://github.com/overtake/TelegramSwift).

gilgongo:
I had a quick chat with my British Library licence manager friend today. He's not a FOSS  expert because the library is mainly concerned with artistic works, but knows a lot about CC and commercial licences and how they work in the "real world". If we did a Nextcloud thing (publishing a disclaimer, essentially) then this shows an intention which a court would almost certainly take into account in the massively unlikely situation anyone tried to make a legal fuss.

emlynmac:
So from my perspective, I’m happy to include these additional terms and publish to the App Store under my corporation, understanding the very small risk here. The most likely outcome in the event of a conflict would be the removal of the app from the store. 
If we want to have an iOS version working, then we have to release via the store

gilgongo:
@emlynmac There's some confusion about where and how to display licencing info in github repos as it's an issue that's come up before. We don't have a LICENCE.md but do have what I think github prefers which is as link from README.md to COPYING.md: https://github.com/jamulussoftware/jamulus/blob/master/COPYING

Our licence is also displayed on the github "code" page.

emlynmac:
So there is one other potential option for iOS/macOS store compliance etc. 
Creating a brand new Apple only client app using native swift and making it compatible with the jamulus server API. 
This would then be independent from the original code and not interfering with the existing project

ann0see:
Yes, sure. But that would involve a lot of development and the need to maintain two compatible clients.

emlynmac:
> Yes, sure. But that would involve a lot of development and the need to maintain two compatible clients.

There will be fresh code but arguably simpler because it only needs to work for two very closely compatible targets using the same interfaces. I'm kinda curious to see how much work at this point. Maybe I'll have a look...

pljones:
Remember, anything that's compatible with protocol.h/protocol.cpp will, arguably, be derived from them...

emlynmac:
> Remember, anything that's compatible with protocol.h/protocol.cpp will, arguably, be derived from them...

I think there's a definite line between copying the contents of the file vs defining against the interface to the server.

pljones:
The only "published" definition, however, of that interface is subject to the GPL 2.0 terms.

emlynmac:
> The only "published" definition, however, of that interface is subject to the GPL 2.0 terms.

And if you were to really be thorough, then you could reverse-engineer the protocol from packet sniffing. At this point, it's a rather academic conversation which is detracting from the real issue of whether or not the existing codebase can get to the App Store or not.

If there is no comfort with publishing to the App Store due to licensing restrictions, then there really is no point in doing an iOS version as it will never be available. The Mac version can at least be signed, but again not in the store.

ann0see:
> If there is no comfort with publishing to the App Store due to licensing restrictions, then there really is no point in doing an iOS version as it will never be available.

At least not easily available. However, sideloading or TestFlight Publishing might be possible?

emlynmac:
> > If there is no comfort with publishing to the App Store due to licensing restrictions, then there really is no point in doing an iOS version as it will never be available.
> 
> At least not easily available. However, sideloading or TestFlight Publishing might be possible?

Hmm, side loading = jailbreaking. So likely no.
Testflight = somebody being ok with their apple developer account being used for that and having builds that expire after 90 days. Neither of these are workable IMHO.

I am beginning to think that it would be prudent to move the main network protocol definition to a separate repo under MIT licence and then have the jamulus repo pull that in for the current software.
Then for the apple-world stuff, we can build something specific for those and leverage the App Store and performance increase from not having QT in the binary.

I can create signed macOS builds for the jamulus main, but we can have something totally isolated in terms of licence and code for the app stores. This would likely be Apache 2 or MIT licensed.

Pros
* minimal existing code is made MIT (server interface), 
* apple users get an efficient, signed client,
* easier to get existing apple developers helping out
* App Store releases
* Simplification of existing codebase - no iOS parts
* Server API changes clearer to see as committed to a separate repository

Cons
- Two code bases instead of one
- Need to MIT licence the message API

ann0see:
Sideloading != jailbreaking but close. Semi advanced users can do that e.g. via AltStore. 

Pro: Shared code-base Con: not a long term solution. 

> I am beginning to think that it would be prudent to move the main network protocol definition to a separate repo under MIT licence

The question here is if that’s allowed. 
***
As far as I know, you’re one of the rare (or only?) Apple developers in the Jamulus community.

emlynmac:
> Sideloading != jailbreaking but close. Semi advanced users can do that e.g. via AltStore.
Sorry, there's no way to side load on iOS without jailbreaking.

emlynmac:
>As far as I know, you’re one of the rare (or only?) Apple developers in the Jamulus community.

I use Typescript, Dart, C, C++, Java, Swift on a fairly regular basis. Given the headaches the team are starting to see with making iOS work, I'm proposing what I think is the best option.

emlynmac:
>>I am beginning to think that it would be prudent to move the main network protocol definition to a separate repo under MIT licence

>The question here is if that’s allowed.

Well, here's the interesting part. You can totally reverse-engineer the protocol as it is un-encrypted - meaning it would be possible to build a completely independent client and/or server. That would be much _faster_ done with shared headers for the protocol definition and better for all involved really - more exposure, more growth etc.

ann0see:
Thanks for the clarification. 

So we have two options:

1. Separate app for macOS and iOS without Qt (I know that moving away from Qt was a discussion. Especially with the latest bugs on macOS, it might be a way to go)
2. Use the existing code and include the statement from above. I think the risk - if it even exists - is not too high and would only result in a removal of the app. So this would be the way I‘d suggest for now.

emlynmac:
Is it time to bring this PR out of WiP?

gene96817:
A few questions....

1- re: "no comfort in publishing to the Apple store"... would the alternatives be more comfortable? 

2- How will the iOS version improve our community? I cringe at the thought it will create more support issues. Is this iOS for smartphones or for tablets(iPads)?

3- What will be the differences in user experience between desktop and iOS? 
     I am not interested in supporting a user that expects a simpler interface with the same musical experience as users with the desktop. (I guess this question assumes it is for smartphones and much smaller screens.)

emlynmac:
> 1- re: "no comfort in publishing to the Apple store"... would the alternatives be more comfortable?

Not sure what you're getting at here. Can you elude what you consider the alternatives to be?

>2- How will the iOS version improve our community? I cringe at the thought it will create more support issues. Is this iOS for smartphones or for tablets(iPads)?

If you support more platforms, you get more issues as more users use it. If there's a strong concern here around android / iOS users being more vocal, then maybe mobile support should _not_ be part of the main code base.

>3- What will be the differences in user experience between desktop and iOS? I am not interested in supporting a user that expects a simpler interface with the same musical experience as users with the desktop. (I guess this question assumes it is for smartphones and much smaller screens.)

This sounds like you have no interest in a mobile development - and if this is something that divides the jamulus contributors, then it does make sense to drop mobile support and introduce mobile clients for those interested in pursuing such.

gene96817:
sorry to sound so negative. 
regarding 1: I didn't see any alternatives proposed that would simplify (or reassure) the concerns of publishing an iOS app.

From my personal experience, when a chorus decides to use Jamulus, there are always a few members that don't want to learn Jamulus and expect a magic solution with no learning curve. I see the desire for a smartphone app to be part of the desire for magic. 

I am concerned about having members that will insist on participating while mobile (actually, while on the run or driving). This attitude is disruptive to video meetings. I expect this use case to be equally disruptive to a rehearsal.

I am very interested in iOS as a solution for a tablet. Then there is enough screen space to deliver a similar experience to a desktop. I have singers that use a tablet for all their computing needs. (In this case, I expect them to use an ethernet connection.)

gilgongo:
@gene96817 I think that while your concerns are legitimate, I see putting Jamlus in the app store as more "strategic" than anything else. Having it there would raise our profile, both for mobile and other platforms, I hope. It might also stimulate efforts to re-architect Jamulus in ways that would benefit all platforms (eg possibly moving away from Qt). Whether is actually works on mobile would be a secondary consideration in my (I hope not too cynical!) opinion.

ann0see:
So the missing points here are:

Ok note by Volker and Peter (Volker as original developer and Peter who is not comfortable: https://github.com/jamulussoftware/jamulus/pull/1874#issuecomment-860270841). Softins, hoffie, gilgongo and me are ok with it. 

I'd see the following as an argument why we can still follow the spirit of FLOSS (personal opinion):

1. The source is still available. Everybody who wants to have the source can still get it and compile Jamulus for macOS and iOS himself. Nobody should and can stop him/her.
2. We do release binaries besides the app store. Nobody forces anyone to use either of them. 
3. @emlynmac should make every source change available, but we see from his code signing PR that he does contribute back. So I don't see an issue there.
3. Apple does add specific restrictions which can be overcome easily: nobody forces you to use the AppStore app. Even on iOS you can bypass the app store in certain ways.
4. AppStore distribution makes this software more accessible (which is in fact one of the core values of the "hacker ethic" which the floss movement follows). Inexperienced users have an easy update mechanism and don't need to worry about signing issues.
5. It's a rather theoretical issue and we can remove the app if we have complaints.

pljones:
> Having it there would raise our profile

That has absolutely no attraction for me.

gilgongo:
@pljones 

> That has absolutely no attraction for me.

That may be because you already know about Jamulus :-)

pljones:
> That may be because you already know about Jamulus :-)

More that anything related to Apple tends to make me uncomfortable.

emlynmac:
> > That may be because you already know about Jamulus :-)
> 
> More that anything related to Apple tends to make me uncomfortable.

Maybe you could elaborate a bit on why that is the case; are the specific things you feel uncomfortable about?

pljones:
Let's just say every time I've been near any Apple product, I've left it wishing I'd never bothered.

emlynmac:
> Let's just say every time I've been near any Apple product, I've left it wishing I'd never bothered.

So it's a personal bias then, rather than something we can quantify?

pljones:
As I said.

gene96817:
There is one strongly positive aspect of Apple products.... no ASIO drivers.... the one issue that makes it hard to get new users working on Jamulus.

pljones:
That applies to devices like Raspberry Pis and other Linux-based systems, of course.  And they're easier to build for and support than Apple.

ann0see:
Yes, but there are still people who use Apple products ;-). I can understand your dislike, however it also exists the other way round (dislike of Windows or Linux (I know a bunch of persons who dislike Linux)). So how to move forward?

gene96817:
There will always be partisans for Windows or Apple or Linux.
It is interesting to see a lot of the younger IT and cyber folks loving Macs because they have a Linux (-like) system running under the MacOS. Microsoft is planning their next major Windows release to use Linux under the hood. The key question is will they change their driver layer enough to get rid of AISO.

Anyway, I propose that we publish Jamulus for iOS in the Apple store. For developers that dislike Apple, please hold your nose and let us proceed. The visibility will be good for Jamulus. And there are iPad users waiting for a way to use Jamulus. Most of all, the work has already been done. Let's not waste that good work.

gene96817:
> That applies to devices like Raspberry Pis and other Linux-based systems, of course. And they're easier to build for and support than Apple.

Absolutely true. if you are a programmer. Most musicians are not programmers and prefer their computers to be appliances.

pljones:
> Absolutely true. if you are a programmer. Most musicians are not programmers and prefer their computers to be appliances.

Or buy one: https://musicbridge.world/product/

ann0see:
Hmm yes. But if you already own an iPad, why would you want to invest into a raspberry pi/... I think the problem is that nowadays people expect everything to be mobile and easy. The raspberry Pi approach is a solution, but "geeky". Apps have a huge benefit and will move the project forward. If you view multiple videos, read articles, ... related to the topic of online jamming, we have disadvantages which come up all the time:

1. Mobile support (Android AND iOS)
2. Missing signatures (solvable)
3. WiFi

1 and 2 can be solved easily.

So from the discussion above I can conclude:

1. many people are in favor of mobile support (probably all). 
2. The GPL is a blocker
3. Peter is against adding a message like Nextcloud and doesn't support publishing the App on the AppStore.

--> since we can't find a consensus, does it even make sense to ask Volker or move mobile support for iOS based on the code of this repo forward?

If not, feel free to close this PR, merge the iOS sound support (at least we have a solid basis then) and:

1. Don't publish Jamulus on the AppStore
2. Allow Jamulus for iOS only via compillation and document it so that it's easily followable by newbies
3. Think about a solution (--> open discussion)

pljones:
> Peter is against adding a message like Nextcloud and doesn't support publishing the App on the AppStore.

I'm not _against_ publishing on the AppStore if it can be done with full compliance with GPL2.0 as is.  I'm not a lawyer but I don't like the ambiguity that comes from saying "Well, we're stating GPL2.0 except..." - the moment I hear an "except", I read it as a weakness that's going to be exploited, undermining the purpose of the GPL2.0.

ann0see:
> if it can be done with full compliance with GPL2.0 as is

Yes. That's the problem and the reason why this PR was raised. 

We need to ensure that the code is still GPL protected but also available through the AppStore. 

The easiest solution is a statement like Nextcloud uses. I assume GitHub user marinofaggiana committed this file. Not going to tag him here but maybe we'd like to contact him?

gene96817:
I suggest we submit the app clearly stating our GPL2.0 requirement. If Apple rejects the app, ask them what is needed to comply with BOTH GPLv2 and the app store. No need for us to be guessing.

ann0see:
@pljones I've now added 

In other words this means: As long as you comply with the GPL in all other respects, follow all requirements like providing
the source code and the text of the license, we will not object your distribution of Jamulus via the
Apple AppStore.

(similar to nextcloud).

LICENSE.APPLEAPPSTORE
```diff
@@ -0,0 +1,17 @@
+Jamulus contributors and maintainers are aware that the GPL and Apple App Store terms
+of service might conflict. Specific rights the GPL guarantees might not be possible
+to be granted by Apple AppStore distribution. Copyright holders of the source code do not
+want this legal conflict to prevent the distribution of Jamulus on the Apple App Store.
+
+Every contributor to the Jamulus source code and copyright holder shall not sue license
+violations soely based on conflicts between the GPL and the App Store guidelines which prevent
```
emlynmac, line 7: typo: solely

emlynmac (commented):
one little typo...

ann0see:
Sorry for tagging @corrados here. I've just received his E-Mail:
```
Hi <MY NAME>,

good to see Jamulus progressing :-)

if you are ok with Jamulus being published on the AppStore.
I have released the code under the GPL. This should allow you to decide it for yourself, right? I am no lawyer by any means so, I have no idea what consequences it has to publish a GPL software in the AppStore.

Anyway, publishing Jamulus in the AppStore is ok for me.

Best Regards,

Volker
```
He seems to be ok. with Jamulus being published on the AppStore.

LICENSE.APPLEAPPSTORE
ann0see: Done.

ann0see:
@pljones do we want to include this one now or how can we move this one forward?

pljones:
I still don't know what it means, having never dealt with the Apple app store.

The current change is like adding a chunk of code to Jamulus and saying "This code makes Jamulus better" but without explaining _how_ ("better" in what way) at all.

What needs doing is to make _explicit_ where there could be conflicts and how those conflicts are to be handled - not just a vague "oh, we'll sweep things under the carpet" statement.

gilgongo:
@pljones If I undertand correctly (and I may not), this rests on four points:

1. Putting Jamulus in the App Store is a violation of the GPL because Section 10 says, "You may not impose any further restrictions on the recipients' exercise of the rights granted herein." The [FSF is clear](https://www.fsf.org/blogs/licensing/more-about-the-app-store-gpl-enforcement ) that you should **not** put GPL-ed apps in the App Store for this reason.

2. The peole who wrote the Jamulus code have the copyright in that code under the GPL. 

3. In submitting Jamulus to the App Store, a person who does not have the copyright becuase the copyright holders have simply distributed it to them will be re-destributing it in a way that violates that copyright under the GPL.

4. So the proposed "disclaimer" protects that person (or persons who might put derived Jamulus code into the App Store as well) by saying the copyright holders won't persue that person or persons for copyright violation under the GPL as long as they otherwise comply with the terms of the GPL (like re-distrbuting source, etc.)

In so far as contract trumps copyright, this "Jamulus App Store Agreement" sort of works. I am told...

Note thouth that Apple seem to have/had a policy of rejecting/expelling GPL apps if they sneak in. So all this may one day collapse.

atsampson:
As a very occasional contributor I'm a bit concerned about this proposal. What you're now suggesting is changing Jamulus's license from the GPL to the GPL plus an exception which effectively turns it into a permissive license for one specific redistributor.

If you're going to change the license then you need explicit assent from **all** of Jamulus's existing copyright holders - which is quite an extensive list, looking at the source (don't forget things like the icons and the translations). You also need to make it very clear to new contributors that their work will no longer be accepted under a standard license.

I'm also worried that you don't seem to have sought legal advice on the wording of the exception - and I don't see any evidence that Nextcloud have done this either. There are plenty of examples in the past of homebrew license terms not producing the desired effect - it's easy to shoot yourself in the foot here. A few examples with the current wording: what happens if Apple change their ToS in the future to claim more rights that you're not happy with? What about if Apple rename the App Store or sell it to another company? Can a developer who wants to reuse Jamulus code in another project remove the exception terms (GPL exceptions usually have this provision)?

gilgongo:
@atsampson 

> If you're going to change the license then you need explicit assent from **all** of Jamulus's existing copyright holders 

To be clear - what's being proposed is **not** a licence (or dual-licensing) change. Jamulus will continue to be GPL, but we will publish a sort of "public contract" to it that says we won't persue copyright infringment in the specific case of people getting their Jamulus from the App Store.

I too share your concerns, but I think the worst that can (and probably will) happen is that Apple simply reject Jamulus :-)

ann0see:
Please read the updated statement in this PR. I hope it makes things clearer

ann0see:
> permissive license for one specific redistributor.

Does this come across with this addition?

atsampson:
No, what you're proposing really is a license change - you're explicitly granting permission in your statement for Apple to do things, such as redistribute Jamulus with additional restrictions, that they would not be able to do under the existing license. If you're going to make an assertion that the copyright holders won't sue someone for doing X - which is what a license _is_ - then you need to get agreement from the copyright holders to that effect.

I'll emphasise again that it would be a good idea to get legal advice to ensure that your statement really does exactly (and only!) what you intend it to do.

gilgongo:
We are not changing Jamulus's licence for anyone **other than** those who obtain it from the App Store. If you get Jamulus from somewhere else, you are bound by the GPL and only the GPL. That's not changing. As to getting agreement to do the App Store flip: are we not doing that here, now, with this PR? I'm not sure I undertand your point on that. 

All I'd say about getting legal advice is that its a very sensible thing to do :-)

pljones:
I think the problem is "assumed assent" / "assumed consent" may not hold up.  Absent parties cannot really be considered to have agreed to something they are not even aware is happening.  There have been code contributions under the GPL going back over many, many years.  I don't even know if all those contributors are aware we've moved to Github...

> Putting Jamulus in the App Store is a violation of the GPL because Section 10 says, "You may not impose any further restrictions on the recipients' exercise of the rights granted herein." The FSF is clear that you should not put GPL-ed apps in the App Store for this reason.

That _still_ doesn't answer the question: what are these "further restrictions"?  What am I being asked to consent to?

gilgongo:
> That still doesn't answer the question: what are these "further restrictions"? What am I being asked to consent to?

Presumably, these: https://www.apple.com/legal/internet-services/itunes/us/terms.html

But those terms govern the distirbution and use of the Jamulus binary by **users** of Jamulus, not you as a copyright holder of the code. Your rights to the Jamulus code under the GPL are unaffected by this proposal as far as I know. At least, Apple's terms say 
 that "Apple’s delivery of Services or Content does not transfer any commercial or promotional use rights to you, and does not constitute a grant or waiver of any rights of the copyright owners."

And I assume that if somebody had used the App Store for getting any app, they would have agreed to the terms with that already?

pljones:
Whilst the archived terms link from the blog are substantially different (being ten years older) than the current Apple T&C, they still seem pretty much heavy-handedly in violation of the GPL.

For a start, they claim that anything acquired through the App Store is an Apple Service or Apple Content.  That's a claim of ownership or branding:
> This Agreement governs your use of Apple’s services (“Services”), through which you can buy, get, license, rent or subscribe to content, Apps (as defined below), and other in-app services (collectively, “Content”). Content may be offered through the Services by Apple or a third party.

> SERVICES AND CONTENT USAGE RULES
> Your use of the Services and Content must follow the rules set forth in this section (“Usage Rules”). *Any other use of the Services and Content is a material breach of this Agreement.* Apple may monitor your use of the Services and Content to ensure that you are following these Usage Rules.

Basically, that says that Apple _could_ decide to take action against anyone using Jamulus in a manner that they consider a material breach of the agreement, if it was acquired through the App Store.  The "All Services:" section lists a number of additional restrictions.  However, their wording seems to focus on content rather than apps and there's very little stated about what can or can't actually be done.  Even in the section _explicitly_ about apps...

It is made clear that "LICENSED APPLICATION END USER LICENSE AGREEMENT" does not apply as we provide our own licence agreement (the GPL) -- it's not that part that's the issue, though, it's those earlier bits.

There's also
> Apple reserves the right at any time to modify this Agreement and to add new or additional terms or conditions on your use of the Services. Such modifications and additional terms and conditions will be effective immediately and incorporated into this Agreement. Your continued use of the Services will be deemed acceptance thereof.

Effectively, that hands Apple the right to change absolutely everything above that statement as they see fit and then enforce it.

gene96817:
> Apple reserves the right at any time to modify this Agreement and to add new or additional terms or conditions on your use of the Services. Such modifications and additional terms and conditions will be effective immediately and incorporated into this Agreement. Your continued use of the Services will be deemed acceptance thereof.

The last sentence "Your continued use.... will be deemed acceptance...." says if apple changes the terms and conditions inconsistent with our intent, we can remove Jamulus from the store. The removal is our rejection of the new T&C.

pljones:
How does that affect a Jamulus user who acquired "Content" -- Jamulus -- from the App Store and continues to use it?  Do they become bound by Apple's updated T&C?  I believe Apple expect so.

gene96817:
The new T&Cs will apply to new downloads. Since we would/should remove Jamulus from the app store as a rejection of the new T&Cs, there would be no downloads under the new terms. Continued use of the app downloaded before the new T&C would not fall under the new terms. 

In my non-lawyer opinion, the Apple terms mostly affect content created with Jamulus. We don't have any functions that give Apple any oversight over copyright issues. (oversight, meaning the storage and distribution of copyright content, similar to YouTube). I think we are worrying excessively about theoretical problems. If Apple raises any issue, we simply pull Jamulus.
We continue and consistently assert GPLv2.  The only real issue is if Apple can claim they reject GPLv2 licensed content.

gilgongo:
If I undertand correctly @emlynmac if anyone objects to this proposal, then those who do not object will re-write Jamulus for the app store with new code, and submit it that way (hence the wording of the disclaimer, which puzzled me at first).

But meanwhile, I think I can summarise all this such that any objection can be made a bit clearer:

1. Putting Jamulus in the app store is the only practical way of allowing its use on Apple mobile, which is seen as desireable by some people. It would also perhaps make more people aware of Jamulus whether they used it on their mobiles or not.

1. Apple's [T&Cs](https://www.apple.com/legal/internet-services/itunes/us/terms.html) apply to people who **obtain** Jamulus fom the app store. It does not affect the copyright holders' assertion of the GPL, nor does it affect any rights to the Jamlus "brand", music played with it, require a licence change, or any such thing affecting the project.

1. Although Apple will assert the GPL as the "Custom EULA” for Jamulus, it imposes some extra terms. This _act of addition_ is not allowed under the GPL - but there are no _specific_ terms that are being called out as unacceptable to us. Presumably also, anyone using any app store content would already be accepting of these constraints anyway. 

1. Because of the above points, we propose to waive any objections to the fact that Apple would violating the GPL in making Jamulus available on the app store. We'll do this by putting up a notice up to say we don't mind. 

1. Apple reseves the right to change its terms. If they do that in a way that we don't like, we simply pull Jamulus.

1. A copyright holder objecting to what we propose would presumably be declaring their intention to bring a legal case against Apple (and those who obtain Jamulus from the app store?). I have been told informally by somebody who knows a bit about this kind of thing that a court would take our waiver (and our discussion and agreement to it) into account.

pljones:
> We'll do this by putting up a notice up to say we don't mind.

I think your summary above should be included in such a notice as an explanation of why it's believed to be required.

ann0see:
@pljones @gilgongo I think you both have access to the file. I‘m not a native speaker and just wrote this note based on the one from Nextcloud. You can both give proposals to this file (and ofc. link this PR)

gilgongo:
OK I'll see if I can sugggest some more wording.

gilgongo (changes requested):
As per the discussion, the following may be clearer (BTW I'd change the file name to make sure it's a waiver not an additional licence).

APPLEAPPSTORE.LICENCE.WAIVER

The inclusion of Jamulus in the App Store is the only practical way of allowing its use on Apple mobile devices. Although Apple will assert the GPL as the "Custom EULA” for Jamulus, their T&Cs impose some extra conditions. This act of addition when distributing the software is not allowed under the GPL. 

However, Apple's terms apply to those who obtain Jamulus from the App Store. It does not affect the copyright holders' assertion of the GPL, or otherwise affect the Jamulus project: 

https://www.apple.com/legal/internet-services/itunes/us/terms.html

Moreover, there are no provisions in the Apple terms and conditions that we identify as being unacceptable to the copyright holders, or to the Jamulus project overall. Anyone using App Store services would already be accepting Apple’s terms for their use of any service they are obtaining. This includes being unable to re-distribute the software. 

The copyright holders do not want a legal conflict to prevent the distribution of Jamulus on the Apple App Store. Therefore, and in consideration of the above points, we the copyright holders in Jamulus hereby waive any objections to the fact that Apple would be violating the GPL in making Jamulus available on the App Store. 

Apple reserves the right to change its terms. If at any time they do that in a way that we find unacceptable, we will remove Jamulus from the App Store.

This waiver applies solely to the specific case of obtaining Jamulus from the Apple App Store. All other distribution methods must fully comply with the terms of the GPL.

Dissent

Any Jamulus contributor who does not wish their code to be published via the App Store must ensure their code is not so published. This can be achieved for example by removing their code from the Jamulus source, or by using technical measures (e.g. #ifdef) to ensure the code in question will not be compiled and included for App Store releases.

[ends]

ann0see:
> However, Apple's terms apply to those who obtain Jamulus from the App Store. It does not affect the copyright holders' assertion of the GPL, or otherwise affect the Jamulus project:

Are we 100% sure about that? Not a lawyer though…

gilgongo:
@ann0see Could be wrong of course, but I can't see anything in their T&Cs that assert anything over app providers. At least, the use of the term "you" in their terms only refers to those who download stuff, not make it.

emlynmac:
> @ann0see Could be wrong of course, but I can't see anything in their T&Cs that assert anything over app providers. At least, the use of the term "you" in their terms only refers to those who download stuff, not make it.

This was my understanding of the T&C also. @gilgongo I really like the updated wording - it's much clearer as to why.

gilgongo:
The wording could do with some amends for style maybe as it's a bit klunky in places, but I'll wait to see what others think.

APPLEAPPSTORE.LICENSE.WAIVER
```diff
@@ -0,0 +1,19 @@
+The inclusion of Jamulus in the App Store is the only practical way of allowing its use on Apple mobile devices. Although Apple will assert the GPL as the "Custom EULA” for Jamulus, their T&Cs impose some extra conditions. This act of addition when distributing the software is not allowed under the GPL.
+
+However, Apple's terms apply to those who obtain Jamulus from the App Store. It does not affect the copyright holders' assertion of the GPL, or otherwise affect the Jamulus project:
```
ann0see, line 3: We should make clear that this is our point of view/understanding/interpretation of the conflict and don't make it sound like we know better.

gilgongo: What do you mean by "know better" exactly?

ann0see: We‘re doing a (legal) interpretation of the conflict which we see as correct while some lawyers may not.

gilgongo: If there is anything in the statement that might be incorrect, then we need to fix that. We can't have a waiver that we also imply is not a waiver :-)

ann0see: I think after reading through it again, it's clear. There's still one point. See the comment below.

APPLEAPPSTORE.LICENSE.WAIVER
```diff
@@ -0,0 +1,19 @@
...
+The copyright holders do not want a legal conflict to prevent the distribution of Jamulus on the Apple App Store. Therefore, and in consideration of the above points, we the copyright holders in Jamulus hereby waive any objections to the fact that Apple would be violating the GPL in making Jamulus available on the App Store.
+
+Apple reserves the right to change its terms. If at any time they do that in a way that we find unacceptable, we will remove Jamulus from the App Store.
+
+This waiver applies solely to the specific case of obtaining Jamulus from the Apple App Store. All other distribution methods must fully comply with the terms of the GPL.
+
+Dissent
+
+Any Jamulus contributor who does not wish their code to be published via the App Store must ensure their code is not so published. This can be achieved for example by removing their code from the Jamulus source, or by using technical measures (e.g. #ifdef) to ensure the code in question will not be compiled and included for App Store releases.
+
```
ann0see, line 18:
```suggestion
Any Jamulus contributor who does not wish their code to be published via the App Store must ensure their code is not so published. This can be achieved for example by removing their code from the Jamulus source, or by using technical measures (e.g. #ifdef) to ensure the code in question will not be compiled and included for App Store releases.

Any Jamulus user who wants to distribute source code or binaries can use the official non-App Store releases or compile the source code himself.
```

gilgongo: I'm afraid I don't think we can legitimately say that. If somebody wants to distribute source code or binaries, they have to do more than just "use" the non-app store releases. We make it very clear  (I hope!) that  "All other distribution methods must fully comply with the terms of the GPL."

ann0see:
@gilgongo I think it‘s ready then?

gilgongo (approved):
I think we've done our best on this and our intentions are clear at least, so I'm OK with it going live.

ann0see:
Ok. Now we’re missing an approval from some other @jamulussoftware/maindevelopers  too ;-).

Issue #1576, opened 2021-04-29
Warn macOS user that he has disabled microphone access


ann0see:
**Has this feature been discussed and generally agreed?**

No, but it should be an easy task to do. 

**Describe the solution you'd like**

If a macOS user has disabled microphone access by wrongly selecting don’t allow microphone access, it might get confusing why the microphone "doesn’t work".

There should be a warning message which is invoked once you connect to a server which hints you at the problem. 

This message should 

1. Not stop the user from connecting to a server (maybe he’s just a listening client?)
2. Have a button to open privacy settings

Can’t work on it right now, so I‘d be happy if a macOS dev would look at it.

PR #1875, opened 2021-06-14
iOS sound support

ngocdh:

What's included in this PR:
- iOS sound support
- a feature for iOS to switch between external device (when plugged in) and the internal mic
- A crash fix related to the way iOS handles sockets in idle mode

**Context: Fixes an issue?**

As discussed in #1865, here is the new iOS-only PR.

Yes. Compilation for iOS is already documented.

Waiting on review; iOS sound is working. Some more real world tests.

Real world tests; review;

Later on: decide if we do a squash and merge.

## Checklist

**PR-Template applied by @ann0see**


ios/sound.h
```diff
@@ -44,64 +43,20 @@ class CSound : public CSoundBase
...
-    virtual QString LoadAndInitializeDriver ( QString strDriverName, bool );
-
-    QString CheckDeviceCapabilities ( const int iDriverIdx );
-    void    GetAvailableInOutDevices();
-
-    static void callbackMIDI ( const MIDIPacketList* pktlist, void* refCon, void* );
-
-    // AVAudioSession      audioSession;
-
     MIDIPortRef midiInPortRef;
```
ngocdh, line 59: to be removed...

ios/sound.h
```diff
@@ -44,64 +43,20 @@ class CSound : public CSoundBase
...
 
-    QString sChannelNamesInput[MAX_NUM_IN_OUT_CHANNELS];
-    QString sChannelNamesOutput[MAX_NUM_IN_OUT_CHANNELS];
-
     QMutex Mutex;
```
ngocdh, line 61: to be removed...

ios/ios_app_delegate.mm
```diff
@@ -7,10 +6,8 @@ @interface QIOSApplicationDelegate ()
 
 @implementation QIOSApplicationDelegate
 
-
-
-- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
-
+- (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(NSDictionary*)launchOptions
```
pljones, line 9: See, it's lines like this where I feel clang-format hasn't done its job right.
```
- ( BOOL ) application: ( UIApplication* ) application didFinishLaunchingWithOptions: ( NSDictionary* ) launchOptions
```
but I don't  know Objective-C...

ios/sound.mm
```diff
@@ -24,209 +24,277 @@
...
-    midiInPortRef ( static_cast<MIDIPortRef> ( NULL ) )
+#define kOutputBus 0
+#define kInputBus  1
+
+/**
+ This callback is called when sound card needs output data to play. And because Jamulus use the same buffer to store input and output data (input is
+ sent to server, then output is fetched from server), we actually use the output callback to read inputdata first, process it, and then copy the
+ output fetched from server to ioData, which will then be played.
+ */
+static OSStatus recordingCallback ( void*                       inRefCon,
```
pljones, line 35: Why isn't this a method on the application object?

ann0see:
@ngocdh  to keep the commit history clean, you should use amend style changes to the last commit and then git push --force the new commit to your branch ;-).

ios/sound.mm
```diff
@@ -24,209 +24,277 @@
...
-            {
-                // for mixed channels, show both audio channel names to be mixed
-                channelInputName[iCh] =
-                    channelInputName[iSelCH] + " + " + channelInputName[iSelAddCH];
-            }
-        }
+        NSError* audioSessionError = nil;
+
+        [[AVAudioSession sharedInstance] setCategory:AVAudioSessionCategoryPlayAndRecord error:&audioSessionError];
+        [[AVAudioSession sharedInstance] requestRecordPermission:^( BOOL granted ) {
```
pljones, line 108: Indentation appears to be off here.  It should indent four spaces (at least), it's only indented two.  The block type is `[{..}]` rather than `{..}`, which is presumably confusing it.

ios/sound.mm
```diff
@@ -24,209 +24,277 @@
...
+          if ( granted )
+          {
+              // ok
```
pljones, line 111: What if it isn't..?

ios/sound.mm
```diff
@@ -24,209 +24,277 @@
...
+          }
+        }];
+        [[AVAudioSession sharedInstance] setMode:AVAudioSessionModeMeasurement error:&audioSessionError];
     }
-    else
+    catch ( const CGenErr& generr )
     {
-        // regular case: no mixing input channels used
-        lNumInChanPlusAddChan = lNumInChan;
+        qDebug ( "Sound exception ...." );
```
pljones, line 118: Generally I've seen this style in the code:
```
qDebug() << "Sound exception";
```
Is there a reason to do it differently?

ios/sound.mm
ann0see: It should probably show an error message. (Which also doesn't happen on macOS https://github.com/jamulussoftware/jamulus/issues/1576)

ios/sound.mm
ngocdh: Just 2 ways to do the same thing. Although your way requires #include

ios/sound.mm
```diff
@@ -24,209 +24,277 @@
...
     }
-*/
-    // everything is ok, return empty string for "no error" case
-    return "";
+    isInitialized = false;
```
pljones, line 120: Shouldn't this be part of the constructor init right at the top?
```
CSound::CSound ( void ( *fpNewProcessCallback ) ( CVector<short>& psData, void* arg ),
                 void*          arg,
                 const QString& strMIDISetup,
                 const bool,
                 const QString& ) :
    isInitialized ( false ),
    CSoundBase ( "CoreAudio iOS", fpNewProcessCallback, arg, strMIDISetup )
{
...
```

ios/sound.mm
```diff
@@ -24,209 +24,277 @@
...
+int CSound::Init ( const int iCoreAudioBufferSizeMono )
 {
-    /*// apply parameter after input parameter check
-    if ( ( iNewChan >= 0 ) && ( iNewChan < iNumInChanPlusAddChan ) )
+    try
     {
-        iSelInputLeftChannel = iNewChan;
+        // printf("Init sound ..."); <= to check the number of Sound inits at launch
+        // init base class
+        // CSoundBase::Init ( iCoreAudioBufferSizeMono ); this does nothing
```
pljones, line 129: Could do with removing any comments like this.

ios/sound.mm
```diff
@@ -24,209 +24,277 @@
...
+        this->iCoreAudioBufferSizeMono = iCoreAudioBufferSizeMono;
+
+        // set internal buffer size value and calculate stereo buffer size
+        iCoreAudioBufferSizeStereo = 2 * iCoreAudioBufferSizeMono;
+
+        // create memory for intermediate audio buffer
+        vecsTmpAudioSndCrdStereo.Init ( iCoreAudioBufferSizeStereo );
+
+        AVAudioSession* sessionInstance = [AVAudioSession sharedInstance];
```
pljones, line 138: See, I'm happy with this spacing.  It's equivalent of `something = other.methodWithReturnValue();`
...

ios/sound.mm
ngocdh: I agree a constructor should be at the top. Must it though?

Lol just trolling you. I'll update the code

ios/sound.mm
```diff
@@ -24,209 +24,277 @@
...
+
+        // we are going to play and record so we pick that category
+        NSError* error = nil;
+        [sessionInstance setCategory:AVAudioSessionCategoryPlayAndRecord
+                         withOptions:AVAudioSessionCategoryOptionMixWithOthers | AVAudioSessionCategoryOptionDefaultToSpeaker |
+                                     AVAudioSessionCategoryOptionAllowBluetooth | AVAudioSessionCategoryOptionAllowBluetoothA2DP |
+                                     AVAudioSessionCategoryOptionAllowAirPlay
+                               error:&error];
```
pljones, line 146: ...
but the spacing here is wrong.  The equivalent would be `other.method ( thing, otherThing, anotherThing );`

I also don't like `error:&error` - I want more spaces... presumably after all the `:`s except with `:&` where it would be after the `:&` -- `error:& error`.

ios/sound.mm
```diff
@@ -24,209 +24,277 @@
...
+
+        // Initialise
+        status = AudioUnitInitialize ( audioUnit );
+        checkStatus ( status );
+
+        isInitialized = true;
+    }
+    catch ( const CGenErr& generr )
+    {
+        qDebug ( "Sound Init exception ...." );
```
pljones, line 229: And again regarding `qDebug() << "message";`

ios/sound.mm
```diff
@@ -24,209 +24,277 @@
...
+    // call base class
+    CSoundBase::Start();
+    try
+    {
+        OSStatus err = AudioOutputUnitStart ( audioUnit );
+        checkStatus ( err );
+    }
+    catch ( const CGenErr& generr )
+    {
+        qDebug ( "Sound Start exception ...." );
```
pljones, line 246: And here.

ios/sound.mm
```diff
@@ -24,209 +24,277 @@
...
-      vSelectedOutputChannels[1] = iNewChan;
-  }*/
+    try
+    {
+        OSStatus err = AudioOutputUnitStop ( audioUnit );
+        checkStatus ( err );
+    }
+    catch ( const CGenErr& generr )
+    {
+        qDebug ( "Sound Stop exception ...." );
```
pljones, line 259: ...

src/main.cpp
```diff
@@ -539,6 +539,7 @@ int main ( int argc, char** argv )
 #else
 #    if defined( Q_OS_IOS )
     bUseGUI        = true;
+    bIsClient      = true; // Client only - TODO: maybe a switch in interface to change to server?
```
pljones, line 542: Or a separate install for the server?  (MacOS has two bundles.)

ngocdh: Yeah why not. However I'm a bit skeptical about iOS server performance (tried it briefly, processing time was high).

ios/sound.mm
ngocdh: We'll have to work that .clang-format out first

src/socket.cpp
```diff
@@ -28,6 +28,11 @@
 /* Implementation *************************************************************/
 void CSocket::Init ( const quint16 iPortNumber, const quint16 iQosNumber, const QString& strServerBindIP )
 {
+    // first store parameters, in case reinit is required (mostly for iOS)
+    iPortNumber_     = iPortNumber;
```
pljones, line 32: This style isn't used elsewhere.  Please don't introduce new coding styles.  You don't need to do this, either:
```
this.varname = varname;
```
works.

src/socket.cpp
```diff
@@ -142,6 +155,8 @@ void CSocket::Init ( const quint16 iPortNumber, const quint16 iQosNumber, const
 
         QObject::connect ( this, &CSocket::ServerFull, pServer, &CServer::OnServerFull );
     }
+
+    bIsInitRan = true; // QObject::connect once only
```
pljones, line 159: If there are things that need to happen multiple times and other things that should only happen once - put them in separate methods.  Then you can decide which to call.  This method can then call the "multiple" times method - but only itself be called once.

src/socket.cpp
```diff
@@ -188,12 +203,22 @@ void CSocket::SendPacket ( const CVector<uint8_t>& vecbySendBuf, const CHostAddr
...
-                 sizeof ( sockaddr_in ) );
+        if ( sendto ( UdpSocket,
+                      (const char*) &( (CVector<uint8_t>) vecbySendBuf )[0],
+                      iVecSizeOut,
+                      0,
+                      (sockaddr*) &UdpSocketOutAddr,
+                      sizeof ( sockaddr_in ) ) < 0 )
+        {
+            // qDebug("Socket send exception - mostly happens in iOS when returning from idle");
+            Init ( iPortNumber_, iQosNumber_, strServerBindIP_ ); // reinit
```
pljones, line 214: As earlier, variable names here need correcting.

src/socket.h
```diff
@@ -77,7 +77,11 @@ class CSocket : public QObject
     void Close();
 
 protected:
-    void Init ( const quint16 iPortNumber, const quint16 iQosNumber, const QString& strServerBindIP );
+    void    Init ( const quint16 iPortNumber, const quint16 iQosNumber, const QString& strServerBindIP );
+    quint16 iPortNumber_;
```
pljones, line 81: As earlier.

ios/sound.mm
pljones: That's why we've code style guidelines that say "do it like it's done elsewhere" rather than "do whatever you like".

ios/sound.mm
pljones: Yeah... mostly it's just irritation having had to adapt to the spacing in the rest of Jamulus ;P

src/main.cpp
pljones: If a Pi4 can do it, a recent iOS device should be powerful enough, I'd have thought?

ann0see: This hints at a thought I had a while ago: could we add a GUI button to switch into server mode? Probably possible, but this might require some big changes.

Building two bundles for macOS is hacky.

pljones: We have two GUIs.

We have two main services.

We bundle all four into a single executable and rely on command line options to pick which of the four modes to operate in.

We could have four smaller executables with the shared code in a separate library.  (Again, part of what I'd like to see in the future - restructuring the code like that will help get the GUI and service layers split more neatly.)

ngocdh:
> @ngocdh  to keep the commit history clean, you should use amend style changes to the last commit and then git push --force the new commit to your branch ;-).

I actually have no idea what you're talking about due to my very limited experience, but thanks I'll keep that in mind.

ann0see:
No problem ;-). 

Let’s say you did a commit and changed a typo. After committing and pushing, you realise that there’s another typo. Ok, let’s do another commit. Oh no. Now I accidentally deleted a closing bracket —> next commit. Argh, no there’s another typo —> commit etc. As you see, you now have lots of unneeded commits. 

What if you could modify these commits and just fix typos in one commit? 

That’s the place where you could use  `git commit —amend`: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

And then `git push —force` the new commit history and be happy about a clean and understandable history. 

Last year (or this year (?)) I also committed a bunch of unneeded commits to my repo and it was part of a PR. It got merged as is into master and now there’s a lot of garbage in there.

src/socket.cpp
ngocdh: ok thanks!

ios/sound.mm
ngocdh: I was following an iOS sound code example, didn't really think about it. Thanks to you, I put the function inside CSound, and it still works!

ios/sound.mm
ngocdh: I say let's just let those style issues slide for now. When we have a better .clang-format, it's just one command away to correct the .mm style for the whole project.

ios/sound.mm
ngocdh: ident 4 actually

ios/sound.mm
ngocdh: I actually have a valid reason to limit the libraries included. Initially, the iOS code had a few more unnecessary #includes, and @ann0see iPhone couldn't even launch Jamulus because it took too long. Launched after removing libs, so I prefer to stay away from #include when possible.

The `qDebug() <<` syntax requires `#include <QDebug>`.

src/socket.cpp
ngocdh: Thanks to you I've found a way to make it look a lot better now, no more `if ( bIsInitRan )`.

src/socket.cpp
ngocdh: Done

ios/sound.mm
ngocdh: hmm I misunderstood.

The reason was simple: I didn't know how/when to put it in the constructor init. Not a developer at heart.

ngocdh:
Ok so I went thru the very thorough reviews from @pljones. Now my code is much better, but I'm refraining from sending a commit to keep history clean because there's still the clang-format to sort out.

I say let's just let those style issues slide for now. When we have a better .clang-format, it's just one command away to correct the ObjC style for the whole project.

What does the team think?

ann0see:
You could amend https://github.com/jamulussoftware/jamulus/pull/1875/commits/6ac16f6c3517e16996d425225a745a66f682816f

pljones:
Yep, we'll have to comply with the exist clang-format checks on this PR.

ios/sound.mm
```diff
@@ -24,209 +24,277 @@
...
-    // ??
+    // Now, we have the samples we just read sitting in buffers in bufferList
+    // Process the new data
+    pSound->processBufferList ( &bufferList, pSound ); // THIS IS WHERE vecsStereo is filled with data from bufferList
 
-    // call base class
-    //CSoundBase::Start();
-}
+    // release the malloc'ed data in the buffer we created earlier
+    free ( bufferList.mBuffers[0].mData );
```
pljones, line 66: There should be _no_ (absolute) allocation and freeing of memory inside the audio processing loop.  It's a major source of performance hits.  Admittedly, I'm not familiar with how it's done for other platforms but if possible, allocate the buffer once when the app starts, then re-use it.

ios/sound.mm
pljones: The code here on github still has the method as a static global.  Did you miss your change getting committed and pushed?

ios/sound.mm
pljones: I meant following the `{` at the end of the line - i.e. what's currently line 109 is only two spaces indented from line 108, whereas it should be 4.

ngocdh: I see now!

ios/sound.mm
ngocdh: I didn't commit, working on --amend

ios/sound.mm
ngocdh: You might be right. I was following this example: https://github.com/Epskampie/ios-coreaudio-example/blob/master/IosAudioController.m which really helped me.
About performance, I've been playing with my iPhone to listen to people and jam for hours and didn't really have any problem.
I'll keep that in mind and see if I can do anything to make it better.

ngocdh:
New code! Huge thanks to @pljones and @ann0see !!!

src/socket.cpp
```diff
@@ -26,8 +26,52 @@
 #include "server.h"
 
 /* Implementation *************************************************************/
-void CSocket::Init ( const quint16 iPortNumber, const quint16 iQosNumber, const QString& strServerBindIP )
+
+// Connections -------------------------------------------------------------
```
pljones, line 30: This is much better and much clearer.

ios/sound.mm
ngocdh: I was trying to use a pointer to store the buffer address, so that the buffer only needs to be initialized once and can be reused. However, since `recordingCallback` is a static function, the pointer must also be static. When compiling in xCode, I had this linker error (couldn't access that pointer from CSound or something).

Do you know how to overcome this error?

pljones: I'd suggest making recordingCallback _not_ static, as a member of a single-instance class it should still be unique - so make sure the class it belongs to _is_ single-instance.  Then the buffer pointer should be a member of the same single-instance class, pointing to memory allocated when that class is instantiated.

I don't know iOS or ObjectiveC, though, to know how best to make that happen.

ann0see:
Thanks. I‘d love to have a second review from @emlynmac here since he‘s an iOS dev.

ios/sound.mm
ngocdh: I tried, but it seems the callback has to be static :(

ngocdh:
> Thanks. I‘d love to have a second review from @emlynmac here since he‘s an iOS dev.

Good idea. Especially if @emlynmac has tricks to boost up launch time that would be great.

ios/sound.mm
```diff
@@ -24,209 +24,275 @@
...
+    try
+    {
+        NSError* audioSessionError = nil;
 
-  //GetAvailableInOutDevices();
+        [[AVAudioSession sharedInstance] setCategory:AVAudioSessionCategoryPlayAndRecord error:&audioSessionError];
+        [[AVAudioSession sharedInstance] requestRecordPermission:^( BOOL granted ) {
+            if ( granted )
+            {
+                // ok
```
emlynmac, line 47: Might want to handle the case where permissions are not granted?

ios/sound.mm
```diff
@@ -24,209 +24,275 @@
...
+        pSound->vecsTmpAudioSndCrdStereo[2 * i + 1] = (short) ( pData[2 * i + 1] * _MAXSHORT ); // right
     }
-*/
-    // everything is ok, return empty string for "no error" case
-    return "";
+    pSound->ProcessCallback ( pSound->vecsTmpAudioSndCrdStereo );
 }
 
-void CSound::SetLeftInputChannel  ( const int iNewChan )
+// TODO - CSound::Init is called multiple times at launch to verify device capabilities.
```
emlynmac, line 122: This smells like maybe the calling code needs to be reworked?

ios/sound.mm
```diff
@@ -24,209 +24,275 @@
...
+        // we are going to play and record so we pick that category
+        NSError* error = nil;
+        [sessionInstance setCategory:AVAudioSessionCategoryPlayAndRecord
+                         withOptions:AVAudioSessionCategoryOptionMixWithOthers | AVAudioSessionCategoryOptionDefaultToSpeaker |
+                                     AVAudioSessionCategoryOptionAllowBluetooth | AVAudioSessionCategoryOptionAllowBluetoothA2DP |
+                                     AVAudioSessionCategoryOptionAllowAirPlay
+                               error:&error];
+
+        // using values from jamulus settings 64 = 2.67ms/2
+        NSTimeInterval bufferDuration = iCoreAudioBufferSizeMono / 48000.0; // yeah it's math
```
emlynmac, line 147: This is not defined somewhere already? 48k magic number?

ios/sound.mm
```diff
@@ -24,209 +24,275 @@
...
+        status      = AudioUnitSetProperty ( audioUnit, kAudioOutputUnitProperty_EnableIO, kAudioUnitScope_Input, kInputBus, &flag, sizeof ( flag ) );
+        checkStatus ( status );
+
+        // Enable IO for playback
+        status = AudioUnitSetProperty ( audioUnit, kAudioOutputUnitProperty_EnableIO, kAudioUnitScope_Output, kOutputBus, &flag, sizeof ( flag ) );
+        checkStatus ( status );
+
+        // Describe format
+        AudioStreamBasicDescription audioFormat;
+        audioFormat.mSampleRate       = 48000.00;
```
emlynmac, line 182: And probably best to have a constant variable as it is reused a lot

src/main.cpp
emlynmac: I would expect no real demand / need to run a server on an iOS device.
I see a strong use for a client here.

emlynmac (commented):
Some comments.
I'd be curious to see what the differences end up being between the macOS and iOS core audio interface layers. Can they be the same file with defines ultimately?

ios/sound.mm
ngocdh: Yes, see #1727

ios/sound.mm
ngocdh: Yeah same as #1576, when permission is not granted, app continues to work, just without sound.
Maybe an improvement in CSoundBase: a signal to CClient for example.

ios/sound.mm
ann0see: If it is the sample rate, there should be a constant somewhere. SYSTEM_SAMPLE_RATE or PREFERRED_SAMPLE_RATE need to check the Code.

ios/ios_app_delegate.mm
pljones: Is that "too many" as in "won't compile" or as in "not normal"?

emlynmac: Oh it's just purely a stylistic comment - there are _religious_ discussions about how to format Obj-C....
It makes no difference to compilation.

ios/sound.mm
emlynmac: #include changes don't change the launch time.

If the function call requires the library to work then I don't see how you can avoid it.

As far as start up performance goes, you have to not block the main thread for long periods or you'll be killed by the system. It's pretty established that this is the ::init call happening too many times.

ngocdh: > #include changes don't change the launch time.

It seemed to do though. We did a test with @ann0see iphone which at first couldn't launch because it took too long. Removed some unused libs, and it launched.
 
> If the function call requires the library to work then I don't see how you can avoid it.

The thing here is qDebug("something") can be called without `#include QDebug`, while `qDebug() << "something"` can't. Although I'm sure QDebug is already included in other ios-unrelated parts of the code, I'm just trying to optimize because ios launch time is yet to be improved.

> As far as start up performance goes, you have to not block the main thread for long periods or you'll be killed by the system. It's pretty established that this is the ::init call happening too many times.

Yeah I proposed multithreading...

I actually see another potential issue: MAX_NUM_CHANNELS (default value=150). If I'm not mistaken, at launch, Jamulus creates MAX_NUM_CHANNELS faders in the mixer, and hide unused ones. 

However, seeing that other OSes don't seem to have problem with this, the problem might be CSound::Init, which is specific to each system.

ios/sound.mm
ngocdh: ha ha got it working! The secret is the pointer has to be called like this: `pSound->pointer`, not just `pointer`. Now 1 malloc only in constructor, which is freed in destructor. It seems to work, but I'm testing a bit before committing.

ngocdh: Also I checked the MacOS sound code, and there's mallocs there ...

ngocdh:
> I'd be curious to see what the differences end up being between the macOS and iOS core audio interface layers. Can they be the same file with defines ultimately?

They are very different. Moreover macOS is much more capable than iOS, thus code is more complicated. So, combining them into one file complicates things a lot.

ios/sound.mm
ngocdh: updated!

ios/sound.h
```diff
@@ -1,8 +1,8 @@
 /******************************************************************************\
- * Copyright (c) 2004-2020
+ * Copyright (c) 2004-2021
  *
  * Author(s):
- *  ann0see based on code from Volker Fischer
+ *  ann0see and ngocdh based on code from Volker Fischer
```
ann0see, line 5: In fact, you could remove me here since you did most of the work:

```suggestion
 *  ngocdh based on code from Volker Fischer
```

ios/sound.mm
```diff
@@ -1,8 +1,8 @@
 /******************************************************************************\
- * Copyright (c) 2004-2020
+ * Copyright (c) 2004-2021
  *
  * Author(s):
- *  ann0see based on code from Volker Fischer
+ *  ann0see and ngocdh based on code from Volker Fischer
```
ann0see, line 5:
```suggestion
 *  ngocdh based on code from Volker Fischer
```

ios/sound.mm
```diff
@@ -24,209 +24,279 @@
...
 /* Implementation *************************************************************/
-CSound::CSound ( void           (*fpNewProcessCallback) ( CVector<short>& psData, void* arg ),
+CSound::CSound ( void ( *fpNewProcessCallback ) ( CVector<short>& psData, void* arg ),
                  void*          arg,
                  const QString& strMIDISetup,
-                 const bool     ,
+                 const bool,
                  const QString& ) :
     CSoundBase ( "CoreAudio iOS", fpNewProcessCallback, arg, strMIDISetup ),
-    midiInPortRef ( static_cast<MIDIPortRef> ( NULL ) )
```
ann0see, line 35: So there's no midi support probably.

ios/sound.mm
ann0see: Doesn't need to be addressed here. Probably use another PR for that.

ios/sound.mm
```diff
@@ -24,209 +24,279 @@
...
+            else
+            {
+                //TODO - signal to CClient?
+            }
+        }];
+        [[AVAudioSession sharedInstance] setMode:AVAudioSessionModeMeasurement error:&audioSessionError];
+    }
+    catch ( const CGenErr& generr )
+    {
+        qDebug ( "Sound exception ...." );
```
ann0see, line 58: This is not very useful... at least we should output what the error was.

ios/sound.mm
```diff
@@ -24,209 +24,279 @@
...
+        }
+        else
+        {
+            unsigned long lastInput = sessionInstance.availableInputs.count - 1;
+            [sessionInstance setPreferredInput:sessionInstance.availableInputs[lastInput] error:&error];
+        }
+    }
+    catch ( const CGenErr& generr )
+    {
+        qDebug ( "Sound dev change exception ...." );
```
ann0see, line 292: All these errors don't seem to be shown to the user. Shouldn't we output them as a warning message?

ios/sound.h
ngocdh: I'll keep credit where credit is due :)

ios/sound.mm
ngocdh: iOS can do MIDI. Not me and my PR though :D

ios/sound.mm
ngocdh: Ok will do

ios/sound.mm
ngocdh: using QMessageBox?

ann0see: Yes, to keep in Qt not native objective C.

ios/sound.mm
ann0see: ;-).

ann0see:
I've just applied my PR template to make the process more visible.

ngocdh:
Thanks @ann0see

ios/sound.mm
```diff
@@ -46,13 +46,27 @@
...
         [[AVAudioSession sharedInstance] setMode:AVAudioSessionModeMeasurement error:&audioSessionError];
     }
     catch ( const CGenErr& generr )
     {
-        qDebug ( "Sound exception ...." );
+        QMessageBox::warning( nullptr, "Sound exception", generr.GetErrorText() );
     }
+
+    buffer.mNumberChannels    = 2;
+    buffer.mData              = malloc ( 256 * sizeof ( Float32 ) * 2 ); // max size
```
pljones, line 62: Is `2` in the `* 2` the same as the `= 2` on the previous line?  If it is, could you use `buffer.mNumberChannels` to be clearer?

Also, what's "m" on the front of the variable names?  I've not seen it in the code base before?

pljones:
I think the main thing I'm not comfortable with is the hardware-specific code creeping into src/client.cpp (Input Device Id, Built In Mic Id) that feel like they should be handled - somehow - in the iOS sound code.

ios/sound.mm
ngocdh: Yes it’s the same. Thanks I’ll update the code.

“m” is from Apple: https://developer.apple.com/documentation/coreaudiotypes/audiobuffer/1421973-mnumberchannels?changes=l_3_8&preferredLanguage=occ

pljones: Ah right.

ngocdh:
> I think the main thing I'm not comfortable with is the hardware-specific code creeping into src/client.cpp (Input Device Id, Built In Mic Id) that feel like they should be handled - somehow - in the iOS sound code.

The only thing that code does is call the CSound code which handles the action. The real code is in iOS CSound, just like for Android if we want to implement the same feature.

pljones:
Yes but audio devices shouldn't appear _anywhere_ outside the sound code.  They should be exposed using the existing interface.

Look at JACK vs CoreAudio vs ASIO.  All expose different devices without needing to have special device-specific code in src/client.cpp.

If there are bits of hardware that need to be exposed, add a "virtual device", add it to the device list from the OS, and expose it on the list of available devices.

ngocdh:
> If there are bits of hardware that need to be exposed, add a "virtual device", add it to the device list from the OS, and expose it on the list of available devices.

You're asking me to do what I did in a much complicated way. I would have to detect new device/lost device events which I don't know how to do. 

And more importantly, a mobile user would have to click settings, go to the Setup tab if not selected, click on the device combobox, select the device, click ok to confirm, and eventually Close to quit the settings dialog.  

Imagine you're on mobile and want to quickly switch from your instrument to the phone's mic to talk for a moment before continuing with the instrument, that's definitely not the best workflow.

pljones:
I'm asking you do what needs doing to keep Jamulus working the way Jamulus works.  What you're doing makes future changes harder than they need be.  Putting the effort in now, once, will save that repeated effort in the future.  Code cleanliness is important.

emlynmac:
> Imagine you're on mobile and want to quickly switch from your instrument to the phone's mic to talk for a moment before continuing with the instrument, that's definitely not the best workflow.

I think this is a perfect use case for something else that I want to get in; ability to use multi-channel multi interface sound. I'm referring to https://github.com/jamulussoftware/jamulus/discussions/1840 and although this is primarily about multi-input interfaces, it would be great to be able to have multiple interfaces mapped through the Jamulus channel mixer. Then you'd be able to quickly unmute your voice channel and then carry on playing etc.

pljones:
Take a look into PortAudio, then.  That should be how things are handled.

ngocdh:
I've removed the feature. I will implement what @pljones suggested if I have time. However, @emlynmac I doubt multiple simultaneous hardware inputs is possible in iOS.

ann0see:
@ngocdh I think you can also remove https://github.com/jamulussoftware/jamulus/blob/e0cac75ee0d5ef0a6f85f6b88367abab9b659853/Jamulus.pro#L201

ngocdh:
I did. Are you reviewing @emlynmac code?

ann0see:
Urgh. Sorry

ngocdh:
Ok the switch device feature is back, this time implemented as a "virtual device", with device change events handling.

ios/sound.mm
```diff
@@ -24,209 +24,323 @@
 #include "sound.h"
 #include <AVFoundation/AVFoundation.h>
 
+#define kOutputBus 0
```
pljones, line 27: 1. Should be in `ios/sound.h`
2. Style should be block caps
3. Naming should be something like `IOS_DEVICE_OUTPUT_BUS` and `IOS_DEVICE_INPUT_BUS`, (if "DEVICE" is appropriate).

ios/sound.mm
```diff
@@ -24,209 +24,323 @@
...
-    //CSoundBase::Stop();
+    return noErr;
 }
 
-void CSound::GetAvailableInOutDevices()
+void CSound::processBufferList ( AudioBufferList* inInputData, CSound* pSound ) // got stereo input data
 {
-  // https://developer.apple.com/documentation/avfoundation/avaudiosession/1616557-availableinputs?language=objc?
+    QMutexLocker locker ( &pSound->MutexAudioProcessCallback );
+    Float32*     pData = static_cast<Float32*> ( inInputData->mBuffers[0].mData );
```
pljones, line 112: Are we processing float audio or integer audio in Jamulus?  If it's still integer on master, can this be integer, too, please.

ios/sound.mm
```diff
@@ -24,209 +24,323 @@
...
+        checkStatus ( status );
+
+        // Describe format
+        AudioStreamBasicDescription audioFormat;
+        audioFormat.mSampleRate       = SYSTEM_SAMPLE_RATE_HZ;
+        audioFormat.mFormatID         = kAudioFormatLinearPCM;
+        audioFormat.mFormatFlags      = kAudioFormatFlagsNativeEndian | kAudioFormatFlagIsFloat | kAudioFormatFlagIsPacked;
+        audioFormat.mFramesPerPacket  = 1;
+        audioFormat.mChannelsPerFrame = 2;  // steoreo, so 2 interleaved channels
+        audioFormat.mBitsPerChannel   = 32; // sizeof float32
```
pljones, line 189: Again, if we're processing PCM 16 bit fixed, what needs changing here?  (And anywhere else in `audioFormat`.)

ios/sound.mm
```diff
@@ -24,209 +24,323 @@
...
+        status = AudioUnitSetProperty ( audioUnit, kAudioOutputUnitProperty_EnableIO, kAudioUnitScope_Output, kOutputBus, &flag, sizeof ( flag ) );
+        checkStatus ( status );
+
+        // Describe format
+        AudioStreamBasicDescription audioFormat;
+        audioFormat.mSampleRate       = SYSTEM_SAMPLE_RATE_HZ;
+        audioFormat.mFormatID         = kAudioFormatLinearPCM;
+        audioFormat.mFormatFlags      = kAudioFormatFlagsNativeEndian | kAudioFormatFlagIsFloat | kAudioFormatFlagIsPacked;
+        audioFormat.mFramesPerPacket  = 1;
+        audioFormat.mChannelsPerFrame = 2;  // steoreo, so 2 interleaved channels
```
pljones, line 188: `steoreo` -> `stereo`

ios/sound.mm
pljones: Hm.  You're following Apple SDK guidelines, not Jamulus ones.  For something like there, where the Apple SDK shouldn't care, the Jamulus style guide should take precedence, I think.

ios/sound.mm
```diff
@@ -24,209 +24,323 @@
...
-            // copy packet and send it to the MIDI parser
-            CVector<uint8_t> vMIDIPaketBytes ( midiPacket->length );
-            for ( int i = 0; i < midiPacket->length; i++ )
-            {
-                vMIDIPaketBytes[i] = static_cast<uint8_t> ( midiPacket->data[i] );
-            }
-            pSound->ParseMIDIMessage ( vMIDIPaketBytes );
-
-            midiPacket = MIDIPacketNext ( midiPacket );
+            iDriverIdx = i;
```
pljones, line 329: If the loop can stop, put a `break` in here.

ios/sound.mm
ngocdh: Check this https://github.com/jamulussoftware/jamulus/blob/e0cac75ee0d5ef0a6f85f6b88367abab9b659853/mac/sound.cpp#L929

pljones: That's not Jamulus sound, that's `mac/sound.cpp`.

ngocdh: Yes, I took the code from there.

ngocdh:
Thanks @pljones, feel free to tell me when you're all done with the reviews :P

ios/sound.mm
pljones: Yes, but it should match the code in the main part of Jamulus to avoid unnecessary data conversions which slow down processing.

ios/sound.mm
```diff
@@ -24,209 +24,323 @@
...
+
+        // create memory for intermediate audio buffer
+        vecsTmpAudioSndCrdStereo.Init ( iCoreAudioBufferSizeStereo );
+
+        AVAudioSession* sessionInstance = [AVAudioSession sharedInstance];
+
+        // we are going to play and record so we pick that category
+        NSError* error = nil;
+        [sessionInstance setCategory:AVAudioSessionCategoryPlayAndRecord
+                         withOptions:AVAudioSessionCategoryOptionMixWithOthers | AVAudioSessionCategoryOptionDefaultToSpeaker |
```
emlynmac, line 143: Do we really want these options? I would have thought to use jamulus we'd want exclusive and not over bluetooth or airplay due to latency concerns.

ngocdh: In case you want to join as Listener, delay is not an issue.
Bluetooth (and probably AirPlay too) is not working properly though because of sample rate difference.

ann0see:
@ngocdh @pljones I‘d really like to have iOS sound compilation support in the next release. Since I don’t see any regressions, can we merge this ad beta?

pljones:
To approve this, I'd want to see:
- rebase and squash, plus confirmation it still works as expected
- an open issue to address the outstanding review comments in future releases (not necessarily "in the next version)
- a clear notice that it's experimental

ann0see:
@emlynmac @softins sorry for the ping. Could you please have a look at this PR? 

I can rebase and test it on my iPhone SE.

emlynmac:
I've been keeping an eye on this PR - I'd like to see it resolved against master first

ngocdh:
Team, I'd love to make this PR happen too, the sooner the better so that we don't have to review this over and over again. 

Could you be specific about what need to be done here?
- changes according to latest comments from @pljones <- should be ok
- conflicts resolution <- same
- rebase and squash: I don't know how to. Is there a way someone could do this on my behalf, or something like a change suggestion that I could simply "Accept"?

Anything else?

ann0see:
Didn’t watch this, but maybe this helps with the rebase: https://m.youtube.com/watch?v=_UZEXUrj-Ds


Of course, every main developer could do the rebase for you.

emlynmac:
> Team, I'd love to make this PR happen too, the sooner the better so that we don't have to review this over and over again.
> 
> Could you be specific about what need to be done here?
> 
> * changes according to latest comments from @pljones <- should be ok
> * conflicts resolution <- same
> * rebase and squash: I don't know how to. Is there a way someone could do this on my behalf, or something like a change suggestion that I could simply "Accept"?
> 
> Anything else?

PR has conflicts with master branch.

first, switch to `master` and make sure you have the latest changes locally.
Then, on your branch, do a `git rebase master` and make sure all the changes are applied properly. You'll need to resolve conflicts as you go. 
Then, force push to your branch and we should see only your changes as if they were applied to the tip of the current master branch. 

The PR should then be much cleaner.

ann0see:
To me it seems as if the conflicts are mainly due to the IPv6 changes in src/socket.*: https://github.com/jamulussoftware/jamulus/pull/1938

pljones:
> Then, on your branch, do a `git rebase master` and make sure all the changes are applied properly. You'll need to resolve conflicts as you go.

I've learnt from bitter experience always to use `git rebase -i master` and check that only the commits I expect are listed.  Git _sometimes_ pick up other changes in history which can make a complete pickle of the resultant merge.

ann0see:
I think the only outstanding issue are the conflicts which I am not 100% comfortable to resolve.

softins:
> I think the only outstanding issue are the conflicts which I am not 100% comfortable to resolve.

OK, since I did the IPv6 changes, I'll resolve the conflicts and push back to the branch.

softins:
I have rebased this PR to the current master and force-pushed. It compiles fine on Ubuntu, and we'll see the CI results.

Please could someone try compiling it for iOS and running?

ann0see (approved):
Could compile and run it on my iPhone SE (1st gen); didn't test interface yet and the "switch microphone" option doesn't seem to be there anymore, but all in all, I'm ok with this state.

ann0see:
> rebase and squash, plus confirmation it still works as expected

done by @softins tested by @ann0see 

> an open issue to address the outstanding review comments in future releases (not necessarily "in the next version")

I'm not up to date with this PR. I think the following should be done in future:
1. Fix the speed issue with multithreading (Jamulus is quite slow on my iPhone SE). Todo: Research how to do multi threading with the sound handling on iOS and also other OS; maybe worth a separate issue
2. Make GUI more mobile friendly; Todo: Research and test a little bit with QML or work on each GUI item one by one and make it more touch friendly
3. Anything else by @pljones can be added to an issue

@ngocdh could you please open an issue on that?

> a clear notice that it's experimental

A change log entry would be sufficient, I think.

ios/sound.mm
pljones: OK, we're leaving this as is - we'll work out what to do when clang-format gets involved...

ios/sound.mm
pljones: Still outstanding.

ios/sound.mm
pljones: Seems to be still outstanding.

ios/sound.mm
pljones: And another.  Did a commit get missed?

src/socket.cpp
```diff
@@ -241,57 +266,73 @@ void CSocket::SendPacket ( const CVector<uint8_t>& vecbySendBuf, const CHostAddr
         // char vector in "const char*", for this we first convert the const
         // uint8_t vector in a read/write uint8_t vector and then do the cast to
         // const char *)
-        if ( HostAddr.InetAddr.protocol() == QAbstractSocket::IPv4Protocol )
+
+        for ( int tries = 0; tries < 2; tries++ ) // retry loop in case send fails on iOS
```
ann0see, line 270: Wouldn't an ifdef for iOS be better here (if it's only needed on iOS)?

pljones:
There's a few comments a made that it looked like were intended to have been fixed -- but they don't appear to have been.  I'm not sure if in the rebasing some commit got lost or what.

ann0see:
Hmm. Maybe. But the comments should still be there.

ngocdh:
I haven't done anything about your last batch of comments @pljones, was waiting for more :) I'm on it.
@softins thank you for the rebase.
@ann0see you can change device in Settings, just like in other OSes.

ann0see:
Thanks. Just saw it.

src/socket.cpp
ngocdh: It's @softins code. I don't think iOS needs it though, but it might be a good idea to leave the loop there for all OSes, just as a safety mesure.

ngocdh:
> Please could someone try compiling it for iOS and running?

My Mac is having trouble synchronizing with test devices. I sadly can't do anything right now.

ios/sound.mm
ngocdh: Ok I had a look here. Actually iOS AudioBuffer uses float, so I had to work with float. Then when sending/receiving data to/from CSound, conversion to/from integer is done. 
We just can't choose to use integer only.

ngocdh:
Ok, so I check @pljones comments and made two (really) minor changes accordingly: a typo (steoreo -> stereo) and and a break in a loop. The third comment couldn't be taken into account because iOS works with float, not integer, so conversion is required.

Now if you're all ok with the above, I will send my last commit (hopefully) for this PR.

src/socket.cpp
emlynmac: I thought the whole point of using UDP was that it was lossy. How did a retry mechanism end up being needed here?

ann0see:
I‘d be ok with that

src/socket.cpp
ngocdh: @emlynmac in iOS when you lock the device, all inactive sockets are closed. When device is unlocked, and user tries to connect to a server for example, the app crashes because of the closed socket. 
As far as I understand, the status check in the loop only concerns the socket (because it can not know if packet reached its destination), if socket is closed, then status < 0, that's when we need to reinit the socket and send the packet again.

softins (approved):

ann0see:
@softins thanks.

Before merging, I‘d like to have an approval by @pljones too.

src/socket.cpp
gene96817: @emlynmac  Application level reliability is common for ensuring a message gets through. However, for audio/video stream, retry/retransmission is not used.

emlynmac: > @emlynmac Application level reliability is common for ensuring a message gets through. However, for audio/video stream, retry/retransmission is not used.

This is handled elsewhere for messages that need it.

@ngocdh Thanks - I see.

ann0see:
@pljones or @ngocdh could you please open an issue on the outstanding tasks?

ngocdh:
> @pljones or @ngocdh could you please open an issue on the outstanding tasks?
> 

My outstanding issue is already described in #1727 https://github.com/jamulussoftware/jamulus/discussions/1727

pljones:
I don't think I can review the code here.  It's not following the way the rest of Jamulus works, it's Apple-specific.  So it needs someone who knows the platform to review it.

ann0see (approved):
@emlynmac can you Review and approve this?

emlynmac:
> @emlynmac can you Review and approve this?

I don't have any means to test this code; the fact that QT is so old coupled with the version of Xcode required means you cannot build on Big Sur.
I don't have any issues with the code, but not being able to test it means I can't really verify it either.

pljones:
OK, can we mark the build on Apple Store as experimental?  Or does that break Apple Store distribution?

emlynmac:
> OK, can we mark the build on Apple Store as experimental? Or does that break Apple Store distribution?

There will be no apple store distribution - https://github.com/jamulussoftware/jamulus/pull/1874 needs resolving.
You may not publish a beta to the App Store also. Testflight is the means for distribution beta versions.

ngocdh:
> so that's a no-go until we push up to more recent versions of QT / Xcode that work on Big Sur.

I'm confused: you can't build because you must maintain the same environment as Jamulus autobuild?

emlynmac:
> > so that's a no-go until we push up to more recent versions of QT / Xcode that work on Big Sur.
> 
> I'm confused: you can't build because you must maintain the same environment as Jamulus autobuild?

Well, it _should_ always be the same environment as the official build yes. But, additionally, the QT5 build environment is not functional with big sur. I used to have an old laptop with Catalina on it, but that is no more...

ann0see:
Hmm. I‘d say since building for IOS worked and the CI is green, this should be OK to merge.

PR #1884, opened 2021-06-16
Moved Custom Directory Server combobox from ClientSettingsDlg toConnectDlg window (discussion #1869)

dcorson-ticino-com:
Moved Custom Directory Server combobox from ClientSettingsDlg toConnectDlg window (discussion #1869)

Changelog: Moved Custom Directory Server combobox from ClientSettingsDlg to ConnectDlg window (discussion #1869)

- changing the Custom server automagically selects Custom server type
- connect button same size as on ClientDlg
- the possibility to delete an entry in the combobox now works again

CustomDir

CustomDir2

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -8,8 +8,11 @@
   (contributed by @jujudusud, @henkdegroot)
 
 - Server: A single click on the server application in the systemtray now opens the Jamulus Server main window (#1722).
-  (contrbuted by @henkdegroot)
-  
+  (contributed by @henkdegroot)
+
+- Bug Fix: Mute myself has been made more consistent (#1838).
+  (contributed by @ngocdh)
+
 ### 3.8.0 (2021-06-02) ###
 
 - The term "Central server" has been replaced with "Directory server" (#1407, #1715, #1629).
--- a/autobuild/linux/autobuild_deb_1_prepare.sh
+++ b/autobuild/linux/autobuild_deb_1_prepare.sh
@@ -13,4 +13,4 @@ echo "Update system..."
 sudo apt-get -qq update
 
 echo "Install dependencies..."
-sudo apt-get -qq -y install devscripts build-essential debhelper libjack-jackd2-dev qtbase5-dev qttools5-dev-tools
+sudo apt-get -qq -y install devscripts build-essential debhelper libjack-jackd2-dev qt6-qtbase-dev qt6-qtbase-dev-tools
--- a/autobuild/linux/autobuild_deb_1_prepare.sh
+++ b/autobuild/linux/autobuild_deb_1_prepare.sh
@@ -13,4 +13,4 @@ echo "Update system..."
 sudo apt-get -qq update
 
 echo "Install dependencies..."
-sudo apt-get -qq -y install devscripts build-essential debhelper libjack-jackd2-dev qt6-qtbase-dev qt6-qtbase-dev-tools
+sudo apt-get -qq -y install devscripts build-essential debhelper libjack-jackd2-dev qt5-qtbase-dev qt5-qtbase-dev-tools
--- a/autobuild/linux/autobuild_deb_1_prepare.sh
+++ b/autobuild/linux/autobuild_deb_1_prepare.sh
@@ -13,4 +13,4 @@ echo "Update system..."
 sudo apt-get -qq update
 
 echo "Install dependencies..."
-sudo apt-get -qq -y install devscripts build-essential debhelper libjack-jackd2-dev qt5-qtbase-dev qt5-qtbase-dev-tools
+sudo apt-get -qq -y install devscripts build-essential debhelper libjack-jackd2-dev qtbase5-dev qttools5-dev-tools
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -493,11 +493,6 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     QObject::connect ( &ClientSettingsDlg, &CClientSettingsDlg::AudioChannelsChanged, this, &CClientDlg::OnAudioChannelsChanged );
 
-    QObject::connect ( &ClientSettingsDlg,
-                       &CClientSettingsDlg::CustomCentralServerAddrChanged,
-                       &ConnectDlg,
-                       &CConnectDlg::OnCustomCentralServerAddrChanged );
-
     QObject::connect ( &ClientSettingsDlg, &CClientSettingsDlg::NumMixerPanelRowsChanged, this, &CClientDlg::OnNumMixerPanelRowsChanged );
 
     QObject::connect ( this, &CClientDlg::SendTabChange, &ClientSettingsDlg, &CClientSettingsDlg::OnMakeTabChange );
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -664,7 +664,6 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
 void CClientSettingsDlg::showEvent ( QShowEvent* )
 {
     UpdateDisplay();
-    //    UpdateCustomCentralServerComboBox();
 
     // set the name
     pedtAlias->setText ( pClient->ChannelInfo.strName );
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -664,7 +664,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
 void CClientSettingsDlg::showEvent ( QShowEvent* )
 {
     UpdateDisplay();
-//    UpdateCustomCentralServerComboBox();
+    //    UpdateCustomCentralServerComboBox();
 
     // set the name
     pedtAlias->setText ( pClient->ChannelInfo.strName );
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -325,15 +325,6 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     cbxInputBoost->setWhatsThis ( strInputBoost );
     cbxInputBoost->setAccessibleName ( tr ( "Input Boost combo box" ) );
 
-    // custom directory server address
-    QString strCentrServAddr = "<b>" + tr ( "Custom Directory Server Address" ) + ":</b> " +
-                               tr ( "Leave this blank unless you need to enter the address of a directory "
-                                    "server other than the default." );
-
-    lblCentralServerAddress->setWhatsThis ( strCentrServAddr );
-    cbxCentralServerAddress->setWhatsThis ( strCentrServAddr );
-    cbxCentralServerAddress->setAccessibleName ( tr ( "Directory server address combo box" ) );
-
     // current connection status parameter
     QString strConnStats = tr ( "Audio Upstream Rate depends on the current audio packet size and "
                                 "compression setting. Make sure that the upstream rate is not "
@@ -406,11 +397,6 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
 
     // language combo box (corrects the setting if language not found)
     cbxLanguage->Init ( pSettings->strLanguage );
-
-    // init custom directory server address combo box (max MAX_NUM_SERVER_ADDR_ITEMS entries)
-    cbxCentralServerAddress->setMaxCount ( MAX_NUM_SERVER_ADDR_ITEMS );
-    cbxCentralServerAddress->setInsertPolicy ( QComboBox::NoInsert );
-
     // update new client fader level edit box
     edtNewClientLevel->setText ( QString::number ( pSettings->iNewClientFaderLevel ) );
 
@@ -624,16 +610,6 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                        this,
                        &CClientSettingsDlg::OnGUIDesignActivated );
 
-    QObject::connect ( cbxCentralServerAddress->lineEdit(),
-                       &QLineEdit::editingFinished,
-                       this,
-                       &CClientSettingsDlg::OnCentralServerAddressEditingFinished );
-
-    QObject::connect ( cbxCentralServerAddress,
-                       static_cast<void ( QComboBox::* ) ( int )> ( &QComboBox::activated ),
-                       this,
-                       &CClientSettingsDlg::OnCentralServerAddressEditingFinished );
-
     QObject::connect ( cbxLanguage, &CLanguageComboBox::LanguageChanged, this, &CClientSettingsDlg::OnLanguageChanged );
 
     QObject::connect ( cbxInputBoost,
@@ -688,7 +664,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
 void CClientSettingsDlg::showEvent ( QShowEvent* )
 {
     UpdateDisplay();
-    UpdateCustomCentralServerComboBox();
+//    UpdateCustomCentralServerComboBox();
 
     // set the name
     pedtAlias->setText ( pClient->ChannelInfo.strName );
@@ -931,26 +907,6 @@ void CClientSettingsDlg::OnEnableOPUS64StateChanged ( int value )
 
 void CClientSettingsDlg::OnFeedbackDetectionChanged ( int value ) { pSettings->bEnableFeedbackDetection = value == Qt::Checked; }
 
-void CClientSettingsDlg::OnCentralServerAddressEditingFinished()
-{
-    // if the user has selected and deleted an entry in the combo box list,
-    // we delete the corresponding entry in the directory server address vector
-    if ( cbxCentralServerAddress->currentText().isEmpty() && cbxCentralServerAddress->currentData().isValid() )
-    {
-        pSettings->vstrCentralServerAddress[cbxCentralServerAddress->currentData().toInt()] = "";
-    }
-    else
-    {
-        // store new address at the top of the list, if the list was already
-        // full, the last element is thrown out
-        pSettings->vstrCentralServerAddress.StringFiFoWithCompare ( NetworkUtil::FixAddress ( cbxCentralServerAddress->currentText() ) );
-    }
-
-    // update combo box list and inform connect dialog about the new address
-    UpdateCustomCentralServerComboBox();
-    emit CustomCentralServerAddrChanged();
-}
-
 void CClientSettingsDlg::OnSndCrdBufferDelayButtonGroupClicked ( QAbstractButton* button )
 {
     if ( button == rbtBufferDelayPreferred )
@@ -989,21 +945,6 @@ void CClientSettingsDlg::UpdateDisplay()
         // clear text labels with client parameters
         lblUpstreamValue->setText ( "---" );
         lblUpstreamUnit->setText ( "" );
-    }
-}
-
-void CClientSettingsDlg::UpdateCustomCentralServerComboBox()
-{
-    cbxCentralServerAddress->clear();
-    cbxCentralServerAddress->clearEditText();
-
-    for ( int iLEIdx = 0; iLEIdx < MAX_NUM_SERVER_ADDR_ITEMS; iLEIdx++ )
-    {
-        if ( !pSettings->vstrCentralServerAddress[iLEIdx].isEmpty() )
-        {
-            // store the index as user data to the combo box item, too
-            cbxCentralServerAddress->addItem ( pSettings->vstrCentralServerAddress[iLEIdx], iLEIdx );
-        }
     }
 }
 
--- a/src/clientsettingsdlg.h
+++ b/src/clientsettingsdlg.h
@@ -83,7 +83,6 @@ public slots:
     void OnAutoJitBufStateChanged ( int value );
     void OnEnableOPUS64StateChanged ( int value );
     void OnFeedbackDetectionChanged ( int value );
-    void OnCentralServerAddressEditingFinished();
     void OnNewClientLevelEditingFinished() { pSettings->iNewClientFaderLevel = edtNewClientLevel->text().toInt(); }
     void OnInputBoostChanged();
     void OnSndCrdBufferDelayButtonGroupClicked ( QAbstractButton* button );
@@ -109,6 +108,5 @@ public slots:
 signals:
     void GUIDesignChanged();
     void AudioChannelsChanged();
-    void CustomCentralServerAddrChanged();
     void NumMixerPanelRowsChanged ( int value );
 };
--- a/src/clientsettingsdlgbase.ui
+++ b/src/clientsettingsdlgbase.ui
@@ -30,7 +30,7 @@
       </sizepolicy>
      </property>
      <property name="currentIndex">
-      <number>1</number>
+      <number>2</number>
      </property>
      <property name="usesScrollButtons">
       <bool>true</bool>
@@ -957,33 +957,6 @@
               </spacer>
              </item>
              <item>
-              <widget class="QLabel" name="lblCentralServerAddress">
-               <property name="text">
-                <string>Custom Directory Server Address:</string>
-               </property>
-              </widget>
-             </item>
-             <item>
-              <widget class="QComboBox" name="cbxCentralServerAddress">
-               <property name="editable">
-                <bool>true</bool>
-               </property>
-              </widget>
-             </item>
-             <item>
-              <spacer name="verticalSpacer_4">
-               <property name="orientation">
-                <enum>Qt::Vertical</enum>
-               </property>
-               <property name="sizeHint" stdset="0">
-                <size>
-                 <width>20</width>
-                 <height>13</height>
-                </size>
-               </property>
-              </spacer>
-             </item>
-             <item>
               <widget class="QLabel" name="lblNewClientLevel">
                <property name="text">
                 <string>New Client Level</string>
@@ -1287,7 +1260,6 @@
   <tabstop>sldNetBuf</tabstop>
   <tabstop>sldNetBufServer</tabstop>
   <tabstop>chbEnableOPUS64</tabstop>
-  <tabstop>cbxCentralServerAddress</tabstop>
   <tabstop>edtNewClientLevel</tabstop>
   <tabstop>cbxInputBoost</tabstop>
   <tabstop>chbDetectFeedback</tabstop>
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -96,8 +96,8 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
 
     // custom directory server address
     QString strDirServAddr = "<b>" + tr ( "Custom Directory Server Address" ) + ":</b> " +
-                               tr ( "Leave this blank unless you need to enter the address of a directory "
-                                    "server other than the default." );
+                             tr ( "Leave this blank unless you need to enter the address of a directory "
+                                  "server other than the default." );
 
     lblCustomDirectoryAddress->setWhatsThis ( strDirServAddr );
     cbxCustomDirectoryAddress->setWhatsThis ( strDirServAddr );
@@ -159,8 +159,7 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     UpdateCustomDirectoryServerComboBox();
 
     // install event handler to catch return keys in the combobox edit line
-    cbxCustomDirectoryAddress->installEventFilter(this);
-
+    cbxCustomDirectoryAddress->installEventFilter ( this );
 
     // setup timers
     TimerInitialSort.setSingleShot ( true ); // only once after list request
@@ -188,20 +187,14 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
                        this,
                        &CConnectDlg::OnCentServAddrTypeChanged );
 
-    QObject::connect ( cbxCustomDirectoryAddress,
-                       &QComboBox::textActivated,
-                       this,
-                       &CConnectDlg::OnCustomDirectoryTypeChanged );
+    QObject::connect ( cbxCustomDirectoryAddress, &QComboBox::textActivated, this, &CConnectDlg::OnCustomDirectoryTypeChanged );
 
     QObject::connect ( cbxCustomDirectoryAddress->lineEdit(),
                        &QLineEdit::editingFinished,
                        this,
                        &CConnectDlg::OnDirectoryServerAddressEditingFinished );
 
-    QObject::connect ( cbxCustomDirectoryAddress->lineEdit(),
-                       &QLineEdit::returnPressed,
-                       this,
-                       &CConnectDlg::NoReturnKeypress );
+    QObject::connect ( cbxCustomDirectoryAddress->lineEdit(), &QLineEdit::returnPressed, this, &CConnectDlg::NoReturnKeypress );
 
     // check boxes
     QObject::connect ( chbExpandAll, &QCheckBox::stateChanged, this, &CConnectDlg::OnExpandAllStateChanged );
@@ -217,16 +210,19 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     QObject::connect ( &TimerReRequestServList, &QTimer::timeout, this, &CConnectDlg::OnTimerReRequestServList );
 }
 
-bool CConnectDlg::eventFilter(QObject *object, QEvent *event)
+bool CConnectDlg::eventFilter ( QObject* object, QEvent* event )
 {
     // event handler to catch return keys in the combobox and call EditingFinished
     //   otherwise the connect button would also be be called
-    if (object == cbxCustomDirectoryAddress && event->type() == QEvent::KeyPress) {
-        QKeyEvent *keyEvent = static_cast<QKeyEvent *>(event);
-        if ( (keyEvent->key()==Qt::Key_Enter) || (keyEvent->key()==Qt::Key_Return) )  {
+    if ( object == cbxCustomDirectoryAddress && event->type() == QEvent::KeyPress )
+    {
+        QKeyEvent* keyEvent = static_cast<QKeyEvent*> ( event );
+        if ( ( keyEvent->key() == Qt::Key_Enter ) || ( keyEvent->key() == Qt::Key_Return ) )
+        {
             OnDirectoryServerAddressEditingFinished();
             return true;
-        } else
+        }
+        else
             return false;
     }
     return false;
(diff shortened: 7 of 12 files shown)


dcorson-ticino-com:
Interesting that even after doing a Clang-format locally and pushing it the style check fails.

pljones:
Could you put screenshots of the result in the PR, please.

dcorson-ticino-com:
Is it possible to remove earlier commits without removing later ones.
I see now that the autobuild problems come from the first 4 commits.  
They should not be in this branch, they were for the Qt6 stuff.

ann0see:
git rebase maybe?

ann0see:
Btw: why do we even need a cancel button in the connect dialog?

pljones:
Not sure I like the change, over all.  That could _well_ be because I'm very used to how things work at the moment.

I don't mind the combo box also being available from the connect dialogue, per se.  But that's about it.  I'd almost say it should be disabled on the connect dialogue unless the list selected is custom.

It should remain available through settings and the two would then need to stay sync'd.

I don't like the changed button size.  I don't like the large button size on the main window, for that matter - it looks... "D'uh - you're too stupid for menus..." tone.  Which isn't nice.

The connect dialogue needs to be easily closeable.  Leaving it open can have a serious adverse effect on your jamming and on the other people - every one of the servers in the list gets sent a ping ever couple of seconds by every client that's viewing the list, using up lots of bandwidth and introducing jitter.

dcorson-ticino-com:
I implemented that which seemed to me to be the result of the discussion.
Of course as you described is possible, if we should decide that would be better I have no problem to change it. 
On the other hand I am not a fan of having multiple places to define one thing; either in the Connect window or the Setup window.
As to the button, you may see that as a silent insult, but I have seen too many users completely stymied because that had to use a menu to set their user info.  I don't think it is out of place to make the most important button somewhat larger.

dcorson-ticino-com:
There is apparently a difference between the clang-format defs used by the autobuild and those in the repo or I am missing something.
Who can help ?

pljones:
> Who can help ?

It looks weird...  Try just patching it the way it's asking for...

dcorson-ticino-com:
After this autobuild didn't work because of some GitHub problem I forced it to run manually. 
It was successful, but GitHub is apparently not able to show that here......

pljones:
I've tried giving it another kick.

dcorson-ticino-com:
Hmmm...
Your kick seems to be more convincing.

pljones:
Getting the kick made seemed... more complicated... than previously.

gilgongo:
The interaction on this is somewhat unclear when the custom directory address is still showing when not in use. Might only showing it conditonally be a clearer?

Also - a minor point, but the absence of an explicit add/submit button when entering the first directory server address might also be a little confusing. Will everyone doesn't realise they have to go back to their mouse to click out, or hit return?

https://user-images.githubusercontent.com/4561747/122640142-b9574000-d0f5-11eb-83d1-0cb0885d60b3.mp4

dcorson-ticino-com:
Hi Jonathan,
Is that a mock-up that you made?
My code doesn't look like that or work like that.

If we make it as plj suggested, which I think is a good idea actually, just not that which I understood from the discussion, I would inactivate, gray out, and not remove from the screen.
Other than that, the edit line works as it had, except that deleting an entry by erasing it, which was always in the code but for some reason didn't work, is now working again.

gilgongo:
@dcorson-ticino-com Yes, it's a mockup from screenshots. 

> My code doesn't look like that or work like that.

I know :-) I'm suggesting a change to it. 

> I would inactivate, gray out, and not remove from the screen.

Can you explain why? I have explained why I suggest hiding it.  

> what is your take on the bigger connect button?

I don't have an opinion on it yet, because I don't know what the reason for making it bigger is.

dcorson-ticino-com:
> Can you explain why? I have explained why I suggest hiding it.

Because if I tell someone to set the custom directory in the connect window and they don't see anything at all about a directory, it will be frustrating. 
I think we could also call the directory combobox "Directory" and not list as we never talk about a "List" anywhere that I am aware of.
Maybe the directory combobox and the custom directory combobox should also be on the same line.

My reason to make the button bigger was just to have visual similarity to the clientdlg screen.

pljones:
> Maybe the directory combobox and the custom directory combobox should also be on the same line.

I actually try to keep my connect dialog fairly small.  Putting too many things onto it makes it less useful as I see fewer servers.  I do use the filter, so I don't want that to go.  But I keep the over width only just wide enough for the filter to be useful.  So there's no room for me on the same line.

I'd really prefer not to lose any more vertical space.  That's one of the reasons I dislike the larger button - it loses two servers off the list.  Being in London, there are a _lot_ of _busy_ servers local to me, so I need vertical space.  Adding the custom directory combo box is going to eat up space that, most of the time, I won't be interested it.  It's very rare for me to need to _change_ the value - that's why it got onto the _Advanced_ Setup tab.

For people who are constantly jumping between different custom directories, they can leave the settings dialogue open on the Advanced Setup tab pretty easily.  (Yes, that takes up more space...)  @gilgongo's idea of the combo only appearing when "Custom" is selected might also be okay...  (The the UI needs to be better than the mock up... label the combo, have "set" and "delete" buttons rather than "magic" ways of interaction...)

pljones:
The other way might be to put it at the bottom, something like:

```
|                                                   |
+---------------------------------------------------+
Server address:   [                    ]    [Connect]
Custom directory: [                    ]     [Cancel]
```

gilgongo:
The trouble is that the vast majority of Jamulus users have no need to use or undertand a custom directory address. It's always a bit of a slipperly slope adding things to main UI that are only aimed at a majority in case it confuses (or inconveniences, as @pljones implies) the majority. What's in this PR seems a bit of an inforamation heirachy issue to me.

I guess "Directory" can replace "List" (which was originally intended to be verb the same way as "Filter" but maybe that's inconsistent style). So in my suggested UI, you can tell people to "Select custom from the directory menu" and the field will appear. 

@pljones 

> (The the UI needs to be better than the mock up... label the combo, have "set" and "delete" buttons rather than "magic" ways of interaction...)

Yes, this is what I was hinting at when I said the absence of an explicit add/submit button when entering the first directory server address might also be a little confusing. Glad you agree :-)

gilgongo:
Putting it all at the bottom might work, although the "Choose your genre..." feel to the UI would be a bit upturned by that.

pljones:
> Putting it all at the bottom might work I guess as it would be more ignorable by the majority.

It's also sort of comparable with the existing "Server Address" field for accessing private servers.

ann0see:
Yes, but it will lead to confusion for private server users: Do I need to input this number into "directory" or "server"?

dcorson-ticino-com:
My first attempt at changing the dlg had the custom directory combobox at the bottom, but I didn't like it and moved it to the top.
If, however, our goal is not only to put this control in this dialog, but also to conserve window surface then I would do the following:  custom directory at the bottom, show all musicians also at the bottom, eliminate the cancel button like all the other menus close with the top-right X, retaining, of course, the fact that the window closes at connection.
There is unused space at the bottom and the filter field could be much larger that way.
As the existing PR is not liked I will make a new suggestion in this direction.

ann0see:
Although it's unrelated, we might want to change the list selection to be editable. 

You'd have a pencil button right next to the "List" selection box which allows you to add multiple custom directories.

gilgongo:
@pljones 

Having a play with that idea, but the trouble with putting them at the bottom is that the "connect" button then becomes ambigous, and we don't want two connect buttons:

image

I think a conditional reveal at the top is the way to go because the in interaction establishes the connection between the word "custom" in the combo and the address you need to enter. The vast majority of people could then ignore it. 

image

gilgongo:
@ann0see 

> You'd have a pencil button right next to the "List" selection box which allows you to add multiple custom directories.

I wondered about that too but wasn't sure if that would get tricky in the UI. What would happen when you hit the pencil?

@dcorson-ticino-com 

> If, however, our goal is not only to put this control in this dialog, but also to conserve window surface then I would do the following:

Can you sketch out what you mean?

dcorson-ticino-com:
> Although it's unrelated, we might want to change the list selection to be editable.
> You'd have a pencil button right next to the "List" selection box which allows you to add multiple custom directories.

That would be cleanest in some ways putting all directory servers at the same level, but would change all the logic.  
But why add a pencil, it could be edited as are the Server address and Custom Directory comboboxes right now?

dcorson-ticino-com:
> Having a play with that idea, but the trouble with putting them at the bottom is that the "connect" button then becomes ambigous, and we don't want two connect buttons:

This is already the case, if a server in the list is not selected the server address entry is used.
Nothing changes for as we know it now.

dcorson-ticino-com:
New push

CustomDir4
CustomDir3

gilgongo:
The "connect" button would then be ambigous because the visual connection between "Custom" in the combo list and "Custom directory" at the bottom would be lost.

dcorson-ticino-com:
For some reason the Clang-format definitions used by the autobuild are not the same as what I got from the repo.
Have they been changed in the last couple of days ?

ann0see:
> I don't think the reason for continusously exposing the input field in the UI is strong enough to justify this change.

Me too. I would not like to have this constantly at the bottom (too much confusion possible)

dcorson-ticino-com:
Another push, but this makes use without reading the directions more difficult IMHO

CustomDir5
CustomDir6

dcorson-ticino-com:
Maybe hide the Server Address if Custom Directory is selected?
Then we have either Server Address or Custom Directory, but never both.

gilgongo:
The problem I think is the desire to show the custom directory input to everyone, coupled with putting it at the bottom, since that separates the directory list control from its effect in the UI.

dcorson-ticino-com:
Custom Directoy moved to top

CustomDir7
CustomDir8

gilgongo:
That build feels clearer on the interaction, since the action of selecting "Custom" is then explained by the appearance for the Directory address drop-down. 

Given the label has been changed from "List" to "Directory" and that association with the "Custom" is established using the interaction, would labelling the drop-down simply "Address" be clearer? 

I do agree with @pljones that the somewhat hidden (and rather fiddly - I couldn't get it to work at first) interaction for actually setting the address value may be confusing. I'd put a button in there that said "Add" perhaps. Deleting is less of a problem because there is a fairly common convention of back-spacing over the value. Assuming we can't adopt the even more common one of a "delete" cross next to the selected value, that is.

ann0see:
@gilgongo 

> What would happen when you hit the pencil?

It would open a separate window where you can edit the addresses of all directories.

pljones:
1) I quite like the dropdown at the top, visible only when needed.
2) "[x] Show All Musicians" should continue to be right-edge aligned; the filter then pads the full remaining width on the top row.
3) Then the "Custom" box appears -- I think it should be labelled "Custom Directory" because changing to "Custom Address" is yet another change for people to absorb unnecessarily.  And it enforces the difference from "Server Address".
4) If we're going to lose the "Cancel" button, the window needs to respond to Escape.  That behaviour has been _explicitly_ removed for this window, currently.  I've just not had a good enough reason to make it do so - but this is one.  In fact, without making it do so, you don't get to remove "Cancel" 😄.
5) I'm not comfortable with the "pencil" ✏️ idea...  I guess it's a little close to the Favourites list idea at the moment, until that's resolved.  I just don't want to see a _third_ related idea...

dcorson-ticino-com:
To be sure I understand you. 
The extra space for the Filter you mentioned above is no longer needed?
I move everything back to how it was before in the first line?

I guess the settings window should get it's escape back too as it also doesn't have a cancel button.

pljones:
> The extra space for the Filter you mentioned above is no longer needed?
> I move everything back to how it was before in the first line?

No.

I'm saying I like this **for the top two lines**
CustomDir7CustomDir8
**so long as** when the window gets wider, the filter gets wider and the Show All Musicians checkbox stays right-aligned.

**HOWEVER* the "Cancel" button has gone and that's made me less happy.  We don't want people to feel trapped.  So I'd actually prefer to keep the "Cancel" button.  It should be the default action.  It should be the bottom right button.  Enter should, by default, close the window without doing anything (i.e. shouldn't change the server you're currently on).  Escape should also, always, close the window without doing anything - currently Escape is ignored.

"Connect" then requires explicit action by the user (so should have an accelerator).  "Connect" connects to the focussed server.  Currently, the default server is the one in "Server Address" and is changed by selecting one from the list.  Double-click or enter on an list entry also connects to that server, bypassing the "Connect" button.

dcorson-ticino-com:
OK, I'll do that.  (I see you added to your description as I write this, most further questions are now clear)
The code for return behaviour will be simple (I think), but opaque.
The Connect button already has an accelerator, Alt-o

What about escape for the clientsettingsdlg as it has no cancel button?

pljones:
> What about escape for the clientsettingsdlg as it has no cancel button?

It _should_ have a "Close" button (not cancel - the changes are made on the fly) and Escape shouldn't work (as it could feel like "cancel").

gilgongo:
> It would open a separate window where you can edit the addresses of all directories:

A separate window in what already is a separate window seems a large bit of UI for something that already has a reasonably common convention: select the list item you want to clear, and either backspace over the selected value, or have a "clear" icon inside the dropdown.

pljones:
It's not "clear selection".  It's "delete from list".  Those two actions need to be kept distinct.  Your animation conflates them uncomfortably for me.

gilgongo:
It's merely a convention adopted by a number of applications. There is a less used one that has delete icons on each item in the drop-down list instead, which may be clearer but I don't know if that's easier or harder to do. This is an operation that's hardly ever going to be performed, after all.

pljones:
> This is an operation that's hardly ever going to be performed, after all.

Agreed.

gilgongo:
> Agreed. And I wouldn't want to do it accidentally.

So perhaps not have the delete icon and require the user to back-space over the value. The alternative, as I say, is to blow this up into a much bigger "management" UI. But I think it's hard to justify that really and also implies we'd need that for both directories as well as server combo-boxes.

jp8:
When I originally suggested the change to Volker, I imagined that the entire concept of a 'Custom' central server could disappear.  The custom ones are no different from the non-custom ones, and could, in my view, be listed together in the same drop-down.

And at the bottom a single server entry field and a single 'connect' button.  If the user enters a server that happens to be a central server, then clicking connect would cause it to be added to the list, with a dialog informing the user that a new 'list' has been added.

jp8:
And yes... somewhere entirely different, in the "settings" pages, could be the functionality that lets you clean up the list of central servers, including:
- removing existing central servers from the list (even removing the default genre ones)
- adding central servers to the list
- re-ordering the list ?
- setting a parental control password so that children cannot just add any central server they happen to hear about on social media

houbsta:
I'm finding it hard to follow, but am I seeing the box with currently selected changing to greyed out (disabled state) when a directory is selected? This is confusing.

pljones:
@jp8 I tend to prefer that, too -- one combo box, with the non-custom servers fixed and uneditable, but if you stick a new value in, it's remembered for good.  You drop the list down, it's grown by that new entry.

If you want to manage it, that's in settings - in place of the existing custom directory server field, you'd have the "management interface" where the additional custom entries could be removed (or new ones added).  I see _that_ as where you'd select one of the custom entries to make it the one you're currently maintaining (doesn't affect the connect dialogue) and you can choose to delete it if you want to.  I'd also like the "management interface" to show things like whether that address returned any servers when you added it - making it easier to delete the typos.  Of course, this is definitely entering the "there's stuff here to discuss" realm...

gilgongo:
@jp8 jp8 

> That is the current user interface for anyone who uses two or more custom central servers.

That is of course true. But it is **also** true that the proportion of Jamulus users who are in that position is tiny. There is a rule of thumb in interface design: don't inconvenience the majority for the sake of the minority. That's not an argument to ignore the minority, but it is an argument to take on certain design principles. In this case, making the central server menu accessible but also ignorable. This is (if you ignore the side-issue of the "clear" icon for now) achieved in @dcorson-ticino-com latest build.

As to the issue of combining the concepts of directories and servers: again, if that can be done (as you seem to imply) "invisibly" to the majority of people who just want to connect to a private server and neither know nor care about directories, then removing the destinction **might** be a good move. But it's not for this PR, which is about moving something out of Settings.

jp8:
@gilgongo hi gilgongo,
I tried to think it through carefully to make sure it wouldn't inconvenience anybody (except the developers ;)

If you saw something in my idea that seemed like it would be inconvenient to any users, could you please identify it so I could have a chance to think about it ?

jp8:
After thinking about it, perhaps the issue could be more simply addressed just by changing the setup pages to allow multiple "Custom" central servers to be entered.  They wouldn't be called Custom of course, they would just show up (as their dns name) in the normal list of Directories.

In that case there would be no changes at all needed on the current Connect dialog.

gilgongo:
@jp8 

> If you saw something in my idea that seemed like it would be inconvenient to any users, could you please identify it so I could have a chance to think about it ?

Sorry, can you (re)post a screenshot? I'm not sure what idea you are referring to here.

dcorson-ticino-com:
> After thinking about it, perhaps the issue could be more simply addressed just by changing the setup pages to allow multiple "Custom" central servers to be entered. They wouldn't be called Custom of course, they would just show up (as their dns name) in the normal list of Directories.
> 
> In that case there would be no changes at all needed on the current Connect dialog.

What am I missing ?  That is already possible.  
You can put at the present up to 12 directory servers in the custom directory server combobox.

gilgongo:
@dcorson-ticino-com I'm assuming it's referring to a management UI to allow deletion and editing.  But I would not be supportive of this becasue a) I think your idea satisfctorily solves the problem as originally stated, and b) as I've shown, there are other ways of editing combo lists that do not involve large amounts of UI invasion.

jp8:
> > After thinking about it, perhaps the issue could be more simply addressed just by changing the setup pages to allow multiple "Custom" central servers to be entered. They wouldn't be called Custom of course, they would just show up (as their dns name) in the normal list of Directories.
> > In that case there would be no changes at all needed on the current Connect dialog.
> 
> What am I missing ? That is already possible.
> You can put at the present up to 12 directory servers in the custom directory server combobox.

No, it's really not already possible.  The original issue is this:  to change to a different custom central server, it is currently necessary to bring up the setup dialog.  it should be possible to change to multiple custom central servers directly from the connect dialog.

jp8:
> 
> Sorry, can you (re)post a screenshot? I'm not sure what idea you are referring to here.
> 

I wish I could.  The description I wrote earlier is:

> When I originally suggested the change to Volker, I imagined that the entire concept of a 'Custom' central server could disappear. The custom ones are no different from the non-custom ones, and could, in my view, be listed together in the same drop-down.
> 
> And at the bottom a single server entry field and a single 'connect' button. If the user enters a server that happens to be a central server, then clicking connect would cause it to be added to the list, with a dialog informing the user that a new 'Directory' has been added.

gilgongo:
@jp8 OK, for the purposes of this PR, can we abaondon the issue of merging directories and servers and look at the other issues intead?

dcorson-ticino-com:
> No, it's really not already possible. The original issue is this: to change to a different custom central server, it is currently necessary to bring up the setup dialog. it should be possible to change to multiple custom central servers directly from the connect dialog.

That is now possible with this PR

jp8:
> @jp8 OK, for the purposes of this PR, can we abaondon the issue of merging directories and servers and look at the other issues intead?

Absolutely.  The only thing is, earlier pljones pointed out that it still makes sense for the 'management' part of the list of directories to remain in the setup dialog.

And after thinking about it, I really agree.  And most of the management part already exists - on the setup screen you can already define multiple custom central servers.  The frustration is that they do not appear on the list in the connect dialog.  Just one entry, named 'Custom' which could be anything and you don't even know what it is, without either trying it, or going back to the setup dialog to see what is currently selected there.

So to address the points raised by pljones, I have now suggested:

1.  leave the connect dialog exactly as it is, except instead of "custom" appearing in the list, each of the (up to 12) custom central servers should appear, by their dns name
2  perhaps a way, on the settings dialog, to clean unwanted entries from the list.

jp8:
> > No, it's really not already possible. The original issue is this: to change to a different custom central server, it is currently necessary to bring up the setup dialog. it should be possible to change to multiple custom central servers directly from the connect dialog.
> 
> That is now possible with this PR

Yes sorry that I wasn't clear.  The PR addresses it really nicely and I don't mean to appear ungrateful.  After carefully considering the comments of pljones I am proposing a simplification.

gilgongo:
> The frustration is that they do not appear on the list in the connect dialog. 

vs

> leave the connect dialog exactly as it is, except instead of "custom" appearing in the list, each of the (up to 12) custom central servers should appear, 

The problem originally stated in this PR was that naive users didn't know how to enter the custom directory address because it was in the Settings (quite why something being in the Settings makes it too hard to set was not explained). So the solution was to put it on the main screen instead. We can do that, and we can make it ignorable as per best practice. The current commit shows that.

But your latest reccommendation seems to want to solve a different problem: one that asserts people find it hard to manage the deletion (and editing?) of the list of directories, but have no peoblem with that list being in the Settings.

Am I right in thinking these are two overlapping, but otherwise incompatible, requirements?

dcorson-ticino-com:
> So to address the points raised by pljones, I have now suggested:
> 
>     1. leave the connect dialog exactly as it is, except instead of "custom" appearing in the list, each of the (up to 12) custom central servers should appear, by their dns name
>        2  perhaps a way, on the settings dialog, to clean unwanted entries from the list.

1. What you are suggesting is not just a simple UI change.  I would want to see a broad consensus here before going in this direction.
2. This is already possible in the custom directory list, delete the server name and then press return.  This is standard usage of an editable combobox as is the way of adding an entry.

Note that if the custom directory list is still kept in the Setup menu this would not address the user problems pointed out by Houbsta.

jp8:
The original request was this

 > Hi Volker - would you consider moving this setting to the connection dialog, instead of the settings dialog?
Personally, any time I need to change to a different central server, I am already on the connection dialog.
It could even be integrated with the drop-down that lists the 'fixed' central servers....

gilgongo:
> Note that if the custom directory list is still kept in the Setup menu this would not address the user problems pointed out by Houbsta.

What are those problems? Please link to the quote.

jp8:
>     2. This is already possible in the custom directory list, delete the server name and then press return.  This is standard usage of an editable combobox as is the way of adding an entry.

Yes you are right.  I wasn't aware of that standard way.

> Note that if the custom directory list is still kept in the Setup menu this would not address the user problems pointed out by Houbsta.

Yes it seems to me the views of pljones and houbsta are directly at odds regarding the preferred location of the UI for the management of custom central servers.

For me personally, when it is possible to access multiple custom central servers from the connect dialog, my request is answered.  If I think about what management functionality might reasonably needed in the future, I lean towards the view of pljones.

jp8:
> > Note that if the custom directory list is still kept in the Setup menu this would not address the user problems pointed out by Houbsta.
> 
> What are those problems? Please link to the quote.

https://github.com/jamulussoftware/jamulus/discussions/1869#discussioncomment-865169

jp8:
> > The frustration is that they do not appear on the list in the connect dialog.
> 
> vs
> 
> > leave the connect dialog exactly as it is, except instead of "custom" appearing in the list, each of the (up to 12) custom central servers should appear,

Both of those statements are mine, and (to me) they say exactly the same thing - that I would like to see all custom central servers on the list in the connect dialog.

> 
> The problem originally stated in this PR was that naive users didn't know how to enter the custom directory address because it was in the Settings (quite why something being in the Settings makes it too hard to set was not explained). So the solution was to put it on the main screen instead. We can do that, and we can make it ignorable as per best practice. The current commit shows that.

In fact the problem originally stated was not about naive users - quite the opposite (I hope).  I complained that in order to change to a different custom central server, it was necessary to open up the setup dialog.  I asked if it could be possible to select another custom central server from the connect dialog instead.

> But your latest reccommendation seems to want to solve a different problem: one that asserts people find it hard to manage the deletion (and editing?) of the list of directories, but have no problem with that list being in the Settings.
> 
> Am I right in thinking these are two overlapping, but otherwise incompatible, requirements?

I think the management of the list of custom central servers could be done either on the connect dialog or the setup dialog.  But yes my latest proposal leaves it on the setup dialog.  I don't think it's incompatible with the (my) original requirement.

gilgongo:
OK good.  

So if I undertand correctly, @houbsta @jp8 @dcorson-ticino-com if we keep the setting up of directories in Settings, along with the currently available behaviour (which allows you can add multiples and delete them). We then reflect those settings in the Directory combo list. 

Will that meets all objectives? It's ignorable by the majority of Jamulus users who have no need of it, and sufficently couples the Settings with the Connection window, without overly invading the latter's UI.

pljones:
My _only_ concern with @jp8's suggestion was
> And at the bottom a single server entry field and a single 'connect' button. If the user enters a server that happens to be a central server, then clicking connect would cause it to be added to the list, with a dialog informing the user that a new 'Directory' has been added.

which conflates a server with a directory.  It would be more appropriate to make the current directory drop down a combo box to allow direct entry of new values.  So adding a new entry would just be "Open Connect and paste this address into the 'Directory:' box", rather than "Open Settings, go to Advanced, paste this address into 'Custom Directory Server Address' box, close Settings, open Connect, select Custom in Directory".

The reason I don't want to conflate the servers and directories is that a directory _can be somewhere you jam_ (all of the Any Genre directories are jam servers, too) - so typing one into the Server Address and hitting connect should connect for jamming like it would for any other jam server.  Doing _anything_ else (even additional) is potentially confusing.

Lastly, I don't know if some key combination within the combobox to select an entry and delete it is obvious UX... it's not something I was aware of before, either, and seems somewhat open to accidental triggering...  (Did I say that already?)

dcorson-ticino-com:
I am against having two places to do the same thing.
That just creates confusion.
We decide for one or the other.

I actually think that after much discussion and many changes this PR is good.
My suggestion for the moment would be that we ship this PR with 3.9.0 and I look into making a single list for default and custom servers for the following version.
The directory list could be made editable and the existing editing features used, add and delete.
I can imagine making the custom servers a different color to indicate that they are editable.
There is, however, more work to be done in the code, because the existing code assumes the existing UI elements.

I share plj's concern of conflating servers and directories, I am not even sure there is an easy and sure way of knowing if one is connected to a server or a directory, which would in that case be very necessary.

gilgongo:
> it would be more appropriate to make the current directory drop down a combo box to allow direct entry of new values. 

OK, which is what the latest commit on this PR does as far as I know. 

> I share plj's concern of conflating servers and directories,

I don't understand this point. You enter server addreses at the bottom of the Connect dialogue - if there's a confusion about that then it would exist already. I thought we'd established the idea of conflating was not a thing for now. https://github.com/jamulussoftware/jamulus/pull/1884#issuecomment-866249465

> I can imagine making the custom servers a different color to indicate that they are editable.

Best practice is not to use colour alone to convey meaning. We'd need some other label as well.

gilgongo:
@pljones  

>  I don't know if some key combination within the combobox to select an entry and delete it is obvious UX.

I don't know either. But it exists as a design pattern in various applications today, so I assume some people have encontered it. The alternatives involve a lot more UI. I've not problem with having a management UI for combo box being our UX gift to the world, but it's not for this PR.

dcorson-ticino-com:
> > I share plj's concern of conflating servers and directories,
> 
> I don't understand this point. You enter server addreses at the bottom of the Connect dialogue - if there's a confusion about that then it would exist already. I thought we'd established the idea of conflating was not a thing for now. [#1884 (comment)](https://github.com/jamulussoftware/jamulus/pull/1884#issuecomment-866249465)

If I understood correctly he suggested combining the Server Address and Custom Server combobox into only one input element, but that was an earlier thought and I guess it is now out of date.

jp8:
> 
> The reason I don't want to conflate the servers and directories is that a directory _can be somewhere you jam_ (all of the Any Genre directories are jam servers, too) - so typing one into the Server Address and hitting connect should connect for jamming like it would for any other jam server. Doing _anything_ else (even additional) is potentially confusing.

A good point.  I suggested the single entry box for servers and central servers because it would take up less space on the connect dialog.

But if management of the list of custom central servers is done on the setup dialog, then there is nothing extra taking up space on the connect dialog.

jp8:
> I am against having two places to do the same thing.
> That just creates confusion.
> We decide for one or the other.

The current proposal is - setup page for managing the list of central servers.  The list of central servers would appear on the connect dialog, but would not be editable from the connect dialog.

jp8:
> > I can imagine making the custom servers a different color to indicate that they are editable.
> 
> Best practice is not to use colour alone to convey meaning. We'd need some other label as well.

It's a completely different discussion and PR for the future, but I will at some point raise a request that all central servers be editable.

jp8:
> If I understood correctly he suggested combining the Server Address and Custom Server combobox into only one input element, but that was an earlier thought and I guess it is now out of date.

Indeed - it was an idea to address pljones concern about too much extra UI elements on the connect dialog.

gilgongo:
@jp8 

The **current** proposal is @dcorson-ticino-com latest commit on this PR, which puts it all in the Connect dialogue. In https://github.com/jamulussoftware/jamulus/pull/1884#issuecomment-866280839 I was just trying to make sure I could disentagle distracting conversations about combining directories and servers, and who had said what, and what problems were being addressed. Apologie, I wasn't trying to imply a decision.

> But if management of the list of custom central servers is done on the setup dialog, then there is nothing extra taking up space on the connect dialog.

And if we stick with @dcorson-ticino-com latest commit on this PR, there is **also** nothing taking up extra taking up space on the connect dialogue, until you select the Custom entry.

dcorson-ticino-com:
> > I am against having two places to do the same thing.
> > That just creates confusion.
> > We decide for one or the other.
> 
> The current proposal is - setup page for managing the list of central servers. The list of central servers would appear on the connect dialog, but would not be editable from the connect dialog.

Yes, that is your proposal, the proposal from Houbsta is have this in the connect window.  I see advantages for both, but if the management can be kept simple, think that in the connect window is more logical (after initially being totally against it in the connect window).

jp8:
> The current proposal is @dcorson-ticino-com latest commit on this PR,

Sorry my mistake.

dcorson-ticino-com:
As I see the picture of the last commit somehow didn't get loaded, here it is:

CustomDir9

CustomDir010

jp8:
> (BTW I think there was some suggestion that IP addresses in the Settings would resolve to DNS names. That'll be the subject of a seaprate PR if we do that at all.)

Not exactly to resolve them.  I'm just conscious that custom central servers have no other name, so they would be listed by their DNS name, or indeed by their IP address if that is what the user enters.

pljones:
@gilgongo 
> there is also nothing taking up extra taking up space on the connect dialogue, until you select the Custom entry.

Wrong.  The proposed UI takes up extra space constantly.  It's just that, unless "Custom" is chosen, the additional input field isn't display.  Making "Directory" a combobox would solve both the issue of needing to go to Settings->Advanced _and_ not change the existing UI layout to achieve.

pljones:
> we keep the setting up of directories in Settings, along with the currently available behaviour (which allows you to add multiples and delete them). We then add the addresses added in those settings in the Directory combo list.

> I am against having two places to do the same thing.

I've not seen anything suggesting _doing the same thing_ in two places.  Like @gilgongo says, Settings->Advanced would be for doing stuff you couldn't do on Connect.  So _explicitly_ not the same thing.

gilgongo:
@pljones 

> Making "Directory" a combobox 

OK, but the standard implementation of a combo box is that it appears blank at first as an input field (albiet it with a cue for a drop down interaction on the right). That's a major emphasis change in the UI that I would predict that would inconvenience the majority of users for the sake of the minority in many cases. And what would the default server list be?

image


In theory, we could have it present as a drop-down, but with an editable entry. I would think that typing into that would be tricky (not least becuase of mouse drift closign the window unless we prevented that). I've not seen that implemented myself.

image

jp8:
Just to be clear, my proposal is for a simple drop-down on the connect dialog, but listing the built-in and custom central servers together.  The word 'Custom' would disappear from the connect dialog UI.  The layout would remain exactly as it is in version 3.8.0

No other changes, as the custom central server list is already editable on the setup dialog.

gilgongo:
@jp8 

> but listing the built-in and custom central servers together. 

 See my video above. Does that do it?

jp8:
> See my video above. Does that do it?

Impossible to say, as you haven't shown what the drop-down looks like after adding some custom central servers.

gilgongo:
> what the drop-down looks like after adding some custom central servers.

image

jp8:
So no, my proposal was to list the built-in and custom central server together, in a single drop-down.  Literally instead of a single entry named 'Custom', there would be multiple entries, named by the DNS name or IP number of the custom central server.

And, just like the current 3.8.0 UI, no editing/management functionality at all on the connect dialog.  Management UI already exists and works perfectly on the setup dialog.

jp8:
connect

So, on the single drop-down on the connect dialog, remove the Custom, and add all of these...

settings

gilgongo:
OK. So this is the divergence of opininon I think. 

Do we:

1. Split the UX of "managing" directories accross Settings and Connect. So, set up the list in Settings, access them in Connect.

or 

2. Have this management UI entirely in the Connect window.

We now have UI that could do either of these without unduly confusing/inconveneicneing the majority and also withuot using any extra space.

I would say that 2 **might** be more aligned to people's mental modal of "setting up" directories before they use Jamulus perhaps. Similar to the way we have other "set and forget" things in Settings, like Mixer Row numbers or skins. Both of which you are unlikely to be changeing with much frequency.

jp8:
The advantages I see for keeping the management on the settings page are:
- it is exactly what is already in place
- it keeps the connect dialog really tidy
- it allows for more complex management later (deleting the built-in genres, re-ordering the list of central servers, parental control, favourites, and other ideas nobody yet thought of)

dcorson-ticino-com:
Yes, but:
- This PR has the custom directory definitions in the connect menu, there is no reason to keep it in settings
- The Connect dialog is not filled with confusion by this PR
- This is about custom servers, it is not the place to do parental control, deleting default directories(?why?), etc.
- By putting the custom directories in the same list as the default directories as I suggested as a future step the interface becomes even simpler.

For me this discussion is becoming redundant, I read over and over the same things and my responses are apparently not being read.  Please excuse me here, I think this PR is ready, there are other more interesting things to do to continue.

gilgongo:
> there is no reason to keep it in settings

@jp8 has pointed out, that there is in fact little reason to move it from Settings, where (we can assume) most users are used to seeing it. And there are others reasons relating to mental model as I've stated here https://github.com/jamulussoftware/jamulus/pull/1884#issuecomment-866675751 

> The Connect dialog is not filled with confusion by this PR

It's even less filled by @jp8 suggestion.

> This is about custom servers, it is not the place to do parental control, deleting default directories(?why?), etc.

Agreed. But none of that is being expressed or implied by @jp8 suggestion.

> By putting the custom directories in the same list as the default directories as I suggested as a future step the interface becomes even simpler.

So why not do it now then? As @jp8 says:

- it is exactly what is already in place [in Settings]
- it keeps the connect dialog really tidy
- it allows for more complex management later

> For me this discussion is going in circles,

Not for me. I think we have a good resolution in fact. Not in the current commit to this PR though, unfortunately.

dcorson-ticino-com:
https://github.com/jamulussoftware/jamulus/discussions/1869#discussion-3410106

gilgongo:
> #1869 (comment)

Yes, but we changed our minds: https://github.com/jamulussoftware/jamulus/pull/1884#issuecomment-866627123. And I think that's for the better given the issues expressed by @pljones and others.

jp8:
After all the simplification, and with the main functionality already done by @dcorson-ticino-com pull request, I am happy to submit a small code modification for the combined combobox.
Screenshot_2021-06-26_07-55-02

But I don't know what is best:
- new independant pull request (but it includes code from @dcorson-ticino-com )
- new pull request that depends on this one (messy, with changes made then reverted)
- send my changes to @dcorson-ticino-com with a request to review and add them to this pull request

dcorson-ticino-com:
Shall I close this now that #1894 is incorporated ?

I just tested #1894, looks good !

pljones:
Personally, I'd say yes.  It's probably better to consider any future enhancements from scratch.

dcorson-ticino-com:
OK, I'm closing.

Issue #1811, opened 2021-06-02
Use constant for JACK like for Jamulus


ann0see:
We use `APP_NAME` for "Jamulus".  I think it would help translators to put "JACK" into a `#define` and use that throughout, too -- probably for a separate issue, though.

_Originally posted by @pljones in https://github.com/jamulussoftware/jamulus/pull/1732#discussion_r637601036_

PR #1887, opened 2021-06-21
Use JACK define constant

ann0see:
Fixes: https://github.com/jamulussoftware/jamulus/issues/1811

After looking a bit at the code, I do not think that it actually has a lot of benefit. Please start %1 to restart %2 ??? This doesn't really help translators in my opinion... There are some other fixes, however, they can be merged in a different PR...

### DIFF
--- a/linux/sound.cpp
+++ b/linux/sound.cpp
@@ -44,8 +44,9 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
 
     if ( pJackClient == nullptr )
     {
-        throw CGenErr ( tr ( "JACK couldn't be started automatically. "
-                             "Please start JACK manually and check for error messages." ) );
+        throw CGenErr ( tr ( "%1 couldn't be started automatically. "
+                             "Please start %1 manually and check for error messages." )
+                             .arg ( JACK_NAME ) );
     }
 
     // tell the JACK server to call "process()" whenever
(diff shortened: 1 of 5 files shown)


linux/sound.cpp
```diff
@@ -61,9 +62,10 @@ void CSound::OpenJack ( const bool bNoAutoJackConnect, const char* jackClientNam
     // check sample rate, if not correct, just fire error
     if ( jack_get_sample_rate ( pJackClient ) != SYSTEM_SAMPLE_RATE_HZ )
     {
-        throw CGenErr ( QString ( tr ( "JACK isn't running at a sample rate of <b>%1 Hz</b>. Please use "
+        throw CGenErr ( QString ( tr ( "%! isn't running at a sample rate of <b>%2 Hz</b>. Please use "
```
ann0see, line 65: Will not work. Is a typo...

linux/sound.cpp
```diff
@@ -207,11 +210,11 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
     // without a Jack server, Jamulus makes no sense to run, throw an error message
     if ( bJackWasShutDown )
     {
-        throw CGenErr ( QString ( tr ( "JACK was shut down. %1 "
-                                       "requires JACK to run. Please restart %1 to "
-                                       "start JACK again. "
-                                       "the JACK output." ) )
```
ann0see, line 213: To do in another PR.

linux/sound.cpp
```diff
@@ -207,11 +210,11 @@ int CSound::Init ( const int /* iNewPrefMonoBufferSize */ )
...
-                            .arg ( APP_NAME ) );
+        throw CGenErr ( QString ( tr ( "%1 was shut down. %2 "
+                                       "requires %1 to run. Please restart %2 to "
+                                       "start %1 again." ) )
```
ann0see, line 215: I think this *removes* context.

ann0see (commented):
Not happy... therefore I stopped.

pljones:
Agreed.  Now I see the result, it's not useful.

ann0see:
Ok. So closing this.

PR #1910, opened 2021-07-13
Multichannel Input using phantom clients

dcorson-ticino-com:
Multichannel Input using phantom clients

Dual Mono-in/Stereo-out added. Note this is not yet working for delay
panned servers. I don't understand the code, still looking on that.

There is no change to the protocol, works with old clients (of course they can't input 
dual mono, but the old client can receive and manipulate the channel).
Using version number 3.9.9 and requires that on the server and client to enable dual mono.

Also compiled with the new ASIOSDK2.3.3 seems to be working fine.

I'm thinking an addition to the protocol between the dual mono input
client and server might be nice to be able to give the phantom channel a
different name and instrument. That could still work with old clients at
the receiving end. Right now I am just appending -L and -R to the name.

Short description of changes

Multichannel Input using phantom clients

Context: Fixes an issue?

Does this change need documentation? What needs to be documented and how?

Status of this Pull Request
Draft and testing, note delay pan is not yet implemented for Dual Mono-in/Stereo-out

What is missing until this pull request can be merged?
Testing, testing, testing
Agreement that we should do multichannel like this
Checklist

I've verified that this Pull Request follows the general code principles
[x ] I tested my code and it does what I want
My code follows the style guide
I waited some time after this Pull Request was opened and all GitHub checks completed without errors.
I've filled all the content above

### DIFF
--- a/autobuild/linux/autobuild_deb_1_prepare.sh
+++ b/autobuild/linux/autobuild_deb_1_prepare.sh
@@ -13,4 +13,4 @@ echo "Update system..."
 sudo apt-get -qq update
 
 echo "Install dependencies..."
-sudo apt-get -qq -y install devscripts build-essential debhelper libjack-jackd2-dev qtbase5-dev qttools5-dev-tools
+sudo apt-get -qq -y install devscripts build-essential debhelper libjack-jackd2-dev qt6-qtbase-dev qt6-qtbase-dev-tools


dcorson-ticino-com:
I am apparently not able to get the correct autobuild_deb_1_prepare.sh committed.
Someone please hit git in the right places so that if accepts to commit the file for qt5 and not qt6.
This is why the linux build is not working.

DavidSavinkoff:
- Going to Qt6 is as disrupting as IPv6.
(especially if only macos needs it. Apple(tm) will lead Jamulus as a whole, into a twisted wreck with the directions it will go to avoid freedom software)
- I'm shocked that you can't compile with Qt5.
- What is going on?

dcorson-ticino-com:
> 
> 
>     * Going to Qt6 is as disrupting as IPv6.
>       (especially if only macos needs it. Apple(tm) will lead Jamulus as a whole, into a twisted wreck with the directions it will go to avoid freedom software)
> 
>     * I'm shocked that you can't compile with Qt5.
> 
>     * What is going on?

Compiling with Qt6 was no problem, its just that git seems to think it always has to use the file for Qt6 even when I have given it the Qt5 file for a Qt5 compile.  Git drives me crazy.  This is why the linux autobuild failed.
I have tried to change this back about 5 times now and it never works. :(

dcorson-ticino-com:
This is now running on my ARM Ubuntu public server.
Try it at Classical>Strings Server.
It works with older clients, but needs a new client from this build (v3.9.9dev) to input in Dual Mono-in/Stereo-out.
Without the new client you will see no difference at all.

npostavs:
It looks like the code you mean to talk about is in your `MultiChannel` branch, but you opened this PR against your `master` branch? (this PR has no changes except for the switch to qt6 deps)

https://github.com/jamulussoftware/jamulus/compare/master...dcorson-ticino-com:MultiChannel

PR #1914, opened 2021-07-14
retain user's directory selection where possible

jp8:

Upon modifications to the vector of custom directories, restore the user's selected directory on the Connect dialog, where possible.  If the previously selected directory has been deleted, select the 'default' directory.

**Context: Fixes an issue?**

addresses discussion #1899

no

Working implementation

Please test and provide comments.

## Checklist

### DIFF
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -547,7 +547,7 @@ void CConnectDlg::OnCustomCentralServerAddrChanged()
     {
         // check if the currently select custom directory still exists in the now potentially re-ordered vector,
         // if so, then change to its new index.  (addresses Issue #1899)
-        int iNewIndex = cbxDirectoryServer->findText ( strPreviousSelection );
+        int iNewIndex = cbxDirectoryServer->findText ( strPreviousSelection, Qt::MatchExactly );
         if ( iNewIndex == INVALID_INDEX )
         {
             // previously selected custom directory has been deleted.  change to default directory
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -548,7 +548,7 @@ void CConnectDlg::OnCustomCentralServerAddrChanged()
         // check if the currently select custom directory still exists in the now potentially re-ordered vector,
         // if so, then change to its new index.  (addresses Issue #1899)
         int iNewIndex = cbxDirectoryServer->findText ( strPreviousSelection );
-        if ( iNewIndex == -1 )
+        if ( iNewIndex == INVALID_INDEX )
         {
             // previously selected custom directory has been deleted.  change to default directory
             pSettings->eCentralServerAddressType = static_cast<ECSAddType> ( AT_DEFAULT );
@@ -949,6 +949,7 @@ void CConnectDlg::UpdateDirectoryServerComboBox()
     {
         if ( pSettings->vstrCentralServerAddress[i] != "" )
         {
+            // add vector index (i) to the combobox as user data
             cbxDirectoryServer->addItem ( pSettings->vstrCentralServerAddress[i], i );
         }
     }
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -211,9 +211,17 @@ void CConnectDlg::RequestServerList()
 
     // update list combo box (disable events to avoid a signal)
     cbxDirectoryServer->blockSignals ( true );
-    // iCustomDirectoryIndex is non-zero only if eCentralServerAddressType == AT_CUSTOM, and represents
-    // the offset into cbxDirectoryServer after the last non-custom directory server
-    cbxDirectoryServer->setCurrentIndex ( static_cast<int> ( pSettings->eCentralServerAddressType ) + pSettings->iCustomDirectoryIndex );
+    if ( pSettings->eCentralServerAddressType == AT_CUSTOM )
+    {
+        // iCustomDirectoryIndex is non-zero only if eCentralServerAddressType == AT_CUSTOM
+        // find the combobox item that corresponds to vstrCentralServerAddress[iCustomDirectoryIndex]
+        // (the current selected custom directory)
+        cbxDirectoryServer->setCurrentIndex ( cbxDirectoryServer->findData ( QVariant ( pSettings->iCustomDirectoryIndex ) ) );
+    }
+    else
+    {
+        cbxDirectoryServer->setCurrentIndex ( static_cast<int> ( pSettings->eCentralServerAddressType ) );
+    }
     cbxDirectoryServer->blockSignals ( false );
 
     // Get the IP address of the directory server (using the ParseNetworAddress
@@ -249,7 +257,8 @@ void CConnectDlg::OnDirectoryServerChanged ( int iTypeIdx )
     // if iTypeIdx != AT_CUSTOM, then iCustomDirectoryIndex MUST be 0;
     if ( iTypeIdx >= AT_CUSTOM )
     {
-        pSettings->iCustomDirectoryIndex = iTypeIdx - AT_CUSTOM;
+        // the value for the index into the vector vstrCentralServerAddress is in the user data of the combobox item
+        pSettings->iCustomDirectoryIndex = cbxDirectoryServer->itemData ( iTypeIdx ).toInt();
         iTypeIdx                         = AT_CUSTOM;
     }
     else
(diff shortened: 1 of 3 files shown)


src/connectdlg.cpp
```diff
@@ -529,15 +538,39 @@ void CConnectDlg::OnServerAddrEditTextChanged ( const QString& )
...
     {
-        // TODO: detect if the currently select custom directory still exists in the now potentially re-ordered vector,
-        //       if so, then change to its new index.  Issue #1899
-        pSettings->eCentralServerAddressType = static_cast<ECSAddType> ( AT_DEFAULT );
-        pSettings->iCustomDirectoryIndex     = 0;
-        RequestServerList();
+        // check if the currently select custom directory still exists in the now potentially re-ordered vector,
+        // if so, then change to its new index.  (addresses Issue #1899)
+        int iNewIndex = cbxDirectoryServer->findText ( strPreviousSelection );
+        if ( iNewIndex == -1 )
```
ann0see, line 551: Shouldn’t -1 be INVALID_INDEX?

jp8: yes !  fixed.

pljones (approved):
Looks good.  Could rebase and squash.

src/connectdlg.cpp
```diff
@@ -529,15 +538,39 @@ void CConnectDlg::OnServerAddrEditTextChanged ( const QString& )
...
     if ( pSettings->eCentralServerAddressType == AT_CUSTOM )
     {
-        // TODO: detect if the currently select custom directory still exists in the now potentially re-ordered vector,
-        //       if so, then change to its new index.  Issue #1899
-        pSettings->eCentralServerAddressType = static_cast<ECSAddType> ( AT_DEFAULT );
-        pSettings->iCustomDirectoryIndex     = 0;
-        RequestServerList();
+        // check if the currently select custom directory still exists in the now potentially re-ordered vector,
+        // if so, then change to its new index.  (addresses Issue #1899)
+        int iNewIndex = cbxDirectoryServer->findText ( strPreviousSelection );
```
ann0see, line 550: Is it possible to have an entry with the same ip/domain twice? Probably not. 

According to the Qt docs:

> Returns the index of the item containing the given text; otherwise returns -1.

What happens if we have

wold.co.uk and wold.co as directories?

ann0see:
I think the "Custom directory server" entry in settings should now be renamed. Custom directory servers or saved custom directory servers?

If the directory is changed in settings, it no longer is in the connect dialog which is a change in behavior. Therefore we should give a thought on how to ensure this is not seen as a bug. 

Tagging @mulyaj and @gilgongo here (for comments on UX).

src/connectdlg.cpp
jp8: There are options on findText do do a partial match.  But the default (if I understand the docs correctly) is QT::MatchExactly

pljones: > Is it possible to have an entry with the same ip/domain twice?

Servers are unique by IP:port.  I run up to a dozen (although most of them are directory servers themselves).

gilgongo:
> I think the "Custom directory server" entry in settings should now be renamed.

I've always been worried that using the "server" word when referring to directories in the UI will lead to confusion. Yes, directory servers are technically "servers" a bit like "domains" are technically "names" along with with address names and host names. But the way they behave for the average user is very different. So by calling them just "directories", it removes any ambiguity or doubt. You don't have to think about "the two types of server that people refer to", just two things that you deal with in the UI, like faders, mute buttons or levels. 

So we should just have have "servers" and "directories" (in the UI at least - in the Server Admin doc we can refer to "directory servers" of course).

pljones:
If you're _running a server_ that's supplying a directory to clients, it's a directory *server*.

gilgongo:
@pljones I know that directory servers are technically servers. But my point is that _normal users_ of Jamulus don't need to think in such "engineering" terms to use them. Somebody singing in a choir who needs to set up their client to connect to a thing doesn't need to know what that thing technically is, and if that thing is named similarly to another thing, then conversations about both can get confusing. We are asking people to remember that there are two types of server, which is unecessary. We have a similar issue with the word "mute" perhaps too, but that's a harder problem.

pljones:
I thought I was agreeing...  From a client perspective, they're selecting a custom directory.

However, any server operator needs to know they're running a server -- otherwise they won't specify `-s` or `--server`, for a start -- so that's a given, not an "engineering term" at all.

gilgongo:
@ann0see  

> If the directory is changed in settings, it no longer changes in the connect dialog which is a change in behavior. 

Not sure I understand. If I change the directory in settings, that (new or changed) directory appears in the genre list drop-down. is that a change?

gilgongo:
@pljones  Oh I thought you were disagreeing. But yes, for the server admins then we should make it clear with "directory server" to distinguish between that and any other type.

ann0see:
> Not sure I understand. If I change the directory in settings, that (new or changed) directory appears in the genre list drop-down. is that a change?

if you use the version of this PR and you change the directory in settings at least for me, the directory does not change on the connect dialog. Say, you selected worldjam.vip in the connect dialog and then moved changed the directory in settings to localhost. The connect dialog would stay on worldjam.vip.

gilgongo:
@ann0see Do you mean if you have the connect diologue open the whole time while you change settings? If so, OK but I doubt that's much of an issue, is it?

That said, is the absence of a submit button for adding (and maybe deleting? not sure) a problem? It feels a bit odd not having it. What if it conditionally appeared so you could "positively" make the change? 


https://user-images.githubusercontent.com/4561747/126062719-5b2301fb-8672-43ad-9753-639c5c087e0f.mp4

pljones:
> 
> 
> @ann0see Do you mean if you have the connect diologue open the whole time while you change settings? If so, OK but I doubt that's much of an issue, is it?
> 
> That said, is the absence of a submit button for adding (and maybe deleting? not sure) a problem? It feels a bit odd not having it. What if it conditionally appeared so you could "positively" make the change?
> button.mp4

Both of those points were addressed on the discussion on which this PR is based.

ann0see:
> Do you mean if you have the connect diologue open the whole time while you change settings? If so, OK but I doubt that's much of an issue, is it?

That’s the old workflow. Since UI wise not much has changed my question is if the behaviour after this PR is confusing.

ann0see:
Maybe we can agree on a compromise: Rename custom Directory Server Address to Custom Directories and document how custom directories can be deleted/modified/... on the website?

jp8:
> Maybe we can agree on a compromise: Rename custom Directory Server Address to Custom Directories and document how custom directories can be deleted/modified/... on the website?

For now I like that idea.  In the future I would like to propose a directory editor that allows:

- re-order the custom directories
- remove custom directories

I still feel strongly, based on quite a lot of experience with users, that adding directories should happen transparently if the user enters a directory into the custom server field.

I don't think there is any realistic possibility that users can understand the difference.  And I don't think users *should* need to understand the difference.

With a UI like I describe, most users, even if they have a few custom directories, will *never* have a need to go and edit their custom directories.

jp8:
Now that I've changed a couple of strings in the UI... how do I alert the translators ?

jp8:
> > If the directory is changed in settings, it no longer changes in the connect dialog which is a change in behavior.

Right.

> Not sure I understand. If I change the directory in settings, that (new or changed) directory appears in the genre list drop-down. is that a change?

If you change the selection in settings, it only changes the order of the combobox items in connect.  Newly added or selected custom directories go to the bottom of the combobox.

softins:
> Now that I've changed a couple of strings in the UI... how do I alert the translators ?

No need. We batch up the translation requests when approaching a release.

gilgongo:
> I still feel strongly, based on quite a lot of experience with users, that adding directories should happen transparently if the user enters a directory into the custom server field.

If you mean combining the in-built "genres" with custom directory addresses, then we've explored that and for various reaons found it too intrusive, haven't we? Please remember that Jamulus users who need to deal with a custom directory (let alone multiple ones, or so many that they need ordering!) are very, very few.

gilgongo:
@pljones 

> Both of those points were addressed on the discussion on which this PR is based.

That's good. So I'm addressing them here then :-)

ann0see (approved):
Not tested & compiled, code seems ok. 

Will approve on the base of more discussion/PRs following

jp8:
> If you mean combining the in-built "genres" with custom directory addresses, then we've explored that and for various reaons found it too intrusive, haven't we?

I didn't find these discussion but I would like to read them - do you have any links ?

> Please remember that Jamulus users who need to deal with a custom directory (let alone multiple ones, or so many that they need ordering!) are very, very few.

I'm not so sure.  I try to help support choirs and online events - both are heavy users of custom directories.

gilgongo:
Those diacussions are here I think, https://github.com/jamulussoftware/jamulus/pull/1884#issuecomment-864391060 but you may be referring to something else.

PR #1889, opened 2021-06-22
Unify JACK wording, fix bug, refactoring

ann0see:
**Short description of changes**

Includes changes of https://github.com/jamulussoftware/jamulus/pull/1887

**Context: Fixes an issue?**

No. Refactoring only.

**Does this change need documentation? What needs to be documented and how?**

No

**Status of this Pull Request**

To be tested and reviewed. It's only a small change though. 

**What is missing until this pull request can be merged?**
Tests...

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want (compiled; didn't try to evoke error messages though)
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

PR #1927, opened 2021-07-25
fix(i18n): localization fixes

BLumia:

Fix some i18n/l10n issues.

**Context: Fixes an issue?**

No.

No.

Waiting on second review and style changes

2nd review.

## Additional information

In the current state, there are some issues that need to be addressed to make the translation process easies. The issues are:

1. We need to avoid splitting the original strings into many `tr()` if they are from the same sentence.
2. We *may* need to avoid using the term "Country" (recommended to use "Region" or something else instead) to avoid potential [politically controversial term](https://en.wikipedia.org/wiki/Taiwan,_China)s.
3. Recommended to use the 2nd parameter of `tr()` to add a comment to the original strings when translator needs to do extra things.

## Checklist

### DIFF
--- a/src/chatdlg.cpp
+++ b/src/chatdlg.cpp
@@ -132,7 +132,7 @@ void CChatDlg::OnAnchorClicked ( const QUrl& Url )
     {
         if ( QMessageBox::question ( this,
                                      APP_NAME,
-                                     tr ( "Do you want to open the link '%1' in an external browser?" ).arg ( "<b>" + Url.toString() + "</b>" ),
+                                     tr ( "Do you want to open the link '%1' in your browser?" ).arg ( "<b>" + Url.toString() + "</b>" ),
                                      QMessageBox::Yes | QMessageBox::No ) == QMessageBox::Yes )
         {
             QDesktopServices::openUrl ( Url );
--- a/src/chatdlg.cpp
+++ b/src/chatdlg.cpp
@@ -132,7 +132,7 @@ void CChatDlg::OnAnchorClicked ( const QUrl& Url )
     {
         if ( QMessageBox::question ( this,
                                      APP_NAME,
-                                     tr ( "Do you want to open the link" ) + " <b>" + Url.toString() + "</b> " + tr ( "in an external browser?" ),
+                                     tr ( "Do you want to open the link '%1' in an external browser?" ).arg ( "<b>" + Url.toString() + "</b>" ),
                                      QMessageBox::Yes | QMessageBox::No ) == QMessageBox::Yes )
         {
             QDesktopServices::openUrl ( Url );
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -273,7 +273,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                "<ul>"
                                "<li>"
                                "<b>" +
-                               tr ( "Mono" ) + "</b> " + tr ( "and " ) + "<b>" + tr ( "Stereo" ) + ":</b> " +
+                               tr ( "%1 and %2:" ).arg ( "<b>" + tr ( "Mono" ) + "</b>", "<b>" + tr ( "Stereo" ) + "</b>" ) +
                                tr ( "These modes use "
                                     "one and two audio channels respectively." ) +
                                "</li>"
@@ -730,11 +730,11 @@ void CClientSettingsDlg::UpdateJitterBufferFrame()
     // update slider value and text
     const int iCurNumNetBuf = pClient->GetSockBufNumFrames();
     sldNetBuf->setValue ( iCurNumNetBuf );
-    lblNetBuf->setText ( tr ( "Size: " ) + QString().setNum ( iCurNumNetBuf ) );
+    lblNetBuf->setText ( tr ( "Size: " ) + QString::number ( iCurNumNetBuf ) );
 
     const int iCurNumNetBufServer = pClient->GetServerSockBufNumFrames();
     sldNetBufServer->setValue ( iCurNumNetBufServer );
-    lblNetBufServer->setText ( tr ( "Size: " ) + QString().setNum ( iCurNumNetBufServer ) );
+    lblNetBufServer->setText ( tr ( "Size: " ) + QString::number ( iCurNumNetBufServer ) );
 
     // if auto setting is enabled, disable slider control
     const bool bIsAutoSockBufSize = pClient->GetDoAutoSockBufSize();
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -235,10 +235,10 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                          "lets you choose input or outputs of your device(s). "
                                          "More information can be found on jamulus.io." );
 
-    QString strSndCardDriverSetupTT = QString ( tr ( "Opens the driver settings. Note: %1 currently only supports devices "
-                                                     "with a sample rate of %2 Hz. "
-                                                     "You will not be able to select a driver/device which doesn't. "
-                                                     "For more help see jamulus.io." ) )
+    QString strSndCardDriverSetupTT = tr ( "Opens the driver settings. Note: %1 currently only supports devices "
+                                           "with a sample rate of %2 Hz. "
+                                           "You will not be able to select a driver/device which doesn't. "
+                                           "For more help see jamulus.io." )
                                           .arg ( APP_NAME )
                                           .arg ( SYSTEM_SAMPLE_RATE_HZ ) +
                                       TOOLTIP_COM_END_TEXT;
--- a/src/serverdlg.cpp
+++ b/src/serverdlg.cpp
@@ -106,16 +106,16 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
     edtLocationCity->setAccessibleName ( tr ( "City where the server is located line edit" ) );
 
     // location country
-    QString strLocCountry = "<b>" + tr ( "Location country" ) + ":</b> " +
-                            tr ( "The country in "
-                                 "which this server is located can be set here. If a country is "
+    QString strLocCountry = "<b>" + tr ( "Country/Region" ) + ":</b> " +
+                            tr ( "The country or region in "
+                                 "which this server is located can be set here. If a location is "
                                  "entered, it will be shown in the connect dialog server list at the "
                                  "clients." );
 
     lblLocationCountry->setWhatsThis ( strLocCountry );
     cbxLocationCountry->setWhatsThis ( strLocCountry );
 
-    cbxLocationCountry->setAccessibleName ( tr ( "Country where the server is located combo box" ) );
+    cbxLocationCountry->setAccessibleName ( tr ( "Combo box for location of this server" ) );
 
     // recording directory
     pbtRecordingDir->setAccessibleName ( tr ( "Display dialog to select recording directory button" ) );
@@ -193,7 +193,7 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
         SystemTrayIcon.setContextMenu ( pSystemTrayIconMenu );
 
         // set tool text
-        SystemTrayIcon.setToolTip ( QString ( APP_NAME ) + tr ( " server" ) );
+        SystemTrayIcon.setToolTip ( tr ( "%1 server", "%1 is the name of the main application" ).arg ( APP_NAME ) );
 
         // show icon of state "inactive"
         SystemTrayIcon.setIcon ( QIcon ( BitmapSystemTrayInactive ) );
@@ -339,7 +339,7 @@ lvwClients->setMinimumHeight ( 140 );
     UpdateGUIDependencies();
 
     // set window title
-    setWindowTitle ( APP_NAME + tr ( " Server" ) );
+    setWindowTitle ( tr ( "%1 Server", "%1 is the name of the main application" ).arg ( APP_NAME ) );
 
     // View menu  --------------------------------------------------------------
     QMenu* pViewMenu = new QMenu ( tr ( "&Window" ), this );
--- a/src/serverdlg.cpp
+++ b/src/serverdlg.cpp
@@ -107,10 +107,9 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
 
     // location country
     QString strLocCountry = "<b>" + tr ( "Country/Region" ) + ":</b> " +
-                            tr ( "The country or region in "
-                                 "which this server is located can be set here. If a location is "
-                                 "entered, it will be shown in the connect dialog server list at the "
-                                 "clients." );
+                            tr ( "Set the country or region where the server is running. "
+                                 "Clients will show this location in their connect dialog's server "
+                                 "list." );
 
     lblLocationCountry->setWhatsThis ( strLocCountry );
     cbxLocationCountry->setWhatsThis ( strLocCountry );
--- a/src/serverdlgbase.ui
+++ b/src/serverdlgbase.ui
@@ -60,7 +60,7 @@
       </sizepolicy>
      </property>
      <property name="currentIndex">
-      <number>0</number>
+      <number>1</number>
      </property>
      <widget class="QWidget" name="tabServerSetup">
       <attribute name="title">
@@ -136,7 +136,7 @@
             <item>
              <widget class="QLabel" name="lblLocationCountry">
               <property name="text">
-               <string>Location: Country</string>
+               <string>Location: Region</string>
               </property>
              </widget>
             </item>
@@ -219,7 +219,7 @@
          <item>
           <widget class="QLabel" name="lblLanguage">
            <property name="text">
-            <string>Language</string>
+            <string extracomment="Consider add &quot;(Lang)&quot; as a suffix to ensure that users who selected the wrong language find the correct button">Language</string>
            </property>
           </widget>
          </item>
--- a/src/serverdlgbase.ui
+++ b/src/serverdlgbase.ui
@@ -60,7 +60,7 @@
       </sizepolicy>
      </property>
      <property name="currentIndex">
-      <number>1</number>
+      <number>0</number>
      </property>
      <widget class="QWidget" name="tabServerSetup">
       <attribute name="title">
--- a/src/soundbase.cpp
+++ b/src/soundbase.cpp
@@ -114,8 +114,9 @@ QString CSoundBase::SetDev ( const QString strDevName )
 
                 // store error return message
                 strReturn = QString ( tr ( "Can't use the selected audio device "
-                                           "because of the following error: " ) ) +
-                            strErrorMessage + QString ( tr ( " The previous driver will be selected." ) );
+                                           "because of the following error: %1 "
+                                           "The previous driver will be selected." )
+                                          .arg ( strErrorMessage ) );
             }
             else
             {
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -533,7 +533,10 @@ CAboutDlg::CAboutDlg ( QWidget* parent ) : CBaseDlg ( parent )
                               "<p><b>" +
                               tr ( "Slovak" ) +
                               "</b></p>"
-                              "<p>Jose Riha (<a href=\"https://github.com/jose1711\">jose1711</a>)</p>" );
+                              "<p>Jose Riha (<a href=\"https://github.com/jose1711\">jose1711</a>)</p>" +
+                              "<p><b>" + tr ( "Simplified Chinese" ) +
+                              "</b></p>"
+                              "<p>Gary Wang (<a href=\"https://github.com/BLumia\">BLumia</a>)</p>" );
 
     // set version number in about dialog
     lblVersion->setText ( GetVersionAndNameStr() );


src/serverdlg.cpp
```diff
@@ -193,7 +193,7 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
         SystemTrayIcon.setContextMenu ( pSystemTrayIconMenu );
 
         // set tool text
-        SystemTrayIcon.setToolTip ( QString ( APP_NAME ) + tr ( " server" ) );
+        SystemTrayIcon.setToolTip ( tr ( "%1 server", "%1 is the name of the main application" ).arg(APP_NAME) );
```
ann0see, line 196:
```suggestion
        SystemTrayIcon.setToolTip ( tr ( "%1 server", "%1 is the name of the main application" ).arg( APP_NAME ) );
```
You should run clang-format on your pc this should fix the style.

src/soundbase.cpp
```diff
@@ -114,8 +114,9 @@ QString CSoundBase::SetDev ( const QString strDevName )
 
                 // store error return message
                 strReturn = QString ( tr ( "Can't use the selected audio device "
-                                           "because of the following error: " ) ) +
-                            strErrorMessage + QString ( tr ( " The previous driver will be selected." ) );
+                                           "because of the following error: %1 "
+                                           "The previous driver will be selected." )
```
ann0see, line 118: Not 100$ sure, but I think the error message might be long and even contain HTML. That’s why the message is at the back.

ann0see (commented):
Thanks. Good catches. 

Could you please add yourself as translator in src/utils.h?

src/serverdlg.cpp
```diff
@@ -106,16 +106,16 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
     edtLocationCity->setAccessibleName ( tr ( "City where the server is located line edit" ) );
 
     // location country
-    QString strLocCountry = "<b>" + tr ( "Location country" ) + ":</b> " +
-                            tr ( "The country in "
```
ann0see, line 110: What about "Country/Region" here too?

src/util.cpp
```diff
@@ -532,7 +532,11 @@ CAboutDlg::CAboutDlg ( QWidget* parent ) : CBaseDlg ( parent )
                               "<p><b>" +
                               tr ( "Slovak" ) +
                               "</b></p>"
-                              "<p>Jose Riha (<a href=\"https://github.com/jose1711\">jose1711</a>)</p>" );
+                              "<p>Jose Riha (<a href=\"https://github.com/jose1711\">jose1711</a>)</p>" +
+                              "<p><b>" +
+                              tr ( "Simplified Chinese" ) +
```
BLumia, line 537: This would fall `clang-format` check but using the result from `clang-format` will make the style not aligned to the format used above in the same file. What should I do or did I missed something?

src/serverdlg.cpp
BLumia: Also sounds okay to me :)

src/util.cpp
ann0see: Hmm. Maybe the clang format versions are different.

Have a look at the patch here: https://github.com/jamulussoftware/jamulus/pull/1927/checks?check_run_id=3154583850

src/chatdlg.cpp
```diff
@@ -132,7 +132,7 @@ void CChatDlg::OnAnchorClicked ( const QUrl& Url )
     {
         if ( QMessageBox::question ( this,
                                      APP_NAME,
-                                     tr ( "Do you want to open the link" ) + " <b>" + Url.toString() + "</b> " + tr ( "in an external browser?" ),
+                                     tr ( "Do you want to open the link %1 in an external browser?" ).arg ( " <b>" + Url.toString() + "</b> " ),
```
ann0see, line 135:
```suggestion
                                     tr ( "Do you want to open the link « %1 » in an external browser?" ).arg ( " <b>" + Url.toString() + "</b> " ),
```

or

```suggestion
                                     tr ( "Do you want to open the link '%1' in an external browser?" ).arg ( " <b>" + Url.toString() + "</b> " ),
```

if that works.

src/clientsettingsdlg.cpp
```diff
@@ -229,13 +229,13 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
...
-                                      tr ( "Hz. "
+    QString strSndCardDriverSetupTT = tr ( "Opens the driver settings. Note: %1"
+                                           " currently only supports devices supporting a sample rate of %2"
+                                           "Hz. "
                                            "You will not be able to select a driver/device which doesn't. "
-                                           "For more help see jamulus.io." ) +
-                                      TOOLTIP_COM_END_TEXT;
+                                           "For more help see "
+                                           "jamulus.io.%3" )
+                                          .arg ( APP_NAME, QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ), TOOLTIP_COM_END_TEXT );
```
ann0see, line 238: No need for .setNum() as far as I remember.

src/clientsettingsdlg.cpp
```diff
@@ -229,13 +229,13 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                          "let you choose input or outputs of your device(s). "
                                          "More information can be found on jamulus.io." );
 
-    QString strSndCardDriverSetupTT = tr ( "Opens the driver settings. Note: " ) + APP_NAME +
-                                      tr ( " currently only supports devices supporting a sample rate of " ) +
-                                      QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ) +
-                                      tr ( "Hz. "
+    QString strSndCardDriverSetupTT = tr ( "Opens the driver settings. Note: %1"
```
ann0see, line 232: The space should be in this line to remain consistent to other files.

src/serverdlg.cpp
```diff
@@ -106,16 +106,16 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
...
+                            tr ( "The country or region in "
                                  "which this server is located can be set here. If a country is "
                                  "entered, it will be shown in the connect dialog server list at the "
                                  "clients." );
 
     lblLocationCountry->setWhatsThis ( strLocCountry );
     cbxLocationCountry->setWhatsThis ( strLocCountry );
 
-    cbxLocationCountry->setAccessibleName ( tr ( "Country where the server is located combo box" ) );
+    cbxLocationCountry->setAccessibleName ( tr ( "Region where the server is located combo box" ) );
```
ann0see, line 118: What about "Combo box for location of this server"?

src/serverdlg.cpp
```diff
@@ -193,7 +193,7 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
         SystemTrayIcon.setContextMenu ( pSystemTrayIconMenu );
 
         // set tool text
-        SystemTrayIcon.setToolTip ( QString ( APP_NAME ) + tr ( " server" ) );
+        SystemTrayIcon.setToolTip ( tr ( "%1 server", "%1 is the name of the main application" ).arg ( APP_NAME ) );
```
ann0see, line 196: Do we need the explanation here?

Maybe better make it clear that %1 is a placeholder for the application name (which will be filled automatically) —> no need for the translator to translate it.

src/serverdlgbase.ui
```diff
@@ -60,7 +60,7 @@
       </sizepolicy>
      </property>
      <property name="currentIndex">
-      <number>0</number>
+      <number>1</number>
```
ann0see, line 63: What’s this change?

src/serverdlgbase.ui
```diff
@@ -219,7 +219,7 @@
          <item>
           <widget class="QLabel" name="lblLanguage">
            <property name="text">
-            <string>Language</string>
+            <string extracomment="Consider add &quot;(Lang)&quot; as a suffix to ensure that users who selected the wrong language find the correct button">Language</string>
            </property>
```
ann0see, line 223: Great! Thanks.

ann0see (changes requested):

src/serverdlgbase.ui
BLumia: I‘m not quite sure since it was changed by the QtDesigner. Not sure if it will break something or not if I manually change it back.

src/serverdlg.cpp
BLumia: Translators will only see "%1 server" which can be confused if no context is provided, it can be helpful if we switch to other translation platforms like Transifex, Crowdin, or Weblate later. I guess it's okay for the current translation workflow (translator will need to clone the repo and Linguist can provide contexts like UI or code.

src/soundbase.cpp
BLumia: It's no different from the older form in action so I guess it's okay?

ann0see: Argh, yes. So it’s ok probably.

src/chatdlg.cpp
```diff
@@ -132,7 +132,7 @@ void CChatDlg::OnAnchorClicked ( const QUrl& Url )
     {
         if ( QMessageBox::question ( this,
                                      APP_NAME,
-                                     tr ( "Do you want to open the link %1 in an external browser?" ).arg ( " <b>" + Url.toString() + "</b> " ),
+                                     tr ( "Do you want to open the link '%1' in an external browser?" ).arg ( " <b>" + Url.toString() + "</b> " ),
```
ann0see, line 135: The question is if single quotes work…

BLumia: If you mean if single quotes can be used inside `tr()` with `%1`, yes it will work fine.

``` cpp
// output will be: "aaa 'bbb' ccc"
qDebug() << tr("aaa '%1' ccc").arg("bbb");
```

ann0see: Ok. Going to trust you on that ;-).

ann0see:
```suggestion
                                     tr ( "Do you want to open the link '%1' in an external browser?" ).arg ( "<b>" + Url.toString() + "</b>" ),
```

src/clientsettingsdlg.cpp
```diff
@@ -229,13 +229,13 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                          "let you choose input or outputs of your device(s). "
                                          "More information can be found on jamulus.io." );
 
-    QString strSndCardDriverSetupTT = tr ( "Opens the driver settings. Note: " ) + APP_NAME +
-                                      tr ( " currently only supports devices supporting a sample rate of " ) +
-                                      QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ) +
-                                      tr ( "Hz. "
+    QString strSndCardDriverSetupTT = tr ( "Opens the driver settings. Note: %1 "
+                                           "currently only supports devices supporting a sample rate of %2"
+                                           "Hz. "
```
ann0see, line 234: For this file see: https://github.com/jamulussoftware/jamulus/pull/1889

src/clientsettingsdlg.cpp
```diff
@@ -267,7 +267,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                "<ul>"
                                "<li>"
                                "<b>" +
-                               tr ( "Mono" ) + "</b> " + tr ( "and " ) + "<b>" + tr ( "Stereo" ) + ":</b> " +
+                               tr ( "%1 and %2:" ).arg ( "<b>" + tr ( "Mono" ) + "</b>", "<b>" + tr ( "Stereo" ) + "</b>" ) +
```
ann0see, line 270: Good idea.

src/clientsettingsdlg.cpp
```diff
@@ -722,11 +722,11 @@ void CClientSettingsDlg::UpdateJitterBufferFrame()
     // update slider value and text
     const int iCurNumNetBuf = pClient->GetSockBufNumFrames();
     sldNetBuf->setValue ( iCurNumNetBuf );
-    lblNetBuf->setText ( tr ( "Size: " ) + QString().setNum ( iCurNumNetBuf ) );
+    lblNetBuf->setText ( tr ( "Size: " ) + QString::number ( iCurNumNetBuf ) );
```
ann0see, line 725: Is that the modern way?

src/serverdlg.cpp
```diff
@@ -106,16 +106,16 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
     edtLocationCity->setAccessibleName ( tr ( "City where the server is located line edit" ) );
 
     // location country
-    QString strLocCountry = "<b>" + tr ( "Location country" ) + ":</b> " +
-                            tr ( "The country in "
+    QString strLocCountry = "<b>" + tr ( "Country/Region" ) + ":</b> " +
+                            tr ( "The country or region in "
                                  "which this server is located can be set here. If a country is "
```
ann0see, line 111:
```suggestion
                                 "which this server is located can be set here. If a location is "
```

src/util.cpp
```diff
@@ -532,7 +532,10 @@ CAboutDlg::CAboutDlg ( QWidget* parent ) : CBaseDlg ( parent )
                               "<p><b>" +
                               tr ( "Slovak" ) +
                               "</b></p>"
-                              "<p>Jose Riha (<a href=\"https://github.com/jose1711\">jose1711</a>)</p>" );
+                              "<p>Jose Riha (<a href=\"https://github.com/jose1711\">jose1711</a>)</p>" +
+                              "<p><b>" + tr ( "Simplified Chinese" ) +
+                              "</b></p>"
```
ann0see, line 537: Hmm. This looks strange. But as far as I know, clang-format complained if you did it as above?

src/clientsettingsdlg.cpp
BLumia: Could do a rebase once that PR get merged :)

src/util.cpp
BLumia: > clang-format complained if you did it as above?

[Yes](https://github.com/jamulussoftware/jamulus/pull/1927#discussion_r676131598)

src/clientsettingsdlg.cpp
BLumia: I actually never used `QString::setNum()` before.

ann0see: MAYBE you don't even need ::number. At least that might be possible.

src/serverdlgbase.ui
```diff
@@ -219,7 +219,7 @@
          <item>
           <widget class="QLabel" name="lblLanguage">
            <property name="text">
-            <string>Language</string>
+            <string extracomment="Consider add &quot;(Lang)&quot; as a suffix to ensure that users who selected the wrong language find the correct button">Language</string>
```
ann0see, line 222: Should be documented somewhere else (for the next release): Translators should add this in their language.

src/clientsettingsdlg.cpp
BLumia: Really? I don't see any documented API for things like `QString(int number)` or similar. You can pass a number to `QString()` because there is a `QString(QChar ch)` but that's not for this use case. For example, `QString(5)` will results in a Unicode character "\u0005", NOT a string with a number "5". `::setNum()` or `QString::number()` should be used instead,

ann0see: Ok. So my assumption was wrong. Probably an issue with my PHP background ;-).

ann0see (approved):
All in all, I think it’s ok. Style wise it’s a bit odd (especially your addition of your name as discussed above). 

Nevertheless: waiting on review from another person.

ann0see:
Todo:

- [ ] QString::number is new (and not yet used anywhere in the code, as far as I know). I‘d accept it. Not sure about others
- [ ] Compilation check on other OS (needed?)
- [ ] Finally: Squash & merge.

BLumia:
> Style wise it’s a bit odd (especially your addition of your name as discussed above).

I can change back to what I did in 87f6770 if preferred as I originally did, but `clang-format` would fail after that.

> QString::number is new (and not yet used anywhere in the code, as far as I know). I‘d accept it. Not sure about others

I can change back to use `QString().setNum()` if preferred as I originally did. I change it to QString::number because you say [it's no need for .setNum()](https://github.com/jamulussoftware/jamulus/pull/1927#discussion_r676139025) so I'm a little confused since it's the only other way to do :/

btw, the *Allow edits by maintainers* option is enabled so feel free to push changes directly to my patchset :)

ann0see:
> but clang-format would fail after that.

That's why we can't change that. So we need to leave it as is.

> I can change back to use QString().setNum()

I think in the comment you mentioned I wasn't clear enough, we usually use:

```
.arg ( APP_NAME )
.arg ( SYSTEM_SAMPLE_RATE_HZ )
```
See: https://github.com/jamulussoftware/jamulus/blob/7fa437292dcfd61d08e4fbb9ebd1cb0c500cd0e4/windows/sound.cpp#L161-L162

However, this does *not* mean the current style cannot be improved.

> btw, the Allow edits by maintainers option is enabled so feel free to push changes directly to my patchset :)

Thanks for the information. Usually we (or at least I) avoid doing that since it might cause confusion.

ann0see:
BTW @BLumia have a look at https://github.com/jamulussoftware/jamulus/releases/tag/r3.8.0devNightly1 this should include your translation.

BLumia:
> > I can change back to use QString().setNum()
> 
> I think in the comment you mentioned I wasn't clear enough, we usually use:
> 
> ```
> .arg ( APP_NAME )
> .arg ( SYSTEM_SAMPLE_RATE_HZ ) // note there's no QString.setNum() here.
> ```
> 
> See:
> 
> https://github.com/jamulussoftware/jamulus/blob/7fa437292dcfd61d08e4fbb9ebd1cb0c500cd0e4/windows/sound.cpp#L161-L162
> 
> However, this does _not_ mean the current style cannot be improved.
 
Oh! I didn't realize you mean chaining `.arg()`. That's also okay and let me know if chaining `arg()` is preferred :)

ann0see:
Hmm. I think chaining .arg() is what we used till now. So I think this should be preferred.

ann0see (approved):
Thank you!

src/clientsettingsdlg.cpp
```diff
@@ -722,11 +724,11 @@ void CClientSettingsDlg::UpdateJitterBufferFrame()
...
```
ann0see, line 727: Maybe this could be done with arg() too. 
```suggestion
    lblNetBuf->setText ( QString ( tr ( "Size: %1" ) ) .arg ( iCurNumNetBuf ) );
```

Similar below. 

But I‘m not a Qt expert.

ann0see:
Tagging @gilgongo for a review here since it concerns some language changing/…

src/clientsettingsdlg.cpp
BLumia: Just `tr("").args()` will be fine :)

```suggestion
    lblNetBuf->setText ( tr ( "Size: " ).arg ( iCurNumNetBuf );
```

src/chatdlg.cpp
```diff
@@ -132,7 +132,7 @@ void CChatDlg::OnAnchorClicked ( const QUrl& Url )
     {
         if ( QMessageBox::question ( this,
                                      APP_NAME,
-                                     tr ( "Do you want to open the link" ) + " <b>" + Url.toString() + "</b> " + tr ( "in an external browser?" ),
+                                     tr ( "Do you want to open the link '%1' in an external browser?" ).arg ( "<b>" + Url.toString() + "</b>" ),
```
gilgongo, line 135: Is "your" a bit less technical?

```suggestion
                                     tr ( "Do you want to open the link '%1' in your browser?" ).arg ( "<b>" + Url.toString() + "</b>" ),
```

src/clientsettingsdlg.cpp
```diff
@@ -229,13 +229,15 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                          "let you choose input or outputs of your device(s). "
                                          "More information can be found on jamulus.io." );
 
-    QString strSndCardDriverSetupTT = tr ( "Opens the driver settings. Note: " ) + APP_NAME +
-                                      tr ( " currently only supports devices supporting a sample rate of " ) +
-                                      QString().setNum ( SYSTEM_SAMPLE_RATE_HZ ) +
-                                      tr ( "Hz. "
+    QString strSndCardDriverSetupTT = tr ( "Opens the driver settings. Note: %1 "
+                                           "currently only supports devices supporting a sample rate of %2"
```
gilgongo, line 233: Is "works" clearer than merely "supports" (it removes the slighty awkward repetition of "support" too)?

```suggestion
                                           "currently only works with devices supporting a sample rate of %2"
```

src/serverdlg.cpp
```diff
@@ -106,16 +106,16 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
     edtLocationCity->setAccessibleName ( tr ( "City where the server is located line edit" ) );
 
     // location country
-    QString strLocCountry = "<b>" + tr ( "Location country" ) + ":</b> " +
-                            tr ( "The country in "
-                                 "which this server is located can be set here. If a country is "
+    QString strLocCountry = "<b>" + tr ( "Country/Region" ) + ":</b> " +
+                            tr ( "The country or region in "
+                                 "which this server is located can be set here. If a location is "
```
gilgongo, line 111: As per style guide, we should try to put things into active case for better readability. So this would be:

"Set the country or region in which this server is located here. Clients will show this location in their connect dialog's server list."

src/clientsettingsdlg.cpp
```diff
@@ -722,11 +724,11 @@ void CClientSettingsDlg::UpdateJitterBufferFrame()
     // update slider value and text
     const int iCurNumNetBuf = pClient->GetSockBufNumFrames();
     sldNetBuf->setValue ( iCurNumNetBuf );
-    lblNetBuf->setText ( tr ( "Size: " ) + QString().setNum ( iCurNumNetBuf ) );
+    lblNetBuf->setText ( tr ( "Size: %1" ).arg ( iCurNumNetBuf ) );
 
     const int iCurNumNetBufServer = pClient->GetServerSockBufNumFrames();
     sldNetBufServer->setValue ( iCurNumNetBufServer );
-    lblNetBufServer->setText ( tr ( "Size: " ) + QString().setNum ( iCurNumNetBufServer ) );
+    lblNetBufServer->setText ( tr ( "Size: " ) + QString::number ( iCurNumNetBufServer ) );
```
ann0see, line 732: Should be as above (.arg()).

src/serverdlg.cpp
```diff
@@ -106,16 +106,16 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
...
                                  "entered, it will be shown in the connect dialog server list at the "
                                  "clients." );
 
     lblLocationCountry->setWhatsThis ( strLocCountry );
     cbxLocationCountry->setWhatsThis ( strLocCountry );
 
-    cbxLocationCountry->setAccessibleName ( tr ( "Country where the server is located combo box" ) );
+    cbxLocationCountry->setAccessibleName ( tr ( "Combo box for location of this server" ) );
```
gilgongo, line 118: Best practice for copy is not to refer to UI or positional elements as they might change (and in this case people might not know what a "combo box" is). Can we just say this instead? I don't know where this text appears exactly though.

```suggestion
    cbxLocationCountry->setAccessibleName ( tr ( "The location of this server" ) );
```

src/serverdlgbase.ui
gilgongo: "suffix" is a bit technical perhaps.

```suggestion
            <string extracomment="Consider adding &quot;(Lang)&quot; at the end so that users who select the wrong language can find the correct button">Language</string>
```

BLumia:
About suggest changes: `supports -> work with` and `an external -> your`:

This patch is intended as a fix for the hard-to-translate separated strings instead of an improvement of original copywriting so I kept the original copywriting as-is unchanged (expect the region thing). Do we really need to tweak the other copywriting in this patch? I believe you can find more issues in the original copywriting strings if you do a review for other strings but IMO it's not related to the topic of this patch.

Let me know if you insist to do these changes in this patch set.

src/serverdlg.cpp
BLumia: These strings set with `setAccessibleName()` are for screen readers, they are actually not displayed on the UI.

src/clientsettingsdlg.cpp
BLumia: Oh, I missed this one, good catch :)

src/clientsettingsdlg.cpp
```diff
@@ -722,11 +724,11 @@ void CClientSettingsDlg::UpdateJitterBufferFrame()
...
```
BLumia, line 731:
```suggestion
    lblNetBufServer->setText ( tr ( "Size: %1" ).arg ( iCurNumNetBufServer ) );
```

src/serverdlg.cpp
```diff
@@ -106,16 +106,16 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
...
 
     // location country
-    QString strLocCountry = "<b>" + tr ( "Location country" ) + ":</b> " +
-                            tr ( "The country in "
-                                 "which this server is located can be set here. If a country is "
+    QString strLocCountry = "<b>" + tr ( "Country/Region" ) + ":</b> " +
+                            tr ( "The country or region in "
+                                 "which this server is located can be set here. If a location is "
                                  "entered, it will be shown in the connect dialog server list at the "
                                  "clients." );
```
BLumia, line 113: > As per style guide, we should try to put things into active case for better readability. So this would be:
>
> "Set the country or region in which this server is located here. Clients will show this location in their connect dialog's server list."

@gilgongo is this correct?

```suggestion
                            tr ( "Set the country or region in which this server is located here. "
                                 "Clients will show this location in their connect dialog's server "
                                 "list." );
```

src/serverdlg.cpp
BLumia: See [here](https://github.com/jamulussoftware/jamulus/pull/1927/files#r684075970).

gilgongo:
> This patch is intended as a fix for the hard-to-translate separated strings

Oh OK, I was just respondind to @ann0see tag as that implied these were new strings.

ann0see:
Yes, although there are no new strings, it’s still good to have the old ones reviewed from a UX POV. 
Maybe not part of this PR then.

ann0see:
@BLumia Could you please rebase this PR since mine was merged?

BLumia:
Squashed the patches locally and rebased them onto the master, done :)

src/serverdlg.cpp
pljones: > server is located

"where the server is running" might be more explicit as to what's actually useful.

I've seen people putting the location the server is expected to be used from (e.g. "Cornwall" for a server that I'd say is almost certainly hosted in London, unless they've got a dedicated fibre from the data centre Jamulus Explorer runs in...  Bristol, which is closer and does have good network links, has a higher ping...).

ann0see:
Thanks

src/serverdlg.cpp
BLumia:
```suggestion
                            tr ( "Set the country or region where the server is running. "
                                 "Clients will show this location in their connect dialog's server "
                                 "list." );
```

@pljones is this correct?

gilgongo: Looks good to me.

gilgongo (approved):
Looks good from my POV.

src/chatdlg.cpp
```diff
@@ -132,7 +132,7 @@ void CChatDlg::OnAnchorClicked ( const QUrl& Url )
     {
         if ( QMessageBox::question ( this,
                                      APP_NAME,
-                                     tr ( "Do you want to open the link" ) + " <b>" + Url.toString() + "</b> " + tr ( "in an external browser?" ),
+                                     tr ( "Do you want to open the link '%1' in your browser?" ).arg ( "<b>" + Url.toString() + "</b>" ),
```
pljones, line 135: I hadn't realised `tr()` returns a QString that can then have `arg()` applied.  This is a nice idiom that could be used further.

src/serverdlg.cpp
pljones: Most of the accessibility string state the nature of the UI component that they are describing - it helps the visually impared handle the screen navigation if they're prompted in this way.

src/serverdlgbase.ui
pljones: It usually happens if you save when exiting the UI Designer having examined a field and not left it in its "clean" state.  I think it should be reverted.

pljones:
There's a tiny revert I'd like done but it looks good other than that.

src/chatdlg.cpp
BLumia: This patchset contains all the strings I found while doing the Simplified Chinese translation that concatenates use `+` instead of a formatted string with `arg()` (since it causes trouble for translating). I'm not sure if I missed some other places but most major strings I found while translating are covered in this PR.

src/serverdlgbase.ui
ann0see: @BLumia could you please revert this as Peter suggested?

BLumia: > It usually happens if you save when exiting the UI Designer having examined a field and not left it in its "clean" state.

Yeah as you can see I did add the `extracomment` field with the designer mode. I didn't touch any other thing so it's probably caused by that. anyway here we go :)

```suggestion
      <number>0</number>
```

PR #1926, opened 2021-07-24
UI menubar: re-arrange existing menu items

jp8:
**Short description of changes**

Re-arrange existing menu items to adhere to industry standard UI guidelines.

**Context: Follows from discussion #1915 

**Does this change need documentation? What needs to be documented and how?**

If this gets merged, I will open a pull request to the documentation section.

**Status of this Pull Request**

Working implementation

**What is missing until this pull request can be merged?**

Please test and provide comments.

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

PR #1934, opened 2021-07-27
[Discussion] UI Menubar changes - pull request 2

jp8:

A collection of small changes to the menus.

CHAT
Chat menubar -> removed for desktop applications
Clear Chat History -> moved to application menubar

FILE
Connection Setup -> renamed to Connect, and assigned Ctrl-O
Load Mixer Channels Setup -> renamed to Load Mix
Save Mixer Channels Setup -> renamed to Save Mix

EDIT
Clear All Stored Solo and Mute Settings -> renamed to Clear Solo and Mute
New Client Level -> renamed to Default Level
Default Level changed from 100 -> 80
Auto-Adjust all Faders -> renamed to Sample and Adjust Faders

VIEW
No User Sorting -> removed
removed reference to 'Users'

**Context**

Follows from discussion #1915 
and from pull request #1926

Yes - to be determined.

DRAFT - (not fully functional)

Discussion
## Checklist

- [ ] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)

### DIFF
--- a/src/chatdlg.cpp
+++ b/src/chatdlg.cpp
@@ -54,23 +54,29 @@ CChatDlg::CChatDlg ( QWidget* parent ) : CBaseDlg ( parent, Qt::Window ) // use 
     edtLocalInputText->setPlaceholderText ( tr ( "Type a message here" ) );
 
     // Menu  -------------------------------------------------------------------
+#if defined( Q_OS_IOS )
     QMenuBar* pMenu     = new QMenuBar ( this );
     QMenu*    pEditMenu = new QMenu ( tr ( "&Edit" ), this );
 
-    pEditMenu->addAction ( tr ( "Cl&ear Chat History" ), this, SLOT ( OnClearChatHistory() ), QKeySequence ( Qt::CTRL + Qt::Key_E ) );
+    // pEditMenu->addAction ( tr ( "Cl&ear Chat History" ), this, SLOT ( OnClearChatHistory() ), QKeySequence ( Qt::CTRL + Qt::Key_E ) );
 
     pMenu->addMenu ( pEditMenu );
-#if defined( Q_OS_IOS )
     QAction* action = pMenu->addAction ( tr ( "&Close" ) );
     connect ( action, SIGNAL ( triggered() ), this, SLOT ( close() ) );
 #endif
 
 #if defined( ANDROID ) || defined( Q_OS_ANDROID )
+    QMenuBar* pMenu     = new QMenuBar ( this );
+    QMenu*    pEditMenu = new QMenu ( tr ( "&Edit" ), this );
+
+    // pEditMenu->addAction ( tr ( "Cl&ear Chat History" ), this, SLOT ( OnClearChatHistory() ), QKeySequence ( Qt::CTRL + Qt::Key_E ) );
+
+    pMenu->addMenu ( pEditMenu );
     pEditMenu->addAction ( tr ( "&Close" ), this, SLOT ( close() ), QKeySequence ( Qt::CTRL + Qt::Key_W ) );
 #endif
 
     // Now tell the layout about the menu
-    layout()->setMenuBar ( pMenu );
+    // layout()->setMenuBar ( pMenu );
 
     // Connections -------------------------------------------------------------
     QObject::connect ( edtLocalInputText, &QLineEdit::textChanged, this, &CChatDlg::OnLocalInputTextTextChanged );
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -291,17 +291,15 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     pEditMenu->addAction ( tr ( "Cle&ar Solo and Mute" ), this, SLOT ( OnClearAllStoredSoloMuteSettings() ) );
 
-    pEditMenu->addAction ( tr ( "Set to Default &Level" ),
-                           this,
-                           SLOT ( OnSetAllFadersToNewClientLevel() ),
-                           QKeySequence ( Qt::CTRL + Qt::Key_L ) );
+    pEditMenu->addAction ( tr ( "Set to Default &Level" ), this, SLOT ( OnSetAllFadersToNewClientLevel() ), QKeySequence ( Qt::CTRL + Qt::Key_L ) );
 
     pEditMenu->addAction ( tr ( "Sample and Adjust &Faders" ), this, SLOT ( OnAutoAdjustAllFaderLevels() ), QKeySequence ( Qt::CTRL + Qt::Key_F ) );
 
     pEditMenu->addSeparator();
 
-    QAction* ClearChatAction = pEditMenu->addAction ( tr ( "Cl&ear Chat History" ), this, SLOT ( OnClearChatHistory() ), QKeySequence ( Qt::CTRL + Qt::Key_E ) );
-    ClearChatAction->setEnabled( false );
+    QAction* ClearChatAction =
+        pEditMenu->addAction ( tr ( "Cl&ear Chat History" ), this, SLOT ( OnClearChatHistory() ), QKeySequence ( Qt::CTRL + Qt::Key_E ) );
+    ClearChatAction->setEnabled ( false );
 
     // View menu  --------------------------------------------------------------
     QMenu* pViewMenu = new QMenu ( tr ( "&View" ), this );
@@ -312,10 +310,8 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     QAction* ByNameAction =
         pViewMenu->addAction ( tr ( "Sort by &Name" ), this, SLOT ( OnSortChannelsByName() ), QKeySequence ( Qt::CTRL + Qt::Key_N ) );
 
-    QAction* ByInstrAction = pViewMenu->addAction ( tr ( "Sort by &Instrument" ),
-                                                    this,
-                                                    SLOT ( OnSortChannelsByInstrument() ),
-                                                    QKeySequence ( Qt::CTRL + Qt::Key_I ) );
+    QAction* ByInstrAction =
+        pViewMenu->addAction ( tr ( "Sort by &Instrument" ), this, SLOT ( OnSortChannelsByInstrument() ), QKeySequence ( Qt::CTRL + Qt::Key_I ) );
 
     QAction* ByGroupAction =
         pViewMenu->addAction ( tr ( "Sort by &Group" ), this, SLOT ( OnSortChannelsByGroupID() ), QKeySequence ( Qt::CTRL + Qt::Key_G ) );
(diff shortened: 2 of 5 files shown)


jp8:
Screenshot_2021-07-27_16-00-39
Screenshot_2021-07-27_16-00-45
Screenshot_2021-07-27_16-00-53
Screenshot_2021-07-27_16-00-58
Screenshot_2021-07-27_16-01-02
Screenshot_2021-07-27_16-01-20
Screenshot_2021-07-27_16-01-33

jp8:
This pull request is not ready for merging.

I've put it up so that people can review, try out the changes, and provide discussion or approval or disapproval.

pljones:
> FILE
> Connection Setup -> renamed to Connect, and assigned Ctrl-O

I've been thinking...  Yes, Ctrl-O is lovely and what I want.  But I want **more**...! 😁 

File->Disconnect should also exist.  I'd like it to be Ctrl-W (like closing a tab in a browser) if that's not already in use.

jp8:
I suspected these might be controversial :)

> > Clear All Stored Solo and Mute Settings -> renamed to Clear Solo and Mute
> 
> That's not clear - it could mean just what's on the screen now. It's not meant to do that - it should be clearing _all_ stored solo and mute settings. Like it says.

I agree it needs to be clear.  Stored where ?

> > New Client Level -> renamed to Default Level
> That's not bad but the menu entry isn't clear. "Set **Faders** to default level" would be better.

ok

> > Default Level changed from 100 -> 80
> Please don't do this. I prefer people to get the input levels correct. This encourages bad behaviour, in my opinion. And it's something easily set in your settings if you have a different opinion.

Yes good point.  But is 100% really the most sane default ?

> 
> > Auto-Adjust all Faders -> renamed to Sample and Adjust Faders
> 
> Why make the label more technical? "Auto-adjust" tells people "if you click this, magic happens" and really, that's all they want. Telling them something (what?) gets "sampled" (what's that) and then the faders are adjusted doesn't provide any more information.

'Auto' gives the impression that you are setting something to automatic.  I'm trying to put myself in the shoes of a chorister who is told to select "auto-adjust" and who doesn't do it because in his mind he already turned that on half an hour ago.

> 
> > VIEW
> > No User Sorting -> removed
> 
> Put that back! I use it and I've actually little need for sorting channels by anything. Sometimes I group by instrument to see what else is on but that's so hit-n-miss it's next to useless.

I removed it for 2 reasons:
1 - it is not 'no sorting' it is sorting by server slot
2 - server slot is an implementation detail that is meaningless to the user

Do you really prefer the channels sorted by server slot ?  If it needs to be in the menu then I need to come up with a wording for it... something like "sort by server-assigned position"

jp8:
Screenshot_2021-07-27_19-21-56
Screenshot_2021-07-27_19-22-02
Screenshot_2021-07-27_19-22-10

ann0see:
Not 100% related, but I think we should add something like "Get latest releases" or "Jamulus downloads..." or "Get updates..." in the help menu. Many users search for an "auto update" function once the "Software update available" message comes up. This should either link to the GH releases page or jamulus.io/ (and later on jamulus.io/download)

pljones:
.. mute, fader ..
> Or perhaps it all belongs on the settings dialog, where it is possible to explain exactly what it is.

Or the settings menu?

> 'Auto' gives the impression that you are setting something to automatic,  ...
> * Quick Mix Adjust

That might do...  "Adjust" is so vague...  Maybe just "Quick Mix"?

> 3 - I wanted to use Ctrl-O for something else ;)

Fine - and I don't believe every menu item needs a hot key, specifically not arcane options like the sorting settings where you'd likely be annoyed if you hit one accidentally.  But it's no reason to remove a feature - and "server channel order" is still the way I like it.  (It's what I'm used to - before there were sort options, this is the order they'd be displayed in.)  From the user's perspective, they're not requesting a sort order - it's left up to the server.  So something meaning "No specific sort" is what's wanted.

> we should add something like "Get latest releases" or "Jamulus downloads..." 

"Check for updates..." and "Jamulus downloads..." would be good, yes.  Despite the former not really being needed - but I think it's probably reassuring to have it.

jp8:
> > Or perhaps it all belongs on the settings dialog, where it is possible to explain exactly what it is.
> 
> Or the settings menu?

Well I was suggesting on a dialog where we have more room for a tooltip or short description.  There is no room for that on a menu - they have to be extremely terse and also extremely clear, and that's what I'm struggling with, for these 'reset' options.

> > 'Auto' gives the impression that you are setting something to automatic,  ...
> > 
> > * Quick Mix Adjust
> 
> That might do... "Adjust" is so vague... Maybe just "Quick Mix"?

Yes I really like it.  Do we need to check it, with the original author of the feature ?

> > 3 - I wanted to use Ctrl-O for something else ;)
> 
> Fine (and a very, very good use indeed) - and I don't believe every menu item needs a hot key, specifically not arcane options like the sorting settings where you'd likely be annoyed if you hit one accidentally. But it's no reason to remove a feature - and "server channel order" is still the way I like it. (It's what I'm used to - before there were sort options, this is the order they'd be displayed in.) From the user's perspective, they're not requesting a sort order - it's left up to the server. So something meaning "No specific sort" is what's wanted.

I know it's the original sort order, but I'm still struggling to understand what it is useful for.  To me it seems similar to asking the spreadsheet program to show me the data rows in order of which memory locations they happen to be allocated to...  like a "sort by pointer value".  I could write that feature into the spreadsheet program (not really, I couldn't) and people who are interested in internal details would surely find it interesting to sort their spreadsheet data that way.  Especially the guy who designed the memory allocation routines would find it quite natural.  But as a user of the spreadsheet program, would you really want that sort option there ?

> 
> > we should add something like "Get latest releases" or "Jamulus downloads..."
> 
> "Check for updates..." and "Jamulus downloads..." would be good, yes. Despite the former not really being needed - but I think it's probably reassuring to have it.

Yes.

jp8:
I also always leave the default sort, so there must be something about sorting by name or instrument that is *not good enough*.

Maybe what is missing is some kind of "sort by arrival time".

ann0see:
New client level default value:

> Yes good point. But is 100% really the most sane default for a brand new user ? I also wish everybody set their input level correctly, but what I've observed is that they don't.

+1.

>  But what exactly does 'stored' mean in this context ?

Doesn't Jamulus store levels of all users you already jammed with? So this function deletes all user levels (including them who are not present currently).

> No User Sorting 

Sort by arrival? But that could also be wrong... I think that's what it does currently.

Leaving the other stuff up to others...

jp8:
> Sort by arrival? But that could also be wrong... I think that's what it does currently.

The server assigns slots as the users arrive, but if somebody leaves, their slot becomes available and is assigned to the next arrival.  So in the end, it's similar to 'by arrival' but only if nobody ever leaves.  If a lot of people are coming and going, it ends up random-ish.

pljones:
> To me it seems similar to asking the spreadsheet program to show me the data rows in order of which memory locations they happen to be allocated to...

No, it's more "Just leave the cells put, I want to type into them", rather than having the spreadsheet sort them each time I change the cell content, which is what you're proposing.  It would be _really annoying_ if that happened.  The difference here is "empty cells" collapse to not being shown.

> The server assigns slots as the users arrive, but if somebody leaves, their slot becomes available and is assigned to the next arrival.

Exactly.

But you're right, I guess it's the next best thing to having people "self-sort" by _always_ appearing on the right-hand end and disappearing from wherever they happen to have ended up.

jp8:
> Just leave the cells put, I want to type into them

Yes you're right, I think that's what I like about it too.  I don't really care what order they are in, but once somebody is in a position I want them to stay there.  It eases the mental load of keeping track of which musician is in which position.

jp8:
Screenshot_2021-07-29_08-26-50
Screenshot_2021-07-29_08-26-54

DavidSavinkoff:
to jp8:
> Now that I think about it I'm not sure which is better. I can see the point of offering 'sort by arrival' and also the current sort by 'server slot'. I think I would even prefer, when a client disconnects, that the channel would just turn grey until a new person joined. That could be implemented without needing to modify the server.
> 
> So you've convinced me that we need to keep it. But I don't know what to call it. Things like "no sorting" and "unsorted" don't really describe it.

It is called Unsorted.
dos command line can be unsorted.
unix command line can be unsorted.
even floppy disks and hard drives can be unsorted.
sorting takes time, and old school'ers can feel clock cycles.
the natural order is comforting and even provides some information.

DavidSavinkoff:
to pljones:
New Client level should Not be 100%, and it should not be 0% for defaults.
A mixer-man generally wants a 0dB output. I found 80% is usually good.
I've seen some here prefer 50% so that new connections can be heard,
and turned up if needed.
100% requires every new connection (person) to be adjusted lower.

pljones:
> A mixer-man generally wants a 0dB output. 

That's a non-sensical statement.  A mix engineer would always state mix levels correctly, dBU, dBV, whatever.  Never "dB".

And it's also not true.

> 100% requires every new connection (person) to be adjusted lower.

No, it doesn't.  It requires you to decide what your preference is and set it in your settings.  And then you get your preferred choice.

jp8:
> dos command line: DIR can be unsorted.
> unix command line: ls can be unsorted.

I think you're right about DOS dir command but ls has a default alphabetical sort.

Anyway I don't think you're proposing MS-DOS as the example user interface standard that we should follow.

DavidSavinkoff:
> > A mixer-man generally wants a 0dB output.
> 
> That's a non-sensical statement. A mix engineer would always state mix levels correctly, dBU, dBV, whatever. Never "dB".

0dB is the ideal input magnitude and output magnitude for all amplifiers, unless they are designed Wrong.

dB is an incredibly useful relative thing the all elecronics engineers, and suppose, audio engineers know.
There is no nonsense here. 0db could be 1mW 1uV or 1000amps, and if your mixer set set to 0dB (to match the arbirarily set 1v output stage input sensitivity) that is what you Will get at your final output stage. If you have a mixerboard with dB markings on the sliders, 0dB will be in the 60% to 80% range.

jp8:
It's all beyond my comprehension but I think I understand the basic principle, at least from a user interface perspective:
- the default value should just leave the signal unmodified (so it should remain at 100, +1 pljones)
- that '100' value should be somewhere around the 75% mark on the slider (so that we can still amplify musicians who didn't set their gain high enough, +1 DavidSavinkoff)

But I can't deal with that in this pull request... I can only create a separate issue to discuss it.

So for this pull request I will need to leave the default 'new client level' alone.

ann0see:
@jp8 I think this has been stale for quite some time (probably because of controversy). I implemented the link to the update page in the update message so my request becomes less important. 

All the other changes should probably be raised as separate pr‘s?
Not closing this yet.

jp8:
Yes I agree separate pull requests is best.

ann0see:
Ok. So closing this then

PR #1938, opened 2021-08-02
IPv6 support, phase 1 (direct connection)

softins:
This implementation of IPv6 on Jamulus is based on #1017 by @jardous, with further enhancements by @softins. It is an alternative to #1455 by @mstolle629.

It has been tested on Linux, Mac and Windows. It is currently running on jamminthejazz, at `privjam.softins.co.uk:22123` (IPv4) and `privjam6.softins.co.uk:22123` (IPv6). There is also a private instance at `privjam.softins.co.uk` and `privjam6.softins.co.uk`, which may be used for testing.

Servers running this version can accept a mix of IPv4 and IPv6 connections.

Server addresses for direct connection can be specified in the following formats:

- `[v6addr]`
- `[v6addr]:port`
- `[v4addr]`
- `[v4addr]:port`
- `v4addr`
- `v4addr:port`
- `hostname`
- `hostname:port`

A literal IPv6 address must always be enclosed in `[ ]`, to avoid ambiguity with a port.

Phase 2 will be the addition of Directory Server support, which will require some
protocol enhancements. Phase 1 is worth adding on its own before Phase 2 is ready.

**Potential ChangeLog entry**

Add support for direct connections using IPv6

**Documentation needed**
- Command line options `-6` and `--enableipv6`
- Address formats as shown above

Working implementation

More testing is always good!

## Checklist


src/socket.cpp
```diff
@@ -40,25 +50,46 @@ void CSocket::Init ( const quint16 iPortNumber, const quint16 iQosNumber, const
...
+        //
+        const char tos = (char) iQosNumber; // Quality of Service
+        setsockopt ( UdpSocket, IPPROTO_IP, IP_TOS, &tos, sizeof ( tos ) );
+    }
+    else
+    {
+        // The IPV6_V6ONLY socket option must be false in order for the socket to listen on both protocols.
+        // On Linux it's false by default on most (all?) distros, but on Windows it is true by default
+        const uint8_t no = 0;
+        setsockopt( UdpSocket, IPPROTO_IPV6, IPV6_V6ONLY, (const char *)&no, sizeof( no ) );
```
atsampson: Missing space after setsockopt (I guess clang-format will pick this up anyway).

src/socket.cpp
```diff
@@ -40,25 +50,46 @@ void CSocket::Init ( const quint16 iPortNumber, const quint16 iQosNumber, const
...
-    UdpSocketInAddr.sin_family = AF_INET;
-    if ( strServerBindIP.isEmpty() )
-    {
-        UdpSocketInAddr.sin_addr.s_addr = INADDR_ANY;
-    }
-    else
+    UdpSocketInAddr.sin_family      = AF_INET;
+    UdpSocketInAddr.sin_addr.s_addr = INADDR_ANY;
+
+    struct sockaddr_in6 UdpSocketIn6Addr;
```
atsampson: The rest of the network code avoids writing `struct` when it's not necessary in C++ (`sockaddr`, `sockaddr_in`); can this and the sizeof below just be `sockaddr_in6`?

src/socket.cpp
```diff
@@ -181,19 +237,36 @@ void CSocket::SendPacket ( const CVector<uint8_t>& vecbySendBuf, const CHostAddr
         // send packet through network (we have to convert the constant unsigned
```
atsampson, line 250: `CVector<T>` subclasses `std::vector<T>`, so can this just use vecbySendBuf.c_data() to get a const pointer rather than copying the vector?

softins: Maybe, but I didn't change that part of the code.

src/socket.cpp
softins: Probably best to review just the differences between the branch point and the final commit.

src/client.cpp
```diff
@@ -31,6 +31,7 @@ CClient::CClient ( const quint16  iPortNumber,
                    const QString& strMIDISetup,
                    const bool     bNoAutoJackConnect,
                    const QString& strNClientName,
+                   const bool     bNEnableIPv6,
```
pljones, line 34: It doesn't need the ugly `N`, generally.

pljones:
Um...  Given IPv4 is always enabled and given that only IPv4 is used to communicate with directory servers (for now), is there any reason not to have IPv6 always enabled?

src/client.cpp
softins: Was just following what appeared to be the majority convention. No strong view either way.

softins:
I wondered that. It would fail on a system that didn't have an IPv6 stack. Such as a Linux box with a file containing "blacklist ipv6" in /etc/modprobe.d/ (I did test that on my Pi). We could try creating an AF_INET6 socket first and falling back to AF_INET, bit would then need to propagate the result of that to a global variable that all classes could consult.

pljones:
Well, rather than set `bEnableIPv6` in `main.cpp` based on a command line parameter, set it based on the result of the check, leaving the rest the same?  (More `bIPv6Available`, in other words.)  Then the it could be invisible to the user.

softins:
I did try something like that earlier, but found that it wasn't easy to pass the result of the test back up from the `CSocket`, given that the client and server sockets are not newed, but instantiated as members of `CClient` or `CServer`.

I could perhaps create a static method of `CSocket` that just tries to create an AF_INET6 socket and returns a success indication, and call that from `main.cpp`

We also need to consider the implications for:

* A client on a computer that is IPv6-capable, but connected via an internet connection that does not support IPv6
* A server that supports IPv6 and has a routable IPv6 address, but the operator does not want their Jamulus server to listen on IPv6, for whatever reason.

pljones:
You've given reasons to switch IPv6 off.  Maybe the command line parameter should do that?

I'm thinking of other software with internet connectivity -- I don't have to tell my web browser or email client whether or not to support IPv6, for example.

ann0see:
> A client on a computer that is IPv6-capable, but connected via an internet connection that does not support IPv6

Best case if host name is given: try to request an A record via DNS. Use ipv4 as fallback. But does Jamulus need to handle that? Isn’t that an OS functionality? Not only Jamulus will have this problem?!

If I‘m giving an ipv6 address it should fail and not try to get a connection over ipv4.

How does Firefox/Chrome/whatever handle this?

ann0see:
> A server that supports IPv6 and has a routable IPv6 address, but the operator does not want their Jamulus server to listen on IPv6, for whatever reason.

Use the bind IP option and only provide an ipv6/ipv4 address?

atsampson:
> How does Firefox/Chrome/whatever handle this?

[Happy Eyeballs](https://datatracker.ietf.org/doc/html/rfc8305).

Firefox does have a configuration option to disable IPv6 entirely to speed things up a bit if you know you're on a network without external IPv6 connectivity (it's still not that unusual to have local IPv6 but not routable to the world).

softins:
> > How does Firefox/Chrome/whatever handle this?
> 
> [Happy Eyeballs](https://datatracker.ietf.org/doc/html/rfc8305).

Certainly very interesting, but quite complex, and I feel it's likely to be overkill for Jamulus. It's primarily concerned with resolving DNS names using both AAAA and A records, and then trying to connect with the returned addresses, preferring IPv6. In Jamulus, an operator of a private server using IPv6 will likely give his invitees a literal address or have a specific AAAA record for it. And once we extend to directory servers, they will return to the client literal IPv6 (if any) and IPv4 addresses, not DNS names.

softins:
> You've given reasons to switch IPv6 off. Maybe the command line parameter should do that?

Maybe, but I think the number of people with an IPv6-capable computer but an IPv4-only internet connection would be a lot larger than the number of people who specifically want to use IPv6. So we would be requiring the much larger number of people to do something different.

ann0see:
> In Jamulus, an operator of a private server using IPv6 will likely give his invitees a literal address or have a specific AAAA record for it. 

What makes you sure about that?

From a UX point of view, I think just giving one address (with AAAA and A records) is the preferable outcome. Maybe - but for debugging purposes only - the admin sends over an IP to force ipv6/ipv4. That’s just my opinion…

For directories we‘re skipping DNS? 

Can we send ping messages over IPv6 and IPv4, and constantly calculate the time difference between both as soon as the connection window is open? The client will then know which one to choose before connecting. 

Issues:
* Worst case: double the ping traffic
* Complexity on the client side
* Constant calculation might not be necessary. We might stop after a few ping messages.

Pro:
* Probably similar to "Happy eyeballs"
* User doesn‘t need to decide in normal mode, but could still use the direct IP to connect to the server

ann0see:
So -6 is a client and server option?

softins:
> So -6 is a client and server option?

Yes.

softins:
> > In Jamulus, an operator of a private server using IPv6 will likely give his invitees a literal address or have a specific AAAA record for it.
> 
> What makes you sure about that?

I'm not sure, it just seems likely to me. Most people I know with private servers haven;t set up a DNS name and just give their friends the IP.

> For directories we‘re skipping DNS?

We can use a DNS name to find a directory, but when a server registers with a directory, it does so just by IP (the directory stores the IP it registered from). The server then tells a requesting client just the IP, not a DNS name, plus other relevant information (private IP, port number, etc).

> Can we send ping messages over IPv6 and IPv4, and constantly calculate the time difference between both as soon as the connection window is open? The client will then know which one to choose before connecting.

That might well be a useful enhancement for the future, but I don't think we should hold up the basic feature waiting for that particular lily to be gilded.

ann0see:
Ok. So what really needs to be tested to get this merged?

IPv6/IPv4 over a dual stack connection seems to work.

softins:
> Ok. So what really needs to be tested to get this merged?

Well I've tested it quite a bit in my environment, on Linux, Windows and Mac, and it's running on my server as mentioned above. I think it's good to go, and that it is beneficial to have `-6` to explicitly turn it on for now. Then others who want to try out IPv6 could do so, without it having unexpected impacts on those who don't.

So it just needs two approvals (which can be tricky), and it could be merged. I would then start thinking about Directory Servers.

> IPv6/IPv4 over a dual stack connection seems to work.

Cool!

pljones:
> it is beneficial to have -6 to explicitly turn it on for now.

Agreed for the "experimental support" version.

src/client.cpp
pljones: I'm not going to block it for that, either.  (It's just on my list of things to get rid of.)

src/socket.cpp
```diff
@@ -39,27 +56,72 @@ void CSocket::Init ( const quint16 iPortNumber, const quint16 iQosNumber, const
...
+        const char tos = (char) iQosNumber; // Quality of Service
+        setsockopt ( UdpSocket, IPPROTO_IPV6, IPV6_TCLASS, &tos, sizeof ( tos ) );
+
+        UdpSocketAddr.sa6.sin6_family = AF_INET6;
+        UdpSocketAddr.sa6.sin6_addr   = in6addr_any;
+        UdpSocketAddrLen              = sizeof ( UdpSocketAddr.sa6 );
+
+        UdpPort = &UdpSocketAddr.sa6.sin6_port; // where to put the port number
+
+        // FIXME: If binding a dual-protocol interface to a specific address, does it cease to be dual-protocol?
```
pljones, line 86: A socket listening on an IPv4 address is only listening for IPv4 traffic.  But the same _interface_ could have a separate socket listening on an IPv6 address, I believe.  The _interface_ has the MAC address (or equivalent).  The protocol has the IPv4 or IPv6 address that goes with the socket layer in use.

At least, that's my understanding.

src/socket.cpp
```diff
@@ -172,6 +232,15 @@ CSocket::~CSocket()
 
 void CSocket::SendPacket ( const CVector<uint8_t>& vecbySendBuf, const CHostAddress& HostAddr )
 {
+    union
```
pljones, line 235: Can this go in a header somewhere as a type, so it can be re-used more easily?

softins: I guess, although that makes it more opaque when reading `socket.cpp`.

src/socket.cpp
softins: > A socket listening on an IPv4 address is only listening for IPv4 traffic. But the same _interface_ could have a separate socket listening on an IPv6 address, I believe.

Yes it could, but Jamulus only expects to listen on one socket. So I think that makes it unable to bind to a specific interface address and remain dual-stack. I don't think it's a problem, unless someone flags up a scenario where it is. The default of binding to `::` will allow it to receive both IPv4 and IPv6 traffic without needing two sockets.

src/socket.cpp
pljones: You could argue that for every definition that isn't in line.

src/socket.cpp
pljones: > Yes it could, but Jamulus only expects to listen on one socket.
> The default of binding to :: will allow it to receive both IPv4 and IPv6 traffic without needing two sockets.

How does an IPv4 client know what address to use?  Does the IPv6 socket "just know" it's got two addresses really?

pljones (approved):

src/util.cpp
```diff
@@ -667,76 +667,118 @@ void CLanguageComboBox::OnLanguageActivated ( int iLanguageIdx )
...
 
     // first try if this is an IP number an can directly applied to QHostAddress
-    if ( !InetAddr.setAddress ( strAddress ) )
+    if ( InetAddr.setAddress ( strAddress ) )
     {
-        // it was no valid IP address, try to get host by name, assuming
+        if ( !bEnableIPv6 && InetAddr.protocol() == QAbstractSocket::IPv6Protocol )
+        {
+            // do not allow IPv6 addresses if not enabled
+            // qInfo() << qUtf8Printable ( QString ( "IPv6 addresses disabled" ) );
```
ann0see, line 735: Do we give an error message to the user somewhere else?

ann0see (commented):
Seems simple. Nevertheless some comments:
1. You're removing the chinese .qm file? Any reason for that?
2. See comments

ann0see:
Since it's not enabled by default, I think we can tag it for 3.8.1.

softins:
> Seems simple. Nevertheless some comments:
> 
> 1. You're removing the chinese .qm file? Any reason for that?

No, must be an accident. I'll correct that before merging.

> 2. See comments

Thanks, will respond individually.

> 3. As it works on my ipv6 connection (and you tested it), I think the functionality is ok.

Cool.

src/serverlist.cpp
```diff
@@ -123,6 +127,16 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
     qDebug() << "Using" << qhaServerPublicIP.toString() << "as external IP.";
     SlaveCurLocalHostAddress = CHostAddress ( qhaServerPublicIP, iNPortNum );
 
+    if ( bEnableIPv6 )
+    {
+        // set the server internal address, including internal port number
+        QHostAddress qhaServerPublicIP6;
+
+        qhaServerPublicIP6 = NetworkUtil::GetLocalAddress6().InetAddr;
+        qDebug() << "Using" << qhaServerPublicIP6.toString() << "as external IPv6.";
```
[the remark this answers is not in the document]
softins: I don't agree with this one. When IP addresses are shown (e.g. by `ip addr` or `host`), they are not quoted, as they are values in their own right, not just strings. It would be like displaying the numeric port number as `'22124'` instead of `22124`.

src/serverlist.cpp
```diff
@@ -123,6 +127,16 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
     qDebug() << "Using" << qhaServerPublicIP.toString() << "as external IP.";
```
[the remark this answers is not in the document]
softins: As above

src/util.cpp
softins: At the moment, with all versions, if an invalid hostname or IP address is entered into the Connect window, the connection fails silently. It would be good to fix that more generally, but I think that should be separate from this PR. Maybe raise it as a specific issue?

src/socket.cpp
```diff
@@ -39,27 +51,72 @@ void CSocket::Init ( const quint16 iPortNumber, const quint16 iQosNumber, const
...
-    if ( strServerBindIP.isEmpty() )
+    if ( bEnableIPv6 )
     {
-        UdpSocketInAddr.sin_addr.s_addr = INADDR_ANY;
+        // try to create a IPv6 UDP socket
+        UdpSocket = socket ( AF_INET6, SOCK_DGRAM, 0 );
+        if ( UdpSocket == -1 )
+        {
+            // IPv6 requested but not available, throw error
+            throw CGenErr ( "IPv6 requested but not available on this system.", "Network Error" );
```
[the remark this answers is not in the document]
softins: I can't see any reference to active vs passive voice in that page, although I know some other wider contexts promote a preference for active voice. In this case, I have a slight preference for the shorter message.

src/socket.cpp
```diff
@@ -99,17 +156,15 @@ void CSocket::Init ( const quint16 iPortNumber, const quint16 iQosNumber, const
...
+        bSuccess = ( ::bind ( UdpSocket, &UdpSocketAddr.sa, UdpSocketAddrLen ) == 0 );
     }
 
     if ( !bSuccess )
     {
         // we cannot bind socket, throw error
-        throw CGenErr ( "Cannot bind the socket (maybe "
-                        "the software is already running).",
-                        "Network Error" );
+        throw CGenErr ( "Cannot bind the socket (maybe the software is already running).", "Network Error" );
```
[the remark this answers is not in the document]
softins: This was just a code formatting change; the original wording was Volker's. Happy to change it if you prefer.

src/socket.cpp
softins: The IPv4 client will connect in the same way as at present, by being told the IP address, either explicitly, or from a hostname lookup, or by a directory server.

Binding an `AF_INET6` socket to `::` (declared as `in6addr_any`) is the IPv6 equivalent of binding an `AF_INET` socket to `0.0.0.0` (`INADDR_ANY`), which means it will accept any incoming packet that arrives on any interface. If the socket has been `setsockopt`ed to dual stack (as we do), it will accept both IPv4 and IPv6 packets.

An IPv4 address mapped to IPv6 uses `0000:0000:0000:0000:0000:ffff` as the first 96 bits, which denotes that the last 32 bits contain an IPv4 address in network byte order. Often expressed like `::ffff:192.168.1.1`. This mapping is used in `CSocket::SendPacket()` and checked for in `CSocket::OnDataReceived()`.

src/serverlist.cpp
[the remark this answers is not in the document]
ann0see: > It would be like displaying the numeric port number as '22124' instead of 22124.

Yes. It’s just cosmetics/personal preference. Since we have spaces before and after the IP and they all follow a standard probably the quotes are not needed.

src/socket.cpp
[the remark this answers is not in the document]
ann0see: I think @gilgongo mentioned it somewhere.

src/socket.cpp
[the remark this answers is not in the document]
ann0see: Yes. Or part of https://github.com/jamulussoftware/jamulus/pull/1889

src/util.cpp
ann0see: Done #1945

src/main.cpp
```diff
@@ -937,6 +955,7 @@ QString UsageArguments ( char** argv )
            "  -Q, --qos             set the QoS value. Default is 128. Disable with 0\n"
            "                        (see the Jamulus website to enable QoS on Windows)\n"
            "  -t, --notranslation   disable translation (use English language)\n"
+           "  -6, --enableipv6      enable IPv6 addressing (IPv4 is always enabled)\n"
```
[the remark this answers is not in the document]
pljones: I don't think this adds anything at all.  It still requires further explanation as to what "direct connection to client/server" means, so it's useless verbosity.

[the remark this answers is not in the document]
ann0see: But do we agree that the original proposal needs more clarification?

src/socket.cpp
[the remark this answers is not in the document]
gilgongo: Maybe I halluciated it then - I recall @mulyaj mentioning it.

ann0see:
As soon as the Chinese qm file is added, I think it’s ready to merge.

src/main.cpp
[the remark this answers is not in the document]
pljones: I'm happy with the wording Tony used.  It's consise and says what it does.

src/socket.cpp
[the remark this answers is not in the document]
mulyaj: Yeah, I've mentioned a preference for active voice. Reading this at first glance, the subject is vague: 

"IPv6 requested..." - who requested?

Then, if the subject is vague, the guideline for concise and specific is broken.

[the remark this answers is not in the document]
pljones: I'd be happy with "IPv6 is not available."  It's issued as a response to the user making a request to use an address.  It doesn't need to tell the user they used an address, just why it hasn't worked.

src/main.cpp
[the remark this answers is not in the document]
ann0see: I still think it's not clear. Probably the question is:
*Does the user need to know that connections to public servers is still IPv4 only?*
If we want to be specific: Yes. If we think it confuses the user: No.
Maybe it's worth mentioning it at least in the release announcement.

ann0see (approved):
Since it works and only one critical step is missing (.qm file), I‘ll approve it.

src/main.cpp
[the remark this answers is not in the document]
pljones: > Does the user need to know that connections to public servers is still IPv4 only?

Not "public", use the word "registered".  You can register with a directory no one knows about.  That's not particularly public.

To operate the software, no.  They can just select a listed entry from the directory and the software will use the returned IPv4 address.

If you want to use IPv6, you need to:
a) pass "-6" on the command line, thus indicating you understand how to operate with IPv6 addresses, so need no further help
b) enter an IPv6 name or address in the connection dialogue "Server Address", by hand, so you know that you've chosen a specific server rather than a registered one.

As I say, the help text need not say anything else.  This is not the final form for the feature.  It's intended to be somewhat experimental and the documentation should make that clear.  Those who choose to use it should thus be given the credit for understanding what they've chosen to do.

softins:
I have squashed all the commits except for the first one by @jardous. I'll wait until the checks are complete (looks like they may need to be rerun) and merge tomorrow.

ann0see:
I think you should use @jardous E-Mail or jardous@users.noreply.github.com to link his GitHub account to the first commit. This would add him to the contributor entry on GitHub.

softins:
> I think you should use @jardous E-Mail or [jardous@users.noreply.github.com](mailto:jardous@users.noreply.github.com) to link his GitHub account to the first commit. This would add him to the contributor entry on GitHub.

Not exactly sure how to do that.

pljones:
> 
> 
> > I think you should use @jardous E-Mail or [jardous@users.noreply.github.com](mailto:jardous@users.noreply.github.com) to link his GitHub account to the first commit. This would add him to the contributor entry on GitHub.
> 
> Not exactly sure how to do that.

To set the values:
```
git commit --author="Jiri Popek <jardous@users.noreply.github.com>" --date="2021-02-18T12:34:56" ...
```
You'd need to do that during either a `git rebase -i` or manually `git reset --hard master` and reapplying the commits.

softins:
> I think you should use @jardous E-Mail or [jardous@users.noreply.github.com](mailto:jardous@users.noreply.github.com) to link his GitHub account to the first commit. This would add him to the contributor entry on GitHub.

OK, that seems to have worked. Just waiting for the Checks to run and then I'll merge.

softins:
Well although Android and Windows failed this time, they have previously worked with identical code (I diffed the before and after of the rebases to be sure), so I'll go ahead and merge anyway.

jardous:
hi. Feel free to use my real email: ***@***.***
J.

softins:
> hi. Feel free to use my real email: ***@***.*** J.

I already used your generic Github address jardous@users.noreply.github.com, which was enough to link the rebased version of your original commit to your Github ID. Thanks!

Issue #1935, opened 2021-07-29
Mute Myself sometimes leads to maximum level of My channel


jregener62:
**Describe the bug**
'Mute Myself' sometimes leads to maximum level of My channel. In these situations it is not possible to change the level of My channel. It stays on maximum.

**To Reproduce**
1. set 'My channel' to any level below maximum
2. enable 'Mute Myself'

**Expected behavior**
'My channel' should not change level.

**Operating system**
macOS Big Sur 11.4, iMac Late 2015, Apogee Element 46

**Version of Jamulus**
3.8.0

ann0see:
I think this bug was already fixed: #1838 so hopefully it’ll be part of the next version

ann0see:
Tagged a nightly release for you: https://github.com/jamulussoftware/jamulus/releases/tag/r3.8.0devNightly1

Once it's built, scroll down to assets, download and install the build and test the bug is fixed.

jregener62:
Thanks. Don't see the issue anymore so far. I will verify the fix for some time and let you know the test results then.

ann0see:
Great! I‘ll close it then. If you experience it again, just re-open it and give a comment

jregener62:
Unfortunately I see the issue again. In this case I was the only participant on a server with 'Mute Myself' enabled. The volume of my channel raised to maximum when another participant joined the server. Workaround is to re-join the server.

ann0see:
Ok. Strange. Could you please update the above bug description? 

So your bug is reproducible? Did you check the linked Pull request and verify that the other bug is not occurring? If that’s the case, we need to re-open the other issue.

jregener62:
Unfortunately not reproducible every time but only sometimes. I'll try to create a configuration where the issue is always reproducible.

jregener62:
I've found a reproducible situation, at least during one Jamulus' runtime. The issue only happens on several servers. Unfortunately I don't see a regularity here but I hope it is a hint to a possible cause of the bug. Please see attached screen movie (with sound).

ann0see:
- Just edited your post to make the link clickable -


Yes, this looks like a bug. @npostavs knows a bit more about the (old) bug, so tagging it here. 

I don’t think it‘s server related since mute myself is a client feature?

jregener62:
I definitely have different results on several servers, although this feature and this bug is not likely to be related to server functionality. As a next test I would like to backup and then reset all settings but I cannot find a pref file on Mac. Any hints?

[...]

PR #1940, opened 2021-08-05
Display server version after name in mixer board

softins:
Add the version of the connected server after its name in the title of the Audio Mixer Board

**Context: Fixes an issue?**

Some issues are dependent on the version of the server the client is connected to. See for example the discussion at https://github.com/jamulussoftware/jamulus/issues/1935#issuecomment-893483589. At the moment, it is necessary to check [Jamulus Explorer](https://explorer.jamulus.io) to determine the version of a server. With this change, the version will be displayed after the server name in the client window, making diagnosis of issues easier. It might also encourage server operators to update their versions.

Documentation not really critical, as it should be self-explanatory. Servers before 3.5.5 will be shown as `<3.5.5`, as that is the first version that sends its version number to the client.

Working implementation.

Should be ready.

## Checklist

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -836,6 +836,7 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent ) :
     iRunningNewClientCnt ( 0 ),
     iNumMixerPanelRows ( 1 ), // pSettings->iNumMixerPanelRows is not yet available
     strServerName ( "" ),
+    strServerVersion ( "" ),
     eRecorderState ( RS_UNDEFINED ),
     eChSortType ( ST_NO_SORT )
 {
@@ -931,6 +932,12 @@ void CAudioMixerBoard::SetServerName ( const QString& strNewServerName )
         // which is most striking (we use filled blocks and upper case letters).
         setTitle ( u8"\u2588\u2588\u2588\u2588\u2588  " + tr ( "T R Y I N G   T O   C O N N E C T" ) + u8"  \u2588\u2588\u2588\u2588\u2588" );
     }
+}
+
+void CAudioMixerBoard::SetServerVersion ( const QString& strNewServerVersion )
+{
+    // store the current server name
+    strServerVersion = strNewServerVersion;
 }
 
 void CAudioMixerBoard::SetGUIDesign ( const EGUIDesign eNewDesign )
@@ -1096,10 +1103,16 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
 void CAudioMixerBoard::UpdateTitle()
 {
     QString strTitlePrefix = "";
+    QString strTitleSuffix = " (<3.5.5)";
 
     if ( eRecorderState == RS_RECORDING )
     {
         strTitlePrefix = "[" + tr ( "RECORDING ACTIVE" ) + "] ";
+    }
+
+    if ( !strServerVersion.isEmpty() )
+    {
+        strTitleSuffix = " (" + strServerVersion + ")";
     }
 
     // replace & signs with && (See Qt documentation for QLabel)
@@ -1109,7 +1122,7 @@ void CAudioMixerBoard::UpdateTitle()
     QString strEscServerName = strServerName;
     strEscServerName.replace ( "&", "&&" );
 
-    setTitle ( strTitlePrefix + tr ( "Personal Mix at: " ) + strEscServerName );
+    setTitle ( strTitlePrefix + tr ( "Personal Mix at: " ) + strEscServerName + strTitleSuffix );
     setAccessibleName ( title() );
 }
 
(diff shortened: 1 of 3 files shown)


pljones:
image
I think the mixer title is already a bit over-crowded, especially when recording is active.  I'd rather have it in the window title bar, if that's feasible (i.e. works cross-platform reliably).

softins:
> image
> I think the mixer title is already a bit over-crowded, especially when recording is active. I'd rather have it in the window title bar, if that's feasible (i.e. works cross-platform reliably).
> 
> And yes, there's already more going on in the window title... I guess it's that I feel the mixer title should be that - saying "this is the mix at the server you're connected to" and nothing else. I didn't like putting the recording active there, really... The window title is describing the server (i.e. name, how many users). Adding the version there just feels more in context.

I disagree. I think users are going to be more aware of what is displayed in the mixer title than the window title (I certainly am). It is also the same on each platform then. I would support replacing `[RECORDING ACTIVE]` with just `[RECORDING]`, but that's a different issue.

> It would also be a much smaller change... :)

Really? This PR isn't really a big change anyway. Most of it would be needed the same whether putting it in the window title or the mixer title.

But either way, it's worth displaying the server version in principle, yes?

pljones:
> But either way, it's worth displaying the server version in principle, yes?

Oh yes.  And I was thinking if it's in the title bar, adding the client version after "Jamulus" would also make sense.  Then getting a report from a user could ask for the full title bar text.

softins:
> > But either way, it's worth displaying the server version in principle, yes?
> 
> 
> 
> Oh yes.  And I was thinking if it's in the title bar, adding the client version after "Jamulus" would also make sense.  Then getting a report from a user could ask for the full title bar text.

I don't really think we need to put the client version on display all the time. For most people, who aren't developers, the client version will only change when they install a new version. And it is easily available already via About Jamulus.

The server version, in contrast, changes depending on where you connect to.

I still prefer putting the server version in the mixer title rather than the window title.

pljones:
I'd question then why the number of users is in the window title - it's to do with the server you're connected to and changes more than the server version does.

softins:
> I'd question then why the number of users is in the window title - it's to do with the server you're connected to and changes more than the server version does.

Maybe, but I never look at it.

gilgongo:
While I realise the client version is also in the Help panel, I think contrasting the client version with the server version helps disambiguate what's being referred to (given the possible confusion in showing a version number in the client that relates to the server). So I'd go for having it in the title bar as "ServerName (v3.6.1) - 7 users - Jamulus (v3.8.2)" as we have the space there. We could of course (also?) put it in the Help, but that does seem a bit hidden away. 

Also agree with @pljones that the best we can hope for is to use this as a diagnostic. I doubt many people would notice it proactively, as it were.

That said, might we be cheeky and for servers before 3.5.5 show that as `(legacy)`?

softins:
> While I realise the client version is also in the Help panel, I think contrasting the client version with the server version helps disambiguate what's being referred to (given the possible confusion in showing a version number in the client that relates to the server). So I'd go for having it in the title bar as "ServerName (v3.6.1) - 7 users - Jamulus (v3.8.2)" as we have the space there.

I don't like that at all, it makes it far too cluttered. The Help About is fine for getting the client version number when needed. I think the server name was added to the title bar to help those running multiple clients connected to different servers (e.g. WorldJam people).

> We could of course (also?) put it in the Help, but that does seem a bit hidden away.

If you mean the version of the connected server, that doesn't belong in the Help, as it isn't about the client.

> Also agree with @pljones that the best we can hope for is to use this as a diagnostic. I doubt many people would notice it proactively, as it were.

I still feel strongly that the better place for the server version is immediately after the server name in the mixer title, as that is the context of "where you are connected to" that people would notice most.

But I haven't got the energy to keep arguing about it, so am tempted just to close the PR.

> That said, might we be cheeky and for servers before 3.5.5 show that as `(legacy)`?

I wondered about something like that or "(VERY OLD)", but decided against it as it is a value judgement and would also require translation. So I settled on the neutral and factual "(<3.5.5)".

gilgongo:
> I don't like that at all, it makes it far too cluttered.

You just said you don't look at it :-)

gilgongo:
I don't have a problem with it being in the mixer panel, it was only that if we wanted to put it in the title, we'd need to disambiguate the server version fromt the client version (since the convention with a version number there is that it would refer to the client).

pljones:
There's too much on the mixer title bar already for my tastes.

So it's a matter of taste and preference, here, by the sound of it.

There's no technical reason to do one thing rather than the other.

At which point it raises the question: what problem is it solving:
>  With this change, the version will be displayed after the server name in the client window, making diagnosis of issues easier.

And that doesn't provide guidance on which to prefer, either.

gilgongo:
I don't think it's completely either/or, hence my point about the need to pair the client version with the server one if we put it in the title bar. I think we could have just the server version if you put it in the mixer panel.

PR #1992, opened 2021-09-05
It-update for new Release

dzpex:
Update IT translation for New Release

**Context: Fixes an issue?**

## Checklist


src/res/translation/translation_it_IT.ts
```diff
@@ -1504,7 +1504,7 @@ E&apos; stato disattivato l&apos;audio del tuo canale ed inserito il &quot;Disat
     <message>
         <location filename="../../clientsettingsdlg.cpp" line="329"/>
         <source>Custom Directory Server Address</source>
-        <translation type="unfinished">Indirizzo personalizzato del Server di Directory</translation>
+        <translation>Indirizzo personalizzato del Server principale</translation>
```
pljones, line 1507: Please note the English meaning is no long "central".  Registering with a directory server is a distributed process.  There are multiple directory servers, many private ones (more than there are public ones).  So the meaning here in Italian should reflect that usage.

dzpex: "elenco" is used for a listed items...

In italian directory can be translate with principal folder or main container...that rapresent a "Server Principale"

The directory server is the server that are usefull for user that want to connect on a public server..

If it is not the right interpretation I can modify the translation.

pljones: > The directory server is the server that are usefull for user that want to connect on a public server..

Not necessarily "public".  Only servers listed in the public directories are public servers.  There are private directories listing private servers, too.

gilgongo: OT but I agree "Custom" isn't a very good term since it implies they might have some features that are not present in other directories. Maybe "Private directory" instead? So everything is public unless otherwise stated ("I've set up a private directory." "My private server is a bit slow." "There are a lot of empty servers listed." "I can't find a good public server"). This would also bring it conceptually into line with private servers. We don't talk about "custom servers" after all.

pljones: Yeah, that's the idea.  (And then, at some point, all the variable names and perhaps method names, in the code need updating, too -- I only did the user-facing bits for the initial change.)

gilgongo (approved):
Seems OK to me - would like to get this in today as we're supposed to be freezing now.

src/res/translation/translation_it_IT.ts
dzpex: > Basically, a server that provides a directory listing is a directory server. (And it can be public - i.e. built into Jamulus - or private, what is still called "Custom" but shouldn't be... I've yet to win that one...)

For me any Server Directory contains "public" server...For me a private server was a server that is used only by a band with a ip address distribuited at all members. 

I know that not all Server Directory are public....but if i know the address of a Server Directory i can access at all server listed into them

pljones: > For me any Server Directory contains "public" server...For me a private server was a server that is used only by a band with a ip address distribuited at all members.

That's simply wrong.

A directory is contains registered servers - i.e. servers that registered with that directory.  Those registered servers are only _public_ if they register with a public directory.

dzpex:
I cannot find another way to translate Server Directory.... if you want i can change it to English and not translate it in italian at this time....

dzpex:
please close this pull request and use  #2044 for Merge

Thanls!

PR #2048, opened 2021-09-27
Improve channel lookup and logging

### LINKED ISSUE #2046: [3.8.1beta1] Existing connection gets allocated to a new channel
**Describe the bug**

It seems like in some circumstances, when a new connection arrives, existing connections get assigned into the same channel. When this happens, the symptom is:

- On the client, the affected users gets these symptoms:

  - The sound begins stuttering.
  - The user receives a welcome message again.
  - The users appears multiple times in the client list.

  It basically disrupted the whole session.

- On the server, we get weird logs that said that several people connected on the same second, but the number of connected people does not increase.

  ```
  2021-09-25 00:03:36, 112.**.112.120, connected (15)
  2021-09-25 00:03:36, 171.**.89.130, connected (15)
  2021-09-25 00:03:36, 171.**.24.59, connected (15)
  2021-09-25 00:03:36, 182.**.53.34, connected (16)
  ```

**To Reproduce**
~~Run a server with `-u 25` for max 25 users, and get several people to join. It will take a while.~~ [See this comment for a repro](https://github.com/jamulussoftware/jamulus/issues/2046#issuecomment-928096090).

**Expected behavior**
Server should go smoothly.

**Screenshots**

**Operating system**
Linux, built from source

**Version of Jamulus**
3.8.1beta

**Additional context**
I suspect that there could be a bug with this PR:

- https://github.com/jamulussoftware/jamulus/pull/1979

So, I reverted the PR (`git revert 15ee6093df26efdb0cbf264e5f8dd34c6785dea6 -m 1`), recompiled, and deployed it to my server. As of writing, I did not encounter the problem since.

softins:

In `CHostAddress::Compare()`, does an explicit comparison of IPv4 addresses instead of just subtracting them. This avoids a potential signed overflow inconsistency.

Also improves the logging of connections by fetching the channel count at the time of connection and passing it through to the logging, instead of fetching it later in the logging thread.

**Context: Fixes an issue?**

May fix #2046 

No, code improvement only

Working, but needs testing.

## Checklist

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -922,7 +922,7 @@ int CHostAddress::Compare ( const CHostAddress& other ) const
     quint32 thisAddr  = InetAddr.toIPv4Address();
     quint32 otherAddr = other.InetAddr.toIPv4Address();
 
-    return (int) thisAddr - (int) otherAddr;
+    return thisAddr < otherAddr ? -1 : thisAddr > otherAddr ? 1 : 0;
 }
 
 // Instrument picture data base ------------------------------------------------
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -562,7 +562,7 @@ void CServer::SendProtMessage ( int iChID, CVector<uint8_t> vecMessage )
     Socket.SendPacket ( vecMessage, vecChannels[iChID].GetAddress() );
 }
 
-void CServer::OnNewConnection ( int iChID, CHostAddress RecHostAddr )
+void CServer::OnNewConnection ( int iChID, int iTotChans, CHostAddress RecHostAddr )
 {
     QMutexLocker locker ( &Mutex );
 
@@ -637,7 +637,7 @@ void CServer::OnNewConnection ( int iChID, CHostAddress RecHostAddr )
     DoubleFrameSizeConvBufOut[iChID].Reset();
 
     // logging of new connected channel
-    Logging.AddNewConnection ( RecHostAddr.InetAddr, GetNumberOfConnectedClients() );
+    Logging.AddNewConnection ( RecHostAddr.InetAddr, iTotChans );
 }
 
 void CServer::OnServerFull ( CHostAddress RecHostAddr )
--- a/src/server.h
+++ b/src/server.h
@@ -181,6 +181,8 @@ public:
 
     bool PutAudioData ( const CVector<uint8_t>& vecbyRecBuf, const int iNumBytesRead, const CHostAddress& HostAdr, int& iCurChanID );
 
+    int GetNumberOfConnectedClients();
+
     void GetConCliParam ( CVector<CHostAddress>& vecHostAddresses,
                           CVector<QString>&      vecsName,
                           CVector<int>&          veciJitBufNumFrames,
@@ -261,7 +263,6 @@ protected:
     void                  InitChannel ( const int iNewChanID, const CHostAddress& InetAddr );
     void                  FreeChannel ( const int iCurChanID );
     void                  DumpChannels ( const QString& title );
-    int                   GetNumberOfConnectedClients();
     CVector<CChannelInfo> CreateChannelList();
 
     virtual void CreateAndSendChanListForAllConChannels();
@@ -413,7 +414,7 @@ signals:
 public slots:
     void OnTimer();
 
-    void OnNewConnection ( int iChID, CHostAddress RecHostAddr );
+    void OnNewConnection ( int iChID, int iTotChans, CHostAddress RecHostAddr );
 
     void OnServerFull ( CHostAddress RecHostAddr );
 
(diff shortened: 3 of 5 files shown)


src/util.cpp
```diff
@@ -922,7 +922,7 @@ int CHostAddress::Compare ( const CHostAddress& other ) const
     quint32 thisAddr  = InetAddr.toIPv4Address();
     quint32 otherAddr = other.InetAddr.toIPv4Address();
 
-    return (int) thisAddr - (int) otherAddr;
+    return thisAddr < otherAddr ? -1 : thisAddr > otherAddr ? 1 : 0;
```
dtinth, line 925: I saw that in IPv6 branch uses `memcmp`, I suppose that can be used as well for consistency?

pljones: That would probably make more logical sense.

softins: IPv6 uses `memcmp()` because the addresses are 16 bytes long. IPv4 addresses can be compared whole as they fit a 32-bit integer, so that seems more efficient than `memcmp()` which would be comparing a byte at a time.

pljones: I'd expect `memcmp` to be optimised for the platform its running on and the data sizes being compared.

dtinth: Here’s some quick and dirty benchmark. I am using Google Benchmark library (`apt install libbenchmark-dev`):

```cpp
#include <benchmark/benchmark.h>
#include <string.h>
#include <stdlib.h>

static void BM_ManualComparison(benchmark::State &state)
{
    srand(1);
    for (auto _ : state)
    {
        state.PauseTiming();
        unsigned int a = rand();
        unsigned int b = rand();
        state.ResumeTiming();
        benchmark::DoNotOptimize(a < b ? -1 : a > b ? 1 : 0);
    }
}
BENCHMARK(BM_ManualComparison);

static void BM_Memcmp(benchmark::State &state)
{
    srand(1);
    for (auto _ : state)
    {
        state.PauseTiming();
        unsigned int a = rand();
        unsigned int b = rand();
        state.ResumeTiming();
        benchmark::DoNotOptimize(memcmp(&a, &b, sizeof(unsigned int)));
    }
}
BENCHMARK(BM_Memcmp);

BENCHMARK_MAIN();
```

Running benchmark with `-O3` resulted in roughly the same amount of CPU time per iteration. Your results may vary:

```sh-session
$ g++ -lbenchmark bench.cpp -O3 && ./a.out

Running ./a.out
Run on (1 X 2900 MHz CPU )
CPU Caches:
  L1 Data 64K (x1)
  L1 Instruction 64K (x1)
  L2 Unified 512K (x1)
  L3 Unified 16384K (x1)
***WARNING*** Library was built as DEBUG. Timings may be affected.
-----------------------------------------------------------
Benchmark                    Time           CPU Iterations
-----------------------------------------------------------
BM_ManualComparison        329 ns        325 ns    2138302
BM_Memcmp                  326 ns        325 ns    2130022
```

softins: That's certainly an interesting result! :)

On a little-endian architecture such as Intel, the comparison by `memcmp()` must necessarily give different results than a direct comparison of a pair of `int` or `unsigned int` variables. In this particular application that doesn't actually matter, as the order is immaterial, provided it is consistent.

Personally, in this instance, I prefer the comparison of `unsigned int` directly (as in the current PR) rather than `memcmp()`, but not strongly enough to argue about it. If the consensus is to prefer `memcmp()`, then fair enough.

I repeated the test on my RPi4:

```sh-session
tony@pi:~/benchmark $ g++ -lbenchmark bench.cpp -O3 && ./a.out
2021-09-28 17:54:48
Running ./a.out
Run on (4 X 1500 MHz CPU s)
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
***WARNING*** Library was built as DEBUG. Timings may be affected.
-----------------------------------------------------------
Benchmark                    Time           CPU Iterations
-----------------------------------------------------------
BM_ManualComparison       2458 ns       2432 ns     285846
BM_Memcmp                 2443 ns       2425 ns     289761
tony@pi:~/benchmark $ ./a.out
2021-09-28 17:55:26
Running ./a.out
Run on (4 X 1500 MHz CPU s)
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
***WARNING*** Library was built as DEBUG. Timings may be affected.
-----------------------------------------------------------
Benchmark                    Time           CPU Iterations
-----------------------------------------------------------
BM_ManualComparison       2464 ns       2434 ns     288050
BM_Memcmp                 2487 ns       2463 ns     286765
```

So again there is not much to choose, and the two runs were different in the opposite way!

dtinth: Personally, I am fine with either way; I don't have a strong opinion — just pointing out that using `memcmp` will make both IPv4 and IPv6 code more consistent with negligible performance penalty, although consistency in this specific case may not be so important to warrant a change.

(By the way I am already deploying the code with patches from this PR, as it is right now, to my servers for testing tomorrow.)

pljones (approved):

PR #2059, opened 2021-10-05
Fix some duplicate accelerator keys

softins:
Small changes after manual review of the translations we have received so far.


softins:
> Assuming you've run your dupe checker over the chages, too? :)

Yes, this is based on master, after the merges that were done the other day. It doesn't include sk, which I'll check later.

softins:
> Rebase and push now the Linux build should work, please :).

Done - just waiting for the build to run.

PR #2060, opened 2021-10-07
Preserve environment variables when building

edoardopirovano:

This PR changes the building script to pass the `--preserve-env` option to `debuild`. This ensures the environment variables needed by CodeQL during the build are present.

**Context: Fixes an issue?**

Fixes: https://github.com/github/codeql-action/issues/760

It does not - no changes to the project itself, only to the build script used in CI.

Working implementation.

Ready to merge.

## Checklist

### DIFF
--- a/distributions/build-debian-package-auto.sh
+++ b/distributions/build-debian-package-auto.sh
@@ -31,4 +31,4 @@ echo "${VERSION} building..."
 
 sed -i "s/é&%JAMVERSION%&è/${VERSION}/g" debian/control
 
-debuild -b -us -uc
+debuild --preserve-env -b -us -uc


edoardopirovano:
> Seems to fix the Linux build. Is `--preserve-env` safe? Should we have more controls in place along with this change?

In my opinion, this is relatively safe because the build is happening in a fresh GitHub Actions environment where you know all the environment variables that have been set because there is relatively little happening before the build.

If you aren't comfortable with preserving all of the environment, we could replace this with a few instances of `--preserve-envvar <varname>` in order to preserve just the variables that CodeQL needs. In particular, I think `--preserve-envvar "LD_PRELOAD" --preserve-envvar "ODASA_TRACER_CONFIGURATION" --preserve-envvar "CODEQL_*"` would do the trick.

However, I think this is less neat and also more subject to future breakage if we change the set of environment variables we use (although I expect the `CODEQL_*` rule will catch everything we might add in future).

edoardopirovano:
> Oh, I should have asked @edoardopirovano to provide changes and contributors updates, too.

Sorry, I'm not sure what you mean by that?

Issue #2099, opened 2021-10-30
Prevent automatic start of Jamulus after installation on windows when no audio driver available


henkdegroot:
**Has this feature been discussed and generally agreed?**
Enhancement has been mentioned in #2088 

**Describe the solution you'd like**
The last screen of the windows installer contains an option to automatically start Jamulus after the installation completed.
This option is enabled by default and can be disabled by the user (if not desired to start).
The installer allows you to install Jamulus on a system which does not have a audio driver (asio or jack) installed. In this scenario is does not make sense to automatically start Jamulus after the installation completed, as it will fail due to the missing audio driver.

This enhancement applies to the windows installer only and will set the option to automatically start Jamulus after the installation to unchecked when there was no audio driver detected and the user decided to go ahead with the installation.

henkdegroot:
@gilgongo , can you assign this to me please?

henkdegroot:
@gilgongo, actually do you think there will ever be a scenario that you want to auto start if you don't have a correct driver installed during installation?
As if not, we could just remove the auto start option completely for this scenario.
Leaving it there of course when there is a driver detected.

gilgongo:
Yes that would seem potentially less confusing as we know that starting the client won't work when there is a problem with the driver. I suppose it's just possible that the subsequent errors encountered might give you some clues if it wasn't immediately obvious why you had a problem perhaps? But that seems unlikely.

ann0see:
@henkdegroot I'd suggest you to read through https://github.com/jamulussoftware/jamulus/pull/841 to get a few insights in the installer.

gilgongo:
Is this also related @henkdegroot ? Maybe look at it in the same PR? https://github.com/jamulussoftware/jamulus/issues/1756

henkdegroot:
@gilgongo, it is partial related. As the dialog in #1756 is the dialog which will be presented when Jamulus run automatically after the install (without having an ASIO driver present). However #1756 is more about "later". I.e. I may have removed the ASIO driver from my system (or it is not working anymore) and then when I start Jamulus I get this error message.

Perhaps adding a link in that dialog referring to the installation for windows page on jamulus.io would be nice?

gilgongo:
> Perhaps adding a link in that dialog referring to the installation for windows page on jamulus.io would be nice?

Yes, always good to have some "further info" available for when things don't go as expected.

Issue #860, opened 2021-01-18
Remember if user opened Jamulus on last install/skipped ASIO warning


ann0see:
From: #841 
The new windows installer currently only remembers if the user checked "Install desktop icon". Should we also save if the user checked the checkbox to open Jamulus or skipped the ASIO warning?

pljones:
I'd prefer _all_ the choices saved for the next run's default settings - a "single-click install with last used settings" is what I'm after.

ann0see:
Ok. It could get annoying in one specific case: Install Jamulus on Windows as a server only. Each time you update Jamulus, it would warn you. But it's always possible to skip it.

Concerning the "Run Jamulus" button, this it should be easy to solve, although I don't know if it is possible to do it via a section.

pljones:
> Install Jamulus on Windows as a server only. Each time you update Jamulus, it would warn you. But it's always possible to skip it.

Better solution: run the server on Linux ;)

ann0see:
> Or you could have a "Server only install" on an "Advanced install options" page and have that get remembered and mean you didn't get asked about missing ASIO drivers...

... which would make the installer more complicated. Since I don't think many people run a server on Windows, I'm not sure if this is needed.

I thought about adding a counter to the registry. If someone clicks "Yes" let's say on 4 installs, the message would then be skipped.

pljones:
> ... which would make the installer more complicated.

Yeah, was joking about that, too ;).  The counter might be confusing -- "Suddenly it doesn't ask me..." or then they decide to switch to running the server in the cloud and use that computer for their client and forget it doesn't have an ASIO driver...

Keep it simple for now - maybe have a note ("Not required for server-only use" or similar).

ann0see:
Ok. So can you add it to the warning message in the installer? I'm currently testing some stuff with codacy (code quality and security testing)

gene96817:
Thanks for all the work you have been doing.
I get calls for help when the user is confused after doing an install (because it doesn't seem to work).
With the installer remembering information from the previous run, how do I get the user to do a "clean install", meaning do an install without remembering anything?

ann0see:
A clean install is possible by running the uninstaller before then running the installer.

But the more important question is: what doesn't work?

The user might be confused by the error message "Due to a bug [...] 64 Bit version is installed to a 32 bit directory..." This message might not be clear enough.

ann0see:
@gene96817 did my answer help you?

[...]

PR #2103, opened 2021-11-01
Prevent setting 'Run Jamulus' if driver not found

henkdegroot:

Installer option to run the Jamulus Client after installation will be set to unchecked when the installation is performed without an audio driver

**Context: Fixes an issue?**

Fixes: #2099 

No documentation update required

Local build tested with and without ASIO driver and JACK installed

Nothing

## Checklist


windows/installer.nsi
```diff
@@ -413,6 +417,8 @@ FunctionEnd
             MessageBox MB_YESNO|MB_ICONEXCLAMATION "$(JACK_EXIT_NO_DRIVER)" /sd IDNO IDYES SkipMessage
                 Abort
        SkipMessage:
+            StrCpy $bRunApp "0"           ; set the app not to run after install as there is no audio driver
+            StrCpy $bShowRunAppOption "0" ; do not show the run the app option in the finish page
```
pljones, line 421: Why do we need two separate variables?

henkdegroot: bRunApp is controlling the launch of the client at the end of the installation and is also linked to the checkbox in the installer gui which the user can uncheck to prevent the client to start.
bShowRunAppOption is used to remove the checkbox and text from the gui.

The bRunApp is also set by reading from the windows registry (on init of the installer) and controls the initial setting of the checkbox (checked or not checked).

Maybe this text works better for line 420 & 462:

`set the client not to run after install as it will fail with an error message due to the absence of an audio driver`

Open to suggestion...

pljones: OK, so should bRunApp get its value changed?  If it's there to represent the user choice (possibly retrieve from the registry), does bShowRunAppOption just mean it won't get referenced?

henkdegroot: bShowRunAppOption means it will not show the tick box option in the dialog regardless if the status is checked or not checked and this status is controlled via the bRunApp value.
If bRunApp value would be set to 1 and the option would only be hidden from the dialog, it would still execute the client. Hence why I believe two variables are needed for this.

pljones: Maybe make the boolean more explicit -- `bCannotRunAppNoDriver`, initialised to false.  If the driver isn't available set it to true.  Only offer to run the app if it's false.  I'd rather have `bRunApp` called something more explicit if it is the user choice, too -- `bUserChoseRunApp` or something, defaulting to true, overwritten from the reg and then overwritten by the popup choice, if presented and stored back to the registry.  The popup wouldn't show and the app wouldn't run if the driver wasn't available.  The app would only run if `bCannotRunAppNoDriver` was false and `bUserChoseRunApp` was true.

henkdegroot: That would mean I need to change this:

`ShowWindow $mui.FinishPage.Run $bShowRunAppOption  ; option will only be displayed if an audio driver was detected during install`

into an if statement. So it would be something like:

${If} $bCannotRunAppNoDriver== 0  ; option will only be displayed if an audio driver was detected during install
       ShowWindow $mui.FinishPage.Run 1
${EndIf}

Okay?

pljones: I was thinking more it would goto some point after showing the prompt and handling the result.

henkdegroot: After showing the prompt (dialog window) to the user that there is no driver found and the user selects Yes to continue, that is when this show/notshow option is set.

The installer continues (shows a few more dialogs) and at the final page there are 2 options presented (both with tick boxes).
One for creating a desktopshortcut and one to run jamulus after install.

So there is not a separate popup displayed to the user asking if they want to run jamulus or not. 
The ShowWindow in NSIS is not showing a window, it shows a part of the text in the dialog.

To my knowledge the result will need to be handled in the FinishPage.Show function.

pljones: > The installer continues (shows a few more dialogs) and at the final page there are 2 options presented (both with tick boxes).
> One for creating a desktopshortcut and one to run jamulus after install.

Ah yes.  I'd forgotten the desktop shortcut (I detest them... best forgotten).

_Ideally_, the option to run would either not be shown or be disabled if the driver were not available -- but I see you can't just not show the dialog.  Perhaps two paths are needed with two slightly different dialogs -- one when "run the app" is possible and one when it isn't?

henkdegroot: To me adding an extra dialog for sounds a bit overkill and would most likely add more complexity (making it even less clear) to the start of the nsis script (where the sequence of the different dailogs/page is setup).

What about using a third value for the bRunApp? 
0 = means do not run be user choice (or from registry)
1 = means do run be user choice (the default)
2 = means do not run and do not show the option in the finish page dialog

When no driver is detected, we then set bRunApp to 2.
This can then be used to control the display of the option and at the same it will treat the option as unset as it is not equal to 1.

Do you know the history of why the registry value is read? When I remove jamulus, the registry key is also removed.
Is it perhaps to deal with a potential OS restart which can be required in some cases?

ann0see: No. It's a "clever" way to save the user choices. At the beginning the registry key is read, a nsis variable is sto, the key is deleted (by removal of Jamulus), since the variable is still set NSIS knows what to do and at the end we save the user choice in the registry again: https://github.com/jamulussoftware/jamulus/issues/860

Personally, I'd suggest just to unset the "run jamulus" checkbox if the ASIO error message was skipped but not save the state of the checkbox in the registry.

henkdegroot: @ann0see , thank you! I get the logic now.....interesting to see what was discussed in #860. 

My first though on this was also to just unset the "run jamulus" checkbox. The state of the checkbox would then be saved a 0 in the registry. Then thought about why would you ever "run jamulus" if you don't have the audio driver installed, as it would fail. So to prevent another issue stating a user gets an error message when he runs the client after installing, I thought it would be nicer to have it removed from the dialog in this case.

henkdegroot: @pljones  I would like to complete this PR. What is your view on changing this just to have the option to run Jamulus at the end of the install unticked when there is no driver found? Meaning leaving the option visible in the dialog and just make sure the state is set to not checked.

pljones: It would mean clarifying the title here and reviewing #2099 to make sure nothing was promised that's not being delivered -- else we'll close that issue incorrectly.

henkdegroot: So instead of "Disable run client" the title should be something like "Remove auto run client option"?

pljones: "Prevent setting 'Run Jamulus' if driver not found"?

It doesn't _automatically_ run Jamulus ever.  It offers the user a choice of whether or not to run Jamulus.  What we're saying is that
1) If the user has _never_ installed Jamulus before and all is good, default to "Yes, run Jamulus"
2) If the user _has_ installled Jamulus before and all is good, default to the registry setting read at the start of the install
3) If there was a driver error, default to "No, do not run Jamulus"

There is still the question of what to do in the registry.  In my view, you do what it currently does -- save the user choice.  If they leave it as "No" after the error, then it's saved as "No".  If they change it to "Yes" (or it defaulted to "Yes", or it was read as "Yes" and there was no error), then it's saved as "Yes".

henkdegroot: So you mean leave the option available in the dialog and just change the "default" state?

The logic you indicate does already exist and I share your view as well, as it makes sense to save whatever the user has selected.
Only thing which currently not exists is the solution is to save the "Yes" in the registry, but that is because the option is not available in the dialog with the current solution.

pljones: I don't really see the point of disabling the ability to run Jamulus from the installer.  It doesn't prevent the user immediately running the installed Jamulus, after all.

henkdegroot: Okay, will update the PR to just leave the option in the dialog and just change the "checked" option to "not checked" when there is no audio driver detected during the installation and the user continued the installation.

pljones (approved):
Looks nice and simple.

ann0see (approved):
Looks quite simple.

pljones:
@henkdegroot Are you okay to fix the conflicts?

henkdegroot:
@pljones I wasn't sure if I could/should do that or not. I have have resolved the conflicts now.

ann0see:
The JACK build failed…

henkdegroot:
> The JACK build failed…

It failed because JACK could not be installed in the prepare for windows step.
Can you rerun the build workflow?

henkdegroot:
@ann0see, pretty annoying but this time NSIS download failed. This seems to be happening very often lately, I don't know why.
Good news is that the jack install did work as normal.

Can you do another re-run please?

henkdegroot:
Finally passed the autobuild 👍

PR #2096, opened 2021-10-26
Double-check feedback detection enable state

JohannesBrx:
**Short description of changes**

This is a quick fix for issue #2066 and has been discussed there. It redundantly check whether the feedback detection is enabled in the UI. An assumption is that the timer for resetting the feedback detection 3 seconds after connecting does not work reliably on some operating systems.

**Context: Fixes an issue?**

Fixes: #2066 

**Does this change need documentation? What needs to be documented and how?**

No documentation required.

**Status of this Pull Request**

**What is missing until this pull request can be merged?**

Maybe the timer could be replaced by a comparison with the connection time plus the feedback protection time as suggested by @npostavs and @pljones.

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [ ] I tested my code and it does what I want
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [ ] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

PR #2120, opened 2021-11-12
Add missing initialisation of bDetectFeedback

### LINKED ISSUE #2119: Feedback detection may happen repeatedly on first connection if disabled.
_Originally posted by @softins in https://github.com/jamulussoftware/jamulus/discussions/2066#discussioncomment-1632868_

1. `bDetectFeedback` only gets set to `true`, and the 3-second timer started, IF `bEnableFeedbackDetection` is also `true`. Since the OP said the latter setting is already `false`, this can't be happening.
2. `bDetectFeedback` only gets set to `false` within `CClientDlg::OnTimerDetectFeedback()` and `CClientDlg::Disconnect()`. Once this has happened, the level check and subsequent popup CAN'T happen.
3. However, `bDetectFeedback` is NOT getting initialised when the `CClientDlg` object is instantiated. This means that, depending on the OS and runtime environment, it could randomly contain any value when the program is started up. UNTIL the first time the client disconnects from a server, after which it will be fine.

So this means that after starting the program, and connecting for the first time to a server, the problem may or may not occur, depending on whether the uninitialised `bDetectFeedback` happens to be zero or non-zero. After disconnecting from the server, it should then be possible to connect to any server without the problem occurring for the rest of the time the program is running. Exiting and restarting the program will allow the issue to occur again for the duration of the next connection.

So the correct fix is to initialise `bDetectFeedback` to `false` in the constructor for `CClientDlg`. Will raise PR.

softins:
Due to missing initialisation, if feedback detection is _disabled_, then on the first connection after starting the program,
it is possible that all loud noises get detected as feedback. This PR fixes that by adding the initialisation.

**Context: Fixes an issue?**

Fixes #2119

No, bug fix only.

Working

Ready to merge.

## Checklist

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -39,6 +39,7 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     pSettings ( pNSetP ),
     bConnectDlgWasShown ( false ),
     bMIDICtrlUsed ( !strMIDISetup.isEmpty() ),
+    bDetectFeedback ( false ),
     bEnableIPv6 ( bNEnableIPv6 ),
     eLastRecorderState ( RS_UNDEFINED ), // for SetMixerBoardDeco
     eLastDesign ( GD_ORIGINAL ),         //          "


pljones (approved):

softins:
This fix renders #2096 unnecessary

ann0see (approved):
Didn’t test it but looks ok.

PR #1767, opened 2021-05-27
Use dch instead of hack to generate changelog

ann0see:
See: https://github.com/nefarius2001/jamulus/blob/autobuild/stable2/distributions/build-debian-package-auto.sh

Probably a lot more could be improved here.

PR #2138, opened 2021-11-26
Use dch instead of hack to generate changelog

npostavs:

Use dch to generate Debian changelog instead of handrolled shell calls

**Context: Fixes an issue?**

This is an alternate version of #1767

No.

Working implementation, a sample output is [changelog.dch.log](https://github.com/jamulussoftware/jamulus/files/7620907/changelog.dch.log). (updated with Nov 29th changes)

~~Possibly the "initial release" line needs to be removed somehow.~~

## Checklist

- [X] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [X] I tested my code and it does what I want
- [X] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [X] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [X] I've filled all the content above

### DIFF
--- a/.github/actions_scripts/getChangelog.pl
+++ b/.github/actions_scripts/getChangelog.pl
@@ -5,15 +5,27 @@ use warnings;
 # get the file path
 my $ChangeLogFile = $ARGV[0];
 # get the version
-my $Version = quotemeta $ARGV[1];
-# open the file (slurp mode): https://perlmaven.com/slurp
+my $Version = $ARGV[1];
+# open the file
 open my $fh, '<', $ChangeLogFile or die;
-$/ = undef;
-my $ChangeLog = <$fh>;
-close $fh;
 
-if ( $ChangeLog =~ m/^### $Version (.*?)$(.*?)^###/sm ) {
-    print $2;
-} else {
-    print "No changelog found for this version.";
+my $logversion = "";
+my $entry;
+while (my $line = <$fh>) {
+   if ($line =~ /^### (\S+)[^#]*###$/m) {
+       # The version for this section of the ChangeLog
+       $logversion = $1;
+       next;
+   }
+   if ($logversion eq $Version) {
+       if ($line =~ /^- (.*)$/) { # beginning of entry
+           $entry = $1;
+       } elsif ($line =~ /^ ( \S.*)$/) { # continuation of entry
+           $entry .= $1;
+       } else { # otherwise, separator between entries
+           print "${entry}\n" if $entry;
+           $entry = undef;
+       }
+   }
 }
+close $fh;
--- a/.github/actions_scripts/getChangelog.pl
+++ b/.github/actions_scripts/getChangelog.pl
@@ -2,12 +2,12 @@
 
 use strict;
 use warnings;
-# get the file path
-my $ChangeLogFile = $ARGV[0];
-# get the version
-my $Version = $ARGV[1];
+# get arguments
+my ($ChangeLogFile, $Version, $FormatOption) = @ARGV;
 # open the file
 open my $fh, '<', $ChangeLogFile or die;
+
+my $SingleLineEntries = ($FormatOption or '') eq '--line-per-entry';
 
 my $logversion = "";
 my $entry;
@@ -18,13 +18,17 @@ while (my $line = <$fh>) {
        next;
    }
    if ($logversion eq $Version) {
-       if ($line =~ /^- (.*)$/) { # beginning of entry
-           $entry = $1;
-       } elsif ($line =~ /^ ( \S.*)$/) { # continuation of entry
-           $entry .= $1;
-       } else { # otherwise, separator between entries
-           print "${entry}\n" if $entry;
-           $entry = undef;
+       if ($SingleLineEntries) {
+           if ($line =~ /^- (.*)$/) { # beginning of entry
+               $entry = $1;
+           } elsif ($line =~ /^ ( \S.*)$/) { # continuation of entry
+               $entry .= $1;
+           } else { # otherwise, separator between entries
+               print "${entry}\n" if $entry;
+               $entry = undef;
+           }
+       } else {
+           print $line;
        }
    }
 }
--- a/.github/actions_scripts/getChangelog.pl
+++ b/.github/actions_scripts/getChangelog.pl
@@ -11,13 +11,20 @@ my $SingleLineEntries = ($FormatOption or '') eq '--line-per-entry';
 
 my $logversion = "";
 my $entry;
+my $matchedLogversion = 0;
 while (my $line = <$fh>) {
    if ($line =~ /^### (\S+)[^#]*###$/m) {
        # The version for this section of the ChangeLog
+       if ($matchedLogversion) {
+           # We've processed the lines corresponding to $Version, and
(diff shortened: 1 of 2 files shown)


softins:
Firstly, thank you @npostavs for addressing the build warnings about the `debian/changelog` entry errors, which had been bothering me for some time. However, I think the proposed change is incomplete, as it introduces other issues:

1. The modification to `.github/actions_scripts/getChangelog.pl` changes the output from MarkDown to a single line per change.

    This is certainly useful for passing each line to `dch`, but the `getChangeLog.pl` script is also used in `.github/actions_scripts/analyse_git_reference.py` to create a MarkDown file containing the version release notes. `analyse_git_reference.py` is called from `.github/workflows/autobuild.yml` in the section `create_release`.

    The new output format will not be compatible with this requirement.

    Maybe instead leave `getChangeLog.pl` unchanged, and create instead a new script (e.g. `getChangeLines.pl`?) for use with `dch`, which could be called from `distributions/build-debian-package-auto.sh`? Or else provide a parameter to the script?

2. The `debian/changelog` produced by the new version of `build-debian-package-auto.sh` changes the maintainer information from what was used previously. After some experimentation, I found that I preferred the output if I changed the `dch` lines as follows:

    ```
    DEBFULLNAME=GitHubActions DEBEMAIL=noemail@example.com dch --nomultimaint --newversion "${VERSION}" ''
    ```

    (or some appropriate values - the above matches the old script) and:

    ```
    dch --nomultimaint "$entry"
    ```

softins:
After further experimentation, this is the current state of my `build-debian-package-auto.sh`:

```
#!/bin/sh -e

# Create deb files

cp -r debian ..
cd ..

# get the jamulus version from pro file
VERSION=$(cat Jamulus.pro | grep -oP 'VERSION = \K\w[^\s\\]*')

export DEBFULLNAME=GitHubActions DEBEMAIL=noemail@example.com

# Generate Changelog
echo -n generating changelog
rm -f debian/changelog
dch --create --package jamulus --empty --newversion "${VERSION}" ''
perl .github/actions_scripts/getChangelog.pl ChangeLog "${VERSION}" | while read entry
do
    echo -n .
    dch "$entry"
done
echo

echo "${VERSION} building..."

sed -i "s/é&%JAMVERSION%&è/${VERSION}/g" debian/control

debuild --preserve-env -b -us -uc
```

It uses `--create`, `--package`, `--newversion` and `--empty` to create a new `debian/changelog` file. The use of `--empty` causes a warning, which the man page says is expected, but it does avoid the spurious "Initial version" log entry.

It exports `DEBFULLNAME` and `DEBEMAIL` so they are available to all calls of `dch`. This also means `--nomultimaint` is not necessary.

npostavs:
> 1. but the `getChangeLog.pl` script is also used in `.github/actions_scripts/analyse_git_reference.py`

Ah, I missed that.

> Or else provide a parameter to the script?

I went with this option.

> After further experimentation, this is the current state of my `build-debian-package-auto.sh`:

Added to the PR.

ann0see:
@npostavs what is the Next action on this? Review, testing?

npostavs:
Yeah, I believe the code is ready, and I have addressed all of @softin's concerns.

.github/actions_scripts/getChangelog.pl
```diff
@@ -2,18 +2,34 @@
...
 open my $fh, '<', $ChangeLogFile or die;
-$/ = undef;
-my $ChangeLog = <$fh>;
-close $fh;
 
-if ( $ChangeLog =~ m/^### $Version (.*?)$(.*?)^###/sm ) {
-    print $2;
-} else {
-    print "No changelog found for this version.";
+my $SingleLineEntries = ($FormatOption or '') ne '--line-per-entry';
```
softins, line 10: I've just tested this and it appears to be the wrong way up.
```suggestion
my $SingleLineEntries = ($FormatOption or '') eq '--line-per-entry';
```

softins (changes requested):
Only the one change required. Everything else looks ok, and I have successfully build a deb.

.github/actions_scripts/getChangelog.pl
npostavs: Oh, I had fixed that in my working copy where I was testing, but forgot to commit it. Done now.

distributions/build-debian-package-auto.sh
```diff
@@ -2,30 +2,24 @@
...
-echo "" >> debian/changelog
-echo " -- GitHubActions <noemail@example.com> ${DATE} +0001" >> debian/changelog
-echo "" >> debian/changelog
-cat distributions/debian/changelog >> debian/changelog
-
-# patch the control file
-# cp the modified control file here
-
-cp distributions/autobuilddeb/control debian/control
+export DEBFULLNAME=GitHubActions DEBEMAIL=noemail@example.com
```
ann0see, line 11: Concerning the E-Mail, is there a standard mail saying basically "we don’t publish an e-mail address"?

ann0see, line 11:
```suggestion
export DEBFULLNAME=JamulusAutobuild DEBEMAIL=noreply@example.com
```

ann0see (commented):
Just a minor thought. It’s still not optimal, but probably better then now.

distributions/build-debian-package-auto.sh
softins: The original corresponds to what was in the old method. I would be inclined not to change that in this PR, but to consider it separately.

softins (approved):
Looks fine now, thank you!

ann0see (commented):
Did you also test if the automated change log detection for GH releases still works? Seems ok, but better test it.

Urgh. Didn't read... Should be ok.

.github/actions_scripts/getChangelog.pl
```diff
@@ -2,18 +2,34 @@
...
 
-if ( $ChangeLog =~ m/^### $Version (.*?)$(.*?)^###/sm ) {
-    print $2;
-} else {
-    print "No changelog found for this version.";
+my $SingleLineEntries = ($FormatOption or '') eq '--line-per-entry';
+
+my $logversion = "";
+my $entry;
+while (my $line = <$fh>) {
```
ann0see, line 14: I *think* we should stop looping through the file if we already found the correct version ($logversion changed to the next version). Then we could also check if a version wasn't found as soon as we reached the EOF and output some error message.

ann0see: @npostavs could you please insert a break statement to stop looping through unneeded lines?

ann0see (changes requested):

.github/actions_scripts/getChangelog.pl
npostavs: Oh yep, done.

ann0see (approved):
Looks good.

.github/actions_scripts/getChangelog.pl
ann0see: Thanks!

PR #2151, opened 2021-12-07
Keep the list of free channels in sorted order

### LINKED ISSUE #2144: Server channel numbers are not reset if last client to disconnect is > 0
**Describe the bug**
jamulus-server 3.8.1 apparently introduced a bug where channel numbers are not reset when server goes idle, except when last client disconnecting is channel 0.  This shows up when using MIDI control of client.

**To Reproduce**
restart jamulus-server 3.8.1:
connect client "A"
connect client "B" which is started with --crtlmidich arg.  name will show as "1:B" in gui.
disconnect client "A"
disconnect client "B".  Server is now idle,
connect client "B" name will show as "1:B" in GUI. Channel numbers were not reset on idle.
connect client "A"
disconnect client "B"
disconnect client "A".  Server is now idle,
connect client "B".  name will show as "0::B" in GUI.  Channel numbers were reset on idle if last client to disconnect was channel 0.

This causes problems with midi control.
Under jamulus-server 3.8.0, channel ids are always reset on server idle.

edit:  I could not reset the public server that gave me channel 12, even by connecting with a client that got channel 0 and disconnecting.  So exact contditions required to reset channel numbers appear to be more complex.

**Expected behavior**

channel numbers should always be reset to start at 0 when server goes idle,

**Screenshots**

after connecting to an idle public 3.8.1 server.  In this case I got channel 12, so 8-channel MIDI controller cannot operate.
image

**Operating system**

linux (raspberry-pi os, ubuntu)

**Version of Jamulus**

3.8.1

**Additional context**

3.8.0 server does not exhibit this problem

softins:

Maintain the list of free channel IDs in sorted order, so that the lowest available channel ID will be used for a new client.

The fix has two parts, the first of which is not really necessary, but makes the output of `DumpChannels()` nicer:

* Reverse the polarity of the comparison in the binary search, so that the list of active channel IDs has the lowest ID first (this should have been done originally).
* When moving the channel IDs downwards in the list while freeing a channel, instead of stopping at the end of the active channels, keep going until the correct ordered place is found to insert the freed channel ID.

**Context: Fixes an issue?**

Fixes #2144 

No documentation - internal enhancement only.

Tested and ready to merge.

Nothing

## Checklist

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1502,7 +1502,7 @@ int CServer::FindChannel ( const CHostAddress& CheckAddr, const bool bAllowNew )
             return vecChannelOrder[t];
         }
 
-        if ( cmp < 0 )
+        if ( cmp > 0 )
         {
             l = t + 1;
         }
@@ -1574,13 +1574,14 @@ void CServer::FreeChannel ( const int iCurChanID )
         {
             --iCurNumChannels;
 
-            // move channel IDs down by one starting at the bottom and working up
-            while ( i < iCurNumChannels )
+            // move channel IDs down by one starting at the freed channel and working up the active channels
+            // and then the free channels until its position in the free list is reached
+            while ( i < iCurNumChannels || ( i + 1 < iMaxNumChannels && vecChannelOrder[i + 1] < iCurChanID ) )
             {
                 int j              = i++;
                 vecChannelOrder[j] = vecChannelOrder[i];
             }
-            // put deleted channel at the end ready for re-use
+            // put deleted channel in the vacated position ready for re-use
             vecChannelOrder[i] = iCurChanID;
 
             // DumpChannels ( __FUNCTION__ );


softins:
The operation of the list can be observed for testing by temporarily uncommenting the two calls to `DumpChannels()` in `src/server.cpp`.

src/server.cpp
```diff
@@ -1575,12 +1575,13 @@ void CServer::FreeChannel ( const int iCurChanID )
...
             // move channel IDs down by one starting at the bottom and working up
-            while ( i < iCurNumChannels )
+            // also keep the free channels in sorted order by moving lower-numbered ones down
+            while ( i < iCurNumChannels || ( i + 1 < iMaxNumChannels && vecChannelOrder[i + 1] < iCurChanID ) )
             {
                 int j              = i++;
                 vecChannelOrder[j] = vecChannelOrder[i];
             }
-            // put deleted channel at the end ready for re-use
+            // put deleted channel in the correct position ready for re-use
```
pljones, line 1584: This all now seems very cumbersome.  If it's going to sit _somewhere_ in `vecChannelOrder`, why not just leave it where it was?

softins: Because the binary search relies on all the active channels being listed before all the free ones. All we are changing is how many IDs we shunt down the list when freeing, so that the freed ID is placed in order instead of just at the start of the free list. Doesn't seem cumbersome to me at all :)

pljones: So you're saying it was (allocated | free):
```
1 2 3 4 5   ||
1 3 4 5 | 2 ||
1 3 4 | 5 2 ||
3 4 | 1 5 2 ||
```
and now it's
```
1 2 3 4 5   ||
1 3 4 5 | 2 ||
1 3 4 | 2 5 ||
3 4 | 1 2 5 ||
```
?  I think the comment doesn't explain anything.

softins: Yes, exactly that. I'll change the comment to say "put deleted channel in the _vacated_ position ready for re-use"

src/server.cpp
```diff
@@ -1575,12 +1575,13 @@ void CServer::FreeChannel ( const int iCurChanID )
             --iCurNumChannels;
 
             // move channel IDs down by one starting at the bottom and working up
```
pljones, line 1577: > at the bottom

No it doesn't.  It starts at `i` where `vecChannelOrder[i] == iCurChanID` (i.e. the freed channel).

so rather than
```
// move channel IDs down by one starting at the bottom and working up
// also keep the free channels in sorted order by moving lower-numbered ones down
```
have
```
// move channel IDs down by one starting at the freed channel and working up the current channels
// and then free channels until its position in the free list is reached
```

softins: OK, yes, to be accurate, you're right. When I originally wrote the comment I was just talking about the direction of scanning (bottom to top when freeing, top to bottom when making space to allocate). I'll update the comment.

pljones (approved):

ann0see (approved):
Looks ok.

PR #2168, opened 2021-12-10
Don't imply that multiple drivers have errors

ann0see:

If there's only one driver with an error message, there's no need to imply that multiple devices have errors. This change is done to simplify the error message.

The following is not exactly what my (fixed) code looks like.
error message

**Context: Fixes an issue?**

Idea from https://github.com/jamulussoftware/jamulus/discussions/2163#discussioncomment-1788094
No

Draft. It remains to test what happens if lNumDevs > 1.
Untested since it's just a quick fix. Will test it by checking the artifacts.
## Checklist

### DIFF
--- a/src/soundbase.cpp
+++ b/src/soundbase.cpp
@@ -174,7 +174,7 @@ QString CSoundBase::SetDev ( const QString strDevName )
                 sErrorMessage += "<b>" + GetDeviceName ( i ) + "</b>: " + vsErrorList[i] + "<br><br>";
             }
 
-#ifdef _WIN32
+#if defined( _WIN32 ) && !defined( WITH_JACK )
             // to be able to access the ASIO driver setup for changing, e.g., the sample rate, we
             // offer the user under Windows that we open the driver setups of all registered
             // ASIO drivers
--- a/src/soundbase.cpp
+++ b/src/soundbase.cpp
@@ -167,23 +167,32 @@ QString CSoundBase::SetDev ( const QString strDevName )
         if ( !vsErrorList.isEmpty() )
         {
             // create error message with all details
-            QString sErrorMessage = "<b>" + QString ( tr ( "No usable %1 audio device found." ) ).arg ( strSystemDriverTechniqueName ) +
-                                    "</b><br><br>" + tr ( "These are all the available drivers with error messages:" ) + "<ul>";
+            QString sErrorMessage = tr ( "<b>No usable %1 audio device found.</b><br><br>" ).arg ( strSystemDriverTechniqueName );
 
             for ( int i = 0; i < lNumDevs; i++ )
             {
-                sErrorMessage += "<li><b>" + GetDeviceName ( i ) + "</b>: " + vsErrorList[i] + "</li>";
+                sErrorMessage += "<b>" + GetDeviceName ( i ) + "</b>: " + vsErrorList[i] + "<br><br>";
             }
-            sErrorMessage += "</ul>";
 
 #ifdef _WIN32
             // to be able to access the ASIO driver setup for changing, e.g., the sample rate, we
             // offer the user under Windows that we open the driver setups of all registered
             // ASIO drivers
-            sErrorMessage += "<br/>" + tr ( "Do you want to open the ASIO driver setup to try changing your configuration to a working state?" );
+            sErrorMessage += "<br>" + tr ( "You may be able to fix errors in the driver settings." );
 
-            if ( QMessageBox::Yes == QMessageBox::information ( nullptr, APP_NAME, sErrorMessage, QMessageBox::Yes | QMessageBox::No ) )
+            // for Windows use a QMessageBox to show the error
+            QMessageBox  qmASIOWarningBox;
+            QPushButton* btnASIOSettings = qmASIOWarningBox.addButton ( tr ( "Open ASIO settings" ), QMessageBox::AcceptRole );
+
+            qmASIOWarningBox.addButton ( QMessageBox::Cancel );
+            qmASIOWarningBox.setText ( sErrorMessage );
+            qmASIOWarningBox.setIcon ( QMessageBox::Warning );
+            qmASIOWarningBox.exec();
+
+            if ( qmASIOWarningBox.clickedButton() == btnASIOSettings )
             {
+                // TODO: This or some related code aborts the app even if the driver is valid after user changes.
+                // This should be handled differently. For reference, please see https://github.com/jamulussoftware/jamulus/pull/2168
                 LoadAndInitializeFirstValidDriver ( true );
             }
 


gilgongo:
Do you think we can remove the ambiguity in the first part of the message? The driver(s) in question my in fact be perfectly usable, it's just that there is a problem with them that may or may not be temporary. 

So if only one driver is erroring, just present this error message with an option to open the settings:

image

If multiple drivers are erroring, try to let the errors speak for themselves:

image

In general, long dialogue messages tend not to get read though. I find that had to accept, and don't really know why it is, but it's a very observable and common thing. So the other approach might be not to show the errors themselves and instead just have a call to action to open the settings? Depends on how helpful the messages are I guess:

image

pljones:
I like the first two.  In the second, I'd add "If needed" at the start of the guidance.

I don't like the third one because it's far too vague to decide what to do.  "Maybe something is wrong with one of your sound devices but I'm not going to tell you which one or what the problem was."  Thanks....

gilgongo:
@pljones Ah yes the last one doesn't indicate which card has the fail. OK that's no good then :-)

pljones:
> BTW why "if needed"? Seems redundant. Or implies you may not want to fix the fact that Jamulus doesn't work for you?

Because you list drivers that failed.  If I'm not using those drivers, why would I want to correct them?  The
> You may be able to correct these in the driver settings

makes it sound like it _could_ **need** doing.  It's better to make it clear the user needs to decide what steps to take: we're not telling them to fix it - just that something went wrong that _could_ need fixing.

gilgongo:
Oh I see - in "plural" errors case, "If needed... " would be clearer there but it still seems odd in the "singular" error case to me.

Agree about the first and last line prominence. Generally though, popups that spawn popups are seen as a bad thing as you can lose track or the parent or child if you accidentally click on something (and I think it's a bit of a mare for those using assistive tech). Also don't forget the ASIO settings is yet another "popop" in this context. But I guess if there's more than about three error messages it might be OK (the best solution would be an expandable panel within the dialogue but I assume that's not going to be possible).

> makes it sound like it should be looked at and may be fixable.

I disagree. The sense is that may or may not be issues to fix, and our lack of specificity clues this as well. Any other wording will be unacceptably long and complicated and we'll be back where we started.

src/soundbase.cpp
```diff
@@ -166,22 +166,38 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
 #ifdef _WIN32
             // to be able to access the ASIO driver setup for changing, e.g., the sample rate, we
             // offer the user under Windows that we open the driver setups of all registered
             // ASIO drivers
-            sErrorMessage += "<br/>" + tr ( "Do you want to open the ASIO driver setup to try changing your configuration to a working state?" );
+            sErrorMessage += "<br/>" + tr ( "You may be able to fix errors in the driver settings." );
+            // for Windows use a QMessageBox to show the error
+
+            QMessageBox mbWinErrorBox = QMessageBox::information ( nullptr, APP_NAME, sErrorMessage, QMessageBox::Cancel );
+            QPushButton*             btnASIOSettings = mbWinErrorBox.addButton ( tr ( "Open ASIO settings" ), QMessageBox::AcceptRole );
```
ann0see, line 198: Sorry for being unsure about the syntax here.

src/soundbase.cpp
```diff
@@ -166,30 +166,38 @@ QString CSoundBase::SetDev ( const QString strDevName )
         if ( !vsErrorList.isEmpty() )
         {
             // create error message with all details
-            QString sErrorMessage = "<b>" + QString ( tr ( "No usable %1 audio device found." ) ).arg ( strSystemDriverTechniqueName ) + "</b><br>";
+            QString sErrorMessage = "<b>";
```
pljones, line 169: Just declare it here - the floating `"<b>"` makes the code hard to follow.

ann0see: Ok. Still struggeling with the syntax though.

src/soundbase.cpp
```diff
@@ -166,30 +166,38 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
                 {
-                    sErrorMessage += "<li><b>" + GetDeviceName ( i ) + "</b>: " + vsErrorList[i] + "</li>";
+                    sErrorMessage += "<b>" + GetDeviceName ( i ) + "</b>: " + vsErrorList[i] + "<br>";
                 }
-                sErrorMessage += "</ul>";
             }
             else
             {
-                 sErrorMessage += "<b>" + GetDeviceName ( 0 ) + "</b>: " + vsErrorList[0] ;
+                sErrorMessage += tr ( "Couldn't initialize audio driver (using %1)</b><br>"
```
pljones, line 184: Line 173:
> `"Sound driver errors (using %1) found:</b><br>"`

Line 184:
> `"Couldn't initialize audio driver (using %1)</b><br>"`

Use the same string in both places.  Ideally the less technical one.  The first doesn't need to end with ":" as it no longer introduces a list.

Maybe
> `"Could not find a sound driver (using %1) that could be used.</b><br>"`

(and do away entirely with `"Your sound device is not compatible in its current state."` as it would then be somewhat redundant).

src/soundbase.cpp
```diff
@@ -166,30 +166,38 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
+                sErrorMessage += "<b>" + GetDeviceName ( 0 ) + "</b>: " + vsErrorList[0];
             }
 
-
 #ifdef _WIN32
             // to be able to access the ASIO driver setup for changing, e.g., the sample rate, we
             // offer the user under Windows that we open the driver setups of all registered
             // ASIO drivers
-            sErrorMessage += "<br/>" + tr ( "You might be able to fix this error by changing settings of your ASIO driver. Do you want to open these settings now?" );
+            sErrorMessage += "<br/>" + tr ( "You may be able to fix errors in the driver settings." );
```
pljones, line 194: This seems worthless.

src/soundbase.cpp
```diff
@@ -166,30 +166,38 @@ QString CSoundBase::SetDev ( const QString strDevName )
         if ( !vsErrorList.isEmpty() )
```
pljones, line 166: Lines 154 to 165 are irrelevant if this we enter this `if`.  They should be placed after the block.

src/soundbase.cpp
```diff
@@ -166,30 +166,38 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
+            sErrorMessage += "<br/>" + tr ( "You may be able to fix errors in the driver settings." );
+            // for Windows use a QMessageBox to show the error
+
+            QMessageBox::information mbWinErrorBox ( nullptr, APP_NAME, sErrorMessage, QMessageBox::Cancel );
+            QPushButton*             btnASIOSettings = mbWinErrorBox.addButton ( tr ( "Open ASIO settings" ), QMessageBox::AcceptRole );
 
-            if ( QMessageBox::Yes == QMessageBox::information ( nullptr, APP_NAME, sErrorMessage, QMessageBox::Yes | QMessageBox::No ) )
+            if ( mbWinErrorBox.clickedButton() == btnASIOSettings )
             {
                 LoadAndInitializeFirstValidDriver ( true );
```
pljones, line 202: `vsErrorList = LoadAndInitializeFirstValidDriver ( true );` and then check the list size.  If it's zero (and I've not checked the code to make sure I'm right here...), `return strReturn` rather than continuing.

ann0see: I think there's more work needed then just that. See the commit I mentioned: Probably it's in a deeper level.

ann0see:
Ok. Maybe something like that would be ok:

image

src/soundbase.cpp
```diff
@@ -166,24 +166,36 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
-            if ( QMessageBox::Yes == QMessageBox::information ( nullptr, APP_NAME, sErrorMessage, QMessageBox::Yes | QMessageBox::No ) )
+            if ( qmASIOWarningBox.clickedButton() == btnASIOSettings )
             {
-                LoadAndInitializeFirstValidDriver ( true );
+                vsErrorList = LoadAndInitializeFirstValidDriver ( true );
+            }
+
+            if ( vsErrorList.isEmpty() ) {
+                return strReturn;
             }
```
ann0see, line 199: As mentioned before, I think this is not enough.

src/soundbase.cpp
```diff
@@ -166,24 +166,36 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
-            sErrorMessage += "<br/>" + tr ( "Do you want to open the ASIO driver setup to try changing your configuration to a working state?" );
+            sErrorMessage += "<br>" + tr ( "You may be able to fix errors in the driver settings." );
+            // for Windows use a QMessageBox to show the error
+            QMessageBox qmASIOWarningBox;
+            QPushButton *btnASIOSettings = qmASIOWarningBox.addButton ( tr ( "Open ASIO settings" ), QMessageBox::AcceptRole );
+
+            qmASIOWarningBox.addButton ( QMessageBox::Cancel );
+            qmASIOWarningBox.setText ( sErrorMessage );
+            qmASIOWarningBox.setIcon ( QMessageBox::Warning );
+            qmASIOWarningBox.exec();
```
ann0see, line 190: Probably this can be simplified somehow

pljones: I think there's a constructor that allows most of those things to be set?

src/soundbase.cpp
pljones: It looks like it's what's needed and removes my concern at https://github.com/jamulussoftware/jamulus/pull/2168/files#r770944946

src/soundbase.cpp
ann0see: Yes, I think/thought so too. However I couldn't find it in the documentation (or the code gave an error). The problem was with adding a "custom" button.

src/soundbase.cpp
ann0see: Might the problem be that this code is run once at startup and also during runtime if someone changes the sound devices?

src/soundbase.cpp
```diff
@@ -166,24 +166,35 @@ QString CSoundBase::SetDev ( const QString strDevName )
         if ( !vsErrorList.isEmpty() )
         {
             // create error message with all details
-            QString sErrorMessage = "<b>" + QString ( tr ( "No usable %1 audio device found." ) ).arg ( strSystemDriverTechniqueName ) +
-                                    "</b><br><br>" + tr ( "These are all the available drivers with error messages:" ) + "<ul>";
+            QString sErrorMessage = tr ( "<b>Could not find compatible sound driver (using %1).</b><br><br>" ).arg ( strSystemDriverTechniqueName );
```
gilgongo, line 169: The "... (using ASIO)." seems a bit confusing here as it implies a driver might be found using something else. Is there a reason we're saying this?

pljones: It won't necessarily say "ASIO" - it provides meaningful information for diagnosing a problem preventing the use of the program.

src/soundbase.cpp
pljones: Before line 197 was introduced, the variable set there wasn't referenced until after this if block.  However, now line 197 _is_ there, it makes sense.

src/soundbase.cpp
gilgongo: OK, but we have a button that says "Open ASIO settings", so is that button conditional on what `%1` is? This PR is to help people who aren't technical deal with the problem. If we qualify an error message in this way, they'd be excused for wondering why it's qualified, should they be using something else, etc.

pljones: No -- it only appear on Windows.  (It should also only appear for ASIO -- that needs fixing.)

gilgongo: I still think "Could not find compatible sound driver (using ASIO)" feels a bit odd. Like it's expecting them to switch to something that isn't ASIO where the problem might not arise.

pljones: But that's entirely true.  Switching to Jamulus for Jack on Windows might solve the problem.

gilgongo: It would be less mysterious to just make it the subject in that case and say, "Could not find a compatible ASIO sound driver." That is, unless we want also to say "Maybe you should try using [other system]." which I doubt.

src/soundbase.cpp
```diff
@@ -166,24 +166,35 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
 
             for ( int i = 0; i < lNumDevs; i++ )
             {
-                sErrorMessage += "<li><b>" + GetDeviceName ( i ) + "</b>: " + vsErrorList[i] + "</li>";
+                sErrorMessage += "<b>" + GetDeviceName ( i ) + "</b>: " + vsErrorList[i] + "<br><br>";
             }
-            sErrorMessage += "</ul>";
 
 #ifdef _WIN32
```
pljones, line 176: Is this defined if we're running JACK on Windows?  If so, it's the wrong symbol.

src/soundbase.cpp
pljones: Maybe go back to `No usable %1 audio device found.` then?

gilgongo: That reads better than a parenthesised suffix (I think). 

BTW re. "audio device", it's probably the subject of a different ticket, but for such an important thing we should keep to a standard term. The [style guide](https://jamulus.io/contribute/Style-and-Tone) says we should be using "sound hardware" here but "audio device" does feel a bit more conventional somehow.

pljones:
Before approval, this PR should show behavour when:
- results for _all_ platforms (iOS, Android, MacOS, Linux, Window JACK and Windows ASIO)
- results when the final `throw CGenErr ( sErrorMessage );` occurs
- results when the final `return strReturn;` occurs
- okay to say "platform tested but could not achieve failure", though, I suppose...

This is needed to ensure it remains sane...

src/soundbase.cpp
ann0see: I think so. Similar to the ASIO Device setting button.

ann0see:
I think it still throws an exception without returning. We‘d need some more debugging, but that would be out of scope of this PR.

src/soundbase.cpp
pljones: OK, probably needs a bug raised to fix it, then.

src/soundbase.cpp
ann0see: Ok. Fixed that.

> The whole flow is really bad, I think - huge block of platform-specific code at level that should be platform-independent.

I'm a bit unsure, but I remember there was a discussion about fixing this. Wasn't @npostavs involved?

npostavs: I guess pljones' remark at https://github.com/jamulussoftware/jamulus/pull/1378#issuecomment-812866670 is similar.

ann0see:
No longer sure if this is moving into the right direction. Probably the exception handling needs to be part of another PR.

src/soundbase.cpp
```diff
@@ -166,24 +166,35 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
 
-            if ( QMessageBox::Yes == QMessageBox::information ( nullptr, APP_NAME, sErrorMessage, QMessageBox::Yes | QMessageBox::No ) )
+            qmASIOWarningBox.addButton ( QMessageBox::Cancel );
+            qmASIOWarningBox.setText ( sErrorMessage );
+            qmASIOWarningBox.setIcon ( QMessageBox::Warning );
+            qmASIOWarningBox.exec();
+
+            if ( qmASIOWarningBox.clickedButton() == btnASIOSettings )
+            {
+                vsErrorList = LoadAndInitializeFirstValidDriver ( true );
```
ann0see, line 192: I think we should revert this to the original state and do functional changes here later.
Related to https://github.com/jamulussoftware/jamulus/issues/2238

src/soundbase.cpp
ann0see: Ok. Fixed this now.

src/soundbase.cpp
ann0see: Done. This should be handled in another PR.

src/soundbase.cpp
```diff
@@ -167,23 +167,32 @@ QString CSoundBase::SetDev ( const QString strDevName )
...
+
+            qmASIOWarningBox.addButton ( QMessageBox::Cancel );
+            qmASIOWarningBox.setText ( sErrorMessage );
+            qmASIOWarningBox.setIcon ( QMessageBox::Warning );
+            qmASIOWarningBox.exec();
+
+            if ( qmASIOWarningBox.clickedButton() == btnASIOSettings )
             {
+                // TODO: This or some related code aborts the app even if the driver is valid after user changes.
+                // This should be handled differently. For reference, please see https://github.com/jamulussoftware/jamulus/pull/2168
```
ann0see, line 195: @pljones wrote a TODO comment here.

ann0see (commented):
Rebased this PR.

pljones:
Overall, far too much of the code appears to be trying to handle Windows-ASIO specific issues in `src/soundbase.cpp`, rather than `windows/sound.cpp`.  I'd rather _any_ improvements were focussed on removing ASIO code from `src/*`.

ann0see (commented):
I think the text part should be split out into another PR (and the addition of the open ASIO Settings button overthought again)

ann0see:
Closing in favour of https://github.com/jamulussoftware/jamulus/pull/2496 we should probably split this PR.

PR #2173, opened 2021-12-15
Use the default server name in the mixer

pljones:

Edge-case for user using no server details and clicking "connect" from the connect window.  They could be looking at _any_ of the directories when they hit connect.  For example:
image

Hitting "Connect" used to title the mixer "Directory Server" (or translation).  Now it titles it with the default directory name as more of a clue:
image

No.

**Translations** needed.

Working as expected.

Should just work (except for translations changing).

## Checklist

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -716,13 +716,13 @@ void CClientDlg::OnConnectDlgAccepted()
             strMixerBoardLabel = strSelectedAddress;
 
             // special case: if the address is empty, we substitute the default
-            // directory server address so that a user which just pressed the connect
+            // directory address so that a user who just pressed the connect
             // button without selecting an item in the table or manually entered an
             // address gets a successful connection
             if ( strSelectedAddress.isEmpty() )
             {
                 strSelectedAddress = DEFAULT_SERVER_ADDRESS;
-                strMixerBoardLabel = tr ( "Directory Server" );
+                strMixerBoardLabel = tr ( "%1 Directory" ).arg ( DirectoryTypeToString ( AT_DEFAULT ) );
             }
         }
 


pljones:
Awaiting this over on https://github.com/pljones/jamulus
* patch/ocd-reorder-of-settings
* patch/ocd-reorder-of-serverdlg-bits
These two pretty much don't change anything, just put the code through a "mental sort", so I could follow it better.  Mostly because they're areas I'm planning to change... and indeed...
* feature/server-list-persistence-ui
Adds a UI to allow the server list persistence file to be set.  Plus changing rather a lot of other stuff to make changes in the UI not need the server stopped and started.

ann0see:
How/where is AT_DEFAULT set?

Also: shouldn’t we rename the directory server at any genre to match your label?

pljones:
> How/where is AT_DEFAULT set?

`grep` is your friend.  `src/util.h` line 547.

> Also: shouldn’t we rename the directory server at any genre to match your label?

Not given this is for an edge case, I don't think.  I'm only changing it from what it was so it's clearer you're not necessarily somewhere related to what you were looking at.

ann0see:
Hmm. I think the main problem is that connecting to the Any Genre 1 directory while being in Any Genre 2 or similar is the confusing part. And this can't be fixed with this (rather visual) approach. OFC, it's much better with your approach, but I think we should rather change functionality of this button if we're in another genre...

ann0see (approved):
Although untested, the code looks simple enough to approve.

ann0see:
BTW: Thanks for the "next big thing" work on Jamulus.

softins (approved):
Just tried this out and it works fine, so I'm happy to approve, as it doesn't change existing functionality.

But I do think the existing functionality is wrong. An improvement would be to connect to the directory server of the currently-displayed list, but I think the right behaviour going forward would be: if no server in the list is selected, AND the server address box is blank, the Connect button is disabled.

pljones:
> Just tried this out and it works fine, so I'm happy to approve, as it doesn't change existing functionality.
> 
> But I do think the existing functionality is wrong. An improvement would be to connect to the directory server of the currently-displayed list, but I think the right behaviour going forward would be: if no server in the list is selected, AND the server address box is blank, the Connect button is disabled.

Yes, it's a bit weird _now_, the way it is.  (Before multiple servers, it just connected to Volker's server, of course, which sort of made sense.)

I'd like to have "Escape" work on the connect dialog along with getting rid of this "default" behavour, too.  (There was a patch to turn off "Escape" on _all_ the dialogs - I'm not entirely sure why...)

PR #2177, opened 2021-12-22
Reorder of serverdlg bits

pljones:

This starts to get to making _actual changes_ in behaviour of the server dialog code.

**I _STRONGLY_ recommend viewing each commit separately before viewing the full diff.**  The aim was to keep each change small, self-contained and then test so that I was confident I'd not broken anything along the way.

(Admittedly, skimming the full diff now, it doesn't look too bad...)

**Context: Fixes an issue?**

The aim is multifold but primarily to make it easier to tie the UI as displayed to the code that implements it.  This means code gets moved around quite a lot, meaning the diff looks like inserts/deletes, unfortunately -- hence the recommendation above.

It also fixes a few bits (like adding missing "What's This" help).
image -> image

New / changed translation(s).  May need new screenshots.  (I've been using it too long to remember, now.)

Ready to merge (IMO).

## Checklist

### DIFF
--- a/src/serverdlg.cpp
+++ b/src/serverdlg.cpp
@@ -62,9 +62,9 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
                                            "to make sure that all servers in the connect dialog server list are "
                                            "actually available." ) );
 
-    // register server status label
+    // server registration status label
     lblRegSvrStatus->setWhatsThis ( "<b>" + tr ( "Register Server Status" ) + ":</b> " +
-                                    tr ( "If the Make My Server Public check box is checked, this will show "
+                                    tr ( "If the Register Server check box is checked, this will show "
                                          "whether registration with the directory server is successful. If the "
                                          "registration failed, please choose another server list." ) );
 
--- a/src/serverdlg.cpp
+++ b/src/serverdlg.cpp
@@ -379,37 +379,36 @@ lvwClients->setMinimumHeight ( 140 );
     // check boxes
     QObject::connect ( chbRegisterServer, &QCheckBox::stateChanged, this, &CServerDlg::OnRegisterServerStateChanged );
 
-    QObject::connect ( chbStartOnOSStart, &QCheckBox::stateChanged, this, &CServerDlg::OnStartOnOSStartStateChanged );
-
     QObject::connect ( chbEnableRecorder, &QCheckBox::stateChanged, this, &CServerDlg::OnEnableRecorderStateChanged );
 
-    // delay panning
+    QObject::connect ( chbStartOnOSStart, &QCheckBox::stateChanged, this, &CServerDlg::OnStartOnOSStartStateChanged );
+
     QObject::connect ( chbEnableDelayPanning, &QCheckBox::stateChanged, this, &CServerDlg::OnEnableDelayPanningStateChanged );
 
     // line edits
-    QObject::connect ( edtCustomDirectory, &QLineEdit::editingFinished, this, &CServerDlg::OnCustomDirectoryEditingFinished );
-
     QObject::connect ( edtServerName, &QLineEdit::textChanged, this, &CServerDlg::OnServerNameTextChanged );
 
     QObject::connect ( edtLocationCity, &QLineEdit::textChanged, this, &CServerDlg::OnLocationCityTextChanged );
 
-    // combo boxes
-    QObject::connect ( cbxLocationCountry,
-                       static_cast<void ( QComboBox::* ) ( int )> ( &QComboBox::activated ),
-                       this,
-                       &CServerDlg::OnLocationCountryActivated );
+    QObject::connect ( edtCustomDirectory, &QLineEdit::editingFinished, this, &CServerDlg::OnCustomDirectoryEditingFinished );
 
+    // combo boxes
     QObject::connect ( cbxDirectoryType,
                        static_cast<void ( QComboBox::* ) ( int )> ( &QComboBox::activated ),
                        this,
                        &CServerDlg::OnDirectoryTypeActivated );
 
+    QObject::connect ( cbxLocationCountry,
+                       static_cast<void ( QComboBox::* ) ( int )> ( &QComboBox::activated ),
+                       this,
+                       &CServerDlg::OnLocationCountryActivated );
+
     QObject::connect ( cbxLanguage, &CLanguageComboBox::LanguageChanged, this, &CServerDlg::OnLanguageChanged );
 
     // push buttons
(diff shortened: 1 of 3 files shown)


pljones:
And once this is done
- https://github.com/pljones/jamulus/tree/feature/server-list-persistence-ui
This updates the Server UI to add the Server List persistence file -- and makes some other significant changes, like getting rid of the "register server" checkbox and using the drop down with a new `AT_NONE` type (meaning not registered).

pljones:
image
Something I've not done is change "Custom Directory" to (or back to...) "Custom Directory Server Address"... but in context here, I think that would be a good idea.

src/serverdlg.cpp
```diff
@@ -154,6 +154,13 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
 
     // Tab: options
 
+    // Interface Language
+    QString strWTLanguage = "<b>" + tr ( "Language" ) + ":</b> " + tr ( "Select the language to be used for the user interface." );
+    lblLanguage->setWhatsThis ( strWTLanguage );
+    cbxLanguage->setWhatsThis ( strWTLanguage );
+
+    cbxLanguage->setAccessibleName ( tr ( "Language combo box" ) );
```
softins, line 152: Is the term "combo box" generally understood by non-programmers who need accessibility?

src/serverdlg.cpp
```diff
@@ -353,6 +352,8 @@ lvwClients->setMinimumHeight ( 140 );
     // update GUI dependencies
     UpdateGUIDependencies();
 
+    UpdateRecorderStatus ( QString::null );
```
softins, line 355: For compatibility with both Qt5 and Qt6, this should be `QString()` instead of `QString::null`.

src/serverdlg.cpp
```diff
@@ -180,6 +170,15 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
                                          tr ( "Click the button to clear the currently selected recording directory. "
                                               "This will prevent recording until a new value is selected." ) );
 
+    // custom directory
+    QString strCustomDirectory = "<b>" + tr ( "Custom Directory" ) + ":</b> " +
```
softins, line 174: I agree with your comment arising from the fact that "Directory" is used in two different ways in close proximity (filesystem directory for recording, and directory server for registration). I don't think it needs changing all the way back to "Custom Directory Server Address", but I think "Custom Directory Server" would be clearer. And similarly for the associated What's This and accessibility texts.

softins (commented):
Just a few initial comments for consideration before doing a final review. Looks good in general.

src/serverdlg.cpp
pljones: Given how most of the AccessibleName values are set, I think it's the standard we're adhering to.  The server ones seem a bit eratic but the client ones seem fairly strict about the AN specifying what the widget type is.  Whether that's correct in UI/UX/Accessibility terms, I don't know.

@gilgongo can you comment?

src/serverdlg.cpp
pljones: This is just moving code...  There would be other places:
```
src/serverdlg.cpp:    UpdateRecorderStatus ( QString::null );
src/serverdlg.cpp:    UpdateRecorderStatus ( QString::null );
src/serverdlg.cpp:    UpdateRecorderStatus ( QString::null );
src/serverdlg.cpp:    UpdateRecorderStatus ( QString::null );
src/serverdlg.cpp:    if ( pServer->GetRecorderErrMsg() != QString::null )
src/serverdlg.cpp:        UpdateRecorderStatus ( QString::null );
src/serverdlg.cpp:    if ( pServer->GetRecorderErrMsg() != QString::null || pServer->GetRecordingDir() != "" )
src/serverdlg.cpp:        UpdateRecorderStatus ( QString::null );
src/serverdlg.cpp:                edtCurrentSessionDir->setText ( sessionDir != QString::null ? sessionDir : "" );
src/serverdlg.cpp:        if ( strRecordingDir == QString::null )
src/recorder/jamcontroller.cpp:        bRecorderInitialised = ( strRecorderErrMsg == QString::null );
src/recorder/jamcontroller.cpp:        strRecorderErrMsg    = QString::null;
src/recorder/jamrecorder.cpp: * @return QString::null on success else the failure reason
src/recorder/jamrecorder.cpp:    QString   errmsg = QString::null;
```

src/serverdlg.cpp
pljones: I prefer the longer for as it's the "Server Address" for a "Custom Directory".  Not a "Custom" "Directory Server".

src/serverdlg.cpp
softins: OK, we can deal with that issue separately then.

src/serverdlg.cpp
softins: OK, so did your original comment mean you are dealing with it in the next PR rather than this one?

pljones: Yes.  Remaining after this now looks like:
- feature/server-list-persistence-ui
The main update to get the server list persistence file working, plus amending server registration to lose the checkbox
- patch/language-selection
Small layout change for the Server UI language selector
- patch/recorder-adjustments
Some changes in the layout for the Server Recorder (it was more crowded than needed), plus the labelling change referred to here and getting rid of `QString::null` (as that was all server recorder code).

src/serverdlg.cpp
softins: OK, yes, I see a lot of others are labelled "xxx combo box" too, so that's fine. Any change would be a separate exercise.

softins (approved):

ann0see:
Hmm. Not 100% sure about the concept of setting a list to "Unregistered". Basically it seems a bit unintuitive if we think about the private/registered(=public) server concept. How do we ensure that the user knows "Unregistered" = private server? I think the behaviour we had until now was a little bit more intuitive.

ann0see:
… and I mean, putting the custom server textbox into another tab feels a bit strange.

pljones:
> Hmm. Not 100% sure about the concept of setting a list to "Unregistered". Basically it seems a bit unintuitive if we think about the private/registered(=public) server concept. How do we ensure that the user knows "Unregistered" = private server? I think the behaviour we had until now was a little bit more intuitive.

It solves an existing defect - go read the discussions.

pljones:
> … and I mean, putting the custom server textbox into another tab feels a bit strange.

It hasn't moved.

ann0see:
> Documentation...

Yes, probably we need to rewrite some parts of the public/private server region then.

ann0see (approved):

PR #2187, opened 2021-12-27
Enhance version text

henkdegroot:
GPL version included in version text.
Add Qt reference and version to command line version output.
Added a newline at the start for windows only to print version details nice in command prompt.

**Context: Fixes an issue?**
No

No

Complete

Nothing

## Checklist

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -410,17 +410,17 @@ CAboutDlg::CAboutDlg ( QWidget* parent ) : CBaseDlg ( parent )
     // libraries used by this compilation
     txvLibraries->setText ( tr ( "This app uses the following libraries, resources or code snippets:" ) + "<br><p>" +
                             tr ( "Qt cross-platform application framework" ) +
-                            ", <i><a href=\"http://www.qt.io\">http://www.qt.io</a></i></p>"
+                            ", <i><a href=\"https://www.qt.io\">https://www.qt.io</a></i></p>"
                             "<p>Opus Interactive Audio Codec"
-                            ", <i><a href=\"http://www.opus-codec.org\">http://www.opus-codec.org</a></i></p>"
+                            ", <i><a href=\"https://www.opus-codec.org\">https://www.opus-codec.org</a></i></p>"
                             "<p>" +
                             tr ( "Audio reverberation code by Perry R. Cook and Gary P. Scavone" ) +
-                            ", 1995 - 2004, <i><a href=\"http://ccrma.stanford.edu/software/stk\">"
+                            ", 1995 - 2021, <i><a href=\"https://ccrma.stanford.edu/software/stk\">"
                             "The Synthesis ToolKit in C++ (STK)</a></i></p>"
                             "<p>" +
                             tr ( "Some pixmaps are from the" ) +
                             " Open Clip Art Library (OCAL), "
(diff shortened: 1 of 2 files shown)


henkdegroot:
@pljones can you please check if you are happy with the addition of the Qt details like this or if you would like to see some more text (i.e. Qt framework).

src/util.cpp
```diff
@@ -1886,7 +1890,15 @@ QString GetVersionAndNameStr ( const bool bWithHtml )
         strVersionText += "\n *** ";
     }
 
-    strVersionText += QCoreApplication::tr ( "Released under the GNU General Public License (GPL)" );
+    strVersionText += QCoreApplication::tr ( "Released under the GNU General Public License version 2 (GPLv2)" );
+
+    if ( !bWithHtml )
```
pljones, line 1895: Shouldn't `bWithHtml` only affect formatting, not content?

src/util.cpp
```diff
@@ -1886,7 +1890,15 @@ QString GetVersionAndNameStr ( const bool bWithHtml )
         strVersionText += "\n *** ";
     }
 
-    strVersionText += QCoreApplication::tr ( "Released under the GNU General Public License (GPL)" );
+    strVersionText += QCoreApplication::tr ( "Released under the GNU General Public License version 2 (GPLv2)" );
```
atsampson, line 1893: I don't think that's correct - the license is GPL v2 **or later**. (There are projects out there which are GPL v2 only, e.g. the Linux kernel.)

pljones: I'm fairly sure Jamulus has _never_ said "or later" in its licence terms.

src/util.cpp
henkdegroot: I have used this, because it did not make to me to have the Qt details shown in the part of the GUI about dialog.
Good point about bWithHtml being more just about formatting.

Any objections in including the Qt reference in the GUI here as well?

src/util.cpp
henkdegroot: > run from the command line in the situation we're addressing here.

Related to the bWithHtml option (if that is to be treated as only to be used for formatting), this would mean duplicating the text which is already part of the about dialog. 
And hints to a preferred solution for this?

> For Qt, according to https://doc.qt.io/qt-5/lgpl.html , I would say we can reference that page

Just to get it correct, you just mean adding the link to the line for Qt?

pljones: I mean specify that the we use the Qt libraries, the version of those that we use in the build in question (we use two - both 5.x - I believe), and link to the Qt LGPL page given.

I also notice in `COPYING` that's not all -- can you check there to see.

(In the distant past, when _everyone_ built Jamulus from source, `COPYING` was all that was needed.  Now that we supply binaries, we need to ensure sufficient detail is provided to comply with any licence terms we specify or comply with.)

henkdegroot: Added more details and also changed the bWithHtml to bDisplayInGui (for clarity around the usage).
Here is a sample of the console output:
version-text-console

Please note the Qt version is dynamic, so will show the version used when compiled.

pljones: Looks great!

pljones (approved):

pljones:
@henkdegroot Thanks for taking the time to clean things up, including checking all the licences were up to date, too. 😃

softins (approved):
Looks fine to me

PR #2198, opened 2021-12-29
Add System Sound Setup Button for Windows and macOS

### LINKED ISSUE #2197: Add System Sound Setup button to Audio-/Network-settings for Windows and macOS
Feature has been discussed in #2181 

Add a System Sound Setup button to the Audio-/Network-settings function for Windows and macOS, so that you get easy access from within Jamulus.
When you click the button, the system sound setup dialog will be launched in a separate process so that is does not interfere with Jamulus. 

Sample screenshot showing what this might look like on windows:
system-sound-windows

Sample screenshot showing what this might look like on macOS:
system-sound-macOS

henkdegroot:
System Sound Setup button added to the Audio-/Network-setting functions.
This opens the System Sound Setup dialog on Windows and macOS.

**Context: Fixes an issue?**
Fixes #2197 

Documentation needs updating to reflect the change and screenshots need to be updated as well.

Sample screenshot for Windows:
system-sound-windows

Complete

Testing on different OS is appreciated. I have only tested using virtual machines.

## Checklist

### DIFF
--- a/src/clientsettingsdlg.h
+++ b/src/clientsettingsdlg.h
@@ -37,6 +37,7 @@
 #include <QLayout>
 #include <QButtonGroup>
 #include <QMessageBox>
+#include <QProcess>
 #include "global.h"
 #include "util.h"
 #include "client.h"
@@ -68,6 +69,17 @@ protected:
     void    UpdateDirectoryServerComboBox();
     void    UpdateAudioFaderSlider();
     QString GenSndCrdBufferDelayString ( const int iFrameSize, const QString strAddText = "" );
+
+    //Define strings for launching System Sound Setup
+#ifdef _WIN32
+    QString OpenSoundSettingsCmd = "control mmsys.cpl,,1";
+#elif defined ( Q_OS_MACX )
+    QString OpenSoundSettingsCmd = "open /System/Library/PreferencePanes/Sound.prefPane";
+#elif defined ( __linux__ )
+    QString OpenSoundSettingsCmd = "gnome-control-center sound";
+#else
+    QString OpenSoundSettingsCmd = "";
+#endif
 
     virtual void showEvent ( QShowEvent* );
 
@@ -106,6 +118,7 @@ public slots:
     void OnTabChanged();
     void OnMakeTabChange ( int iTabIdx );
     void OnAudioPanValueChanged ( int value );
+    void OnSoundSetupClicked();
 
 signals:
     void GUIDesignChanged();
--- a/src/clientsettingsdlg.h
+++ b/src/clientsettingsdlg.h
@@ -70,12 +70,12 @@ protected:
     void    UpdateAudioFaderSlider();
     QString GenSndCrdBufferDelayString ( const int iFrameSize, const QString strAddText = "" );
 
-    //Define strings for launching System Sound Setup
+    // Define strings for launching System Sound Setup
 #ifdef _WIN32
     QString OpenSoundSettingsCmd = "control mmsys.cpl,,1";
-#elif defined ( Q_OS_MACX )
+#elif defined( Q_OS_MACX )
     QString OpenSoundSettingsCmd = "open /System/Library/PreferencePanes/Sound.prefPane";
-#elif defined ( __linux__ )
+#elif defined( __linux__ )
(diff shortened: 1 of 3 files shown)


ann0see:
After viewing the screenshots, I have a small concern: does the button confuse newcomers since it might imply similarity to ASIO settings?

How can we ensure - except for labelling - that the user grasps that there is in fact a difference between ASIO device settings and the system settings?

henkdegroot:
> How can we ensure - except for labelling -

Open to suggestions here, though I would like to believe the button text is clear enough. New Users may not even know what ASIO device settings is.

ann0see:
> New Users may not even know what ASIO device settings is

exactly. But maybe this button does exactly what they need. I‘d assume the first thing they’d do if there’s no sound is to open system settings while on fact they misconfigured ASIO4ALL.

ann0see:
Not sure if moving the button above "device" would help?

henkdegroot:
> above "device" would help?

That makes sense to me. Should we ask others for input on this?

src/clientsettingsdlg.h
```diff
@@ -69,6 +70,17 @@ class CClientSettingsDlg : public CBaseDlg, private Ui_CClientSettingsDlgBase
     void    UpdateAudioFaderSlider();
     QString GenSndCrdBufferDelayString ( const int iFrameSize, const QString strAddText = "" );
 
+    // Define strings for launching System Sound Setup
+#ifdef _WIN32
+    QString OpenSoundSettingsCmd = "control mmsys.cpl,,1";
+#elif defined( Q_OS_MACX )
+    QString OpenSoundSettingsCmd = "open /System/Library/PreferencePanes/Sound.prefPane";
+#elif defined( __linux__ )
+    QString OpenSoundSettingsCmd = "gnome-control-center sound";
```
pljones, line 79: What if I'm not running Gnome?  In fact, what if I'm running a remote X server with JACK netaudio out to the local PC?

henkdegroot: > What if I'm not running Gnome?

Really not my area of expertise, this is the only thing I could find to launch a dialog to get the sound settings display.
No experience with a remote Jamulus, would you then also need access to the local built sound settings?
Due to the many different use cases of Linux, perhaps we should not have this button for Linux but just for macOS and Windows.

ann0see: Ok. I think we should open qjackctl anyways...

henkdegroot: > qjackctl

Qjackctl is not really System Sound Settings, is it? Is it really needed to have a button to open qjackctl? Perhaps Qjackctl should be linked to the ASIO settings button? Obvious the button text should be updated as well for that to be correct.

pljones: The problem is that this is sound system specific and we're trying to do the change in the generic part of the sound code.

Ideally that whole panel would be managed per sound system and the generic code would know nothing about it.

We definitely should _not_ try to shoe-horn in mixed concepts like opening qJackCtl with the ASIO Setting button!

henkdegroot: I have update the PR to be just for Windows and macOS now.

pljones:
> EDIT: Thinking about this more, not sure if moving the button to the top would make any difference. As when it is the top most button, it will be likely the first button a person will try/use.

Below is better.

But what about a button to open the control panel for my USB sound card, so I can adjust my mic input gain there?

henkdegroot:
> open the control panel for my USB sound card

Do you have an example? Are these not controlled via the system sound setting?

The intend for this button is to allow quick/easy access to a setting/dialog which most users have a hard time finding.

pljones:
```
"C:\Program Files (x86)\Echo FireWire\Echo FireWire Console.exe"
```
image

henkdegroot:
> "System sounds" isn't necessarily _the right place_ to go to

Well for the internal microphone it is, and most people using the internal/built in sound cards are not that technical.
These people will have a hard time to setup any shortcut.....

Guess we leave it as is.....

ann0see:
Why closing so quickly? Maybe something else could be found?

henkdegroot:
> Maybe something else could be found?

Hope so, just adding the button like this seems causing confusion and might send users down the wrong path.

PR #2199, opened 2021-12-29
Server UI changes

pljones:

My initial goal was to add to the server UI a way of setting (and changing in flight) the server list persistence file for a Jamulus directory.

However, during that change, it became apparent that the design of the interface between GUI and service layers was not geared to such in flight changes - the server is more expectant that it will be run with a configuration and left running "as is".

CHANGELOG: UI: Amend server registration, add server persistence

**Context: Fixes an issue?**

* Jamulus Directory improvements: UI to set / amend / remove the server list persistence file;

* Jamulus Server improvements: the server can be switched between:
    - unregistered
    - registered as a public server
    - registered with a custom directory
    - run as a Jamulus Directory

without restart, through the UI.  That configuration will be saved and restored in the server inifile.  If set to unregistered, the server will start up _unregistered_, rather than registered in the default directory.  You can set the custom directory without registering, so you can set that whilst unregistered then switch to registering with the custom directory, and never register as a public server.

The UI has changed both in layout and behaviour.  All server documentation probably needs review as a result.

I've been running it for months in headless mode.  I've used it on Windows and Linux with the GUI and find it works much better than 3.8.1.

More testing.  UX feedback.

## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -557,6 +557,9 @@ int main ( int argc, char** argv )
     }
 #endif
 
+    // TODO create settings in default state, if loading from file do that next, then come back here to
+    //      override from command line options, then create client or server, letting them do the validation
+
     if ( bIsClient )
     {
         if ( ServerOnlyOptions.size() != 0 )
@@ -593,12 +596,19 @@ int main ( int argc, char** argv )
 
         if ( bUseGUI )
         {
-            if ( strDirectoryServer.isEmpty() )
+            // by definition, when running with the GUI we always default to registering somewhere but
+            // until the settings are loaded we do not know where, so we cannot be prescriptive here
+
+            if ( !strServerListFileName.isEmpty() )
             {
-                // per definition: if we are in "GUI" server mode and no directory server
-                // address is given, we use the default directory server address
-                strDirectoryServer = DEFAULT_SERVER_ADDRESS;
-                qInfo() << qUtf8Printable ( QString ( "- default directory server set: %1" ).arg ( strDirectoryServer ) );
+                qInfo() << "Note:"
+                        << "Server list persistence file will only take effect when running as a directory server.";
+            }
+
+            if ( !strServerListFilter.isEmpty() )
+            {
+                qInfo() << "Note:"
+                        << "Server list filter will only take effect when running as a directory server.";
             }
         }
         else
@@ -607,51 +617,11 @@ int main ( int argc, char** argv )
             if ( !strIniFileName.isEmpty() )
             {
                 qWarning() << "No initialization file support in headless server mode.";
+                strIniFileName = "";
             }
-        }
+            // therefore we know everything based on command line options
 
-        if ( strDirectoryServer.isEmpty() )
-        {
-            // per definition, we must be a headless server and ignoring inifile, so we have all the information
-
-            if ( !strServerListFileName.isEmpty() )
-            {
-                qWarning() << "Server list persistence file will only take effect when running as a directory server.";
-                strServerListFileName = "";
-            }
-
-            if ( !strServerListFilter.isEmpty() )
-            {
-                qWarning() << "Server list filter will only take effect when running as a directory server.";
-                strServerListFilter = "";
-            }
-
-            if ( !strServerPublicIP.isEmpty() )
-            {
-                qWarning() << "Server Public IP will only take effect when registering a server with a directory server.";
-                strServerPublicIP = "";
-            }
-        }
-        else
-        {
-            // either we are not headless and there is an inifile, or a directory server was supplied on the command line
-
-            // if we are not headless, certain checks cannot be made, as the inifile state is not yet known
-            if ( !bUseGUI && strDirectoryServer.compare ( "localhost", Qt::CaseInsensitive ) != 0 && strDirectoryServer.compare ( "127.0.0.1" ) != 0 )
-            {
-                if ( !strServerListFileName.isEmpty() )
-                {
-                    qWarning() << "Server list persistence file will only take effect when running as a directory server.";
-                    strServerListFileName = "";
-                }
-
-                if ( !strServerListFilter.isEmpty() )
-                {
-                    qWarning() << "Server list filter will only take effect when running as a directory server.";
-                    strServerListFileName = "";
-                }
-            }
-            else
+            if ( strDirectoryServer.compare ( "localhost", Qt::CaseInsensitive ) == 0 || strDirectoryServer.compare ( "127.0.0.1" ) == 0 )
             {
                 if ( !strServerListFileName.isEmpty() )
                 {
@@ -709,14 +679,39 @@ int main ( int argc, char** argv )
                     }
                 }
             }
-
-            if ( !strServerPublicIP.isEmpty() )
+            else
             {
-                QHostAddress InetAddr;
-                if ( !InetAddr.setAddress ( strServerPublicIP ) )
+                if ( !strServerListFileName.isEmpty() )
                 {
-                    qWarning() << "Server Public IP is invalid. Only plain IP addresses are supported.";
+                    qWarning() << "Server list persistence file will only take effect when running as a directory server.";
+                    strServerListFileName = "";
+                }
+
+                if ( !strServerListFilter.isEmpty() )
+                {
+                    qWarning() << "Server list filter will only take effect when running as a directory server.";
+                    strServerListFileName = "";
+                }
+            }
+
+            if ( strDirectoryServer.isEmpty() )
+            {
+                if ( !strServerPublicIP.isEmpty() )
+                {
+                    qWarning() << "Server Public IP will only take effect when registering a server with a directory server.";
                     strServerPublicIP = "";
+                }
+            }
+            else
+            {
+                if ( !strServerPublicIP.isEmpty() )
+                {
+                    QHostAddress InetAddr;
+                    if ( !InetAddr.setAddress ( strServerPublicIP ) )
+                    {
+                        qWarning() << "Server Public IP is invalid. Only plain IP addresses are supported.";
+                        strServerPublicIP = "";
+                    }
                 }
             }
         }
@@ -881,10 +876,6 @@ int main ( int argc, char** argv )
                     CLocale::LoadTranslation ( Settings.strLanguage, pApp );
                 }
 
-                // update server list AFTER restoring the settings from the
-                // settings file
-                Server.UpdateServerList();
-
                 // GUI object for the server
                 CServerDlg ServerDlg ( &Server, &Settings, bStartMinimized, nullptr );
 
@@ -902,11 +893,12 @@ int main ( int argc, char** argv )
                 // only start application without using the GUI
                 qInfo() << qUtf8Printable ( GetVersionAndNameStr ( false ) );
 
-                // enable server list if a directory server is defined
-                Server.SetServerRegistered ( !strDirectoryServer.isEmpty() );
-
-                // update serverlist
-                Server.UpdateServerList();
+                // CServerListManager defaults to AT_NONE, so need to switch if
+                // strDirectoryServer is wanted
+                if ( !strDirectoryServer.isEmpty() )
+                {
+                    Server.SetDirectoryType ( AT_CUSTOM );
+                }
 
                 pApp->exec();
             }
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -684,12 +684,6 @@ void CServer::OnAboutToQuit()
 
     Stop();
 
-    // if server was registered at the directory server, unregister on shutdown
-    if ( GetServerRegistered() )
-    {
-        Unregister();
-    }
-
     if ( bWriteStatusHTMLFile )
     {
         WriteHTMLServerQuit();
--- a/src/serverdlg.cpp
+++ b/src/serverdlg.cpp
@@ -52,24 +52,36 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
 
     lvwClients->setAccessibleName ( tr ( "Connected clients list view" ) );
 
-    // Make My Server Public flag
-    chbRegisterServer->setWhatsThis ( "<b>" + tr ( "Make My Server Public" ) + ":</b> " +
-                                      tr ( "If the Make My Server Public check box is checked, this server registers "
-                                           "itself at the directory server so that all users of the application "
-                                           "can see the server in the connect dialog server list and "
-                                           "connect to it. The registration of the server is renewed periodically "
-                                           "to make sure that all servers in the connect dialog server list are "
-                                           "actually available." ) );
+    // Server list selection combo box
+    QString strDirectoryTypeAN = tr ( "Directory Type combo box" );
+    lblDirectoryType->setAccessibleName ( strDirectoryTypeAN );
+    cbxDirectoryType->setAccessibleName ( strDirectoryTypeAN );
+
+    QString strDirectoryTypeWT = "<b>" + tr ( "Directory" ) + ":</b> " +
+                                 tr ( "Select '%1' not to register your server with a directory." ).arg ( DirectoryTypeToString ( AT_NONE ) ) +
+                                 "<br>" + tr ( "Select one of the genres to register with that directory." ) + "<br>" +
+                                 tr ( "Or select '%1' and specify a Custom Directory address on the "
+                                      "Options tab to register with a custom directory." )
+                                     .arg ( DirectoryTypeToString ( AT_CUSTOM ) ) +
+                                 "<br><br>" +
+                                 tr ( "For any value except '%1', this server registers "
+                                      "with a directory so that a %2 user can select this server "
+                                      "in the client connect dialog server list when they choose that directory." )
+                                     .arg ( DirectoryTypeToString ( AT_NONE ) )
+                                     .arg ( APP_NAME ) +
+                                 "<br><br>" +
+                                 tr ( "The registration of the server is renewed periodically "
+                                      "to make sure that all servers in the connect dialog server list are "
+                                      "actually available." );
+    lblDirectoryType->setWhatsThis ( strDirectoryTypeWT );
+    cbxDirectoryType->setWhatsThis ( strDirectoryTypeWT );
 
     // server registration status label
     lblRegSvrStatus->setWhatsThis ( "<b>" + tr ( "Register Server Status" ) + ":</b> " +
-                                    tr ( "If the Register Server check box is checked, this will show "
-                                         "whether registration with the directory server is successful. If the "
-                                         "registration failed, please choose another server list." ) );
-
-    cbxDirectoryType->setWhatsThis ( "<b>" + tr ( "Server List Selection" ) + ":</b> " +
-                                     tr ( "Selects the server list (i.e. directory server address) in which your server will be added." ) );
-    cbxDirectoryType->setAccessibleName ( tr ( "Server list selection combo box" ) );
+                                    tr ( "When a value other than \"%1\" is chosen for Directory, this will show "
+                                         "whether registration is successful. If the registration failed, "
+                                         "please choose a different directory." )
+                                        .arg ( DirectoryTypeToString ( AT_NONE ) ) );
 
     // server name
     QString strServName = "<b>" + tr ( "Server Name" ) + ":</b> " +
@@ -171,13 +183,34 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
                                               "This will prevent recording until a new value is selected." ) );
 
     // custom directory
-    QString strCustomDirectory = "<b>" + tr ( "Custom Directory" ) + ":</b> " +
-                                 tr ( "The custom directory is the IP address or URL of the directory "
-                                      "server at which the server list of the connection dialog is managed." );
+    QString strCustomDirectory = "<b>" + tr ( "Custom Directory address" ) + ":</b> " +
+                                 tr ( "The Custom Directory address is the address of the directory "
+                                      "holding the server list to which this server should be added." );
 
     lblCustomDirectory->setWhatsThis ( strCustomDirectory );
     edtCustomDirectory->setWhatsThis ( strCustomDirectory );
     edtCustomDirectory->setAccessibleName ( tr ( "Custom Directory line edit" ) );
+
+    // server list persistence file name (directory server only)
+    pbtServerListPersistence->setAccessibleName ( tr ( "Server List Filename dialog push button" ) );
+    pbtServerListPersistence->setWhatsThis ( "<b>" + tr ( "Server List Filename" ) + ":</b> " +
+                                             tr ( "Click the button to open the dialog that allows the "
+                                                  "server list persistence file name to be set. The user Jamulus is running as "
+                                                  "needs to be able to create the file name specified "
+                                                  "although it may already exist (it will get overwritten on save)." ) );
+
+    edtServerListPersistence->setAccessibleName ( tr ( "Server List Filename text box (read-only)" ) );
+    edtServerListPersistence->setWhatsThis ( "<b>" + tr ( "Server List Filename" ) + ":</b> " +
+                                             tr ( "The current value of server list persistence file name. The user Jamulus is running as "
+                                                  "needs to be able to create the file name specified "
+                                                  "although it may already exist (it will get overwritten on save). "
+                                                  "Click the button to open the dialog that allows the "
+                                                  "server list persistence file name to be set." ) );
+
+    tbtClearServerListPersistence->setAccessibleName ( tr ( "Clear the server list file name button" ) );
+    tbtClearServerListPersistence->setWhatsThis ( "<b>" + tr ( "Clear Server List Filename" ) + ":</b> " +
+                                                  tr ( "Click the button to clear the currently selected server list persistence file name. "
+                                                       "This will prevent persisting the server list until a new value is selected." ) );
 
     // start minimized on operating system start
     chbStartOnOSStart->setWhatsThis ( "<b>" + tr ( "Start Minimized on Operating System Start" ) + ":</b> " +
@@ -247,6 +280,7 @@ lvwClients->setMinimumHeight ( 140 );
 
     // directory type combo box
     cbxDirectoryType->clear();
+    cbxDirectoryType->addItem ( DirectoryTypeToString ( AT_NONE ) );
     cbxDirectoryType->addItem ( DirectoryTypeToString ( AT_DEFAULT ) );
     cbxDirectoryType->addItem ( DirectoryTypeToString ( AT_ANY_GENRE2 ) );
     cbxDirectoryType->addItem ( DirectoryTypeToString ( AT_ANY_GENRE3 ) );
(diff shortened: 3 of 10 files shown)


src/serverdlg.cpp
```diff
@@ -52,24 +52,36 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
...
-                                           "connect to it. The registration of the server is renewed periodically "
-                                           "to make sure that all servers in the connect dialog server list are "
-                                           "actually available." ) );
+    // Server list selection combo box
+    QString strDirectoryTypeAN = tr ( "Directory Type combo box" );
+    lblDirectoryType->setAccessibleName ( strDirectoryTypeAN );
+    cbxDirectoryType->setAccessibleName ( strDirectoryTypeAN );
+
+    QString strDirectoryTypeWT = "<b>" + tr ( "Directory" ) + ":</b> " +
+                                 tr ( "Select '%1' to leave your server unregistered." ).arg ( DirectoryTypeToString ( AT_NONE ) ) + "<br>" +
```
ann0see, line 61:
```suggestion
                                 tr ( "Select '%1' to leave your Server in private mode i.e. not showing it on a server list." ).arg ( DirectoryTypeToString ( AT_NONE ) ) + "<br>" +
```

src/util.h
```diff
@@ -558,8 +559,8 @@ inline QString DirectoryTypeToString ( EDirectoryType eAddrType )
 {
     switch ( eAddrType )
     {
-    case AT_CUSTOM:
-        return QCoreApplication::translate ( "CClientSettingsDlg", "Custom" );
+    case AT_NONE:
+        return QCoreApplication::translate ( "CServerDlg", "Unregistered" );
```
ann0see, line 563:
```suggestion
        return QCoreApplication::translate ( "CServerDlg", "Private/Unregistered" );
```

ann0see: AT_NONE could, better should say Private/Unregistered since we‘re talking about private servers in the docs – even though it’s not the exact technical description. However I think private is quicker to understand than unregistered.

Will test it as soon as I find time

ann0see (commented):
Thank you. Will try soon.

src/serverdlg.cpp
pljones: No!!!!!   Please...

There's no such thing as "private mode".

src/util.h
pljones: As above.

src/serverdlg.cpp
gilgongo: I think @ann0see is right at least since the term "private server" has been used so far to mean servers not listed in a directory: https://jamulus.io/wiki/Running-a-Server#server-types So we'd have to change the docs to explain what the "unregistered" server state means, and retire the use of the term "private server" - quite a big change!

@pljones statement "Even a registered server can be for private use" is interesting though if I understand correctly, since to make the term "unregistered server" work on the grounds that you can have a server registered on a custom directory that isn't public, then I think that would make a "custom directory" a "private directory" as per https://github.com/jamulussoftware/jamulus/discussions/2100 (essentially, I think explaining servers and directories as both having the property of "public" and "private" makes it easier to understand them).

ann0see: Unregistered= technical, 100% correct description. However it’s "Jamulus dev speech" and not intuitively understandable. 

Private = not 100% correct, but at least understandable and used commonly. Not only our website uses this word.

ann0see: What about using the label next to the select box to say "Not registered" and renaming "Unregistered" as Private?

pljones: >  make a "custom directory" a "private directory" 

No!

_Some_ custom directories are private.

_All_ private directories are custom.

Can you see the difference?

pljones: > What about using the label next to the select box to say "Not registered" and renaming "Unregistered" as Private?

You mean have a checkbox with a different label doing nothing?  And make a totally misleading cosmetic change to the dropdown?

pljones: Basically, a server operator **must** comprehend the concept of server registration.  Yes, it's a Jamulus-specific term.  And that's because it's a Jamulus-specific concept, unfortunately.

Maybe your countries don't maintain voter registries?  You don't have to register to vote?

I think the idea of "registration" is pretty natural for a normal adult but maybe that's just my heritage.

I can see that "Unregistered" may imply some "active" sense whereas "Not registered" is passive (and thus more accurate), however.

pljones: That does raise an interesting future project, if someone wants to do it -- adding a client whitelist to both server and directory.

ann0see: I think there’s a misunderstanding here:
In German there is a label next to the dropdown box. This says "Registered" if one selects Any Genre 1, Any genre 2,…

 if one selects "Unregistered" this label says "Unregistered" (or in German "Nicht registriert") too. IMHO: that’s waste of space. 

My suggestion:

Let the drop down say "Private" and the label say "Unregistered".

And no: in my country you don’t need to register to vote ;-). 

> I think the idea of "registration" is pretty natural for a normal adult but maybe that's just my heritage.

Yes. However I think there’s no need a musician who is running a server (without much IT knowledge) must understand the concept of Directories.

Another compromise:
What about changing Unregistered to "Don’t publish"?

pljones: Sorry, I disagree.  You've not said anything that changes my position.  I think "register" and "directory" are totally natural.

ann0see: > That does raise an interesting future project, if someone wants to do it -- adding a client whitelist to both server and directory. Possibly lower priority than getting IPv6 support finished.

Yep. That’s somehow related to what I attempted with my eduTools fork

pljones: Just to be clear:
> This says "Registered" if one selects Any Genre 1, Any genre 2,… if one selects "Unregistered" this label says "Unregistered" 

No, the label isn't just saying "Registered" or "Unregistered".  It's reporting the result of the attempt to registered.  If the directory is full, you'll see an error, for example, saying why your registration failed.  "Unregistered" could simply be blank, of course, if the user chooses "Not registered".  But it is _necessary_ information to the server operator when attempting to register, so it cannot be removed.

ann0see: > Not registering a server does not make it private. You could have it posted all over major world news feeds and Facebook

Totally valid.

We had a checkbox saying "Publish my server in the server list" or similar. This was quite clear.

Now we have "Not registered" in the drop down which gives a lot less information. 
Probably I‘m unhappy since the word "publish" was removed. Hence my suggestion: "Don’t publish".
Why? Unregistered doesn’t fit well to the term "private" . Private is present everywhere around Jamulus although it’s incorrect but we don’t have a fitting, non-technical term.

 I do not think we can get rid of the term "private" and introducing a new term is confusing.

gilgongo: > How is "Unregistered" harder to understand than "Not registered"?

We seem to be talking about several different things. The UI change itself is not controversial as it's needed for the functionality to work I think. But this PR is also implying a possibly large change to the terminology we have been using in the documentation, and calling into question (for the first time I think) our use of terms like "private". Or have I not understood?

Put it this way: if this PR is to go ahead with the server states of "Registered" and "Unregistered", does [anything here](https://jamulus.io/wiki/Running-a-Server#server-types) need changing? It feels like it does because the reader wouldn't be able to make a direct connection between what was in the UI and what's in the docs.

DavidSavinkoff: @ann0see said: We had a checkbox saying "Publish my server in the server list" or similar. This was quite clear.

Publish may be a better word than Register. I consulted a thesaurus for "publish" and found more good alternatives. "Publish" puts power in the hands of the user, whereas "Register" vaguely gets your server on some server list. Yes! ... "Publish my server in the server list" !

gilgongo: @ann0see 

> Private is present everywhere around Jamulus although it’s incorrect but we don’t have a fitting, non-technical term.

Exactly. It's not a precise meaning, but it's accurate enough for the context in which we use it. Arguing against the use of "private" is not going to get us anywhere.

gilgongo: > Publish may be a better word than Register. 

Perhaps, if we're going to come up with new terms for existing states of the server, then we need to be looking a the docs as well as this PR. We seem to be implying a pretty large change.

pljones: This is the **3.8.1** UI:

"Server Setup" tab:
image
```
- server mode chosen
- initialization file name: ../Jamulus-wip/ini-server
- selected port number: 56881
- default directory server set: anygenre1.jamulus.io
```
On start up, the server _immediately_ registered in the default directory.  However, once the settings were loaded, it then unregistered again due to the checkbox being off.

"Make my server public"
-- This is incorrect.  Just because the checkbox is checked, does _not_ mean a server is public.
"Register my server in the Server List"
-- This was an on/off switch for registering from back in the day when there was a single directory - you registered or not - so fine.  But life moves on.

"List [DISABLED]"
-- I have to enable registration to change where I register - meaning I have to register in the _wrong_ server list before I can choose the right one!  Bad design.

"Unregistered"
-- This is the latest change in registration state -- from registered with the default directory to unregistered.  The checkbox _causes_ this status because the server _actively_ unregisters (tells the directory to remove it from the server list) successfully.

"Options" tab:
image
"Custom Directory Server Address"
-- is rather long winded - but at least the input box is active.

pljones: This is the UI with this change:

Start up log:
```
- server mode chosen
- initialization file name: ini-server
- selected port number: 56881
```

"Server Setup" tab:
image

"Directory [Not registered]"
-- "Directory" because that's the current term we're using in the client, so this makes it consistent.  "Not registered" because "register" is the term that has been used since Jamulus first had a server list and, with this item selected, the server _does not register_ on start up at all.

"Not registered"
-- As the "result" of registration - in this case, no registration was attempted, so it makes sense.  It's going to say that if you switch from registered to no longer registered, too.

"Options" tab:
image
"Custom Directory Server Address [ ---value--- ]"
-- I've not changed it yet.  I did change it -- then had second thoughts.  This is a "Custom Directory" "server address" - i.e. the "data type" is "server address" and the "usage" is "Custom Directory", so the full text gives the best context as to what value to enter.  The "What's this?" text on the field is wrong, I just noticed...  Any UI/UX changes here, I'd rather keep separate at this stage.

"{Server List Filename} [| --- value --- |] {<X}"
--- Finally the "new" bit - allows choosing a server list file name using Qt file chooser (click button).  The "clear" button clears the value.

pljones: > "Publish my server in the server list" 
I don't like "publish" as it conflicts with the idea of private directories and private servers registering with them.  A _published_ private server?  It seems wrong.

gilgongo: image

1. Generally, the convention with drop-downs is they describe their own state (in this case "registered/not registered"). So would it be possible to make that message only appear if the directory is full? Looks a bit odd repeated like that.

2. Not a big deal, but the convention we and others also take is that the drop-down's label is the "thing" that's being changed by the dropdown. So to the uninitiated, this looks like it's the directory that's not registered, not the server. 

----
image

3. To make it look a bit more compact, you could label the address field "Custom Directory Server" and put "Enter address" into the "hint text" of the address field perhaps?

4. What does the server list file do?

----

5. On vocabulary: With "private" and "public" have gone from the UI I think it makes things less clear. I understand that Jamulus doesn't technically make anything private, and that a "public" server might be on a private directory or whatever,  but that's not the point. The terms are **accurate enough**  (shorthand, if you like) to give the _average_ user (not custom server operators, who are just going to have to read the manual) the important idea that if they make a server "public" it can be seen by others without the server operator doing anything else. And if they make it "private" then by default nobody can see it. I don't think we should be moving away from the current vocabulary because it's pretty much at the heart of Jamulus's UX.

pljones: >     1. Generally, the convention with drop-downs is they describe their own state (in this case "[directory name]/not registered"). So would it be possible to make that message only appear if the directory is full? Looks a bit odd repeated like that.

"Successfully unregistered" might be better than "Not registered" if the server was previously registered, else blank if it hasn't been registered, as it's the result of the action taken as a result of changing the dropdown selection.  The dropdown says what state you _want_ the status label says what happened as a result of trying to achieve that for you.

> 
>     2. Not a big deal, but the convention we and others also take is that the drop-down's label is the "thing" that's being changed by the dropdown. So to the uninitiated, this looks like it's the directory that's not registered, not the server.

Yes, that's unfortunately true -- "Server registration" maybe as the label?  I wanted it to be the same as in the client for familiarity and consistency.

>     3. To make it look a bit more compact, you could label the address field "Custom Directory Server" and put "Enter address" into the "hint text" of the address field perhaps?

You mean label "Custom Directory" (no "server" on that) and "Enter server address" into the hint text.  I guess that would work.  I'll give it a go - in a separate, later change.

>     4. What does the server list file do?

Go read the PR for that change.

>     5. On vocabulary: With "private" and "public"  gone from the UI I think it makes things less clear.

We **must not** use "private".  There's no two ways about it.  It's morally wrong, technically wrong and potentially legally wrong (IANAL).

> I understand that Jamulus doesn't technically make anything private, and that a "public" server might be on a private directory or whatever,  but that's not the point. 

Hm...

> The terms are **accurate enough**

No.  They are not "accurate enough" if they are misleading people into believing something that is simply untrue.

> (shorthand, if you like) to give the _average_ user (not custom server operators, who are just going to have to read the manual) the important idea that if they make a server "public" it can be seen by others without the server operator doing anything else. And if they make it "private" then by default nobody can see it. 

> I don't think we should be moving away from the current vocabulary because it's pretty much at the heart of Jamulus's UX.

I don't believe the "current vocabulary" as you put has been in place long enough to matter.  The concepts of a directory, multiple directories and custom directories have not even been stable over the last two and a half years, so how you can come up with some idea there's a "current vocabulary" that everyone understands is just stupidly unbelievably annoying.

Forget any ideas you've decided are right "just because" and look at reality.

pljones:
Jamulus client - connects your sound system to the internet!  Let's jam!

Jamulus server - somewhere Jamulus clients connect to for jams

Server list - a list of Jamulus servers retrieved from a Jamulus directory

Jamulus directory - a service with which a Jamulus server can choose to _register_ and from which a Jamulus client can request a server list (list of Jamulus servers)

Public directory - all of the built in Jamulus directories, some custom directories

Custom directory - a Jamulus directory not built into the Jamulus server or client

Private directory - a custom directory that restricts access for server registration and server list retrieval

Private server - a server (optionally registered with a directory) that restricts access for client connection

NOTE: "Privacy" (i.e. running as a private directory or private server) is not directly supported by Jamulus - firewall set up is required.  (When thinking about this, remember `--showallservers` will return the full list of servers registered at a directory, regardless of whether the client can access them; hence saying the directory must also restrict client access.)

src/serverdlg.cpp
gilgongo: > Go read the PR for that change.

Sorry, I meant how is the user supposed to know what that does? But I am a bit unclear as to the use case myself too. I guess they'd just have to read the manual, but there might be other ways of presenting it with some help text perhaps. 

> We must not use "private". There's no two ways about it.

OK, but the use of the term (also paired with "public", which I suppose is similarly imprecise by your logic) is all over the docs and in the forums. So if we are to retire them on the grounds you state, what do we say [here](https://jamulus.io/wiki/Running-a-Server#server-types) and [here](https://jamulus.io/wiki/Running-a-Server#running-a-private-server) and [elsewhere](https://github.com/jamulussoftware/jamulus/discussions?discussions_q=private)? I really don't think anyone is being "misled" as you say.

pljones: See here:
https://github.com/jamulussoftware/jamulus/pull/2199#issuecomment-1003943082

pljones: > > Go read the PR for that change.
> 
> Sorry, I meant how is the user supposed to know what that does? But I am a bit unclear as to the use case myself too. I guess they'd just have to read the manual, but there might be other ways of presenting it with some help text perhaps.

Right-click, "What's this?" or read the online documentation (which needs updating, of course).

pljones:
For the "Directory" drop down:
- "None" when no directory is selected
- One of the built in public directories
- Or a "Custom" for the specified custom directory

For the status value:
- Initially blank (i.e. if the user starts with "Directory" -> "None", it will remain blank)
- On registering, the result of attempting to register (so, usually "Registered" but sometimes some other message - but initially, IIRC "Registering" until the successful response and then some other value on failure)
- If the user chooses to switch "Directory" to "None", set the state to "Unregistered" (active - the server tells any directory it was registered in to remove it and, regardless of previous registration success, stops trying to register) -- it needs _not_ to become blank but I'd be happy enough with "Not registered" if "Unregistered" is unpalletable.

src/serverdlg.cpp
gilgongo: OK I'll have a look at the docs. Meanwhile, the "server list file" is used to preserve the list of servers on a directory between reboots, is that right?

pljones: > Meanwhile, the "server list file" is used to preserve the list of servers on a directory between reboots, is that right?

Yes, the change to add persistence for the server list added the file.  That went into 3.8.1.

The biggest problem I hit here was that most of the server settings were "set and forget, restart the server with a new command line" settings... even some of the UI ones had no effect until restart...  All that will change...

gilgongo: > That went into 3.8.1.

Ah yes, the `--directoryfile` option. So if you don't specify this and restart the directory, then any servers registered on it will need to reconnect (that might be a better way of explaining it perhaps?).

pljones: Yup.  It's great :).  OK, if I take my server down for a reboot, the hosted directories are offline for about 45 seconds.  But without it, the server lists come up empty and it takes up to 15 minutes for the servers to run their regular re-registration and the list to fill.  With it, the lists are full when the directory starts up.

gilgongo: So is the reason why we don't have it by default to do with not being able to easily write to the filesystem?

pljones: No, like all new features - the reason we don't change behaviour by default is that we preserve backward compatibility.

pljones:
Hm.  Might have to rescind my idea of
> Initially blank (i.e. if the user starts with "Directory" -> "None", it will remain blank)

The box holds a place holder of `STATUS` from the `.ui` layout, which gets overwritten with the current status -- i.e. "Not registered" if you're not registered - on start up, along with several other field updates (there's a method, `CServerDlg::UpdateGUIDependencies` that grabs current server values and stuffs them into the UI each time something changes - and on start up).

Currently, from start up:
```
$ ./Jamulus -s -i ini-server -p 56881
- server mode chosen
- initialization file name: ini-server
- selected port number: 56881
Using "192.168.1.19" as external IP.
Recording state not initialised
Recording state not initialised
libGL error: No matching fbConfigs or visuals found
libGL error: failed to load driver: swrast
```
image
(That last paragraph is there in 3.8.1 -- I'm not sure anyone looking at "What's This?" text wants to know that...)
image

image
Then switch to custom...
```
Server Registration Status update: Registration requested
```
image
... but that wasn't a directory:
```
Server Registration Status update: Registration failed
```
image
Switch back to "None" until I get the right Custom Directory address:
```
Server Registration Status update: Not registered
```
image
Cut-n-paste the right value...
image
Switch to "Custom":
```
Server Registration Status update: Registration requested
Server Registration Status update: Registered
```
image
(with the slightest flicker of "Registration requested" before succeeding).

Switch back to "None" now:
```
Server Registration Status update: Not registered
```
image

pljones:
We're also planning client Connect window changes:
- Server list, first column, "Server Name" becomes "Name"
- Server Address input box label becomes "Address"
- What's This? for the label, "If you know the IP address or URL of a server..." (we don't support URL format...) becomes "Where a server isn't in any of the server lists, you can enter the address (for example, _example.org:22524_ or _192.168.1.54_) you want to connect to here, or select one of the most recent values entered from the drop down list"
- Accessibility help needs a similar update

(Client Settings/Advanced Custom directories "What's This?" looks good to me -- seems I've not changed my mind _again_ since I wrote it... or I have but I've got back to where I was at that point...)

pljones:
Hm.  Can we try not running all of the builds in parallel?  It might be causing the fails.  If it doesn't fix it, we can switch back to parallel...

ann0see:
Probably out of scope, but having the recorder directory in the Options tab while having the (by default disabled) recorder in the configuration tab seems very confusing. We don't hint the user at setting a recording directory in the server configuration tab, so by default it looks like the recording feature is not available (and this can't be fixed).

pljones:
> Probably out of scope, but having the recorder directory in the Options tab while having the (by default disabled) "enable recorder" checkbox in the configuration tab seems very confusing. We don't hint the user at setting a recording directory, so by default it looks like the recording feature is not available (and this can't be fixed).

The whole Server UI is a bit clumsy.  Pretty much _everything_ for a server is "set once and forget".  You could almost have the list of clients filling one tab and a second tab "Settings" for everything.

pljones:
Given no one's come up with an alternative that makes sense to registering when you register with a directory, can we go ahead with this?

I was hoping to get this as well as
* patch/language-selection
* patch/recorder-adjustments

into the coming release but it looks like that's been missed.

gilgongo:
3.8.2 is supposed to be a bug fix release, but I don't suppose we need to stick to that.

hoffie:
> The UI has changed both in layout and behaviour. All server documentation probably needs review as a result.

Given that this PR touches many files and deliberately changes layout and behavior, I'd tend towards targeting 3.9.0. Such changes might come unexpected for users, e.g. see the unexpected (?) sandboxing side effects of the Mac signing stuff.

ann0see:
Can this be merged right after the upcoming release? If yes, please do so.

pljones:
I'll rebase (and squash shortly).

ann0see:
Great!

pljones:
Hm.  I changed the commit comment.  Let me try that again.

ann0see:
I merged some PRs, maybe that's why?

pljones:
> I merged some PRs, maybe that's why - although I doubt that.

No, I probably got lost somewhere locally - I had to redo what I was doing at one point.

src/util.h
```diff
@@ -560,7 +560,7 @@ inline QString DirectoryTypeToString ( EDirectoryType eAddrType )
     switch ( eAddrType )
     {
     case AT_NONE:
-        return QCoreApplication::translate ( "CServerDlg", "Unregistered" );
+        return QCoreApplication::translate ( "CServerDlg", "None" );
```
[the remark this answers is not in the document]
ann0see: -ignore-

pljones:
It'll be easier to document this once it's stable on master.  So does anyone want to approve it? :)

ann0see (approved):
Although only roughly tested a few weeks ago, I think we should merge this (and tag a nightly in a week or so)

src/server.cpp
```diff
@@ -684,12 +684,6 @@ void CServer::OnAboutToQuit()
 
     Stop();
```
softins, line 686: Was it deliberate to remove the `if ( GetServerRegistered() ) { Unregister(); }` here? I can't see it has moved to anywhere else.

src/main.cpp
```diff
@@ -593,65 +596,31 @@ int main ( int argc, char** argv )
...
-            if ( !strIniFileName.isEmpty() )
-            {
-                qWarning() << "No initialization file support in headless server mode.";
-            }
-        }
-
-        if ( strDirectoryServer.isEmpty() )
-        {
-            // per definition, we must be a headless server and ignoring inifile, so we have all the information
+            // by definition, when running with the GUI we always default to registering somewhere but
```
softins, line 599: This comment suggests that you can't have an unregistered ("private") server when running with the GUI. That's not true, surely (maybe it used to be?). Could the comment be clarified?

src/main.cpp
```diff
@@ -593,65 +596,31 @@ int main ( int argc, char** argv )
...
-                    qWarning() << "Server list persistence file will only take effect when running as a directory server.";
-                    strServerListFileName = "";
-                }
-
-                if ( !strServerListFilter.isEmpty() )
-                {
-                    qWarning() << "Server list filter will only take effect when running as a directory server.";
-                    strServerListFileName = "";
-                }
+                qWarning() << "No initialization file support in headless server mode.";
```
softins, line 619: Maybe add `strIniFileName = "";` similarly to the other options, for consistency? I know the original code that was moved here didn't do it.

softins (commented):
Had a look through and made some initial comments. Would really like to compile and try it out, but as I said, can't do that will Wednesday.
It looks like the "Register" checkbox has been replaced by a "None" option in the directory list. That means if the user wants to unregister their server, it can't remember where it was registered to, and the user needs to choose the directory afresh when re-registering. With an on/off checkbox, the preferred directory could be left set.

src/main.cpp
pljones: No, it says what it defaults to, not what's available.

src/server.cpp
hoffie: 9222c202158deff4b18f75975355f931a9dbaa18 was the original change doing this and `ServerListManager.SetDirectoryType()` seems to handle `Unregister()` if necessary: https://github.com/jamulussoftware/jamulus/blob/master/src/serverlist.cpp#L269-L275

It's called in https://github.com/jamulussoftware/jamulus/pull/2199/files#diff-c044ba02ccf7862f1e2f56436d5b92baa7bb35601791846655fb07ac253a8697R951

So this looks fine to me.

src/settings.cpp
```diff
@@ -859,6 +811,82 @@ QString directoryAddress = GetIniSetting ( IniXMLDocument, "server", "centralser
...
+    if ( !CommandLineOptions.contains ( "--centralserver" ) && !CommandLineOptions.contains ( "--directoryserver" ) )
+    {
+        // custom directory
+        // CServerListManager defaults to command line argument (or "" if not passed)
+        // Server GUI defaults to ""
+        QString directoryAddress = "";
+        // clang-format off
+// TODO compatibility to old version < 3.8.2
+directoryAddress = GetIniSetting ( IniXMLDocument, "server", "centralservaddr", directoryAddress );
+        // clang-format on
```
hoffie, line 824: Not a huge fan of this break-identation-to-get-attention scheme personally (I know that the original code uses this in several places).

hoffie (approved):
The code looks good to me, but I haven't done any real-world tests so far. I'll comment again when I get time to do those. At least one additional person should do them before merging, I guess. :)

Meta: This PR is rather large and was not easy to review for me. I did most of the initial reading via `git log -wp 75e4ad0c100bce605c14d9fe0f0729143fc70bc2..a50f9e2f62b3bae8acaf4bfde1b0918b53db7aad` (earlier state before rebasings) as it was much simpler for me to follow the individual commits.

What I think we should do for future refactoring PRs is:
- Make refactorings only do refactorings. It's way harder to judge whether code reorganizations are correct if they are mixed with new functionality.
- Use and keep individual commits with a polished subject and description so that they can be merged without squashing.

Thanks for doing this though, it looks much cleaner and logical to me now.

src/settings.cpp
pljones: > in several places

"Consistently" might be a better way of putting it ;).

src/main.cpp
pljones: Done

gilgongo:
For website screenshot URLs. Ignore.

server-1

server-2

Issue #2186, opened 2021-12-25
Do not show ASIO Device settings and open ASIO settings on JACK version of Jamulus


ann0see:
Documenting for future fixing. In theory this is a good first issue ;-). However the description is for the team.

Currently we still show the ASIO device settings button (probably somewhere in settings) and the open ASIO settings button in soundbase.cpp even if we compile the JACK version on windows. We should not use #ifdef _WIN32 in these files.

See:

Is this defined if we're running JACK on Windows?  If so, it's the wrong symbol.

_Originally posted by @pljones in https://github.com/jamulussoftware/jamulus/pull/2168#discussion_r773478884_

gilgongo:
Just a meta point: issues in "triage" are for when we've not decided if we want to deal with them yet. Issues with a milestone can't be in triage, so i've moved this to "backlog" as I assume we've decided we will work on it.

pljones:
Just checking -- this has 3.9.0 but looks like it went into 3.8.2, is that right?

ann0see:
Yes, I think so.

hoffie:
Yep. I've updated the milestone.

PR #2215, opened 2022-01-05
Remove ASIO button from Audio/Network settings when using JACK

henkdegroot:
Remove the ASIO Settings button in the Audio/Network Setting dialog when using JACK or when using macOS.

**Context: Fixes an issue?**
Fixes: #2186
The ASIO sound dialog (from soundbase.cpp) is not displayed when using JACK.
Therefore it is only needed to remove the button from the settings dialog.

No

Complete

Nothing

## Checklist

### DIFF
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -117,13 +117,14 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     chbAutoJitBuf->setAccessibleName ( tr ( "Auto jitter buffer switch" ) );
     chbAutoJitBuf->setToolTip ( strJitterBufferSizeTT );
 
+#if !defined( WITH_JACK )
     // sound card device
     lblSoundcardDevice->setWhatsThis ( "<b>" +
-                                       QString ( tr ( "Sound Card Device" ) + ":</b> " +
-                                                 tr ( "The ASIO driver (sound card) can be selected using "
-                                                      "%1 under the Windows operating system. Under macOS/Linux, no sound "
-                                                      "card selection is possible. If the selected ASIO driver is not valid "
-                                                      "an error message is shown and the previous valid driver is selected." ) )
+                                       QString ( tr ( "Device" ) + ":</b> " +
+                                                 tr ( "Under the Windows operating system the ASIO driver (sound card) can be "
+                                                      "selected using %1. If the selected ASIO driver is not valid an error "
+                                                      "message is shown and the previous valid driver is selected. "
+                                                      "Under macOS the input and output hardware can be selected." ) )
                                            .arg ( APP_NAME ) +
                                        "<br>" +
                                        tr ( "If the driver is selected during an active connection, the connection "
@@ -132,7 +133,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
 
     cbxSoundcard->setAccessibleName ( tr ( "Sound card device selector combo box" ) );
 
-#if defined( _WIN32 ) && !defined( JACK_REPLACES_ASIO )
+#    if defined( _WIN32 )
     // set Windows specific tool tip
     cbxSoundcard->setToolTip ( tr ( "If the ASIO4ALL driver is used, "
                                     "please note that this driver usually introduces approx. 10-30 ms of "
@@ -143,7 +144,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                     "driver, make sure to connect the ASIO inputs in the kX DSP settings "
                                     "panel." ) +
                                TOOLTIP_COM_END_TEXT );
-#endif
+#    endif
 
     // sound card input/output channel mapping
     QString strSndCrdChanMapp = "<b>" + tr ( "Sound Card Channel Mapping" ) + ":</b> " +
@@ -166,6 +167,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     cbxLOutChan->setAccessibleName ( tr ( "Left output channel selection combo box" ) );
     cbxROutChan->setWhatsThis ( strSndCrdChanMapp );
     cbxROutChan->setAccessibleName ( tr ( "Right output channel selection combo box" ) );
+#endif
 
     // enable OPUS64
     chbEnableOPUS64->setWhatsThis (
@@ -229,6 +231,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                       .arg ( APP_NAME ) +
                                   TOOLTIP_COM_END_TEXT;
 
+#if defined( _WIN32 ) && !defined( WITH_JACK )
     // Driver setup button
     QString strSndCardDriverSetup = "<b>" + tr ( "Sound card driver settings" ) + ":</b> " +
                                     tr ( "This opens the driver settings of your sound card. Some drivers "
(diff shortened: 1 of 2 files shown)


src/clientsettingsdlg.cpp
```diff
@@ -132,7 +132,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
```
pljones, line 132: It looks to me like lines 120 onwards through 146 are ASIO-specific code.  Can they _all_ go into a `#if` block?  (Else hide the controls?)

src/clientsettingsdlg.cpp
```diff
@@ -388,10 +388,10 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     chbDetectFeedback->setAccessibleName ( tr ( "Feedback Protection check box" ) );
 
     // init driver button
```
pljones, line 390: Is there any other code relating to the button that should have the same treatment?  It's a bit odd only to change the label or hide the button - we shouldn't even have the "click" handling code if the button is hidden.

src/clientsettingsdlg.cpp
henkdegroot: Guess that would be possible, also the channel mapping does not really apply when using JACK. Am I correct?
As if so, there is more to be removed when using JACK.

src/clientsettingsdlg.cpp
henkdegroot: Guess hiding the button (when the button was introduced) was the easy thing to do without impacting/changing too much code and because the button is not available, you cannot trigger the click handling. Will check the code and update accordingly.

henkdegroot: Do you think these changes should be code for JACK on Mac as well? 
As in that case the code should use the "WITH_JACK" in the if block instead of "JACK_REPLACES_ASIO".

src/clientsettingsdlg.cpp
pljones: Any code that doesn't apply for JACK also doesn't apply for JACK on Windows.

src/clientsettingsdlg.cpp
pljones: I'm not familiar with the preprocessor defines created in the Jamulus.pro for these situations.

henkdegroot: Going to update and use WITH_JACK. This is defined on all builds (when JACK is used) and will included _WIN32 as the button is only applicable to Windows.

src/clientsettingsdlg.cpp
pljones: Hm, `!defined( WITH_JACK )` covers iOS and Android.

You need something that says "MacOS or (Windows and not JACK)"

src/clientsettingsdlg.cpp
```diff
@@ -132,7 +133,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
 
     cbxSoundcard->setAccessibleName ( tr ( "Sound card device selector combo box" ) );
 
-#if defined( _WIN32 ) && !defined( JACK_REPLACES_ASIO )
+#    if defined( _WIN32 )
     // set Windows specific tool tip
```
pljones, line 137: These commects explicitly mention ASIO - so you should leave in the JACK_REPLACES_ASIO, I think?

src/clientsettingsdlg.cpp
```diff
@@ -166,6 +167,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     cbxLOutChan->setAccessibleName ( tr ( "Left output channel selection combo box" ) );
     cbxROutChan->setWhatsThis ( strSndCrdChanMapp );
     cbxROutChan->setAccessibleName ( tr ( "Right output channel selection combo box" ) );
+#endif
 
     // enable OPUS64
     chbEnableOPUS64->setWhatsThis (
```
pljones, line 173: Random point of note.

The UI has "(preferred)" against 128 samples.  The `strSndCrdBufDelay` text says `64 samples: The preferred setting`
image
image
Clever...

src/clientsettingsdlg.cpp
henkdegroot: I only used this once (line 120). All other 'if' lines that have WITH_JACK have a second check. 
And this relates to the WhatsThis text for the Device selection and channel mapping.

Do you believe iOS and Android do not support multiple devices?
And channel mapping, I also left this in place just in case it will be possible to have more than 2. I know for sure that channel mapping does not apply to JACK, but the rest may or may not show this depending on connect hardware/installed drivers.

For a moment I had doubt...and wanted to make the WhatsThis  separate for Windows and macOS. If you believe I should add the OS here, do you agree it would be better to do this as well?

src/clientsettingsdlg.cpp
henkdegroot: Well spotted! The website does not include the "preferred" text for this text. Agree this should be a separate issue.

src/clientsettingsdlg.cpp
pljones: ASIO lets you choose ASIO driver, because multiple ASIO drivers can register and the app talks to the ASIO driver.

JACK doesn't work that way - the app talks to JACK, JACK manages multiple devices.

MacOS has this weird system for audio aggregate devices that I don't understand and I don't know how it's managed.

As far as I'm aware, Android and iOS don't let you know anything about the sound devices, just use them.

Again, that's why all this code _should not be here_ -- it should be in the platform-specific sound code, then it would be clear what needs to be said for each case.

pljones: Each sound system implementation could provide a panel with all the necessary widgets specifically for that sound system.

If there was nothing that could be done, the panel could be empty and set to hidden.

Then the common sound code could supply the system-specific panel to the settings code for inserting into the UI.

henkdegroot: > MacOS has this weird system for audio aggregate devices that I don't understand and I don't know how it's managed.

In macOS you get the option to select which input/output combination you want to use supported by your system. So there is functionality linked to this drop down list. I have not seen the channel mapping on macOS, but I also haven't got a macOS running with multiple driver/outputs to select from.

I am still tempted to just use "WITH_JACK" for this, because I know for sure that it will not change the existing behaviour and you won't have to change/update if you suddenly have another OS that would support multiple devices/drivers.

My Android device it just list the Device as "Oboe" and no other option available. Similar as how to Linux is just showing "Jack" in the drop down and on Android I can not get to the tooltip and whatisthis detail in this dialog (or better..I don't know how to).

Not sure how you want to proceed with this. This issue/pr is just about removing the ASIO button when using JACK, and it seems we are trying to get a lot more done in this now.
Initially I just added removing the ASIO tooltip/whatisthis, as if the button is not there, then these should not be included as well.

Perhaps we should do another issue/pr for the "removal (hiding)" of the Device (with dropdown) for other platforms?

pljones: OK for this PR, `WITH_JACK` will do.

Just to repeat... it worries me that there's leakage of sound system-specific code into the generic code.  Each time something like this comes up, it triggers the question "Where else have we got it wrong?" (and it might not be just the sound code that's affected - there's potentially for leakage from the sound code to the client code in general, as demonstrated here).

I would like some work done to clean this up - move all the sound system-specific back into the subdirectories for each system, wrap that in a common wrapper in the common code and the common code where anything sound-related is needed.

But that's a task for another time.

pljones (approved):

src/clientsettingsdlg.cpp
henkdegroot: > I would like some work done to clean this up - move all the sound system-specific back into the subdirectories

Sound like a good thing but also some massive work.

pljones (approved):

softins (approved):
Looks ok and the CI has no errors, so happy to approve.

PR #2223, opened 2022-01-07
Correct minimum macOS version for legacy install

### LINKED ISSUE #2143: Jamulus MacOS Legacy installer does not work on Yosemite 10.10.5
**Describe the bug**

Tried to download and install the "[jamulus_3.8.1_mac_legacy.dmg](https://github.com/jamulussoftware/jamulus/releases/download/r3_8_1/jamulus_3.8.1_mac_legacy.dmg)" macOS legacy version, and install it on MaxOS X Yosemite Version 10.10.5.

Start is denied with error, stating that I cannot start Jamulus with the current version MacOS X 10.10.5, and it would require version 10.13.

**To Reproduce**

Download Jamulus legacy version
try to install on macOS X 10.10.5

**Expected behavior**

Legacy version to run on 10.10.5

**Screenshots**

2021-12-04 16_26_45-EIVOK Tralala _ Jitsi Meet

**Operating system**

MacOS X Yosemite 10.10.5

**Version of Jamulus**

3.8.1

**Additional context**

We tried it several times, with deinstalling and re-installing. Always the same effect

softins:
Allows recent Qt to set the macOS minimum target version, and provides a separate plist file for the Legacy build, that correctly specifies 10.10 as the minimum macOS version. (3.8.1 wrongly specified 10.13.0 as the minimum version for the legacy installer)

**Context: Fixes an issue?**

Fixes #2143 

No documentation needed.

Tested building on Mojave with both 5.9.9 and 5.15.2 of Qt. ~Awaiting results of Github CI.~

Test the assets created by the CI.

## Checklist

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -157,7 +157,11 @@ win32 {
         MACOSX_BUNDLE_ICON.path = Contents/Resources
         QMAKE_BUNDLE_DATA += MACOSX_BUNDLE_ICON
     } else {
-        QMAKE_INFO_PLIST = mac/Info-make.plist
+        equals(QT_VERSION, "5.9.9") {
+            QMAKE_INFO_PLIST = mac/Info-make-legacy.plist
+        } else {
+            QMAKE_INFO_PLIST = mac/Info-make.plist
+        }
     }
 
     LIBS += -framework CoreFoundation \
--- a/mac/Info-make-legacy.plist
+++ b/mac/Info-make-legacy.plist
@@ -21,7 +21,7 @@
         <key>LSApplicationCategoryType</key>
 	    <string>public.app-category.music</string>
         <key>LSMinimumSystemVersion</key>
-        <string>${MACOSX_DEPLOYMENT_TARGET}</string>
+        <string>10.10</string>
         <key>CFBundleExecutable</key>
         <string>@EXECUTABLE@</string>
         <key>CFBundleDevelopmentRegion</key>
--- a/mac/Info-xcode.plist
+++ b/mac/Info-xcode.plist
@@ -21,7 +21,7 @@
 	<key>LSApplicationCategoryType</key>
 	<string>public.app-category.music</string>
 	<key>LSMinimumSystemVersion</key>
-    <string>10.13.0</string>
+	<string>${MACOSX_DEPLOYMENT_TARGET}</string>
 	<key>CFBundleExecutable</key>
 	<string>${EXECUTABLE_NAME}</string>
 	<key>CFBundleDevelopmentRegion</key>


mac/Info-make.plist
```diff
@@ -21,7 +21,7 @@
         <key>LSApplicationCategoryType</key>
 	    <string>public.app-category.music</string>
         <key>LSMinimumSystemVersion</key>
-        <string>10.13.0</string>
+        <string>${MACOSX_DEPLOYMENT_TARGET}</string>
```
softins, line 24: Yes, this is indeed supposed to be `${MACOSX_DEPLOYMENT_TARGET}`, not `@MACOSX_DEPLOYMENT_TARGET@` due to the `sed` command generated in the `Makefile`.

Jamulus.pro
```diff
@@ -157,7 +157,11 @@ win32 {
         MACOSX_BUNDLE_ICON.path = Contents/Resources
         QMAKE_BUNDLE_DATA += MACOSX_BUNDLE_ICON
     } else {
-        QMAKE_INFO_PLIST = mac/Info-make.plist
+        equals(QT_VERSION, "5.9.9") {
```
pljones, line 160: The only thing I can ask about this change is whether an explicit `equals 5.9.9` here is correct.  Does this apply "before 5.10", for example?

softins: 5.9.9 is the version of Qt we use for a legacy build. There is a qmake function for comparing version numbers, but it only became available in 5.10, so wouldn't work here. I thought the simplest solution was just to compare with the actual version number we specify in the autobuild workflow, as it's for the benefit of the CI, not for people's own builds.

pljones: Assuming "5.9.9" appears only here in the Jamulus.pro file, that's probably okay, really.

We should really get the auto build running for _all_ targets in the Jamulus.pro file to make sure what it claims to offer works.  I notice there's a "jackonmac" target.  Should we also build that and does it need both legacy and non-legacy builds?

softins: > Assuming "5.9.9" appears only here in the Jamulus.pro file, that's probably okay, really.

It only appears here and in `.github/workflows/autobuild.yml`. So long as they match, it should be fine.

> We should really get the auto build running for _all_ targets in the Jamulus.pro file to make sure what it claims to offer works. I notice there's a "jackonmac" target. Should we also build that and does it need both legacy and non-legacy builds?

I haven't seen a demand for jackonmac. I guess the very few that might want it would build it themselves.
From `git log --patch Jamulus.pro`, it looks like Volker added it in 2016, for version 3.4.2. It is probably years since it was last tested. But that's out of scope for this PR.

pljones: Yep, I don't like stale code though, so it triggers an urge to get stuff tested...

DavidSavinkoff: The lowest Qt version yet.
I build Jamulus with Qt 5.6.1 on IA32 CentOS 6.
Thus you could specify Qt 5.6.1 instead of Qt 5.9.9 (unless MacOS can't).

> > Assuming "5.9.9" appears only here in the Jamulus.pro file, that's probably okay, really.
> 
> It only appears here and in `.github/workflows/autobuild.yml`. So long as they match, it should be fine.

pljones (approved):

ann0see (approved):
Seems good.

Issue #2222, opened 2022-01-07
64 and 128 sample size is indicated to be preferred (shown in different text)


henkdegroot:
**Describe the bug**

Different samples size is preferred.

**To Reproduce**

Open the GUI audio/network-settings and check the text shown in the GUI vs the text shown in the WhatsThis details.

**Expected behavior**

Either the same setting should be indicated as being preferred or the WhatsThis text may have the "preferred" reference removed (which will be inline with the Usermanual on the website).

**Screenshots**

![sample-size-preferred](https://user-images.githubusercontent.com/13550012/148596561-e1ca3b09-1a78-4192-8b32-9b6cede309b3.png)

**Operating system**

Win10

**Version of Jamulus**

3.8.1

henkdegroot:
@pljones, thank you for your always thorough review of PR's and spotting this!

henkdegroot:
Any preferences for which solution to implement?

pljones:
Historically, "128, preferred" has always been the case (predating the existance of "small network buffers").

Both the 64 samples for the audio buffers and the "small network buffers" options increase CPU load and potential jitter (particularly on Windows).  It may be perfectly okay to run a RPi4 with 64 samples and small network buffers but a five year old Windows laptop using ASIO4ALL might struggle to keep up with 128/"normal".

I'm inclined to remove "preferred" as it's system-specific.

henkdegroot:
@pljones, in the PR I have removed the "preferred" text from the whats is this text but left the one in the GUI. Not sure if you wanted that to be remove as well.

pljones:
It should be consistent: i.e. the list in the "What's This?" should show the user what they see on screen and explain what it means (and why it says it, which helps them remember what it means).  The radio button labels are explained in the "What's This?" text -- and should match.  So if we're not saying "preferred" in the "What's This?" text, we shouldn't say "preferred" in the label text.

PR #2232, opened 2022-01-10
Buffer Delay WhatisThis text update

henkdegroot:

Remove preferred text from "What Is This?" for Buffer Delay.

**Context: Fixes an issue?**

Fixes: #2222 

Yes, minor text update in the Software Manual would be advised to keep the "64, 128 and 256" samples options in-line with the details in the GUI.

Complete.
The next round of translation should apply this change to the local languages as well.

Nothing

## Checklist

### DIFF
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -119,12 +119,11 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
 
 #if !defined( WITH_JACK )
     // sound card device
-    lblSoundcardDevice->setWhatsThis ( "<b>" +
-                                       QString ( tr ( "Device" ) + ":</b> " +
-                                                 tr ( "Under the Windows operating system the ASIO driver (sound card) can be "
-                                                      "selected using %1. If the selected ASIO driver is not valid an error "
-                                                      "message is shown and the previous valid driver is selected. "
-                                                      "Under macOS the input and output hardware can be selected." ) )
+    lblSoundcardDevice->setWhatsThis ( "<b>" + tr ( "Device" ) + ":</b> " +
+                                       tr ( "Under the Windows operating system the ASIO driver (sound card) can be "
+                                            "selected using %1. If the selected ASIO driver is not valid an error "
+                                            "message is shown and the previous valid driver is selected. "
+                                            "Under macOS the input and output hardware can be selected." )
                                            .arg ( APP_NAME ) +
                                        "<br>" +
                                        tr ( "If the driver is selected during an active connection, the connection "
@@ -152,9 +151,9 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                      "input or output channel, the Input Channel Mapping and Output "
                                      "Channel Mapping settings are visible." ) +
                                 "<br>" +
-                                QString ( tr ( "For each %1 input/output channel (left and "
-                                               "right channel) a different actual sound card channel can be "
-                                               "selected." ) )
+                                tr ( "For each %1 input/output channel (left and "
+                                     "right channel) a different actual sound card channel can be "
+                                     "selected." )
                                     .arg ( APP_NAME );
 
     lblInChannelMapping->setWhatsThis ( strSndCrdChanMapp );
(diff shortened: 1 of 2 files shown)


src/clientsettingsdlg.cpp
```diff
@@ -186,7 +186,7 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                 "<br>" + tr ( "Three buffer sizes are supported" ) +
```
pljones, line 186: ... the change itself here is okay -- but the whole `strSndCrdBufDelay` text could probably do with rewriting...

henkdegroot: Apart from the 128 Samples text, it is now in-line with the web-site. The 128 Samples on the web-site includes the preferred text (which is matching with the text shown in the GUI).

pljones: First up, the extra use of `QString()` could be done away with in several places as I think `tr()` will happily take `.arg()` directly.

Then there's the fact checks...
> Three buffer sizes are supported

Hm.  If the sound card doesn't allow the app to set the buffer size, the app happily copes with the sound card buffer size.  So this isn't entirely true.  One could go as far as saying "Wrong!".  Indeed, it finishes by saying "(Jamulus) will still work with this setting".  (Arguable it's explained that "supported" means "one of these three" and "unsupported" means "not one of these three".)

> 256 samples: Should only be used on very slow computers or with a slow internet connection.

Not really sure the sound card buffer size is going to help with a slow network connection, I'm afraid...  It'll still either be fast enough or not fast enough.  There are other settings that _do_ affect network bandwidth but this doesn't... so scores another "Wrong!".

pljones: Lines 214 to 218 also look spurious...

pljones: My approximate understanding of the processing is...

If Jamulus is using 64 sample network buffers, it can read a 64 samples from the audio buffer each time they're ready, i.e. once per audio buffer input - reads once and sends a 64 sample audio buffer, reads twice for a 128 sample buffer, four times for a 256 sample buffer, and so on - and then sends them after each read.

If Jamulus is using 128 sample network buffers, it can read 128 samples from the audio buffer each time they're ready, i.e. once per audio buffer input - it needs to read two 64 audio sample buffers before sending, or reads once for a 128 sample buffer, or can read and send twice for 256 sample buffer, and so on.

Input and Output buffer sizes have to match for Audio and Network.  Audio and Network buffers do _not_ have to match.  Network buffer size has to be 64 or 128.  Audio buffer size isn't constrained.

_Latency_ in this sytem comes from a combination of factors:
1. The size of the Audio buffer: this should be as low as possible, obviously, but having a value lower than the size of the network buffer introduces negligible improvement and introduces CPU overhead and I/O.
2. The size of the Network buffer: this is more significant as it affects the delay introduced by jitter buffering
3. Jitter buffering: this is even more significant as it magnifies the network buffer delay
4. Ping: this adds to the other factors but is outside Jamulus's control

The unit of "delay" here is the time span represented by a sample at 48KHz (or, rather 64 or 128 samples, or whatever).

Note that actual jitter is also outside Jamulus's control - the jitter buffers are there to alleviate the effect of jitter to a more acceptable level.

Network _bandwidth_ is unlikely to be a problem for most Client systems - the highest data rate is well under 1Mbps (around 900Kbps at 64 samples, high quality stereo).  Increasing the audio buffer size _does_ drop the data rate - with 256 samples, low quality mono audio, it can go under 200Kbps.

The bandwidth usage, however, does not affect the latency.

henkdegroot: Excellent explained! Now it even starts to make sense to me. Thank you. Not sure how to capture this in the what is this text...

pljones (approved):

src/clientsettingsdlg.cpp
pljones: No - it shouldn't be in the "What's This?" (and scare people ;) ).  What's there now is "good enough" - if still too long.

It should be something short and relevant, really...  Hm...  Maybe something like:

"When enabled, this control allows selecting the size of the buffer Jamulus uses for handling sound.  For details on how the buffers affect latency (delay in when you hear sound), see the \[wiki](link...).  The values here are:
- 64 samples (...): this gives lower latency when 'small network buffer' is selected
- 128 samples (...): this gives lower CPU and network usage, suggested when 'small network buffer' is not selected
- 256 samples (...): this gives still lower CPU and network usage at the expense of a small increase in latency, suggested when for low bandwidth connections or slow computers.

When not enabled, Jamulus does not have control of the buffer size but similar considerations apply to those mentioned above.  Set up your sound device appropriately."

pljones: And another thing - separate improvement...
image
"Device" on the "Audio/Network Setup" tab really, really needs to say "Sound Device" (or Audio Device).  Yes, "everybody knows" it's not the network device...

henkdegroot: @pljones, I did not notice this until now (when I was fixing a conflict which I merged incorrect yesterday). All should now be back in sync with master and this change has been applied as well. I used Audio Device, because of the TAB name referring to Audio as well.

ann0see: > 10.67ms (256) […] slow computers

I don’t think it’s mainly about "Slow computers" from my experience it’s mainly about cheap audio devices…

henkdegroot: > I don’t think it’s mainly about "Slow computers"

Hence why I removed both text "Slow computers" and "Slow internet" and now just mention that 256 should only be used when 64 or 128 is causing issues :).

ann0see (approved):

Issue #2190, opened 2021-12-28
--help goes to stderr instead of stdout


QJKX:
Debian, Version 3.8.1

jamulus --help
prints text to stderr instead of stdout.

ditto
jamulus --version

Expected behavior:

text I expect to see should go to stdout.
only errors and debug info should go to stderr.

https://unix.stackexchange.com/questions/8813/should-the-usage-message-go-to-stderr-or-stdout

https://stackoverflow.com/questions/7977852/what-are-the-conventions-for-stdout-stderr-messages

ann0see:
Hi! Thank you for the report and sorry for the delay. Does it go to both?

pljones:
I've got a fix for this.  Although it'll be a while.

```
#include <iostream>
...
<< qUtf8Printable ( UsageArguments ( argv[0], options ) ) << "\n";
...
```

QJKX:
Hi!
>Does it go to both?

So by default I think a console shows both stdout and stderr. If messages were going to both you'd see them twice, and that doesn't happen.

ann0see:
Ok. Just looked at the code: https://github.com/jamulussoftware/jamulus/blob/2f7146b1812da58e6984cb6a2771b98cc273ba53/src/main.cpp#L131

It seems as if due to reasons I don't know this is the only cli option not going to qInfo.

pljones:
> going to qInfo.

`qInfo()` doesn't help.

deithohx:
I'm a qt/c++ novice but the qInfo qCritical functions look like stuff to do with the [logging framework](https://doc.qt.io/qt-5/qmessagelogger.html). I would expect logging to be a toggleable side effect of what the program is doing.


With `--version` and `--help` (and maybe `--serverinfo`) the output text is the purpose of the program. So I think the simple and correct thing to do is to just write to stdout with something like [qTextStream](https://doc.qt.io/qt-5/qtextstream.html).

Cribbing an example
    QTextStream(stdout) << QString("Hello world");

QJKX:
(I'll try this comment again 😬)

I'm a qt/c++ novice but the qInfo qCritical functions look like stuff to do with the [logging framework](https://doc.qt.io/qt-5/qmessagelogger.html). I would expect logging to be a toggleable side effect of what the program is doing.

[...]

PR #2244, opened 2022-01-14
Output Help and Version information to StdOut

henkdegroot:

When the command line option for help or version is used, the text output will now be sent to stdout (instead of stderr).
Also a missing linefeed added to the version text.

**Context: Fixes an issue?**

Fixes: #2190
Output is only changed when the provided argument is valid. An incorrect argument still prints the help text to stderr, which is (as far as I can tell) the expected behaviour.

No

Complete

Nothing

## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -24,6 +24,7 @@
 
 #include <QCoreApplication>
 #include <QDir>
+#include <iostream>
 #include "global.h"
 #ifndef HEADLESS
 #    include <QApplication>
@@ -121,15 +122,15 @@ int main ( int argc, char** argv )
         if ( ( !strcmp ( argv[i], "--help" ) ) || ( !strcmp ( argv[i], "-h" ) ) || ( !strcmp ( argv[i], "-?" ) ) )
         {
             const QString strHelp = UsageArguments ( argv );
-            qInfo() << qUtf8Printable ( strHelp );
+            std::cout << qUtf8Printable ( strHelp );
             exit ( 0 );
         }
 
         // Version number ------------------------------------------------------
         if ( ( !strcmp ( argv[i], "--version" ) ) || ( !strcmp ( argv[i], "-v" ) ) )
         {
-            qCritical() << qUtf8Printable ( GetVersionAndNameStr ( false ) );
-            exit ( 1 );
+            std::cout << qUtf8Printable ( GetVersionAndNameStr ( false ) );
+            exit ( 0 );
         }
 
         // Common options:
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -1927,6 +1927,7 @@ QString GetVersionAndNameStr ( const bool bDisplayInGui )
         strVersionText += "\n *** ";
         strVersionText += "\n *** ";
         strVersionText += QCoreApplication::tr ( "Copyright (C) 2005-2021 The Jamulus Development Team" );
+        strVersionText += "\n";
     }
 
     return strVersionText;


pljones (approved):

mirabilos:
Oh no! Help output always goes to stderr in proper Unix utilities!

pljones:
Well, GNU gets it wrong, then.
```
peter@fs-peter:~/git/Jamulus-wip$ cat --help 2>/dev/null
Usage: cat [OPTION]... [FILE]...
Concatenate FILE(s) to standard output.
...
peter@fs-peter:~/git/Jamulus-wip$ cat --help 2>/dev/null
Usage: cat [OPTION]... [FILE]...
Concatenate FILE(s) to standard output.
```

mirabilos:
Interesting, but an outlier. I’ve not seen this before, even with GNU tools.

pljones:
Oh, and:
```
peter@fs-peter:~/git/Jamulus-wip$ bash --help >/dev/null
peter@fs-peter:~/git/Jamulus-wip$
```
... so that's the first two I could think of, both use stdout.

pljones:
```
peter@fs-peter:~/git/Jamulus-wip$ file `which ffmpeg`
/usr/bin/ffmpeg: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=30732af2c96354365e8ffaddc817360ec5f3effc, for GNU/Linux 3.2.0, stripped
peter@fs-peter:~/git/Jamulus-wip$ ffmpeg --help | less
```
... gets nicely piped to `less`.  If it put the help on stderr, it would miss the pipe.

softins:
> Interesting, but an outlier. I’ve not seen this before, even with GNU tools.

```
tony@pi:~ $ ls --help >/tmp/out 2>/tmp/err
tony@pi:~ $ ls -l /tmp/out /tmp/err
-rw-r--r-- 1 tony tony    0 Feb 28 16:47 /tmp/err
-rw-r--r-- 1 tony tony 7744 Feb 28 16:47 /tmp/out
```

```
tony@pi:~ $ sed --help >/tmp/out 2>/tmp/err
tony@pi:~ $ ls -l /tmp/out /tmp/err
-rw-r--r-- 1 tony tony    0 Feb 28 16:51 /tmp/err
-rw-r--r-- 1 tony tony 1838 Feb 28 16:51 /tmp/out
```

```
tony@pi:~ $ cat /etc/os-release
PRETTY_NAME="Raspbian GNU/Linux 11 (bullseye)"
NAME="Raspbian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"
```

mirabilos:
Tony Mountifield dixit:

>PRETTY_NAME="Raspbian GNU/Linux 11 (bullseye)"

Hm, I’m not so much interested in the new ones, but okay.

I’ve learnt that output for interactive users belongs to stderr.
I had lumped help and version output with that unconditionally,
for example in wtf(1).

I’ve tested Debian 5 (lenny) and, indeed, coreutils output
--help and --version to stdout (and gtar --usage), but the
error messages when invalid options/arguments are passed,
properly, end on stderr.

I can see people not wanting to type ls --help 2>&1 | less
but this i̲s̲ unixy standard (compress -h, for example, outputs
its help on stderr; so does nc, openssl(1), ssh…) even though
most unixy tools don’t have a help flag (you’re supposed to
RTFM).

GNU tools as apparent on BSD are split: RCS has no help flag,
CVS and gdb output to stderr, gcc and as to stdout. perl stdout,
same for lynx.

So, I guess, for --help I’m fine (starting now) with either but
will continue emitting to stderr, myself.


Similarily, version flags are rare. Let’s see…

stdout: awk, captoinfo(ncurses), RCS(GNU), flex, compress,
host(BIND4/BSD), binutils/GCC(GNU), less, OpenSSL, Perl, xmlwf

stderr: cpan(Perl), grep, host(BIND9/Debian), units, wtf

(Note that compress probably gained these flags for GNU gzip
compatibiliy only and did not natively have them.)

We have an interesting schism in the Perl world, and BIND even
moved from stdout to stderr, a majority uses stdout though.

Again, I can see people wanting to do var=$(tool -V) but I’ve
seen var=$(gnu_tool --version 2>&1) in the wild (some time ago).

One benefit of version to stderr is that you can continue running;
if you output the version to stdout, you’ve got to exit immediately
afterwards (and ensure not to print anything else before, i.e. the
help flag also must immediately exit).

I guess I’ll allow stdout --version iff it immediately exits.


Given that I’ll revert the reversal of this PR in the next Debian
upload as it fulfills these criteria (but continue advocating to
put them to stderr, to people who will listen anyway ☻).

POSIX doesn’t specify any version, at least, output for utilities
either.

bye,
//mirabilos
-- 
“It is inappropriate to require that a time represented as
 seconds since the Epoch precisely represent the number of
 seconds between the referenced time and the Epoch.”
	-- IEEE Std 1003.1b-1993 (POSIX) Section B.2.2.2

PR #2250, opened 2022-01-18
Disable Power Savings on Windows

henkdegroot:

Disable Windows Power Saving to prevent performance issues when the app (as client or server) is minimized in Windows 11

**Context: Fixes an issue?**

Fixes: #2233 

No

Complete

Tested on Windows 11, waiting for autobuild to complete to repeat test on Windows 10

## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -42,6 +42,11 @@
 #    include "mac/activity.h"
 extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
 #endif
+#if defined( _WIN32 )
+#    include "windows.h"
+#    include "powrprof.h"
+#    pragma comment( lib, "powrprof.lib" )
+#endif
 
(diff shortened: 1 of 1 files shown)


softins:
If this is verified as fixing the issue and not causing any bad side-effects, it would be good to include in 3.8.2

henkdegroot:
The testing I performed was not correct, which made me believe this would fix the issue.
My test was to compile locally and then start the jamulus server from the powershell window I had open. This powershell is started with admin rights. In this case, when jamulus is launched from the command/powershell prompt, the problem does not occur. While starting it from a non-admin command prompt or launching it using the shortcut is causing the issue.

So now also found another solution to the problem, which is to run jamulus server as administrator. Obvious it does not need admin rights, but Windows 11 seems to treat programs running with admin rights differently.

Closing this PR, as it further investigation is needed.

PR #2256, opened 2022-01-20
GUI: Load translations as early as possible to enable translated startup errors

### LINKED ISSUE #2041: Initial "no JACK connection" message has no translation
**Describe the bug**
When Jamulus is started without JACK the message appears:
image

and it is not translated

**To Reproduce**
Just launch Jamulus but without JACK and with localisation other than English

**Operating system**
Arch Linux

### Jamulus Version 3.8.1beta1dev-4b5c6aaa-dirty

**Additional context**
Seems to me like this message is invoked before `QTranslator` loads its stuff.

hoffie:
This PR moves the translation loading to the earliest possible time. As translation loading depends on settings, this also implies that settings loading is moved to an earlier stage. However, this was only possible after splitting the setting loading into early settings (those which do not refer to `pClient` or `pServer`; this includes translations) and those who do rely on the respective object. The split means that the config has to be parsed twice. The alternative would be caching the parsed config, which occupies (a small amount of) memory and is prone to becoming out of date.

The alternative of the settings change would be to modify the Client (and possibly) Server initialization logic. This would imply moving most of the work out of CSound constructors into their own functions which could be explicitly called at a slightly later stage (after loading translations). However, this would be much more code moving as it would have to be done per-platform. Therefore, I decided to go the Settings reorg route.

**Context: Fixes an issue?**
This PR ensures that translations are available when the sound backends are started. It therefore fixes issues with untranslated sound-related error messages during startup (Fixes #2041).

CHANGELOG: GUI: Make startup errors translatable by loading translations as early as possible

No.

Working as expected on Linux/jack.

Other platforms should be tested:
- [ ] WIndows ASIO
- [ ] Windows JACK
- [ ] Mac
- [ ] (Android)
- [ ] (iOS)

In addition:
- [ ] It should be decided whether the usability improvements (translated startup error messages) are worth the code reorganization.

## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -795,6 +795,18 @@ int main ( int argc, char** argv )
         if ( bIsClient )
         {
             // Client:
+
+            // load settings from init-file (command line options override)
+            CClientSettings Settings ( strIniFileName );
+            Settings.LoadEarlySettings ( CommandLineOptions );
+
+            // load translation
+            if ( bUseGUI && bUseTranslation )
+            {
+                CLocale::LoadTranslation ( Settings.strLanguage, pApp );
+                CInstPictures::UpdateTableOnLanguageChange();
+            }
+
             // actual client object
             CClient Client ( iPortNumber,
                              iQosNumber,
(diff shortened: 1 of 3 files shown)


pljones:
Overall, I'd rather this waited (for however long), until my current settings rewrite is done: part of what I'm doing moves the whole of settings loading earlier, so this would become partly redundant.  I'm not going to block it, though, as I've no idea how long I'll take finishing what I'm doing.

pljones:
I'd also avoid putting too much into having non-GUI text translated.  Much of it is technical, and it's the value in the application source code that's needed for finding where it was issued from, not a Google translation back from the reporter's language to English, which might not be a good match.

hoffie:
> I'd also avoid putting too much into having non-GUI text translated. Much of it is technical, and it's the value in the application source code that's needed for finding where it was issued from, not a Google translation back from the reporter's language to English, which might not be a good match.

I fully agree. The main point of this PR was to fix the lack of translations for GUI startup errors. I think it would be nice to fix those for 3.9.0. This PR would do that but currently has conflicts.

@pljones Should I work on resolving them or is it realistic to get your other reworkings which would help here merged for 3.9.0?

ann0see:
Personally I wouldn’t put much effort into this PR and rather wait for the re-ordering to happen.

hoffie:
Ok. I'll close this for now then.

ann0see:
Yes. Did the same with FDroid (doesn’t seem realistic for me at the moment unfortunately). Your code is still there.

Issue #2247, opened 2022-01-17
Prepare Release 3.8.2


ann0see:
**Target timeline**
Scheduled feature freeze / Start of translation process: 2022-01-27
Targeted translation completion date: ~~2022-02-06~~ 2022-02-13
~~Approximate~~ release date: 2022-02-20
Current state: Released  

**Checklist**
- [ ] Assign this issue to the release shepherd who is in charge of managing this checklist.
- [x] Pin this issue
- [x] Ensure that all issues/PR targeted for this release are done by checking the Project board with [the appropriate filter for this release](https://github.com/orgs/jamulussoftware/projects/2?card_filter_query=milestone%3A%22release+3.8.2%22). Remind main developers to review entries in *Waiting on team* state. 
- [x] Agree to de-tag unfinished Issues/PRs.
- [x] Declare a freeze for code and website by updating this Issue and adding a comment. PRs can still be worked on and may get reviewed, but must not be merged unless agreed explicitly.
- [x] Check the `needs documentation` label for any outstanding PRs flagged for this release and remove that label if done.
- [x] Check ./Jamulus -h output against the Include-Client/Server-Commands.md pages (wiki/en) and man page (distributions/Jamulus.1). Update if necessary.
- [x] Write the Changelog based on the list of PRs in "done" state ([example for 3.8.1](https://github.com/orgs/jamulussoftware/projects/2?card_filter_query=milestone%3A%22release+3.8.1%22+is%3Amerged))
- [x] Start Website translations
  - [x] Check for broken links with a link checker on the `next-release` branch and fix them
  - [x] Open a Pull Request from `next-release` to release, set it as "Draft", sanity check for conflicts and any obvious problems. 
  - [x] Declare a full freeze of the `next-release` and `release` branch. No changes should be made from now on to ensure translators don't have to work twice.
  - [x] Check if the list of translators in `tools/create-translation-issues.sh`. Make sure issue text is up-to-date. Add any URLs that will need localisation into the "New/Changed screenshots" section.
  - [x] Create a translation issue for each language with `tools/create-translation-issues.sh` using `web` argument (see notes in script). 
  - [x] If anyone finds critical issues now, all translators must be made aware of them and all languages should be updated.
- [x] Start App translations
  - [x] Generate `.ts` files in master via `lupdate`
  - [x] Check if the list of translators in `tools/create-translation-issues.sh` is up-to-date
[...]

PR #2264, opened 2022-01-22
Update year to 2022

hoffie:

- [x] Automatical updates in translation source files:
`sed -re 's/-2021/-2022/g' -i src/res/translation/*.ts`

- [x] Manual verification/updates of all files found via:
`grep 2021 src/ *.cpp *.h -r |& grep -vP '\.(svg|qm)'`

- [ ] What do to about copyright code headers? I think one is supposed to update them when the file changes, but we currently do not enforce that. `src/main.cpp` still mentions 2020). This does not necessarily have to be done as part of this PR.

**Context: Fixes an issue?**

Thanks to @henkdegroot for the pointer. Release issue: #2247

No.

Ready.

Review.

## Checklist

cc @softins


pljones:
Ideally we'd have a check that any (source) file changed had the Copyright header correct - right latest year and the person making the change included in the list of authors for that file - that would fail the build, like clang-format verification.  Although it's always arguable whether that should be a build breaker.

hoffie:
> Ideally we'd have a check that any (source) file changed had the Copyright header correct - right latest year and the person making the change included in the list of authors for that file - that would fail the build, like clang-format verification. Although it's always arguable whether that should be a build breaker.

Thinking some more, we could do that automatically based on git history as well. I will try to come up with something.

hoffie:
> > Ideally we'd have a check that any (source) file changed had the Copyright header correct - right latest year and the person making the change included in the list of authors for that file - that would fail the build, like clang-format verification. Although it's always arguable whether that should be a build breaker.
> 
> Thinking some more, we could do that automatically based on git history as well. I will try to come up with something.

PR is updated and contains a small tool to update all known (?) global copyright strings, translation sources and copyright headers -- the latter based on `git log`.

The script is idem-potent, so it's safe to run anytime. We should still discuss where to run it, e.g.:
- a CI job on merges to master
- manually as part of the release checklist

ann0see:
What about a scheduled job?

hoffie:
> What about a schedueled job?

I'm currently looking into running this (and auto-committing) after any pushes to master (which hopefully includes Github merges?). The copyright header part is relevant during the whole year, so this would make sense. I'm testing in my private repos and will report back if I have something working.

linux/sound.cpp
```diff
@@ -1,5 +1,5 @@
 /******************************************************************************\
- * Copyright (c) 2004-2020
+ * Copyright (c) 2004-2021
```
ann0see, line 2: There’s something wrong here!

ann0see (changes requested):
Seems inconsistent:

linux/sound.cpp
hoffie: >  or would you really use the "last edited" date? I don’t see too much benefits from not just using the current year.

This was the intention and is what's currently implemented, yes.
Having read some more, the benefit of doing this 100% accurate seems low and the recommendation is to update all files consistenly.

I guess I should drop this logic again to simplify things.

https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html

hoffie:
> What about a scheduled GH actions job? Best case would be that the job automatically opens a PR on 01-01-xxxx

Yes, that should work, there's a `schedule` event for workflows. I still tend to trigger on pushes to master instead:
- The year should only be updated if there are significant changes. Significant changes should always come via a push.
- A PR might introduce new code with wrong copyright dates. Therefore, it makes sense to run after each PR.

We should still check if we want to auto-open a PR (a little bit more effort) or simply auto-fix via a commit by CI (simple, but less transparent).

ann0see:
Doing it via commit once a year might result in unintended conflicts. I‘d definitely be in favour of a PR.

hoffie:
> Doing it via commit once a year might result in unintended conflicts. I‘d definitely be in favour of a PR.

I prefer a PR for transparency as well, especially considering that this should only happen a few times (maybe only once) a year. I have just updated the code and the PR description accordingly.
Removing Draft status.

linux/sound.cpp
hoffie: The updated logic no longer inserts file-specific dates, it always updates to the current year. Resolving.

tools/update-copyright-notices.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -eu
+
+YEAR=$(date +%Y)
+echo "Updating global copyright strings..."
+sed -re 's/(Copyright.*20[0-9]{2}-)[0-9]{4}/\1'$YEAR'/g' -i src/res/translation/*.ts src/util.cpp src/aboutdlgbase.ui
```
ann0see, line 6: Although a bit picky (and we still have a bunch of years left): but could you use a less specific regex in terms of not specifying the prefix 20 for the year?

.github/workflows/update-copyright-notices.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  check-and-update-copyright-notices:
+    if: ${{ github.event_name == 'push' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: ./tools/update-copyright-notices.sh
+      - run: |
+          git config --global user.email "actions@github.com"
```
ann0see, line 19: Does this need to be set explicitly?

.github/workflows/update-copyright-notices.yml
```diff
@@ -0,0 +1,39 @@
...
+          git config --global user.name "github-actions[bot]"
+          pr_branch=update-copyright-notices-$(date +%Y-%m-%d-%H-%M-%S)
```
ann0see, line 21: I‘d prefer something shorter like `updateCopyrightYear/<year>`

.github/workflows/update-copyright-notices.yml
```diff
@@ -0,0 +1,39 @@
...
+          git checkout -b ${pr_branch}
+          git add .
+          git commit -m "Update copyright notices for $(date +%Y)" || exit 0
```
ann0see, line 24:
```suggestion
          git commit -m "AUTO: Update copyright notices for $(date +%Y)" || exit 0
```

Similar to what the website does.

ann0see (commented):
Thanks for all the automation stuff you’ve doing ;-).

ann0see:
@hoffie 

could the PR say something like:

This automated Pull Request updates the copyright messages throughout the source.

Output of […]

.github/workflows/update-copyright-notices.yml
hoffie: Yes. The defaults error out and passing `git commit --author` is not sufficient either.

ann0see:
Although unrelated, the contributor checking script could be automated the same way

tools/update-copyright-notices.sh
hoffie: Done, although I've kept the `2`. I really want to avoid that the script touches an unintended place with its modifications.

.github/workflows/update-copyright-notices.yml
hoffie: Thinking some more, we can even drop the year altogether. There should only ever be one active PR. If a new one is needed, the existing one can be updated (= the branch can be force-pushed). I've updated this accordingly.

.github/workflows/update-copyright-notices.yml
hoffie: Done (although the fact that it's automated should already be clear by looking at the author).

tools/update-copyright-notices.sh
ann0see: Understandable 😀.

.github/workflows/update-copyright-notices.yml
ann0see: Yes. Just a consistency thing.

.github/workflows/update-copyright-notices.yml
```diff
@@ -0,0 +1,42 @@
+name: Update copyright notices
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
ann0see, line 7: Do we need to run this on PR close with the simplified logic?
I‘d prefer to run it once/twice a year. Preferably in January.

hoffie: Not sure, I'd still run it on every push/merge because each PR may introduce old headers again.

Indepdendent of the frequency we can drop the deletion logic if we are OK to permanently have an updateChangelogNotices branch in jamulussoftware/jamulus.

ann0see: Ok. Got your concern. I wouldn’t drop the deletion logic since it might clutter the repo. Maybe we‘d even need to ensure different names of the branch since conflicting history on a local branch and on GH might confuse people pulling the repo.

hoffie:
> could the PR say something like:
> 
> This automated Pull Request updates the copyright messages throughout the source.
> 
> Output of […]

Done. Examples above have been updated to show the updated texts/behavior.

> Although unrelated, the contributor checking script could be automated the same way

Yes. Can you open a seperate issue?

.github/workflows/update-copyright-notices.yml
hoffie: Heh, that was the initial reason for the rather specific branch names. However, having those would clutter people's checkouts (at least when they check them out). I'd keep it like it is now (single branch name) as noone should usually check out those automatic repos. If they do, they'd hopefully know why. :)

ann0see: Yes. Great!

.github/workflows/update-copyright-notices.yml
```diff
@@ -0,0 +1,46 @@
...
+          git checkout -b ${pr_branch}
+          git add .
+          git commit -m "AUTO: Update copyright notices for $(date +%Y)" || exit 0
+          git push origin "+${pr_branch}"
+          existing_pr=$(gh pr list --head "${pr_branch}" --json number --jq '.[].number')
+          if [[ $existing_pr ]]; then
+            gh pr comment "${existing_pr}" --body "PR has been updated again."
+            exit 0
+          fi
+          gh pr create --base test-master --head "${pr_branch}" --title "Update copyright notice(s) for $(date +%Y)" --body $'This automated Pull Request updates the copyright notices throughout the source.\n\nThis PR is the result of a `./tools/update-copyright-notices.sh` run.'
```
ann0see, line 33: You'll probably change test-master to master once this is finished?

ann0see (approved):

.github/workflows/update-copyright-notices.yml
```diff
@@ -0,0 +1,46 @@
...
+          git config --global user.name "github-actions[bot]"
+          pr_branch=updateCopyrightNotices
+          git checkout -b ${pr_branch}
+          git add .
+          git commit -m "AUTO: Update copyright notices for $(date +%Y)" || exit 0
+          git push origin "+${pr_branch}"
+          existing_pr=$(gh pr list --head "${pr_branch}" --json number --jq '.[].number')
+          if [[ $existing_pr ]]; then
+            gh pr comment "${existing_pr}" --body "PR has been updated again."
+            exit 0
```
ann0see, line 31: I'd prefer a if/else instead of exiting here (this makes the code clearer)

.github/workflows/update-copyright-notices.yml
```diff
@@ -0,0 +1,46 @@
...
+          git config --global user.email "actions@github.com"
+          git config --global user.name "github-actions[bot]"
+          pr_branch=updateCopyrightNotices
+          git checkout -b ${pr_branch}
+          git add .
+          git commit -m "AUTO: Update copyright notices for $(date +%Y)" || exit 0
+          git push origin "+${pr_branch}"
+          existing_pr=$(gh pr list --head "${pr_branch}" --json number --jq '.[].number')
+          if [[ $existing_pr ]]; then
+            gh pr comment "${existing_pr}" --body "PR has been updated again."
```
ann0see, line 30: @gilgongo not sure if this message is clear enough?

.github/workflows/update-copyright-notices.yml
hoffie: Done.

.github/workflows/update-copyright-notices.yml
hoffie: Umm, I had missed that. Thanks! Fixed.

.github/workflows/update-copyright-notices.yml
hoffie: What should be changes/added?

ann0see: We should clarify *why* the PR was updated. Is it possible to link the related commit? Then something like:

Updated this PR because of commit xyz
At least for me it wasn't quite clear why this message would come up.

hoffie: > We should clarify why the PR was updated.

Is it really relevant? The updated diff is easily visible from the PR. We don't explain the *why* in more depth when initially creating the PR either. The answer is simple: The script becomes active (opens a PR / updates it) when there's something to fix. This might be due to the year rolling over or due to some old copyright header being introduced.

Looking up the commit which triggered is probably possible in a simple way. But remember that this is push-based, so it will not be easy to tell if it's this commit or rather some parent.

To sum it up: It would certainly be possible, but it would lead to way more code/complexity.

ann0see: Ok. So you'd say it's not worth it. I'm still not happy with a message saying (Yeah, this PR just got updated). What speaks against omitting it completely? What speaks for having it? I mean, you'll always see on GH that the PR got force-pushed and hence updated.

hoffie: Well, it's an edge case (IMO).
I was mostly thinking about protecting against unwanted merges.
E.g.:
- PR is reviewed and close to being merged
- Script updates PR
- PR is merged, thinking that it had been reviewed, without anyone noticing that it was changed meanwhile

This can happen with human PRs as well, but humans might act in good faith and provide a pointer, so I figured the script should try do the same. :)

pljones: In which case 
> "PR has been updated again."

could be
> "PR was updated by tools/update-copyright-headers"

or similar, to make it clear.

hoffie: Done.

.github/workflows/update-copyright-notices.yml
softins: Personally, I am a big fan of early exit/return, and find it much clearer.

If there is nothing further to do, why not exit at that point, rather than have the reader need to follow through the else/fi to find there is nothing else below?

hoffie: > Personally, I am a big fan of early exit/return, and find it much clearer.

Same here. I guess this instance is less clear than others though as there are two code paths where neither is the clear default. If the script was to be extended, I couldn't predict which of those two paths it would be.

.github/workflows/update-copyright-notices.yml
ann0see: Ok. Probably it’s ok then.

.github/workflows/update-copyright-notices.yml
```diff
@@ -0,0 +1,46 @@
...
+          pr_branch=updateCopyrightNotices
+          git checkout -b ${pr_branch}
+          git add .
+          git commit -m "AUTO: Update copyright notices for $(date +%Y)" || exit 0
+          git push origin "+${pr_branch}"
+          existing_pr=$(gh pr list --head "${pr_branch}" --json number --jq '.[].number')
+          if [[ $existing_pr ]]; then
+            gh pr comment "${existing_pr}" --body "PR has been updated by tools/update-copyright-notices.sh."
+          else
+            gh pr create --base master --head "${pr_branch}" --title "Update copyright notice(s) for $(date +%Y)" --body $'This automated Pull Request updates the copyright notices throughout the source.\n\nThis PR is the result of a `tools/update-copyright-notices.sh` run.'
```
pljones, line 32: Could you `\` wrap this line, please.  (We don't have clang-format for YAML :) )

hoffie: Hrm, even when wrapping the text stays rather long. I've therefore moved the text to a variable and build that in two lines. Should be better now.

tools/update-copyright-notices.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -eu
+
+YEAR=$(date +%Y)
+echo "Updating global copyright strings..."
+sed -re 's/(Copyright.*2[0-9]{3}-)[0-9]{4}/\1'$YEAR'/g' -i src/res/translation/*.ts src/util.cpp src/aboutdlgbase.ui
+
+echo "Updating copyright comment headers..."
+for file in $(find -regex '.*\.\(cpp\|h\|mm\)' -not -regex '\./\(\.git\|libs/\|moc_\|ui_\).*'); do
```
softins, line 9: I just tried this out and found it had the same issue as the recent clang_format one: it found all the files in build directories too.
```suggestion
for file in $(find android ios linux mac src windows -regex '.*\.\(cpp\|h\|mm\)' -not -regex '\./\(\.git\|libs/\|moc_\|ui_\).*'); do
```

hoffie: Thansk, updated.

pljones (approved):

softins (approved):

Issue #2274, opened 2022-01-25
Parts of PR #1994 (wrong display of Unicode characters at line wrap) got lost


hoffie:
**Describe the bug**

`TruncateString()` in `src/util.cpp` is currently unused and generates a compiler warning on Mac due to this.
This function was initially introduced in ce370e95d43b2051ba20160e5ec50f13b2de5748 via #1994, master merge commit 51589960a80d91b8d5c681349ac21a91d53fefdd. While the function is still there, all the other code from this commit is missing.

I suspect that a (later) bad merge has caused this.

After lots of tracing directly on master, I had a look at the PR's commits again and something seems fishy there:
- ce370e95d43b2051ba20160e5ec50f13b2de5748 adds `TruncateString` + the call in `src/util.cpp` (as expected)
- a8853b6a7deb19e3596fb47bbb72a5baf29c4c7a is a merge commit which still contains the function definition, but lacks any call of the function

So, I believe that the merge commit unwittingly dropped the call. This is no surprise though, as the merge commit likely had severe conflicts to do a parallel huge code reorganization through d94fd69e1404a874beec53d6a721abcd74caf8fb, which moved `CMusProfDlg::OnAliasTextChanged` (where the lost change was) to `CClientSettingsDlg::OnAliasTextChanged`.

- [x] Find the root cause
- [x] Attempt to find further affected PRs/commits: As the problem occured directly within the PR, I don't assume any systematic problems.
- [x] Restore any missing commits #2275

**Version of Jamulus**

master @ 25018800b704cdf23bffa837911a485309ca216e

**Additional context**

djfun:
Before removing the function, please check what happened to the context where it was used.

hoffie:
I have found the likely reason for the missing caller and have updated the issue description accordingly. Will submit a PR shortly.

PR #1994, opened 2021-09-05
Fix for #1451 - unicode in mixer should wrap on character boundaries (take 2)

pljones:
**Short description of changes**

@djfun wrote a fix but didn't raise a PR, so here's the PR.

**Context: Fixes an issue?**

Yes.  #1688.  This still happens.

**Does this change need documentation? What needs to be documented and how?**

No.

**Status of this Pull Request**

Needs testing.

**What is missing until this pull request can be merged?**

Needs testing.

## Checklist
- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [ ] I tested my code and it does what I want
- [ ] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [ ] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [ ] I've filled all the content above

----

I've now created a branch off jamulussoftware/jamulus master and merged @djfun's changes to that, then fixed the conflicts.

I'll try it out now.

PR #2275, opened 2022-01-25
Client: Fix wrong truncation of unicode aliases

hoffie:

If the 16th char of the user-selected alias was a multi-byte unicode char, it would be incorrectly split in half, rendering it invalid. This commit makes the truncation unicode-aware.

The fix was initially developed in ce370e95d43b2051ba20160e5ec50f13b2de5748. It got lost during PR merge in commit a8853b6a7deb19e3596fb47bbb72a5baf29c4c7a, most likely during conflict resolution due to code reorganizations in d94fd69e1404a874beec53d6a721abcd74caf8fb.

This commit re-applies the fix to the new place (`CClientSettingsDlg::OnAliasTextChanged` instead of the no longer
existing `CMusProfDlg::OnAliasTextChanged`). It also adds a function declaration to the header to make inclusion by other files possible in the first place. This also required dropping the `static` and `inline` keywords.

Co-authored-by: @djfun

See #2274 for more context.

**Context: Fixes an issue?**

Fixes: #1994 (for real)
Fixes: #2274

No. ChangeLog has already been slightly adjusted to include this issue number.

Ready, should go into 3.8.2.

Needs review.

## Checklist

- [x] I tested my code and it does what I want (Linux)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -18,8 +18,8 @@
 - Client: Added Conductor to list of instruments (#2140).
   (contributed by @henkdegroot)
 
-- Client: Fix wrong display of Unicode characters at line wrap (#1994).
-  (contributed by @djfun, @pljones)
+- Client: Fix wrong display of Unicode characters at line wrap and settings screen (#1994, #2274).
+  (contributed by @djfun, @pljones, @hoffie)
 
 - Client: Fixed incorrect operation of feedback detection on first connect in run (#2120).
   (contributed by @softins)
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -1093,7 +1093,7 @@ void CClientSettingsDlg::OnAliasTextChanged ( const QString& strNewName )
     else
     {
         // text is too long, update control with shortened text
-        pedtAlias->setText ( strNewName.left ( MAX_LEN_FADER_TAG ) );
+        pedtAlias->setText ( TruncateString ( strNewName, MAX_LEN_FADER_TAG ) );
     }
 }
 
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -667,7 +667,7 @@ void CLanguageComboBox::OnLanguageActivated ( int iLanguageIdx )
     }
 }
 
-static inline QString TruncateString ( QString str, int position )
+QString TruncateString ( QString str, int position )
 {
     QTextBoundaryFinder tbfString ( QTextBoundaryFinder::Grapheme, str );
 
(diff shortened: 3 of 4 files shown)


softins:
> It also moves the related inline function to the header to make it possible to include in the first place.

Did you mean "inline" rather than "include"?

As a minor thing, I'm not a fan of non-trivial code being in a `.h` rather than a `.cpp`. Does this function really need to be inline?

hoffie:
> > It also moves the related inline function to the header to make it possible to include in the first place.
> 
> Did you mean "inline" rather than "include"?

What I meant is that without moving the file to the (included) `uti.h` header, it is not usable outside of `util.cpp`. I have adjusted the PR and the commit text in the hope that it is clearer now.

> As a minor thing, I'm not a fan of non-trivial code being in a `.h` rather than a `.cpp`. Does this function really need to be inline?

I agree, but I initially tried to keep the code as-unmodified as possible. Keeping the `static` definition required me to move the whole function to the header. When dropping `static`, one would still get compiler warnings due to `inline` (probably because [`inline`only makes sense in header files for reusable code](https://stackoverflow.com/questions/1421666/qt-creator-inline-function-used-but-never-defined-why/1421730#1421730)?).
I don't see a reason for inlining though. This is not a performance-critical function and even if it was, compilers are usually better at deciding this, I guess...

So, I've dropped both `static` and `inline` now, kept the function in `util.cpp` and added the declaration to `util.h`.

djfun:
the original issue mentions the test string "Mr. Hu 🍽", where the last character is problematic

softins:
> the original issue mentions the test string "`Mr. Hu 🍽`", where the last character is problematic

Yes, that was the one that showed it up when trying to line-wrap the name on the mixer display, and was already fixed. But this PR is addressing truncation after 16 characters in the settings window, which I couldn't make go wrong in a prior version, with a multi-byte char at the end.

djfun:
> > the original issue mentions the test string "`Mr. Hu 🍽`", where the last character is problematic
> 
> Yes, that was the one that showed it up when trying to line-wrap the name on the mixer display, and was already fixed. But this PR is addressing truncation after 16 characters in the settings window, which I couldn't make go wrong in a prior version, with a multi-byte char at the end.

Try putting this string in the settings window alias field and then insert characters before the multi-byte character until you get a question mark character at the end.

hoffie:
> I was unable to demonstrate the problem in a version prior to this fix, so couldn't do a before and after comparison.

Sorry, should have documented my test case as it also took me some time to understand and reproduce the problem.

Before this fix: `12345678901234😍` works, adding `5` makes Jamulus display an invalid character (`?`)
After this fix:  `12345678901234😍` works, adding `5` properly discards the whole Unicode character at the end (no `?`).

pljones (approved):
Thanks!

PR #2292, opened 2022-01-26
src/buffer.cpp: Avoid CodeQL warning about multiplication overflow

softins:

Satisfy CodeQL multiplication overflow warnings by changing a loop variable from `int` to `long`

**Context: Fixes an issue?**

Fixes some high-severity code alerts raised by Github, even though they wouldn't be encountered in practice.

No documentation needed

Trivial change, ready to merge

Validate that the warnings on `src/buffer.cpp` are no longer raised by Github

## Checklist

### DIFF
--- a/src/buffer.cpp
+++ b/src/buffer.cpp
@@ -146,6 +146,7 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
         const int iNumBlocks = /* floor */ ( iInSize / iBlockSize );
 
         // copy new data in internal buffer
+        // iBlock is a long to avoid overflowing a mulitplication later in the code
         for ( long iBlock = 0; iBlock < iNumBlocks; iBlock++ )
         {
             // extract sequence number of current received block (per definition
@@ -263,6 +264,7 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
         // copy new data in internal buffer
         const int iNumBlocks = iInSize / iBlockSize;
 
+        // iBlock is a long to avoid overflowing a mulitplication later in the code
         for ( long iBlock = 0; iBlock < iNumBlocks; iBlock++ )
         {
             // for simultion buffer only update pointer, no data copying
--- a/src/buffer.cpp
+++ b/src/buffer.cpp
@@ -146,7 +146,7 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
         const int iNumBlocks = /* floor */ ( iInSize / iBlockSize );
 
         // copy new data in internal buffer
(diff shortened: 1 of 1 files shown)


ann0see:
At least the warning in the files tab disappeared. Hopefully changing the type doesn’t introduce any bugs. But I doubt that. CodeQL still seems to find issues, but I can’t find them. Probably we’ll see after a merge

src/buffer.cpp
```diff
@@ -146,7 +146,7 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
         const int iNumBlocks = /* floor */ ( iInSize / iBlockSize );
 
         // copy new data in internal buffer
-        for ( int iBlock = 0; iBlock < iNumBlocks; iBlock++ )
+        for ( long iBlock = 0; iBlock < iNumBlocks; iBlock++ )
```
hoffie, line 149:
```suggestion
        // iBlock is a long to avoid overflowing a mulitplication later in the code
        for ( long iBlock = 0; iBlock < iNumBlocks; iBlock++ )
```

src/buffer.cpp
```diff
@@ -263,7 +263,7 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
         // copy new data in internal buffer
         const int iNumBlocks = iInSize / iBlockSize;
 
-        for ( int iBlock = 0; iBlock < iNumBlocks; iBlock++ )
+        for ( long iBlock = 0; iBlock < iNumBlocks; iBlock++ )
```
hoffie, line 266:
```suggestion
        // iBlock is a long to avoid overflowing a mulitplication later in the code
        for ( long iBlock = 0; iBlock < iNumBlocks; iBlock++ )
```

hoffie (approved):
The fix looks correct to me (after having read some [background](https://www.ti.com/lit/an/spra683/spra683.pdf?ts=1643269623888)).
As I understand, this PR changes the type of two loop variables in a way so that a later multiplication involving this loop variable is forced to occur in the larger type (long) instead of the type of all the operands (int).

I guess one other, potentially more readable solution would be casting one of the operands to `long` in place, but I assume that this is -- at least theoretically -- less efficient instead of doing it once per loop as you do.

I guess it could use a comment to avoid later "improvements" where someone wonders why it is a `long` and reverts back to `int`. See suggestions.

CHANGELOG: Bug Fix: A hypothetical CodeQL warning about a multiplication overflow in sound buffer handling has been fixed.

softins:
In the longer term, I would prefer to eliminate multiplications within array indices. I prefer to traverse with suitably-incremented simple indices or pointers. But the immediate need was to eliminate the red warnings in Github.

ann0see (approved):
Ok. Let‘s try it.

PR #1999, opened 2021-09-07
Qt6

dcorson-ticino-com:
Compiles using Qt6 and Qt5
- changes in string constants
- certain functions have other parameters under Qt6
- the biggest change is that Qt6 uses other locale codes than Qt5, functions added to convert between Qt5 and Qt6 locale
- some Qt functions changed names

Compiles using Qt6 and Qt5

Maybe Qt6 is friendlier with MacOS

**Status of this Pull Request**
This compiles and works under Win10

**What is missing until this pull request can be merged?**
Testing under the other OSes

## Checklist

- [ ] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) at least my local clang-format, we will see with the autobuild
- [ ] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [ ] I've filled all the content above

Issue #490, opened 2020-08-03
Server List Is Not Accessible With Screen Reader on Mac 


chigkim:
1. Open server list to join.
2. Turn on VoiceOver with command+f5 (or command+triple click power on newer mac with with touch bar)
3. Try navigating the list with tab and arrows
None of the QTreeWidgetItems are not announced as being selected.
However, if you move the mouse cursor to one of the item and press control+option+f5, you'll hear what's under the mouse.
It seems like QTreeWidgetItems are accessible, but they don't appear as children of QTreeWidget to accessibility API.

corrados:
> However, if you move the mouse cursor to one of the item and press control+option+f5, you'll hear what's under the mouse.

I have tried this but it does not work on my Mac. Maybe it is an issue of my old MacOS.

> It seems like QTreeWidgetItems are accessible, but they don't appear as children of QTreeWidget to accessibility API.

That sounds to me like a but in Qt since the QTreeWidgetItems are always children of QTreeWidget, right? Or is there a possibility to add QTreeWidgetItems in a QTreeWidget if they are not children?

chigkim:
Sorry I'm not sure why control+option+f5 didn't work for you.
When a sighted person manually move my mouse and puts it on one of the 
server, and if I press control+option+f5, VO announced the item for me.
So far, my dirty hack around was to add a each item as a button to the 
tree. This has its own problem where you can't expand with right arrow 
to look at the users, but at least people can look at the servers and 
join one.
Inside CConnectDlg::SetServerList
```
for (int i=0; i<lvwServers->columnCount(); i++) {
     lvwServers->setItemWidget(pNewListViewItem, i, new 
QPushButton(pNewListViewItem->text(i)));
}
```
Inside CConnectDlg::SetPingTimeAndNumClientsResult
```
dynamic_cast<QPushButton*>(lvwServers->itemWidget(pCurListViewItem, 
1))->setText(pCurListViewItem->text(1));
dynamic_cast<QPushButton*>(lvwServers->itemWidget(pCurListViewItem, 
2))->setText(pCurListViewItem->text(2));
```

corrados:
I just tried it out with the screen reader which is built-in the Windows 10 operating system. That screen reader can read all list items without any problem. You can even let it read the connected musicians names.

So I assume that we have an issue which is Qt and Mac related. I don't think I can fix anything in the Jamulus software to get it working.

[...]

PR #2299, opened 2022-01-27
Code: Add Qt6 compatibility

hoffie:

This PR adds the necessary changes to allow building with both Qt5 and Qt6 (Qt6 **compatibility**). It does not switch the autobuild targets (although it does contain changes/refactorings to make that switch simple).

The vast amount of development was done by @dcorson-ticino-com.
@softins had already helped with rebasing and his branch was of huge help to validate the changes in my branch.
I've tried to reflect those authorships in the commits, either as commit author or as Co-authored-by. If you're unhappy with any of this, please ping me and I'll adjust accordingly.

Two larger parts by @dcorson-ticino-com's tree are not part of mine, because I haven't found a reason to include that:
1. [The Jamulus.pro update with win64 logic](https://github.com/jamulussoftware/jamulus/pull/1999/files#diff-3b059d403dbbf0c4c127642ee3318bc2c3b10f70f929b71e2746c4a6ef907244R126). I don't think it is necessary and I couldn't get it to work like that at all. The existing build logic already handles both 32bit and 64bit, so I think we should be fine.
1. [The `enum class ESvrRegStatus` change](https://github.com/jamulussoftware/jamulus/pull/1999/files#diff-40ab92941994ec9418dbb76ae46d5aef0e4cb8b034f2be2b31562bb9aca8070fR576). For me, the existing code compiled properly for both Qt5 and Qt6, but maybe I'm missing something?

Other than that, this PR differs in the following ways:
- git history: Several parts of development iterations have been squashed together. At the same time, history has been split into independent changes. Therefore, it's **very important that this PR is NOT squashed**.
- Mac Dock Badge Label handling has been restored
- Autobuild code has been reworked to cover both the Qt6 use case and our existing requirements on `master`
- I've done some minor cleanups such as moving Qt header includes to .h files instead of .cpp.

**Context: Fixes an issue?**

- Qt6 support is mandatory as long as we use Qt, because there will be no more open source Qt5 releases. Relevant discussion: https://github.com/jamulussoftware/jamulus/discussions/1795
- Platform-specific bugs have come up ([Mac performance issues/window focus issues](https://github.com/jamulussoftware/jamulus/issues/1791), [Windows 11 sound issues](https://github.com/jamulussoftware/jamulus/discussions/2233)) without us knowingly breaking anything. So there's at least a very small hope that simply updating our framework to a version which officially supports those platforms could help there as well.
- Apparently there is [hope for Jamulus to become more screenreader-friendly](https://github.com/jamulussoftware/jamulus/issues/490#issuecomment-1022566525)

- [ ] We should check if COMPILING.md or similar documentation should be updated.

CHANGELOG: Build: Initial support for building with Qt6 has been implemented (@dcorson-ticino-com, @softins)

Ready, I'm not planning further changes right now.

Testing. Most of the changes are build-related and either work (compile) or don't (CI failure). I'm pretty confident that this PR is pretty safe for Qt5 and is a huge improvement for Qt6 (from "doesn't compile" to "compiles and runs"). There might be some bugs in Qt6 builds for sure.
There's one thing that should get explicit testing: The commit `Mac: Use native Dock Badge handling` drops Qt's macextras usage in favor of a small native ObjC wrapper. Therefore:
- [x] Create tests for the different flag handling code conversion approaches
- [x] Do Country code conversion in Settings to ensure that updates continue working
- [x] Confirm that https://github.com/jamulussoftware/jamulus/discussions/2148#discussioncomment-2287669 is fixed
- [x] Mac: Confirm that Dock Label updates still work properly
- [ ] Reviews: I suggest reviewing by commit, not (only) by full diff. I have added explanations to most of the commits.

All in all, the full diff is delightfully small and I hope to get this merged soon after 3.8.2 is done.

## Checklist


softins (commented):
Looking good so far. One small point. Maybe the conditional:
```
#if QT_VERSION >= 0x060000
```
should be changed to:
```
#if QT_VERSION >= QT_VERSION_CHECK( 6, 0, 0 )
```
for consistency with the existing version checks?

src/recorder/creaperproject.cpp
```diff
@@ -62,19 +62,19 @@ CReaperItem::CReaperItem ( const QString& name, const STrackItem& trackItem, con
 
     QTextStream sOut ( &out );
```
pljones, line 64: All of the recorder files _want_ to flush eash line.

src/recorder/jamcontroller.cpp
```diff
@@ -91,15 +91,15 @@ void CJamController::SetRecordingDir ( QString newRecordingDir, int iServerFrame
     {
         pJamRecorder         = new recorder::CJamRecorder ( newRecordingDir, iServerFrameSizeSamples );
         strRecorderErrMsg    = pJamRecorder->Init();
```
pljones, line 93: I've a change for this waiting on two other PRs.

src/serverlogging.cpp
```diff
@@ -70,7 +70,7 @@ void CServerLogging::operator<< ( const QString& sNewStr )
     {
         // append new line in logging file
         QTextStream out ( &File );
```
pljones, line 72: The log file wants to flush on each write.

pljones:
This PR appears to mix up a number of different changes.  Please split out those changes to the build scripts from changes to the source code.

Note that most of the changes to the source for Qt6 _that are wanted_ have been done -- most of those included here do not appear to be wanted.

src/serverlogging.cpp
hoffie: I agree, I mis-interpreted `File.flush()` which (probably?) only affects the underlying file, but not the `QTextStream`. I've updated this part to call `out.flush()` (but dropped the `File.flush` as I understand [it is redundant in that case](https://github.com/qt/qtbase/blob/5558eb4edb62ddc920fd5f4171e732fb7cbe53a7/src/corelib/serialization/qtextstream.cpp#L72)).

hoffie:
Thanks for your comments, @softins and @pljones.

@softins 
> Maybe the conditional:  `#if QT_VERSION >= 0x06000` should be changed to `#if QT_VERSION >= QT_VERSION_CHECK( 6, 0, 0 )` for consistency with the existing version checks?

That's certainly more readable and consistent. Updated, thanks!

@pljones 
> This PR appears to mix up a number of different changes. Please split out those changes to the build scripts from changes to the source code.

What do you mean by *split out* exactly? I've already invested lots of time to split those changes by their intention and did make them separate commits with explanations why they are needed. Do you mean to split those commits even further?
Or would you prefer them as individual PRs? I can do that, but the only comments regarding that question so far were that [too many PRs would be more cumbersome(https://github.com/jamulussoftware/jamulus/discussions/1795#discussioncomment-2043838), which is why I went the two-PR route.

So yes, this PR does contain both changes to source code and build logic because its intention is to provide Qt6 compatibility. It tries hard not to change existing behavior though. That's what #2300 is about.

It's also correct that this PR includes seemingly unrelated changes (refactorings), but I consider them a necessity. For example, the build script re-organization (especially on Windows) is necessary in order to allow building with two different Qt versions (while still keeping those two the same in this PR). 
If you think that moving the build-related updates to the other PR (#2300) would be better, I can do that, of course. I think that's simply two different approaches. I had tried to distinguish between "safe to merge after review with no visible changes to release artifacts" vs. "change releases to Qt6", but distinguishing by type of change (code vs. build logic) would probably work as well, yeah. In the end, the separation into different commits is already there, so it's only a matter about how to distribute them among PRs.

> Note that most of the changes to the source for Qt6 _that are wanted_ have been done -- most of those included here do not appear to be wanted.

Not sure I follow. Do you suggest to drop specific changes from the PR, if so, which? In what way would they not be wanted?
I can only think of one previous merged PR with the goal of improving Qt6 compatibility, which was about `QRegularExpression`. That was certainly necessary, but not sufficient for Qt6 compatibility. This PR tries to add what's missing.

softins:
If these changes are all related to the same objective (which I think they are), I don't see any point in splitting them into separate PRs for the sake of it. That just makes unnecessary work, which would have the same end result once they had all been merged. It's not as if the changes in this PR are huge and very many.

pljones:
> @pljones
> 
> > This PR appears to mix up a number of different changes. Please split out those changes to the build scripts from changes to the source code.
> 
> What do you mean by _split out_ exactly? I've already invested lots of time to split those changes by their intention and did make them separate commits with explanations why they are needed. Do you mean to split those commits even further? Or would you prefer them as individual PRs? I can do that, but the only comments regarding that question so far were that [too many PRs would be more cumbersome](https://github.com/jamulussoftware/jamulus/discussions/1795#discussioncomment-2043838), which is why I went the two-PR route.
> 
> So yes, this PR does contain both changes to source code and build logic because its intention is to provide Qt6 compatibility. It tries hard not to change existing behavior though. That's what #2300 is about.
> 
> It's also correct that this PR includes seemingly unrelated changes (refactorings), but I consider them a necessity. For example, the build script re-organization (especially on Windows) is necessary in order to allow building with two different Qt versions (while still keeping those two the same in this PR). If you think that moving the build-related updates to the other PR (#2300) would be better, I can do that, of course. I think that's simply two different approaches. I had tried to distinguish between "safe to merge after review with no visible changes to release artifacts" vs. "change releases to Qt6", but distinguishing by type of change (code vs. build logic) would probably work as well, yeah. In the end, the separation into different commits is already there, so it's only a matter about how to distribute them among PRs.

The point is to keep the opportunity for introducing problems as discrete as possible.  If you're changing the _build_, then you know - by definition - any problems introduced into Jamulus's behaviour are because of changes to the build process, _not_ the code.

If you change both at the same time, you have more possible places a single problem could be introduced, making it harder to diagnose.

> 
> > Note that most of the changes to the source for Qt6 _that are wanted_ have been done -- most of those included here do not appear to be wanted.
> 
> Not sure I follow. Do you suggest to drop specific changes from the PR, if so, which? In what way would they not be wanted? I can only think of one previous merged PR with the goal of improving Qt6 compatibility, which was about `QRegularExpression`. That was certainly necessary, but not sufficient for Qt6 compatibility. This PR tries to add what's missing.
> 

None of the recorder changes are "missing" because they're either in another PR already or not wanted, as far as I know.

> Edit: Maybe you are referring to small changes such as `Document dmgbuild as reason for staying on macOS 10.15`. Those are technically not needed for Qt6 compatibility, of course. I was just trying to follow a [Clean Code principle](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html). If such changes should rather be in their own PRs, I'd find that rather tedious, but if that's the consensus then I could do that as well, but they'll often be hard to work with due to a likely tendency for conflicts.

hoffie:
> The point is to keep the opportunity for introducing problems as discrete as possible. If you're changing the build, then you know - by definition - any problems introduced into Jamulus's behaviour are because of changes to the build process, not the code.
> If you change both at the same time, you have more possible places a single problem could be introduced, making it harder to diagnose.

I fully agree there, that's why I've split everything into manageable pieces.
Do you suggest to convert those 16 commits into individual PRs? As said, I can do it, but I don't see the benefit to do that "discrete" part on a PR level instead of on a commit level. This would also be likely 16 PRs where most of those would have to be merged in a specific order as they would probably depend on each other.

> None of the recorder changes are "missing" because they're either in another PR already or not wanted, as far as I know.

Could you be more specific, please? Are you referring to #1999 (which was not ready for merge and was closed)? This PR is based on @dcorson-ticino-com's work in #1999. Are you referring to other PRs? If so, which? Or had you been planning to open those?
What part exactly is "not wanted"?
I can only further improve this PR if I know what to do and currently I don't.

hoffie:
> Do you suggest to convert those 16 commits into individual PRs? As said, I can do it, but I don't see the benefit to do that "discrete" part on a PR level instead of on a commit level. This would also be likely 16 PRs where most of those would have to be merged in a specific order as they would probably depend on each other.

What I could do is: Move the refactoring parts in a single refactoring PR (along with other autobuild refactorings I'm currently doing). Would that be better?
It will still have the downside that this PR would depend on the refactoring PR.

pljones:
> > Do you suggest to convert those 16 commits into individual PRs? As said, I can do it, but I don't see the benefit to do that "discrete" part on a PR level instead of on a commit level. This would also be likely 16 PRs where most of those would have to be merged in a specific order as they would probably depend on each other.
> 
> What I could do is: Move the refactoring parts in a single refactoring PR (along with other autobuild refactorings I'm currently doing). Would that be better? It will still have the downside that this PR would depend on the refactoring PR.

That's what I'm suggesting, yes.  Do not try to do everything at once.  That's why I have a stack of PRs _waiting_.

hoffie:
Done. I have limited the scope of this PR to changes to the actual code. I've opened #2328 to cover the (auto)build changes. #2300 is unchanged for now and I suggest we'd ignore it for now (that's why it's a draft). I'll open yet another PR when the other (non-Qt6) autobuild refactorings are done.

src/audiomixerboard.cpp
```diff
@@ -747,8 +747,13 @@ void CChannelFader::SetChannelInfos ( const CChannelInfo& cChanInfo )
     // Country flag icon -------------------------------------------------------
     if ( cChanInfo.eCountry != QLocale::AnyCountry )
     {
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 0, 0 )
+        QLocale::Country eC = CLocale::LocaleQt5toQt6 ( cChanInfo.eCountry );
```
pljones, line 751: Generally, single letter variable names are avoided.  This should be `eCountry`.

src/recorder/creaperproject.cpp
pljones: Actually, in that case, could we drop the `; sOut` from most of the lines and just have one big statement...
```
sOut << "    <ITEM " << '\n'
     << "      FADEIN 0 0 0 0 0 0" << '\n'
     ...
     << "      >" << endl;

sOut.flush();
```
(I'd still prefer the double flush.)

src/util.h
```diff
@@ -634,7 +634,7 @@ inline QString svrRegStatusToString ( ESvrRegStatus eSvrRegStatus )
         return QCoreApplication::translate ( "CServerDlg", "Requirements not fulfilled" );
     }
 
-    return QString ( QCoreApplication::translate ( "CServerDlg", "Unknown value " ) ).append ( eSvrRegStatus );
```
pljones, line 637: Translations affected.

hoffie: Yep. If this is merged for 3.9.0 it should be no problem though, right?
If you think it's ok to merge for 3.8.2, then we could manually fix the .ts files in this case or ignore it for now. The latter should be OK as well as this is a code path which is never supposed to be taken ("last resort fallback").

src/audiomixerboard.cpp
hoffie: Thanks, updated.

src/recorder/creaperproject.cpp
hoffie: > Actually, in that case, could we drop the ; sOut from most of the lines and just have one big statement...

Done (here and in line 118; the other places with 3 lines or less did not make sense, especially since `clang-format` would start putting those into a single line then).

> (I'd still prefer the double flush.)

Do you mean `endl`? If so, why, is it any safer that way?
What should I do -- add a `#define` as initially thought? It sounded overkill and looked not that readable to me.

src/util.h
pljones: I'd stick with 3.9.0 for this.

src/connectdlg.cpp
```diff
@@ -411,8 +411,11 @@ void CConnectDlg::SetServerList ( const CHostAddress& InetAddr, const CVector<CS
 
         if ( vecServerInfo[iIdx].eCountry != QLocale::AnyCountry )
         {
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 0, 0 )
+            QString strCountryToString = QLocale::countryToString ( CLocale::LocaleQt5toQt6 ( vecServerInfo[iIdx].eCountry ) );
```
pljones, line 415: Hm.  Reading this now raises concerns.

We need backwards compatibility -- but I don't think this is the right way to achieve it.  The content of `vecServerInfo[iIdx].eCountry` should be Qt6-compatible at this point in the code, so however that vec is populated should be doing the translate.

I think that would mean fewer changes to the code.

src/recorder/creaperproject.cpp
pljones: It was probably something I read on StackOverflow and may have been about Qt4 (as that's what we were targetting when this code was written).  So it may not apply now.  I think the concern was that sometimes there would be an intermediate buffer, so the first flush (endl) flushed to the file (which was buffered) and the second flush said "Oh, the first buffer is clean, I better flush the file buffer to disk".  Or something.

It's fairly important to avoid missing the opportunity to save what's being written here -- the audio has been saved but the track timings relative to each other aren't easy to recover (but I've a script which helps).

src/serverlogging.cpp
pljones: As noted above, I'd understood at some point that both were needed - the text stream was buffer and then the underlying file was separately buffered.  The text stream flush gets to into the file but not necessarily to disk, which is why I think it was like this.

Again, as above, that might have been a Qt4 idiom that no longer applies - but it's worth checking if it's applicable in Qt5, at least, as we won't be moving to Qt6 for a long time, I expect.

src/connectdlg.cpp
hoffie: So, the origin of `vecServerInfo` is:
CProtocol::CLServerListReceived -> CClient::CLServerListReceived -> CConnectDlg::OnCL(Red)ServerListReceived -> CConnectDlg::SetServerList

Not sure if it'll really be fewer changes, but I agree that it would be more correct to do it on the lowest possible level.

@pljones So I guess the best place would be protocol.cpp, so that the contract would be that the caller always puts in / gets out a value which is valid for the currently used Qt version, right?

hoffie: I have pushed that now.
Old: https://github.com/jamulussoftware/jamulus/pull/2299/commits/4292008f78f553d37a71d56be42d9d64180d2266
New: https://github.com/jamulussoftware/jamulus/pull/2299/commits/0c6055cde8443215eb4f201c2999472c78de118b

So, the new approach (protocol layer handling) requires more changes. However:
- It handles more cases (the previous approach seemed like it didn't handle the Server UI, for example)
- It includes the helper script which updates the code translation table in order to make that reproducible and maintainable.
- I believe the new approach as suggested by you (@pljones) is better at using the right layer of abstraction instead of having to think about this in all the places where this is used.

There's one caveat though: The country flag handling currently hardcodes a code to flag name mapping. According to the comments, this is due to Qt4 compatibility. The code already includes code which should work for Qt5+. So, for Qt6 support in the flag handling we've got two options:
- The protocol layer gives us Qt-native (i.e. Qt6 in Qt6) country codes now. As the hardcoded `switch` uses Qt5 country codes, we would have to convert *back*. I've currently implemented this approach because it limits the changes to the Qt6-specific `#ifdef`.
- We could drop the hardcoding (I don't think we are Qt4-compatible anymore anyway?) in favor of Volker's Qt-native code which should (and seems to) work with both Qt5 and Qt6. I think this is the better solution. I'm just unsure if we should do that within this PR or not.

pljones: We no longer support build on a Qt 4 environment.  However, for clarity, does the comment mean that or is it talking about something else, like a Qt4 client talking (via a server) to a Qt5 client?

hoffie: I'm pretty sure it's about the implementation. To be extra safe, I'm going to build a test case which compares the output of the commented out logic with the hardcoded values.

hoffie: **Test environment**
Special Jamulus builds for Qt5 and Qt6 based on current Qt6-compat branch from this PR with a patch to output code+flag for all supported country codes. Details:
<details>
<summary>Setup instructions/Patch</summary>

```
$ git checkout 9207e341ccc34be93ebde130d44282b3cbdad9fd
$ patch -p1
diff --git a/src/main.cpp b/src/main.cpp
index 3ae7aab2..f2b8a2e7 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -50,6 +50,27 @@ extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
 
 {
+    bool bUseTable = argc > 1 ? std::string ( argv[1] ) != "qt" : true;
+    qInfo() << "qt version:" << QT_VERSION_STR;
+    qInfo() << "mode:" << ( bUseTable ? "table" : "qt" );
+    for ( int iCurCntry = static_cast<int> ( QLocale::AnyCountry ); iCurCntry < static_cast<int> ( QLocale::LastCountry ); iCurCntry++ )
+    {
+        if ( static_cast<QLocale::Country> ( iCurCntry ) == QLocale::AnyCountry )
+        {
+            // continue;
+        }
+
+        if ( !CLocale::IsCountryCodeSupported ( iCurCntry ) )
+        {
+            continue;
+        }
+
+        qInfo() << "code" << iCurCntry << "-" /*<< QLocale::countryToString ( static_cast<QLocale::Country> ( iCurCntry ) ) << " - " */
+                << CLocale::GetCountryFlagIconsResourceReference (
+                       CLocale::QtCountryToWireFormatCountry ( static_cast<QLocale::Country> ( iCurCntry ) ),
+                       bUseTable );
+    }
+    exit ( 0 );
 
 #if defined( Q_OS_MACX )
     // Mnemonic keys are default disabled in Qt for MacOS. The following function enables them.
diff --git a/src/util.cpp b/src/util.cpp
index 364dd86e..f442272c 100644
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -1246,7 +1246,7 @@ bool CLocale::IsCountryCodeSupported ( uint iCountryCode )
 #endif
 }
 
-QString CLocale::GetCountryFlagIconsResourceReference ( const QLocale::Country eCountry /* Always a Qt5 (!) code */ )
+QString CLocale::GetCountryFlagIconsResourceReference ( const QLocale::Country eCountry /* Always a Qt5 (!) code */, bool bUseTable )
 {
     QString strReturn = "";
 
@@ -1255,6 +1255,9 @@ QString CLocale::GetCountryFlagIconsResourceReference ( const QLocale::Country e
     {
         return ":/png/flags/res/flags/flagnone.png";
     }
+
+    if ( bUseTable )
+    {
         // NOTE: The following code was introduced to support old QT versions. The problem
         //       is that the number of countries displayed is less than the one displayed
         //       with the new code below (which is disabled). Therefore, as soon as the
@@ -1827,15 +1830,15 @@ QString CLocale::GetCountryFlagIconsResourceReference ( const QLocale::Country e
             return "";
         }
         return strReturn;
-
+    }
+    else
+    {
         // AT LEAST QT 4.8 IS REQUIRED:
-    /*
+
         // There is no direct query of the country code in Qt, therefore we use a
         // workaround: Get the matching locales properties and split the name of
         // that since the second part is the country code
-    QList<QLocale> vCurLocaleList = QLocale::matchingLocales ( QLocale::AnyLanguage,
-            QLocale::AnyScript,
-            eCountry );
+        QList<QLocale> vCurLocaleList = QLocale::matchingLocales ( QLocale::AnyLanguage, QLocale::AnyScript, eCountry );
 
         // check if the matching locales query was successful
         if ( vCurLocaleList.size() > 0 )
@@ -1860,7 +1863,7 @@ QString CLocale::GetCountryFlagIconsResourceReference ( const QLocale::Country e
             }
         }
         return strReturn;
-    */
+    }
 }
 
 QMap<QString, QString> CLocale::GetAvailableTranslations()
diff --git a/src/util.h b/src/util.h
index 56bb6ff6..413b4552 100644
--- a/src/util.h
+++ b/src/util.h
@@ -809,7 +809,7 @@ protected:
 class CLocale
 {
 public:
-    static QString                 GetCountryFlagIconsResourceReference ( const QLocale::Country eCountry /* Always a Qt5 (!) code */ );
+    static QString GetCountryFlagIconsResourceReference ( const QLocale::Country eCountry /* Always a Qt5 (!) code */, bool bUseTable = true );
     static QMap<QString, QString>  GetAvailableTranslations();
     static QPair<QString, QString> FindSysLangTransFileName ( const QMap<QString, QString>& TranslMap );
     static void                    LoadTranslation ( const QString strLanguage, QCoreApplication* pApp );
^D

$ make distclean; qmake && make -j && mv Jamulus Jamulus-qt5
$ make distclean; qmake6 && make -j && mv Jamulus Jamulus-qt6
$ ./Jamulus-qt5 table &> /tmp/qt5-table
$ ./Jamulus-qt5 qt &> /tmp/qt5-qt
$ ./Jamulus-qt6 table &> /tmp/qt6-table
$ ./Jamulus-qt6 qt &> /tmp/qt6-qt
```
</details>

**Qt5: Table (existing) vs. `QLocale::matchingLocales`**
Result: `QLocale::matchingLocales` supports 53 more flags and doesn't lack a single one.
```
diff -u0 /tmp/qt5-{table,qt} | grep -v '@@' | grep -c '^-code'
53
```
<details>
<summary>diff -u0 /tmp/qt5-{table,qt} | grep -v '@@'</summary>

```diff
--- /tmp/qt5-table	2022-02-08 01:02:36.072798047 +0100
+++ /tmp/qt5-qt	2022-02-08 01:02:36.086131545 +0100
-mode: table
+mode: qt
-code 4 - ""
+code 4 - ":/png/flags/res/flags/as.png"
-code 7 - ""
+code 7 - ":/png/flags/res/flags/ai.png"
-code 9 - ""
+code 9 - ":/png/flags/res/flags/ag.png"
-code 16 - ""
+code 16 - ":/png/flags/res/flags/bs.png"
-code 19 - ""
+code 19 - ":/png/flags/res/flags/bb.png"
-code 22 - ""
+code 22 - ":/png/flags/res/flags/bz.png"
-code 24 - ""
+code 24 - ":/png/flags/res/flags/bm.png"
-code 31 - ""
+code 31 - ":/png/flags/res/flags/io.png"
-code 40 - ""
+code 40 - ":/png/flags/res/flags/ky.png"
-code 45 - ""
-code 46 - ""
+code 45 - ":/png/flags/res/flags/cx.png"
+code 46 - ":/png/flags/res/flags/cc.png"
-code 51 - ""
+code 51 - ":/png/flags/res/flags/ck.png"
-code 60 - ""
+code 60 - ":/png/flags/res/flags/dm.png"
-code 70 - ""
+code 70 - ":/png/flags/res/flags/fk.png"
-code 72 - ""
+code 72 - ":/png/flags/res/flags/fj.png"
-code 80 - ""
+code 80 - ":/png/flags/res/flags/gm.png"
-code 84 - ""
+code 84 - ":/png/flags/res/flags/gi.png"
-code 87 - ""
+code 87 - ":/png/flags/res/flags/gd.png"
-code 89 - ""
+code 89 - ":/png/flags/res/flags/gu.png"
-code 94 - ""
+code 94 - ":/png/flags/res/flags/ht.png"
-code 107 - ""
+code 107 - ":/png/flags/res/flags/jm.png"
-code 112 - ""
+code 112 - ":/png/flags/res/flags/ki.png"
-code 121 - ""
+code 121 - ":/png/flags/res/flags/lr.png"
-code 129 - ""
+code 129 - ":/png/flags/res/flags/mw.png"
-code 131 - ""
+code 131 - ":/png/flags/res/flags/mv.png"
-code 134 - ""
+code 134 - ":/png/flags/res/flags/mh.png"
-code 140 - ""
+code 140 - ":/png/flags/res/flags/fm.png"
-code 144 - ""
+code 144 - ":/png/flags/res/flags/ms.png"
-code 149 - ""
+code 149 - ":/png/flags/res/flags/nr.png"
-code 158 - ""
-code 159 - ""
-code 160 - ""
+code 158 - ":/png/flags/res/flags/nu.png"
+code 159 - ":/png/flags/res/flags/nf.png"
+code 160 - ":/png/flags/res/flags/mp.png"
-code 164 - ""
+code 164 - ":/png/flags/res/flags/pw.png"
-code 171 - ""
+code 171 - ":/png/flags/res/flags/pn.png"
-code 180 - ""
-code 181 - ""
-code 182 - ""
-code 183 - ""
+code 180 - ":/png/flags/res/flags/kn.png"
+code 181 - ":/png/flags/res/flags/lc.png"
+code 182 - ":/png/flags/res/flags/vc.png"
+code 183 - ":/png/flags/res/flags/ws.png"
-code 193 - ""
+code 193 - ":/png/flags/res/flags/sb.png"
-code 199 - ""
-code 200 - ""
+code 199 - ":/png/flags/res/flags/sh.png"
+code 200 - ":/png/flags/res/flags/pm.png"
-code 203 - ""
+code 203 - ":/png/flags/res/flags/sj.png"
-code 213 - ""
+code 213 - ":/png/flags/res/flags/tk.png"
-code 215 - ""
+code 215 - ":/png/flags/res/flags/tt.png"
-code 218 - ""
-code 219 - ""
-code 220 - ""
+code 218 - ":/png/flags/res/flags/tm.png"
+code 219 - ":/png/flags/res/flags/tc.png"
+code 220 - ":/png/flags/res/flags/tv.png"
-code 226 - ""
+code 226 - ":/png/flags/res/flags/um.png"
-code 229 - ""
-code 230 - ""
+code 229 - ":/png/flags/res/flags/vu.png"
+code 230 - ":/png/flags/res/flags/va.png"
-code 233 - ""
-code 234 - ""
-code 235 - ""
+code 233 - ":/png/flags/res/flags/vg.png"
+code 234 - ":/png/flags/res/flags/vi.png"
+code 235 - ":/png/flags/res/flags/wf.png"
```
</details>

**Qt6: Table (existing + Qt6-back-translation in 9207e341ccc34be93ebde130d44282b3cbdad9fd) vs. `QLocale::matchingLocales`**
Result: `QLocale::matchingLocales` supports 53 more flags and doesn't lack a single one.
```
diff -u0 /tmp/qt6-{table,qt} | grep -v '@@' | grep -c '^-code'
53
```
<details>
<summary>diff -u0 /tmp/qt6-{table,qt} | grep -v '@@'</summary>

```diff
--- /tmp/qt6-table	2022-02-08 01:02:36.096131669 +0100
+++ /tmp/qt6-qt	2022-02-08 01:02:36.106131793 +0100
-mode: table
+mode: qt
-code 6 - ""
+code 6 - ":/png/flags/res/flags/as.png"
-code 9 - ""
+code 9 - ":/png/flags/res/flags/ai.png"
-code 11 - ""
+code 11 - ":/png/flags/res/flags/ag.png"
-code 14 - ""
+code 14 - ":/png/flags/res/flags/um.png"
-code 20 - ""
+code 20 - ":/png/flags/res/flags/bs.png"
-code 23 - ""
+code 23 - ":/png/flags/res/flags/bb.png"
-code 26 - ""
+code 26 - ":/png/flags/res/flags/bz.png"
-code 28 - ""
+code 28 - ":/png/flags/res/flags/bm.png"
-code 36 - ""
-code 37 - ""
+code 36 - ":/png/flags/res/flags/io.png"
+code 37 - ":/png/flags/res/flags/vg.png"
-code 51 - ""
-code 52 - ""
+code 51 - ":/png/flags/res/flags/ky.png"
+code 52 - ":/png/flags/res/flags/tc.png"
-code 58 - ""
+code 58 - ":/png/flags/res/flags/cx.png"
-code 65 - ""
-code 66 - ""
+code 65 - ":/png/flags/res/flags/ck.png"
+code 66 - ":/png/flags/res/flags/vi.png"
-code 75 - ""
+code 75 - ":/png/flags/res/flags/dm.png"
-code 78 - ""
+code 78 - ":/png/flags/res/flags/zw.png"
-code 89 - ""
+code 89 - ":/png/flags/res/flags/fk.png"
-code 91 - ""
+code 91 - ":/png/flags/res/flags/fj.png"
-code 98 - ""
+code 98 - ":/png/flags/res/flags/gm.png"
-code 103 - ""
+code 103 - ":/png/flags/res/flags/gi.png"
-code 106 - ""
+code 106 - ":/png/flags/res/flags/gd.png"
-code 108 - ""
+code 108 - ":/png/flags/res/flags/gu.png"
-code 114 - ""
+code 114 - ":/png/flags/res/flags/ht.png"
-code 118 - ""
+code 118 - ":/png/flags/res/flags/cc.png"
-code 121 - ""
+code 121 - ":/png/flags/res/flags/vu.png"
-code 132 - ""
+code 132 - ":/png/flags/res/flags/jm.png"
-code 138 - ""
+code 138 - ":/png/flags/res/flags/ki.png"
-code 147 - ""
+code 147 - ":/png/flags/res/flags/lr.png"
-code 155 - ""
+code 155 - ":/png/flags/res/flags/mw.png"
-code 157 - ""
-code 158 - ""
+code 157 - ":/png/flags/res/flags/tv.png"
+code 158 - ":/png/flags/res/flags/mv.png"
-code 161 - ""
+code 161 - ":/png/flags/res/flags/mh.png"
-code 166 - ""
+code 166 - ":/png/flags/res/flags/fm.png"
-code 171 - ""
+code 171 - ":/png/flags/res/flags/ms.png"
-code 178 - ""
+code 178 - ":/png/flags/res/flags/nr.png"
-code 186 - ""
-code 187 - ""
-code 188 - ""
+code 186 - ":/png/flags/res/flags/nu.png"
+code 187 - ":/png/flags/res/flags/nf.png"
+code 188 - ":/png/flags/res/flags/mp.png"
-code 194 - ""
+code 194 - ":/png/flags/res/flags/pw.png"
-code 200 - ""
+code 200 - ":/png/flags/res/flags/pm.png"
-code 205 - ""
+code 205 - ":/png/flags/res/flags/pn.png"
-code 211 - ""
+code 211 - ":/png/flags/res/flags/wf.png"
-code 217 - ""
+code 217 - ":/png/flags/res/flags/sh.png"
-code 220 - ""
-code 221 - ""
-code 222 - ""
-code 223 - ""
+code 220 - ":/png/flags/res/flags/kn.png"
+code 221 - ":/png/flags/res/flags/lc.png"
+code 222 - ":/png/flags/res/flags/vc.png"
+code 223 - ":/png/flags/res/flags/ws.png"
-code 235 - ""
+code 235 - ":/png/flags/res/flags/sb.png"
-code 237 - ""
+code 237 - ":/png/flags/res/flags/va.png"
-code 246 - ""
+code 246 - ":/png/flags/res/flags/sj.png"
-code 249 - ""
+code 249 - ":/png/flags/res/flags/tk.png"
-code 257 - ""
+code 257 - ":/png/flags/res/flags/tt.png"
```
</details>

**Is a single flag image used for multiple codes?**
No:
```
grep -oP '":.*"' /tmp/qt* | sort | uniq -c | grep -v '1 '
```


**Is any flag from the reference (Qt5, table) missing in any other scenario (Qt5/QLocale, Qt6/Table, Qt6/Qlocale)?**
```
$ for flag in $(grep -oP '":.*"' /tmp/qt5-table | sort); do for scenario in /tmp/qt5-qt /tmp/qt6-table /tmp/qt6-qt; do grep -qF "$flag" "$scenario" || echo "$scenario is missing $flag"; done; done
/tmp/qt6-table is missing ":/png/flags/res/flags/zw.png"
```
- [ ] This will have to be investigated as it might point towards a problem in the code translation table. Looks like the reason might be that Qt5 uses `240` for two different entries. Should probably adjust the logic to last-one-wins to work around that?
```
$ grep 240 /usr/include/qt/QtCore/qlocale.h
        Duala = 240,
        Zimbabwe = 240,
$ grep Zimbabwe /usr/include/qt6/QtCore/qlocale.h
        Zimbabwe = 261,
        LastTerritory = Zimbabwe,
$ grep Duala /usr/include/qt6/QtCore/qlocale.h
        Duala = 71,
```
**Summary**
It looks like we could safely switch to the `QLocale::matchingLocales` approach as it delivers all of what the table delivers + 53 additional flags.
- [ ] We should check that those new flag assignments are correct.

ann0see:
On Windows 11 there seem to be strange lines on the GUI which disappear on click. It looks as if I marked some invisible button:
image

image

src/connectdlg.cpp
hoffie: The issue for Zimbabwe missing was a off-by-one error in the translation table generation script and its output. After fixing this, the last check from the previous comment does not yield anything anymore (as expected) in 3ac8c17d2df0bec436f3f6c34e29c82771e7a5f6.

However, since switching to the `QLocale`-based lookups is a change in functionality (it adds 53 countries/flags!), I'm splitting this out into its own PR.

dcorson-ticino-com:
> On Windows 11 there seem to be strange lines on the GUI which disappear on click. It looks as if I marked some invisible button:

Very weird, there is nothing at that position in the GUI, does it appear when there are users, you are connected?

I hope it is not caused my MS's ever deeper split between processing and display which has made some versions of MS Office unusable.  As I have completely stopped using MS Office I don't know if they have improved things recently, but using certain versions of Excel one never knew if the screen display corresponded to the data and spent half ones time provoking screen updates.

ann0see:
Yes. These lines also appear if I‘m connected (but at different positions)

dcorson-ticino-com:
Could it be a question of display driver update?

src/protocol.cpp
```diff
@@ -2772,6 +2772,12 @@ bool CProtocol::GetStringFromStream ( const CVector<uint8_t>& vecIn, int& iPos,
     return false; // no error
 }
 
+QLocale::Country CProtocol::GetCountryFromStream ( const CVector<uint8_t>& vecIn, int& iPos )
```
pljones, line 2775: Mmm, I'd leave this as the 2 byte value until it's been through `CLocale::WireFormatCountryToQtCountry`.  Turning it into a _current_ Qt QLocale::Country isn't right: it's always a Qt 5 (or less) QLocale, so it's invalid if, say, Qt8 changes QLocale::Country into something the `static_cast<QLocale::Country> ( GetValFromStream ( vecIn, iPos, 2 ) )` would fail for.

I can't see the point of doing half the work here and half the work there.

Here, just get the ushort value off the stream and pass it to CLocale::WireFormatCountryToQtCountry to get the country.

src/protocol.cpp
```diff
@@ -2884,3 +2890,9 @@ void CProtocol::PutStringUTF8OnStream ( CVector<uint8_t>& vecIn, int& iPos, cons
         PutValOnStream ( vecIn, iPos, static_cast<uint32_t> ( sStringUTF8[j] ), 1 );
     }
 }
+
+void CProtocol::PutCountryOnStream ( CVector<uint8_t>& vecIn, int& iPos, QLocale::Country eCountry )
```
pljones, line 2894: Same again, here, `CLocale::QtCountryToWireFormatCountry` should hide everything and return a `ushort`.

src/protocol.cpp
hoffie: Um, thanks for that suggestion. I have no idea why I did it the way I did initially. I fully agree with what you said.

I've now
- renamed `CLocale::WireFormatCountryToQtCountry` to `CLocale::WireFormatCountry`**`Code`**`ToQtCountry` and updated it to return an `unsigned short` instead of `QLocale::Country`
- removed the obsolete, redundant casting in `CProtocol::GetCountryFromStream`

Can you re-check? Is `us` the proper prefix for `unsigned short`? `usCountry...` has potential for confusion, but I've got no idea what else to call it.

src/protocol.cpp
hoffie: Thanks. I've
- renamed `CLocale::QtCountryToWireFormatCountry` to `CLocale::QtCountryToWireFormatCountry`**`Code`** and updated it to take an `unsigned short` instead of `QLocale::Country`
- removed the obsolete, redundant casting in `CProtocol::PutCountryOnStream`

src/main.cpp
```diff
@@ -971,7 +974,7 @@ QString UsageArguments ( char** argv )
            "  -L, --licence         show an agreement window before users can connect\n"
            "  -m, --htmlstatus      enable HTML status file, set file name\n"
            "  -o, --serverinfo      registration info for this server.  Format:\n"
-           "                        [name];[city];[country as QLocale ID]\n"
+           "                        [name];[city];[country as Qt5 QLocale ID]\n"
```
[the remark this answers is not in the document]
hoffie: Opened #2362 as an ugliness tracker. :)

hoffie:
@ann0see @dcorson-ticino-com I've opened #2363 to track the Windows/Mac glitches. Let's continue analysis there.
I think we should keep this PR focused on getting the code right (which is tedious enough). Strangeness may come up on Qt6 builds and we should track and fix this of course, but I think it's easier to handle if we do that in separate issues.

src/protocol.cpp
pljones: Yeah, we don't use a lot of unsigned shorts...  And I agree, `usCountry` is not nice.  `wfCountry` (wire-format)? :)

src/util.cpp
```diff
@@ -1211,619 +1211,656 @@ CInstPictures::EInstCategory CInstPictures::GetCategory ( const int iInstrument
 }
 
 // Locale management class -----------------------------------------------------
-QString CLocale::GetCountryFlagIconsResourceReference ( const QLocale::Country eCountry )
+QLocale::Country CLocale::WireFormatCountryCodeToQtCountry ( unsigned short eCountryCode )
```
pljones, line 1214: `eCountryCode` there wants _not_ to be `e...`.

src/util.cpp
```diff
@@ -1211,619 +1211,656 @@ CInstPictures::EInstCategory CInstPictures::GetCategory ( const int iInstrument
...
     }
-    else
+    // NOTE: The following code was introduced to support old QT versions. The problem
+    //       is that the number of countries displayed is less than the one displayed
+    //       with the new code below (which is disabled). Therefore, as soon as the
+    //       compatibility to the very old versions of QT is not required anymore, use
+    //       the new code.
+    // COMPATIBLE FOR OLD QT VERSIONS -> use a table:
+    QString strISO3166   = "";
+    uint    iCountryCode = static_cast<uint> ( eCountry );
```
pljones, line 1265: So here we're using a `uint` (not `unsigned short`) and calling the variable `i...` (without a `u`)...

Hmm!  Not consistent.

src/util.cpp
```diff
@@ -1211,619 +1211,656 @@ CInstPictures::EInstCategory CInstPictures::GetCategory ( const int iInstrument
...
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 0, 0 )
+    // When using Qt6, the protocol layer has already translated the original Qt5 codes
+    // to their Qt6 equivalents. However, the table below expects Qt5 codes,
+    // so we'll have to translate back...:
+    iCountryCode = qt6CountryToQt5Table[iCountryCode];
+#endif
+
+    switch ( iCountryCode )
```
pljones, line 1273: Is the switch really better than a `Map<unsigned short, String>` lookup?  (Obviously, only define the Map for Qt6, only do the lookup for Qt6.)

The nice thing then is we can mirror it with a `Map<String, unsigned short>` for the server info country lookup.  (Here two maps, one for Qt5, one for Qt6.)

src/util.h
```diff
@@ -809,10 +809,40 @@ class CInstPictures
...
+    static QString                 GetCountryFlagIconsResourceReference ( const QLocale::Country eCountry /* Always a Qt5 (!) code */ );
     static QMap<QString, QString>  GetAvailableTranslations();
     static QPair<QString, QString> FindSysLangTransFileName ( const QMap<QString, QString>& TranslMap );
     static void                    LoadTranslation ( const QString strLanguage, QCoreApplication* pApp );
+    static QLocale::Country        WireFormatCountryCodeToQtCountry ( unsigned short usCountryCode );
+    static unsigned short          QtCountryToWireFormatCountryCode ( const QLocale::Country eCountry );
+    static bool                    IsCountryCodeSupported ( unsigned short iCountryCode );
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 0, 0 )
+    // ./tools/qt5-to-qt6-country-code-table.py generates these lists:
+    constexpr int const static qt5CountryToQt6Table[] = {
```
pljones, line 821: Can we rename these `wireFormatTo...` and `...ToWireFormat`?

henkdegroot:
Just want to report that it seems there is some issue with my country being changed from in settings, after moving to the Qt6 version. My normal setup was running 3.8.1 and my country was set to Netherlands. Then I installed the Qt6 build from here: https://github.com/hoffie/jamulus/actions/runs/1809196472, and started Jamulus. Now my country changed to Mayotte.

Not sure if this already should work correct, but just dropping a note here.

src/connectdlg.cpp
hoffie: The flag-related change is now split into #2409. Will set this PR to draft (also for the unresolved settings stuff) and will rebase/continue once #2409 is done.

src/protocol.cpp
hoffie: I've ran a quick grep over the current code and the `i` prefix is used for anything-int, independent of signedness or width. Therefore, I've reverted this back to `iCountryCode` for now.

src/connectdlg.cpp
hoffie: With #2409 merged and this PR rebased, I'm marking this Flag-related thread as resolved. It also cleared out several other review comments as the suggestions were around code which has been eliminated by now. :)

ann0see:
@hoffie I checked the country code checkbox as I think it has been fixed.

You'll need to rebase this PR.

hoffie:
> @hoffie I checked the country code checkbox as I think it has been fixed.

Only the Flag-related improvements in #2409 are resolved. The checklist item was supposed to be about the fact the the Settings handling also needs to consider this. This isn't done yet. I've therefore unchecked this item again. Maybe I'll have time for some more work on this PR tomorrow.

> You'll need to rebase this PR.

Done.

.github/autobuild/ios.sh
```diff
@@ -19,7 +19,14 @@ setup() {
     else
         echo "Install dependencies..."
         python3 -m pip install "aqtinstall==${AQTINSTALL_VERSION}"
-        python3 -m aqt install-qt --outputdir "${QT_DIR}" mac ios "${QT_VERSION}" --archives qtbase qttools qttranslations qtmacextras
+        # Install actual ios Qt:
+        python3 -m aqt install-qt --outputdir "${QT_DIR}" mac ios "${QT_VERSION}" --archives qtbase qttools qttranslations
+        if [[ ! "${QT_VERSION}" =~ 5\..* ]]; then
```
pljones, line 24: Are you using the _QT_ version to check the _target platform_?  That doesn't seem right.  Why not have parameter or separate variable?

Jamulus.pro
```diff
@@ -211,7 +210,13 @@ win32 {
     # liboboe requires C++17 for std::timed_mutex
     CONFIG += c++17
 
-    QT += androidextras
+    lessThan(QT_MAJOR_VERSION, 6) {
+        QT += androidextras
+    } else {
+        # Qt >=6 does not have an androidextras module anymore.
```
pljones, line 216: "anymore" -> "any more"

.github/autobuild/ios.sh
hoffie: Nah, it's quite something else: This is all about the ios target platform.
The issue is that Qt5 ships qmake as part of the aqt installation, while Qt6 only ships a shim which redirects to the mac/desktop qmake. Therefore, this is needed additionally for Qt6.

Maybe the comment should be improved?

pljones: Yes :)

hoffie: Done.

Jamulus.pro
hoffie: Done.

src/util.h
hoffie: Done.

hoffie:
- Rebased
- Added calls to translate country codes back and forth for ini settings (both clients and servers). This should fix the problem reported by @henkdegroot and @ashstrahle in https://github.com/jamulussoftware/jamulus/discussions/2148#discussioncomment-2287669.

@pljones I'm no longer sure if the "wireFormat" suffix/prefix is still sane now that this is used for ini files as well. I struggle to find a better name though."externalCountryCode"? "protocolCountryCode"? "jamulusCountryCode"? Keep "wireFormat"?

pljones:
> @pljones I'm no longer sure if the "wireFormat" suffix/prefix is still sane now that this is used for ini files as well. I struggle to find a better name though."externalCountryCode"? "protocolCountryCode"? "jamulusCountryCode"? Keep "wireFormat"?

I'd ask _why_ the value stored in the inifile goes through this - if it's because its to keep to "protocol format" for the value, then "wireFormat" is reasonably evocative.  If the protocol ever changed, would we change the inifile?

hoffie:
> I'd ask why the value stored in the inifile goes through this - if it's because its to keep to "protocol format" for the value, then "wireFormat" is reasonably evocative.

Well, until now, country codes have been an implementation detail of both the inifile and the protocol. Both happen to use Qt5's country representation. As such, the implementation details became part of external contracts. While the protocol part cannot be fixed without long migration periods, there are more options for the inifile as it is local to a specific client.

Some approaches which come to mind:
1. Always store the Qt-native code.
   That's what has been done until now and is going to cause breakage (wrong country) when migrating from Qt5 to Qt6 builds (or when switching back-and-forth for whatever reasons). If we wanted to stay Qt-native in this place, we would have to add additional information how this number is to be interpreted (i.e. is it Qt5 or Qt6). I'd really like to avoid it as it would require translation logic in both directions.
2. Stop using implementation details and store something like the ISO country code.
   In general, I'd be open to this idea. It would require inifile migration code though. It would require more code/logic and would still require the Qt5-to-Qt6 code conversion as part of the migration logic. It's certainly an improvement, but it doesn't have to be necessarily done as part of this PR.
3. Accept that Jamulus made Qt5 country codes part of its contract in multiple places (protocol, inifile, serverinfo #2362).
   That's the simplest solution for now, but obviously doesn't really improve the status quo. It would be a compatibility helper, nothing more, nothing less. That's what's currently included in this PR.

Personally, I'd like to go with (3) to finally get this PR merged (after all, the scope is adding compatibility and not fixing  such non-optimal decisions of the past [easy to say in hindsight...;) ]). If wanted, I could work on (2) in the near future though.

> If the protocol ever changed, would we change the inifile?

I assume that the current protocol is not going to change this part. If at all, it's going to be "forked", i.e. Qt5 codes extended for new countries if really required. This would have to be reflected in the inifile in the same way.
Any future iteration of the protocol should probably get rid of the semantic Qt5 dependency (e.g. by using ISO codes instead) and it would only be logical for the inifile to do the same.
In short: Yes.

So I guess with that reasoning we could keep saying wireFormat* even for the inifile?

pljones:
> So I guess with that reasoning we could keep saying wireFormat* even for the inifile?

Yes, I think the argument is that the reason to stick that number in the inifile is that we're tied to it by the on-the-wire protocol value -- it's the wire format country code.  That it's also the Qt5 (actually, Qt4...) value isn't the point: we're stuck with it so Server A can register with Directory B and Client C and Client D can identify each other, the servers and directories, without needless translation.

Not being able to put "GB" or "US" or "CN" or whatever for a country code has been wrong since the begining... it's no more wrong now :).

hoffie:
@pljones Thanks. Would this PR then be ready for a final round of reviews? I'm not planning any further changes right now.
@ann0see Could you check "Mac: Confirm that Dock Label updates still work properly"? It would be sufficient/important to check the Qt5 builds from this PR, not mainly Qt6 builds.

ann0see:
Jamulus Usercount

@hoffie I can confirm on my VM that the artifact indeed shows the user count. I think the new code even shows it in the launchpad now.

.github/autobuild/ios.sh
```diff
@@ -19,7 +19,17 @@ setup() {
...
+            # Starting with Qt6, ios' qtbase install does no longer include a real qmake binary.
+            # Instead, it is a script which invokes the mac desktop qmake.
+            # As of aqtinstall 2.0.6 / 03/2022, desktop qtbase has to be installed manually:
+            python3 -m aqt install-qt --outputdir "${QT_DIR}" mac desktop "${QT_VERSION}" --archives qtbase
+            # Additionally, Qt6 has renamed the Mac binary location from clang_64 to macos.
```
ann0see, line 29: I know you opened an issue related to this at aqtinstall. Is there anything which needs to be changed here?

src/clientsettingsdlg.cpp
```diff
@@ -561,21 +561,31 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     for ( int iCurCntry = static_cast<int> ( QLocale::AnyCountry ); iCurCntry < static_cast<int> ( QLocale::LastCountry ); iCurCntry++ )
     {
         // exclude the "None" entry since it is added after the sorting
-        if ( static_cast<QLocale::Country> ( iCurCntry ) != QLocale::AnyCountry )
+        if ( static_cast<QLocale::Country> ( iCurCntry ) == QLocale::AnyCountry )
```
ann0see, line 564: Is the comment still valid here? You're now inverting the condition and use IsCountrySuppoerted()

src/main.cpp
```diff
@@ -21,6 +21,9 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
  *
 \******************************************************************************/
+#ifdef _MSC_VER
+#    pragma comment( linker, "/subsystem:windows /ENTRY:mainCRTStartup" )
+#endif
```
ann0see, line 26: Could you please add a comment to this what this does Linker Flags? Maybe move it elsewhere down if possible

src/serverdlg.cpp
```diff
@@ -308,12 +308,22 @@ lvwClients->setMinimumHeight ( 140 );
     for ( int iCurCntry = static_cast<int> ( QLocale::AnyCountry ); iCurCntry < static_cast<int> ( QLocale::LastCountry ); iCurCntry++ )
     {
         // add all countries except of the "Default" country
-        if ( static_cast<QLocale::Country> ( iCurCntry ) != QLocale::AnyCountry )
+        if ( static_cast<QLocale::Country> ( iCurCntry ) == QLocale::AnyCountry )
```
ann0see, line 311: As above

src/util.cpp
```diff
@@ -1202,7 +1202,42 @@ CInstPictures::EInstCategory CInstPictures::GetCategory ( const int iInstrument
...
+#else
+    return (unsigned short) eCountry;
+#endif
+}
+
+bool CLocale::IsCountryCodeSupported ( unsigned short iCountryCode )
+{
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 0, 0 )
+    // On newer Qt versions there might be codes which do not have a Qt5 equivalent.
+    // We have no way to support those sanely right now.
```
ann0see, line 1231: Should be a // TODO: comment for future versions (if there's a workaround found)

src/util.h
```diff
@@ -648,7 +648,7 @@ inline QString svrRegStatusToString ( ESvrRegStatus eSvrRegStatus )
         return QCoreApplication::translate ( "CServerDlg", "Requirements not fulfilled" );
     }
 
-    return QString ( QCoreApplication::translate ( "CServerDlg", "Unknown value " ) ).append ( eSvrRegStatus );
+    return QString ( QCoreApplication::translate ( "CServerDlg", "Unknown value %1" ) ).arg ( eSvrRegStatus );
```
ann0see, line 651: Unknown value of what? The registration status msg? Although not related to this PR, this is not very user friendly.

tools/qt5-to-qt6-country-code-table.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
```
ann0see, line 1: Did you check this file via Codacy?

ann0see (commented):
The artifact (macOS) runs with Qt5. I think we should do a real world test after this is merged.

ann0see (approved):
Did a smoke client and server test on Linux. I hope we can merge this soon.

src/serverdlg.cpp
hoffie: Still applies, IMO.

src/util.cpp
hoffie: Hrm, the comment just intends to explain why we do a sanity check here. We don't know if and in what way this might become an issue, so there's no specific TODO right now, I'd say?

hoffie (commented):
- [ ] Check Windows loader adjustment -- can it be dropped?
- [ ] Check aqt update #2597  -- can the symlink workaround be dropped?
- [ ] flakes8 *.py

.github/autobuild/ios.sh
hoffie: I have tested the removal in https://github.com/hoffie/jamulus/commit/5adde603213b936539b77ad5b4f82a8f460ff28e and it works. Therefore, I've removed the symlink workaround here as well (the manual desktop install is still required).

tools/qt5-to-qt6-country-code-table.py
hoffie: Those two hints have been fixed now.

Jamulus.pro
```diff
@@ -88,6 +88,14 @@ win32 {
...
             winmm.lib \
             ws2_32.lib
+        greaterThan(QT_MAJOR_VERSION, 5) {
+            # Qt5 had a special qtmain library which took care of forwarding the MSVC default WinMain() entrypoint to
+            # the platform-agnostic main().
+            # Qt6 is still supposed to have that lib under the new name QtEntryPoint. As it does not seem
+            # to be effective when building with qmake, we are rather instructing MSVC to use the platform-agnostic
+            # main() entrypoint directly:
+            QMAKE_LFLAGS += /subsystem:windows /ENTRY:mainCRTStartup
+        }
```
hoffie, line 98: @ann0see 
> Could you please add a comment to this what this does Linker Flags? Maybe move it elsewhere down if possible

Github makes it hard to reply to your [previous thread](https://github.com/jamulussoftware/jamulus/pull/2299#discussion_r849874437) as the code has been changed/moved, therefore replying here:

I've done further tests and research regarding this. The above comment basically explains all the reasoning. I also agree with you regarding placement: main.cpp might be a logical place as the build flags are about the main function, but it's rather surprising to find linker flag adjustments in a .cpp file. Therefore, I have moved this to Jamulus.pro where other linker flags are set.

src/main.cpp
hoffie: Marking resolved in favor of this thread: https://github.com/jamulussoftware/jamulus/pull/2299#discussion_r851670516

softins (approved):
I've read through the diffs and checked the CI, so happy to approve.

I haven't been able to do any tests myself though.

ann0see:
Hmm. I'd be more happy if every platform is checked again before this is merged.

hoffie:
The number of platform-specific changes is low, the most invasive one (Mac badge handling) has been tested. I'd say that the general nightly/beta phase is sufficient.
This PR has technically two approvals so would be fine to merge, but as @pljones had lots of valuable input here I'd prefer if he can take a final look as well before merging.

ann0see:
@pljones I'm going to do some final tests here (Linux + Windows) and then merge this PR. You can of course do another after-merge review.

ann0see:
Windows ASIO didn't show anything going obviously wrong. I just noticed - again - that the meter style is now round LEDs by default, I think.

ann0see:
Ok. Did the merge now since no real end user visible changes came up.

ann0see:
Compiled via Qt6.3 on Windows and the images look a bit lo-res.

jujudusud:
Do you want me to build it on ArchLinux with Qt 6.3?

ann0see:
Not yet – unless you can verify that Qt6 doesn't introduce visible changes/bugs.

ann0see:
Not yet – unless you can verify that Qt6 doesn't introduce visible changes/bugs.

PR #2300, opened 2022-01-27
Autobuild: Use Qt6 for Windows 64bit

hoffie:

This PR switches the Autobuilds (and therefore: the releases!) to use Qt 6 for Windows 64bit builds.
The PR depends on the already merged PRs #2299 (initial Qt6 code compatibility) and #2328 (build logic Qt6 compatibility)
It only contains three small additional commits:
- Autobuild: Windows: Build with Qt 6.~3.1~5.0 on 64bit
  32bit builds remain on 5.15.2, as Qt6 is not available there.

Code-wise this is the smaller change. Change-wise it is the more relevant, more invasive one.

**Context: Fixes an issue?**

See #2299

CHANGELOG: Build: Windows 64bit releases use Qt 6.5.0 now.

Ready for testing

- [x] Wait for merge of #2299 (Build would fail otherwise)
- [x] Wait for merge of #2328 (Merge will conflict otherwise)
- [x] Testing of all officially supported and affected release artifacts on relevant platforms (via beta/rc)
  - [ ] Windows 10
  - [x] Windows 11
  - [x] Mac
- [x] Check if Qt6 drops support for any older platform versions (docs: [Qt5](https://doc.qt.io/qt-5/supported-platforms.html), [Qt6](https://doc.qt.io/qt-6/supported-platforms.html))
  - [x] Qt6 no longer officially supports Windows 7 and 8.1 (see comments)
  - [x] Qt6 no longer supports macOS versions on or before 10.13. I guess that's OK because we still have the legacy build for those users? (@softins: "So if we build with Qt6, High Sierra users on 10.13 would need to start using the legacy build.")

I'm hoping to merge this for 3.9.0, but if any critical issues are found or if there's too little testing, we should postpone it.

## Checklist

### DIFF
--- a/.github/autobuild/windows.ps1
+++ b/.github/autobuild/windows.ps1
@@ -20,7 +20,7 @@ $DownloadCacheDir = 'C:\AutobuildCache'
 # The following version pinnings are semi-automatically checked for
 # updates. Verify .github/workflows/bump-dependencies.yaml when changing those manually:
 $Qt32Version = "5.15.2"
-$Qt64Version = "5.15.2"
+$Qt64Version = "6.4.3"
 $AqtinstallVersion = "3.1.5"
 $JackVersion = "1.9.22"
 $Msvc32Version = "win32_msvc2019"


pljones:
Again, please can the _build_ and _code_ changes be split.

hoffie:
The PR only contains two more commits compared to #2999. It's based on #2999 (that's why I noted it above and marked it as Draft).

Those two commits are have trivial changes and only apply to the build logic, so I'm not sure what to split.
Linking them here, but links might change after rebasing:
[Autobuild: Windows: Build with Qt 6.2.2 on 64bit ](https://github.com/jamulussoftware/jamulus/pull/2300/commits/f06a7d3fcb0e5dd0316e01da99939da9091a4dfd)
[Autobuild: Mac: Build with Qt 6.2.2 + Xcode 12.4](https://github.com/jamulussoftware/jamulus/pull/2300/commits/1411e169bf42893c1670c78e98d2e19533b6a54f)

pljones:
Can you raise the PR against the appropriate point, rather than master, if it's not the full set of commits?

hoffie:
> Can you raise the PR against the appropriate point, rather than master, if it's not the full set of commits?

Yeah, this would be nice, but do you know how?
After some research, [there doesn't seem to be first-class support for this](https://github.com/github/feedback/discussions/4477#discussioncomment-1039627). What I could do is:
- Push #2299 to a new pr2299 branch in jamulussoftware/jamulus. I'm a bit hesistant to do that though...
- Then I could base this PR (#2300) on pr2299. The "virtual" refs such as pull/ don't seem to work in the selector.

pljones:
Mmm, no, in that case it's best not to raise the PR until master is ready to receive it, I think.  Otherwise it's not really easy to read.  There's little worth having a draft PR if it can't enable pre-review, either.

hoffie:
I've rebased to master after merge of #2328. The two relevant commits apply cleanly. Therefore, this PR now looks and feels like it should (commit list, diff). Caveat: It no longer contains #2299, therefore, build is expected to fail until #2299 is merged and #2300 is rebased.

A tree with both #2299 and #2300 can be found at [hoffie/jamulus@Qt6-pr2299-and-pr2300](https://github.com/hoffie/jamulus/actions/runs/1785992773). Artifacts should be available [here](https://github.com/hoffie/jamulus/actions/runs/1785992773) shortly.

hoffie:
I've rebased this PR on current master. As #2299 has been merged, the build of this PR now succeeds as well.
I've also updated it to use the latest Qt6 (6.3.0).
I've also hidden all the outdated comments regarding PR branch points.
I've also updated it to switch iOS as well.

ann0see:
Good to know. I just wanted to add that Qt6.3.0 worked fine on my rehearsal yesterday without UI glitches.

ann0see:
Installing the signed ipa on my phone works, but the app crashes. Hopefully it's not related to the Qt 6 move

hoffie:
Any chance to cross-check a recent Qt5-based build?
Also, some kind of debug log would be helpful of course :)

ann0see:
Latest master doesn't crash. I didn't find the crash logs for the version of this PR yet

ann0see:
I think we could move the macOS non-legacy build to use Qt 6.3 only at first. I think that's the version which needs it the most (fixed UI glitches, ARM support).

ann0see:
@hoffie can we make this macOS only for now?

pljones:
I'll just point out again that moving beyond Qt 5.2 will drop me as a developer.

ann0see:
@pljones should we build macOS only on Qt6 (this allows us to have the legacy version as fall back) since that’s the platform which shows the most glitches on old versions of Qt)

pljones:
I know nothing about Qt 6 or MacOS, so I can't comment.

ann0see:
Ok. But would you be ok with a macOS Qt6 build. This will NOT drop compatibility to Qt5 but increase compatibility with recent macOS versions. If you and @hoffie are ok with that, I suggest closing this PR and making The necessary changes for macOS non legacy

pljones:
If it's just the Github builder that's affect, I've no problem with the change (as it doesn't affect anything I do).

I'm not sure where, e.g. debian, for example, stands on Qt6.  So we might want to take care around what gets built with Qt6.

ann0see:
Yes. Just macOS on GitHub. I'll raise an alternative PR then

pljones:
> Yes. Just macOS on GitHub. I'll raise an alternative PR then

#2672 is the MacOS-only version, I think.  Should this PR be closed in favour of that?

ann0see:
I think the only blocker is that iOS is crashing. Windows worked fine last time I tried it.

pljones:
The next question, then, is is this a blocker for 3.9.0 - or can it wait for 3.9.1?

ann0see:
iOS can't go in. Windows yes

.github/workflows/autobuild.yml
```diff
@@ -155,13 +155,13 @@ jobs:
 
           - config_name:            iOS (artifacts)
             target_os:              ios
-            building_on_os:         macos-10.15
-            base_command:           QT_VERSION=5.15.2 ./.github/autobuild/ios.sh
+            building_on_os:         macos-12
+            base_command:           QT_VERSION=6.3.1 ./.github/autobuild/ios.sh
```
ann0see, line 159: iOS needs to stay on the old Qt5 version due to the crashes.

hoffie:
> This now has conflicts. Are we still hoping to get it ready for 3.9.0? (It doesn't touch anything the translators need, so there's time.)

Rebased now. I've dropped the Mac part (as it was already merged as part of the other PR). Only Windows and iOS remain.
I've also updated the PR to use 6.3.1.

I certainly would like to get it in.

@ann0see: It's been some time... can you add some details about the iOS issue? I've only found a single reference (https://github.com/jamulussoftware/jamulus/issues/2363#issuecomment-1110757457). Maybe you can also retry with fresh artifacts from this PR as Qt, Xcode and macOS have been updated meanwhile. If there are still issues, I'm happy to drop iOS from this PR for now.

With regards to Windows, the open questions/blockers are:
- Are we fine with dropping support for Windows 7 and Windows 8? Dropping support does not imply that Jamulus won't work, it just means that there *might* be issues. I think dropping Windows 7 support is fine as it is no longer supported by Microsoft. The latest Windows 8.1 patch levels still seem to be supported until 2023. I guess we could do it and if there are issues, we could provide a legacy build (still hoping that we don't have to...)?
- Can someone test on Windows 10 and/or Windows 11?

pljones:
> * Are we fine with dropping support for Windows 7 and Windows 8? Dropping support does not imply that Jamulus won't work, it just means that there _might_ be issues. I think dropping Windows 7 support is fine as it is no longer supported by Microsoft. The latest Windows 8.1 patch levels still seem to be supported until 2023. I guess we could do it and if there are issues, we could provide a legacy build (still hoping that we don't have to...)?

I'd rather we went to having a "Legacy" build for out of support Windows operating systems, using Qt5, so that Jamulus can still be used on those platforms.  For in support platforms, I think we should retain support in the main build.  That means if we _need_ Qt6 for later Windows builds, that should be separate and labelled with the OS version that needs it (e.g. Win11).

ann0see:
Agree. I think windows 7 must still be supported somehow.

ann0see:
Finally found out (again) how to view iOS crash logs. 
You need macOS and The console App.
I think that's the error:
image

We probably need to change the application class.

pljones:
OK, that sounds like it'll need some (more) platform-specific knitting in `main.cpp`.

I see no call to `UIApplicationMain` in `main()`.  Also, for iOS, we use `QApplication*` to hold the reference rather than `QCoreApplication*`, which is used everywhere else:
```
784     QApplication* pApp       = new QApplication ( argc, argv );
```
(Should really be changed.)

ann0see:
@hoffie Please drop iOS being built with Qt6. I think this should be part of the next release. Windows can be updated

hoffie:
I agree that iOS cannot be merged right now. I'll open a separate issue for the further investigation.
I'm not sure if we have a common understanding with regards to Windows. The options I see are:

1. Continue using Qt5 for Windows. Downside: Qt5 is no longer supported. Personally, I'd like to avoid that as it pushes unsupported software to users and may put us under pressure in case of Qt5 security issues.
2. Use Qt6 for Windows. Downside: Windows 7 and Windows 8 are no longer supported. That's what this PR does (besides iOS) and what I would suggest (chances are still good that it works on Windows 7 and Windows 8; combined with the fact that both should be niche by now and are nearing end of support, I'd say that this is acceptable, but I'd understand if someone argues otherwise)
3. Provide two builds. Downside: More builds (resources), more user confusion (more download options)
  a. Qt6-first: Main build is Qt6, a legacy Qt5 build would be advertised as Win7/8 build (that would be similar to our Mac strategy; this would be my second choice).
  b. Qt5-first: Main build is Qt5, a "next-gen" Qt6 build would be advertised as Win11 build (I think that was @pljones's tendency)

pljones:
I'm torn both ways.  As a developer on Windows, I'm going to have to get Qt6 built from source and patched from source (although I've not kept my Qt5 install particularly current) if I don't want to set up an account to get the binaries (which I don't).

That's really the only reason I'd rather stick with Qt5 - it's just a pain to get to Qt6 and I've not the time to sort out a build system that's going to work reliably.  It's hassle I can do without.

But I do accept that Qt6 is the supported version and we should be using it for all supported Windows versions for the _main_ Jamulus build.

hoffie:
> it's just a pain to get to Qt6 and I've not the time to sort out a build system that's going to work reliably. It's hassle I can do without.

Understandable. But as we are keeping Qt5-compatibility for the forseeable future, nothing has to change right now, I think.

> Qt6 is the supported version and we should be using it for all supported Windows versions for the main Jamulus build.

Ok, I'll update the PR. We can still add a secondary legacy build later.

hoffie:
- Opened #2711 for the iOS TODOs.
- Updated the PR to drop the iOS part.

ann0see:
I think we don't officially support anything before Windows 10 however we should test at least win 8.1 with Qt 6.3

hoffie:
Yeah. Unless one of the developers has access to a Win8.1 machine, we should call that out in the beta/rc announcement.

ann0see:
I probably do, but it's only 32 Bit. Let's ask for someone testing it via the forum.

ann0see:
Windows 11 works as expected (or at least almost as expected). The pan Pots are a bit low res, but that might just be a glitch.

ann0see (approved):

ann0see:
Just tested a Windows 7 VM 64 bit. Qt6 does drop win 7 support with the error described here. https://www.thewindowsclub.com/createdxgifactory2-could-not-be-found-or-located maybe the nsis installer can just install 32 bit.

ann0see:
Windows 8 also shows a warning: 

image

ann0see:
If we merge this and decide to go Win 10+ only, we need to update CONTRIBUTING.md. 

Not that I'm 100% convinced doing that and dropping support...

pljones:
I have a bad feeling about this...

ann0see:
Let's postpone it for the next release. I think we should do something to support win 7 and 8.1 but still have Win 10 and 11 use Qt6. If that's a separate legacy build or the 32 Bit version can still be decided

pljones:
@hoffie can you confirm you're okay with that?  I'd rather we get this right than push it for 3.9.0 as it's not really giving anything immediate for the user.

hoffie:
I'm fine with it. It doesn't matter too much if we do it now or in one of the upcoming releases. I think it's important to have a plan and that seems to be the case.
This is something for a minor release and not for a patch level release though, so I'm adding it to 3.10.0.

pljones (dismissed):

pljones:
Should this be converted to a draft for now, then?

ann0see:
@hoffie would you be ok with the following idea:
1. If we run the installer on 64 Bit win 10 or later, we use the 64 Bit Qt6 build
2. if we run any earlier OS on 64 Bit, we give a warning and install the 32 Bit "Legacy" Qt5 version.

hoffie:
> would you be ok with the following idea:

In general, yes, but I'm not sure if it addresses all potential problems.
What about Win7/8 64bit setups? They would start failing.
If we changed those to 32bit, this would be a downgrade. In addition, it would likely change binary paths which implies problems for ASIO...

I guess we could switch the main 64bit build to Qt6, but build a 64bit Qt5 binary as well and install it based on the OS version. This would imply additional installer logic and a third Windows build run though.

pljones:
> This would imply additional installer logic and a third Windows build run though.

I'd suggest a _new_ "Win10+ 64" build (with Qt6 support) and keeping a separate "Win Legacy" build (with 32/64 Qt5 support - i.e. "as now").

We already have a similar set up for MacOS (plus the MacOS multi-silicon support), so it's not really like this is any more complicated.

ann0see:
While I agree my idea would be strictly speaking a downgrade, I think it’s easier since we don‘t have two builds and can be made almost invisible. (User friendliness in terms of non ASIO4ALL users. [I think we have a work around for 32 bit ASIO4ALL…])

ann0see:
A new "legacy" installer – in my opinion – is less automated and thus less user friendly.

pljones:
I think it's easier for users to understand:
- mass market: get the "main" Win 10+ installer
- minority market who can't / won't run Win 10+ / prefer Qt5: get the "legacy" installer (which will work perfectly well on Win10+ most likely, too, of course...)

ann0see:
It's still one download more.
Since we're doing the same for macOS, I'm ok with that too. It should still not be a main build, and I don't think we need JACK Support for Win legacy.

hoffie:
> While I agree my idea would be strictly speaking a downgrade, I think it’s easier since we don‘t have two builds and can be made almost invisible. (User friendliness in terms of non ASIO4ALL users. [I think we have a work around for 32 bit ASIO4ALL…])

I think that one only works in the other direction (32bit -> 64bit) though, so it might have to be modified to be effective here?

To be honest, I'm fairly disconnected from the Windows world. I've tried to gather some stats:
- [Windows 10 was at >70% Windows market share in 2020](https://www.computerworld.com/article/3199373/windows-by-the-numbers-windows-10-rolls-on-past-70.html)
- [Windows 7 is out of support (except for special contracts which are likely not relevant for Jamulus users)](https://en.wikipedia.org/wiki/Windows_7)
- [Windows 8 is still supported until 01/2023](https://en.wikipedia.org/wiki/Windows_8), but [Steam lists <1% market share](https://www.windowslatest.com/2022/06/05/windows-11-is-gaining-some-decent-momentum-in-desktop-market-share/) and [other sources don't list Windows 8 anymore at all](https://www.thurrott.com/windows/269495/windows-11-market-share-adduplex-june-2022)

Based on that, Windows <10 seems rather niche by now. I'd be perfectly fine with moving the main build to Qt6 (and therefore making it Windows 10+ only) and providing a legacy build for 3.10.0 (which is likely not to happen before 11/2022, rather later).

Unless there are very, very good reasons, I'd argue for dropping support for unsupported operating systems if they cause any kind of overhead on our side (which an additional build does). Unsupported operating systems should not be connected to the Internet. Therefore, some future version (3.11.x) might drop support for Windows before version 10 later in 2023. If someone really wants to use Jamulus on such unsupported versions they can still continue to do so using the archived versions.

ann0see:
Time has passed and Windows 8.1 is now out of support. I'm now no longer pressing that hard for Windows 8.1 support. I'd suggest to spin up a discussion explaining the pros and cons of dropping "easy" 64 Bit Qt5 support.

ann0see:
So the suggestion would be:

* Provide a 64 Bit only Qt6 build; advertise it as main build for Win 10 and up
* Provide a 32 Bit Qt5 build similar to the macOS legacy build. 

This would probably also reduce the Windows installer size and complexity.

ann0see:
Um drafting as I believe this should go in the next release.

ann0see:
Yes. It drops Win 8.1 and 7 support.

ann0see (commented):
@pljones I think we should also update to the latest Qt6 version here. As Windows 8.1 is no longer officially supported, I think we are ok to drop support.

pljones:
Does it just need conflicts resolved and review?

ann0see:
Ok. If we merge this, we need to add an auto update like on macOS

ann0see:
https://github.com/jamulussoftware/jamulus/actions/runs/4813082936/jobs/8569162877?pr=2300#step:10:415 Failed...

pljones (approved):
OK, my new Win 11 PC will need to have Qt6 on, then, for development.  Once I get that far with the (re-)installs...

ann0see:
I believe that the scripts need to be updated with better logic to detect the difference between QT 5 and QT 6.5. Probably the regular expressions are not exact enough like in the macOS case. I still think - after testing - we can merge this.

ann0see:
Ok. Windows 11 Still seems to work. I'm currently downloading a Win 10 VM to try it there too.

ann0see (approved):
Ok. The Windows 10 VM seems ok too. I'll squash the two commits, merge it and then edit the CONTRIBUTING file.

ann0see:
Dropping needs documentation as this will be part of the ChangeLog and is documented.

PR #2313, opened 2022-01-28
Update zh_CN app translation for 3.8.2

### LINKED ISSUE #2311: Update zh_CN app translation for 3.8.2
Hi @BLumia,
We are getting ready for the 3.8.2 release. No further changes to translatable strings are currently planned for this release.

We would be happy if you updated the Jamulus software translations for **zh_CN** by **2022-02-06**.

Please
- Update your fork from `jamulussoftware/jamulus` `master` and create a working branch
- Update translations using Qt Linguist in your fork,
- Commit and push your changes and reference this Issue,
- Open a Pull Request before 2022-02-06

Suggested commit and PR message:
```
Update zh_CN app translation for 3.8.2

Fixes #2311
```

Note: Some of the displayed "untranslated" strings will actually not be new, but will rather be shown due to technical reformatting (e.g. %1). You may be able to easily update them using the "vanished" translations list.

Further documentation can be found in [TRANSLATING.md](https://github.com/jamulussoftware/jamulus/blob/master/TRANSLATING.md).

Thanks for contributing to Jamulus!

BLumia:
Update zh_CN app translation for 3.8.2

Fixes #2311

### DIFF
--- a/src/res/translation/translation_zh_CN.ts
+++ b/src/res/translation/translation_zh_CN.ts
@@ -44,12 +44,12 @@
     <message>
         <location filename="../../util.cpp" line="425"/>
         <source>Flag icons by Mark James</source>
-        <translation type="unfinished"></translation>
+        <translation>旗帜图标作者为 Mark James</translation>
     </message>
     <message>
         <location filename="../../util.cpp" line="494"/>
         <source>For details on the contributions check out the %1</source>
-        <translation type="unfinished"></translation>
+        <translation>关于贡献者的详情请参见 %1</translation>
     </message>
     <message>
         <location filename="../../util.cpp" line="495"/>
@@ -109,7 +109,7 @@
     <message>
         <location filename="../../util.cpp" line="546"/>
         <source>About %1</source>
-        <translation type="unfinished"></translation>
+        <translation>关于 %1</translation>
     </message>
     <message>
         <source>About </source>
@@ -241,12 +241,12 @@
     <message>
         <location filename="../../audiomixerboard.cpp" line="178"/>
         <source>The fader tag identifies the connected client. The tag name, a picture of your instrument and the flag of your location can be set in the main window.</source>
-        <translation type="unfinished"></translation>
+        <translation>推子标签指示了连接到的客户端。标签名称、乐器图案和您所在位置的旗帜可在主窗口中进行设置。</translation>
     </message>
     <message>
         <location filename="../../audiomixerboard.cpp" line="187"/>
         <source>Mixer channel country/region flag</source>
-        <translation type="unfinished"></translation>
+        <translation>混音通道的国家/地区旗帜</translation>
     </message>
     <message>
         <location filename="../../audiomixerboard.cpp" line="253"/>
@@ -722,7 +722,7 @@
     <message>
         <location filename="../../clientdlg.cpp" line="207"/>
         <source>If this LED indicator turns red, you will not have much fun using the %1 software.</source>
-        <translation type="unfinished"></translation>
+        <translation>如果此 LED 指示器变红，您可能无法很愉快的使用 %1。</translation>
     </message>
     <message>
         <location filename="../../clientdlg.cpp" line="234"/>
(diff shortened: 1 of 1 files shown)


src/res/translation/translation_zh_CN.ts
```diff
@@ -3275,62 +3275,62 @@ We muted your channel and activated &apos;Mute Myself&apos;. Please solve the fe
...
     <message>
         <location filename="../../util.cpp" line="1906"/>
         <source>Foundation; either version 2 of the License, or (at your option) any later version.</source>
-        <translation type="unfinished"></translation>
+        <translation>无论您依据的是本授权的第 2 版，或（您可选的）任一日后发行的版本。</translation>
     </message>
     <message>
         <location filename="../../util.cpp" line="1908"/>
         <source>There is NO WARRANTY, to the extent permitted by law.</source>
-        <translation type="unfinished"></translation>
```
BLumia, line 3298: We once again separated once single paragraph into multiple `tr()`s which make translation become harder. Please avoid splitting one sentence into two `tr()`s :/

hoffie: Thanks for the feedback and sorry for the extra efforts.
Is it possible for you to work around this for 3.8.2?
The reason for splitting is formatting. One possible workaround for you would be to take all parts of this block, merge them (on paper), translate them and split them again in a way which makes sense for Chinese that still attempts to stay within column width limits (~70 chars). I understand that this is non-optimal form a translator point of view.

Looks like this was introduced in @8fb8ce768.
@henkdegroot Any ideas if and how we could work around this in future releases? Can you open a seperate issue for this?

cc @softins

henkdegroot: This update relates to displaying proper license details in the console window when using the --version.
I think this cannot be avoid, as otherwise the print in the console will not be nice.
I don't think this will be updated very often, so hope it is just a "pain" now and not anymore in the next release.

If there is a better option, please let me know and I am happy to implement.

I would advise to use Qt Linguist for the translation update, so you can see the lines which belong together.

BLumia: > Is it possible for you to work around this for 3.8.2?

Already in this PR, but no fixed line wrap alignment through.

> The reason for splitting is formatting.

This is a paragraph of legal text referring to the GPL, but not actual GPL license content, so IMHO the ~70 chars line wrap rule doesn't need to be applied here. If it is the legal text of GPL (which isn't), it shouldn't be translatable according to [here](https://www.gnu.org/licenses/translations.html).

> I think this cannot be avoid, as otherwise the print in the console will not be nice.

By the way even in a single `tr()` you can still use regular line breaks (`\n`) to make it a multi-line translatable string and also achieve the ~70 chars line wrap goal, and use the 2nd argument of `QCoreApplication::tr()`  to tell the translators to apply the ~70 chars width line wrap rule.

hoffie: @BLumia I've opened #2319 to track this. Thanks for your input!

hoffie (approved):
PR looks fine to me, thanks again! :)

- [x] Translator listed in the src/util.cpp, optionally add link to PR or code
- [ ] Punctuation and spacing consistent: Can't tell ;)
- [x] Signal words consistent (“ASIO”, “Buffer”)
- [x] App translations: No untranslated strings (grep unfinished -5 src/res/translation/translation_$TRANSLATION*.ts)
- [x] App translations: Only a single .ts file checked in (.qm in addition is also OK)
- [x] Installer translations: Passes tools/check-wininstaller-translations.sh

src/res/translation/translation_zh_CN.ts
```diff
@@ -2330,7 +2330,7 @@ We muted your channel and activated &apos;Mute Myself&apos;. Please solve the fe
     <message>
         <location filename="../../connectdlg.cpp" line="87"/>
         <source>If you know the IP address or URL of a server, you can connect to it using the Server name/Address field. An optional port number can be added after the IP address or URL using a colon as a separator, e.g, %1. The field will also show a list of the most recently used server addresses.</source>
```
ann0see, line 2332: Unrelated, but @pljones didn‘t we want to remove the URL part or replace it?

ann0see (approved):
Seems good.

PR #2321, opened 2022-01-29
Packaging: Debian: Update repo URL, contact and Copyright

hoffie:
**Short description of changes**

Our debian packaging metadata still referred to https://github.com/corrados/...
It also still listed Volker as point of contact.
This PR changes those references to jamulussoftware and team@jamulus.io.
It also updates tools/update-copyright-notices to check the Debian Copyright: line as well.

cc @mirabilos I think you once said that you use independent control files for the official packaging. You might want to check if updates are needed there as well (or you might have spotted it earlier than us ;)).

**Context: Fixes an issue?**

/

**Does this change need documentation? What needs to be documented and how?**

No.
CHANGELOG: SKIP
(@pljones Or would you rather add this to the Changelog?)

**Status of this Pull Request**

Ready when CI is green.

**What is missing until this pull request can be merged?**

/

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- [ ] ~My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency)~ 
- [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

PR #2325, opened 2022-01-30
debian/control: Fix License header & show Volker's copyright

hoffie:

@trebmuh pointed out in https://github.com/jamulussoftware/jamulus/pull/2321#discussion_r795157739 that we should probably keep Volker listed as Copyright holder in the `debian/control` file for the past years. I struggled what years to put in, but came up with the following:

```
Copyright:
 2021-2022 The Jamulus Development Team and contributors
 2004-2021 Volker Fischer and contributors
```
I hope that this addresses all concerns.
2004 was the begin of Jamulus according to `main.cpp`.
2021 was when [Volker left](https://github.com/jamulussoftware/jamulus/discussions/928) and the Jamulus Development Team took over.
`and contributors` hopefully covers the fact that both in previous times and nowadays lots of code is provided by individual contributors. That would hopefully cover @pljones's early contributions, as well as @corrados' recent contribution.

This PR also fixes the `License:` field, which stated `GPL-2`, while it should be `GPL-2 or later` according to the file header (e.g. `src/main.cpp`). [The proper tag for that is `GPL-2+`](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#license-syntax).

**Context: Fixes an issue?**

https://github.com/jamulussoftware/jamulus/pull/2321#discussion_r795157739

No.
CHANGELOG: SKIP

Ready.

/

## Checklist

- [ ] ~My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency)~ 

### DIFF
--- a/tools/update-copyright-notices.sh
+++ b/tools/update-copyright-notices.sh
@@ -10,4 +10,4 @@ for file in $(find android ios linux mac src windows -regex '.*\.\(cpp\|h\|mm\)'
 		sed -re 's/(\*.*Copyright.*[^-][0-9]{4})(\s*-\s*\b[0-9]{4})?\s*$/\1-'"${YEAR}"'/' -i "${file}"
 done
 
-sed -re 's/(Copyright:\s*[0-9]{4}-)[0-9]{4}/\1'"${YEAR}"'/' -i distributions/debian/copyright
+sed -re 's/^( [0-9]{4}-)[0-9]{4}( The Jamulus)/\1'"${YEAR}"'\2/' -i distributions/debian/copyright
--- a/distributions/debian/copyright
+++ b/distributions/debian/copyright
@@ -5,16 +5,16 @@ Source: https://github.com/jamulussoftware/jamulus/
 
 Files: *
 Copyright: 2004-2022 The Jamulus Development Team
-License: GPL-2
+License: GPL-2+
 
 Files: debian/*
 Copyright:
  2020 Tormod Volden <debian.tormod@gmail.com>
  2020 Olivier Humbert <trebmuh@tuxfamily.org>
  2019 Marc Landolt jr <debian@marclandolt.ch>
-License: GPL-2
+License: GPL-2+
 
-License: GPL-2
+License: GPL-2+
  This package is free software; you can redistribute it and/or modify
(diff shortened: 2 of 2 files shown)


trebmuh (approved):

trebmuh:
LGTM

tools/update-copyright-notices.sh
```diff
@@ -10,4 +10,4 @@ for file in $(find android ios linux mac src windows -regex '.*\.\(cpp\|h\|mm\)'
 		sed -re 's/(\*.*Copyright.*[^-][0-9]{4})(\s*-\s*\b[0-9]{4})?\s*$/\1-'"${YEAR}"'/' -i "${file}"
```
pljones, line 10: Here and line 6 are also brought into question, perhaps?

hoffie: Do you mean the regexp? I've adjusted the other instance because it's a different format (multi-line value in debian/control)

softins (approved):
Looks good to me

tools/update-copyright-notices.sh
pljones: I just meant does the same reasoning - splitting the Copyright line - also apply to the file headers on the source, as well as the debian copyright message.  I'm not sure most of them credit the contributors (although I've tended to add myself).

hoffie: Ah, yeah.
I've also thought about those multiple times, but didn't come to a good solution.

Yes, most contributors are missing.

- We could add everyone based on git history, but this will get out of hands. `src/utils.cpp` would show 33 author lines, from a quick glance (`git log  --format='format:%an' src/util.cpp  | sort -u | wc -`).
- Listing only Volker tends to become less valid over time.
- Dropping authors altogether and listing them somewhere else instead (`src/utils.cpp` does) would also imply to drop Volker's name, which does not sound nice to me.
- Maybe a compromise would be `Volker Fischer, The Jamulus Development Team and contributors`?

pljones: Probably worth raising as a separate issue for now.

pljones (approved):

PR #1688, opened 2021-05-15
Add Meter Style Selection Option

henkdegroot:
This PR adds the option in the My Profile tab to allow the Meter Style type to be selected by the user.

Fixes #1658

PR #2352, opened 2022-02-06
Audiomixerboard: Auto-reduce size when switching between meter styles

### LINKED ISSUE #2351: Audiomixerboard: Round LEDs -> Small LEDs fails to reduce width
> But when changing from "small LED's" to "Round LED's" the fader width is adjusted, but when changing back to "small LED's" it stays wide. Only switching to "Narrow Bar" gives the narrow fader again. 
> And "Small LED's" seem to fit into the same fader width as "Narrow Bar", as changing from "Narrow Bar" to "Small LED's" does not change the total fader width.

> But when changing the meter type the fader width should always depend on the selected type, and it should not depend on a previous selected meter type.

_Originally posted by @pgScorpio in https://github.com/jamulussoftware/jamulus/issues/1688#issuecomment-1030714460_

hoffie:

There are two different base meter styles (LED vs. bars) with multiple variants each (regular, slim, round). The base meter styles are implmeneted as widgets and are always part of a `QStackedLayout`. The selected meter style is shown via `setCurrentWidget()`. This hides the inactive widget, but the inactive widget is still there and is part of `QStackedLayout`'s size calculations which always uses the maximum size over all widgets, not just the currently visible one.

This leads to problems when the currently visible meter widget is narrower than the hidden counterpart. In that case, the `QStackedLayout` is wider than expected.

Technically, the following transitions show this issue:
- Select Bar, then Small LEDs (or start with Small LEDs), or
- Select LEDs, then Narrow Bar (or start with Narrow Bar)

However, the latter case already had a workaround (choosing Narrow Bar [silently adjusted details of the LED meter as well](https://github.com/jamulussoftware/jamulus/blob/ebb1e9f278e6ba08cd717fdd2ae3692aeeb327d3/src/levelmeter.cpp#L132-L136)).

The proper fix is using a subclass of QStackedLayout which calculates its sizeHint solely based on the currently visible widget.
This fixes the Bar->Small LEDs transition and makes the LEDs->Narrow Bar workaround redundant, which has therefore been removed.

**Context: Fixes an issue?**

Fixes #2351
Initially reported here: https://github.com/jamulussoftware/jamulus/pull/1688#issuecomment-1030714460

No ChangeLog entry as the bug was not part of a release. The PR adds itself to the existing ChangeLog entry.
CHANGELOG:SKIP

Ready for testing/review

- [ ] Confirm that it fixes the bug (@pgScorpio)
- [ ] Confirm correctness with someone familiar with the initial PR (@henkdegroot?)

## Checklist

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,8 +1,8 @@
 
 ### 3.8.2beta1dev <- NOTE: the release version number will be 3.8.2 ###
 
-- Client: Added selection option for level meter style (#1688).
-  (contributed by @henkdegroot)
+- Client: Added selection option for level meter style (#1688, #2352).
+  (contributed by @henkdegroot, @hoffie, @pgScorpio)
 
 - Client: On Windows, if no driver found while installing, the "Run Jamulus"
   option will not be checked (#2103).
--- a/src/levelmeter.cpp
+++ b/src/levelmeter.cpp
@@ -61,9 +61,9 @@ CLevelMeter::CLevelMeter ( QWidget* parent ) : QWidget ( parent ), eLevelMeterTy
     pBarMeter->setFormat ( "" );                        // suppress percent numbers
 
     // setup stacked layout for meter type switching mechanism
-    pStackedLayout = new QStackedLayout ( this );
-    pStackedLayout->addWidget ( pLEDMeter );
-    pStackedLayout->addWidget ( pBarMeter );
+    pMinStackedLayout = new CMinimumStackedLayout ( this );
+    pMinStackedLayout->addWidget ( pLEDMeter );
+    pMinStackedLayout->addWidget ( pBarMeter );
 
     // according to QScrollArea description: "When using a scroll area to display the
     // contents of a custom widget, it is important to ensure that the size hint of
@@ -103,7 +103,7 @@ void CLevelMeter::SetLevelMeterType ( const ELevelMeterType eNType )
         {
             vecpLEDs[iLEDIdx]->SetColor ( cLED::RL_BLACK );
         }
-        pStackedLayout->setCurrentIndex ( 0 );
+        pMinStackedLayout->setCurrentIndex ( 0 );
         break;
 
     case MT_SLIM_LED:
@@ -112,7 +112,7 @@ void CLevelMeter::SetLevelMeterType ( const ELevelMeterType eNType )
         {
             vecpLEDs[iLEDIdx]->SetColor ( cLED::RL_SLIM_BLACK );
         }
-        pStackedLayout->setCurrentIndex ( 0 );
+        pMinStackedLayout->setCurrentIndex ( 0 );
         break;
 
     case MT_SMALL_LED:
@@ -121,21 +121,12 @@ void CLevelMeter::SetLevelMeterType ( const ELevelMeterType eNType )
         {
             vecpLEDs[iLEDIdx]->SetColor ( cLED::RL_SMALL_BLACK );
         }
-        pStackedLayout->setCurrentIndex ( 0 );
+        pMinStackedLayout->setCurrentIndex ( 0 );
         break;
 
     case MT_BAR:
-        pStackedLayout->setCurrentIndex ( 1 );
-        break;
-
     case MT_SLIM_BAR:
-        // set all LEDs to disabled, otherwise we would not get our desired small width
-        for ( int iLEDIdx = 0; iLEDIdx < NUM_LEDS_INCL_CLIP_LED; iLEDIdx++ )
-        {
-            vecpLEDs[iLEDIdx]->SetColor ( cLED::RL_DISABLED );
-        }
-
-        pStackedLayout->setCurrentIndex ( 1 );
+        pMinStackedLayout->setCurrentIndex ( 1 );
         break;
     }
 
--- a/src/levelmeter.h
+++ b/src/levelmeter.h
@@ -29,7 +29,6 @@
 #include <QTimer>
 #include <QLayout>
 #include <QProgressBar>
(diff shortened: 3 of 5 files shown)


hoffie:
Something like this explicitly sets a border similar to what it looks like for me for the *Bar* style, but I'm not sure it would be a good idea as it probabl overrides default operating styles:
```diff
diff --git a/src/levelmeter.cpp b/src/levelmeter.cpp
index 1244daf3..4658d1b3 100644
--- a/src/levelmeter.cpp
+++ b/src/levelmeter.cpp
@@ -173,6 +173,7 @@ void CLevelMeter::SetBarMeterStyleAndClipStatus ( const ELevelMeterType eNType,
         if ( bIsClip )
         {
             pBarMeter->setStyleSheet ( "QProgressBar        { border:        2px solid red;"
+                                       "                      border-radius: 4px;"
                                        "                      margin:        1px;"
                                        "                      padding:       1px;"
                                        "                      width:         15px;"
@@ -181,7 +182,9 @@ void CLevelMeter::SetBarMeterStyleAndClipStatus ( const ELevelMeterType eNType,
         }
         else
         {
-            pBarMeter->setStyleSheet ( "QProgressBar        { margin:     1px;"
+            pBarMeter->setStyleSheet ( "QProgressBar        { border:        2px solid #ccc;"
+                                       "                      border-radius: 4px;"
+                                       "                      margin:        1px;"
                                        "                      padding:       1px;"
                                        "                      width:         15px; }"
                                        "QProgressBar::chunk { background:    green; }" );
```

Also, based on my CSS understanding, I don't know why it would be necessary at all. But maybe Qt is specific or I'm missing some detail.

henkdegroot:
I believe that the stylesheet is not reset to a default or that it remembers what the previous setting was, as if it remove all previous set values. Think it needs explicit to be set again when changing. 

> _Round LEDs_ and _Small LEDs_ should probably use the narrow meter style.

Round LEDs can be selected in any skin mode. So don't think they should always use the "narrow" style, but think in "Compact" mode it could.

henkdegroot:
Small LEDs now seem to work as originally would have like to implement it. Nice work @hoffie and thanks for fixing (tested with the build generate from this PR).

So at least we have an option to keep the smallest space for the fader when switching between narrow bar and small leds.

Other styles still changes the width, but to be honest I would not call that a bug. It is done by "design".

hoffie:
> I believe that the stylesheet is not reset to a default or that it remembers what the previous setting was, as if it remove all previous set values. Think it needs explicit to be set again when changing.

I suspected something like this, but found contrary information here: https://www.qtcentre.org/threads/38841-How-to-reset-a-stylesheet-to-default?p=178450#post178450 (just a forum thread though, so might be inaccurate).

Do you have an idea what to set `border` to (see my patch from the comment above)?

> > _Round LEDs_ and _Small LEDs_ should probably use the narrow meter style.
> 
> Round LEDs can be selected in any skin mode. So don't think they should always use the "narrow" style, but think in "Compact" mode it could.

Ok, I'll look into it.

henkdegroot:
> Do you have an idea what to set `border`

I have seen really odd behaviours sometimes it can be fixed by explicit setting the border option to no. But then I have also seen that, when doing it, is suddenly inherits some style/spacing from other item it is placed in.

pljones:
> No ChangeLog entry as the bug was not part of a release:
> CHANGELOG:SKIP

Pardon?  Do you mean it's a fix to a bug introduced within the release cycle?  That means tagging the new feature with this PR as well, then.

hoffie:
> Round LEDs can be selected in any skin mode. So don't think they should always use the "narrow" style, but think in "Compact" mode it could.

Hmm, ok. I've moved Round LEDs back to the regular code path.

I think I've found the real reason for the size changes: The stacked layout seems to take the width of the widest member. I'm currently looking for ways around that.

henkdegroot:
I remember now I that I tried to only "modify" the meterstyle when changing the appearance and not also updating the "skin" style settings. It seems almost not possible to not have a link between them to keep a consistent/view layout.

hoffie:
Short update: I still think that the QStackedLayout behavior is the issue. I'm currently experimenting with the following:
- Subclassing QStackedLayout to only consider the width of the visible widget
- Restricting the maximum size of the hidden meter
- Trying different SizeConstraints
- Removing the hidden meter from the layout while not in use

I think the following is code which only exists due to this exact bug and I don't think it is a nice solution. It 
https://github.com/jamulussoftware/jamulus/blob/master/src/levelmeter.cpp#L132-L136 (introduced in cf7a01d231a5b9a910a58732e990a211de6fa360)

hoffie:
I think I now understand the issue and have a fix which looks proper and seems to work. I've updated the PR along with its description.
@henkdegroot @pgScorpio: Can you please re-check?


@pljones 
> Do you mean it's a fix to a bug introduced within the release cycle? That means tagging the new feature with this PR as well, then.

Yep, I've added the PR to the existing ChangeLog entry.

softins (approved):
I think this is OK. On my Pi, I've compiled and compared this commit with its parent commit, and the sizing is more consistent with this PR. When changing meter modes, I only observed any changes in the overall width of a channel when in Compact mode. In other modes, the channel width was determined by the name box, and was not affected by changes to the LED/Bar.

pljones (approved):
That looks like a really nice fix!  (Not tested yet.)

pgScorpio:
> Or try the assets [here](https://github.com/jamulussoftware/jamulus/actions/runs/1805413277)

This one look OK to me!

henkdegroot:
Seem working fine to me as well. Nice work!

Issue #2353, opened 2022-02-06
Improve naming meterstyle options


henkdegroot:
**Describe the bug**

New feature in 3.8.2 beta allows the selection of different meterstyle options. Current options are:
LEDs
Bar
Narrow Bar
Round LEDs
Small LEDs

**To Reproduce**

Run jamulus and open the profile settings tab.

**Expected behavior**

Options suggested in #1688 by @pgScorpio (from testing the beta release)

Bar (Narrow)
Bar (Wide)
LED Bar (or LED's (Bar) or Bar (LED's) to be consistent in using a suffix)
LED's (Small)
LED's (Big)

or (Though I personally like the previous one better)

Narrow Bar
Wide Bar
LED Bar
Small LED's
Big LED's

**Decided to implement** 

Bar (narrow)
Bar (wide)
LEDs (stripe)
LEDs (round, small)
LEDs (round, big)

**Version of Jamulus**

3.8.2beta1dev

**Additional context**

Changing this does mean English text changes, and could potentially need local language update as well.

henkdegroot:
My preference would be this option:

Bar (Narrow)
Bar (Wide)
LEDs (Bar)
LEDs (Small)
LEDs (Big)

Like the fact the first part of the option indicates the "type" and the part in brackets the "shape". Not a big fan of the bracket though.

pgScorpio:
> Like the fact the first part of the option indicates the "type" and the part in brackets the "shape". Not a big fan of the bracket though.

Yes the idea was grouping by type and shape...
But feel free to use i.e. a hyphen instead of brackets :
Bar-Narrow
Bar-Wide
LEDs-Bar
LEDs-Small
LEDs-Big

pljones:
I don't like using "Bar" for two different things: "Bar-Narrow" (or "-Wide") means the vertical strip meter.  "LEDs-Bar" then should mean a vertical strip of LEDs..?  Uh?  No.

So it needs to use a different word.  I suggested "stripe" (or "strip").

pgScorpio:
> I don't like using "Bar" for two different things: "Bar-Narrow" (or "-Wide") means the vertical strip meter. "LEDs-Bar" then should mean a vertical strip of LEDs..? Uh? No.

Uh, Yes! 
Just google LED Bar and LED Strip, almost the same results, but still it says nothing about shape, but "LED bar" IS a usual name for this type of VU meter (Just Google LED VU), so maybe use that instead?

[...]

PR #2356, opened 2022-02-07
Improve Meterstyle Naming options

henkdegroot:

Improve Meterstyle Naming and Sort Order

**Context: Fixes an issue?**

Fixes: #2353

No

Complete

Needs review, tested on Windows

## Checklist


src/clientsettingsdlg.cpp
```diff
@@ -440,11 +440,11 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
...
-    cbxMeterStyle->addItem ( tr ( "LEDs" ) );       // MT_LED
-    cbxMeterStyle->addItem ( tr ( "Bar" ) );        // MT_BAR
-    cbxMeterStyle->addItem ( tr ( "Narrow Bar" ) ); // MT_SLIM_BAR
-    cbxMeterStyle->addItem ( tr ( "Round LEDs" ) ); // MT_SLIM_LED
-    cbxMeterStyle->addItem ( tr ( "Small LEDs" ) ); // MT_SMALL_LED
+    cbxMeterStyle->addItem ( tr ( "Bar (Narrow)" ) );  // MT_SLIM_BAR
+    cbxMeterStyle->addItem ( tr ( "Bar (Wide)" ) );    // MT_BAR
+    cbxMeterStyle->addItem ( tr ( "LEDs (Stripe)" ) ); // MT_LED
+    cbxMeterStyle->addItem ( tr ( "LEDs (Small)" ) );  // MT_SMALL_LED
+    cbxMeterStyle->addItem ( tr ( "LEDs (Big)" ) );    // MT_SLIM_LED
```
hoffie, line 447: - The distinction still sounds a bit strange to me. Example: "You you like your LEDs rather as Stripes, Small or Big?" The words in brackets do not describe the same type of thing. I've got no better idea though. The following would be more accurate, but probably to long:
  - LEDs (striped)
  - LEDs (round, small)
  - LEDs (round, big)
  This may quickly get into the bike-shedding area, so we should do the best we can but should merge nevertheless even if we think it's not perfect... :)

- Should the adjectives be lowercase? (No idea, not a native speaker)

hoffie:
- [x] Tested on Linux: All labels do what they say. The proper value is also restored upon restart.

src/clientsettingsdlg.cpp
pljones: > You you like your LEDs rather as Stripes, Small or Big?

What is "rather" doing in that?  And why make "Stripe" plural?  You're changing things and then complaining it doesn't sound consistent... ;)

pljones (approved):
Looks okay.

src/clientsettingsdlg.cpp
henkdegroot: Happy to update this, and would like to update the enum entries at the same time with similar logic.
So should we take this proposal (with lowercase) as the option to go for?

(I am not a native speaker as well)

hoffie: > What is "rather" doing in that? And why make "Stripe" plural? You're changing things and then complaining it doesn't sound consistent... ;)

I was trying to place the different options in an English sentence to see how they would sound.
What I was trying to say:
- *Stripe* is not a size. There could be small or big stripes (in theory, I'm not suggesting we add them)
- *Small* and *Big* are sizes, but don't imply any form (in contrast to *Stripe*)

henkdegroot:
> E.g. `MT_BAR_NARROW` instead of `MT_SLIM_BAR`

@hoffie, the "slim/small" LED reference is also used in the bitmap references, i.e. BitmCubeSlimLedGreen, BitmCubeSmallLedGreen. Should these be updated as well when the enum entries are updated?

src/clientsettingsdlg.cpp
henkdegroot: I don't we have to worry about the total length of the text, as the "language" box just below the setting in the GUI will be wider. Good point that Small and Big no longer indicate the shape, so adding "Round" would make sense to me.

hoffie:
> the "slim/small" LED reference is also used in the bitmap references, i.e. BitmCubeSlimLedGreen, BitmCubeSmallLedGreen. Should these be updated as well when the enum entries are updated?

I think it the naming should be as consistent as possible, yes. In general, aim for the Principle of least surprise, I'd say. :)
As far as I can see, those variables are also new (as in: they've not been part of a final release), so there should be little risk of breaking things, right?

So, yes, I'd be in favor of naming those references with the same words which are used in other places (enum, labels).

pljones:
One thing, when this is released, we'll still need to highlight that anyone who's tested a post-3.8.1 build that had the new config item will need to expect that to mean something different following this change.

hoffie (approved):
Thanks a lot for working on this!
All of the changes sound meaningful. I've re-tested the behavior and everything looks fine.

This is a huge improvement at code readability. :)

Should be squash-merged.

src/levelmeter.cpp
```diff
@@ -357,14 +357,14 @@ CLevelMeter::cLED::cLED ( QWidget* parent ) :
...
-    BitmCubeSmallLedYellow ( QString::fromUtf8 ( ":/png/LEDs/res/LEDYellowSmall.png" ) ),
-    BitmCubeSmallLedRed ( QString::fromUtf8 ( ":/png/LEDs/res/LEDRedSmall.png" ) )
+    BitmCubeRoundSmallLedBlack ( QString::fromUtf8 ( ":/png/LEDs/res/LEDBlackSmall.png" ) ),
+    BitmCubeRoundSmallLedGreen ( QString::fromUtf8 ( ":/png/LEDs/res/LEDGreenSmall.png" ) ),
+    BitmCubeRoundSmallLedYellow ( QString::fromUtf8 ( ":/png/LEDs/res/LEDYellowSmall.png" ) ),
+    BitmCubeRoundSmallLedRed ( QString::fromUtf8 ( ":/png/LEDs/res/LEDRedSmall.png" ) ),
+    BitmCubeRoundBigLedBlack ( QString::fromUtf8 ( ":/png/LEDs/res/CLEDBlackSmall.png" ) ),
+    BitmCubeRoundBigLedGreen ( QString::fromUtf8 ( ":/png/LEDs/res/CLEDGreenSmall.png" ) ),
+    BitmCubeRoundBigLedYellow ( QString::fromUtf8 ( ":/png/LEDs/res/CLEDYellowSmall.png" ) ),
+    BitmCubeRoundBigLedRed ( QString::fromUtf8 ( ":/png/LEDs/res/CLEDRedSmall.png" ) )
```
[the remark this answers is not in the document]
henkdegroot: > to see "Big" being mapped to "Small",

Good catch. The Small icons already existed....and is one of the reason why I used the "odd" names in the beginning. Though when started, it was still associated to a skin choice as well and later changed to changing the meter style as a separate option.

pljones (approved):

PR #2345, opened 2022-02-05
Autobuild: Refactor and add dependency pinning

hoffie:
**Short description of changes**

This PR refactors the autobuild logic and introduces version pinning of external dependencies:
Pinning has partly been implemented already (Qt, NSIS, ASIOSDK, Android Command Line Tools, Android NDK, Android SDK). In addition, this PR introduces version pinning for JACK and aqtinstaller.

**Context: Fixes an issue?**

This is the basis for #2284 (for 3.9.0). The goal is to merge this for 3.8.2.

**Does this change need documentation? What needs to be documented and how?**

- [x] Yes, we should have a reminder for updating dependencies. See #2346

CHANGELOG: Internal: Release build process has been refactored and uses pinned dependencies for better reproducibility now.

**Status of this Pull Request**

Ready.

**What is missing until this pull request can be merged?**

- [x] "Review" (Code has been already approved as part of #2284; Code is unchanged)
- [x] Approvals to merge for 3.8.2?

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

PR #2336, opened 2022-02-03
Add missing xcode_version to CodeQL build for Mac

softins:
**Short description of changes**

Ensure CodeQL for Mac builds with the same xcode as the Artifacts.

**Context: Fixes an issue?**

For consistency, and avoids Qt warnings in the build output.

**Does this change need documentation? What needs to be documented and how?**

No

**Status of this Pull Request**

Ready

**What is missing until this pull request can be merged?**

Check the CI builds for Mac (CodeQL)

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [ ] I tested my code and it does what I want
~- [ ] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) ~ N/A
- [ ] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [ ] I've filled all the content above

PR #2283, opened 2022-01-25
Move documentation files to docs folder

ann0see:
**Short description of changes**
Moves/deletes some files to a newly created folder. Maybe we need to overthink the documentation structure. Having too many  files in the root folder is confusing.

**Context: Fixes an issue?**
Not yet.

**Does this change need documentation? What needs to be documented and how?**

**Status of this Pull Request**

Proof of concept (not to be merged soon)

**What is missing until this pull request can be merged?**
Discussion

## Checklist

Unrelated

PR #2370, opened 2022-02-09
Add missing ChangeLog entries

softins:
CHANGELOG: SKIP

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -50,6 +50,9 @@
 - Client: Accelerator key clash between Settings button and Settings menu corrected (#2248).
   (contributed by @henkdegroot)
 
+- Client: Added option code 'o' in --ctrlmidich for "Mute Myself" (#2334).
+  (contributed by @henkdegroot)
+
 - Client/Server: Output from --help and --version is now sent to StdOut instead of StdErr (#2244).
   (contributed by @henkdegroot)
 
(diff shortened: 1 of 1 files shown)


hoffie (approved):
I think we should also add the following:
- #2356 should be added to the existing levelmeter entry (first entry)
- #2283 contains a `CHANGELOG` line in its PR, but the script hasn't been run yet, so it's still missing. We can run the script or take the entry from there.
- #2336 should probably be added to the existing entry for #2345
- Translation PRs are still missing. The changelog script doesn't have special logic for this yet, I might build it though.

I may have time to put that together later today if you want. We can either merge this PR as-is and add another one for the missing stuff or we can extend the scope of this one.

softins:
@hoffie I've added your first three items, but not the translation PRs.

hoffie:
I've updated (but not yet pushed) the Changelog script to handle that. The output is still incomplete as some PRs are not merged yet.

```
- GUI: Translations have been updated
  * Dutch, by @henkdegroot (#2314)
  * French, by @jujudusud (#2329, #2320)
  * German, by @ann0see (#2349)
  * Italian, by @dzpex (#2327)
  * Polish, by @SeeLook (#2331)
  * Portuguese Brazilian, by @melcon (#2358)
  * Simplified Chinese, by @BLumia (#2313)
  * Slovak, by @jose1711 (#2348)
  * Spanish, by @ignotus666 (#2326)
  * Swedish, by @softins (#2364)
```

Issue #2372, opened 2022-02-09
Include Jamulus for Android in Fdroid (as listening client)


ann0see:
In order to move the Android support forward, including Jamulus in an App Store was requested: https://github.com/jamulussoftware/jamulus/discussions/1971
I've opened a inclusion request on Fdroid, an alternative App Store for FLOSS software on Android:
https://gitlab.com/fdroid/rfp/-/issues/2011#note_837311154

**Inclusion in Fdroid?**
According to some comments, Jamulus for Android is usable via 4G/5G even for joining Jams. However, even having an Android app as listening client opens up multiple new opportunities: For example live-streaming would be made easier; also some people - like me - like listening while traveling. I'd prefer Fdroid to the PlayStore for now, since I assume more experienced users use Fdroid and inclusion in the PlayStore needs someone to pay a specific amount of money.

**Inclusion of disclaimer**
I would totally suggest to add a disclaimer to the app as mentioned in the GitLab comment: Something like "Using a cellular or ethernet connection is crucial for using Jamulus. Do not use WiFi and please be aware that audio quality is different from device to device".

If you are totally against inclusion/adding a disclaimer, please state why. The Android version is not yet 100% stable, but I think good enough to be provided as easy to get listen only client.

Next actions:

* I would like to start working on the following: Build a metadata structure for Fdroid: https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/
* Find someone to test Jamulus from Fdroid via Cellular and a good wired Headphone.

pljones:
Ooooh.... "Latest tab criteria":
>    Name
>    Summary
>    Description
>    License
>    a What’s New entry for at least one release
>    at least one graphic (screenshot or a feature graphic)
>    **at least one of the above translated**

Does that mean we can show screen shots of Jamulus in different languges to pass the lab test?

ann0see:
I think it's the metadata. But I think one language (en-US) is enough

The bot in the post I raised says "en-US is required!"

pljones:
Well, we won't get into their "Latest" tab without the translation(s).  It only needs to be one translation (en-US and en-GB?) of the summary...

ann0see:
I mean, we can probably include one or two translations. But I don't want to put the burden on translators to translate much.

pljones:
I reckon "nice" would be:
- everything required under en-US
[...]

PR #2373, opened 2022-02-09
Add metadata for fdroid build

ann0see:
Adds metadata for Fdroid. Related to #2372 

**Context: Fixes an issue?**
Fixes: #2372 

Maybe? Not sure if we want to link the Fdroid build on the (hidden) Android installation page.

Ready for review

Review. This is just a start PR to get the first metadata into the repo to start the Fdroid build.

## Checklist

- [ ] I tested my code and it does what I want (needs to be done by Fdroid)


metadata/en-US/full_description.txt
```diff
@@ -0,0 +1,11 @@
+This is a listening client for Jamulus. Please note that you need to use a <b>cellular or ethernet connection</b> in order to use this app. WiFi will lead to massively degraded audio quality. Also note that not all devices provide low enough audio latency for Jamming, which is why you should use this App as listening client.
```
ann0see, line 1: Is this too much pushing people back?

metadata/de-DE/title.txt
```diff
@@ -0,0 +1 @@
+Jamulus (Zuhörer Client)
```
ann0see, line 1: I don't really like this translation

hoffie: Don't have a good suggestion either.
- Wiedergabe-Client
- Player
- Passiver Client

metadata/en-US/full_description.txt
pljones: tl;dr version: "Jamming: YMMV".

Yes, I think it's pushing too much.  I'd say maybe that optimisation on Android has not yet been addressed and that experiences over various networking approaches may vary.  (Is WiFi bad for just listening?  Really?)

I've not actually attempted _use_ at all with the Android version, only verified an early version could connect and "appear to be working".

metadata/de-DE/title.txt
pljones: Deos it need an adjective?  If we're not supporting installation as server on Android, I guess that does need stating, though.

pljones: (I mean "Jamulus (Client)" would be good enough, right?  Looks like "Kunde" wouldn't be the right term.)

metadata/en-US/full_description.txt
pljones: I also notice Google had already found these translations when I went to translate the German to English and gave me this exact text..!

metadata/de-DE/title.txt
ann0see:
```suggestion
Jamulus (Wiedergabe-Client)
```

I think that fits best.

metadata/en-US/full_description.txt
ann0see: Interesting. I translated English to German by hand ;-). 

Ok. So what would you suggest?

```suggestion
This is a listening client for Jamulus. Please note that you should to use a <b>cellular or ethernet connection</b>, since WiFi will lead to degraded audio quality. Also note that latency optimization on Android has not been addressed and experiences will vary from device to device. Therefore this should be seen as listening client.
```

softins: > Maybe
> 
> This is a listening client for Jamulus. Because latency optimization on Android has not been addressed, experiences will vary from device to device, so trying to participate in jams isn't recommended. For best performance, listen only over an ethernet or (4G?)/5G cellular connection (as WiFi, even for listening, isn't suitable).
> 
> * Is 4G suitable?
> * Is WiFi _not_ suitable just for listening?

I would have thought with auto-set jitter buffer, that both a good 4G signal and Wifi are fine for just listening. I have listened over Wifi with no problems, and have also participated in Zoom calls with the remote person on 4G in good quality, so would expect adequate performance for Jamulus, if latency is not an issue.

pljones: [802.11ac](https://en.wikipedia.org/wiki/Wi-Fi#Versions_and_generations) or ax WiFi, I guess, rather than earlier.

ann0see: On iOS 4G is perfectly suitable, On Android I think it’s a bit too much latency for untrained musicians if they want to play along (I didn’t try on a modern device).

pljones: OK, based on the above, I'd make it _very_ undirective.

Welcome to Jamulus. Latency optimization on Android has not been addressed and experiences will vary from device to device, so trying to participate in jams should only be attempted when you're sure you have a good connection.  The usual recommendations apply - a stable connection is essential (ethernet, 4G/5G cellular, preferably not WiFi).  Try listening only first and see if the latency is suitable before trying to join in and jam.

ann0see: What about 

```suggestion
Welcome to Jamulus. please note, latency optimization on Android has not been addressed and experiences will vary strongly from device to device. Trying to participate in jams should only be attempted when you're sure you have a good connection. The usual recommendations apply - a stable connection (4G/5G cellular; Ethernet. Stable, good quality WiFi signal might work, but is discouraged) and wired, non-Bluetooth headphones are essential. Try listening only first and see if the latency is suitable before trying to join in and jam.
```

hoffie:
Are we allowed to put other files into `metadata/`? If yes, I'd suggest placing a one-line `README` there which explains what this directory does.
`This directory is used by the Android F-Droid builds at https://....` or something.

metadata/en-US/full_description.txt
ann0see:
```suggestion
Welcome to Jamulus. Please note, latency optimization on Android has not been addressed and experiences will vary strongly from device to device. Trying to participate in jams should only be attempted when you're sure you have a good connection. The usual recommendations apply - a stable connection (4G/5G cellular; Ethernet; WiFi discouraged) and wired, non-Bluetooth headphones are essential. Try listening only first and see if the latency is suitable before trying to join in and jam.
```

pljones: Looks good to me.

pljones:
> I tested my code and it does what I want (needs to be done by Fdroid)

Do they have the facility to validatity check proposed entries before acceptance?  If so, it'd be good to do that before going ahead with this.

ann0see:
After some more research, Fdroid seems to require us building Qt from source. That's certainly not what I would like to do since it will make the android build even slower probably _ and in my opinion is not worth it. Not sure if it's worth integrating it in the official Fdroid repo. I'd maybe wait for someone else picking up the Qt build process or include the APK in a self-hosted repo (similar to a proposed but not yet worked on debian repo).

ann0see:
Uploading the screenshots here as GitHub assets:
Screenshot_20220218-212920_Jamulus
Screenshot_20220218-213009_Jamulus
Screenshot_20220218-213031_Jamulus

pljones:
We need better font handling on Android...

The issue with Qt is the Qt license... it's not _entirely_ GPL...

PR #2257, opened 2022-01-20
tools: Add ChangeLog helper script

hoffie:
**Short description of changes**

This PR adds release tooling to validate the ChangeLog for completeness.
- find-missing-entries: Will find missing PRs/Issues by scanning the Github milestone and the `git log`.
- add-missing-entries: Will edit the ChangeLog and add those found entries.
- group-entries: Will group newly added (or even existing) entries by their category while using a pre-defined category order (within the script).

**Context: Fixes an issue?**

**Does this change need documentation? What needs to be documented and how?**

- [x] The `CHANGELOG:` syntax should be suggested in the PR template. This is done as part of this PR.
- [x] Translation logic still has whitespace issues
- [ ] Existence of this tool should be documented in the release checklist: https://github.com/jamulussoftware/jamuluswebsite/pull/708

**Status of this Pull Request**

Works, but open to improvements.

**What is missing until this pull request can be merged?**

Reviews.

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- [ ] ~My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency)~  (no bash style guide)
- [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

---

> Why is #1993 listed here? It was tagged for 3.8.1, not 3.8.2, but was actually closed, not merged, and replaced with #1994. Just thought it might be something to check in your tools.

_Originally posted by @softins in https://github.com/jamulussoftware/jamulus/issues/2249#issuecomment-1017745618_

It's listed because it's part of the history in master (see a8853b6a7deb19e3596fb47bbb72a5baf29c4c7a). Might be a stray reference.
The tool uses Github PRs and their milestone to find relevant PRs. However, we might have forgotten to assign a milestone. That's why the script also scans `git log` and uses the references there as well.

Finally, let this PR be handled properly in the tool...:
CHANGELOG: SKIP

PR #2061, opened 2021-10-07
Improve accelerator key detection

softins:
**Short description of changes**

Merge base class with its derived class for dialogs (e.g. CClientDlgBase and CClientDlg)
Don't skip unfinished translations, as they still get used.

**Context: Fixes an issue?**

Just a tool used during the release process. It is just an aid to human checking, and not used for building the software.

**Does this change need documentation? What needs to be documented and how?**

No

**Status of this Pull Request**

Ready

**What is missing until this pull request can be merged?**

Ready to merge

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- n/a My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- n/a I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

PR #2380, opened 2022-02-11
ChangeLog: Add translations for 3.8.2 & #2061

hoffie:

- Adds translation PRs/authors
- Adds #2061 (@softins Should we?)
- Some punctuation consistency fixes.

Uses #2257 (the helper can now build the translation list automatically).

**Ready for merge?**
- [ ] There may still be a translation PR coming from @jujudusud. If we do rc1 before, we should probably merge this PR first. Else, we can delay the merge and add the new reference here.

CHANGELOG: SKIP

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -44,7 +44,7 @@
 - Client: On Windows, hide the ASIO button in the version that uses JACK instead (#2215).
   (contributed by @henkdegroot)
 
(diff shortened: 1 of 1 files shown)


ann0see:
Merging. I think we shouldn’t wait until the translators are ready. This can be done in another PR.

PR #2393, opened 2022-02-14
Build: Generate qm files and embedded resource during build

softins:

Generate the translation qm files during build and remove them from github.

**Context: Fixes an issue?**

The qm files are binary dependants of the ts text files. Historically they have been generated manually
and stored in github. Qt versions from 5.12 include qmake rules for generating and embedding them.

This revisits previous closed PR #1303

Working with Qt >= 5.12. ~If we need to keep supporting older Qt, we will need a compatible
set of fallback rules.~ Fallback rules added.

~Support for Qt < 5.12, or a decision to stop supporting those versions.~
Checking that all artifacts build and run correctly, particularly Mac Legacy.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ Jamulus.xcodeproj
 jamulus_plugin_import.cpp
 .github_release_changelog.md
 /debian/
+src/res/qmake_qmake_qm_files.qrc
+src/res/qrc_qmake_qmake_qm_files.cpp
+src/res/translation/*.qm
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -22,7 +22,7 @@ contains(VERSION, .*dev.*) {
 
 CONFIG += qt \
     thread \
-    lrelease
+    lrelease embed_translations
 
 QT += network \
     xml \
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -707,19 +707,6 @@ DISTFILES += ChangeLog \
     src/res/io.jamulus.jamulus.png \
     src/res/io.jamulus.jamulus.svg \
     src/res/io.jamulus.jamulusserver.svg \
-    src/translation/translation_de_DE.qm \
-    src/translation/translation_fr_FR.qm \
-    src/translation/translation_ko_KR.qm \
-    src/translation/translation_pt_PT.qm \
-    src/translation/translation_pt_BR.qm \
-    src/translation/translation_es_ES.qm \
-    src/translation/translation_nb_NO.qm \
-    src/translation/translation_nl_NL.qm \
-    src/translation/translation_pl_PL.qm \
-    src/translation/translation_it_IT.qm \
-    src/translation/translation_sv_SE.qm \
-    src/translation/translation_sk_SK.qm \
-    src/translation/translation_zh_CN.qm \
     src/res/CLEDBlack.png \
     src/res/CLEDBlackSmall.png \
     src/res/CLEDDisabledSmall.png \
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -1171,7 +1171,7 @@ QMAKE_EXTRA_TARGETS += clang_format
 
 equals(QT_MAJOR_VERSION, 5) {
     lessThan(QT_MINOR_VERSION, 12) {
-        message(Using extra lrelease rules for old Qt5)
+        message(Using extra lrelease rules for old Qt5 *** IGNORE the RCC errors below ***)
 
         qtPrepareTool(QMAKE_LRELEASE, lrelease)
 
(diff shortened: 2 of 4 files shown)


softins:
Currently in Qt < 5.12, compilation still succeeds, but no translations are available in the executable.

pljones:
Is the legacy MacOS build affected?

softins:
> Is the legacy MacOS build affected?

Yes, at the moment it would build but have no translations. So if we want to continue legacy support in 3.9 there is more work to do yet.

ann0see:
@softins this needs to be rebased.

softins:
> @softins this needs to be rebased.

Yes, I still haven't worked out what to do about embedding translations for Qt < 5.12. I haven't yet got it to work, and the limitation appears to be built into `qmake`.

ann0see:
There's nothing like a system("lrelease") call I think? Could we run a bash script?

hoffie:
> There's nothing like a system("lrelease") call I think? Could we run a bash script (or extend the autobuild scripts to run lrelease before compiling)

We can run system commands from .pro files. However, `embed_translations` does more than what `lrelease` does. Integrating `lrelease` directly in `.pro` files for older Qt releases should be possible with the logic linked above.

softins:
The problem with `qmake` on Qt < 5.12 is that it complains if the `.qm` files listed in a resource file do not already exist, and will refuse to add them to the resource `.cpp` file it generates. This all happens at the time of `qmake Jamulus.pro`, so if the `.qm` files are not generated until the subsequent run of `make`, they do not get embedded.

From 5.12 onwards, `qmake` doesn't insist that the `.qm` files exist already. This change was presumably part of adding the `embed_translations` functionality, as it wouldn't work without.

I haven't had time to think of a way forward yet for old Qt.

pljones:
> The problem with `qmake` on Qt < 5.12 is that it complains if the `.qm` files listed in a resource file do not already exist, and will refuse to add them to the resource `.cpp` file it generates. This all happens at the time of `qmake Jamulus.pro`, so if the `.qm` files are not generated until the subsequent run of `make`, they do not get embedded.

How are the `.qm` files generated?  Is there a way to run, e.g.
* `make distclean || true`
* `qmake <args that ignore qm files>`
* `make <but just the .qm files please>`
* `qmake <args that require qm files>`
* `make <main run>`

softins:
Managed to get it working on Qt 5.11, finally. Waiting to see whether the autobuilds all work, particularly Mac Legacy.

I will probably squash before merging, but for now the individual commits might be of interest.

softins:
Managed to get the Windows build working, on a debug branch. Still working on the Android build. `CONFIG += lrelease embed_translations` doesn't appear to play nicely with multi-architecture builds. Unless there is another variable that needs setting.

pgScorpio:
It looks like several problems mentioned here are related to the problems I discovered in [this](https://github.com/jamulussoftware/jamulus/pull/2588) PR.

pgScorpio:
> > @softins this needs to be rebased.
> 
> Yes, I still haven't worked out what to do about embedding translations for Qt < 5.12. I haven't yet got it to work, and the limitation appears to be built into `qmake`.

This is probably because in Jamulus.pro the path to the qm files does not match the path of the generated qm files, since lrelease creates subfolders in it's destination folder for debug_and_release or multitarget builds. (See Jamulus.pro of #2588 ) 
So if qmake also generates Makefile.Xxx files the lrelease output folder will also have Xxx subfolders!

pljones:
De-tagging from 3.9.0 - we can try to get it into 3.9.1.

ann0see:
> How are the .qm files generated? Is there a way to run, e.g.

>make distclean || true
>qmake <args that ignore qm files>
>make <but just the .qm files please>
>qmake <args that require qm files>
>make <main run>

Could we do something like that in and just add the necessary commands to builds which use an older Qt version to the respective autobuild scripts? Then we‘d add documentation in COMPILING.md about what you need to do on old Qt versions

pljones:
Hi @softins,

Do you know if you'll be able to start looking at this again in the short term?  We're looking to keep 3.9.1 "light" on translation impact anyway (although it introduces a whole new translation platform for the team) but we'd need to decide if this was in or out before going into the release cycle.

Thanks.

softins:
I'm afraid I still have very little time over the next few weeks, so I think it's best to leave this as pending, and do the 3.9.1 release without it, unless someone else wants to pick it up.

pljones:
OK bumping to 3.10.0.

ann0see:
I mean, what speaks against executing the necessary commands in the mac.sh build script and documenting that Qt < 5.12 needs the files to be generated manually?

pljones:
> I mean, what speaks against executing the necessary commands in the mac.sh build script and documenting that Qt < 5.12 needs the files to be generated manually?

That means we have different build processes depending on platform.  I don't understand why lrelease can't be installed on the MacOS build environment if needed, though, as part of the build process (and cached, ideally).

ann0see:
As I think we should be slowly deprecating macOS legacy, I think we could live with an English only legacy build. Would anyone reject that? It's not ideal, but still "ok"

ann0see:
Ok. I've now rebased this PR.

ann0see:
Closing as stale. If there's further development, please re-open

softins:
After two years and a lot of experimentation and head scratching, I have finally solved this for all builds. I will open a fresh PR for it.

PR #2398, opened 2022-02-16
Android: Set proper versionCode and versionName

### LINKED ISSUE #1760: [android] update filas
**Describe the bug**
trying to update the 380beta to the rc fails on android 9

**To Reproduce**
install rc over beta

**Expected behavior**

**Screenshots**

**Operating system**
Android 9 on Moto g6 (XT1925-5)

**Version of Jamulus**
380beta to 380rc, not sure if beta1 or beta2

**Additional context**
installation of rc working after deinstallation of 380beta

hoffie:

- Autobuild: Allow overriding checkout's fetch-depth
- Android: Set ANDROID_VERSION_{CODE,NAME}
  - android/AndroidManifest.xml uses specific macros (%%INSERT_VERSION...). Those macros are replaced by the ANDROID_VERSION_* qmake variables, which were previously unset and therefore defaulted to "1". This made it impossible to update the APK without uninstalling first according to #1760.
  - This commit populates these variables. The user-visible ANDROID_VERSION_NAME is set similar to VERSION for other platforms. The Android-internal ANDROID_VERSION_CODE is set to the number of git commits in the current branch. This seems to be the most straightforward solution which guarantees unique, increasing integers within the scope of the master branch. Feature branches will break this logic, so developers may still be required to uninstall/reinstall.
  - In order to get the commit count, the autobuild Android logic had to be modified to fetch the repo in full depth instead of only the most-recent commit.
  - Tested via `apktool d` and by checking `apktool.yml`.
- Jamulus.pro: Move ANDROID_ABIS to conditional (The variable was set globally, but is only required for the Android target. Git history does not show a specific reason for placing it globally (a3558aa61).)
- Jamulus.pro: Use DISTFILES for AndroidManifest.xml
  - Official docs suggest using DISTFILES: https://doc.qt.io/qt-5/deployment-android.html
  - OTHER_FILES is undocumented, while DISTFILES is. https://doc.qt.io/qt-5/qmake-variable-reference.html#distfiles https://stackoverflow.com/questions/38102160/in-qt-when-should-you-use-resources-vs-distfiles-vs-other-files/38102991#38102991

**Context: Fixes an issue?**

Fixes #1760

No.
CHANGELOG: Android package version has been fixed and should make APK updates easier.

Ready.

I've targeted 3.9.0.
In theory, I think this would be safe to merge for 3.8.2 as it mostly touches Android-specific build logic, which is experimental anyway. It also touches generic autobuild logic, although I'd say that the worst-case would be unnecessary git repo fetches during autobuilds.

## Checklist

- [x] I tested my code and it does what I want (via apktool)


hoffie:
@NickHyHo Can you confirm that this PR fixes your issue? You can get the APK from the "Checks" listed above.
@ann0see Were you able to reproduce the issue with previous APKs? Can you confirm that the APK from this PR works better?

NickHyHo:
@hoffie First, a Government Health Warning: My coding days are long (30+ years!) gone, and my training/experience on GitHub is zilch, but here goes... Managed to find an APK "3.8.2rc1dev-c50f092 (1)" which I installed ok, but I have a feeling that it made itself at home as a brand new package... so I then found APK "3.8.2rc1dev-26b235b0 (4749)" hot off the press, attempted to install, got "This package is already installed ... {2x package details} ... Do you want to install the update", so I did attempt install, but got "App not installed as package conflicts with an existing package." It's now a bit late here (UK) so I won't try to untangle myself until tomorrow... Since I'm junior at GitHub, please could you point me directly to the APK you'd like me to try, I'll then clear down, install the official Jamulus build, put some settings in (so I know if they're maintained during update!), then try to install your new APK on top... Or another process if you wish!

NickHyHo:
@hoffie Doh! I used wrong device last night (2 similar phones, tired person, any excuse...). This morning I used the _correct_ device, attempted install of 3.8.2rc1dev-26b235b0 (4749) over 3.8.0, but it still ended with "App not installed". I don't know if I used the correct Release Candidate, but my erroneous work last night might suggest that there's still a problem anyway?

pljones:
> I don't know if I used the correct Release Candidate, but my erroneous work last night might suggest that there's still a problem anyway?

The assets from the latest run are here (as far as I can tell...):
https://github.com/jamulussoftware/jamulus/actions/runs/1855822646

NickHyHo:
@pljones Thanks yes, that's the one, so I think it's not quite fixed yet...

hoffie:
@NickHyHo Thanks for testing. So this PR doesn't seem to be enough to fix the issue. I still believe it's right and improves things (we were always version 1 before that PR), but there's no hurry in getting it merged. I'm leaving it open for merge after 3.8.2.
I think we should move further debugging to #1760 again.

ann0see:
Unfortunately I don’t have access to the Android device anymore.

.github/workflows/autobuild.yml
```diff
@@ -168,6 +169,7 @@ jobs:
         uses:                       actions/checkout@v2
         with:
           submodules:               true
+          fetch-depth:              ${{ matrix.config.checkout_fetch_depth || '1' }}
```
softins, line 172: Does the `checkout_fetch_depth: '0'` further up evaluate here as false or true?

Jamulus.pro
```diff
@@ -203,6 +203,10 @@ win32 {
     LIBS += -framework AVFoundation \
         -framework AudioToolbox
 } else:android {
+    ANDROID_ABIS = armeabi-v7a arm64-v8a x86 x86_64
+    ANDROID_VERSION_NAME = $$VERSION
+    ANDROID_VERSION_CODE = $$system(git log --oneline | wc -l)
+
```
softins, line 209: I couldn't see in the build log where these get used. If they are used silently, then perhaps they could be echoed too?

softins (commented):
Looks good so far, but I'd like to try an installation, and maybe even a build myself.

.github/workflows/autobuild.yml
hoffie: I haven't found explicit documentation, but the CI confirms that it does what it should: [Android uses a full checkout](https://github.com/jamulussoftware/jamulus/runs/5267232585?check_suite_focus=true#step:3:40) while the other targets (e.g. [Linux](https://github.com/jamulussoftware/jamulus/runs/5267232613?check_suite_focus=true#step:3:40)) only fetch the most recent commit:

Therefore, I assume that `'0'` as a non-empty string evaluates to `true` (probably like Python).

softins: > Therefore, I assume that `'0'` as a non-empty string evaluates to `true` (probably like Python).

And unlike Perl and PHP!

Jamulus.pro
hoffie: Yep, they're qmake-internal variables. I've added a `message()` call to output them.

NickHyHo: Just tried installing jamulus_3.8.2dev-5f108fd_android.apk over 3.8.0 ... result is as before, i.e. version number is reported as being different (see screenshot attached), but install ends with message "App not installed". I suspect this is "as expected" at the moment! BTW, if you want me to check any functionality on any new build let me know - I have several Android devices at my disposal, and can force a new build to install by uninstalling any old one, of course.
Screenshot_20220222-091013_Total Commander

NickHyHo: Sorry, first time I've posted an image - didn't realise it'd come out that big on screen!

hoffie: (I've edited your comment to use ``to reduce the size a bit.

> result is as before, i.e. version number is reported as being different (see screenshot attached)
> but install ends with message "App not installed". I suspect this is "as expected" at the moment!

Thanks for the screenshot. Yes, seeing the real version number here is definitely an improvement. And yes, it's expected that this PR does not solve all update problems on Android. While I had hoped for this to be the case, the PR doesn't claim to do it :)

> BTW, if you want me to check any functionality on any new build let me know - I have several Android devices at my disposal, and can force a new build to install by uninstalling any old one, of course.

Great, thanks! I'll likely ask again as part of #2404.

hoffie:
@ann0see @softins IMO we could merge this soon. Android is still experimental and both `apktool` and @NickHyHo's tests/screenshot show that the version handling seems to have been improved with this PR. I think the chance that this PR makes anything worse is rather low. We should have a close look at the Autobuild fetch depth change, but there are at least logs which show that it works as expected.

ann0see (approved):
I can't test it at the moment, but I think it's ok.

softins (approved):
The changes look ok, and I was able to install the apk on my phone. Settings:Apps reports the correct version number, and I can run the app. Some issues in the app itself to address, but they are not related to this PR, so can be investigated separately.

ann0see:
@pljones and another fix (prefixed with Android: now)

PR #2207, opened 2022-01-02
Switch to create-dmg

ann0see:
**Short description of changes**

Just a draft on how we could switch to create-dmg. Probably it's worth changing more in the macOS build process.
**Context: Fixes an issue?**

Fixes: #1797 
**Does this change need documentation? What needs to be documented and how?**
No

**Status of this Pull Request**

Working semi-tested version 
**What is missing until this pull request can be merged?**
Someone with a Mac should test if it does what it should. Especially the placing of files in the .dmg
And it should be squashed and merged
## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

PR #2420, opened 2022-02-21
Build: Cache brew/create-dmg install

### LINKED ISSUE #2412: Build: Cache Mac's brew install (create-dmg)
**Describe the task**

#2284 introduced caching of build dependencies.
@ann0see suggested in https://github.com/jamulussoftware/jamulus/pull/2284#issuecomment-1046296291 that we should check if we can cache brew-installed code such as create-dmg (#2207) as well.

- [x] Find brew file system location
- [x] Check cache key

**Operating system**

Mac

hoffie:

The Mac build installs create-dmg via brew. The install is already version-pinned.
This PR caches the resulting install and adds the file which does the version pinning to the cache key.

**Context: Fixes an issue?**

Fixes #2412

No
CHANGELOG: SKIP
(I've included this PR in https://github.com/jamulussoftware/jamulus/pull/2207#issuecomment-1046294191)

Ready.

Reviews.

## Checklist

### DIFF
--- a/mac/deploy_mac.sh
+++ b/mac/deploy_mac.sh
@@ -69,9 +69,8 @@ build_installer_image()
 {
     # Install create-dmg via brew. brew needs to be installed first.
     # Download and later install. This is done to make caching possible
-    local create_dmg_version="1.0.9"
-    brew extract --version="${create_dmg_version}" create-dmg homebrew/cask
-    brew install /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/Formula/create-dmg@"${create_dmg_version}".rb
+    brew_install_pinned "create-dmg" "1.0.9"
+
     # Get Jamulus version
     local app_version="$(sed -nE 's/^VERSION *= *(.*)$/\1/p' "${project_path}")"
 
@@ -91,6 +90,29 @@ build_installer_image()
       "${deploy_path}/$1-${app_version}-installer-mac.dmg" \
       "${deploy_path}/"
 
+}
+
+brew_install_pinned() {
+    local pkg="$1"
+    local version="$2"
+    local pkg_version="${pkg}@${version}"
+    local brew_bottle_dir="${HOME}/Library/Cache/jamulus-homebrew-bottles"
+    local formula="/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/Formula/${pkg_version}.rb"
+    echo "Installing ${pkg_version}"
+    mkdir -p "${brew_bottle_dir}"
+    pushd "${brew_bottle_dir}"
+    if ! find . | grep -qF "${pkg_version}--"; then
+        echo "Building fresh ${pkg_version} package"
+        brew developer on  # avoids a warning
+        brew extract --version="${version}" "${pkg}" homebrew/cask
+        brew install --build-bottle --formula "${formula}"
+        brew bottle "${formula}"
+        # In order to keep the result the same, we uninstall and re-install without --build-bottle later
+        # (--build-bottle is documented to change behavior, e.g. by not running postinst scripts).
+        brew uninstall "${pkg_version}"
+    fi
+    brew install "${pkg_version}--"*
+    popd
 }
 
 
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -183,7 +183,7 @@ jobs:
             /usr/local/opt/qt
             ~/Library/Caches/pip
             ~/Library/Cache/jamulus-homebrew-bottles
-          key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/workflows/autobuild.yml', 'autobuild/mac/artifacts/autobuild_mac_1_prepare.sh', 'autobuild/mac/codeQL/autobuild_mac_1_prepare.sh') }}-${{ matrix.config.cmd1_prebuild }}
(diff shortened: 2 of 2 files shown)


hoffie:
Doesn't work yet, probably some path missing:
https://github.com/jamulussoftware/jamulus/runs/5267841353?check_suite_focus=true#step:9:1294

hoffie:
First run:
- Cache miss: https://github.com/jamulussoftware/jamulus/runs/5273453517?check_suite_focus=true#step:4:12
- Download: https://github.com/jamulussoftware/jamulus/runs/5273453517?check_suite_focus=true#step:9:1637

Second run:
- Cache hit: https://github.com/jamulussoftware/jamulus/runs/5273624229?check_suite_focus=true#step:4:17
- No redownload: https://github.com/jamulussoftware/jamulus/runs/5273624229?check_suite_focus=true#step:9:1637

pljones:
Mac installs mean nothing to me.  Was there something specific you were thinking of?

hoffie:
> Mac installs mean nothing to me. Was there something specific you were thinking of?

Hm, if I only knew... I might have confused this PR with another one. Sorry. Will ping softins once all problems with this PR are fixed.

mac/deploy_mac.sh
```diff
@@ -70,8 +70,22 @@ build_installer_image()
...
+    local formula="/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/Formula/create-dmg@${create_dmg_version}.rb"
+    echo "Installing create-dmg@${create_dmg_version}"
+    mkdir -p "${brew_bottle_dir}"
+    pushd "${brew_bottle_dir}"
+    if ! find . | grep -qF "create-dmg@${create_dmg_version}--"; then
+        echo "Building fresh create-dmg@${create_dmg_version} package"
+        brew developer on  # avoids a warning
+        brew extract --version="${create_dmg_version}" create-dmg homebrew/cask
+        brew install --build-bottle --formula "${formula}"
+        brew bottle "${formula}"
```
ann0see, line 83: Isn't it duplicate work if we run reinstall?

Why can't we just use an if/else?

mac/deploy_mac.sh
```diff
@@ -70,8 +70,22 @@ build_installer_image()
...
+    fi
+    # use reinstall to cover both the fresh build (needs reinstall) and the cached builds (install would be sufficient):
+    brew reinstall "create-dmg@${create_dmg_version}--"*
```
ann0see, line 86: Related to comment above

mac/deploy_mac.sh
hoffie: The problem is that `--build-bottle` seems to disable postinst scripts. I'm not sure if `aqt` uses those at all, but it might at some point, so I'd rather be on the safe side here.
There's also the option of running the postinst scripts explicitly. I chose the reinstall path though as it most closely ensures that both paths are really similar.

I should probably add a comment though.

Side note: The PR still doesn't work as it should, therefore it's still a draft again. When testing with a fake build (no compile), the logic works, while in the real build it doesn't (= always a fresh build). I'm probably missing something.

hoffie: Changed that part now to make it clearer (uninstall package & add comment + regular install later).
Testing still not finalized, will continue tomorrow.

hoffie:
Finally.

First run:
- Cache miss: https://github.com/jamulussoftware/jamulus/runs/5349900281?check_suite_focus=true#step:4:10
- Fresh create-dmg download/build: https://github.com/jamulussoftware/jamulus/runs/5349900281?check_suite_focus=true#step:9:1627

Second run:
- Cache hit: https://github.com/jamulussoftware/jamulus/runs/5349974155?check_suite_focus=true#step:4:15
- Use of existing create-dmg package: https://github.com/jamulussoftware/jamulus/runs/5349974155?check_suite_focus=true#step:9:1628

pljones (commented):
I can't find the Action in hoffie:actions-cache-create-dmg for this to check but it seems clean in the PR build.

ann0see (commented):
To me it still seems a bit strange to install and then uninstall the thing. But I'll probably need to be happy with that. Another possible step would be to ask on the brew forum how they do caching.

hoffie:
> I can't find the Action in hoffie:actions-cache-create-dmg for this to check but it seems clean in the PR build.

The links I provided pointed to two different runs as part of this PR.
I don't think the branch as its own runs in my repo as the branch name doesn't start with autobuild*.

> To me it still seems a bit strange to install and then uninstall the thing. 

If you felt better with the other path (`if (no package) { download; build; install; manual postinstall } else { install }`) I can adjust that. The downside would be that the gap between cache-hit and cache-miss runs would be larger. Should I change it?

> Another possible step would be to ask on the brew forum how they do caching.

The cache part isn't that complex at all. What makes most of the lines is the version-specific download, build and packaging.

ann0see:
Hmm. If your approach is better performance wise, we should go with that (and make the comment even clearer why we didn't choose the other approach).

mac/deploy_mac.sh
```diff
@@ -93,6 +92,28 @@ build_installer_image()
...
+    mkdir -p "${brew_bottle_dir}"
+    pushd "${brew_bottle_dir}"
+    if ! find . | grep -qF "${pkg_version}--"; then
+        echo "Building fresh ${pkg_version} package"
+        brew developer on  # avoids a warning
+        brew extract --version="${version}" "${pkg}" homebrew/cask
+        brew install --build-bottle --formula "${formula}"
+        brew bottle "${formula}"
+        # uninstall again to ensure that we can install the package freshly without leftovers afterwards:
+        brew uninstall "${pkg_version}"
```
ann0see, line 111: Something like "since brew install doesn't run postinst scripts, we are uninstalling it before later on installing it again [...]"

ann0see (approved):

PR #2397, opened 2022-02-16
Build scripts: Use file name for grep directly

ann0see:
**Short description of changes**

Code Style: Do not use cat | grep due to performance reasons: https://bencane.com/2013/08/19/grepping-a-file-without-using-cat-and-grep-other-tricks/
Found by Codacy

**Context: Fixes an issue?**
No. Just code style

**Does this change need documentation? What needs to be documented and how?**

No
**Status of this Pull Request**

Waiting on CI

**What is missing until this pull request can be merged?**

Reviewing. 

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [ ] I tested my code and it does what I want
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [ ] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [ ] I've filled all the content above

PR #2421, opened 2022-02-21
Build: Fix Mac app version

hoffie:

The build logic used `grep -oP` which does not seem to be supported by macOS' `grep`. Therefore, keep using `sed` instead.

Broken in: 3a630330e113dc77c736762c08fb53cad0904ba8 (#2397)

**Context: Fixes an issue?**

```
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
```
- grep error: https://github.com/jamulussoftware/jamulus/runs/5247842643?check_suite_focus=true#step:6:1652
- missing version: https://github.com/jamulussoftware/jamulus/runs/5247842643?check_suite_focus=true#step:7:16

No.

Ready.

CI/Review.

## Checklist

### DIFF
--- a/mac/deploy_mac.sh
+++ b/mac/deploy_mac.sh
@@ -73,7 +73,7 @@ build_installer_image()
     brew extract --version="${create_dmg_version}" create-dmg homebrew/cask
     brew install /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/Formula/create-dmg@"${create_dmg_version}".rb
     # Get Jamulus version
-    local app_version="$(grep -oP 'VERSION = \K\w[^\s\\]*' Jamulus.pro)"
+    local app_version="$(sed -nE 's/^VERSION *= *(.*)$/\1/p' "${project_path}")"
 
     # Build installer image
 


hoffie:
Test successful:
- No unexpected `grep` usage text anymore
- Output file is named correctly (contains the version): https://github.com/jamulussoftware/jamulus/runs/5267742959?check_suite_focus=true#step:10:16

ann0see (approved):

mac/deploy_mac.sh
```diff
@@ -73,7 +73,7 @@ build_installer_image()
     brew extract --version="${create_dmg_version}" create-dmg homebrew/cask
     brew install /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/Formula/create-dmg@"${create_dmg_version}".rb
     # Get Jamulus version
-    local app_version="$(grep -oP 'VERSION = \K\w[^\s\\]*' Jamulus.pro)"
+    local app_version="$(sed -nE 's/^VERSION *= *(.*)$/\1/p' Jamulus.pro)"
```
softins, line 76: I think to be consistent with earlier versions and the rest of the script, this should use `${project_path}`. Line 98 below doesn't actually check the directory we are running in, as `${project_path}` is an absolute path, so the `-f` test always succeeds.
```suggestion
    local app_version="$(sed -nE 's/^VERSION *= *(.*)$/\1/p' ${project_path})"
```

hoffie: > I think to be consistent with earlier versions and the rest of the script, this should use ${project_path}

Thanks, changed. We should wait for another CI run though.

> Line 98 below doesn't actually check the directory we are running in, as ${project_path} is an absolute path, so the -f test always succeeds.

Yeah, but I'd rather not touch this as part of this PR. I plan to submit my other autobuild changes really soon.

softins: > > I think to be consistent with earlier versions and the rest of the script, this should use ${project_path}
> 
> Thanks, changed. We should wait for another CI run though.

Yes, I'd like to test install the Mac artifact with the new `create-dmg`. Unfortunately, non-release runs don't make the artifacts available until the last job has finished, afaik.

> > Line 98 below doesn't actually check the directory we are running in, as ${project_path} is an absolute path, so the -f test always succeeds.
> 
> Yeah, but I'd rather not touch this as part of this PR. I plan to submit my other autobuild changes really soon.

Sure. My point was only to illustrate why it was important to use `${project_path}` instead of a simple filename, as it doesn't actually ensure we are in the correct directory.

softins (approved):
Successfully installed the dmg from this build, so happy to approve.

ann0see:
Probably a CHANGELOG: SKIP
(just edit this message if we don't skip it)

PR #2442, opened 2022-02-28
Make server GUI default to "None" directory if not set.

### LINKED ISSUE #2441: Server GUI directory should default to "None" (not registered)
I have just tried this code for the first time (in fact it was the code for #2427 I tried, which incorporates these changes), and compared with 3.8.2, using default empty ini files to ensure all defaults were used:

```
/usr/local/bin/Jamulus -s -i /tmp/x.ini
./Jamulus -s -p 22125 -i /tmp/y.ini
```

The 3.8.2 defaulted to "Not registered", with the "make my server public" checkbox unticked, which is good.

The newer version defaulted to registering with Any Genre 1, which I think is not good. The directory drop-down should default to "None", if not set in the command line or ini file, so that a new server does not register with a directory until the user asks it to.

_Originally posted by @softins in https://github.com/jamulussoftware/jamulus/pull/2199#discussion_r815875125_

softins:

If the directory is not set in the ini file or the command line, the server should not automatically register to All Genres 1. It should default to not registering until the user actually chooses a directory.
This PR sets the default for the GUI accordingly.

CHANGELOG: Server: Set default directory to "None" when not set in ini file or command line.

**Context: Fixes an issue?**

Fixes #2441 

Probably not.

Tested and ready

## Checklist

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -840,11 +840,10 @@ directoryAddress = GetIniSetting ( IniXMLDocument, "server", "centralservaddr", 
     else
     {
         // clang-format off
-// TODO compatibility to old version < ?????
-// only the case that manual was set in old ini must be considered
-if ( GetFlagIniSet ( IniXMLDocument, "server", "defcentservaddr", bValue ) && !bValue )
+// TODO compatibility to old version < 3.4.7
+if ( GetFlagIniSet ( IniXMLDocument, "server", "defcentservaddr", bValue ) )
 {
-    directoryType = AT_CUSTOM;
+    directoryType = bValue ? AT_DEFAULT : AT_CUSTOM;
 }
 else
             // clang-format on
@@ -867,6 +866,15 @@ else
         {
             directoryType = static_cast<EDirectoryType> ( iValue );
         }
+
+        // clang-format off
+// TODO compatibility to old version < 3.9.0
+// override type to AT_NONE if servlistenabled exists and is false
+if (  GetFlagIniSet ( IniXMLDocument, "server", "servlistenabled", bValue ) && !bValue )
+{
+    directoryType = AT_NONE;
+}
+        // clang-format on
     }
 
     pServer->SetDirectoryType ( directoryType );
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -829,11 +829,10 @@ directoryAddress = GetIniSetting ( IniXMLDocument, "server", "centralservaddr", 
 
     // directory type
     // CServerListManager defaults to AT_NONE
-    // Server GUI also defaults to AT_NONE (previously AT_DEFAULT, when registration was controlled by a checkbox)
     // Because type could be AT_CUSTOM, it has to be set after the address to avoid multiple registrations
     EDirectoryType directoryType = AT_NONE;
 
-    // if command line is set, use it
+    // if a command line Directory server address is set, set the Directory Type (genre) to AT_CUSTOM so it's used
     if ( CommandLineOptions.contains ( "--centralserver" ) || CommandLineOptions.contains ( "--directoryserver" ) )
     {
         directoryType = AT_CUSTOM;
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -829,9 +829,9 @@ directoryAddress = GetIniSetting ( IniXMLDocument, "server", "centralservaddr", 
 
     // directory type
     // CServerListManager defaults to AT_NONE
-    // Server GUI defaults to AT_DEFAULT
+    // Server GUI also defaults to AT_NONE (previously AT_DEFAULT, when registration was controlled by a checkbox)
     // Because type could be AT_CUSTOM, it has to be set after the address to avoid multiple registrations
-    EDirectoryType directoryType = AT_DEFAULT;
+    EDirectoryType directoryType = AT_NONE;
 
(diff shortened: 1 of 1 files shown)


src/settings.cpp
```diff
@@ -829,9 +829,9 @@ directoryAddress = GetIniSetting ( IniXMLDocument, "server", "centralservaddr",
 
     // directory type
     // CServerListManager defaults to AT_NONE
-    // Server GUI defaults to AT_DEFAULT
+    // Server GUI also defaults to AT_NONE (previously AT_DEFAULT, when registration was controlled by a checkbox)
```
pljones, line 832: I'd delete the comment entirely now - I think it's probably more confusing (trying to decypher past history).  It also makes the comment on the next line easier to follow.  (With less distraction.)

src/settings.cpp
```diff
@@ -829,9 +829,9 @@ directoryAddress = GetIniSetting ( IniXMLDocument, "server", "centralservaddr",
...
     // Because type could be AT_CUSTOM, it has to be set after the address to avoid multiple registrations
-    EDirectoryType directoryType = AT_DEFAULT;
+    EDirectoryType directoryType = AT_NONE;
 
     // if command line is set, use it
```
pljones, line 836: > if command line is set, use it

Mm, I use too few words...

> If a command line Directory server address is set, set the Type to AT_CUSTOM so it's used

might work better...

src/settings.cpp
```diff
@@ -840,15 +840,6 @@ directoryAddress = GetIniSetting ( IniXMLDocument, "server", "centralservaddr",
     }
     else
     {
-        // clang-format off
```
pljones, line 843: Just removing this isn't enough to preserve behaviour.

If `servlistenabled` _exists_ and is _true_ but there's no value for `defcentservaddr`, we should set `directoryType` to `AT_DEFAULT`, I think.  (That way it's like having not changed `AT_DEFAULT` to `AT_NONE` on line 834.)

softins: Is there a reason to preserve such behaviour so explicitly? Would the situation you described ever occur more than once, or would the ini file written on exit no longer give that situation for subsequent runs? If so, I doubt the extra logic is worth it.

I suspect the situation described would only ever occur when the user has updated from an earlier version, the first time they run the new version. They are likely to be looking at their settings anyway, and to set them as desired.

src/settings.cpp
softins: OK, will do.

src/settings.cpp
softins: Sure, I'll add that.

src/settings.cpp
pljones: All of the "backwards compatibility" in settings.cpp only happens once.

softins: > If ( `servlistenabled` _exists_ and is _true_ ) _AND_ ( there's no value for `defcentservaddr` _OR_ `defcentservaddr` is set and true ), we should set `directoryType` to `AT_DEFAULT`, I think. (That way it's like having not changed `AT_DEFAULT` to `AT_NONE` on line 834.)

Hmm, I don't think we should do that if `defcentservaddr` is missing, only if it's set and true. `defcentservaddr` will not be in any ini file written by 3.4.7 or later.

Happy to add the logic for `serverlistenabled == true && defcentservaddr == true` to set `AT_DEFAULT`. It would only be triggered for someone upgrading from 3.4.6 or earlier directly to 3.9.0 (if such a rare person were to exist).

pljones: Indeed rare.  Server settings are only loaded when there's a UI and there are precious few Server UI cases from back then that are listed (zero, I'd predict).  Perhaps we should announce that a two step upgrade is recommended from cases older than 3.8.2 - upgrade to 3.8.2 and then the next release, and drop compatibility with older cases.  It would clean up settings.cpp a lot.

(Of course, we can't do that for the protocol stuff so easily.)

softins: > Indeed rare. Server settings are only loaded when there's a UI and there are precious few Server UI cases from back then that are listed (zero, I'd predict). Perhaps we should announce that a two step upgrade is recommended from cases older than 3.8.2 - upgrade to 3.8.2 and then the next release, and drop compatibility with older cases. It would clean up settings.cpp a lot.

I think that would be good. My immediate concern is whether I need any changes to this PR to get merged. Do you have more work on settings in your NBT?

softins: OK, I've added the backward compatibility for < 3.4.7 by checking for `defcentservaddr` first if it exists, and setting to `AT_DEFAULT` or `AT_CUSTOM` as appropriate, and also for < 3.9.0 by testing for `servlistenabled` exists and is false _after_ doing the other settings from the ini file (but not if using the command line). I'm pretty sure this gives proper backward compatibility for all versions.

pljones (approved):

src/settings.cpp
pljones: Yep, pretty sure that should do it.

hoffie (approved):
Looks good to me.

Tested:
- [x] This PR with empty ini: registration is disabled.
- [x] 3.8.2 with registration disabled stays disables with this PR
- [x] 3.8.2 with registration on public directory stays like this with this PR
- [x] 3.8.2 with registration on custom directory stays like this with this PR

PR #2444, opened 2022-03-01
Build: Parallelize compilation on Android, Mac (CodeQL) and Debian

hoffie:

This PR modifies the `make` calls to use a `-jN` paramter to parallelize builds.
- Mac already used this.
- Windows' nmake doesn't support `-j`, but I'll try to do something similar using `jom`.
- This PR adds
  - Android
  - Debian
  - Mac (CodeQL)

CHANGELOG: Internal: Speed up build process by using parallelization.

**Context: Fixes an issue?**

Performance

No

Ready.
- [x] Linux build times are down from [8m22s](https://github.com/jamulussoftware/jamulus/runs/5366322032?check_suite_focus=true) on master to [5m5s](https://github.com/hoffie/jamulus/runs/5383531494?check_suite_focus=true) in this build
- [x] Mac codeQL build times are down from [10m13s](https://github.com/jamulussoftware/jamulus/runs/5366322090?check_suite_focus=true) on master to [8m6s] in this build
- [ ] Android build times are down from [32m28s](https://github.com/jamulussoftware/jamulus/runs/5366321973?check_suite_focus=true) on master to [???](https://github.com/hoffie/jamulus/runs/5383531308?check_suite_focus=true) in this build

Reviews.

## Checklist

### DIFF
--- a/autobuild/android/autobuild_apk_2_build.sh
+++ b/autobuild/android/autobuild_apk_2_build.sh
@@ -20,7 +20,7 @@ echo .
 echo .
 echo .
 echo .
-/opt/android/android-ndk/prebuilt/linux-x86_64/bin/make
+/opt/android/android-ndk/prebuilt/linux-x86_64/bin/make -j "$(nproc)"
 echo .
 echo .
 echo .
--- a/autobuild/mac/codeQL/autobuild_mac_2_build.sh
+++ b/autobuild/mac/codeQL/autobuild_mac_2_build.sh
@@ -20,7 +20,7 @@ echo "Building... qmake"
 qmake
 
 echo "Building... make"
-make
+make -j "$(sysctl -n hw.ncpu)"
 
 
 echo "Done"
--- a/distributions/debian/rules
+++ b/distributions/debian/rules
@@ -11,8 +11,8 @@ override_dh_auto_configure:
 
 override_dh_auto_build:
 	cd src/res/translation && lrelease *.ts
-	cd build-gui && make
-	cd build-nox && make
+	cd build-gui && make -j "$$(nproc)"
+	cd build-nox && make -j "$$(nproc)"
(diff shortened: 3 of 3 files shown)


hoffie:
> Interesting it works. I tried it once and the CI ran out of resources.

Maybe hardware specs have changed (Mac seems to use 3 cores, other platforms 2 cores) or another `-j` value had been chosen in the past? As said, the Mac builds already used this, so it didn't break anything at least.

> But if it works now that’s great (I.e the build time gets ok-ish time wise).

The examples show an improvement, but there is lots of flakiness in build times, so it's hard to tell for sure and how much the benefit is. Anyway, some of the build scripts are also used locally and not only in autobuild, so it makes sense to add it anyway.

ann0see:
I think @pljones or I used plain -j

pljones (approved):
Build is green (but I still get confused whether it's using the "to be merged" CI components when doing CI on CI components or whether it's using the existing ones).

ann0see (approved):
I think it uses the one in the PR.

hoffie:
> I think @pljones or I used plain -j

That's what I use locally as well. I thought it had some automatism for figuring out a good `N` value, but it seems like it would just start anything it could ("similar to a fork bomb" is what I read elsewhere). So that might explain it.

> but I still get confused whether it's using the "to be merged" CI components when doing CI on CI components or whether it's using the existing ones

Pretty sure it's the to-be-merged ones. Besides .github/, Github has no way to figure out what else is "CI related". I think you might be referring to the logic that PRs (`pull_request` event) do not run in the security context of the target repo (jamulussoftware/jamulus). IOW, they do not have access to a GITHUB_SECRET with repo access. See https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

Issue #2346, opened 2022-02-05
Autobuild: Add Github Action for dependency update PRs


hoffie:
**Describe the task**

We have several external dependencies which are version-locked for reproducibility and security. We should regularly check those for updates. A Github Action scheduled job (cron-style) could do that. It could automatically submit a PR with the suggested update.

- [ ] Github Actions (`uses: `)
- [ ] Pinned dependencies in the build process (aqt, Qt, JACK), see #2345

ann0see:
Dependabot?

hoffie:
> Dependabot?

Yeah, might be worth a look wrt Github Actions usage. I would not expect it to catch the depds from our build scripts though, unless I'm missing something?

ann0see:
Maybe we can set it up to do that

hoffie:
From https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies/configuring-dependabot-security-updates#supported-repositories

> **[](https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies/configuring-dependabot-security-updates#supported-repositories)Supported repositories**
> Repository contains dependency manifest file from a package ecosystem that GitHub supports: "[Supported package ecosystems](https://docs.github.com/en/github/visualizing-repository-data-with-graphs/about-the-dependency-graph#supported-package-ecosystems)"

So it looks like:
- [x] Github Actions could work.
- [x] pip (for aqt) could work (although we would have to create and use a `pipfile.lock`

That'd mean that the following would require manual work (at least):
- [ ] choco
- [ ] Qt

ann0see:
> The workflow does not handle most of the Android pinnings (haven't looked into it so far).

I think we should list them in this issue.

I just found out that create-dmg is also missing. So macOS and Android probably need some further investigation.

ann0see:
Not sure what is still outstanding here. Tagging for looking at it in the next release.

ann0see:
--> moving to next release

pljones:
This is something an AI Agent should be able to do pretty easily and the result should be pretty easy to verify.

But I'm dropping from 4.0.0 until there's a set of PRs.

PR #2445, opened 2022-03-01
Build: Use jom to parallelize Windows builds

hoffie:

This PR installs [Qt jom](https://wiki.qt.io/Jom) via `choco` and uses it in place of `nmake`. `jom` supports parallel builds and this PR enables them.

CHANGELOG: Internal: Speed up build process by using jom with parallelization instead of nmake in autobuilds.

**Context: Fixes an issue?**

Performance

No.

- [x] Windows build times are down from [8m52s](https://github.com/jamulussoftware/jamulus/runs/5366322321?check_suite_focus=true) on master to [6m16s](https://github.com/hoffie/jamulus/runs/5383808077?check_suite_focus=true) in this build [[5m54s in the most recent run]](https://github.com/jamulussoftware/jamulus/runs/5387615640?check_suite_focus=true) (only compared JACK, but ASIO should be similar)

Note: Build times do vary so take these numbers with a grain of salt.

CI/Reviews

## Checklist

### DIFF
--- a/autobuild/windows/autobuild_windowsinstaller_1_prepare.ps1
+++ b/autobuild/windows/autobuild_windowsinstaller_1_prepare.ps1
@@ -46,6 +46,7 @@ $AqtinstallVersion = "2.0.5"
 $JackVersion = "1.9.17"
 $Msvc32Version = "win32_msvc2019"
 $Msvc64Version = "win64_msvc2019_64"
+$JomVersion = "1.1.2"
 
 if ( Test-Path -Path $QtDir )
 {
@@ -68,6 +69,9 @@ else
     Install-Qt ${Qt32Version} ${Msvc32Version} ${QtDir}
 }
 
+choco config set cacheLocation $ChocoCacheDir
+choco install --no-progress -y jom --version "${JomVersion}"
+
 
 #################################
 ###  ONLY ADD JACK IF NEEDED  ###
@@ -75,8 +79,6 @@ else
 
 if ($BuildOption -Eq "jackonwindows")
 {
-    choco config set cacheLocation $ChocoCacheDir
-
     echo "Install JACK2 64-bit..."
     # Install JACK2 64-bit
     choco install --no-progress -y jack --version "${JackVersion}"
--- a/windows/deploy_windows.ps1
+++ b/windows/deploy_windows.ps1
@@ -234,7 +234,16 @@ Function Build-App
         "-o", "$BuildPath\Makefile")
 
     Set-Location -Path $BuildPath
-    Invoke-Native-Command -Command "nmake" -Arguments ("$BuildConfig")
+    if (Get-Command "jom.exe" -ErrorAction SilentlyContinue)
+    {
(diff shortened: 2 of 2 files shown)


windows/deploy_windows.ps1
```diff
@@ -234,7 +234,15 @@ Function Build-App
...
     Set-Location -Path $BuildPath
-    Invoke-Native-Command -Command "nmake" -Arguments ("$BuildConfig")
+    if (Get-Command "jom.exe" -ErrorAction SilentlyContinue)
+    {
+        echo "Building with jom /J ${Env:NUMBER_OF_PROCESSORS}"
+        Invoke-Native-Command -Command "jom" -Arguments ("/J", "${Env:NUMBER_OF_PROCESSORS}", "$BuildConfig")
+    }
+    else
+    {
+        Invoke-Native-Command -Command "nmake" -Arguments ("$BuildConfig")
```
ann0see, line 244: I‘d give a message saying that jom couldn’t be found so the build is now done with nmake which might be slower.

hoffie:
Quick test from a branch which has the `deploy_windows.ps1` changes from this PR, but does not install `jom` in the prepare script: https://github.com/hoffie/jamulus/runs/5387646863?check_suite_focus=true#step:9:126

(This was before @ann0see's suggestion to add an explicit echo for this case, so don't search for it ;))

- [x] Build logic still works even without `jom` installed

ann0see (approved):
Ok. So approving in these terms.

ann0see:
In Linux I thought multiple files were compiled in parallel (at least locally) and I don’t know if you can judge via CI log.

hoffie:
x

> In Linux I thought multiple files were compiled in parallel (at least locally) and I don’t know if you can judge via CI log.

There are two places where "parallelization" can happen:
- By running compiler invocations in parallel (that's what `make -j` and `jom /N` do. That's not necessarily visible in build logs, yep.
- By compiling multiple source files in a single compiler invocation. This isn't parallelization, but it'll likely reduce overall compiler warm up times, `.h` parsing, etc. This should be visible in the build logs as the invocation would have to change for that. That's where `/MP` would be relevant. But if we don't expect benefits, I'm a bit hesistant to add it.

autobuild/windows/autobuild_windowsinstaller_1_prepare.ps1
```diff
@@ -46,6 +46,7 @@ $AqtinstallVersion = "2.0.5"
 $JackVersion = "1.9.17"
 $Msvc32Version = "win32_msvc2019"
 $Msvc64Version = "win64_msvc2019_64"
+$JomVersion = "1.1.2"
```
pljones, line 49: Do we have a documented process in place to review pinned versions from time to time?

(Don't I love coming up with unconnected issues...)

hoffie: > Do we have a documented process in place to review pinned versions from time to time?

Not yet, but there's at least *one* reminder: #2346

PR #2447, opened 2022-03-02
Server: Fix outdated comment about QThreadPool [trivial]

hoffie:
We don't use QThreadPool, but rather our own CThreadPool.

CHANGELOG: SKIP

**Context: Fixes an issue?**

Misleading comment.

No.

Ready.

Needs approvals which confirm that my understanding is correct that we don't use QThreadPool anymore.


softins (approved):
I can't see any reference to QThreadPool in the source.

ann0see (approved):

pljones:
Probably explains why task distribution across threads is weird...

Issue #2429, opened 2022-02-23
Releases: Rename .debs (drop _ubuntu_, dash for headless)


hoffie:
**Has this feature been discussed and generally agreed?**

Initial discussion here: https://github.com/jamulussoftware/jamulus/pull/2423#issuecomment-1048915034

**Describe the solution you'd like**

- [x] Check that there is no technical reason to name the package like that
- [ ] ~Rename jamulus_3.8.2_ubuntu_amd64.deb to jamulus_3.8.2_amd64.deb~
- [ ] ~Rename headless version accordingly~
- [ ] Replace _ by - for headless to match the package name.
- [ ] grep the source for required updates (autobuild scripts use `ubuntu`)

Reasoning: We provide the .deb to support Debian and Ubuntu. `.deb` already hints that it targets these distributions. Adding `ubuntu` is confusing.

**Describe alternatives that have been considered**

Keep as-is

ann0see:
Related to https://github.com/jamulussoftware/jamulus/issues/795

ann0see:
Actually I don’t remember why we named it "Ubuntu" probably because most users are assumed to use Ubuntu and we build on Ubuntu. I think we could rename it. This might imply changes in unofficial scripts.

ann0see:
@sthenos I think you wrote a script for WorldJam. You might need to change the install links one da

ann0see:
Ah. Ok. So basically it’s semi-officially supported. It might make sense to put it into the installscripts repo.

ann0see:
Current status:

I am not feeling a strong need of removing or not removing "ubuntu" from the file name.
Hoffie is slightly pro, softins is slightly con, so I assume the decision is a draw. 

Suggestions involved
* Renaming all files by appending the OS as a folder
* Renaming ubuntu to linux or debian
* Doing nothing

I think renaming to debian would be good although we can’t for sure guarantee Debian compatibility

hoffie:
Yep. I've crossed out the no langer planned TODO items.
jamulus_headless -> jamulus-headless would still help with consistency with the actual package name, so I've left this TODO item open.

ann0see:
So the only thing left is the rename?

pljones:
a) If the "ubuntu" is staying, can the issue title be changed, please?
b) @softins, if this went ahead, would you be able to update Simon's Script after the release?
c) Who maintains the "ubuntu" build script?

softins:
[...]

PR #2452, opened 2022-03-02
Remove Ubuntu from artifact name

ann0see:
Fixes: https://github.com/jamulussoftware/jamulus/issues/2429

Removes "ubuntu" from the artifact name.

CHANGELOG: Remove confusing _ubuntu suffix from .deb files to avoid users thinking that the artifacts are only compatible with Ubuntu linux.

**Context: Fixes an issue?**
Partly fixes: #2429 

Yes. We need to update the installation links on the website.

Draft. 

Needs resolving of #2429 first.

## Checklist

### DIFF
--- a/autobuild/linux/autobuild_deb_3_copy_files.sh
+++ b/autobuild/linux/autobuild_deb_3_copy_files.sh
@@ -44,14 +44,14 @@ echo ""
 #move/rename headless first, so wildcard pattern matches only one file each
 echo ""
 echo ""
-artifact_deploy_filename_1=jamulus_headless_${jamulus_buildversionstring}_ubuntu_amd64.deb
+artifact_deploy_filename_1=jamulus_headless_${jamulus_buildversionstring}_amd64.deb
 echo "Move/Rename the built file to deploy/${artifact_deploy_filename_1}"
 mv "${THIS_JAMULUS_PROJECT_PATH}"/../jamulus-headless*_amd64.deb "${THIS_JAMULUS_PROJECT_PATH}"/deploy/"${artifact_deploy_filename_1}"
 
 #move/rename normal second
 echo ""
 echo ""
-artifact_deploy_filename_2=jamulus_${jamulus_buildversionstring}_ubuntu_amd64.deb
+artifact_deploy_filename_2=jamulus_${jamulus_buildversionstring}_amd64.deb
 echo "Move/Rename the built file to deploy/${artifact_deploy_filename_2}"
 mv "${THIS_JAMULUS_PROJECT_PATH}"/../jamulus*_amd64.deb "${THIS_JAMULUS_PROJECT_PATH}"/deploy/"${artifact_deploy_filename_2}"
 


pljones (approved):

softins (changes requested):
Looking at the list of artifacts built, I don't think just removing `ubuntu` is right. All the other platforms have "win", "mac", "android", etc., in their names. Without the "ubuntu", it is not clear that these artifacts are even for Linux. But the problem is, if we replace "ubuntu" with just "linux", it implies that the packages are generically for any Linux, rather than just for Ubuntu or Debian. So I'm not sure what the best solution is.

softins:
Since the packages are built on Ubuntu, we're only really guaranteeing that platform is compatible. We could say "ubuntu_debian", but then we would need to verify that the packages actually install and work properly on Debian 10 and 11. I'm actually leaning towards preserving the status quo.

ann0see:
@hoffie should decide if this should be changed or not.

hoffie:
The reason why I suggested to do this was [Debian naming suggestions](http://manpages.ubuntu.com/manpages/bionic/man1/dpkg-name.1.html):

```
root@e522311634d9:/# dpkg-name jamulus_3.8.2-1_ubuntu_amd64.deb 
dpkg-name: info: moved 'jamulus_3.8.2-1_ubuntu_amd64.deb' to './jamulus_3.8.2-1_amd64.deb'
```

However, those are only suggestions, so I don't expect this to be technically relevant.

I agree that this is just one perspective on the naming. The distinction of the different platforms in the Artifact list is very relevant as well, maybe even more relevant.

One would to satisfy both needs

ann0see:
> Ubuntu/jamulus_3.8.2-1_amd64.deb

I doubt GitHub releases supports this. Moreover, I'm not sure if we need to stick that closely to the names? If we put it on a repo, maybe. But the file name could also be renamed there.

hoffie:
> > Ubuntu/jamulus_3.8.2-1_amd64.deb
> 
> I doubt GitHub releases supports this.

Just tried it, looks like you're right. `/ ` is converted to a dot (`.`). So it won't work like that.

> Moreover, I'm not sure if we need to stick that closely to the names?

No, it's a suggestion/convention.

pljones:
My main objection to using `_ubuntu` in the place it was in the name is that it can imply that the `.deb` is an _official_ Ubuntu artifact.

`Splashtop_Streamer_Ubuntu_14.04_v2.2.5.1-4_amd64.deb`

Putting it in a different place would be better -- here the "Ubuntu" isn't part of the build designator but part of the software package name.  So "Jamulus for Debian" would be "jamulus_debian_3.9.0-1_amd64.deb" (or whatever CPU).

softins:
> My main objection to using `_ubuntu` in the place it was in the name is that it can imply that the `.deb` is an _official_ Ubuntu artifact.

I wouldn't infer that, any more than the Mac and Windows version are official from Apple or Microsoft. Just a designation of the intended target.

I would prefer to put this PR on hold for now until we've been able to think/talk a bit more about it.

pljones:
The official debian release is named `jamulus_3.8.2+dfsg-1_amd64.deb`; any release as part of Ubuntu (rather than a ppa) that repackaged that would have "ubuntu" in the package name (as well as the Ubuntu release identifier) to indicate it was an official Ubuntu release.  ppa releases don't have "ubuntu", to show they're not official.

Windows, MacOS, etc don't use `.deb` files, so arguing their common naming conventions are different isn't really relevant.

ann0see:
I'd be ok with merging or closing this PR (not feeling a strong need of changing it).

ann0see:
Closing as discussion should continue in the respective issue

PR #2451, opened 2022-03-02
Refactor linux autobuild

ann0see:
**Short description of changes**
Moves some files around to be consistent with Windows/Android build
Removes unneeded files

Changelog entry might need to be split.

**Context: Fixes an issue?**
Fixes: #2430 

**Does this change need documentation? What needs to be documented and how?**

**Status of this Pull Request**
Finished
**What is missing until this pull request can be merged?**
Needs a second review
## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

PR #2460, opened 2022-03-04
Build: Include jamulus-server.desktop in Debian/Ubuntu builds and fix Linux SVG icon installation location

hoffie:
- Moves the existing .svg icons to /usr/share/icons/hicolor/scalable/apps/ (from /usr/share/icons/hicolor/512x512/apps, which sounds wrong for SVGs). This affects all Linux installations.
- Adds the existing jamulus-server.desktop file to the .deb packaging (but not to headless). This is Debian-only.

Both combined should the effect of having a working Jamulus Server desktop icon in the .deb package.

CHANGELOG: Linux: Added the Jamulus Server desktop icon to the Debian/Ubuntu build and fixed SVG icon installation location.

**Context: Fixes an issue?**

Related: https://github.com/jamulussoftware/jamulus/pull/2451#issuecomment-1058326425

No.

Ready, resulting .deb has been checked if the file is there -- it is. I can't test the UI (icon rendering) myself though.

Ready.

## Checklist

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -410,7 +410,14 @@ win32 {
         }
         ICONSDIR = $$absolute_path($$ICONSDIR, $$PREFIX)
         icons.path = $$ICONSDIR
-        icons.files = distributions/jamulus.png distributions/jamulus.svg distributions/jamulus-server.svg
+        icons.files = distributions/jamulus.png
+
+        isEmpty(ICONSDIR_SVG) {
+            ICONSDIR_SVG = share/icons/hicolor/scalable/apps/
+        }
+        ICONSDIR_SVG = $$absolute_path($$ICONSDIR_SVG, $$PREFIX)
+        icons_svg.path = $$ICONSDIR_SVG
+        icons_svg.files = distributions/jamulus.svg distributions/jamulus-server.svg
 
         isEmpty(MANDIR) {
             MANDIR = share/man/man1
@@ -419,7 +426,7 @@ win32 {
         man.path = $$MANDIR
         man.files = distributions/Jamulus.1
 
-        INSTALLS += target desktop icons man
+        INSTALLS += target desktop icons icons_svg man
     }
 }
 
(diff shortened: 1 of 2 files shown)


distributions/debian/jamulus.install
```diff
@@ -1,3 +1,4 @@
 usr/bin/jamulus
 usr/share/applications/jamulus.desktop
+usr/share/applications/jamulus-server.desktop
```
ann0see, line 3: Unfortunately it doesn't show the server icon for me. Probably the icon is not set correctly?

ann0see (changes requested):

distributions/debian/jamulus.install
hoffie: There's no .png in the tree, I'll check whether the .svgs can be used.

ann0see: Ah yes. .svg's would be even better.

Jamulus.pro
```diff
@@ -410,7 +410,14 @@ win32 {
         }
         ICONSDIR = $$absolute_path($$ICONSDIR, $$PREFIX)
         icons.path = $$ICONSDIR
-        icons.files = distributions/jamulus.png distributions/jamulus.svg distributions/jamulus-server.svg
+        icons.files = distributions/jamulus.png
+
+        isEmpty(ICONSDIR_SVG) {
+            ICONSDIR_SVG = share/icons/hicolor/scalable/apps/
+        }
```
hoffie, line 417: I've no idea why we are checking `ICONSDIR` (and others) for emptiness. Those variables do not seem magical (i.e. they are not pre-set by qmake, as far as I can see). Therefore the intention was probably to make them overridable. Anyway, I'm following this path and I'm using the same pattern (probably same reasoning what @atsampson did in 69e3286057a94574ba7b61e7f93ec812251775d2 when first adding it for `ICONSDIR`).

Fear of breakage is what prevents me from removing that. :)

distributions/debian/jamulus.install
hoffie: The .svgs were already installed by the `Makefile`, but to the wrong location (IMO). I've fixed that as well and adjusted the debian `.install` file. Can you test again once the build is done?

ann0see (approved):
Ok. Works now.

pljones (approved):
Based on @ann0see's OK.

PR #2470, opened 2022-03-07
Autobuild: Clean up workflow description

hoffie:
This is a small, split-out part of the announced autobuild refactoring.

- Autobuild: Drop commented out code and duplicate newlines
- Autobuild: Shorten branch [matching syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)
- Autobuild: Drop comments which repeat step names
- Autobuild: Unify identation
- Autobuild: Reorganize, shorten and strip redundant comments
- Autobuild: Shorten if expressions
  [`${{ ... }}` is optional for if: entries](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif). Therefore drop it in order to reduce clutter.
- Autobuild: Line-wrap long if statements
- Autobuild: Rename uses_codeql -> run_codeql

CHANGELOG: Autobuild: Refactor for better readability

**Context: Fixes an issue?**

Improve readability/maintenance.

No.

Ready when CI is green.

Ready when CI is green.

Note: I suggest reading individual commit diffs if unsure. `git show -w` and Github's whitespace-ignoring diff may also help during review.

## Checklist

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -39,17 +39,13 @@ jobs:
       version_name:                 ${{ steps.jamulus-build-vars.outputs.RELEASE_VERSION_NAME }}
 
      steps:
-         # Checkout code
          - name:                    Checkout code
            uses:                    actions/checkout@v2
 
-         # Set variables
-         # Determine release / pre-release
          - name:                    Get Jamulus build info, determine actions & variables
            run:                     python3 ${{ github.workspace }}/.github/actions_scripts/analyse_git_reference.py
            id:                      jamulus-build-vars
 
-         # remove release, if it exists (with this releasetag)
          - name:                    Remove release, if existing (for branches)
            if:                      ${{ contains(steps.jamulus-build-vars.outputs.PUBLISH_TO_RELEASE, 'true') }}
            continue-on-error:       true
@@ -59,7 +55,7 @@ jobs:
              tag_name:              ${{ steps.jamulus-build-vars.outputs.RELEASE_TAG }}
            env:
              GITHUB_TOKEN:          ${{ secrets.GITHUB_TOKEN }}
-         # create release (empty, filled by next jobs)
+
          - name:                    'Create Release ${{steps.jamulus-build-vars.outputs.RELEASE_TAG}}  {{steps.jamulus-build-vars.outputs.RELEASE_TITLE}}'
            if:                      ${{ contains(steps.jamulus-build-vars.outputs.PUBLISH_TO_RELEASE, 'true') }}
            id:                      create_release_step
@@ -158,21 +154,18 @@ jobs:
 
     runs-on:                        ${{ matrix.config.building_on_os }}
     steps:
-
       - name:                       Select Xcode version for Mac
         if:                         ${{ matrix.config.target_os == 'macos' || matrix.config.target_os == 'ios' }}
         uses:                       maxim-lobanov/setup-xcode@v1
         with:
           xcode-version: ${{ matrix.config.xcode_version }}
 
-      # Checkout code
       - name:                       Checkout code
         uses:                       actions/checkout@v2
         with:
           submodules:               true
           fetch-depth:              ${{ matrix.config.checkout_fetch_depth || '1' }}
 
-      # Enable caching of downloaded dependencies
       - name:                       "Cache Mac dependencies"
         if:                         ${{ matrix.config.target_os == 'macos' }}
         uses:                       actions/cache@v2
@@ -205,7 +198,6 @@ jobs:
             /opt/android/android-ndk
           key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/workflows/autobuild.yml', 'autobuild/android/autobuild_apk_1_prepare.sh', 'autobuild/android/install-qt.sh') }}-${{ matrix.config.cmd1_prebuild }}
 
-      # Prepare (install QT & dependencies)
       - name:                       "Prepare for ${{ matrix.config.config_name }}"
         if:                         ${{ matrix.config.cmd1_prebuild }}
         run:                        ${{ matrix.config.cmd1_prebuild }}
@@ -214,7 +206,6 @@ jobs:
           jamulus_project_path:     ${{ github.workspace }}
           jamulus_buildversionstring: ${{ needs.create_release.outputs.version_name }}
 
-      # Initialize CodeQL tools for code-scanning for security
       - name:                       Initialize CodeQL
         if:                         matrix.config.uses_codeql
         uses:                       github/codeql-action/init@v1
@@ -224,7 +215,6 @@ jobs:
           # Learn more:
           # https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#changing-the-languages-that-are-analyzed
 
-      # Build
       - name:                       "Build for ${{ matrix.config.config_name }}"
         id:                         step_macos_build
         if:                         ${{ matrix.config.cmd2_build }}
@@ -246,7 +236,6 @@ jobs:
           jamulus_project_path:       ${{ github.workspace }}
           jamulus_buildversionstring: ${{ needs.create_release.outputs.version_name }}
 
-      # Upload Artifact to Job
       - name:                       Upload Artifact 1 to Job
         if:                         ${{ steps.step_cmd3_postbuild.outputs.artifact_1 }}
         uses:                       actions/upload-artifact@v2
@@ -256,7 +245,6 @@ jobs:
           retention-days:           31
           if-no-files-found:        error # 'warn' or 'ignore' are also available, defaults to `warn`
 
-      # Upload Artifact to Job
       - name:                       Upload Artifact 2 to Job
         if:                         ${{ steps.step_cmd3_postbuild.outputs.artifact_2 }}
         uses:                       actions/upload-artifact@v2
@@ -266,7 +254,6 @@ jobs:
           retention-days:           31
           if-no-files-found:        error # 'warn' or 'ignore' are also available, defaults to `warn`
 
-      # Notarize macOS build, if needed
       - name: "Notarize macOS Release Build"
         if: ${{ (steps.step_cmd3_postbuild.outputs.artifact_1 != '') && (steps.step_macos_build.outputs.macos_signed == 'true') && contains(needs.create_release.outputs.publish_to_release, 'true') }}
         id: notarize-macOS-app
@@ -276,7 +263,7 @@ jobs:
           primary-bundle-id: io.jamulus.Jamulus
           appstore-connect-username: ${{ secrets.NOTARIZATION_USERNAME }}
           appstore-connect-password: ${{ secrets.NOTARIZATION_PASSWORD }}
-      # Apply successful notarization
+
       - name: "Staple macOS Release Build"
         if: ${{ (steps.step_cmd3_postbuild.outputs.artifact_1 != '') && (steps.step_macos_build.outputs.macos_signed == 'true') && contains(needs.create_release.outputs.publish_to_release, 'true') }}
         id: staple-macOS-app
@@ -284,7 +271,6 @@ jobs:
         with:
           product-path: deploy/${{ steps.step_cmd3_postbuild.outputs.artifact_1 }}
 
-      # Upload Artifact to Release
       - name:                       Upload Artifact 1 to Release
         if:                         ${{ (steps.step_cmd3_postbuild.outputs.artifact_1 != '') && contains(needs.create_release.outputs.publish_to_release, 'true') }}
         id:                         upload-release-asset1
@@ -297,7 +283,6 @@ jobs:
           asset_name:               ${{ steps.step_cmd3_postbuild.outputs.artifact_1 }}
           asset_content_type:       application/octet-stream
 
-      # Upload Artifact to Release
       - name:                       Upload Artifact 2 to Release
         if:                         ${{ (steps.step_cmd3_postbuild.outputs.artifact_2 != '') &&  contains(needs.create_release.outputs.publish_to_release, 'true') }}
         id:                         upload-release-asset2
@@ -310,7 +295,6 @@ jobs:
           asset_name:               ${{ steps.step_cmd3_postbuild.outputs.artifact_2 }}
           asset_content_type:       application/octet-stream
 
-      # Run CodeQL tools for code-scanning for security
       - name:                       Perform CodeQL Analysis
         if:                         matrix.config.uses_codeql
         uses:                       github/codeql-action/analyze@v1
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -85,7 +85,7 @@ jobs:
              cmd1_prebuild:         "./autobuild/android/autobuild_apk_1_prepare.sh"
              cmd2_build:            "./autobuild/android/autobuild_apk_2_build.sh"
              cmd3_postbuild:        "./autobuild/android/autobuild_apk_3_copy_files.sh"
-             uses_codeql:           true
+             run_codeql:            true
              # Jamulus.pro needs to count git history length for android versioning:
              checkout_fetch_depth:  '0'
 
@@ -95,7 +95,7 @@ jobs:
              cmd1_prebuild:         "./autobuild/linux/autobuild_deb_1_prepare.sh"
              cmd2_build:            "./autobuild/linux/autobuild_deb_2_build.sh"
              cmd3_postbuild:        "./autobuild/linux/autobuild_deb_3_copy_files.sh"
-             uses_codeql:           true
+             run_codeql:            true
 
            - config_name:           MacOS (codeQL)
              target_os:             macos
@@ -103,7 +103,7 @@ jobs:
              cmd1_prebuild:         "./autobuild/mac/codeQL/autobuild_mac_1_prepare.sh 5.15.2"
              cmd2_build:            "./autobuild/mac/codeQL/autobuild_mac_2_build.sh"
              cmd3_postbuild:        false
-             uses_codeql:           true
+             run_codeql:            true
              xcode_version:         12.1.1
 
            - config_name:           MacOS (artifacts)
@@ -113,7 +113,7 @@ jobs:
              cmd1_prebuild:         "./autobuild/mac/artifacts/autobuild_mac_1_prepare.sh 5.15.2"
              cmd2_build:            "./autobuild/mac/artifacts/autobuild_mac_2_build.sh sign_if_possible"
              cmd3_postbuild:        "./autobuild/mac/artifacts/autobuild_mac_3_copy_files.sh"
-             uses_codeql:           false
+             run_codeql:            false
              xcode_version:         12.1.1
 
            - config_name:           iOS (artifacts)
@@ -122,7 +122,7 @@ jobs:
              cmd1_prebuild:         "./autobuild/ios/artifacts/autobuild_ios_1_prepare.sh 5.15.2"
              cmd2_build:            "./autobuild/ios/artifacts/autobuild_ios_2_build.sh"
              cmd3_postbuild:        "./autobuild/ios/artifacts/autobuild_ios_3_copy_files.sh"
-             uses_codeql:           false
+             run_codeql:            false
              xcode_version:         12.1.1
 
            - config_name:           MacOS Legacy (artifacts)
@@ -131,7 +131,7 @@ jobs:
              cmd1_prebuild:         "./autobuild/mac/artifacts/autobuild_mac_1_prepare.sh 5.9.9"
              cmd2_build:            "./autobuild/mac/artifacts/autobuild_mac_2_build.sh do_not_sign"
              cmd3_postbuild:        "./autobuild/mac/artifacts/autobuild_mac_3_copy_files.sh legacy"
-             uses_codeql:           false
+             run_codeql:            false
              # For Qt5 on Mac, we need to ensure SDK 10.15 is used, and not SDK 11.x.
              # Xcode 12.1 is the most-recent release which still ships SDK 10.15:
              # https://developer.apple.com/support/xcode/
@@ -145,7 +145,7 @@ jobs:
              cmd1_prebuild:         powershell .\autobuild\windows\autobuild_windowsinstaller_1_prepare.ps1
              cmd2_build:            powershell .\autobuild\windows\autobuild_windowsinstaller_2_build.ps1
              cmd3_postbuild:        powershell .\autobuild\windows\autobuild_windowsinstaller_3_copy_files.ps1
-             uses_codeql:           true
+             run_codeql:            true
 
            - config_name:           Windows JACK (artifact)
              target_os:             windows
@@ -153,7 +153,7 @@ jobs:
              cmd1_prebuild:         powershell .\autobuild\windows\autobuild_windowsinstaller_1_prepare.ps1 -BuildOption jackonwindows
              cmd2_build:            powershell .\autobuild\windows\autobuild_windowsinstaller_2_build.ps1 -BuildOption jackonwindows
              cmd3_postbuild:        powershell .\autobuild\windows\autobuild_windowsinstaller_3_copy_files.ps1 -BuildOption jackonwindows
-             uses_codeql:           false
+             run_codeql:            false
 
     runs-on:                        ${{ matrix.config.building_on_os }}
     steps:
@@ -210,7 +210,7 @@ jobs:
           jamulus_buildversionstring: ${{ needs.create_release.outputs.version_name }}
 
       - name:                       Initialize CodeQL
-        if:                         matrix.config.uses_codeql
+        if:                         matrix.config.run_codeql
         uses:                       github/codeql-action/init@v1
         with:
           languages: 'cpp'
@@ -306,5 +306,5 @@ jobs:
           asset_content_type:       application/octet-stream
 
       - name:                       Perform CodeQL Analysis
-        if:                         matrix.config.uses_codeql
+        if:                         matrix.config.run_codeql
         uses:                       github/codeql-action/analyze@v1
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -1,6 +1,5 @@
 #### Automatically build and upload releases to GitHub ####
 
-
 # see analyse_git_reference.py for implementation of the logic:
 #     for every push to a branch starting with "autobuild": (can be used during development for tighter supervision of builds)
 #          - do CodeQl while building for every platform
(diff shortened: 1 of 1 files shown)


.github/workflows/autobuild.yml
```diff
@@ -45,29 +42,25 @@ jobs:
 
 
      steps:
-         # Checkout code
          - name:                    Checkout code
            uses:                    actions/checkout@v2
 
-         # Set variables
-         # Determine release / pre-release
          - name:                    Get Jamulus build info, determine actions & variables
```
ann0see, line 48:
```suggestion
         - name:                    Get Jamulus build/release info, determine actions & variables
```

Although not part of this PR, I don't like this description at all. What is "determine actions"?

.github/workflows/autobuild.yml
```diff
@@ -251,77 +233,81 @@ jobs:
...
-        if:                         ${{ (steps.step_cmd3_postbuild.outputs.artifact_1 != '') && contains(needs.create_release.outputs.publish_to_release, 'true') }}
+        if:                         >-
+                                    steps.step_cmd3_postbuild.outputs.artifact_1 != '' &&
+                                    contains(needs.create_release.outputs.publish_to_release, 'true')
         id:                         upload-release-asset1
         uses:                       actions/upload-release-asset@v1
         env:
           GITHUB_TOKEN:             ${{ secrets.GITHUB_TOKEN }}
         with:
-          upload_url:               ${{ needs.create_release.outputs.upload_url }} # See this blog post for more info: https://jasonet.co/posts/new-features-of-github-actions/#passing-data-to-future-steps
```
ann0see, line 306: Maybe leave this link here and delete it below.

.github/workflows/autobuild.yml
```diff
@@ -216,29 +204,23 @@ jobs:
             /opt/android/android-ndk
           key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/workflows/autobuild.yml', 'autobuild/android/autobuild_apk_1_prepare.sh', 'autobuild/android/install-qt.sh') }}-${{ matrix.config.cmd1_prebuild }}
 
-      # Prepare (install QT & dependencies)
       - name:                       "Prepare for ${{ matrix.config.config_name }}"
```
hoffie, line 207: What about this:

```suggestion
      - name:                       "Install build dependencies for ${{ matrix.config.config_name }}"
```

.github/workflows/autobuild.yml
hoffie: Hm, I'd rather like to remove it. Reasoning:
- While this might have been a new feature at some point, it no longer is. It's standard Github Workflow syntax and not some kind of hard to find trick or something. Anyone who touches the workflow should read the documentation. I'm considering this to be a rather foundational thing and we don't link foundational things in other places either (e.g. C++ or Qt books in source files ;)).
- I don't know why it's worth to document this feature (the "outputs" logic) right here. The whole workflow is sprinkled with `outputs` references as it's such a basic idiom. So if we document it at all, it should be documented at the very top of the file. Maybe I'm missing something, but `grep -cF '.outputs.' .github/workflows/autobuild.yml` returns 33 places where outputs are either set or read. I don't see why those two occurences would be special.

.github/workflows/autobuild.yml
hoffie:
```suggestion
         - name:                    Determine release version, type and prerelease variables
```

At least, that's what the script does. I'm open to other wordings though. :)

.github/workflows/autobuild.yml
hoffie: > Of course the "can also be warn,…" can be removed but the links to the documentation or blog posts should remain.

(I've replied to that directly in the other thread)

> Some comments might be trivial (Checkout code,…)

I'm a huge fan of comments to explain specialities. I'm annoyed by comments which state the obvious. They just make the file longer and make me think "umm, I've just read that already". It's a [code smell](https://moderatemisbehaviour.github.io/clean-code-smells-and-heuristics/comments/c3-redundant-comment.html).
You've already mentioned one extreme case. Here's another.
https://github.com/jamulussoftware/jamulus/blob/master/.github/workflows/autobuild.yml#L238-L239

Thanks for the specific comments, that's much easier to work with. :)

.github/workflows/autobuild.yml
ann0see: Yes. That's better.

.github/workflows/autobuild.yml
ann0see: Yes. I think that's better (which would also imply a change of all the script file names...)

.github/workflows/autobuild.yml
ann0see: Ok. Probably I still think of the old times where the .yaml was basically only a three step procedure (and where I wanted it to be clarified by nefarius). 

So I'll need to be happy with it.

hoffie: > So I'll need to be happy with it.

Well, I don't want to say "it must be like that". I want to outline what my goals are. If I'm missing something or if there are better approaches, I'm open to that. But it should have some justification :)

ann0see: The thing is  - in my opinion - the whole autobuild file has become complicated (there are many specialities. Let me do another review on how I would comment the code)

.github/workflows/autobuild.yml
```diff
@@ -1,6 +1,5 @@
 #### Automatically build and upload releases to GitHub ####
```
ann0see, line 1:
```suggestion
```

.github/workflows/autobuild.yml
```diff
@@ -1,6 +1,5 @@
...
 
-
 # see analyse_git_reference.py for implementation of the logic:
```
ann0see, line 3: Maybe a brief comment like

"This file builds the source and produces artifacts for every operating system. If requested it creates a release and uploads the binaries to GitHub releases. Moreover, it tests the code with CodeQl"

.github/workflows/autobuild.yml
```diff
@@ -14,21 +13,19 @@
...
-      - "autobuild*"  # for developers: branches starting with autobuild will be built and evaluated on each push
-      - "autobuild/*" # for developers: branches starting with autobuild will be built and evaluated on each push
+      # For developers: Branches starting with autobuild will be built and evaluated on each push.
+      - "autobuild**"
       - "master"
-    #  - "*"
-  pull_request: # The branches below must be a subset of the branches in "push"
+  pull_request:
     branches:
+      # These branches must be a subset of the branches in "push":
```
ann0see, line 28:
```suggestion
      # Only automatically run the build on pushes to master to avoid wasting ressources
```

.github/workflows/autobuild.yml
```diff
@@ -14,21 +13,19 @@
...
       - master
 
 name:                               Auto-Build
```
ann0see, line 31: Below at create_release:

Check if we are doing a release or just a normal build. This must be done before actually building the app to clarify where to upload the binaries and if we need to create a release.

.github/workflows/autobuild.yml
```diff
@@ -45,29 +42,25 @@ jobs:
...
            uses:                    actions/checkout@v2
 
-         # Set variables
-         # Determine release / pre-release
-         - name:                    Get Jamulus build info, determine actions & variables
+         - name:                    Determine release version, type and prerelease variables
            run:                     python3 ${{ github.workspace }}/.github/actions_scripts/analyse_git_reference.py
            id:                      jamulus-build-vars
 
-         # remove release, if it exists (with this releasetag)
```
ann0see, line 58: I think this was mainly a bug-fix or related to pushing a release tag? Not sure why this was needed again (ie. why we needed to check for an existing release. But it's good to have it anyways.)

.github/workflows/autobuild.yml
```diff
@@ -80,42 +73,40 @@ jobs:
              draft:                 false
 
 
-         ### CANCEL ### can be used for development concerning release-creation
-         #- name:                    Cancelthrougherroe
-         #  run:                     myundefinedfunction
-
-
   release_assets:
```
ann0see, line 76: Start the actual build process.

.github/workflows/autobuild.yml
```diff
@@ -80,42 +73,40 @@ jobs:
...
     name:                           Build assets for ${{ matrix.config.config_name }}
```
ann0see, line 77:
```suggestion
    name:                           Build for ${{ matrix.config.config_name }}
```

.github/workflows/autobuild.yml
```diff
@@ -157,35 +148,32 @@ jobs:
...
         with:
-          xcode-version: ${{ matrix.config.xcode_version }}
+          xcode-version:            ${{ matrix.config.xcode_version }}
 
-      # Checkout code
       - name:                       Checkout code
         uses:                       actions/checkout@v2
         with:
           submodules:               true
           fetch-depth:              ${{ matrix.config.checkout_fetch_depth || '1' }}
```
ann0see, line 174: Set up build environment. This downloads dependencies if needed or uses those cached from previous runs.

.github/workflows/autobuild.yml
```diff
@@ -251,77 +233,81 @@ jobs:
 
       - name:                       "Post-Build for ${{ matrix.config.config_name }}"
```
ann0see, line 234: Should probably be named something like

Move files... But that's not part of this PR though.

.github/workflows/autobuild.yml
```diff
@@ -251,77 +233,81 @@ jobs:
...
         id:                         step_cmd3_postbuild
-        if:                         ${{ matrix.config.cmd3_postbuild }}
+        if:                         matrix.config.cmd3_postbuild
         run:                        ${{ matrix.config.cmd3_postbuild }} ${{ github.workspace }}
         env:
           jamulus_project_path:       ${{ github.workspace }}
           jamulus_buildversionstring: ${{ needs.create_release.outputs.version_name }}
 
-      # Upload Artifact to Job
```
ann0see, line 260: Upload/Release binaries

.github/workflows/autobuild.yml
```diff
@@ -251,77 +233,81 @@ jobs:
...
-        uses: devbotsxyz/xcode-notarize@d7219e1c390b47db8bab0f6b4fc1e3b7943e4b3b
+          if-no-files-found:        error
+
+      - name:                       "Notarize macOS Release Build"
+        if:                         >-
+                                    steps.step_cmd3_postbuild.outputs.artifact_1 != '' &&
+                                    steps.step_macos_build.outputs.macos_signed == 'true' &&
+                                    contains(needs.create_release.outputs.publish_to_release, 'true')
+        id:                         notarize-macOS-app
+        uses:                       devbotsxyz/xcode-notarize@d7219e1c390b47db8bab0f6b4fc1e3b7943e4b3b
```
ann0see, line 266: Why is this done after the upload artifact call? Couldn't there also be signed non-release artifacts?

.github/workflows/autobuild.yml
```diff
@@ -251,77 +233,81 @@ jobs:
...
+                                    steps.step_cmd3_postbuild.outputs.artifact_1 != '' &&
+                                    steps.step_macos_build.outputs.macos_signed == 'true' &&
+                                    contains(needs.create_release.outputs.publish_to_release, 'true')
+        id:                         staple-macOS-app
+        uses:                       devbotsxyz/xcode-staple@v1
         with:
-          product-path: deploy/${{ steps.step_cmd3_postbuild.outputs.artifact_1 }}
+          product-path:             deploy/${{ steps.step_cmd3_postbuild.outputs.artifact_1 }}
 
-      # Upload Artifact to Release
```
ann0see, line 298: If requested upload artifacts to release.

ann0see (commented):
I think I'm probably commenting a lot about the structure? But please see below:

.github/workflows/autobuild.yml
hoffie: Hm, currently it only runs on release builds. It could probably also be run earlier. Not sure if there are any reasons for not signing dev builds, but as this runs on @emlynmac's fork and not on ours anyway, I'm not sure how deeply we should care...? :)

Maybe open a separate issue for that?

.github/workflows/autobuild.yml
hoffie: Might also be relevant if we encounter any issues with the PR run and have to re-start it. I'd keep it (especially as part of this PR).

.github/workflows/autobuild.yml
hoffie: Hm, I'd prefer improving the name instead. What about this:

```diff
- name:                          Prepare Auto-Build/Release
+ name:                          Create Github release if this is a release build
```

.github/workflows/autobuild.yml
hoffie: Hm, what would be the added value of such a comment? It's just a more elaborate re-phrasing of the `name:` below to me.

.github/workflows/autobuild.yml
hoffie: Github-wise the term is `assets`, so if we want to provide details, we should probably keep this as-is?

.github/workflows/autobuild.yml
hoffie: Yes, that could indeed be improved to match what's actually done in the step (in another PR).

.github/workflows/autobuild.yml
hoffie: This sounds like some kind of headline to me, but technically there is no such thing. All of that is done as part of the `release_assets` job. It doesn't feel right to add such a "block comment" in a place where there is no block.

What are you mainly after at? Maybe adding the "Download dependencies" part to the rather inaccurate "Prepare" step later would be beneficial? I'll add a suggestion below.

I'm all for clarifying the structure of this file. Github's workflow syntax already provides us with a way to technically annotate each step (`name:`). I think its use is even mandatory. Therefore, I think we should do all the useful optimizations there and don't add another level of structure via comments.
Improving `name:` has the benefits of also improving the CI logs in the same go. A `grep '[ -] name:' .github/workflows/autobuild.yml | grep -v artifact` can also be used to get a sense of the structure of this workflow.

.github/workflows/autobuild.yml
```diff
@@ -216,29 +204,23 @@ jobs:
             /opt/android/android-ndk
           key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/workflows/autobuild.yml', 'autobuild/android/autobuild_apk_1_prepare.sh', 'autobuild/android/install-qt.sh') }}-${{ matrix.config.cmd1_prebuild }}
 
-      # Prepare (install QT & dependencies)
-      - name:                       "Prepare for ${{ matrix.config.config_name }}"
-        if:                         ${{ matrix.config.cmd1_prebuild }}
+      - name:                       "Install build dependencies for ${{ matrix.config.config_name }}"
```
hoffie, line 207:
```suggestion
      - name:                       "Download and install build dependencies for ${{ matrix.config.config_name }}"
```

.github/workflows/autobuild.yml
hoffie: The step in question actually just uploads the artifact to the Actions job. I think the `name:` already covers that. The upload to the release does not happen here, but in a later step.

.github/workflows/autobuild.yml
hoffie: Background: https://github.com/jamulussoftware/jamulus/pull/978#issuecomment-778540855

So... now I understand what was probably meant by the original comment. But rather than explaining Why (CodeQL mandates that) it explains the What (the fact that this list is a subset of the other can be derived without the comment). I've now added a comment to explain the reasoning. I've also added the above suggestion.

.github/workflows/autobuild.yml
softins: Looking at the workflow, I think this name is slightly misleading. This step happens first, before the builds, so it isn't actually creating the Github release, just setting up variables to tell later steps to generate one. So "Prepare" or "Setup" is better, but happy for the rest of the name to be made clearer.

.github/workflows/autobuild.yml
```diff
@@ -14,178 +15,168 @@
...
 
 name:                               Auto-Build
 jobs:
   create_release:
-     name:                          Prepare Auto-Build/Release
-     runs-on:                       ubuntu-20.04
-     outputs:
+    # Check if we are doing a release or just a normal build.
+    # This must be done before actually building the app to find out where to upload the binaries and if we need to create a Github release.
+    name:                           Create Github release if this is a release build
```
hoffie, line 38: > Looking at the workflow, I think the new name is slightly misleading. This step happens first, before the builds, so it isn't actually creating the Github release, just setting up variables to tell later steps to generate one. So "Prepare" or "Setup" is better, but happy for the rest of the name to be made clearer.

@softins By *new name* you are referring to the line which this comment is attached to, right?

This job does indeed create the Github release (although it doesn't fill it). The step some lines below is also named like that.

Maybe this would be better:
```suggestion
    name:                           Create empty Github release if this is a release build
```

hoffie: Addition: I just realized that the changelog is generated by `anaylze_git_references.py`. Generating a file in an "analyze_" script certainly violates the principle of least surprise...

As this is both related to this PR and #2471, I think I'll open a separate Issue to track this and work on it when both PRs are merged. I've adjust the name of the `analyse_git_reference.py` step for now.

softins: > > Looking at the workflow, I think the new name is slightly misleading. This step happens first, before the builds, so it isn't actually creating the Github release, just setting up variables to tell later steps to generate one. So "Prepare" or "Setup" is better, but happy for the rest of the name to be made clearer.
> 
> @softins By _new name_ you are referring to the line which this comment is attached to, right?
> 
> This job does indeed create the Github release (although without the artifacts, of course). The step some lines below is also named like that.
> 
> Example: https://github.com/jamulussoftware/jamulus/runs/5265414702?check_suite_focus=true#step:5:1

Ah, ok, I misunderstood. I thought it was just preparing variables for later use. So it's actually creating a release container that will later receive the assets.

My only other comment there would be that the `name` is getting rather long, and becoming more like a comment or description. Can it be made more concise but still accurate?

.github/workflows/autobuild.yml
ann0see: Yes. That's good.

.github/workflows/autobuild.yml
```diff
@@ -14,178 +15,168 @@
...
-             release_name:          ${{ steps.jamulus-build-vars.outputs.RELEASE_TITLE }}
-             body_path:             ${{ github.workspace }}/autoLatestChangelog.md
-             prerelease:            ${{ steps.jamulus-build-vars.outputs.IS_PRERELEASE }}
-             draft:                 false
-
-
-         ### CANCEL ### can be used for development concerning release-creation
-         #- name:                    Cancelthrougherroe
-         #  run:                     myundefinedfunction
+      - name:                       'Create Release ${{steps.jamulus-build-vars.outputs.RELEASE_TAG}}  {{steps.jamulus-build-vars.outputs.RELEASE_TITLE}}'
```
ann0see, line 65: I think we could even drop the variables here (most times they're not even filled).

.github/workflows/autobuild.yml
ann0see: Yes, but I'm unsure if it adds any value.

.github/workflows/autobuild.yml
ann0see: > Improving name: has the benefits of also improving the CI logs

Good point. And yes 'Prepare' sounds impercise.

.github/workflows/autobuild.yml
ann0see: +1

.github/workflows/autobuild.yml
ann0see: @emlynmac could you please comment why this is here and not before any upload has started?

emlynmac: This step is to notarize the build, which has to happen after the build is done and the artifacts created.
The signing of the macOS build is a two-step process; build with certs then notarize the full thing with Apple.

ann0see: Is this basically a step which transfers something (hashes?) to apple?

ann0see:
BTW: See the nightly: https://github.com/jamulussoftware/jamulus/runs/5487103016?check_suite_focus=true

The Create Release step doesn't even output RELEASE_TITLE

.github/workflows/autobuild.yml
```diff
@@ -14,178 +15,169 @@
...
```
ann0see, line 66: I think a $ is missing:

```suggestion
      - name:                       'Create Release ${{steps.jamulus-build-vars.outputs.RELEASE_TAG}}  ${{steps.jamulus-build-vars.outputs.RELEASE_TITLE}}'
```

ann0see (changes requested):

.github/workflows/autobuild.yml
ann0see: Great. Thanks.

.github/workflows/autobuild.yml.orig
```diff
@@ -0,0 +1,318 @@
+# This file builds the source and produces artifacts for every supported platform.
```
ann0see, line 1: What is this file? Probably a backup?

.github/workflows/autobuild.yml
hoffie: > My only other comment there would be that the name is getting rather long, and becoming more like a comment or description. Can it be made more concise but still accurate?

Hrm, you are right. I struggle to find good wording though. Maybe we should drop the imperative for job-level names.

If just adjusted it like that:
```suggestion
    name:                           Build vars & Github release (if required)
```

.github/workflows/autobuild.yml.orig
hoffie: Argh. Yep. Removing.

.github/workflows/autobuild.yml
hoffie: Ok, removed. It also matches the first job's name better, now that the imperative has been dropped.

ann0see (approved):

ann0see:
I think I‘m happy now v

hoffie:
Currently still cleaning up the git history. As @softins had some comments as well I'll wait with merging until tomorrow (or explicit feedback).

.github/workflows/autobuild.yml
emlynmac: There's a bunch of information about how notarization works here: https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution

.github/workflows/autobuild.yml
```diff
@@ -213,29 +206,23 @@ jobs:
             /opt/android/android-ndk
           key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/workflows/autobuild.yml', 'autobuild/android/autobuild_apk_1_prepare.sh', 'autobuild/android/install-qt.sh') }}-${{ matrix.config.cmd1_prebuild }}
 
-      # Prepare (install QT & dependencies)
-      - name:                       "Prepare for ${{ matrix.config.config_name }}"
-        if:                         ${{ matrix.config.cmd1_prebuild }}
+      - name:                       Download and install build dependencies for ${{ matrix.config.config_name }}
```
softins, line 209: Do we need to say "Download and"? Downloading is probably implicit.

.github/workflows/autobuild.yml
```diff
@@ -246,79 +233,83 @@ jobs:
...
+        id:                         staple-macOS-app
+        uses:                       devbotsxyz/xcode-staple@v1
         with:
-          product-path: deploy/${{ steps.step_cmd3_postbuild.outputs.artifact_1 }}
+          product-path:             deploy/${{ steps.step_cmd3_postbuild.outputs.artifact_1 }}
 
-      # Upload Artifact to Release
-      - name:                       Upload Artifact 1 to Release
-        if:                         ${{ (steps.step_cmd3_postbuild.outputs.artifact_1 != '') && contains(needs.create_release.outputs.publish_to_release, 'true') }}
+      - name:                       Upload Artifact 1 to Release if this is a release build
```
softins, line 285: Not sure the "if this is a release" is needed. If it's not a release, the step will be shown in the log as not executed anyway. Same for artifact 2 below.

softins (approved):
Looking good. I left a couple of comments regarding step naming, but they are only suggestions, not deal-breakers :)

Approving anyway.

.github/workflows/autobuild.yml
hoffie: I had added that as a response to feedback by @ann0see.
@ann0see Should we keep "Download and" in or should I shorten it again?

.github/workflows/autobuild.yml
hoffie: I might have added this in response to feedback as well, but I've lost track by now...
@ann0see Should I shorten it as well? :)

ann0see: Probably it’s redundant so yes, you can remove it.

.github/workflows/autobuild.yml
ann0see: Maybe use Set up? But that would be close to the unwanted „Prepare“ again.

hoffie: Ok, done.

.github/workflows/autobuild.yml
hoffie: Done. Will merge on CI green.

.github/workflows/autobuild.yml
ann0see: @hoffie sorry for the small modification request: Could you please add a comment here that this needs to be done after the build is created?

hoffie: Not sure I understand.
Do you mean that the binaries can only be signed once they're built? That would be rather obvious to me (similar for other instances: We don't explicitly highlight that the build dependencies have to be prepared before the build can start).

Or do you mean that the binaries can only be signed once the artifact has been uploaded to the job? I don't think this is the case (or if it is, I don't understand why). The Notarize step references `product-path:             deploy/${{ steps.step_cmd3_postbuild.outputs.artifact_1 }}`. This looks like a filesystem path to me and not like a job artifact path. So I guess the order could be turned around. Adding a comment which states the contrary would seem strange then :)

Maybe you could send a follow-up PR with a suggested wording?

(In general, I think a PR which touches specific code should leave the code in better state than it was before. However, we should be cautious to avoid trying to push to much *additional* things which *could* or *should* be improved into PRs. PRs are getting large and take lots of iterations until they can get merged...)

ann0see: Yes sure. @emlynmac should open a PR then since he knows what to do. I think we can move it around - at least from my readings.

ann0see:
Ok. Ready to merge after Android build is finished.

ann0see:
@hoffie please merge (squash merge?) this if it's finished.

hoffie:
> squash merge?

Although the PR contains more than a few commits, I'd like to keep it separate to keep the information what was changed for what reasons. I've already squashed all the same-type adjustments (wording improvements are a single commit, for example).

Issue #300, opened 2020-05-27
Intercom to individual user


robsco:
If I join a channel, it would be handy to "shout" to a friend in the room, but privately, so not to disturb the group.

Could this be possible?

"Hey Bill, I'll be over in room XYZ when you're done in here."

Think of it as virtually walking up to someone and telling them something, rather than send a text message which they may not read for a while.

Snayler:
Wouldn't that disturb the user you're talking to? When I need to talk with a friend in the middle of a session, I just message him through other means, like whatsapp.

robsco:
That would be the idea, but a "do-not-disturb" could be implemented too, if you really never wanted anyone to talk to you while you're playing.

Whatsapp and other means may work sometimes, but I've had many occasions where I can't get hold of a friend because they're not checking their phone, but did want interrupting.

This would negate the need to keep checking your whatsapp, or the chat window, etc. especially useful if you're not in front of the pc, or don't keep checking your phone.

corrados:
The implementation of the feature would be quite compilcated. Right now at the server a separate mix for each client is calculated. In case of an intercom, you would have to mute your signal for all other clients and only ship that audio data to the desired client. That would mean to add protocol messages for enabling/disabling the intercom and some modifications in the server signal processing. At the same time some intercom button would have to be put on each channel in the client. That would mean you have another button (which may confuse some users). Then you have a lot of special cases. E.g. if intercom is activated and the other user disconnects.

The question is how many people would actually use that. Normally you would write these things in the chat dialog that you, e.g., want to leave the server and go to another server. That would not disturb the audio session.

I think it is too much work to do and too little benefit.

robsco:
OK, no worries if it's too difficult.

corrados:
I don't think this will ever be implemented in Jamulus so I will close this Issue now.

PR #2485, opened 2022-03-09
tools/changelog-helper: Ignore non-MERGED PRs & ignore already listed PRs faster

hoffie:

- tools/changelog-helper: Ignore non-MERGED PRs

- tools/changelog-helper: Ignore already listed PRs faster

CHANGELOG: Internal: Improved Changelog release tooling.

**Context: Fixes an issue?**

https://github.com/jamulussoftware/jamulus/pull/1865#issuecomment-1063468921

Ready.

Ready.

## Checklist

### DIFF
--- a/tools/changelog-helper.sh
+++ b/tools/changelog-helper.sh
@@ -126,7 +126,9 @@ check_or_add_pr() {
         return
     fi
     checked_ids[$id]=1
-    if grep -qF "#$id" <<<"$changelog"; then
+    if grep -qE "#$id\>" <<<"$changelog"; then
+        # Changelog already lists this PR ID -> nothing to do
+        # (\> ensures that we only match full, standalone IDs)
         return
     fi
     local json=$(gh pr view "${id/#/}" --json title,author,state)
--- a/tools/changelog-helper.sh
+++ b/tools/changelog-helper.sh
@@ -126,9 +126,14 @@ check_or_add_pr() {
         return
     fi
     checked_ids[$id]=1
-    local json=$(gh pr view "${id/#/}" --json title,author)
+    local json=$(gh pr view "${id/#/}" --json title,author,state)
+    local state=$(jq -r .state <<<"${json}")
     local title=$(jq -r .title <<<"${json}" | sanitize_title)
     local author=$(jq -r .author.login <<<"${json}")
+    if [[ "${state}" != "MERGED" ]]; then
+        echo "-> Ignoring PR #${id} as state ${state} != MERGED"
+        return
+    fi
     if grep -qF "#$id" <<<"$changelog"; then
         return
     fi
--- a/tools/changelog-helper.sh
+++ b/tools/changelog-helper.sh
@@ -126,15 +126,15 @@ check_or_add_pr() {
         return
     fi
     checked_ids[$id]=1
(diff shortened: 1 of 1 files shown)


ann0see (approved):

tools/changelog-helper.sh
```diff
@@ -126,10 +126,15 @@ check_or_add_pr() {
         return
     fi
     checked_ids[$id]=1
-    local json=$(gh pr view "${id/#/}" --json title,author)
+    if grep -qF "#$id" <<<"$changelog"; then
```
pljones, line 129: That will find `#300` if the list contains `#3000` through to `#3009`.  A more precise match would be good.

hoffie: Good point. Will probably add `\b` or something around that later.

pljones (changes requested):
Line 129 needs a better matcher.

tools/changelog-helper.sh
hoffie: Updated (no \b though).

tools/changelog-helper.sh
```diff
@@ -126,7 +126,7 @@ check_or_add_pr() {
         return
     fi
     checked_ids[$id]=1
-    if grep -qF "#$id" <<<"$changelog"; then
+    if grep -qP "#$id([^0-9]|$)" <<<"$changelog"; then
```
pljones, line 129: Doesn't `"#$id\>"` work?  I don't think I'd want it matching `#12AB`, either.

hoffie: > Doesn't `"#$id\>"` work?

I didn't know about that -- thanks! It works in `-E` mode (may `\b` might have worked there, too). Updated.

tools/changelog-helper.sh
```diff
@@ -126,10 +126,15 @@ check_or_add_pr() {
         return
     fi
     checked_ids[$id]=1
-    local json=$(gh pr view "${id/#/}" --json title,author)
+    if grep -qE "#$id\>" <<<"$changelog"; then
```
pljones, line 129: Can you add a comment on what this does and why, please.

pljones: And, perhaps - like the "not merged" below - echo a message.

tools/changelog-helper.sh
softins: I think `\b` will only work in `-P` (PCRE) mode, if the installed version of `grep` supports it. I prefer `\<` and `\>` if I'm not actually writing perl or PHP.

tools/changelog-helper.sh
hoffie: > Can you add a comment on what this does and why, please.

Done.

> And, perhaps - like the "not merged" below - echo a message.

Hm, I'm not convinced. It was generally nice to see that repeated invocations of this script added less and less stuff (e.g. when run again later in the release process). If we generate a line of output of each PR even if it was already there, the output becomes harder to read, IMO. If you feel strongly, I can add it though.

pljones: Yeah, that's why I was hesitant.  Probably best without.

pljones (approved):

Issue #2474, opened 2022-03-09
CI: Run `shellcheck` and `shfmt` on shell scripts


hoffie:
**What is the current behaviour and why should it be changed?**

We currently don't have any automatic QA on shell scripts.

**Describe possible approaches**

To keep the growing number of shell scripts manageable and consistent, we should run `shellcheck` as a static shell script code analyzer to catch mistakes and `shfmt` to apply a consistent coding style (indentation, etc.), similar to `clang-format`.

- #2547
- #2582
- [ ] https://github.com/hoffie/jamulus/tree/ci-check-shell-scripts

<details>
<summary>Example output (should be empty when we introduce it in CI):</summary>

```
$ shellcheck $(find -name '*.sh' | grep -v libs/)

In ./linux/deploy_deb.sh line 19:
perl .github/actions_scripts/getChangelog.pl ChangeLog "${VERSION}" --line-per-entry | while read entry
                                                                                             ^--^ SC2162 (info): read without -r will mangle backslashes.


In ./mac/deploy_mac.sh line 45:
    qmake "${project_path}" -o "${build_path}/Makefile" "CONFIG+=release" ${@:2}
                                                                          ^----^ SC2068 (error): Double quote array expansions to avoid re-splitting elements.


In ./mac/deploy_mac.sh line 46:
    local target_name=$(sed -nE 's/^QMAKE_TARGET *= *(.*)$/\1/p' "${build_path}/Makefile")
          ^---------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./mac/deploy_mac.sh line 47:
    local job_count="$(sysctl -n hw.ncpu)"
          ^-------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./mac/deploy_mac.sh line 74:
    local app_version="$(sed -nE 's/^VERSION *= *(.*)$/\1/p' "${project_path}")"
          ^---------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./mac/deploy_mac.sh line 121:
    echo Please run this script from the Qt project directory where $(basename "${project_path}") is located.
                                                                    ^---------------------------^ SC2046 (warning): Quote this to prevent word splitting.


In ./mac/deploy_mac.sh line 122:
    echo Usage: mac/$(basename "${0}")
                    ^----------------^ SC2046 (warning): Quote this to prevent word splitting.


In ./mac/deploy_mac.sh line 135:
[...]

PR #2489, opened 2022-03-10
Improve code style in raspijamulus.sh

ann0see:
Codacy complained that quoting is not right in raspijamulus.sh. This PR tries to fix it. I did not test it on a Raspberry Pi, but on my machine it doesn't seem to run. 
This shouldn't touch the refactoring by @hoffie 

CHANGELOG: Refactor raspijamulus.sh file

**Context: Fixes an issue?**
Not known.

Waiting for test.

Needs testing on a raspberry pi.

## Checklist

### DIFF
--- a/distributions/raspijamulus.sh
+++ b/distributions/raspijamulus.sh
@@ -1,16 +1,19 @@
 #!/bin/bash
 
 # This script is intended to setup a clean Raspberry Pi system for running Jamulus
-OPUS="opus-1.3.1"
+# This needs to be run from the distributions/ folder
+
+readonly OPUS="opus-1.3.1"
 NCORES=$(nproc)
+readonly NCORES
 
 # install required packages
-pkgs='alsamixergui build-essential qtbase5-dev qttools5-dev-tools libasound2-dev cmake libglib2.0-dev'
-if ! dpkg -s $pkgs >/dev/null 2>&1; then
+readonly pkgs=(alsamixergui build-essential qtbase5-dev qttools5-dev-tools libasound2-dev cmake libglib2.0-dev)
+if ! dpkg -s "${pkgs[@]}" >/dev/null 2>&1; then
   read -p "Do you want to install missing packages? " -n 1 -r
   echo
-  if [[ $REPLY =~ ^[Yy]$ ]]; then
-    sudo apt-get install $pkgs -y
+  if [[ ${REPLY} =~ ^[Yy]$ ]]; then
+    sudo apt-get install "${pkgs[@]}" -y
     # Raspbian 10 needs qt5-default; Raspbian 11 doesn't need or provide it
     if ! qtchooser -list-versions | grep -q default; then
       sudo apt-get install qt5-default -y
@@ -22,11 +25,11 @@ fi
 if [ -d "${OPUS}" ]; then
   echo "The Opus directory is present, we assume it is compiled and ready to use. If not, delete the opus directory and call this script again."
 else
-  wget https://archive.mozilla.org/pub/opus/${OPUS}.tar.gz
-  tar -xzf ${OPUS}.tar.gz
-  rm ${OPUS}.tar.gz
-  cd ${OPUS}
-  if [ ${OPUS} == "opus-1.3.1" ]; then
+  wget https://archive.mozilla.org/pub/opus/"${OPUS}.tar.gz"
+  tar -xzf "${OPUS}.tar.gz"
+  rm "${OPUS}.tar.gz"
+  cd "${OPUS}" || exit 1
+  if [ "${OPUS}" == "opus-1.3.1" ]; then
     echo "@@ -117,13 +117,19 @@ void validate_celt_decoder(CELTDecoder *st)
  #ifndef CUSTOM_MODES
     celt_assert(st->mode == opus_custom_mode_create(48000, 960, NULL));
@@ -51,7 +54,7 @@ else
     patch celt/celt_decoder.c opus_patch_file.diff
   fi
   ./configure --enable-custom-modes --enable-fixed-point
-  make -j${NCORES}
+  make "-j${NCORES}"
   mkdir include/opus
   cp include/*.h include/opus
   cd ..
@@ -62,10 +65,10 @@ if [ -d "jack2" ]; then
   echo "The Jack2 directory is present, we assume it is compiled and ready to use. If not, delete the jack2 directory and call this script again."
 else
   git clone https://github.com/jackaudio/jack2.git
-  cd jack2
+  cd jack2 || exit 1
   git checkout v1.9.20
-  ./waf configure --alsa --prefix=/usr/local --libdir=$(pwd)/build
-  ./waf -j${NCORES}
+  ./waf configure --alsa --prefix=/usr/local "--libdir=$(pwd)/build"
+  ./waf "-j${NCORES}"
   mkdir build/jack
   cp build/*.so build/jack
   cp build/common/*.so build/jack
@@ -84,7 +87,7 @@ fi
 # compile Jamulus with external Opus library
 cd ..
 qmake "CONFIG+=opus_shared_lib raspijamulus headless" "INCLUDEPATH+=distributions/${OPUS}/include" "QMAKE_LIBDIR+=distributions/${OPUS}/.libs" "INCLUDEPATH+=distributions/jack2/common" "QMAKE_LIBDIR+=distributions/jack2/build/common" Jamulus.pro
-make -j${NCORES}
+make "-j${NCORES}"
 
 # get first USB audio sound card device
 ADEVICE=$(aplay -l|grep "USB Audio"|tail -1|cut -d' ' -f3)
@@ -111,7 +114,7 @@ fi
 
 # start Jack2 and Jamulus in headless mode
 export LD_LIBRARY_PATH="distributions/${OPUS}/.libs:distributions/jack2/build:distributions/jack2/build/common"
-distributions/jack2/build/jackd -R -T --silent -P70 -p16 -t2000 -d alsa -dhw:${ADEVICE} -p 128 -n 3 -r 48000 -s &
+distributions/jack2/build/jackd -R -T --silent -P70 -p16 -t2000 -d alsa "-dhw:${ADEVICE}" -p 128 -n 3 -r 48000 -s &
 ./Jamulus -n -i ${JAMULUSINIFILE} -c anygenre3.jamulus.io &
 
 echo "###---------- PRESS ANY KEY TO TERMINATE THE JAMULUS SESSION ---------###"


distributions/raspijamulus.sh
```diff
@@ -22,10 +22,10 @@ fi
 if [ -d "${OPUS}" ]; then
   echo "The Opus directory is present, we assume it is compiled and ready to use. If not, delete the opus directory and call this script again."
 else
-  wget https://archive.mozilla.org/pub/opus/${OPUS}.tar.gz
-  tar -xzf ${OPUS}.tar.gz
-  rm ${OPUS}.tar.gz
-  cd ${OPUS}
+  wget https://archive.mozilla.org/pub/opus/"${OPUS}".tar.gz
+  tar -xzf "${OPUS}".tar.gz
```
pljones, line 26: Style note...  I'd prefer to see the quoting around the _entire_ parameter, so here `"${OPUS}.tar.gz"` and later `"-j${NCORES}"`.

ann0see: I'm not sure if this is valid (I can't really test it either since my laptop is intel based).

pljones: If `-j3` is valid and `NCORES=3` then `"-j${NCORES}"` better be identical to `-j3` or there's a bug in the shell.

distributions/raspijamulus.sh
```diff
@@ -6,11 +6,11 @@ NCORES=$(nproc)
...
 # install required packages
 pkgs='alsamixergui build-essential qtbase5-dev qttools5-dev-tools libasound2-dev cmake libglib2.0-dev'
-if ! dpkg -s $pkgs >/dev/null 2>&1; then
+if ! dpkg -s "${pkgs}" >/dev/null 2>&1; then
   read -p "Do you want to install missing packages? " -n 1 -r
   echo
-  if [[ $REPLY =~ ^[Yy]$ ]]; then
-    sudo apt-get install $pkgs -y
+  if [[ ${REPLY} =~ ^[Yy]$ ]]; then
+    sudo apt-get install "${pkgs}" -y
```
hoffie, line 13: I don't think this can work:
```
root@e7169628f80d:/# x="vim screen"; apt install "$x"
[...]
E: Unable to locate package vim screen
```

This should either be kept unquoted or use the proper expansion syntax (not sure if I remember correctly, but it could have been `"${pkgs[@]}"` when the pkgs definition would be changed into an array).

hoffie (changes requested):

distributions/raspijamulus.sh
pljones: Yes, you're going to have to use an array if any element of `pkgs` could have a space in the name and there are multiple arguments.  Never use one variable to represent multiple arguments.

softins: But this doesn't apply when we know what is in `pkgs` because it was set explicitly in line 8.

distributions/raspijamulus.sh
```diff
@@ -6,11 +6,11 @@ NCORES=$(nproc)
 
 # install required packages
 pkgs='alsamixergui build-essential qtbase5-dev qttools5-dev-tools libasound2-dev cmake libglib2.0-dev'
-if ! dpkg -s $pkgs >/dev/null 2>&1; then
+if ! dpkg -s "${pkgs}" >/dev/null 2>&1; then
```
softins, line 9: This doesn't work either. Same comment as on line 13.

distributions/raspijamulus.sh
```diff
@@ -22,10 +22,10 @@ fi
...
   echo "The Opus directory is present, we assume it is compiled and ready to use. If not, delete the opus directory and call this script again."
 else
-  wget https://archive.mozilla.org/pub/opus/${OPUS}.tar.gz
-  tar -xzf ${OPUS}.tar.gz
-  rm ${OPUS}.tar.gz
-  cd ${OPUS}
+  wget https://archive.mozilla.org/pub/opus/"${OPUS}.tar.gz"
+  tar -xzf "${OPUS}.tar.gz"
+  rm "${OPUS}.tar.gz"
+  cd "${OPUS}"
```
softins, line 28: Again, no need to quote `${OPUS}`, because we know what is in it, from line 4

distributions/raspijamulus.sh
```diff
@@ -51,7 +51,7 @@ else
     patch celt/celt_decoder.c opus_patch_file.diff
   fi
   ./configure --enable-custom-modes --enable-fixed-point
-  make -j${NCORES}
+  make -j"${NCORES}"
```
softins, line 54: And `${NCORES}` is just a number. Again, quoting is unnecessary and cumbersome.

softins (changes requested):
I don't think many of these items are needed. I don't know this Codacy, but it seems to be applying some rules without any knowledge of whether they should be applicable in this particular context.

For variables that have been set explicitly, one can know they don't contain spaces that are part of a pathname, and extra quoting is either unnecessary or even incorrect. It is only when a variable has been set by user input or globbing for filenames that extra quoting and possibly array syntax are necessary.

distributions/raspijamulus.sh
pljones: I wonder if Codacy understands `readonly` and then would work out that the expression is safe?  Somehow I doubt it.

hoffie:
> I don't think many of these items are needed. I don't know this Codacy, but it seems to be applying some rules without any knowledge of whether they should be applicable in this particular context.
> 
> For variables that have been set literally, one can know they don't contain spaces that are part of a pathname, and extra quoting is either unnecessary or even incorrect. It is only when a variable has been set by user input or globbing for filenames that extra quoting and possibly array syntax are necessary.

In general, I agree that the quoting is optional in many places and quotes may even make it a bit harder to read. At the same time, depending on the script size it becomes harder and harder to keep in mind which things are safe to use without quotes and which are not. I think shellcheck (#2474) suggests doing the same -- quote or write it in a way which makes it obvious that expansion is wanted (i.e. use arrays). Maybe Codacy even uses Shellcheck, I don't know.

So, if we want some kind of static analysis, we either have to accept that everything has to be quoted or we have to sprinkle ignore comments into the scripts.
We'd not be the only ones who go a "quote everything" route (if we do):

https://tldp.org/LDP/abs/html/quotingvar.html
> When referencing a variable, it is generally advisable to enclose its name in double quotes.

https://unix.stackexchange.com/questions/68694/when-is-double-quoting-necessary
> Second, it is far easier to use double quotes all the time than to remember when they are needed

distributions/raspijamulus.sh
pljones: Line 8 is still ugly bash script.  It would be better as `pkgs=(name0 name1 ...)` and then `"${pkgs[@]}` to properly expand it.  Relying on not having spaces in order for expansion to work _is_ bad coding.

distributions/raspijamulus.sh
pljones: Isn't the point of populating a variable that you _don't_ have to know its value to use it?  In which case you should make sure you're using it safely.  (And again, if this is a constant, use `readonly` to declare it.)

distributions/raspijamulus.sh
pljones: Using `declare -ri NCORES=(whatever)` might allow Codacy to realise it's an integer constant, too.

distributions/raspijamulus.sh
ann0see: Why do we even ask?

ann0see:
I *think* codacy uses spellcheck to detect quoting.

ann0see:
Please see my new try. I've now been able to test it in a Debian docker container and it seems to run (except for jack obviously). Couldn't check on real Pi yet - but I'll try in a VM.

distributions/raspijamulus.sh
```diff
@@ -22,10 +23,10 @@ fi
...
   echo "The Opus directory is present, we assume it is compiled and ready to use. If not, delete the opus directory and call this script again."
 else
-  wget https://archive.mozilla.org/pub/opus/${OPUS}.tar.gz
-  tar -xzf ${OPUS}.tar.gz
-  rm ${OPUS}.tar.gz
-  cd ${OPUS}
+  wget https://archive.mozilla.org/pub/opus/"${OPUS}.tar.gz"
+  tar -xzf "${OPUS}.tar.gz"
+  rm "${OPUS}.tar.gz"
+  cd "${OPUS}" || exit
```
softins, line 29: If exiting due to an error, use a non-zero exit status, e.g. `exit 1`. Similarly below on line 66. Then the caller can tell the script failed, if necessary.

softins (commented):
One comment so far. I'll try it on my Pi in the next day or two before approving.

distributions/raspijamulus.sh
ann0see: Done.

distributions/raspijamulus.sh
```diff
@@ -62,10 +65,10 @@ if [ -d "jack2" ]; then
   echo "The Jack2 directory is present, we assume it is compiled and ready to use. If not, delete the jack2 directory and call this script again."
 else
   git clone https://github.com/jackaudio/jack2.git
-  cd jack2
+  cd jack2 || exit
```
hoffie, line 68: @ann0see Needs a ` 1` as well, as @softins hinted on the other place. :)

ann0see: Yes. Just added it.

hoffie (approved):
Code looks good to me (haven't tested on a Pi).

distributions/raspijamulus.sh
```diff
@@ -1,16 +1,19 @@
 #!/bin/bash
 
-# This script is intended to setup a clean Raspberry Pi system for running Jamulus
-OPUS="opus-1.3.1"
+# This script is intended to setup a clean Raspberry Pi system for running Jamulus.
+# Please run this script from the distributions/ folder
+
+readonly OPUS="opus-1.3.1"
 NCORES=$(nproc)
```
pljones, line 7: You can say `declare -ri NCORES="$(nproc)"` to declare NCORES both read-only _and_ containing an integer value.

ann0see:
The Raspberry Pi VM is probably as slow as a real Raspberry Pi 1st edition, so I'm going to abort it soon. 32 Bit Intel worked (although I had the feeling that the script did "too much").

distributions/raspijamulus.sh
```diff
@@ -22,10 +25,10 @@ fi
...
 else
-  wget https://archive.mozilla.org/pub/opus/${OPUS}.tar.gz
-  tar -xzf ${OPUS}.tar.gz
-  rm ${OPUS}.tar.gz
-  cd ${OPUS}
+  wget https://archive.mozilla.org/pub/opus/"${OPUS}.tar.gz"
+  tar -xzf "${OPUS}.tar.gz"
+  rm "${OPUS}.tar.gz"
+  cd "${OPUS}" || exit 1
   if [ ${OPUS} == "opus-1.3.1" ]; then
```
pljones, line 32: `"${OPUS}"`

distributions/raspijamulus.sh
ann0see: I did that but spellcheck complained: https://www.shellcheck.net/

distributions/raspijamulus.sh
```diff
@@ -62,10 +65,10 @@ if [ -d "jack2" ]; then
   echo "The Jack2 directory is present, we assume it is compiled and ready to use. If not, delete the jack2 directory and call this script again."
 else
   git clone https://github.com/jackaudio/jack2.git
-  cd jack2
+  cd jack2 || exit 1
   git checkout v1.9.20
-  ./waf configure --alsa --prefix=/usr/local --libdir=$(pwd)/build
-  ./waf -j${NCORES}
+  ./waf configure --alsa --prefix=/usr/local --libdir="$(pwd)"/build
```
pljones, line 70: `"--libdir=$(pwd)/build"` is nicer (or `--libdir="$(pwd)/build"` - but splitting the /build off doesn't look nice.)

distributions/raspijamulus.sh
```diff
@@ -111,7 +114,7 @@ fi
 
 # start Jack2 and Jamulus in headless mode
 export LD_LIBRARY_PATH="distributions/${OPUS}/.libs:distributions/jack2/build:distributions/jack2/build/common"
-distributions/jack2/build/jackd -R -T --silent -P70 -p16 -t2000 -d alsa -dhw:${ADEVICE} -p 128 -n 3 -r 48000 -s &
+distributions/jack2/build/jackd -R -T --silent -P70 -p16 -t2000 -d alsa -dhw:"${ADEVICE}" -p 128 -n 3 -r 48000 -s &
```
pljones, line 117: Again, I'd do `"-dhw:${ADEVICE}"` if it's a single argument.

distributions/raspijamulus.sh
pljones: Oh dear...  What did it complain about?

ann0see:
```
declare -ri NCORES=$(nproc)
            ^-- [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155) (warning): Declare and assign separately to avoid masking return values.
```

ann0see:
Messed a bit with the rebase while trying to move the commits around. It's now all squashed and I hope I didn't loose anything.

hoffie:
> Messed a bit with the rebase while trying to move the commits around. It's now all squashed and I hope I didn't loose anything.

The diff between your force push and the state before looks ok to me:

<details>
<summary>Diff</summary>

```diff
$ git diff f86d21..8795988
diff --git a/distributions/raspijamulus.sh b/distributions/raspijamulus.sh
index 5dabcfc3..71f777db 100755
--- a/distributions/raspijamulus.sh
+++ b/distributions/raspijamulus.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 
 # This script is intended to setup a clean Raspberry Pi system for running Jamulus
+# This needs to be run from the distributions/ folder
+
 readonly OPUS="opus-1.3.1"
 NCORES=$(nproc)
 readonly NCORES
@@ -66,7 +68,7 @@ else
   cd jack2 || exit 1
   git checkout v1.9.20
   ./waf configure --alsa --prefix=/usr/local "--libdir=$(pwd)/build"
-  ./waf -j${NCORES}
+  ./waf "-j${NCORES}"
   mkdir build/jack
   cp build/*.so build/jack
   cp build/common/*.so build/jack
```
</details>

distributions/raspijamulus.sh
pljones: Oh dear, then it is pretty broken.  You _cannot_ declare something readonly and then assign to it...  It's right about assignment masking the exit code:
```
~$ ( declare -ri NCORES=$(nproc) || { echo "foo"; exit 1; }; echo $NCORES )
6
~$ ( declare -ri NCORES=$(barbaz) || { echo "foo"; exit 1; }; echo $NCORES )
barbaz: command not found
0
```
Where the second go should have said "foo", not "0".  So it should equally not like
```
readonly NCORES=$(nproc)
```
on that basis:
```
$ ( readonly NCORES=$(barbaz) || { echo "foo"; exit 1; }; echo $NCORES )
barbaz: command not found

$
```
(no "foo" echoed).  You end up with
```
tempvarnamehere=$(ncores) || exit 1
declare -ri NCORES="$tempvarnamehere"
```
which should keep it happy.  Otherwise you end up having `"...${NCORES}..."` everywhere.

softins (approved):
Tried it on my Pi4 with a fresh image of Raspbian 11, and it works ok.

I'm not sure anyone other than Volker actually uses the script. I see that after building (if necessary), it automatically starts a headless client, connecting to `anygenre3.jamulus.io` (which actually ends up on the directory server for Any Genre 1, because no custom port is given), with the username `Raspi raspberryp` (where the `raspberryp` part comes from the default hostname of the Pi, truncated).

ann0see:
Yes, I think as soon as we have ARM builds, this script should be (re)-moved

PR #2502, opened 2022-03-13
Autobuild: Combine and simplify Windows build scripts

hoffie:

- Move all autobuild/windows/* scripts to a single autobuild/windows.ps1 script which is called for the different stages (setup/build/get-artifacts).
- Condense redundant parameter parsing into a single step
- Simplify jack build naming
- Make coding style more consistent (spaces, curly braces)

CHANGELOG: Autobuild: Refactored and simplified Windows build scripts.
(will be condensed...)

Notes:
- This is a rather large Diff which cannot be easily followed. I'm sorry for that, but I don't see any way to make this better. I'm usually a fan of having individual commits for sub-changes, but in this case it would have required artificially creating "bad" intermediate states just to fix them in a later commit. Therefore, the diff is rather large, but I'm hoping that the resulting file is way more readable than what we had before.
- I've kept the script in autobuild/ as it used to be. We can also think of moving all autobuild scripts (now that they are only a single script) to their platform folder, e.g. move autobuild/windows.ps1 to windows/autobuild.ps1. If wanted, I can do this as part of this PR.
- This PR drops some error checking code which is not relevant for Github-initiated builds as it makes the scripts way shorter and more readable. They error checks sound like they may have been relevant during development or when running locally. However, the scripts were not designed to run locally as they make use of the Github-setup build environments and techniques (e.g. magic set-output lines; it's not impossible, but it wasn't straight-forward either; it's still possible and might be a bit more straight-forward at least).
- Doing this for Windows is obviously just the first step. After this PR, the other platforms will be a bit unaligned in how the scripts are run. I still chose to submit this now because I consider it ready and I think it's easier to review and reason about. PRs for the other platforms will be done in the coming days (even if I was unable to manage to complete them timely for whatever reasons, all platforms are in working state).
- Once all platforms are converted to the new model, I'll simplify the build matrix (it will be enough to supply a single base `command:` entry for each platform). I will also drop a number of unneeded commands (GITHUB_WORKSPACE extra logic).

**Context: Fixes an issue?**

https://github.com/jamulussoftware/jamulus/discussions/2282#discussioncomment-2075961
Related: #2503

No.

- [x] Artifact size should be compared. Result: [Artifacts from this PR](https://github.com/jamulussoftware/jamulus/actions/runs/1977361917) match the size of [artifacts from latest master run](https://github.com/jamulussoftware/jamulus/actions/runs/1977279935)
- [x] It should be confirmed that the JACK binary does in fact use JACK and not ASIO

## Checklist


autobuild/windows.ps1
```diff
@@ -0,0 +1,159 @@
+param(
```
ann0see, line 1: If we really want this to only be GitHub compatible, I'd add a comment here to say that this is a GitHub only script.

Maybe also create a README.md file in autobuild/

ann0see (commented):
I'd still be more happy if we let it in a windows specific folder for now. Maybe in future, we need to add other files like documentation?

autobuild/windows.ps1
hoffie: > If we really want this to only be GitHub compatible, I'd add a comment here to say that this is a GitHub only script.

Good idea, done.

> Maybe also create a README.md file in autobuild/

Not sure. What would we add there? The fact that the directory is called autobuild/ should also be halfway obvious. Maybe we should move it to a more appropriate place altogether? Something like .github/autobuild/ similar to .github/action_scripts?

I also want to re-emphasize that the scripts already used Github-specific logic and this is nothing which is introduced as part of this PR. Therefore I want to limit the work in that direction to a minimum to get this PR merged quickly. I'm not opposed to further improvements afterwards, of course. :)

pljones: I think a one line README.md so people not familiar with Github understand is a good idea.

ann0see: > .github/action_scripts

Yes. Probably that would be best:

./github/autobuild_scripts

ann0see:
I can't find the JACK build artifact of this run. Maybe there's a naming clash?

autobuild/windows.ps1
hoffie: Ok, added a README and used .github/autobuild* now.

hoffie:
> I can't find the JACK build artifact of this run. Maybe there's a naming clash?

Thanks for the pointer. Build step three lacked the necessary `-BuildOption`. Added that now and waiting for CI.

hoffie:
> Build step three lacked the necessary `-BuildOption`. Added that now and waiting for CI.

That fixed it.
I've also compared the resulting binary sizes with the most recent master CI run and their Jamulus.exe sizes match exactly (ASIO: 2532352, JACK: 2514944). See PR description for links.
I'm fairly confident that this PR is good to go now.

ann0see:
Great! It works now.

.github/autobuild/README.md
```diff
@@ -0,0 +1,7 @@
+# Scripts for building Jamulus via GitHub Actions
+
+The scripts in this fodler are used by the Github autobuild actions in .github/workflows/autobuild.yml.
```
ann0see, line 3:
```suggestion
The scripts in this folder are used by the Github autobuild actions in .github/workflows/autobuild.yml.
```

hoffie: FIxed, thanks.

.github/autobuild/windows.ps1
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    echo "Get Qt 64 bit..."
+    Install-Qt "${Qt64Version}" "${Msvc64Version}"
+
+    echo "Get Qt 32 bit..."
+    Install-Qt "${Qt32Version}" "${Msvc32Version}"
+}
+
+Function Ensure-jom
```
ann0see, line 82: I'm not sure if you used "Ensure-jom" because of [approved verbs](https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.2) but I'd prefer something like Install-jom or similar. One could probably argue that it is named "Ensure" due to caching?

.github/autobuild/windows.ps1
```diff
@@ -0,0 +1,169 @@
...
+    choco install --no-progress -y -f --forcex86 jack --version "${JackVersion}"
+    if (!$?)
+    {
+        throw "32bit jack installation failed with exit code $LastExitCode"
+    }
+}
+
+Function Build
+{
+    echo "Build installer..."
```
ann0see, line 113:
```suggestion
    echo "Build app and create installer..."
```

The actual build installer step is in deploy_windows.ps1 I assume this hasn't changed?

.github/autobuild/windows.ps1
```diff
@@ -0,0 +1,169 @@
...
+    {
+        "jackonwindows" { "_jack"; break }
+        "^\S+$"         { "_${BuildOption}"; break }
+        default         { "" }
+    }
+
+    $artifact_deploy_filename = "jamulus_${JamulusVersion}_win${ArtifactSuffix}.exe"
+
+    echo "Copying artifact to ${artifact_deploy_filename}"
+    cp ".\deploy\Jamulus*installer-win.exe" ".\deploy\${artifact_deploy_filename}"
```
ann0see, line 139: Maybe the deploy script could be adapted to create the correct name (but that's for another PR)

.github/autobuild/windows.ps1
```diff
@@ -0,0 +1,169 @@
...
+    "setup"
+    {
+        choco config set cacheLocation $ChocoCacheDir
+        Ensure-Qt
+        Ensure-jom
+        Ensure-JACK
+    }
+    "build"
+    {
+        Build
```
ann0see, line 159: This looks really strange to me (the Powershell syntax): Maybe add a short comment if we plan to continue going the 3 step build. But as I understood, you'd like to move away - so it doesn't matter now probably

.github/autobuild/windows.ps1
hoffie: Thanks for the pointer regarding approved verbs, I didn't know about it.

Yes, I chose that name due to caching. The function `Install-Qt` already existed and performs the actual installation. Therefore, I needed a name for the higher-level function which checks if Qt needs to be installed at all and only calls Install-Qt if necessary.

Ensure-jom and Ensure-JACK have been named to match that.

As it is, it is the most consistent across the stuff which we install.
I can change it to Install- for jom and JACK as actual installations do happen there (even on cached builds -- it's just that the installation source is from the cache). Do you think that would be better?

Ensure-Qt would still have to remain like that though (although I'm open to naming suggestions).

.github/autobuild/windows.ps1
hoffie: Done.
(BTW, this is not a change of this PR ;))

.github/autobuild/windows.ps1
hoffie: I think they'll stay at 2 (setup, build) or even 3 steps (as it is). The initial autobuild logic's reason for separate steps seems to have been the added value regarding runtime statistics. That might still be interesting, so we should probably keep it.

What would you want to have commented exactly? I'm pretty new to PowerShell, but this part sounds like an ordinary switch statement to me?
https://docs.microsoft.com/en-us/powershell/scripting/learn/deep-dives/everything-about-switch?view=powershell-7.2

I've just realized that the official docs use the `switch ( foo )` spacing style (similar to what we do for C++), so I've just adapted this to be consistent as well.

ann0see: In my opinion it looks a bit like that:

switch ($switch)
  case 'switch': switch

I'd prefer something like

"build" {
  # Run deploy script to build Jamulus and installer
  Build
}

The comment is basically just there to make a clear cut between the case and the function.

hoffie: Ah, now I understand. I've just renamed the function -- does that make it more readable?

ann0see (approved):
Yes, I think that ok.

.github/autobuild/windows.ps1
ann0see: Yes.

.github/autobuild/windows.ps1
ann0see: I can’t come up with a better name at the moment. Probably it isn’t worth changing.

pljones (approved):

Issue #2503, opened 2022-03-13
Autobuild: Combine scripts into single script per-platform


hoffie:
**What is the current behaviour and why should it be changed?**

- The autobuild scripts have grown a lot and are hard to reason about.
- Code is duplicated within the same platform (GITHUB_WORKSPACE handling), across platforms (e.g. mac & ios are very similar, yet they are copies).
- They rely on the discouraged, insecure Github feature `ACTIONS_ALLOW_UNSECURE_COMMANDS: 'true'` (which could be worked around, but why do that if we can make it redundant?)

**Describe possible approaches**

My plan is to combine all the platform-specific scripts into a single per-platform script.

- [x] #2502
- [x] #2509
- [x] #2514
- [x] #2521
- [x] #2527
- [x] #2540
- [ ] ~~Check if Mac/iOS can partially re-use stuff from the other script:~~ After looking at the diff, it's not worth it
- [x] Mac/iOS: Shorten artifact_deploy_filename
- [x] Mac: Simplify PATH handling similar to iOS
- [x] Convert autobuild.yml to use a single `command:` in the build matrix & adjust inline comments
- [x] Consider moving the scripts to `.github/autobuild/` instead (see https://github.com/jamulussoftware/jamulus/pull/2502#discussion_r825433890)
- [x] Remove unused Workflow environment variables (e.g. `jamulus_project_path`)
- [x] Consider moving the autobuild-only scripts from `.github/action_scripts` to `.github/autobuild/` or `.github/autobuild_scripts` (@ann0see) as well, contained in #2584 and #2656
- [ ] ~~Consider removing `deploy/` from `autobuild.yml` to allow scripts to set output directly instead of copying stuff around~~: This isn't possible as the name is used both for the path (that's fine) and the name (that's problematic as it should not be a path, obviously).
- [x] Remove `ensure_THIS_JAMULUS_PROJECT_PATH.sh`
- [x] "Install dependencies" -> "Install Qt"
- [x] #2652
- [x] `#!/bin/bash & set -eu everywhere?`
- [x] `powershell ".\windows\deploy_windows.ps1" "C:\Qt\${Qt32Version}" "C:\Qt\${Qt64Version}" @ExtraArgs`
- [x] `export DEBIAN_FRONTEND="noninteractive"` everywhere?
[...]

PR #2521, opened 2022-03-18
Autobuild: Combine and simplify iOS build scripts

hoffie:

- Extract the non-autobuild-related parts into an ios/deploy_ios.sh script, which can run standalone, similar to other platforms (by @ann0see)

- Move all autobuild/ios/* scripts to a single .github/autobuild/ios.sh script which is called for the different stages (setup/build/get-artifacts).

- Condense redundant parameter parsing into a single step

- Make shellcheck-clean

CHANGELOG: Autobuild: Refactored and simplified iOS build scripts.
(will be condensed...)

**Context: Fixes an issue?**

Related: #2503

Ready for review.

Reviews.

## Checklist

- [x] I tested my code and it does what I want: partially


ann0see (approved):

ios/deploy_ios.sh
```diff
@@ -1,27 +1,19 @@
...
-cd "${THIS_JAMULUS_PROJECT_PATH}"
+## Builds an ipa file for iOS. Should be run from the repo-root
 
+# Create Xcode file and build
 qmake -spec macx-xcode Jamulus.pro
 /usr/bin/xcodebuild -project Jamulus.xcodeproj -scheme Jamulus -configuration Release clean archive -archivePath "build/Jamulus.xcarchive" CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO
-# Make a deploy folder
-mkdir deploy
+
 # Generate ipa by copying the .app file from the xcarchive file
```
pljones, line 10: `xcarchive file` looks like a directory.

ios/deploy_ios.sh
```diff
@@ -1,27 +1,19 @@
...
-mkdir deploy/Payload
-cp -r build/Jamulus.xcarchive/Products/Applications/Jamulus.app deploy/Payload/
-cd deploy
+mkdir build/Payload
+cp -r build/Jamulus.xcarchive/Products/Applications/Jamulus.app build/Payload/
+cd build || exit 1
```
pljones, line 13: `set -eu` should mean the `|| exit 1` isn't needed -- it's effectively the same thing, as far as I'm aware.

hoffie: Hmm, [shellcheck made me do this](https://github.com/koalaman/shellcheck/wiki/SC2164) and I assumed that `cd` (as a shell built-in) was special compared to other commands which I know are checked with `set -e`.
I've run a quick test and it looks you are right. I'm going to remove the `|| exit 1`.

This is what happened (probably): I took the script, ran shell check, noted the warning, fixed it and only later noticed that the script did not use `set -eu` and added that. So, rather PEBKAC than a shellcheck bug... :)

ios/deploy_ios.sh
hoffie: I don't know anything about that part, but based on the `cp` invocation, it must be a directory, so I've adjusted the comment as well (cc @ann0see).

ann0see: On macOS it displays as a file although you can cd into it. It’s similar to .app files. They show up as file, but are essentially folders

pljones: A truely broken design...

pljones (approved):
Looks very nice now.

PR #2526, opened 2022-03-19
Autobuild: Add armhf 32bit (Raspberry-compatibe) Linux .deb packages

### LINKED ISSUE #1452: Provide Debian/Ubuntu ARM builds

Multiple discussions, e.g.
https://github.com/jamulussoftware/jamulus/pull/1446#discussion_r607056616
No explicit agreement, as far as I know.

The autobuild process should build arm binaries for Debian/Ubuntu as well.

 

Make arm users (e.g. Raspberry owners) build themselves.

Last I checked, Github's native runners were x86-only. We may either have to cross-compile (not sure if it is possible without obstacles or how simple this would be) or we would need a self-hosted runner.

hoffie:

- Autobuild: Add Linux .deb armhf (Raspberry) 32bit builds
  This uses the existing build logic, but makes it cross-compiler aware.

- Build: Use bash as shell and fail on errors in linux/deploy.deb

CHANGELOG: Autobuild: Added experimental support for armhf (Raspberry-compatible) Linux .deb packages

**Context: Fixes an issue?**

Fixes #1452

Though not required, it would be helpful to advertise this.
- [ ] Docs
- [ ] Release announcement platform list
- [ ] Website downloads

Builds, binary has the proper arch and headless binary works on my RPi 3 Model B.
```
$ file ./usr/bin/jamulus-headless
./usr/bin/jamulus-headless: ELF 32-bit LSB pie executable, ARM, EABI5 version 1 (GNU/Linux), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0, BuildID[sha1]=dd4c54e78824cabcc13e2396f828b078e227f382, stripped
```

- It would be good to have actual tests on other Raspberry models
- It would be good to have actual tests of a Jamulus client

## Checklist

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -94,13 +94,21 @@ jobs:
             # Jamulus.pro needs to count git history length for android versioning:
             checkout_fetch_depth:   '0'
 
-          - config_name:            Linux .deb (artifacts+codeQL)
+          - config_name:            Linux .deb amd64 (artifacts+codeQL)
             target_os:              linux
             building_on_os:         ubuntu-18.04
             cmd1_prebuild:          ./.github/autobuild/linux_deb.sh setup
             cmd2_build:             ./.github/autobuild/linux_deb.sh build
             cmd3_postbuild:         ./.github/autobuild/linux_deb.sh get-artifacts
             run_codeql:             true
+
+          - config_name:            Linux .deb armhf (artifacts)
+            target_os:              linux
+            building_on_os:         ubuntu-18.04
+            cmd1_prebuild:          TARGET_ARCH=armhf ./.github/autobuild/linux_deb.sh setup
+            cmd2_build:             TARGET_ARCH=armhf ./.github/autobuild/linux_deb.sh build
+            cmd3_postbuild:         TARGET_ARCH=armhf ./.github/autobuild/linux_deb.sh get-artifacts
+            run_codeql:             false
 
           - config_name:            MacOS (artifacts+codeQL)
             target_os:              macos
--- a/distributions/debian/rules
+++ b/distributions/debian/rules
@@ -1,13 +1,22 @@
 #!/usr/bin/make -f
 
 export QT_SELECT=qt5
+DEB_TARGET_GNU_TYPE := $(shell dpkg-architecture -qDEB_TARGET_GNU_TYPE)
+DEB_HOST_GNU_TYPE := $(shell dpkg-architecture -qDEB_HOST_GNU_TYPE)
+DEB_BUILD_GNU_TYPE := $(shell dpkg-architecture -qDEB_BUILD_GNU_TYPE)
+
+ifeq ($(DEB_BUILD_GNU_TYPE),$(DEB_HOST_GNU_TYPE))
+	QMAKE := qmake
+else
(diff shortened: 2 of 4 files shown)


.github/workflows/autobuild.yml
```diff
@@ -94,14 +94,22 @@ jobs:
             # Jamulus.pro needs to count git history length for android versioning:
             checkout_fetch_depth:   '0'
 
-          - config_name:            Linux .deb (artifacts+codeQL)
+          - config_name:            Linux .deb x86_64 (artifacts+codeQL)
```
hoffie, line 97: The debian name (and our .deb artifact name) is amd64. So we could name this amd64 instead of x86_64 to make this consistent.
Counter-point: amd64 is unintuitive for people who don't want to think about CPU architectures. :)

pljones: I think it's better to be consistent with the platform naming conventions - as that's the target audience.  We're likely not the only software they'll be dealing with, so doing something unexpected will cause more confusion than not.

hoffie: Ok. Changed to amd64 now.

ann0see:
So we need to run this on every PR? I‘d like to save resources and only build on merge to master and release tags. Yes there’s the argument that we don’t See breakage soon – however the intel build is very similar.

hoffie:
> So we need to run this on every PR? I‘d like to save resources and only build on merge to master and release tags. Yes there’s the argument that we don’t see breakage soon – however the intel build is very similar.

I understand the desire to save ressources, but we should have a consistent plan there. When is a build different enough to allow for skipping it? We build Windows for 32bit & 64bit as well. Aren't they very similar? We build headless on each run as well, isn't it very similar?
I see a huge benefit of building all supported architectures on all code/build-related PRs.

Basically, I think we should have this discussion here:
https://github.com/jamulussoftware/jamulus/discussions/2282#discussioncomment-2046262

... and do in this PR what has been done for the other platforms as well. :)

pljones (approved):

distributions/debian/rules
```diff
@@ -1,13 +1,22 @@
...
+endif
 
 %:
 	dh $@
 
 override_dh_auto_configure:
-	mkdir -p build-gui && cd build-gui && qmake "CONFIG+=noupcasename" PREFIX=/usr ../Jamulus.pro
-	mkdir -p build-nox && cd build-nox && qmake "CONFIG+=nosound headless" TARGET=jamulus-headless PREFIX=/usr ../Jamulus.pro
+	mkdir -p build-gui && cd build-gui && $(QMAKE) "CONFIG+=noupcasename" PREFIX=/usr ../Jamulus.pro
+	mkdir -p build-nox && cd build-nox && $(QMAKE) "CONFIG+=nosound headless" TARGET=jamulus-headless PREFIX=/usr ../Jamulus.pro
```
ann0see, line 19: Worth a separate issue, but this should also be somehow restructured (e.g. build/gui and build/nox (better build/headless))

ann0see (approved):
Approving since the CI seems to be ok, but should still be tested on a real Raspberry Pi.

softins (approved):
The changes look sane. I haven't studied them in fine detail.

I downloaded the full artifact on my Pi4 with Raspbian 11 (32 bit), and it installed with no problems. I then ran it, did a few tests and listened to a session, and it seems to work fine.

ann0see:
Since the related documentation issue is done, removing needs documentation

PR #2527, opened 2022-03-19
 Autobuild: Combine and simplify Android build scripts

hoffie:

- Move all autobuild/android/* scripts to a single
  .github/autobuild/android.sh script which is called for the different stages
  (setup/build/get-artifacts).

- Make shellcheck-clean

- Drop wget dependency and use curl for all downloads

- Keep variables as local as possible; only export what's necessary

- Move further hardcoded versions to variables

And as this is also the last PR which converts the platform-specific scripts, this includes the now possible cleanup:

- Autobuild: Remove unused autobuild/ensure_THIS_JAMULUS_PROJECT_PATH.sh
  Now that all platforms have been converted to the new scheme (.github/autobuild/*), the autobuild directory only contains the
ensure_THIS_JAMULUS_PROJECT_PATH.sh script which is no longer used.

CHANGELOG: Autobuild: Refactored and simplified Android build scripts.
(will be condensed...)

**Context: Fixes an issue?**

Related: #2521

No.

Ready for review.

Reviews.

## Checklist

- [x] I tested my code and it does what I want: @ann0see tested it


.github/autobuild/android.sh
```diff
@@ -0,0 +1,112 @@
...
+        setup_qt
+        ;;
+    build)
+        build_app_as_apk
+        ;;
+    get-artifacts)
+        pass_artifact_to_job
+        ;;
+    *)
+        echo "Unknown stage '${1:-}'"
```
ann0see, line 110: Although this would apply to all scripts: We should probably output something like "NOTE: This script is designed to run on GitHub Actions, not locally. Please run the deploy scripts in the respective OS folder"

.github/autobuild/android.sh
```diff
@@ -0,0 +1,112 @@
...
+    echo "Environment variable jamulus_buildversionstring has to be set to a valid version string"
+    exit 1
+fi
+
+setup_ubuntu_dependencies() {
+    export DEBIAN_FRONTEND="noninteractive"
+
+    sudo apt-get -qq update
+    sudo apt-get -qq --no-install-recommends -y install build-essential zip unzip bzip2 p7zip-full curl chrpath \
+        libxkbcommon-x11-0 openjdk-8-jre-headless openjdk-8-jdk-headless
```
ann0see, line 34: Why do we need libxkbcommon-x11? According to the Debian documentation this is related to the keyboard somehow?

.github/autobuild/android.sh
```diff
@@ -0,0 +1,112 @@
...
+}
+
+setup_android_sdk() {
+    mkdir -p "${ANDROID_BASEDIR}"
+
+    if [[ -d "${COMMANDLINETOOLS_DIR}" ]]; then
+        echo "Using commandlinetools installation from previous run (actions/cache)"
+    else
+        mkdir -p "${COMMANDLINETOOLS_DIR}"
+        curl -s -o downloadfile "https://dl.google.com/android/repository/commandlinetools-linux-${COMMANDLINETOOLS_VERSION}_latest.zip"
```
ann0see, line 44: We should check the md5 hash on downloads in my opinion (but that's also not related).

.github/autobuild/android.sh
```diff
@@ -0,0 +1,112 @@
...
+}
+
+build_app_as_apk() {
+    QT_DIR="${QT_BASEDIR}/${QT_VERSION}/android"
+    MAKE="${ANDROID_NDK_ROOT}/prebuilt/${ANDROID_NDK_HOST}/bin/make"
+
+    "${QT_DIR}/bin/qmake" -spec android-clang
+    "${MAKE}" -j "$(nproc)"
+    "${MAKE}" INSTALL_ROOT=android-build -f Makefile install
+    "${QT_DIR}"/bin/androiddeployqt --input android-Jamulus-deployment-settings.json --output android-build \
```
ann0see, line 84: I'd rename android-build to build/ to be consistent to Windows.

ann0see (commented):
Thanks! I don't have an android device with me at the moment, but the CI looks ok. See my - partly unrelated - comments below.

.github/autobuild/android.sh
hoffie: No idea. Seems like it was part of the initial addition in a0a2f1205. I can try to remove it and see what happens.
jre is also redundant if jdk is installed.

.github/autobuild/android.sh
hoffie: Hmm...
- Regarding download errors (corruptions), we should be safe due to TCP (lost or corrupted packets) and TLS (mangled packets). Both would fail/retry if transmission errors occur.
- Regarding security, I'm not sure what the benefit would be. Where would we get the hash from? From the very Google page as well?

ann0see: Do it once locally and then put it up into the script?

hoffie: Well, but where do I get the hash from in the first place? I would have to download from the very same source, so the only protection benefit would be a benefit over time, i.e. if Google became malicious or a malicious actor could replace the file there.
We aren't handling that attack vector anywhere else either (Qt, choco, pip, Android SDK contents, ...), do we?

I'm not opposed to doing that per-se, but I'd rather have a discussion and do it in all places if we decide to do it. Remember that it will carry a rather big maintenance burden though.

I think the current consensus is:
- Ensure authenticity/integrity of downlods by using HTTPS
- Use trusted download URLs
- Pin versions, especially for user-generated content (e.g. pip, choco) [and elsewhere for reproducibility reasons)

ann0see: Probably this is #1737 then.

.github/autobuild/android.sh
hoffie: Done. Let's hope I haven't introduced new errors... Waiting for CI. :)

ann0see: Seems good.

ann0see (approved):
Didn't test due to the lack of a real android device but it compiles, so it should be ok?

pljones (approved):

PR #1921, opened 2021-07-17
Do not run ci if only md files are changed

ann0see:
This was done to avoid running the ci if it's unneeded.

**Short description of changes**

Avoid running autobuild if only .md files are changed. 

**Context: Fixes an issue?**

Fixes: https://github.com/jamulussoftware/jamulus/issues/1806

**Does this change need documentation? What needs to be documented and how?**

Internal: Do not run CI if .md files are changed. This was done to avoid unneeded runs of the CI.

**Status of this Pull Request**

Working implementation

**What is missing until this pull request can be merged?**
Review from two maintainers.

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want: https://github.com/ann0see/jamulus/commits/autobuild/test-ci-md
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

PR #2529, opened 2022-03-19
Move entitlements file to mac folder

ann0see:
**Short description of changes**
Moves the Jamulus.entitlements file to the mac/ folder to clean up the root folder

CHANGELOG: Internal: Moved entitlements file to mac/ folder to clean up the repo root.

**Context: Fixes an issue?**
No. Clean up only.

**Does this change need documentation? What needs to be documented and how?**
No.

**Status of this Pull Request**
Should be tested by @emlynmac if it still builds correctly. Also it reminds me of the sandbox writing issue (https://developer.apple.com/forums/thread/666634)

**What is missing until this pull request can be merged?**
Needs testing - especially with a signed build.

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

PR #2532, opened 2022-03-20
Do not run CI on documentation changes

ann0see:
Stops running the CI if the documentation changes. There is no need of running the CI on non source files. New version of #1921 I think this is a quick fix (and can be update ofc.)

CHANGELOG: Stop running Autobuild if only documentation is updated to avoid wasting computation time.

**Context: Fixes an issue?**
No. Related to https://github.com/jamulussoftware/jamulus/discussions/2282#discussioncomment-2046262

No

Ready for review

## Checklist

- [x] I tested my code and it does what I want (https://github.com/ann0see/jamulus/actions/runs/2013115499 and https://github.com/ann0see/jamulus/commits/autobuild/noRunDocs are small tests)

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -27,9 +27,27 @@ on:
       # CodeQL requires every branch from on.pull_request to be part of on.push as well in order to run comparisons.
       # We also need master here to trigger builds on PR merge to master and manual pushes (e.g. as part of the release process):
       - "master"
+    paths-ignore:
+      - '**README.md'
+      - 'docs/**'
+      - 'SECURITY.md'
+      - 'COMPILING.md'
+      - 'COPYING'
+      - 'APPLEAPPSTORE.LICENCE.WAIVER'
+      - '.github/ISSUE_TEMPLATE/*'
+      - '.github/pull_request_template.md'
   pull_request:
     branches:
       - master
+    paths-ignore:
+      - '**README.md'
+      - 'docs/**'
+      - 'SECURITY.md'
+      - 'COMPILING.md'
+      - 'COPYING'
+      - 'APPLEAPPSTORE.LICENCE.WAIVER'
+      - '.github/ISSUE_TEMPLATE/*'
+      - '.github/pull_request_template.md'
 
 name:                               Auto-Build
 jobs:


ann0see:
Peter‘s alternative approach might also work, but this is faster to implement.

hoffie (approved):
Seems safe to do.

Note: README.md is part of Jamulus.pro's DISTFILES, so I guess it's part of `make dist`, but we don't run that on CI.

`paths-ignore` doesn't work for tags per the docs, so even if we ignore too much, we should be on the safe side for releases.

The duplication is not nice, but I don't see a way around.
Doing trigger-based ignores has the advantage of being able to still run the Action manually if this logic is wrong for some reason.

hoffie:
Maybe I'm the only one caring about `git log` cleanliness, but I'd prefer context in the commit message.
`Autobuild: Don't run on README/docs-only commits` (the current message isn't clear that it's about the auto build logic)

softins:
Just wondering, what about `ChangeLog`? And are the other `*.md` files in the root dir going to move into `docs/` at some point?

pljones:
> Just wondering, what about `ChangeLog`? And are the other `*.md` files in the root dir going to move into `docs/` at some point?

ChangeLog, README.md and any licence-related files should stay at root level, otherwise discovery proves more difficult.  (Particularly auto-discovery.)

ann0see:
Probably true 😀. Not sure what the gold standard is.

pljones:
I _think_ all the remaining top-level docs need to stay there.  
* APPLEAPPSTORE.LICENCE.WAIVER and COPYING are licence-related
* README.md is definitely needed there
* Anyone looking for COMPILING.md shouldn't have to look to far - it's a "first place of reference" for those who are doing it themselves
* SECURITY.md and CONTRIBUTING.md could perhaps move unless used by Github for "project completeness scoring" or whatever

ann0see:
> should ChangeLog be included in the files that do not trigger a rebuild?

I excluded it for the reason that I think the Debian build uses it and it’s related to the release process

ann0see:
> Jamulus.entitlements - no idea what this is or why it's at top level.

#2529

PR #2533, opened 2022-03-21
Android: Incomplete Qt6 support

hoffie:

I'm creating this PR to document and archive the needed steps to get Android built with Qt6. Sadly, I'm hitting a dead end and will not continue this PR for now.

This PR is based on #2527. Until that PR is merged, its changed will show up here as well.

State of this PR:
- [x] Qt archive updates: Qt5 provided packages for all 4 common Android ABIs. Qt6 requires installation of individual packages. In addition, it requires installation of the desktop Linux qtbase package in order to have a matching `qmake` binary. In addition that requires the Qt `icu` build because standard Debian(Ubuntu versions don't fit.
- [x] Qt6 uses Gradle 7.2 which requires Java 11
- [x] Qt6 requires the Debian package libglib2.0.0-0
- [x] [Qt6 removes several features from the `AndroidManifest.xml` logic. They can simply be dropped.](https://doc.qt.io/qt-6/android-manifest-file-configuration.html#qt-permissions-and-features).
  - [ ] Most other code has been hidden behind version checks, while this change hasn't. I'm not sure what it causes for Qt5.
CHANGELOG: 
- [x] Jamulus uses QtAndroid (qtandroidextras package). Qt6 no longer ships that. [The workaround is using the private API instead](https://bugreports.qt.io/browse/QTBUG-84382?focusedCommentId=566019&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-566019). This requires `QT += core-private`, it requires a `#define QtAndroid QtAndroidPrivate` and other ugly logic. Even after that, it won't work out of the box as types are slightly different and *sync API versions have been dropped. Those changes are no longer part of the PR. I tried dropping the Android-specific code altogether in favor of adding the required permission (`RECORD_AUDIO`) directly to the `AndroidManifest.xml`.
  - [ ] I suspect that this might not work as there are hints that some permissions (and maybe exactly this one) are only granted when the App asks explicitly.
- [x] Qt6 requires `minSdkVersion=23` which has been bumped accordingly.
- [ ] Qt6 drops support for building muti-ABI APKs in one go via `ANDROID_ABIS`. [There's no plan to re-add this feature for `qmake`-based builds](https://bugreports.qt.io/browse/QTBUG-95990). [6.3.0 re-adds support for this, but only for cmake-based projects.](https://www.qt.io/blog/android-multi-abi-builds-are-back). This is the main reason why I stopped with this PR. I don't see a sane way forward. Some ideas for the future:
  - Attempt to generate a single APK from multiple `qmake/make` runs. I tried that as part of the PR with no success. The necessary native libraries are built and kept there, but `androiddeployqt` still errors out.
  - There are docs for workarounds, which I haven't investigated yet: https://codereview.qt-project.org/c/qt/qtdoc/+/372498/7/doc/src/platforms/android/android-publishing-to-googleplay.qdoc
  - We could consider switching to cmake (at least for Android). It seems like the future of `qmake` is a bit unclear for post-Qt6 anyway.
  - We could stop distributing a single .apk and would distribute separate .apks instead (better build times, but worse user experience)

**Context: Fixes an issue?**

Qt6 support.

Broken and won't be fixed in the short term.

Multi-ABI support, see above.


hoffie:
As noted in the description, this PR has only been created for archival reasons. Therefore, I'm closing it.

PR #2534, opened 2022-03-21
Autobuild: Don't run building ARM on non-releases

ann0see:
This prevents unneeded builds on ARM since the intel builds already 
check the same code. We still run this on betas and everything which creates a GitHub release.

CHANGELOG: SKIP

**Context: Fixes an issue?**
No, https://github.com/jamulussoftware/jamulus/discussions/2282#discussioncomment-2046262 is related

No.

Probably worth a discussion; however my standpoint is: We don't need to check building on ARM until we do beta releases. Also we could ensure that the workflow_dispatch event runs ARM builds (but that isn't implemented yet).

Review.

## Checklist

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -102,7 +102,8 @@ jobs:
             cmd3_postbuild:         ./.github/autobuild/linux_deb.sh get-artifacts
             run_codeql:             true
 
-          - config_name:            Linux .deb armhf (artifacts)
+          - config_name:            Linux .deb armhf (artifacts, release only)
+            if:                     contains(needs.create_release.outputs.publish_to_release, 'true')
             target_os:              linux
             building_on_os:         ubuntu-18.04
             cmd1_prebuild:          TARGET_ARCH=armhf ./.github/autobuild/linux_deb.sh setup
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -103,7 +103,7 @@ jobs:
             run_codeql:             true
 
           - config_name:            Linux .deb armhf (artifacts, release only)
-            if:                     contains(needs.create_release.outputs.publish_to_release, 'true')
+            if:                     needs.create_release.outputs.publish_to_release == 'true'
             target_os:              linux
             building_on_os:         ubuntu-18.04
             cmd1_prebuild:          TARGET_ARCH=armhf ./.github/autobuild/linux_deb.sh setup


hoffie:
In the order of preference:
1. We currently run the full autobuild logic for all platforms in order to detect breakage as soon as possible and I think this is reasonable. Yes, this requires resources, but it also provides confidence at the time of PR approval.
2. If we shift that goal to "detect 90% of breakage only and still provide quick feedback" in order to save ressources, I'd rather want a more radical approach:
    - Restrict PR-triggered builds to stripped down builds for the main platforms (Windows 64bit, Mac (no server), Linux (no headless)). This would require additional logic though.
    - Add logic to decide when to run full builds anyway (any .github/{workflow,autobuild}/-related changes, any $PLATFORM/-related changes, any libs/-related changes, any manually triggered runs).
3. Only disable .deb armhf (and arm64, if we add it). It sounds rather arbitrary to me. For example, why would we continue to build 32bit binaries as part of the Windows build? Isn't it just another arch as well? But if accepting this is the way forward to make additional target platform builds possible, then I'd accept it as well, of course.

.github/workflows/autobuild.yml
```diff
@@ -102,7 +102,8 @@ jobs:
             cmd3_postbuild:         ./.github/autobuild/linux_deb.sh get-artifacts
             run_codeql:             true
 
-          - config_name:            Linux .deb armhf (artifacts)
+          - config_name:            Linux .deb armhf (artifacts, release only)
+            if:                     contains(needs.create_release.outputs.publish_to_release, 'true')
```
hoffie, line 106: Just curious, is there any advantage of using contains() vs. just a string comparison via `==`?

ann0see: Not sure? I think it's what nefarius used.

ann0see:
> We currently run the full autobuild logic for all platforms in order to detect breakage as soon as possible and I think this is reasonable. Yes, this requires resources, but it also provides confidence at the time of PR approval.

Yes, but we do a lot of duplication if we're just after the code check:

* JACK build for Windows is very similar to the Linux build
* ARM is basically the same as the intel build

These are two jobs which in my opinion should only run on releases. I don't actually see 10% decrease in quality checking. (I mean in theory you'd need to check all build options for all possible combinations of Qt and OS to be fully sure)

> If we shift that goal to "detect 90% of breakage only and still provide quick feedback" in order to save ressources, I'd rather want a more radical approach

Yes, that's certainly an option, but not sure if it's worth the effort now.


> But if accepting this is the way forward to make additional target platform builds possible, then I'd accept it as well, of course.

In my opinion we can add many build targets for the release, but not for every PR. I find that a waste of resources with low benefit (since we're always checking the same code.)

.github/workflows/autobuild.yml
softins: I can't see any other use of `contains()` in this file. Other similar places just use `==`.
```suggestion
            if:                     needs.create_release.outputs.publish_to_release == 'true'
```

hoffie: I agree it should be changed. Just wanted to say that the workflow *did* use lots of `contains()` until b3f9c2c352cc07b17a6c2a3a1fd210e0bac65818 from 2022-03-23... :)

I think `==` is more appropriate here.

hoffie:
I understand the reasoning, but I'm not really happy with the direction of this PR.
A possible compromise:
1. Let's do all of the following for all non-main targets (i.e. everything except Windows ASIO, Linux deb amd64, macOS & macOS Legacy¹)
2. Keep building all targets for any kind of (pre-)release (as this PR already does)
3. Provide a way for people to build all targets on their branches. Otherwise, it will need extra-work (Autobuild adjustments or continuous tagging) to do any build-related changes.
  a.  Either keep doing all builds on branches named autobuild*
  b. Or at least keep doing all builds on branches which an even more specific name (e.g. autobuild-all/)
4. Disable non-main target builds on PRs by default
  a. Enable them based on a heuristic (each change to .github/{autobuild,workflows}, {windows/linux/mac/...}/). This will still miss platform-specific changes which are in other files (e.g. something in an #ifdef)
  b. Alternatively (or additionally?) provide a way to signal that all builds should be done. I'm thinking of a checkbox in the PR template (default to "No") which is parsed by the build process (e.g. `[ ] This PR contains platform- or build-specific changes, run all build targets`)

¹ The reason for including macOS Legacy is that it provides the very important verification that code changes still work with old Qt versions.

If you don't want to work on that in this depth, I can do it, but we should first have some consensus here (@pljones  - any opinions? ;)).

pljones:
The argument here is that there's no benefit to running CodeQL on both ARM64 and X86_64 architectures, if I understand it - assuming the target OS is the same (i.e. Linux ARM64 vs Linux X86_64 are considered equal, whereas Linux ARM64 and MacOS ARM64 are not).

Do we have evidence to support this?  What _scope_ are we considering - just Jamulus or the libraries it uses, too?

Presumably, where an OS is delivered primarily on ARM32 (e.g. Android), the same argument applies with respect to ARM64?  That is, we're proposing to run CodeQL on the primary architecture for the OS - but for at least one instance of each OS on which Jamulus is delivered.

ann0see:
There are multiple arguments:
- There's no need to run CodeQl on the exact same code base (except for an architecture change) --> Android & Linux
- There's no need to check and build basically exactly same builds twice (Windows JACK = Windows ASIO + Linux build), legacy, ARM & Intel Linux builds on every small change.

ann0see:
@hoffie I'm not planning to do big Autobuild changes at the moment (this PR is simple...).

I think a checkbox is OK, however we could probably also extend the Workflow_dispatch trigger somehow to enable/disable a full build. 
autobuild/ branches could still build the full release build, maybe even master, but IMHO a PR doesn't need that on every commit.

hoffie:
> I think a checkbox is OK, however we could probably also extend the Workflow_dispatch trigger somehow to enable/disable a full build.
> autobuild/ branches could still build the full release build, maybe even master, but IMHO a PR doesn't need that on every commit.

Ok, sounds good to me, maybe I can give this a try in the coming days.

hoffie:
Hmm, just re-checked the code of this PR and I don't understand how it would work. I think you can set `if` as part of the `config` hash, but I don't think it has a special meaning (it's just another freely choosable field name). At least on this PR I still see `armhf`.

On a related note, I've implemented the ideas mentioned above and I think I've gotten it to work in https://github.com/hoffie/jamulus/commit/autobuild-limit-build-targets now.
This will need some further polishing and testing, but after that I'll submit it as a PR.

ann0see:
Ok. So closing this PR then. Feel free to open one based on your work.

PR #2535, opened 2022-03-22
Add rate-limiting for channel gain change messages

### LINKED ISSUE #2492: Effect of MIDI fader audio changes lags behind UI changes
**Describe the bug**
Attempting to use MIDI control for the first time. Launching the Mac client with this command:
`open -a "Jamulus.app" --args --mutestream --ctrlmidich "1;f70*8"`
The effect of my MIDI controller knobs on the corresponding Jamulus fader UI controls is immediate, but the effect on the actual audio level lags behind by ~10 seconds. If I do the same action by manipulating the UI faders with my mouse, there's still a bit of lag but much less.

**To Reproduce**
For example, picking a play whose LEDs show the person playing constantly. With UI fader at zero, I quickly move the MIDI controller knob until the UI fader goes all the way to 100%. After about ~4 seconds, the player's sound level starts slowly rising, reaching full after ~10 seconds. If I do the same action by manipulating the UI with my mouse, the audio level reaches full in ~4 seconds or less.

**Expected behavior**
Manipulating the faders via MIDI should be at least as responsive as manipulating the UI via mouse.

**Screenshots**
N/A

**Operating system**
MacOS Big Sur (11.5.2) on MacBook Pro (Retina, 15-inch Mid 2014), 2.2 GHz Quad-Core Intel Core i7, 16 GB 1600 MHz DDR3 RAM

**Version of Jamulus**
3.8.2 and 3.8.1

**Additional context**
My MIDI controller is a Akai MPK mini mk3

softins:

Adds rate-limiting for channel gain change messages.
This avoids a backlog of messages being queued due to ACK latency,
particularly if using a MIDI controller that send fine-grained
level changes. After sending a gain change message, further gain
changes will be updated locally for ~300ms~ a time and then the latest
value sent to the server. This time will be 50ms by default, or double the
current ping time, whichever is greater.

CHANGELOG: Client: Fix potential long delay in sending fader changes to the server.

**Context: Fixes an issue?**

Fixes #2492 

No, bug fix only

Tested and working

Review and testing on other platforms, and by @bawbgale, who reported the issue

## Checklist

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -173,6 +173,11 @@ CClient::CClient ( const quint16  iPortNumber,
     // start timer so that elapsed time works
     PreciseTime.start();
 
+    // set gain delay timer to single-shot and connect handler function
+    TimerGain.setSingleShot ( true );
+
+    QObject::connect ( &TimerGain, &QTimer::timeout, this, &CClient::OnTimerRemoteChanGain );
+
     // start the socket (it is important to start the socket after all
     // initializations and connections)
     Socket.Start();
@@ -335,15 +340,85 @@ void CClient::SetDoAutoSockBufSize ( const bool bValue )
     CreateServerJitterBufferMessage();
 }
 
+// In order not to flood the server with gain change messages, particularly when using
+// a MIDI controller, a timer is used to limit the rate at which such messages are generated.
+// This avoids a potential long backlog of messages, since each must be ACKed before the next
+// can be sent, and this ACK is subject to the latency of the server connection.
+//
+// When the first gain change message is requested after an idle period (i.e. the timer is not
+// running), it will be sent immediately, and a 300ms timer started.
+//
+// If a gain change message is requested while the timer is still running, the new gain is not sent,
+// but just stored in newGain[iId], and the minGainId and maxGainId updated to note the range of
+// IDs that must be checked when the time expires (this will usually be a single channel
+// unless channel grouping is being used). This avoids having to check all possible channels.
+//
+// When the timer fires, the channels minGainId <= iId < maxGainId are checked by comparing
+// the last sent value in oldGain[iId] with any pending value in newGain[iId], and if they differ,
+// the new value is sent, updating oldGain[iId] with the sent value. If any new values are
+// sent, the timer is restarted so that further immediate updates will be pended.
+
 void CClient::SetRemoteChanGain ( const int iId, const float fGain, const bool bIsMyOwnFader )
 {
+    QMutexLocker locker ( &MutexGain );
+
     // if this gain is for my own channel, apply the value for the Mute Myself function
     if ( bIsMyOwnFader )
     {
         fMuteOutStreamGain = fGain;
     }
 
+    if ( TimerGain.isActive() )
+    {
+        // just update the new value for sending later;
+        // will compare with oldGain[iId] when the timer fires
+        newGain[iId] = fGain;
+
+        // update range of channel IDs to check in the timer
+        if ( iId < minGainId )
+            minGainId = iId; // first value to check
+        if ( iId >= maxGainId )
+            maxGainId = iId + 1; // first value NOT to check
+
+        return;
+    }
+
+    // here the timer was not active:
+    // send the actual gain and reset the range of channel IDs to empty
+    oldGain[iId] = newGain[iId] = fGain;
     Channel.SetRemoteChanGain ( iId, fGain );
+
+    maxGainId = 0;
+    minGainId = MAX_NUM_CHANNELS;
+
+    // start timer to delay sending further updates
+    TimerGain.start ( GAIN_DELAY_PERIOD_MS );
+}
+
+void CClient::OnTimerRemoteChanGain()
+{
+    QMutexLocker locker ( &MutexGain );
+    bool         bSent = false;
+
+    for ( int iId = minGainId; iId < maxGainId; iId++ )
+    {
+        if ( newGain[iId] != oldGain[iId] )
+        {
+            // send new gain and record as old gain
+            float fGain = oldGain[iId] = newGain[iId];
+            Channel.SetRemoteChanGain ( iId, fGain );
+            bSent = true;
+        }
+    }
+
+    // if a new gain has been sent, reset the range of channel IDs to empty and start timer
+    if ( bSent )
+    {
+        maxGainId = 0;
+        minGainId = MAX_NUM_CHANNELS;
+
+        TimerGain.start ( GAIN_DELAY_PERIOD_MS );
+    }
 }
 
 bool CClient::SetServerAddr ( QString strNAddr )
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -304,6 +304,8 @@ void CClient::OnCLPingReceived ( CHostAddress InetAddr, int iMs )
         const int iCurDiff = EvaluatePingMessage ( iMs );
         if ( iCurDiff >= 0 )
         {
+            iCurPingTime = iCurDiff; // store for use by gain message sending
+
             emit PingTimeReceived ( iCurDiff );
         }
     }
@@ -388,11 +390,7 @@ void CClient::SetRemoteChanGain ( const int iId, const float fGain, const bool b
     oldGain[iId] = newGain[iId] = fGain;
     Channel.SetRemoteChanGain ( iId, fGain );
 
-    maxGainId = 0;
-    minGainId = MAX_NUM_CHANNELS;
-
-    // start timer to delay sending further updates
-    TimerGain.start ( GAIN_DELAY_PERIOD_MS );
+    StartDelayTimer();
 }
 
 void CClient::OnTimerRemoteChanGain()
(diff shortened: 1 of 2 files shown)


softins:
Tested with single and grouped channels, and the Wireshark trace examined.

src/client.cpp
```diff
@@ -335,15 +340,84 @@ void CClient::SetDoAutoSockBufSize ( const bool bValue )
...
+        }
+    }
+
+    // if a new gain has been sent, reset the range of channel IDs to empty and start timer
+    if ( bSent )
+    {
+        maxGainId = 0;
+        minGainId = MAX_NUM_CHANNELS;
+
+        TimerGain.start ( GAIN_DELAY_PERIOD_MS );
```
pljones, line 419: Why start the timer now?  Wait until after the next update (which may never come).

src/client.cpp
```diff
@@ -335,15 +340,84 @@ void CClient::SetDoAutoSockBufSize ( const bool bValue )
...
+{
+    QMutexLocker locker ( &MutexGain );
+    bool         bSent = false;
+
+    for ( int iId = minGainId; iId < maxGainId; iId++ )
+    {
+        if ( newGain[iId] != oldGain[iId] )
+        {
+            // send new gain and record as old gain
+            Channel.SetRemoteChanGain ( iId, oldGain[iId] = newGain[iId] );
```
pljones, line 408: When does the assignment happen?  Before the call or after?  I'd rather see it on a separate line so I don't have to think about it.

src/client.h
```diff
@@ -73,6 +73,9 @@
 // audio reverberation range
 #define AUD_REVERB_MAX 100
 
+// delay period between successive gain updates (ms)
+#define GAIN_DELAY_PERIOD_MS 300
```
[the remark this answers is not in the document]
softins: No, someone moving faders would be hard pushed to notice a 300ms delay in the sound changing, but it could take nearly that long for the message to be acked from a very distant server. The aim is not to generate and queue messages faster than they could be dequeued for sending. Message N+1 is not sent until message N has been acked. So during the delay period, we just keep track of updates to the requested value and then send only the latest one when the timer fires.

src/client.cpp
softins: Any time we have sent a message (or maybe several grouped messages in the timer handler), we start the time to say we don't want to send any more until it expires. Then we send the latest value if it is different from the last.

src/client.h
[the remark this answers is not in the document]
pljones: I guess if it does become noticed, it's easy enough to adjust.

src/client.cpp
softins: The assignment expression happens first, and the assigned value is used as the parameter. To me it's a common C idiom, and in this case it also avoids repeating the array lookup again (which the compiler may possibly optimise out, but I like to write optimal code of my own).

src/client.cpp
pljones: OK, got it - if this sends, then the other method won't as the mutex prevents it running until this has finished and then the timer will be active, so it'll just store the value.

src/client.cpp
softins: One could do:
```
fGain = oldGain[iId] = newGain[iId];
Channel.SetRemoteChanGain ( iId, fGain );
```
but to me the original is perfectly readable.

src/client.h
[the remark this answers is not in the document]
softins: And if a single update is requested after an idle period (e.g. clicking Mute, or clicking the Fader straight to a target position, if the GUI allows it), then it will still get sent immediately, and after the time fires it will just notice there has been no change.

softins:
Harmless code alerts in Oboe, but Github won't let me dismiss them.

DavidSavinkoff:
I think waiting for 50 milliseconds would give a more realistic response because 300 ms is too long.
Also, the server should make finer level changes for the same amount of time. For example, the client updates level changes every 50 ms if commanded, and the server may smoothly update the received level with several smaller level changes within the following 50 ms.

src/client.cpp
```diff
@@ -173,6 +173,11 @@ CClient::CClient ( const quint16  iPortNumber,
     // start timer so that elapsed time works
     PreciseTime.start();
 
+    // set gain delay timer to single-shot and connect handler function
```
hoffie, line 176: That comment almost repeats the variable/method names and provides little added value, but I guess it's ok.. :)

src/client.cpp
```diff
@@ -335,15 +340,84 @@ void CClient::SetDoAutoSockBufSize ( const bool bValue )
...
+// This avoids a potential long backlog of messages, since each must be ACKed before the next
+// can be sent, and this ACK is subject to the latency of the server connection.
+//
+// When the first gain change message is requested after an idle period (i.e. the timer is not
+// running), it will be sent immediately, and a 300ms timer started.
+//
+// If a gain change message is requested while the timer is still running, the new gain is not sent,
+// but just stored in newGain[iId], and the minGainId and maxGainId updated to note the range of
+// IDs that must be checked when the time expires (this will usually be a single channel
+// unless channel grouping is being used). This avoids having to check all possible channels.
```
hoffie, line 354: As we are talking about worst case n=200 or something and this is not part of the critical path (sound/network processing) and not part of something which scales exponentially (e.g. server-side per-client-per-client stuff), I think I would have gone with the check-all-channels approach in order to keep the logic simpler and shorter.

PS: Now this whole block comment provides value and is well-written! :)

src/client.cpp
hoffie: I find it readable, but it's easy to miss, IMO.
I would have gone with a simple two-step assignment / method call as I don't think this is in a performance-critical path.
The alternative (reuse `fGain`) also looks fine to me.

In the end, I don't feel strongly.

softins:
@softins I just tried that build with a server 160ms away and it works much better! MIDI controller changes take effect very quickly. Thank you!

_Originally posted by @bawbgale in https://github.com/jamulussoftware/jamulus/issues/2492#issuecomment-1075763709_

src/client.cpp
softins: Yes, but I figured if it is possible at little cost to limit the number of checks needed to the specific one or a few out of the 200 channels, it's worth doing. And of course include an explanation, as I did :)

src/client.cpp
softins: OK, I've updated it for clarity to use `fGain`.

src/client.cpp
```diff
@@ -335,15 +340,85 @@ void CClient::SetDoAutoSockBufSize ( const bool bValue )
...
+        }
+    }
+
+    // if a new gain has been sent, reset the range of channel IDs to empty and start timer
+    if ( bSent )
+    {
+        maxGainId = 0;
+        minGainId = MAX_NUM_CHANNELS;
+
+        TimerGain.start ( GAIN_DELAY_PERIOD_MS );
```
hoffie, line 420: Addition to the already closed conversation:
If I get this right, starting the timer right away implies that once the delaying logic kicks in, it needs another 300ms without changes to go back to the default send-immediate logic.
(Just an observation, no request for changes)

hoffie:
I've tested this PR and it does what it claims to do. Wireshark looks as expected.

At the same time, this causes very noticable loudness jumps (similar to artifacts) when changing the fader level (tested via streaming music and listening to the sound coming back from the server while playing with the slider of my channel).
One could argue that the most common slider adjustments would be:
- Slow, but not by large amounts
- Quick, but only to make a (maybe new) participant less loud quickly
Therefore, it might not be too relevant.

With 50ms, it's much smoother again, but obviously causes more protocol messages.
@DavidSavinkoff's proposal to add server-side fading would probably work around that, but I don't see a way to add that without introducing new logic into the very critical sound mixing code path (we do have fading for new channels, but I'm not sure if it's possible to adapt).

In essence, I think
- we should have this PR as it solves a rather nasty problem
- we should try to use the lowest possible constant for delayed updates

src/client.cpp
softins: Yes, that's right, any time we send a gain message (or several, in the case of a group), we start the delay period and collect up further changes until the end of the delay.

softins:
I think any server-side smoothing of gain changes might be a worthwhile improvement, but should be a separate feature and PR, as it would only apply to servers running new code. The current PR addresses the root issue at the client, which is where it originates, and works when connected to any version of server.

I did wonder about having a smaller default delay period, and then increasing it if necessary based on the ping time of the connected server. Maybe 50ms or 100ms by default, and 1.5x or 2x the ping time if greater (to allow a little extra time)? It sounds like that might be a worthwhile improvement.

softins:
The ping time multiplied by a suitable constant could easily be stored in `CClient` by `CClient::OnCLPingReceived()` for use when starting the timer.

softins:
I've added a second commit to default the timer to `max(50ms, pingtime*2)`, and refactored the timer starting into a separate function to be DRY. I tested against my private server in London and a server in Hong Kong and compared the packet traces. As expected, when changing a fader, the London server was sent updates at 50ms intervals, and the HK server at 400ms intervals.

src/client.cpp
hoffie: Yes, I mainly meant the fact that the logic stays active until there's one period with no changes.
I guess it's good as it makes it more deterministic.

hoffie (approved):
Thanks! This behaves much better and looks clean. Nice work :)

PR #2540, opened 2022-03-23
Autobuild: Derive all step commands from a base_command + multiple minor fixes

hoffie:

This is the last bigger PR in this Autobuild cleanup series. It performs several refactorings, cleanups and fixes which only became possible after the previous ground-work or only became apparent by now.
This PR touches multiple aspects of Autobuild, but is solely focused on refactoring without any changes in the output.

- Autobuild: Derive all step commands from a base_command
  Previously, the build matrix had to provide three different command lines for the three different build steps. As all command invocations are normalized across platforms now, we can condense that into a base_command per matrix entry and will just add the build stage as a parameter in the generic invocation. This allows removal of several lines of definitions and makes the
workflow more readable/maintenable.

- Autobuild: Drop irregular contains() usage from Workflow
  The Workflow used the `contains()` function in multiple places. [contains() will either perform an in-array check or a substring check](https://docs.github.com/en/actions/learn-github-actions/expressions#contains). We are not dealing with arrays, nor do we want a substring check when checking for `'true'`. Therefore, use an ordinary string comparison here which is more readable anyways.

- Autobuild: Drop jamulus_project_path from Workflow
  The previous refactorings have adjusted all platform autobuild scripts to stop using this variable. Therefore, we can now safely drop the definition from the Workflow.

- Autobuild: Mac: Set PATH before build step directly
  There's no need to set PATH in the prepare step, pass it back to the workflow, let the workflow pass it back to the build stage and use it only then. Therefore, just set PATH right before it's needed in the build step and drop the Github logic.

- Autobuild: Workflow: Drop discouraged ACTIONS_ALLOW_UNSECURE_COMMANDS
  As no autobuild scripts require setting environment variables across build steps anymore, we can now safely drop this discouraged configuration.
  In case we ever need to set environment variables across steps again, there are better ways to do that by now (env files).

- Autobuild: Rename RELEASE_VERSION_NAME -> BUILD_VERSION
  This renames the RELEASE_VERSION_NAME (and all calculation in analyse_git_reference.py beforehand) to BUILD_VERSION instead. The variable does not contain a name (e.g. "Jamulus"), so that's confusing. It isn't solely used for releases either, so BUILD_ seems like a better fit.

- Autobuild: Rename jamulus_buildversionstring -> JAMULUS_BUILD_VERSION
  - "string" is redundant
  - use uppercase as all other environment variables do
  - using BUILD_VERSION is consistent with the source of the variable (create-release step)
  - prefixing with JAMULUS_ is helpful as other versions are in play in the Workflow as well.

- Autobuild: Windows: Rename the artifact instead of copying
  This is more consistent with the other platforms and might be slightly faster.

- Autobuild: Windows: Use Qt version variables instead of hardcoding

- Autobuild: Linux: Silence apt questions

- Autobuild: Improve dependency installation log messages

- Autobuild: Mark local variables as local

- Autobuild: Workflow: Unify id naming scheme

- Autobuild: Shorten artifact_deploy_filename variable name

CHANGELOG: Autobuild: Simplified build matrix and cleaned up redundant functionality
(also to be condensed with other autobuild PRs)

**Context: Fixes an issue?**

Related: #2503

No.

Ready for review.

Reviews.

## Checklist

- [x] I tested my code and it does what I want:
  - [x] [Build is green](https://github.com/hoffie/jamulus/actions/runs/2027994842)
  - [x] All artifacts are there
  - [x] Artifact size is unchanged compared to latest [`master` run](https://github.com/jamulussoftware/jamulus/actions/runs/2022402481)

### DIFF
--- a/.github/actions_scripts/analyse_git_reference.py
+++ b/.github/actions_scripts/analyse_git_reference.py
@@ -43,7 +43,7 @@ def write_changelog(version):
         f.write(changelog)
 
 
-def get_release_version_name(jamulus_pro_version):
+def get_build_version(jamulus_pro_version):
     if "dev" in jamulus_pro_version:
         name = "{}-{}".format(jamulus_pro_version, get_git_hash())
         print("building an intermediate version: ", name)
@@ -61,14 +61,14 @@ def set_github_variable(varname, varval):
 
 jamulus_pro_version = get_version_from_jamulus_pro()
 write_changelog(jamulus_pro_version)
-release_version_name = get_release_version_name(jamulus_pro_version)
+build_version = get_build_version(jamulus_pro_version)
 
 fullref = os.environ['GITHUB_REF']
 publish_to_release = bool(re.match(r'^refs/tags/r\d+_\d+_\d+\S*$', fullref))
 
-# RELEASE_VERSION_NAME is required for all builds including branch pushes
+# BUILD_VERSION is required for all builds including branch pushes
 # and PRs:
-set_github_variable("RELEASE_VERSION_NAME", release_version_name)
+set_github_variable("BUILD_VERSION", build_version)
 
 # PUBLISH_TO_RELEASE is always required as the workflow decides about further
 # steps based on this. It will only be true for tag pushes with a tag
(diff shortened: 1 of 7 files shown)


pljones (approved):

ann0see (approved):
Looks good - and CI is green.

Issue #2519, opened 2022-03-18
Connection status client and gui can get out of sync.


pgScorpio:
**Describe the bug**

There are several occasions where Client connection and GUI can get out of sync.
If this is the case the gui will show connected state (and disconnect button) while still connected or the gui will show disconnected state (and connect button) while not connected.
Also there are cases where GUI shows the "Disconnect" button, but opens the "Connect" dialog when pressed.

**To Reproduce**

This mostly happens when you change to an invalid device while connected.
It also happens in other cases of device errors but these are harder to reproduce.

**Expected behavior**

GUI should always show the correct state.

**Screenshots**

**Operating system**

**Version of Jamulus**
All Jamulus versions

**Additional context**

I finally found the problem (after a long time), And there are two causes:
1: Connect/Disconnect is implemented in the CClientDlg code, with a lot of calls to CClient (Doesn't belong here, since this is not GUI functionality.)
2: Some functions in CCLient have ambiguous names, so wrong functions are used to check if connected (actually checking Sound.IsStarted)

This issue will be solved in my "sound re-design" repo.

ann0see:
Which we also see in https://github.com/jamulussoftware/jamulus/pull/3249

I hope we'll get to a better design in Jamulus 4

pljones:
@ann0see @mcfnord would the Connect refactor address this?  If so, please link.  Dropping from 4.0.0, though.

mcfnord:
Root cause (static analysis, per the request in #3804). This confirms @pgScorpio's own diagnosis and pins it to the exact lines.

The GUI's notion of "connected" is `CClient::IsRunning()`, which is literally:

```cpp
bool IsRunning() { return Sound.IsRunning(); }
```

[...]

PR #2541, opened 2022-03-23
issue #2538: Adding global messagebox and commandline parameter parsing classes.

pgScorpio:
**Short description of changes**

Added global messagebox and commandline parameter parsing classes to global.h/main.c
Adapted all messageboxes to use the new CMsgBoxes class.

CHANGELOG: 

**Context: Fixes an issue?**

General improvement and preparation for sound-redesign.

**Does this change need documentation? What needs to be documented and how?**

**Status of this Pull Request**

Working...

**What is missing until this pull request can be merged?**

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [ ] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

Issue #2538, opened 2022-03-22
Adding global messagebox and commandline parameter parsing classes.


pgScorpio:
**What is the current behaviour and why should it be changed?**

Currently messageboxes for errors, warnings and info are created all over the code. Disadvantage is that they don't have a parent dialog and are inconsistent in message title.

Also commandline parameters are now parsed by several functions in main and passed via-via through several constructor parameters. A class accessible anywhere in the code would be more efficient and more versatile.

**Describe possible approaches**

All this is in preparation of the "sound-redesign"

A static messagebox class will provide messageboxes that can be opened anywhere in the code, but are always child of CClientdlg and will have a consistant title (including any clientname).

The commandline parsing functions in main will be moved to a (static) class that is accessible anywhere in the code so commandline parameters can be retrieved anywhere in the code, no need to pass then via-via through constructors. 
Also we will be able to use "special" commandline parameters (normally rejected) for debugging and testing purposes.

**Has this feature been discussed and generally agreed?**

The commandline parsing class was already suggested in the comments in the code.

ann0see:
Great! Thanks for opening an issue. I think it is possible to have a PR with only the class and some mi or code changes?

pgScorpio:
@ann0see 

> I think it is possible to have a PR with only the class and some mi or code changes?

As a matter of fact this can be implemented with only changes in global.h and main.cpp.
It does not have to be actually used anywhere else in the code yet... 
Though the message boxes can already be used on several places, that could be another PR. Even someone else could pick up changing all message boxes to the new implementation.)

The no longer needed (nor wanted) commandine parameter passing mainly affects CSound so I will make these changes automatically in following PR's.

ann0see:
I think everything in the scope of this issue can be put into one PR. 
Other people should comment, but I‘d be fine with not splitting the error messages out. 
I‘d not be OK with big visible UI changes or sound related changes in one big PR if not absolutely necessary.

pljones:
> Also commandline parameters are now parsed by several functions in main and passed via-via through several constructor parameters. A class accessible anywhere in the code would be more efficient and more versatile.

I'm working on that.

pgScorpio:
@pljones 

> I'm working on that.

I already have it in my own jamulus repo and I would like to push it but, again, I don't know how to open a PR for this issue and push it to jamulus master.

@ann0see (Hate to say it, but I'm getting really tired of all those github related issues.)

ann0see:
[...]

PR #2550, opened 2022-03-25
Connection status issue #2519

pgScorpio:
In the first stage I just renamed some functions in CClient to make the problem clear.
The second stage should already solve this issue.

First Stage: Make the cause of the problem clear.

Second Stage: 
Moved Connect/Disconnect code from CClientdlg to CClient.
Now using the proper connected checks in several places.
Added bDisconnectAndDisable to CChannel. (For a Client now
Channel.Disconnect() will block audio data and auto disable the channel on disconnected)

CHANGELOG: 
Fix for issue #2519

**Context: Fixes an issue?**

fixes issue #2519

No documentation changes.

## Checklist

- [x ] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x ] I tested my code and it does what I want
- [x ] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [x ] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x ] I've filled all the content above


src/client.h
```diff
@@ -119,10 +119,18 @@ class CClient : public QObject
 
     virtual ~CClient();
 
-    void Start();
-    void Stop();
-    bool IsRunning() { return Sound.IsRunning(); }
-    bool IsCallbackEntered() const { return Sound.IsCallbackEntered(); }
+    void StartConnection(); // ---> pgScorpio: Was Start(), but Start what ? ( Should be Connect() ?)
```
pljones, line 122: Staart the _client_.  It does more than just connection.

src/client.h
```diff
@@ -119,10 +119,18 @@ class CClient : public QObject
...
+    void StopConnection();  // ---> pgScorpio: Was Stop(), but Stop what ?   ( Should be Disconnect() ?)
```
pljones, line 123: Stop the _client_.  It does more than just disconnection.

src/clientdlg.cpp
```diff
@@ -1190,16 +1191,18 @@ void CClientDlg::OnCLPingTimeWithNumClientsReceived ( CHostAddress InetAddr, int
...
+
     // set address and check if address is valid
     if ( pClient->SetServerAddr ( strSelectedAddress ) )
     {
         // try to start client, if error occurred, do not go in
         // running state but show error message
         try
         {
-            if ( !pClient->IsRunning() )
+            if ( !pClient->SoundIsStarted() ) // ---> pgScorpio: Again this is NOT connection started !() )
```
pljones, line 1203: No, it's whether the client is running.

pgScorpio: Exactly what I mean to say! This function does NOT indicate if we are connected to a server, but on;ly if Sound is Started.
And Sound started does say nothing about an connection to the server, since Sound.Start() can have failed (No check on that) but the connection still can be started with Channel.Enable(true) !
This is exactly where it goes wrong with an invalid device selected...
Sound is NOT started, but the Channel is enabled !

pgScorpio:
And auto-build failed on Android, Linux, macOS and iOS ?
I should have to look into this, but that's exactly why I want the Sound-redesign ! 
Nothing changed in any CSound, but still build errors on other OS builds ??? This should never happen...

pgScorpio:
And again the code style check failed, while I have ran `make clang_format` and again the result of the auto-build clang-format is NOT according the code style!

I'm really considering stopping my contribution to jamulus at all. Too much frustration!

ann0see:
Maybe leave the clang format alone until it’s close to finished?

ann0see:
> I'm really considering stopping my contribution to jamulus at all. Too much frustration!

If nobody notices and reports the Problems they can’t be fixed, so yes, it’s painful to start but has a high value.

src/clientdlg.cpp
pljones: It's meant to indicate whether the client has _started_.  Not the state of the sound system.  Not the state of the network connection.  But whether the client has started.

pgScorpio: > It's meant to indicate whether the client has _started_. Not the state of the sound system. Not the state of the network connection. But whether the client has started.

I Think we mean the same, but use different vocabulary ;=))
And yes, I am very persistent in using variable- and function names that clearly state what they do (And unfortunately I encounter a lot of them in jamulus which absolutely don't, some of them even very misleading.)

So I think the Client is already **started** when you run jamulus (without a -s ;=))

It's about if the connection to the server is started, and the original IsRunning() (Used at many places) had nothing to do with the connection to the server, but with Sound. So if Sound.Start() fails (and there is no check on that!) they all think there is no connection to the server, but in reality we actually are connected/connecting to the server, so UI and reality don't match and you can't disconnect, since hitting the connect button will try to start a connection again while we are already connecting.

pgScorpio:
@ann0see 

> Did you see the error? Cross platform development is probably really hard.

Now I do see it! Thanks!. Somehow missed that one in the search and replace :=((, 
I Should have tested it on at least on macOS and Linux (and also headless, which isn't even in the Qt Creator builds, so I had to add it), but it is corrected now.

And Cross platform development Isn't that hard with the propper design and tooling. I've done it for many years, And as a matter of fact it always has been one of my specialties, but unfortunately Qt has little to no support for it and Jamulus is also not really well designed for multi platform either (it clearly has a history), hence my redesign proposal, but It will take a while till we are getting there. As I see it Sound is just the first, most obvious, step....

src/clientdlg.cpp
pljones: No, the Client is definitely _not_ started when you run Jamulus.  It's not even created then.

pgScorpio: > No, the Client is definitely _not_ started when you run Jamulus. It's not even created then.

I meant It is started automatically when you run jamulus  (without a -s ;=)). (Created in main and started with `pApp->exec();`)

As said we are apparently using different vocabulary...
Maybe you call " Started" what I call "Connected" (or "connecting" ?), since I think the client is already started if you see it's user interface and can interact with it. Though it's just not yet connected to a server, it's still "Running" in some way. 

Simple verbs like " Start" and " Stop" are often ambiguous, and adverbs like " Running"  and " Connected" are often even more ambiguous and often interpreted differently depending on the context. 
i.e. When are you "Driving" a car ?, As soon as you take a seat behind the wheel? (Legally seen yes), When you "Started" the engine? (You can drive downhill without even starting the engine), or when your car is really moving ? (If so, also when it's on the back of a truck?, or being towed?) 

So to come back on topic: " Started" does not mean "Running", nor "Connected". And **not "Connected"** is not the same as **not "Running"** (might be "Connecting", so " Started"  but not yet " Connected" ). An easy confusion that can cause a lot of, often hard to find, bugs, like in this case.

But no offence, I'm the last one to say that writing clear code is easy. Even after so many years I'm still struggling almost daily with good, clear, naming's.

ann0see: Don’t quote me on that – since my knowledge of the code is basic – but I think "the client" is not the UI but the CClient class. Maybe talking about the actual class names makes it easier to avoid confusion. 

Also: I wouldn’t write your name next to comments since they might be updated/… in future by someone else who isn’t you. Having you in the git history should clearly show who wrote what if this is actually needed.

src/global.h
```diff
@@ -362,15 +362,214 @@ class CCustomEvent : public QEvent
...
+#define CMDLN_STARTMINIMIZED      "-z",                    "--startminimized"
+#define CMDLN_CONNECT             "-c",                    "--connect"
+#define CMDLN_NOJACKCONNECT       "-j",                    "--nojackconnect"
+#define CMDLN_MUTESTREAM          "-M",                    "--mutestream"
+#define CMDLN_MUTEMYOWN           "--mutemyown",           "--mutemyown"
+#define CMDLN_CLIENTNAME          "--clientname",          "--clientname"
+#define CMDLN_CTRLMIDICH          "--ctrlmidich",          "--ctrlmidich"
+// Backwards compatibilyty:
+#define CMDLN_CENTRALSERVER       "--centralserver",       "--centralserver"
+// pgScorpio: TODO These are NOT in help !:
```
ann0see, line 568: I remember that they are mostly used for debugging/advanced use. Hence it was agreed not to document them.

pgScorpio: @ann0see 

But these options are documented on the [wiki](https://jamulus.io/wiki/Server-Troubleshooting)  and the [man page](https://manpages.debian.org/testing/jamulus/Jamulus.1.en.html)

So should I change this comment to "Debug options: (not in help)" ?

Also I think I'm messing things up, since this is a change from issue #2538. 
And I only just now realize what you ment with "although merging from your master branch to our master branch is discouraged". I didn't branch this one ! So now, after my rebasing these changes also appeared in my branches i guess?
(Sorry, still learning...)

ann0see: If "show all servers" and "analszer console" are really documented somewhere, something I wasn’t aware of changed (@gilgongo do you know anything?).


If you rebase your master branch, sure this will show up in the PR from master (but shouldn’t on this one).

My offer still stands: we can have a short call if you need it).

gilgongo: Yes, the **use** of "show all servers" is mentioned as a way of debugging a problem. But "analyzerconsole" got added to the man page, because... the man page plays by its own rules.

src/channel.h
```diff
@@ -76,7 +76,7 @@ class CChannel : public QObject
 
     void PrepAndSendPacket ( CHighPrioSocket* pSocket, const CVector<uint8_t>& vecbyNPacket, const int iNPacketLen );
 
-    void ResetTimeOutCounter() { iConTimeOut = iConTimeOutStartVal; }
+    void ResetTimeOutCounter() { iConTimeOut = bDisconnectAndDisable ? 1 : iConTimeOutStartVal; }
```
ann0see, line 79: How often do we use short hand if in the source?

src/clientdlg.cpp
```diff
@@ -1270,11 +1213,7 @@ void CClientDlg::Disconnect()
     TimerDetectFeedback.stop();
     bDetectFeedback = false;
 
-    // clang-format off
-// TODO is this still required???
-// immediately update status bar
-OnTimerStatus();
```
ann0see, line 1276: Could you please briefly explain your ideas/changes?

src/global.h
```diff
@@ -362,15 +362,214 @@ class CCustomEvent : public QEvent
...
-                          QString strLongOpt,
-                          double  rRangeStart,
-                          double  rRangeStop,
-                          double& rValue );
+//============================================================================
+// CMsgBoxes class:
+//  Use this static class to show basic Error, Warning and Info messageboxes
+//  For own created message boxes you should still use
+//    CMsgBoxes::MainForm() and CMsgBoxes::MainFormName()
+//============================================================================
```
ann0see, line 370: Is this comment syntax used throughout the code - except for ASIO? We should remain consistent.

src/global.h
```diff
@@ -362,15 +362,214 @@ class CCustomEvent : public QEvent
...
+        {
+            if ( GetStringArgument ( i, strShortOpt, strLongOpt, strArg ) )
+            {
+                return true;
+            }
+
+            i++;
+        }
+
+        return false;
```
ann0see, line 470: Looks a bit like we could use a for loop?

src/main.cpp
```diff
@@ -55,8 +55,88 @@ extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
 
 // Implementation **************************************************************
 
+int    CCommandlineOptions::appArgc = 0;
+char** CCommandlineOptions::appArgv = NULL;
+
+QDialog* CMsgBoxes::pMainForm       = NULL;
+QString  CMsgBoxes::strMainFormName = APP_NAME;
+
+QString UsageArguments ( char** argv )
```
ann0see, line 64: Yeah, looks a lot better ;-).

src/main.cpp
```diff
@@ -78,6 +158,7 @@ int main ( int argc, char** argv )
 #else
     bool bIsClient = true;
 #endif
+    bool         bSpecialOptions             = false; // Any options after this option will be accepted ! (mostly used for debugging purpouses)
```
ann0see, line 161: What's that exactly? Is it a new debugging feature?

src/main.cpp
```diff
@@ -148,7 +235,7 @@ int main ( int argc, char** argv )
         // Common options:
 
         // Initialization file -------------------------------------------------
-        if ( GetStringArgument ( argc, argv, i, "-i", "--inifile", strArgument ) )
+        if ( cmdLine.GetStringArgument ( i, CMDLN_INIFILE, strArgument ) )
```
ann0see, line 238: cmdLine sounds a bit strange and somewhat not consistent to the style - however, I don't know what the current style would want here.

src/main.cpp
```diff
@@ -175,26 +262,8 @@ int main ( int argc, char** argv )
...
-
-        // JSON-RPC secret file name -------------------------------------------
-        if ( GetStringArgument ( argc, argv, i, "--jsonrpcsecretfile", "--jsonrpcsecretfile", strArgument ) )
-        {
-            strJsonRpcSecretFileName = strArgument;
-            qInfo() << qUtf8Printable ( QString ( "- JSON-RPC secret file: %1" ).arg ( strJsonRpcSecretFileName ) );
-            CommandLineOptions << "--jsonrpcsecretfile";
-            continue;
-        }
-
```
ann0see, line 195: Why did that get removed?

src/main.cpp
```diff
@@ -556,14 +575,27 @@ int main ( int argc, char** argv )
             continue;
         }
 
+        // Enable Special Options ----------------------------------------------
+        if ( cmdLine.GetFlagArgument ( i, CMDLN_SPECIAL ) )
```
ann0see, line 579: After thinking a bit more about the new style: I think it would still be good to see the options in the code like before. But I agree that the current way is not good since it leads to a lot of unneeded code. 

Something like a function getting all arguments from the CLI and returning an array of enabled options (or similar) which we loop through to set the options would be another option.

ann0see (commented):
Can't comment much on the code - that should be up to @softins and @pljones but see some comments. How mergable is this PR at the moment?

src/clientdlg.cpp
```diff
@@ -1141,31 +1127,17 @@ void CClientDlg::OnTimerCheckAudioDeviceOk()
     // timeout to check if a valid device is selected and if we do not have
     // fundamental settings errors (in which case the GUI would only show that
     // it is trying to connect the server which does not help to solve the problem (#129))
-    if ( !pClient->IsCallbackEntered() )
+    if ( !pClient->SoundIsRunning() ) // ---> pgScorpio Was pClient->IsCallbackEntered()
```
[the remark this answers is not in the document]
hoffie: I don't think we should put naming history into comments at all. That'll be rather something for the git history, IMO?

hoffie:
- The PR title suggests that this PR is related to a connection status issue. There are logic adjustments in that area (`bDisconnectAndDisable`?). At the same time, the PR description says "No actual code changes". This doesn't match up for me. I guess that's where the work started, but it doesn't appropriately describe the changes in this PR, IMO. :)
- The PR contains a rework of command line flag processing. I don't think that's related? I think we should get better there, but please keep in mind that @pljones has already been working on some kind of refactoring in that area, IIRC.
- The PR contains dialog box refactoring. I don't see how this is related either. I'm not opposed to it at all (but haven't checked more deeply either). I just believe it doesn't belong here.

For my taste, that should be three (or more?) individual PRs, some with explicit Discussions/Issue ahead of time.

src/channel.h
pgScorpio: > How often do we use short hand if in the source?

You mean conditional assignment ?
It is used at several places.
I use it whenever it makes the code more readable... (at least to me ;=))

ann0see: > I use it whenever it makes the code more readable

I usually think it makes code less readable ;-). But in this case I think it's ok - but that depends on if this is used elsewhere too.

src/client.cpp
```diff
@@ -819,50 +803,119 @@ void CClient::OnClientIDReceived ( int iChanID )
...
-    Sound.Start();
+    return false;
 }
 
-void CClient::Stop()
+bool CClient::Disconnect()
+{
+    if ( Channel.IsEnabled() )
+    {
+        /*
```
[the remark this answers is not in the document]
pgScorpio: @ann0see 
> Why is this commented? Could you please remove debug code - if we want to merge this?

What do you mean exactly ?
Most of the comment was already there, I adjusted to the new situation, but I see there is still some comment from stage 1, no longer applicable for stage 2 
I will remove those.
And what "debug" code ? A check on failure of disconnect ? That's essential code to me....

src/clientdlg.cpp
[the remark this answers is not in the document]
pgScorpio: > I don't think we should put naming history into comments at all. That'll be rather something for the git history, IMO?

ann0see already pointed me to that on. I will remove those.

src/clientdlg.cpp
pgScorpio: > Could you please briefly explain your ideas/changes?

I like to use clear function names that are telling what they do...

The next line now calls UpdateSettingsAndChatButtons() which is now also used in OnTimerStatus() and at multiple other locations, instead of calling OnTimerStatus(), or repeating the same code, at multiple locations.

Since the name OnTimerStatus() tells exactly when it's called (Good name for a callback) this suggests that it's used for one thing only and does not give any clue it's used elsewhere too, so future changes to OnTimerStatus() could have unwanted side effects.

src/global.h
pgScorpio: > Is this comment syntax used throughout the code - except for ASIO? We should remain consistent.

AFAIK it was already used at some places (and also in many libraries), but various other styles are used too (so not very consistent anyway)...

Personally I find this the most readable style, especially when quick scrolling through a file. But maybe that is because this was the mandatory style I had to use for many, many years, so it became a habit ;=) 
 
I can adapt if you want, but forgive me if I frequently fall back into those habits ;=)

src/global.h
pgScorpio: > Looks a bit like we could use a for loop?

Yes could just as well be a for loop...

src/main.cpp
pgScorpio: > What's that exactly? Is it a new debugging feature?

Yes since we now can access the command line parameters anywhere in the code those are also very useful for debugging and testing purposes too. But since there is a check in main which would accept only valid, known parameters I introduced the "--special" option to allow any option for debugging and test purposes.

src/main.cpp
pgScorpio: > cmdLine sounds a bit strange and somewhat not consistent to the style - however, I don't know what the current style would want here.

Getting an argument from the Command line.. so cmdLine seams logical to me...
And consistent to what style ? Unfortunately I didn't find a consistent naming style in the code yet...

src/main.cpp
pgScorpio: Good point !!
I have no idea how that happened !

src/main.cpp
pgScorpio: Please explain further...
The code isn't that different from how it was, except for the defines for the short + long names. Is that what you have a problem with ? 
Personally I think it makes the code a lot more readable, and it has another reason... since it is no longer necessary to pass command line parameters as variable true a lot of constructors, this could mean the long and short names could spread throughout the code and that didn't seem a good idea to me, so I use defines instead.

And I think there is hardly any more code, it's just in another place.

pgScorpio:
@hoffie 

I completely understand your confusion, but that's because I terribly messed up some things by opening a PR from my master (Sorry!!). This causing PR's, my master updates and branch rebasing to mingle up. (I hope we can still correct this)

> * The PR title suggests that this PR is related to a connection status issue. There are logic adjustments in that area (`bDisconnectAndDisable`?). At the same time, the PR description says "No actual code changes". This doesn't match up for me. I guess that's where the work started, but it doesn't appropriately describe the changes in this PR, IMO. :)

"No actual code change" was the " first stage" (Just renaming functions to make the problem clear), in the mean time the second stage is committed too (solving the problem, so actually changing the code). Probably this should have been a new PR? I just don't know...

> * The PR contains a rework of command line flag processing. I don't think that's related? I think we should get better there, but please keep in mind that @pljones has already been working on some kind of refactoring in that area, IIRC.
> * The PR contains dialog box refactoring. I don't see how this is related either. I'm not opposed to it at all (but haven't checked more deeply either). I just believe it doesn't belong here.

Yes those are from PR #2541 "Adding global messagebox and commandline parameter parsing classes." that was pushed from my master (Sorry again!!)

> For my taste, that should be three (or more?) individual PRs, some with explicit Discussions/Issue ahead of time.

So as a matter of fact it are two PR's, both with multiple commits, and updates of my own master also seem to appear here now, so I understand all is getting really unclear (also for me) and I hope we can find a way to solve this problem.

Probably it's better to solve my git problems first and than completely start over again, since I guess I'm already spending more time in solving all kind of problems then I did spent on the coding this PR itself.

src/channel.h
pljones: Where's it's a simple "truth" test to pick between two values based on another value, it's okay, in general, I'd say.  (I don't think Volker was keen on it.)  Anything else (e.g. an expression getting evaluated as any part of the `?:` expression).

ann0see:
Closing as stale. If there's further development, please re-open

ann0see:
If I find some time, I'd like to look into this again.

ann0see (changes requested):
I think that we should recreate this PR from scratch.

.gitignore
ann0see, line 1: No changes to be made

src/channel.cpp
```diff
@@ -506,11 +508,20 @@ void CChannel::Disconnect()
     // we only have to disconnect the channel if it is actually connected
     if ( IsConnected() )
     {
+        // for a Client we will block further audio data and disable the channel as soon as disconnected;
+        bDisconnectAndDisable = !bIsServer;
```
ann0see, line 512: Reasoning to be investigated.

src/channel.h
ann0see, line 79: To be investigated. May need comment.

src/chatdlg.cpp
```diff
@@ -140,8 +140,8 @@ void CChatDlg::OnAnchorClicked ( const QUrl& Url )
     // only allow http(s) URLs to be opened in an external browser
     if ( Url.scheme() == QLatin1String ( "https" ) || Url.scheme() == QLatin1String ( "http" ) )
     {
-        if ( QMessageBox::question ( this,
-                                     APP_NAME,
+        if ( QMessageBox::question ( CMsgBoxes::MainForm(),
```
ann0see, line 143: Not part of this issue. In recreation -> don't include

src/client.cpp
```diff
@@ -181,13 +180,6 @@ CClient::CClient ( const quint16  iPortNumber,
     // start the socket (it is important to start the socket after all
     // initializations and connections)
     Socket.Start();
-
```
ann0see, line 184: In depth code review needed to understand changes.

src/client.cpp
```diff
@@ -715,8 +694,13 @@ void CClient::OnSndCrdReinitRequest ( int iSndCrdResetType )
     }
     MutexDriverReinit.unlock();
 
+    if ( !strError.isEmpty() )
```
ann0see, line 697: Investigate if related

src/clientdlg.cpp
```diff
@@ -1059,7 +1044,8 @@ void CClientDlg::OnTimerSigMet()
         // show message box about feedback issue
         QCheckBox* chb = new QCheckBox ( tr ( "Enable feedback detection" ) );
         chb->setCheckState ( pSettings->bEnableFeedbackDetection ? Qt::Checked : Qt::Unchecked );
-        QMessageBox msgbox;
+        QMessageBox msgbox ( CMsgBoxes::MainForm() );
```
ann0see, line 1047: Not part of the issue.

src/clientdlg.cpp
```diff
@@ -1141,31 +1127,17 @@ void CClientDlg::OnTimerCheckAudioDeviceOk()
...
     // fundamental settings errors (in which case the GUI would only show that
     // it is trying to connect the server which does not help to solve the problem (#129))
-    if ( !pClient->IsCallbackEntered() )
+    if ( !pClient->SoundIsRunning() ) // ---> pgScorpio Was pClient->IsCallbackEntered()
     {
-        QMessageBox::warning ( this,
-                               APP_NAME,
-                               tr ( "Your sound card is not working correctly. "
-                                    "Please open the settings dialog and check the device selection and the driver settings." ) );
+        CMsgBoxes::ShowWarning ( tr ( "Your sound card is not working correctly. "
```
ann0see, line 1132: Not related.

src/clientdlg.cpp
```diff
@@ -1141,31 +1127,17 @@ void CClientDlg::OnTimerCheckAudioDeviceOk()
...
+                                      "Please open the settings dialog and check the device selection and the driver settings." ) );
     }
 }
 
 void CClientDlg::OnTimerDetectFeedback() { bDetectFeedback = false; }
 
-void CClientDlg::OnSoundDeviceChanged ( QString strError )
+void CClientDlg::OnSoundDeviceChanged()
 {
-    if ( !strError.isEmpty() )
```
ann0see, line 1157: Check preconditions. Refactored method makes sense.

src/global.h
```diff
@@ -359,18 +359,220 @@ class CCustomEvent : public QEvent
 };
 
 /* Prototypes for global functions ********************************************/
-// command line parsing, TODO do not declare functions globally but in a class
-QString UsageArguments ( char** argv );
+extern QString UsageArguments ( char** argv );
```
ann0see, line 362: Unrelated.

src/serverdlg.cpp
```diff
@@ -571,12 +571,10 @@ void CServerDlg::OnStopRecorder()
     UpdateRecorderStatus ( QString() );
     if ( pServer->GetRecorderErrMsg() != QString() )
     {
-        QMessageBox::warning ( this,
-                               APP_NAME,
```
ann0see, line 575: Unrelated.

src/main.cpp
ann0see, line 1: Probably everything unrelated

src/soundbase.cpp
ann0see, line 1: Unrelated

src/util.cpp
```diff
@@ -495,6 +495,7 @@ CAboutDlg::CAboutDlg ( QWidget* parent ) : CBaseDlg ( parent )
         "<p>RobyDati (<a href=\"https://github.com/RobyDati\">RobyDati</a>)</p>"
         "<p>Rob-NY (<a href=\"https://github.com/Rob-NY\">Rob-NY</a>)</p>"
         "<p>Thai Pangsakulyanont (<a href=\"https://github.com/dtinth\">dtinth</a>)</p>"
+        "<p>Peter Goderie (<a href=\"https://github.com/pgScorpio\">pgScorpio</a>)</p>"
```
ann0see, line 498: Add in separate commit

src/util.cpp
```diff
@@ -666,7 +667,9 @@ void CLanguageComboBox::OnLanguageActivated ( int iLanguageIdx )
     // only update if the language selection is different from the current selected language
     if ( iIdxSelectedLanguage != iLanguageIdx )
     {
-        QMessageBox::information ( this, tr ( "Restart Required" ), tr ( "Please restart the application for the language change to take effect." ) );
```
ann0see, line 669: Unrelated

src/util.h
ann0see, line 1: Unrelated

windows/sound.cpp
ann0see, line 1: Unrelated

ann0see (commented):
Just adding personal comments to sort this.

ann0see:
@pgScorpio I've now opened https://github.com/jamulussoftware/jamulus/pull/3364 Please continue with the other PR.

pgScorpio:
> @pgScorpio I've now opened #3364 Please continue with the other PR.

Sorry, but this was years Ago!

In the mean time, due to health issues and other priorities, I totally stopped programming (and even using Jamules).

ann0see:
Ok. No problem. I hope the team will be able to pick up the changes.

ann0see:
Best wishes for your health.

Issue #2555, opened 2022-03-27
SIGINT handler for Linux headless server does not unregister


pljones:
**Describe the bug**
Stopping the headless Server should unregister from any Directory with which it is registered.  If this is not done, a "ghost entry" remains in the directory list (visible when a client runs with `--showallservers` or in Jamulus Explorer).

Current behaviour is that the headless Server does _not_ unregister.

**To Reproduce**
```
git checkout master
git pull upstream master
make distclean; qmake && make -j3
./Jamulus -s -n -p <openport> -e <Directory server address>
(check the server is visible and operational)
```
![image](https://user-images.githubusercontent.com/1549463/160280000-ba7e615a-ae1f-4213-b7f6-ce98e1b8e7c2.png)
And this shows the log for Jamulus Explorer successfully retrieving the welcome message:
```
 *** Copyright (C) 2005-2022 The Jamulus Development Team

Server Registration Status update: Registration requested
Server Registration Status update: Registered
2022-03-27 12:44:37, 80.85.86.7, connected (1)
2022-03-27 12:44:37,, server idling -------------------------------------
```
Now shut down the server with Ctrl-C:
```
^COnHandledSignal: 2
```
The signal is _handled_ but the Server doesn't send an unregister request, leaving a ghost entry:
![image](https://user-images.githubusercontent.com/1549463/160280079-7f68158f-69b6-44f9-8aa4-2c982fa7dc87.png)
Note the "Ping", "Version" and "OS" are not now listed.

**Expected behavior**
Stopping the headless server should send an unregister request.  This _does_ happen when using the UI:

Server:
```
./Jamulus -s -p <openport> -e <Directory server address>
...
Server Registration Status update: Registration requested
Server Registration Status update: Registered
...
Server Registration Status update: Not registered
```
Directory:
```
Mar 27 12:51:34 fs-peter Jamulus-Directory2[646838]: Requested to register entry for 192.168.1.19:56850 (81.174.155.168:56850): (private)drealm.info
...
Mar 27 12:52:45 fs-peter Jamulus-Directory2[646838]: Requested to unregister entry for 192.168.1.19:56850
```

**Operating system**
Tested on Linux.

**Version of Jamulus**
jamulussoftware/jamulus@56092960

hoffie:
I don't see this issue on 3.8.2. I do see this issue on `master`. Therefore, sounds like a regression.

pljones:
[...]

PR #2556, opened 2022-03-27
#2555 Headless Server should unregister

pljones:
When SIGINT or SIGTERM is received, the headless Server initiates shutdown.  However, this did not previously include unregistering.  This change adds the call to the unregister method.

CHANGELOG: Headless Server should unregister on shutdown (regression in feature/server-list-persistence-ui)

**Context: Fixes an issue?**
Fix for #2555 

No, this should be existing behaviour but isn't.

Tested locally:
```
Server Registration Status update: Registration requested
Server Registration Status update: Registered
2022-03-27 13:39:48, 80.85.86.7, connected (1)
2022-03-27 13:39:48,, server idling -------------------------------------
^COnHandledSignal: 2
Server Registration Status update: Not registered
```

Review.

## Checklist

### DIFF
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -848,6 +848,17 @@ void CServerListManager::OnTimerCLRegisterServerResp()
     }
 }
 
+void CServerListManager::OnAboutToQuit()
+{
+    {
+        QMutexLocker locker ( &Mutex );
+        Save();
+    }
+
+    // Sets the lock - also needs to come after Save()
+    Unregister();
+}
+
 void CServerListManager::SetRegistered ( const bool bIsRegister )
 {
     // we need the lock since the user might change the server properties at
--- a/src/serverlist.h
+++ b/src/serverlist.h
@@ -229,11 +229,7 @@ public slots:
     void OnTimerCLRegisterServerResp();
     void OnTimerIsPermanent() { ServerList[0].bPermanentOnline = true; }
 
-    void OnAboutToQuit()
-    {
-        QMutexLocker locker ( &Mutex );
-        Save();
-    }
+    void OnAboutToQuit();
 
 signals:
     void SvrRegStatusChanged();


src/serverlist.cpp
```diff
@@ -848,6 +848,15 @@ void CServerListManager::OnTimerCLRegisterServerResp()
     }
 }
 
+void CServerListManager::OnAboutToQuit()
+{
+    // Sets the lock
```
ann0see, line 853: Is this comment correct here? Doesn't the lock get set below?

ann0see (commented):
Seems like a trivial change

ann0see:
BTW: I think GitHub needs the `Fixes: #issue` syntax to automatically link an issue.

src/serverlist.cpp
pljones: It's a comment about the line that immediately follows it.  `Unregister()` sets the lock.  It's _why_ the lock isn't taken earlier.

pljones:
> BTW: I think GitHub needs the `Fixes: #issue` syntax to automatically link an issue.

Oh... is that in the template?  I might have deleted it before reading it :).

src/serverlist.cpp
ann0see: Ok. Wasn’t sure if Unregister() sets a lock. It’s a bit unintuitive from the name so it’s not obvious.

hoffie:
> > BTW: I think GitHub needs the `Fixes: #issue` syntax to automatically link an issue.
> 
> Oh... is that mentioned in the template? I might have deleted it before reading it :).

https://github.com/jamulussoftware/jamulus/blob/0bbd298d335ebc26791fb8f7aab304ce8f471fbc/.github/pull_request_template.md?plain=1#L11

hoffie:
> CHANGELOG: Headless Server should unregister on shutdown

As this sounds like a regression (3.8.2 worked correctly in my tests, but master doesn't), this should be listed under the registration refactoring in the changelog as it is no change between any released version of Jamulus.

hoffie (approved):

softins:
Should we understand the reason for the regression first? It looks like `server.cpp` used to call `ServerListManager.Unregister();` but this was removed in 1088c0a933b18c858ca826ccdd3046864a2518a4 that was part of pljones/feature/server-list-persistence-ui, merged on 23 Feb as 7d7b337355dce06c7a3b4df5f04ec1084980f695.

I can't tell whether that removal was a deliberate part of the server list changes, or something that got lost accidentally.

pljones:
> I can't tell whether that removal was a deliberate part of the server list changes, or something that got lost accidentally.

Hm... good find.  Given CServerList::OnAboutToQuit applies regardless of whether a Server is a Directory, I might have been thinking to move it there and it "got lost".  At least we know it used to happen and got removed...

pljones:
> > CHANGELOG: Headless Server should unregister on shutdown
> 
> As this sounds like a regression (3.8.2 worked correctly in my tests, but master doesn't), this should be listed under the registration refactoring in the changelog as it is no change between any released version of Jamulus.

Yep.  Is there a way to do that?

hoffie:
> > > CHANGELOG: Headless Server should unregister on shutdown
> > 
> > As this sounds like a regression (3.8.2 worked correctly in my tests, but master doesn't), this should be listed under the registration refactoring in the changelog as it is no change between any released version of Jamulus.
> 
> Yep. Is there a way to do that?

There's no extra syntax for that, no. It helps using the same text as for the other PR as the grouping will move those entries together so that a human can easily merge it. I wouldn't call out the other PR number in the entry here as it may confuse the script. Not sure if it's worth it to add logic to the script. Let's hope such cases remain seldom. :)

pljones:
I've tagged the change log entry with "regression in feature/server-list-persistence-ui" so it's clearly marked.

softins (commented):
The change looks sensible. In 3.8.2, the call to `Unregister()` was in `CServer::OnAboutToQuit()`, in this PR it is in `CServerListManager::OnAboutToQuit()`, which is arguably a more logical place anyway.

But as I was looking through it, I noticed that we have four different `OnAboutToQuit()` functions, all connected to `QCoreApplication::aboutToQuit`:
- `CServer::OnAboutToQuit()`
- `CSettings::OnAboutToQuit()`
- `CServerListManager::OnAboutToQuit()`
- `CJamRecorder::OnAboutToQuit()` (which is _also_ connected to `CJamController::EndRecorderThread`).

I don't know if there is any control over the order all the above handlers are executed, but it might be worth a sanity check that the order they are called doesn't actually matter.

Issue #2552, opened 2022-03-26
Investigate differences in clang format in CI and Windows 


ann0see:
**Describe the bug**
@pljones and @pgScorpio experienced unexpected clang format failures: see #2550 and #2551 
**To Reproduce**
?

**Expected behavior**
Clang format on Windows should format the same as on the CI. I expect some unknown incompatibility.

**Screenshots**

**Operating system**
Windows and Linux

**Version of Jamulus**
N/A
**Additional context**
@hoffie

pljones:
I'm running it on Linux, not Windows.

ann0see:
Ok. Thought you were developing on Windows too. 

https://github.com/jamulussoftware/jamulus/blob/56092960981a979ec0b73244f13c90b5d55d9695/.github/workflows/coding-style-check.yml#L20

It’s clang format 10 probably?

pljones:
I just tried running the clang-format check workflow on my repo:
https://github.com/pljones/jamulus/runs/5702476683?check_suite_focus=true
run against
https://github.com/pljones/jamulus/commit/b995489be5c48f1d0ee2c89faba7f052efbb5535

So it's weird...

pgScorpio:
I'm using clang-format Version 13.0.1 on Windows and macOS and Version 10.0.0-4ubuntu1 on Linux.
Both give the same result locally, but often auto-build does not have the same result.

1. auto-build clang-format breaks long lines too soon (config says ColomnLimit 150, but auto-build clang-format breaks sooner (just try a line of 149 chars long.).
2. auto-build clang-format does sometimes not align braces with `if` correctly. (See: [here](https://github.com/jamulussoftware/jamulus/runs/5683382273?check_suite_focus=true))
3. auto-build clang-format sometimes changes `type* variable` or `type *variable` to `type * variable`, but not consistently. (See [here](https://github.com/jamulussoftware/jamulus/pull/2407#issuecomment-1046261464))

ann0see:
Thanks for your comment. If I find time for it I’ll check what my clang-format does (but it should also be clang-format 10)

@passing knows clang format well, so he might know what goes wrong.

dcorson-ticino-com:
I had the same problem on my Qt6 work.
clang-format all OK locally on Win10, had errors when run on github.

hoffie:
I didn't have time to investigate all hints yet, but I mainly suspect discrepancies in clang-format versions to be the cause.

[...]

PR #2407, opened 2022-02-19
audio redesign

pgScorpio:
**Short description of changes**

Redesign of all sound handling for different platforms as discussed [here](https://github.com/jamulussoftware/jamulus/discussions/2400).

The user interface should be the same for all platforms as much as possible.
No sound related platform dependent code in CClient. (Also all interface related settings should eventually be handled in CSound).

**Context: Fixes an issue?**

Could also fix: #1378, #2344, #2238, #2496, #2519
Also better solution for: #2168, #2215

**Does this change need documentation? What needs to be documented and how?**

A new sound interface (CSoundBase) should be well documented as this PR evolves.

**Status of this Pull Request**

Work in progress.
Initial stage (Replacing ASIOSDK2 for Windows) could already be merged after reviewing and testing.
A lot of following steps still to be taken. See the [discussion](https://github.com/jamulussoftware/jamulus/discussions/2400).

**What is missing until this pull request can be merged?**

Merging should be done in steps, as will be discussed in this PR.

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [ ] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [ ] I've filled all the content above

PR #2559, opened 2022-03-28
Code: Use clang-format 13

hoffie:
- Coding style: Require clang-format 13
  We currently use clang-format 10 which has shown to produce unexpected output. This is annoying for PR submitters as the CI forces non-optimal code layout, especially when submitters use a later clang-format version which does not show this behavior.
  This commit changes CI to use clang-format 13 and updates relevant clang-format inline docs.

- Code: Run clang-format 13 on all source files

CHANGELOG: Code: Updated clang-format version requirement and CI check to v13 due to bugs in v10.

**Context: Fixes an issue?**

Related: #2552

Yes. This necessary documentation update of CONTRIBUTING.md has part of this PR.

Ready.

## Checklist

-  [x] I tested my code and it does what I want: [CI used v13](https://github.com/jamulussoftware/jamulus/runs/5716100782?check_suite_focus=true#step:4:9)
-  [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency): Heh...

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,12 +1,12 @@
 ---
-# Style for Clang-Format version 10
-# https://releases.llvm.org/10.0.0/tools/clang/docs/ClangFormatStyleOptions.html
+# Style for clang-format version 13
+# https://releases.llvm.org/13.0.0/tools/clang/docs/ClangFormatStyleOptions.html
 BasedOnStyle: LLVM
 AccessModifierOffset: -4
 AlignConsecutiveAssignments: true
 AlignConsecutiveDeclarations: true
 AlignConsecutiveMacros: true
-# Alignment of Consecutive Assignments/Declarations/Macros can be changed to `AcrossEmptyLinesAndComments` with Clang-Format version 12
+# Alignment of Consecutive Assignments/Declarations/Macros can be changed to `AcrossEmptyLinesAndComments` now that we require clang-format >=12
 #AlignConsecutiveAssignments: AcrossEmptyLinesAndComments
 #AlignConsecutiveDeclarations: AcrossEmptyLinesAndComments
 #AlignConsecutiveMacros: AcrossEmptyLinesAndComments
--- a/.github/workflows/coding-style-check.yml
+++ b/.github/workflows/coding-style-check.yml
@@ -9,15 +9,15 @@ on:
 
 jobs:
   clang-format:
-    name: Verify Clang-Format compliance
+    name: Verify clang-format (v13) compliance
     permissions:
       contents: read
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - uses: DoozyX/clang-format-lint-action@2a28e3a8d9553f244243f7e1ff94f6685dff87be
+    - uses: DoozyX/clang-format-lint-action@9ea72631b74e61ce337d0839a90e76180e997283
       with:
-        clangFormatVersion: 10
+        clangFormatVersion: 13
         # When updating the extension list, remember to update
         # Jamulus.pro's CLANG_SOURCES as well.
         extensions: 'cpp,h,mm'
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,6 +24,9 @@ If a feature or function can be achieved in another way by another system or met
 
 You can use your editor's or IDE's clang-format support to accomplish that.
 On the command line, you can run `make clang_format` to do the same before committing.
+Please use the same clang-format version which we use.
+Other versions might produce slightly different output.
+You can find the currently used version in the [workflow definition](https://github.com/jamulussoftware/jamulus/blob/master/.github/workflows/coding-style-check.yml#L20].
 
 Do not use diff/patch to send your code changes but create a Github fork of the Jamulus code and create a Pull Request when you are done.
 
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -856,16 +856,16 @@ if ( GetNumericIniSet ( IniXMLDocument, "server", "centservaddrtype", static_cas
     directoryType = static_cast<EDirectoryType> ( iValue );
 }
 else
-            // clang-format on
-            if ( GetNumericIniSet ( IniXMLDocument,
-                                    "server",
-                                    "directorytype",
-                                    static_cast<int> ( AT_NONE ),
-                                    static_cast<int> ( AT_CUSTOM ),
-                                    iValue ) )
-        {
-            directoryType = static_cast<EDirectoryType> ( iValue );
-        }
+                // clang-format on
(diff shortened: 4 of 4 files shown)


softins:
This has the potential to be a pain. Do we know what version of clang-format is included with various distributions of Linux and macOS? Ubuntu 20 has clang-format version 10. Raspbian 11 (the latest) has clang-format version 11.

When I was developing on Raspbian 10, which only has clang-format 7, it was annoying to have to push on the Pi, fire up an Ubuntu 20 VM, pull to it, do clang-format and push back the changes. It became much nicer when I switched to Raspbian 11, as I could do clang-format natively. Moving to a requirement of clang-format 13 would be a backward step for me.

I suppose I might be able to install a newer clang-format from source, but I prefer to use what is packaged with the distro.

hoffie:
> I'm still not keen.

I understand that regarding the consequences (worsened situation for many Linux users) and I agree that we should try to avoid that. I think this was the main reason for choosing a rather conservative clang-format version when first introduced. However, if we don't update, the only other alternatives I can think of are adding workarounds for wrongly-formatted code or merging PRs despite red warnings (we might be able to reduce the checks to the files which are modified in a PR, though). Do you have any other ideas?

softins:
I haven't encountered any formatting issues yet, and have not studied the examples mentioned in the issue. Has it been determined whether the observed issues are fixed in 11, 12 or 13?

softins:
Maybe if clang-format fails in CI it could raise a PR on the contributor's repo for the required changes? I don't know how hard that would be, or even if it's possible.

hoffie:
> I haven't encountered any formatting issues yet, and have not studied the examples mentioned in the issue. Has it been determined whether the observed issues are fixed in 11, 12 or 13?

The one we've pinpointed exactly (pointer receivers) is fixed in 13. This has already been relevant in a PR by @pgScorpio so far and not on `master`.
Another one is relevant on master and can be seen as part of this PR: https://github.com/jamulussoftware/jamulus/pull/2559/commits/18e45d0505444f17456067a0bd84a64e26aa626d#diff-c044ba02ccf7862f1e2f56436d5b92baa7bb35601791846655fb07ac253a8697

> Maybe if clang-format fails in CI it could raise a PR on the contributor's repo for the required changes? I don't know how hard that would be, or even if it's possible.

I think that's doable, but requires contributors to merge the PR and pull from their own PR (or do it again after the next force push).

softins:
> Another one is relevant on master and can be seen as part of this PR: [18e45d0#diff-c044ba02ccf7862f1e2f56436d5b92baa7bb35601791846655fb07ac253a8697](https://github.com/jamulussoftware/jamulus/commit/18e45d0505444f17456067a0bd84a64e26aa626d#diff-c044ba02ccf7862f1e2f56436d5b92baa7bb35601791846655fb07ac253a8697)

That one looks like clang-format getting confused because of rapid turning off and on of clang-format. I couldn't even tell from scrolling back a bit what the indentation level should be! Is it really necessary to have all possible version compatibility sections un-indented? Might it be better to delimit them with special comments instead and allow them to be subject to clang-format indentation?

hoffie:
I'm not a fan of the un-ident at all, but that's how it is. If there's agreement to change that (@pljones? :)), that would work around *this* part.
Sadly, it's just one of the issues. Another one is this and I can't think of a workaround (despite accepting it or disabling clang-format for it): https://github.com/jamulussoftware/jamulus/pull/2407#issuecomment-1046261464

pljones:
> I'm not a fan of the un-ident at all, but that's how it is. If there's agreement to change that (@pljones? :)), that would work around _this_ part.

Hey, I just live here...  Uh...  In fact, the outdented style for "Look, I'm not normal code, so pay attention" is a bit odd.  Having a `#ifdef` rather than commenting it out (here) would be better - I could have complained in `Jamulus.pro` that you couldn't do both `serveronly` and `testbench`, for example.  I was left having to add a `#ifdef` around a comment, in case someone was working with testbench enabled...

 Sadly, it's just one of the issues. Another one is this and I can't think of a workaround (despite accepting it or disabling clang-format for it): [#2407 (comment)](https://github.com/jamulussoftware/jamulus/pull/2407#issuecomment-1046261464)

pgScorpio:
@hoffie 

> Do you have any other ideas?

Just some thoughts.... (As we always will have to deal with different clang-format-versions I guess.)

- We could make the auto-build check use multiple versions of clang-format, if one of them gives an OK result, the check should be considered OK.
- Can we include the clang-format binaries in the repo (in the OS folders) and let `make clang_format` use those versions. (This would limit the amount of different versions to be checked by auto-build.) 
- Letting `make clang_format` adding a comment in the file with the clang-format version (output of `clang-format --version`) could also help to determine which clang-format version to use for the check.

hoffie:
Thanks for those suggestions.

>  * We could make the auto-build check use multiple versions of clang-format, if one of them gives an OK result, the check should be considered OK.

This would work for new code, but would be annoying for existing code which might flip back and forth dependening on the PR submitter. This increases diff sizes and makes `git blame` less useful.

>  * Can we include the clang-format binaries in the repo (in the OS folders) and let `make clang_format` use those versions. (This would limit the amount of different versions to be checked by auto-build.)

Hm, if we go that route I'd rather point to some canonical source to get the binaries. Git repos usually aren't supposed to contain binaries. We'd need to have at least four versions: Linux x86_64, Linux armhf, Windowx 86_64, Mac x86_64. In addition, `make clang_format` is just one way to format code. Several people rather use the functionality included within their IDEs or editors.

>  * Letting `make clang_format` adding a comment in the file with the clang-format version (output of `clang-format --version`) could also help to determine which clang-format version to use for the check.

Hmm, but I guess this would have the same back-and-forth problem for all non-new files as well? Or would require submitters to have multiple clang-format versions installed in parallel?

hoffie:
What problems would we expect if we switched to v13 as the authoritive version, but advised users only to clang-format the files which they've touched? I think that's what most editors do anyway. We could then drop `make clang_format` again (which does touch all files) and instead advise using [git-clang-format](https://github.com/llvm/llvm-project/blob/main/clang/tools/clang-format/git-clang-format) (which I only found right now).

That way, the only known formatting conflict in current `master` would be the one in [src/settings.cpp](https://github.com/jamulussoftware/jamulus/pull/2559/files#diff-c044ba02ccf7862f1e2f56436d5b92baa7bb35601791846655fb07ac253a8697).

ann0see:
Related: https://github.com/jamulussoftware/jamulus/pull/2600

ann0see:
Closing as I think https://github.com/jamulussoftware/jamulus/pull/2600 is a short term fix. As soon as debian stable moves to a more recent Clang format version, we can re-open this.

Issue #2560, opened 2022-03-28
Can't Load or Save Mixer Channels Setup on Mac


softins:
**Describe the bug**

Load and Save Mixer Channels Setup does not work on Mac release 3.8.2. Nothing happens when the menu items are selected.

**To Reproduce**

Click on File menu, then on either Load Mixer Channels Setup... or Save Mixer Channels Setup...

**Expected behavior**

A finder dialog should appear for selecting the file to load or save.

**Screenshots**

**Operating system**

Reported on Facebook by a Mac M1 user. Also observed by me on Mojave.

**Version of Jamulus**

3.8.2 release - I suspect the same would be true of 3.8.1 (the first signed release), but not of earlier versions.

**Additional context**

A development version 3.8.2dev-xxxxxxxx I already had installed, which was not signed, displays the finder dialog ok. My suspicion is that a required entitlement is missing to allow this function to work in the sandboxed environment. @emlynmac ?

emlynmac:
Where are the setting stored? Are they hard coded to a path somewhere?
My understanding of the sandboxing of the app will change all paths relative to the sandbox. This should mean that functionality is not affected,  but the one time move to a signed version will not have the settings from a version prior to the signing.

softins:
No, it is supposed to open a finder dialog so the user can navigate to the desired directory and specify the filename. Maybe by default it starts in a directory it is not permitted to?

The action functions for those two menu entries are here:
https://github.com/jamulussoftware/jamulus/blob/178de745d6cdf5218df7f12bbb1e3c9d8c2c39c6/src/clientdlg.cpp#L772-L795

The third argument to `getOpenFileName()` and `getSaveFileName()` is supposed to be a `QString&` to specify the directory. Maybe the literal `""` fails in the sandboxing? The documentation says that an empty string is acceptable.

Maybe instead we should pass a variable string that has been initialised to the working directory (possibly `"."`?), which should allow the user to navigate to a different directory if needed, and the running instance to remember that directory for subsequent invocations of the action in the same run? I haven't tried anything, just looking for ideas.

References:
- https://doc.qt.io/qt-5/qfiledialog.html#getOpenFileName
- https://doc.qt.io/qt-5/qfiledialog.html#getSaveFileName

softins:
[...]

Issue #2131, opened 2021-11-23
macOS: Allow r/w access outside of default ini file location


ToolTech:
Previously in pre 3.8.1 I could start multiple clients with different settings and the settings would be stored in separate config files. This way we could start multiple client for different devices and channels on a computer. (OS X)

In 3.8.1 this is mixed up. the names are lost and the settings are not stored. A Majord down step in functionality

softins:
This is because the MacOS version in 3.8.1 is signed with a developer signature, and as a result, apparently has to run in a Sandbox. I believe this is a standard feature of signed apps, but @emlynmac could advise more, as he implemented the signing for us.

So now, the default ini file is stored in `$HOME/Library/Containers/io.jamulus.Jamulus/Data/.config/Jamulus/Jamulus.ini`. If you give a filename to `-i`, then it needs to be a relative filename, not one starting with `/`, `~` or `$HOME`. This will then be relative to `$HOME/Library/Containers/io.jamulus.Jamulus/Data/`, so you could say `-i Config1.ini` to store the config in `.../Data` directly, or `-i .config/Jamulus/Config1.ini` to store it alongside the default ini file.

This definitely needs documenting - apologies - but it was an unforseen side-effect of making the app signed in order to avoid the "Unknown developer" warning on installation.

softins:
BTW, if you have your existing ini files, you can copy them to `~/Library/Containers/io.jamulus.Jamulus/Data/` or `~/Library/Containers/io.jamulus.Jamulus/Data/.config/Jamulus/` and the app should find them.

ToolTech:
And pretty strange to introduce this feature in a minor version ramp

Best regards!

/Anders Modén

softins:
> So you need to copy those files each time you install a new version then.

No, only the first time. The name of the data directory won't change in future versions.

> And pretty strange to introduce this feature in a minor version ramp

It was unexpected - well, it surprised me, at least.

emlynmac:
As we started signing the macOS release, this also means we support app sandboxing on the signed release. This is a security feature of macOS apps more recently, to ensure that they only have access to their files and nothing else.

softins:
> As we started signing the macOS release, this also means we support app sandboxing on the signed release. This is a security feature of macOS apps more recently, to ensure that they only have access to their files and nothing else.

How would this apply to other kinds of app, such as MuseScore, for example, or a text editor, which can read and write arbitrary files?

emlynmac:
[...]

PR #2563, opened 2022-03-28
Disable codeQL on Mac build and add com.apple.security.files.user-selected.read-write entitlements

emlynmac:
Disables codeQL on macOS builds as it breaks signed builds.

Adds entitlements to the app under macOS to access the filesystem as needed.

**Context: Fixes an issue?**
https://github.com/jamulussoftware/jamulus/issues/2560

Ready for testing, use build here:

## Checklist

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -127,7 +127,7 @@ jobs:
             # Stay on 10.15 as long as we use dmgbuild which does not work with 11's hdiutil (?):
             building_on_os:         macos-10.15
             base_command:           QT_VERSION=5.15.2 SIGN_IF_POSSIBLE=1 ./.github/autobuild/mac.sh
-            run_codeql:             true
+            run_codeql:             ${{ secrets.MACOS_CERT == null }}
             xcode_version:          12.1.1
 
           - config_name:            MacOS Legacy (artifacts)
@@ -214,10 +214,7 @@ jobs:
           JAMULUS_BUILD_VERSION:    ${{ needs.create_release.outputs.build_version }}
 
       - name:                       Initialize CodeQL
-        env:
-          MACOS_CERTIFICATE:        ${{ secrets.MACOS_CERT}}
-        # Don't run CodeQL if we have a set up for signing the macOS build
-        if:                         matrix.config.run_codeql && ${{ env.MACOS_CERTIFICATE == '' }}
+        if:                         matrix.config.run_codeql 
         uses:                       github/codeql-action/init@v1
         with:
           languages: 'cpp'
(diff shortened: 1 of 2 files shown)


emlynmac:
I've tested this build: https://github.com/emlynmac/jamulus/suites/5835725008/artifacts/196029418

And it seems to be good. Would be great if @ann0see and @hoffie @softins @pljones could have a look.

pljones:
I'd rather this were two pull requests as the two things it does are unrelated.

emlynmac:
> 2\. It doesn't change the behaviour of the server recording directory as mentioned in [macOS: Allow r/w access outside of default ini file location #2131 (comment)](https://github.com/jamulussoftware/jamulus/issues/2131#issuecomment-989184190). If I click on the `Recording Directory` button, I get a finder dialog in which I can create or navigate to a recording directory, but it will only allow me to navigate within the sandbox, not to any arbitrary directory within my home directory. I assume the new entitlement that was added does not apply to this operation.

This is likely due to _how_ that's done.
The entitlement to read/write is from the user choosing a location / file. This is the way it must be done in a sandboxed app since the all-files was deprecated: https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_files_all

Maybe the server UI can change to have a means to choose a location. I've updated this PR in the meantime to only address the client file access issue.

softins:
The person who originally reported the issue on Facebook has tried the build listed above, and confirmed that it fixes the problem with saving and loading mixer channel files.

softins:
But they just happened to be encountered together. The codeql change would be needed even if we didn't address the reported bug by changing the entitlements.

emlynmac:
> But they just happened to be encountered together. The codeql change would be needed even if we didn't address the reported bug by changing the entitlements.

Yes, this is true; and we need the change in this PR to make this fix actually work. So rather than the overhead of making a whole other one-line PR, builds etc. I see no reason why not to include the change here where it is needed to make the fix work.

emlynmac:
@pljones You can always make a PR for the change to disable codeQL if you want to separate this out, at which time the changes from this PR will become none for that aspect.

softins:
@emlynmac I'm not sure why this was closed (there was no comment), but I raised and merged PR #2564 for the CodeQL part, and have updated your branch to contain just the entitlement change. I kept a backup branch called `bugfix-2560-codeql-tests` in your repo for reference about the attempts to control CodeQL.

You will need to do a `git fetch` to get my changes.

So I'm re-opening this PR ready to merge just the entitlements. I'll update the title in a moment.

OK, it won't let me re-open because I force-pushed to your branch. I might need to create a fresh PR.

PR #2566, opened 2022-03-29
Add extra comments requested in #2564

softins:
**Short description of changes**

Add extra comments requested in #2564 but overlooked before merging

CHANGELOG: SKIP

PR #2569, opened 2022-03-31
Reinstate CodeQL on Mac if not signing the build

softins:

Moves the CodeQL from Mac Legacy back to Mac, and makes it conditional on not doing a signed build.

CHANGELOG: Mac: Make Github workflow skip CodeQL when doing a signed build, to avoid hanging.

**Context: Fixes an issue?**

Replaces #2564 and #2566, and solves the issues encountered when working on #2563.

Probably not

Ready for review

Review

## Checklist

### DIFF
--- a/.github/autobuild/mac.sh
+++ b/.github/autobuild/mac.sh
@@ -82,7 +82,7 @@ case "${1:-}" in
     setup)
         setup
         # check whether signing will be used and prevent
-	 # a return status of 1 from propagating to the script exit status.
+	# a return status of 1 from propagating to the script exit status.
         check_if_signing || true
         ;;
     build)
--- a/.github/autobuild/mac.sh
+++ b/.github/autobuild/mac.sh
@@ -73,6 +73,7 @@ pass_artifact_to_job() {
 case "${1:-}" in
     setup)
         setup
+        prepare_signing
         ;;
     build)
         build_app_as_dmg_installer
--- a/.github/autobuild/mac.sh
+++ b/.github/autobuild/mac.sh
@@ -73,7 +73,7 @@ pass_artifact_to_job() {
 case "${1:-}" in
     setup)
         setup
-        prepare_signing
+        prepare_signing || true
         ;;
     build)
         build_app_as_dmg_installer
--- a/.github/autobuild/mac.sh
+++ b/.github/autobuild/mac.sh
@@ -23,7 +23,7 @@ setup() {
     fi
 }
 
-prepare_signing() {
+check_if_signing() {
     [[ "${SIGN_IF_POSSIBLE:-0}" == "1" ]] || return 1
 
     # Signing was requested, now check all prerequisites:
@@ -34,6 +34,14 @@ prepare_signing() {
     [[ -n "${KEYCHAIN_PASSWORD:-}" ]] || return 1
 
     echo "Signing was requested and all dependencies are satisfied"
+
+    # Tell Github Workflow to disable CodeQL as it interferes with signing
+    echo "::set-output name=disable_codeql::true"
+    return 0
+}
+
+prepare_signing() {
+    check_if_signing || return 1
 
     # Put the cert to a file
     echo "${MACOS_CERTIFICATE}" | base64 --decode > certificate.p12
@@ -73,9 +81,9 @@ pass_artifact_to_job() {
 case "${1:-}" in
     setup)
         setup
-        # set up the macos_signed output if needed, but prevent
-        # a return status of 1 from propagating to the script exit status.
-        prepare_signing || true
+        # check whether signing will be used and prevent
+	# a return status of 1 from propagating to the script exit status.
+        check_if_signing || true
         ;;
     build)
         build_app_as_dmg_installer
--- a/.github/autobuild/mac.sh
+++ b/.github/autobuild/mac.sh
@@ -82,7 +82,7 @@ case "${1:-}" in
     setup)
         setup
         # check whether signing will be used and prevent
-	# a return status of 1 from propagating to the script exit status.
+	 # a return status of 1 from propagating to the script exit status.
         check_if_signing || true
         ;;
     build)
--- a/.github/autobuild/mac.sh
+++ b/.github/autobuild/mac.sh
@@ -73,6 +73,8 @@ pass_artifact_to_job() {
 case "${1:-}" in
     setup)
         setup
+        # set up the macos_signed output if needed, but prevent
+        # a return status of 1 from propagating to the script exit status.
         prepare_signing || true
         ;;
     build)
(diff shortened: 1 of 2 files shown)


softins:
After reviews, I will squash to a single commit before merging. I have left the original commits for review, to show the process of arriving at the solution.

Before the final commit, I pushed to both emlynmac and softins, to compare the runs:
- https://github.com/emlynmac/jamulus/actions/runs/2067750063
- https://github.com/softins/jamulus/actions/runs/2070320595

softins:
And similarly after the final commit:
- https://github.com/emlynmac/jamulus/actions/runs/2070442079
- https://github.com/softins/jamulus/actions/runs/2070443008

.github/autobuild/mac.sh
```diff
@@ -73,6 +73,9 @@ pass_artifact_to_job() {
 case "${1:-}" in
     setup)
         setup
+        # set up the macos_signed output if needed, but prevent
+        # a return status of 1 from propagating to the script exit status.
+        prepare_signing || true
```
hoffie, line 78: Hmm, so we would now be running those steps twice?
Maybe there's some nice way to save that state and skip the second run somehow. Either pass back toe macos_signed output variable as an environment variable or touch some kind of file.

.github/workflows/autobuild.yml
```diff
@@ -122,22 +122,21 @@ jobs:
...
+          - config_name:            MacOS (artifacts+CodeQL)
             target_os:              macos
             # Stay on 10.15 as long as we use dmgbuild which does not work with 11's hdiutil (?):
             building_on_os:         macos-10.15
             base_command:           QT_VERSION=5.15.2 SIGN_IF_POSSIBLE=1 ./.github/autobuild/mac.sh
-            # Disable CodeQL on mac as it interferes with signing the binaries (signing hangs, see #2563 and #2564)
-            run_codeql:             false
+            # run_codeql will be ignored in the steps below if building a signed image
+            # for Mac, as it causes the signing process to hang. See #2563 and #2564.
+            run_codeql:             true
```
hoffie, line 132: Hm, there might be another approach which could work with less additional logic (although I haven't tried it):
Can't we just disable `run_codeql` based on the existence of the secrets?

Seems like [direct secrets access isn't supported for `if` expressions](https://stackoverflow.com/questions/70249519/how-to-check-if-a-secret-variable-is-empty-in-if-conditional-github-actions), but environment variables should work.

```suggestion
            run_codeql:             ${{ env.MACOS_CERT_ID && 'false' || 'true' }}
```
(i.e. this additionally needs a job-level `env:  MACOS_CERT_ID: ${{ secrets.MACOS_CERT_ID}}`)

There is a downside of this approach though: In repos which "enable" signing (= have the required secrets), but don't have a valid signing config (bad password, missing some other secrets), CodeQL will not run although it could. The original approach from this PR handles this more correctly. I'd argue that this is not relevant though. First, there's likely only one repo (emlyn's) which does proper signing. Second, we mainly care for CodeQL on our jamulussoftware repo.

.github/workflows/autobuild.yml
```diff
@@ -212,12 +211,18 @@ jobs:
           key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/workflows/autobuild.yml', '.github/autobuild/android.sh') }}-${{ matrix.config.base_command }}
 
       - name:                       Set up build dependencies for ${{ matrix.config.config_name }}
+        id:                         setup
         run:                        ${{ matrix.config.base_command }} setup
         env:
           JAMULUS_BUILD_VERSION:    ${{ needs.create_release.outputs.build_version }}
+          MACOS_CERTIFICATE:        ${{ secrets.MACOS_CERT}}
```
hoffie, line 218: If we need to keep this at all, the missing space should be added

```suggestion
          MACOS_CERTIFICATE:        ${{ secrets.MACOS_CERT }}
```

.github/autobuild/mac.sh
softins: Well in fact, the only parts of `prepare_signing` that are needed in the setup phase are the initial tests and the setting of the output variable (so that it can be tested in the workflow). So we could make a different function `check_signing` that just does those parts, and maybe even call it from `prepare_signing` to reactor those parts out.

.github/workflows/autobuild.yml
softins: Approaches like this were already tried by @emlynmac without success when developing his recent PR, which got superseded. I'm not sure environment variables are necessarily available at the time the steps are set up, only when they are executed. So I think your suggestion won't work. That is why I looked at having the `setup` phase set an output we could test in an `if` condition on a step, as that is a run-time check.

.github/workflows/autobuild.yml
softins: I just copied and pasted from the Build step. I see the space is missing there too, but I didn't notice before.

.github/workflows/autobuild.yml
hoffie: Ah, I had missed https://github.com/emlynmac/jamulus/commit/1bd0a022a1a242ad207d35c9ccd9a60d70633032 (related [CI failure](https://github.com/emlynmac/jamulus/actions/runs/2055012005)). Seems to be related to the matrix not supporting that. Too bad. :(

hoffie:
Another thing to consider:
@ann0see already introduced a way to run CodeQL on the legacy build (downside: we have to take care of that when removing that, but there's an appropriate comment as a reminder now). If we make CodeQL conditional on the signing, this might block us from or need another reversal if we decide to sign all non-legacy Mac builds (maybe with temporary keys). While there is no concrete PR for that, I think it should be a goal in order to make the autobuild and release code paths as similar as possible. Additionally, Mac arm64/M1 builds seem to require workarounds as long as there is no signature at all.

softins:
> Another thing to consider: @ann0see already introduced a way to run CodeQL on the legacy build (downside: we have to take care of that when removing that, but there's an appropriate comment as a reminder now).

This PR turns off the running of CodeQL on Legacy and removes the comment, as it would no longer be needed. I can't see any reason to run CodeQL on both Mac builds.

.github/autobuild/mac.sh
```diff
@@ -73,6 +81,9 @@ pass_artifact_to_job() {
 case "${1:-}" in
     setup)
         setup
+        # check whether signing will be used and prevent
+	# a return status of 1 from propagating to the script exit status.
```
pljones, line 85: Spacing

ann0see, line 85:
```suggestion
	 # a return status of 1 from propagating to the script exit status.
```

ann0see: Not sure why the spacing didn't work...

ann0see:
Is this still an issue? I believe so?

softins:
Closing this old PR, as the affected files have changed a lot since it was raised.

PR #2571, opened 2022-03-31
Reduce scope of iSndChanIdx

ann0see:

This reduces the scope of iSndChanIdx for windows/macOS. This shouldn't have any negative impact

CHANGELOG: SKIP

**Context: Fixes an issue?**

No.

No.

Not yet tested on Windows. But it should already fail the CI if anything is wrong.

Needs testing on macOS/Windows

## Checklist

### DIFF
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -842,7 +842,6 @@ void CClientSettingsDlg::UpdateSoundDeviceChannelSelectionFrame()
 
     // update input/output channel selection
 #if defined( _WIN32 ) || defined( __APPLE__ ) || defined( __MACOSX )
-    int iSndChanIdx;
 
     // Definition: The channel selection frame shall only be visible,
     // if more than two input or output channels are available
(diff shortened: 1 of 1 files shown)


src/clientsettingsdlg.cpp
```diff
@@ -862,6 +861,9 @@ void CClientSettingsDlg::UpdateSoundDeviceChannelSelectionFrame()
         // input
         cbxLInChan->clear();
         cbxRInChan->clear();
+
+        int iSndChanIdx;
+
         for ( iSndChanIdx = 0; iSndChanIdx < pClient->GetSndCrdNumInputChannels(); iSndChanIdx++ )
```
softins, line 867: As this variable is only used within the loops, it doesn't need to be declared outside the `for()` at all:
```suggestion

        for ( int iSndChanIdx = 0; iSndChanIdx < pClient->GetSndCrdNumInputChannels(); iSndChanIdx++ )
```
And also on line 881 below (Github wouldn't let me attach a suggestion to that line)

ann0see: Probably yes? I still don’t understand why it was that high out

softins: Neither do I. Maybe it was previously used outside of the loops, although it isn't now.

ann0see:
Updated.

hoffie (approved):
Nit: I'm still wishing for commit messages with context which can easily be grokked then browsing the history via `git log --format=one`... :)

softins (approved):

softins:
But a local variable like that is not allocated from the heap. It's allocated on the stack, just by moving the stack pointer: a single instruction updating a register, taking a few ns nowadays, and never with any possibility of blocking.

Issue #2573, opened 2022-04-01
improve maintainability Jamulus.pro


pgScorpio:
**Describe the bug**

In the past month there where a lot of changes to Jamulus.pro, leading to a lot of rebase actions with a high chance on mistakes.
Especially for me since I have a sound-redesign branch with added files for Sound in the OS folders.
(PS: there are also several headers missing in the definitions for ASIO)

The headers and sources for sound (especially linux) are used in multiple places in the Jamulus.pro file and therefore have to be changed in multiple places too.
I suggest we use separate defines for the sound files which can be used at many places, but have to be edited in just one place.

**To Reproduce**

n.a.

**Expected behavior**

Easier maintenance, less prone to mistakes/errors.

**Screenshots**

n.a.

**Operating system**

all

**Version of Jamulus**

n.a.

**Additional context**
General idea:
```
HEADERS_ASIO= ...
SOURCES_ASIO= ...

HEADERS_JACK= ...
SOURCES_JACK= ...

HEADERS_COREAUDIO= ...
SOURCES_COREAUDIO= ...

HEADERS_COREAUDIOIOS= ...
SOURCES_COREAUDIOIOS= ...

HEADERS_OBOE= ...
SOURCES_OBOE= ...

....

HEADERS+=$$HEADERS_JACK
SOURCES+=$$SOURCES_JACK

...

HEADERS+=$$HEADERS_COREAUDIO
SOURCES+=$$SOURCES_COREAUDIO

....

HEADERS+=$$HEADERS_JACK
SOURCES+=$$SOURCES_JACK

```
Another point is that it looks very strange when including "linux" files in a Windows build etc.
So it would be more clear when using separate folders for the sound implementation files like sound/jack, sound/asio, sound/coreaudio, sound/oboe, ...

pgScorpio:
See my [implementation](https://github.com/pgScorpio/jamulus/blob/autobuild/soundRedesignDev/Jamulus.pro) in my sound-redesign branch. 
(which has several added files, so not valid for current master.)

hoffie:
1) I like the idea of having the includes grouped together (although I don't have a strong opinion here)
2) I totally agree regarding the backend-specific directory structure. I don't know why those directories should be at the root of the repo either. `src/sound/{jack,asio,oboe,coreaudio,...}` would seem logical to me. I would certainly support this change.

ann0see:
[...]

Issue #2417, opened 2022-02-20
Clarify ASIO SDK usage


hoffie:
It seems to be non-obvious how the ASIO SDK and its license integrates with Jamulus. We should do everything to make the situation more clear.

- [x] Add a clause on the Windows ASIO client download page saying "Please note that you are not permitted to redistribute this binary without acquiring a [licence agreement from Steinberg](https://www.steinberg.net/developers/)" (and in the About box)
- [x] Make a Windows server download using "nosound" which doesn't touch ASIO SDK
- [x] Verify that JACK Windows builds don't touch ASIO SDK #2443
- [x] Check/update ASIO mentions regarding license requirements in
  - [x] Installer Startup/Splash-screen ("ASIO" + copyrights notice - see 3.b in their Terms, logo is optional)
  - [x] About dialog ("ASIO" + copyrights notice)
  - [x] [README.md](https://github.com/jamulussoftware/jamulus#acknowledgements) (copyrights notice only)
  - [x] Website ("ASIO" + copyrights notice, logo is optional). 
- [ ] Sign the agreement and send it to Steinberg to cover jamulus.io's distribution of ASIO-enabled Windows binaries (we can't cover other usages, sadly)
  - [x] Who from the team would be available to do that? @gilgongo + @pljones for deeper technical stuff
- [x] Agree approach re. system GPL library exception vs additional terms notice 
- [x] Document that users who self-compile and distribute the result should check this for themselves ([README.md](https://github.com/jamulussoftware/jamulus#acknowledgements) as well?)
- [x] Add pointers to the code to be careful about this topic, i.e. *"never copy from ASIO SDK as it makes the situation more complicated"*
  - [x] [windows/deploy_windows.ps1](https://github.com/jamulussoftware/jamulus/blob/master/windows/deploy_windows.ps1#L8) #2443
  - [x] windows/sound.h #2443

References:
See PDF in the [SDK zip](https://download.steinberg.net/sdk_downloads/asiosdk_2.3.3_2019-06-14.zip).

I'm currently don't plan to work on this myself, but I do think that we should move this forward for the next release and have added it to 3.9.0 therefore.

gilgongo:
The approved wording for the copyright statement is: ”ASIO is a trademark and software of Steinberg Media Technologies GmbH”

gilgongo:
I think the _least contentious_ option here is to give anyone who receives the Jamulus source an additional permission under GNU GPL v3 section 7 as follows:

[...]

PR #2575, opened 2022-04-01
Refactoring: Move sound sources into src/sound folder

ann0see:

Moves the sound.cpp / sound.h / sound.mm files to src/sound/

CHANGELOG: Refactoring: move sound API files into src/sound folder. Contributors will now find the sound.cpp/sound.h/sound.mm in the src/sound folder for consistency

**Context: Fixes an issue?**

Related to: https://github.com/jamulussoftware/jamulus/issues/2573

@pgScorpio, I assume you should also have a look at this PR. 

Yes

Waiting on second review

Second review

## Checklist

-  [x] I tested my code and it does what I want (works on Linux and Windows. Since the CI is clean it should be ok)


pgScorpio:
@ann0see 
Yes, I will keep an eye on this PR too...

And if we indeed are going to change the src tree we maybe should have a look if it wouldn't be useful to split up some more   at the same time... i.e. Common, Client, Server, Protocol,....

And would it be possible to get the same structure in the editors (Qt Creator, VS, XCode) too? That would be a great improvement too.

hoffie (commented):
Didn't have a deeper look yet, just some thoughts. As said elsewhere, I like this idea.

- [ ] Did you check translations? It looks like each .ts file contains references to the source path, so I guess they must/should be updated to match the new location to avoid all translations showing up as new? I haven't checked though if this is really the case. Maybe Qt tooling is intelligent enough to update it itself, but it should be checked nevertheless.
- [x] Various checks or scripts contain explicit references to our code directories. As everything is moved to the existing src/, this should be fine already, but I'm mentioning it nevertheless.
- [ ] I didn't think of the fact that windows/ etc. are not gone automatically, as they contain build scripts, platform-specific resources (icons, installers). Some (at least mac/) continues to ship source files. Would be nice to get rid of that as well, but maybe in a follow-up PR...
- [x] I think it would be logical to move the windows/ASIOSDK2 extraction target. Either, it should be moved along with the code which is using it or it could be moved to libs/ where all other third-party stuff lives. As this is closely related to the move of the ASIO sound implementation, I think this should be done in this PR as well. It will likely affect docs and build scripts though.

ann0see:
Hmm. Moving ASIO around involves some bigger changes, I think?

I'd like to have it in libs.

ann0see:
@hoffie just updated the .ts files

ann0see:
> I didn't think of the fact that windows/ etc. are not gone automatically, as they contain build scripts, platform-specific resources (icons, installers). Some (at least mac/) continues to ship source files. Would be nice to get rid of that as well, but maybe in a follow-up PR...

Yes. But this needs a bit of restructuring other places too...

Jamulus.pro
```diff
@@ -1120,6 +1120,6 @@ contains(CONFIG, "disable_version_check") {
 # be sure to update .github/workflows/coding-style-check.yml and .clang-format-ignore as well.
 CLANG_FORMAT_SOURCES = $$files(*.cpp, true) $$files(*.mm, true) $$files(*.h, true)
 CLANG_FORMAT_SOURCES = $$find(CLANG_FORMAT_SOURCES, ^\(android|ios|mac|linux|src|windows\)/)
-CLANG_FORMAT_SOURCES ~= s!^\(windows/\(nsProcess|ASIOSDK2\)/|src/res/qrc_resources\.cpp\)\S*$!!g
+CLANG_FORMAT_SOURCES ~= s!^\(windows/nsProcess/|libs/ASIOSDK2|src/res/qrc_resources\.cpp\)\S*$!!g
```
ann0see, line 1123: @hoffie I think you should check if this is ok here.

pgScorpio:
> Hmm. Moving ASIO around involves some bigger changes, I think?
> 
> I'd like to have it in libs.

I agree that that seems more consistent. oboe is also in libs.
Though I don't see yet why that would involve bigger changes. Changing base folder is changing base folder, whether it is src/sound/asio/ or src/libs/ doesn't make that much of a difference or do I mis something here ?

ann0see:
No. I did that already in this PR - and it seems to build fine.

Jamulus.pro
```diff
@@ -191,28 +191,28 @@ win32 {
...
         !exists(/usr/include/jack/jack.h) {
             !exists(/usr/local/include/jack/jack.h) {
-                 error("Error: jack.h was not found at the usual place, maybe jack is not installed")
+                 error("Error: jack.h was not found at the usual place, maybe JACK is not installed")
             }
         }
-        HEADERS += linux/sound.h
-        SOURCES += linux/sound.cpp
+        HEADERS += src/sound/jack/sound.h
+        SOURCES += src/sound/jack/sound.cpp
```
pgScorpio, line 198: Why still repeating the same lines again and again?
Whenever something changes in the source files needed for a specific sound implementation I would like to have to change that at just one location.
So who not using separate defines for the sound folders as I suggested in [this issue](https://github.com/jamulussoftware/jamulus/issues/2573) ?

So using
```
HEADERS+=$$HEADERS_JACK
SOURCES+=$$SOURCES_JACK
```
etc...

COMPILING.md
```diff
@@ -89,8 +89,8 @@ Most users will probably want to use this method:
 
 ### Compiling only
 
-1. Create a folder under `\windows` called ASIOSDK2
-1. Download the [ASIOSDK](https://www.steinberg.net/asiosdk), open the top level folder in the .zip file and copy the contents into `[\path\to\jamulus\source]\windows\ASIOSDK2` so that, e.g., the folder `[\path\to\jamulus\source]\windows\ASIOSDK2\common` exists
+1. Create a folder under `\libs` called ASIOSDK2
+1. Download the [ASIOSDK](https://www.steinberg.net/asiosdk), open the top level folder in the .zip file and copy the contents into `[\path\to\jamulus\source]\libs\ASIOSDK2` so that, e.g., the folder `[\path\to\jamulus\source]\libs\ASIOSDK2\common` exists
```
pgScorpio, line 93: Isn't it easier if we make the folder libs/ASIOSDK2 already exist, so we can just say:

Download the [ASIOSDK](https://www.steinberg.net/asiosdk), open the top level folder in the .zip file and copy the contents into `[\path\to\jamulus\source]\libs\ASIOSDK2` so that, e.g., the folder `[\path\to\jamulus\source]\libs\ASIOSDK2\common` exists

Jamulus.pro
```diff
@@ -125,19 +125,19 @@ win32 {
             message(Using ASIO.)
             message(Please review the ASIO SDK licence.)
 
-            !exists(windows/ASIOSDK2) {
-                error("Error: ASIOSDK2 must be placed in Jamulus windows folder.")
+            !exists(libs/ASIOSDK2) {
```
pgScorpio, line 128: I think it is better to check for libs/ASIOSDK2/common 

That's conform compiling.md "so that, e.g., the folder `[\path\to\jamulus\source]\libs\ASIOSDK2\common` exists" )

Jamulus.pro
pgScorpio: I think line 1122 could change to `CLANG_FORMAT_SOURCES = $$find(CLANG_FORMAT_SOURCES, ^\(src)/)`
since all sources are under src now ?

ann0see: No. There's still some cpp code in these folders(androiddebug.cpp) for example

pgScorpio: Wouldn't it be logical to move those somewhere under src too ? (perhaps libs/android?)

hoffie: > Wouldn't it be logical to move those somewhere under src too ?

Yes, maybe, but probably not as part of this PR which is titled and branch-named to touch sound files. :)

> (perhaps libs/android?)

It's not a generic lib, rather a platform-specific extension. Also, `libs/` is used for third-party imported code only and it has benefits to keep it that way. As these are not many files, it might be ok to put them directly under src/ and prefix (mac_activity.mm) or suffix (activity_mac.mm) them accordingly. Again... different PR. :)

@ann0see Looks good at first glance, you can run qmake and check the Makefile's clang_format target whether it includes ASIOSDK2 or not (it obviously should not).

ann0see: I checked it yesterday and it seems to be ok.

Jamulus.pro
ann0see: Yes that's a good suggestion but probably part of another change/PR (maybe including some other smaller improvements of the file in general).

COMPILING.md
ann0see: Copying the contents is more work than just extracting it – at least I think that's the case from CLI/Autobuild.

Jamulus.pro
pgScorpio: As a matter of fact I think moving the sound related files is a good reason to change this too at the same time.

COMPILING.md
pgScorpio: The document says "create a folder" and "copy the contents of the archive so that..."
I think copying to an existing (empty) folder decreases the chance on mistakes.

ann0see: Hmm. I now tested what the standard Windows 11 behaviour actually is. Right clicking and starting extract gives a folder in a folder – that's indeed confusing. I'd like not to change the repo but better the description since the deploy_windows script should download the SDK and might need to be modified.

pgScorpio: > since the deploy_windows script should download the SDK and might need to be modified.

I wasn't aware of that. Isn't that against all ASIO rules as discussed in [this](https://github.com/jamulussoftware/jamulus/issues/2417#issuecomment-1068947074) issue)

If it is OK to automatically download it in the the deploy_windows script it could just as well be downloaded as a submodule like oboe.

hoffie: >I wasn't aware of that. Isn't that against all ASIO rules as discussed in https://github.com/jamulussoftware/jamulus/issues/2417#issuecomment-1068947074 issue.

I think "might need to be modified" was related to the build script, not to ASIO SDK files.

> If it is OK to automatically download it in the the deploy_windows script it could just as well be downloaded as a submodule like oboe.

Submodules only work for git repositories. ASIO SDK is only provided as a ZIP download.

pgScorpio: > Submodules only work for git repositories. ASIO SDK is only provided as a ZIP download.

OK not a submodule then, but it still could be downloaded from the make file. Either as a separate make target or as part of the build right ?

pgScorpio:
> move sound API files into src/sound folder

To keep related things together please move soundbase.h and soundbase.cpp to ./Sound too !

ann0see:
@pgScorpio moved the soundbase files too.

Also added a README stub which you can use for documenting your new design + ideas.

Jamulus.pro
ann0see: Fixed.

Jamulus.pro
ann0see: Yes, but I'd really like to have this PR merged soon (and this can be added in another one)

COMPILING.md
ann0see: I mean the deploy_windows.ps1 file already does that?

ann0see:
What's missing here?

pgScorpio (approved):
Looks good to me

hoffie (approved):
Looks good to me. Thanks! :)

ann0see:
@pljones os this ok to be merged? @pgScorpio depends on this being merged such that he can continue

ann0see:
@jamulussoftware/maindevelopers any review here? I think we should merge this.

ann0see:
Rebased. I'll open another PR moving the other files soon

pljones:
One thing I noticed when I was doing my various changes... linux/sound.h depends on src/server.h.  It shouldn't - not sure if you want to go fixing this here, though.  (It means moving the high resolution timer from src/server.* to src/util.* - was a simple cut-n-paste.)

ann0see:
I wouldn't want to include it here. This should just move the files without many changes

pgScorpio:
> One thing I noticed when I was doing my various changes... linux/sound.h depends on src/server.h. It shouldn't - not sure if you want to go fixing this here, though. 

Should be another PR I think

> (It means moving the high resolution timer from src/server.* to src/util.* - was a simple cut-n-paste.)

To keep things clean I would create a new .h/.cpp i.e. 'timers'.
I personally don't call a timer a 'util' and in my opinion util is already becoming a receptacle.

ann0see:
@pljones could you please review this PR?

ann0see:
@hoffie @npostavs since you're both familiar with the code could you please review? I assume this is safe to merge.

Jamulus.pro
```diff
@@ -125,19 +125,19 @@ win32 {
             message(Using ASIO.)
             message(Please review the ASIO SDK licence.)
 
-            !exists(windows/ASIOSDK2) {
-                error("Error: ASIOSDK2 must be placed in Jamulus windows folder.")
+            !exists(libs/ASIOSDK2/common) {
+                error("Error: ASIOSDK2 must be placed in Jamulus \libs folder.")
```
npostavs, line 129: Should be `/libs` or `\\libs`, not `\libs`, I think?

ann0see: Not sure what the Windows convention is?

ann0see: Ok. Changed it. I think `\\` can be used for local paths and network paths. At least it's nothing bad.

Jamulus.pro
```diff
@@ -125,19 +125,19 @@ win32 {
             message(Using ASIO.)
             message(Please review the ASIO SDK licence.)
 
-            !exists(windows/ASIOSDK2) {
-                error("Error: ASIOSDK2 must be placed in Jamulus windows folder.")
+            !exists(libs/ASIOSDK2/common) {
+                error("Error: ASIOSDK2 must be placed in Jamulus \\libs folder such that e.g. \\libs\ASIOSDK2\common exists.")
```
npostavs, line 129: Ah, you added two more unescaped backslashes.

I double checked it, and it seems that qmake treats `\\x` and `\x` the same (with exception of `x = "`), so I guess you could just leave them unescaped, but it still seems cleaner to have it escaped, IMO.

ann0see: Ok.

ann0see:
Note to maintainers: Should be squash-merged.

ann0see:
Since @pgScorpio depends on this and I think it's save, I will merge this in the following days if nobody objects

ann0see:
@jamulussoftware/maindevelopers I will merge this today

PR #2582, opened 2022-04-06
Scripts: Apply a consistent coding style

hoffie:
**Short description of changes**
- Scripts: Add .editorconfig with shfmt style definition
- Scripts: Adjust code which `shfmt` can't handle properly
- Scripts: Run `shfmt -w .`

(A CI check will be in a follow-up PR: https://github.com/hoffie/jamulus/commit/ci-check-shell-scripts)

CHANGELOG: Scripts: Applied a consistent coding style

**Context: Fixes an issue?**

Related: #2474

**Does this change need documentation? What needs to be documented and how?**

No.

**Status of this Pull Request**

Ready.

**What is missing until this pull request can be merged?**

Reviews.

## Checklist

-  [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
-  [x] I tested my code and it does what I want
-  [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
-  [ ] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
-  [ ] I've filled all the content above

PR #2585, opened 2022-04-06
Coding style: Add .editorconfig for cross-language standardized indentation and line endings

hoffie:
This adds a new config file which is supported by multiple editors and IDEs in order to get started with a consistent style, e.g. regarding indentations.
This makes sense for both file types which are covered by linters (.cpp, .sh) as it provides a sane starting point and has additional features (e.g. config for line endings) as well as for file types which do not have linting at all.
Having such a generic config file seems preferable to having none or multiple editor-specific configs.

See:
https://editorconfig.org

CHANGELOG: Coding Style: Added .editorconfig for cross-language standardized indentation and line endings

**Context: Fixes an issue?**

Related: #2552 (cc @pgScorpio)

Having the file is useful on its own as [many tools detect it automatically](https://editorconfig.org/#pre-installed).
In addition, we *could* add it to CONTRIBUTING.mdin order to point users to install editor/IDE plugins, but I'm not sure if it's worth it (cc @ann0see).

Ready

Reviews

## Checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+# See https://editorconfig.org/
+
+# This is the top-most .editorconfig file
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{cpp,h,mm,nsi,pro,ps1,py,sh}]
+indent_style = space
+indent_size = 4
+
+[*.{yaml,yml}]
+indent_style = space
+indent_size = 2
+
+[libs/**]
+ignore = true


.editorconfig
```diff
@@ -0,0 +1,20 @@
...
+
+[*.{cpp,h,mm,nsi,pro,ps1,py,sh}]
+indent_style = space
+indent_size = 4
+
+[*.{yaml,yml}]
+indent_style = space
+indent_size = 2
+
+[libs/**]
```
ann0see, line 19: There are no other parts which should be excluded? Are you sure (docs/)

hoffie: Hm, I don't see any obvious mistakes when editing a .md file from docs/ with vim (and the editorconfig plugin) with this config. I think the line ending/trailing LF/stripping trailing whitespace should apply as well (and works in my tests).

We should keep in mind that this is not about re-formatting existing files, but about keeping the style when creating new files or adding existing files (well, the counter example is `shfmt` which does use this file to decide about the proper formatting for `.sh` files, see #2582).

pgScorpio: Correct that this one is important when creating new files, but just as important, maybe even more important, is the gitconfig file.
See my latest comment on [this issue](https://github.com/jamulussoftware/jamulus/issues/2552#issuecomment-1090465840).
Without the proper gitconfig, on Windows all text files are converted to CR/LF on checkout.

pgScorpio: I think you should check the config for ps1 files. they are typical windows and might expect cr/lf (especially the older Powershells).

hoffie: > Correct that this one is important when creating new files, but just as important, maybe even more important, is the gitconfig file.

That's right, but [.git/config cannot be part of a repository](https://stackoverflow.com/questions/18329621/how-to-store-a-git-config-as-part-of-the-repository/18330114#18330114) AFAIK, so while this is related to this PR, it's not something we can do as part of the repo.

> I think you should check the config for ps1 files. they are typical windows and might expect cr/lf (especially the older Powershells).

They are already supposed to be LF, but might be converted to CRLF on Windows checkouts.
```
$ file .github/autobuild/windows.ps1 
.github/autobuild/windows.ps1: ASCII text
```

```
$ hexdump -c .github/autobuild/windows.ps1  | head -n4
0000000   #       S   t   e   p   s       f   o   r       g   e   n   e
0000010   r   a   t   i   n   g       W   i   n   d   o   w   s       a
0000020   r   t   i   f   a   c   t   s       v   i   a       G   i   t
0000030   h   u   b       A   c   t   i   o   n   s  \n   #       S   e

$ git show master:.github/autobuild/windows.ps1 | hexdump -c | head -n4
0000000   #       S   t   e   p   s       f   o   r       g   e   n   e
0000010   r   a   t   i   n   g       W   i   n   d   o   w   s       a
0000020   r   t   i   f   a   c   t   s       v   i   a       G   i   t
0000030   h   u   b       A   c   t   i   o   n   s  \n   #       S   e

$ git ls-files .github/autobuild/windows.ps1 windows/* --eol
i/lf    w/lf    attr/                 	.github/autobuild/windows.ps1
i/lf    w/lf    attr/                 	windows/NSISCopyRegistryKey/README.txt
i/lf    w/lf    attr/                 	windows/NSISCopyRegistryKey/registry.nsh
i/lf    w/lf    attr/                 	windows/deploy_windows.ps1
i/-text w/-text attr/                 	windows/installer-banner.bmp
i/-text w/-text attr/                 	windows/installer-welcome.bmp
i/lf    w/lf    attr/                 	windows/installer.nsi
i/-text w/-text attr/                 	windows/jamulus-server-icon-2020.ico
i/-text w/-text attr/                 	windows/mainicon.ico
i/lf    w/lf    attr/                 	windows/mainicon.rc
i/lf    w/lf    attr/                 	windows/nsProcess/COPYING
i/lf    w/lf    attr/                 	windows/nsProcess/ConvFunc.h
i/lf    w/lf    attr/                 	windows/nsProcess/README_JamulusChanges.txt
i/lf    w/lf    attr/                 	windows/nsProcess/Readme.txt
i/lf    w/lf    attr/                 	windows/nsProcess/api.h
i/lf    w/lf    attr/                 	windows/nsProcess/nsProcess.c
i/lf    w/lf    attr/                 	windows/nsProcess/nsProcess.sln
i/lf    w/lf    attr/                 	windows/nsProcess/nsProcess.vcxproj
i/lf    w/lf    attr/                 	windows/nsProcess/nsProcess.vcxproj.filters
i/lf    w/lf    attr/                 	windows/nsProcess/nsis_tchar.h
i/lf    w/lf    attr/                 	windows/nsProcess/pluginapi.c
i/lf    w/lf    attr/                 	windows/nsProcess/pluginapi.h
i/lf    w/lf    attr/                 	windows/sound.cpp
i/lf    w/lf    attr/                 	windows/sound.h
```

pgScorpio: > That's right, but [.git/config cannot be part of a repository](https://stackoverflow.com/questions/18329621/how-to-store-a-git-config-as-part-of-the-repository/18330114#18330114) AFAIK

That's right. But you **can** include a .gitconfig file in the root of the repo and let contributers execute `git config --local include.path ../.gitconfig` to include that one in .git/config.

ann0see (approved):

pgScorpio:
@ann0see @pljones @hoffie 

> That's right, but [.git/config cannot be part of a repository](https://stackoverflow.com/questions/18329621/how-to-store-a-git-config-as-part-of-the-repository/18330114#18330114) AFAIK

Aren't we gonna use .gitattributes to force line endings ? 
The main advantage is that .gitattributes CAN be included in the repo.

See [here](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings) or [here](https://stackoverflow.com/questions/9976986/force-lf-eol-in-git-repo-and-working-copy).

ann0see:
@pgScorpio I mean if it works with .gitattributes too, I don't think there's any harm. I'd however not see this as a replacement.

pgScorpio:
> @pgScorpio I mean if it works with .gitattributes too, I don't think there's any harm. I'd however not see this as a replacement.

Indeed not a replacement, but an addition.
Forcing LF endings in .editorconfig is needed when creating new files, but you need .gitattributes to force the line endings on checkout too. (otherwise all text files will still be CR/LF after checkout on windows,)

.editorconfig
hoffie: > That's right. But you **can** include a .gitconfig file in the root of the repo and let contributers execute `git config --local include.path ../.gitconfig` to include that one in .git/config.

Yes, sure, but it defeats the reasoning for Git not supporting that by default. :)

> EDIT: Apparently you also can use .gitattributes to do the same: See [here](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings) or [here](https://stackoverflow.com/questions/9976986/force-lf-eol-in-git-repo-and-working-copy).

Ah, I wasn't aware that those could be part of the repo. This would definitely make sense.

hoffie:
I think we should have `.gitattributes` in the repo, but I don't plan to squash this into this PR as well as this PR does not strictly depend on it (although it certainly benefits from it and works in tandem) and because we might take some time to decide about the proper list.

I'll merge this PR in some days even if it lacks a second review until now.

pljones (approved):

PR #2583, opened 2022-04-06
Global MsgBoxes and Commandline

pgScorpio:
**Short description of changes**

New attempt replacing #2541

Added global messagebox and commandline parameter parsing classes to global.h/main.c
Adapted all messageboxes to use the new CMsgBoxes class.

CHANGELOG: 

**Context: Fixes an issue?**

General improvement and preparation for sound-redesign.

**Does this change need documentation? What needs to be documented and how?**

**Status of this Pull Request**

**What is missing until this pull request can be merged?**

## Checklist

-  [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
-  [x] I tested my code and it does what I want
-  [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
-  [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
-  [x] I've filled all the content above

PR #2598, opened 2022-04-14
Implementation of global messages

pgScorpio:
New implementation of the messageboxes part of #2583

CHANGELOG: Messageboxes will show up related to the instance instead of at system default position and they will have 'clientname' in the window title. In headless mode the messages will appear on the terminal.

**Context: Fixes an issue?**

General improvement and preparation for sound-redesign.

No documentation changes

Waiting on review

- [ ] Review

## Checklist


ann0see:
Great! Thank you!

ann0see:
Could you please add a CHANGELOG entry in your PR description (and a short reasoning why you need this + advantages)

src/messages.cpp
```diff
@@ -0,0 +1,58 @@
...
+tMainform* CMessages::pMainForm = NULL;
+QString    CMessages::strMainFormName;
+
+/******************************************************************************\
+* Message Boxes                                                                *
+\******************************************************************************/
+void CMessages::ShowError ( QString strError )
+{
+#ifndef HEADLESS
+    QMessageBox::critical ( pMainForm, strMainFormName + ": " + QObject::tr ( "Error" ), strError, QObject::tr ( "Ok" ), nullptr );
```
ann0see, line 36:
```suggestion
    QMessageBox::critical ( pMainForm, strMainFormName + ": " + tr ( "Error" ), strError, tr ( "Ok" ), nullptr );
```
Is "Ok" really not an integrated button?

ann0see (commented):
Looks interesting. Are there any examples yet where the new message box class is a benefit?

src/messages.cpp
pgScorpio: > Looks interesting. Are there any examples yet where the new message box class is a benefit already?

Well we first have to merge this one before we can modify the other messagebox calls of coarse...
But it has multiple benefits. From coding point of view showing a message becomes easier and more consistent, just Cmessages::ShowError( 'text' ) and in headless mode the message will now show up on the terminal.
From the users point of view it's especially of use for people running multiple instances, since the messagebox will popup on the instance generating the message and will have the 'clientname' in the title too.

pgScorpio: > Is "Ok" really not an integrated button?

The  standard messagebox has one button and here we just set the button text. 
I don't like to depend on default values, since they might change. In this way I'm always sure what the result will be.

pgScorpio:
> Could you please add a CHANGELOG entry in your PR description (and a short reasoning why you need this + advantages e.g uniform way to show error messages in future)

Done. Please let me know if it's ok this way.

pgScorpio:
Questions:

Do we want a "Jamulus [<clientname>]" prefix to appear on the debug steams too?
(Personally I think this might be usefull.)

Would it be useful to be able to set the button text from the "ShowXxxx" calls (optional parameter defaulting to "Ok")?
(Personally I think we shouldn't for consistency, since different text's might give the expectation of different behavior.)

src/main.cpp
```diff
@@ -990,17 +1002,10 @@ int main ( int argc, char** argv )
     catch ( const CGenErr& generr )
     {
         // show generic error
-#ifndef HEADLESS
-        if ( bUseGUI )
-        {
-            QMessageBox::critical ( nullptr, APP_NAME, generr.GetErrorText(), "Quit", nullptr );
```
pgScorpio, line 996: This is the only place in the code where the messagebox button originally doesn't have a "Ok" text, though the original "Quit" would expect the program to terminate when clicking it, which it doesn't.

If we really want to 'Quit' here it should be a modal messagebox and we should really have to exit on pressing 'Quit'

hoffie (commented):
I think the UI abstraction makes sense, as it saves passing the parent window and the title all over again.
I'm less sure about the headless stuff. My gut feeling is that there are not that many instances where there's duplicated error handling for UI/CLI stuff, but I might be wrong. I think the CLI logging handling should use some reworking as well, so this might cover it. But maybe we should do that on its own as the first step and see if it makes sense to merge that later? My main concern is that if we replaced all qWarning/qInfo calls with the new abstraction, we might start showing solely informational messages as dialogs accidentially?

I think it would be good to have @pljones's judgement here first.

PR note: It seems like this PR adds and includes the infrastructure (src/messages) and updates one caller place. If we decide to add it, it would make sense to update all callers in the same go (separate commit, but same PR).

(Maybe there was some confusion about that in the other PRs regarding what to include. A PR should do one thing. If it aims to improve the message box implementation, it should include all of that, even if that touches lots of different files. It's just that all those changes should be related to the same specific goal.)

src/messages.cpp
ann0see: Ok. Thanks for the note here.

Concerning the QObject::tr() call: I think the code doesn't use it in other places? Should't we stay consistent and use tr()

ann0see: >  I don't like to depend on default values, since they might change. In this way I'm always sure what the result will be.

Yes, however the default values exist for a reason. Also they're translated automatically.

src/main.cpp
ann0see: Is CGenErr critical as in Jamulus would be unusable if this error occurs? Probably not if we didn't quit the program before. Do you consider this a bug you fixed?

src/messages.cpp
```diff
@@ -0,0 +1,58 @@
...
+    qCritical() << "Error: " << strError.toLocal8Bit().data();
+#endif
+}
+
+void CMessages::ShowWarning ( QString strWarning )
+{
+#ifndef HEADLESS
+    QMessageBox::warning ( pMainForm, strMainFormName + ": " + QObject::tr ( "Warning" ), strWarning, QObject::tr ( "Ok" ), nullptr );
+#else
+    qWarning() << "Warning: " << strWarning.toLocal8Bit().data();
```
ann0see, line 47: Why .toLocal8Bit()?

src/messages.h
```diff
@@ -0,0 +1,47 @@
...
+ *      CMessages::MainForm() as parent parameter
+ *  and CMessages::MainFormName() as base of the title parameter
+\******************************************************************************/
+
+#include <QString>
+#ifndef HEADLESS
+#    include <QMessageBox>
+#    define tMainform QDialog
+#else
+#    include <QDebug>
```
ann0see, line 20: What does QDebug include? Does it add bloat on a release build?

ann0see:
> My main concern is that if we replaced all qWarning/qInfo calls with the new abstraction, we might start showing solely informational messages as dialogs accidentially?

I mean you could also add a boolean to only show it via CLI?

ann0see:
> Would it be useful to be able to set the button text from the "ShowXxxx" calls (optional parameter defaulting to "Ok")?

Overload the methods?

src/main.cpp
pgScorpio: > Is CGenErr critical as in Jamulus would be unusable if this error occurs? 

CGenErr is only to retrieve the error message from an exception...

> Probably not if we didn't quit the program before. Do you consider this a bug you fixed?

No I didn't fix anything here. I just note that the button text here is different, but the original button text is misleading.

ann0see: Thanks for clarifying. 

Yes you did fix a linguistic "bug".

pgScorpio:
> I'm less sure about the headless stuff. My gut feeling is that there are not that many instances where there's duplicated error handling for UI/CLI stuff, but I might be wrong. 

There are several places places in the code with ifndef HEADLESS QMessage #else qStream

The QMessageBox class would make those ifdefs unnecessary. Only if you explicitly don't want the message on the terminal in headless mode you should now use an ifdef. And I think there are very few reasons **not** to show errors/warnings/info on the terminal in headless mode.

> My main concern is that if we replaced all qWarning/qInfo calls with the new abstraction, we might start showing solely informational messages as dialogs accidentially?

I'm not sure if I really understand this remark. But we have Error, Warning and Info messageboxes. We would just replace all QMessagebox::xxxx calls with the matching CMessageBox::ShowXxxx call.

> PR note: It seems like this PR adds and includes the infrastructure (src/messages) and updates one caller place. If we decide to add it, it would make sense to update all callers in the same go (separate commit, but same PR).

Sure !, but I did that in the other PR and I was requested NOT to do that again this PR but in a separate PR after this one is merged....
 
> (Maybe there was some confusion about that in the other PRs regarding what to include. A PR should do one thing. If it aims to improve the message box implementation, it should include all of that, even if that touches lots of different files. It's just that all those changes should be related to the same specific goal.)

So there is still some difference in opinion on that. 
Just let me know what we agree upon, new commit or new PR doesn't make that much of a difference to me.

pgScorpio:
> > Would it be useful to be able to set the button text from the "ShowXxxx" calls (optional parameter defaulting to "Ok")?
> 
> Overload the methods to add the ability to do that?

There are several ways to do it, but as said another button text would only seem use full to me if we take specific actions after the message is shown, so that would need another kind of (modal) message box that waits for the button to be pressed. 
So I would rather implement those kind of message boxes too and those would then have a (optional?) button text parameter.
But AFAIK there are just one or two places in the current code where this would be applicable. (and one occurrence would be disappearing after the sound redesign anyway.)

src/messages.cpp
pgScorpio: > Concerning the QObject::tr() call: I think the code doesn't use it in other places? Should't we stay consistent and use tr()

Yes, but since tr is a static QObject function it is only available in a QObject, so since here we are not a QOject here using QObject::tr here is actually the same as tr in a QObject.

src/messages.cpp
pgScorpio: As it is used everywhere, because the debug streams do not support unicode?

src/messages.h
pgScorpio: QDebug includes what's neccesary for the cli streams.

src/messages.cpp
ann0see: Ok (although that's strange...)

ann0see:
I think the idea is still valuable. However to keep the PR view clean, I'll close this for now.

Issue #609, opened 2020-09-20
Manage settings with structs


pljones:
(Yes, I'm still trying to sort this out...)

My mind's eye view is that the server has a set of settings: SServerSettings.  This is _everything_ the server needs to know.

On start up, the structure is created (in main.cpp, defined in server.h).

Then, if GUI-enabled, CSettings populates it with values from the inifile - hence CSettings doesn't get passed a server instance any more - the server doesn't yet exist.

Next, all the values collected from the command line overwrite the appropriate entry in the settings structure.

And then finally, CServer gets instantiated with the SServerSettings instance reference.

To change a server setting, you call the server as you do now and it records that value in SServerSettings&.

When the server terminates, if GUI-enabled, CSettings is passed the SServerSettings reference to store to the inifile.

----

It's actually a fairly straight forward change in the server.  The only concern I have when it comes to the client code... it isn't anywhere near as simple a model.  Much of what's in the inifile isn't even for the client.  So I'd keep the SClientSettings struct "unpolluted", free of GUI settings, perhaps having a separate SClientGUISettings for those.

----

One benefit is dropping all that code I added checking for command line options to prevent the inifile being used.  The logic above handles that without any effort.

It could also help make it clear on what will persist to the inifile -- if it's not in one of the structs, it's not getting saved.

corrados:
Should we really change anything with the settings? It just works as it is right now.

pljones:
It's a "wait and see" -- the setting / command line haven't changed recently but, when I raised this a lot of change was happening.

pljones:
OK, some more thoughts.

A couple of people have wanted to decouple of client GUI from the client itself and have a more programmer-friendly way of controlling the client - effectively, all the calls to the client from the GUI and the `emit`s from the client to the GUI would need replacing with some other asynchronous interface.

Ideally - and what one of the people wanted - this would allow easy custom GUI construction, on top of that interface.

The other requirement was that remote control would be possible.

Whilst the server has _some_ support for running without a GUI, it doesn't provide a full runtime control interface, either.

So, how does this relate to my ideas on the settings?  Well, this was a kind of first step down the path - get the settings clearly separated out and managed by specific calls to update them.

I'm still half thinking "it's not going to happen"...

gene96817:
[...]

PR #2599, opened 2022-04-16
Adding global commandline parameters

pgScorpio:

Commandline parsing is now implemented in a separate class
Commandline parameters now accessible anywhere in the code.

CHANGELOG: Refactoring: Commandline parameters are now accessible anywhere in the code via the CCommandline class.

**Context: Fixes an issue?**

General improvement and preparation for sound-redesign.

No documentation changes required.

## Checklist

### DIFF
--- a/src/global.h
+++ b/src/global.h
@@ -357,20 +357,3 @@ public:
     int iStatus;
     int iChanNum;
 };
-
-/* Prototypes for global functions ********************************************/
-// command line parsing, TODO do not declare functions globally but in a class
-QString UsageArguments ( char** argv );
-
-bool GetFlagArgument ( char** argv, int& i, QString strShortOpt, QString strLongOpt );
-
-bool GetStringArgument ( int argc, char** argv, int& i, QString strShortOpt, QString strLongOpt, QString& strArg );
-
-bool GetNumericArgument ( int     argc,
-                          char**  argv,
-                          int&    i,
-                          QString strShortOpt,
-                          QString strLongOpt,
-                          double  rRangeStart,
-                          double  rRangeStop,
-                          double& rValue );
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -52,8 +52,17 @@ extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
 #ifndef SERVER_ONLY
 #    include "clientrpc.h"
 #endif
+#include "cmdline.h"
 
 // Implementation **************************************************************
+
+QString UsageArguments ( char* argv );
+
+void OnFatalError ( QString errMsg )
+{
+    qCritical() << qUtf8Printable ( errMsg );
+    exit ( 1 );
+}
 
 int main ( int argc, char** argv )
 {
@@ -63,6 +72,10 @@ int main ( int argc, char** argv )
     // Qt will not show these with underline characters in the GUI on MacOS. (#1873)
     qt_set_sequence_auto_mnemonic ( true );
 #endif
+    CCommandline::argc = argc;
+    CCommandline::argv = argv;
+
+    CCommandline cmdLine ( OnFatalError );
 
     QString        strArgument;
     double         rDbleArgument;
@@ -79,6 +92,7 @@ int main ( int argc, char** argv )
 #else
     bool bIsClient = true;
 #endif
+    bool         bSpecialOptions             = false; // Any options after this option will be accepted ! (mostly used for debugging purpouses)
     bool         bUseGUI                     = true;
     bool         bStartMinimized             = false;
     bool         bShowComplRegConnList       = false;
@@ -134,7 +148,7 @@ int main ( int argc, char** argv )
         // Help (usage) flag ---------------------------------------------------
         if ( ( !strcmp ( argv[i], "--help" ) ) || ( !strcmp ( argv[i], "-h" ) ) || ( !strcmp ( argv[i], "-?" ) ) )
         {
-            const QString strHelp = UsageArguments ( argv );
+            const QString strHelp = UsageArguments ( argv[0] );
             std::cout << qUtf8Printable ( strHelp );
             exit ( 0 );
         }
@@ -149,7 +163,7 @@ int main ( int argc, char** argv )
         // Common options:
 
         // Initialization file -------------------------------------------------
-        if ( GetStringArgument ( argc, argv, i, "-i", "--inifile", strArgument ) )
+        if ( cmdLine.GetStringArgument ( i, CMDLN_INIFILE, strArgument ) )
         {
             strIniFileName = strArgument;
             qInfo() << qUtf8Printable ( QString ( "- initialization file name: %1" ).arg ( strIniFileName ) );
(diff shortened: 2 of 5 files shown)


pgScorpio:
Better(?) implementation of the Commandline parameters part of #2538 / #2541

src/global.h
```diff
@@ -357,20 +357,3 @@ class CCustomEvent : public QEvent
...
-bool GetStringArgument ( int argc, char** argv, int& i, QString strShortOpt, QString strLongOpt, QString& strArg );
-
-bool GetNumericArgument ( int     argc,
-                          char**  argv,
-                          int&    i,
-                          QString strShortOpt,
-                          QString strLongOpt,
-                          double  rRangeStart,
-                          double  rRangeStop,
-                          double& rValue );
```
pgScorpio, line 376: From now on include "cmdline.h" to access commandline parameters via the CCommandline class

src/main.cpp
```diff
@@ -52,9 +52,18 @@ extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
...
 // Implementation **************************************************************
 
+QString UsageArguments ( char* argv );
+
+void OnFatalError ( QString errMsg )
+{
+    qCritical() << qUtf8Printable ( errMsg );
+    exit ( 1 );
+}
+
```
pgScorpio, line 66: Note that OnFatalError could be used on many places in main.cpp.

```
    qCritical() << qUtf8Printable ( ... );
    exit ( 1 );
```

Is used in many places...

src/main.cpp
```diff
@@ -1082,69 +1068,3 @@ QString UsageArguments ( char** argv )
...
-            exit ( 1 );
-        }
-
-        return true;
-    }
-    else
-    {
-        return false;
-    }
-}
```
pgScorpio, line 1150: These functions are now implemented in CCommandline.

Note that in the new implementation the 
```
qCritical() << qUtf8Printable ( ... );
exit ( 1 );
``` 
parts are replaced by the CCommandline::onArgumentError function which can, optionally, be passed in the CCommandLine constructor, since we probably don't want the program to exit on parsing wrong (--special?) parameters.

src/main.cpp
```diff
@@ -557,14 +529,28 @@ int main ( int argc, char** argv )
...
-        qCritical() << qUtf8Printable ( QString ( "%1: Unknown option '%2' -- use '--help' for help" ).arg ( argv[0] ).arg ( argv[i] ) );
+        // No exit for options after the "--special" option.
+        // Used for debugging and testing new options...
+        if ( !bSpecialOptions )
+        {
+            // Unknown option ------------------------------------------------------
+            qCritical() << qUtf8Printable ( QString ( "%1: Unknown option '%2' -- use '--help' for help" ).arg ( argv[0] ).arg ( argv[i] ) );
 
 // clicking on the Mac application bundle, the actual application
 // is called with weird command line args -> do not exit on these
```
pgScorpio, line 549: If this is true, how about all other places in main using almost the same code ?

```
qCritical() << qUtf8Printable ( QString (...) );
exit (1);
```
OnFatalError uses this code now too.

pgScorpio:
@pljones and others of the Jamulus team

Could we please have some reviews here?

Also my new settings implementation is now waiting on this one.

pgScorpio:
@ann0see @pljones  @hoffie 

I also have a CCommandlineOptions class ready that reads all commandline options, and checks client/server validity and ranges,
But also that one depends on this PR and #2598 and #2620. 
And I'm also working on a better implementation of [Settings handling](https://github.com/pgScorpio/jamulus/tree/first-big-thing-for-settings) and that one, again, depends on CCommandlineOptions, and so also on all my other PR's
I'm aware of #2617, but could we please get things going ASAP so I can proceed while I still have the time?

ann0see:
@pljones can you review these PRs please, (I think we're missing a C++ developer/reviewer at the moment)

ann0see:
Not sure how to deal with this. I think this is still valid. NOT closing for now.

pljones:
My view remains that _settings_ and _command line_ handling need to be unified and that all the handling for _client_ and _server_ properties (i.e. the values configured either by _settings_ or _command line_) should be handled through calls to the client or server instances.

ann0see:
I still think that this PR is a good starting point. We'd include the config file parsing in the provided class.

pgScorpio:
> My view remains that _settings_ and _command line_ handling need to be unified and that all the handling for _client_ and _server_ properties (i.e. the values configured either by _settings_ or _command line_) should be handled through calls to the client or server instances.
> 
> This approach enables a consistent, UI-independent, development basis going forward. The CLI and GUI - and RPC interface - would be sharing common code for client and server. Where client and server needed common code, that would be separated out. It does mean a lot more work rearchitecting Jamulus, however.

pgScorpio:
oops accidentally closed...

pljones:
> I must disagree, since the inifile is read before any module is initialized there should be no calls to the modules from reading the inifile but the modules should request the inifile values at their initialization!

This requires the client and server to know about the _inifile_, if you're explaining yourself correctly.

I'm suggesting that the client and server know about _settings_ and they "own" their own settings in the sense that any read or update goes through a client or server call.  The implementation details (getting those changes to backing store) should, naturally, be independent.

Naming the inifile would come from parsing the command line.  Loading the inifile would come next.  Instantiating the client or server would come next, passing the settings (the combined inifile and command line values).

pgScorpio:
> I'm suggesting that the client and server know about _settings_ and they "own" their own settings in the sense that any read or update goes through a client or server call. The implementation details (getting those changes to backing store) should, naturally, be independent.

I suggest we have a commandline class that reads commandline values and an inifile class containing 3 classes: app_settings, server_settings and client_settings.

First the commandline class parses the commandline and store these values. This comnandline class is passed to the inifile class which then reads the inifile values

Then the app is initialised using app_settings and client/server  is initialised using their setting classes

app, client and server retrieve their settings by calls to their settings class, which will apply any commandline overrides.

app, client and server can update their settings by calls to the matching settings class during runtime.

At app closing the settings class writes out the new inifile using the stored settings.

ann0see:
Why do you think there should be three classes instead of just one saving the state of CLI/ini file/etc?

ann0see:
I'd rename the commandline class to config as it's not just saving the CLI.

pgScorpio:
> I'd rename the commandline class to config as it's not just saving the CLI.

It should just parse the CLI!

And config is way more than the commandline so it would be more logical if "inifile" was named config as it should contain all configurable settings (inifile as well as commandline).

The commandline can also contain commandline only (i.e. server/client, logging, debugging) parameters and, as so, it should be a separate class accessible anywhere.

ann0see:
Hmm. I mean in the end having one class serving the whole configuration state/providing access to CLI/GUI/Commandline would be beneficial as this allows logical separation.

Do you think that's an issue?

pgScorpio:
> Hmm. I mean in the end having one class serving the whole configuration state/providing access to CLI/GUI/Commandline would be beneficial as this allows logical separation.
> 
> Do you think that's an issue?

Why putting everything in one class?

I think separate classes are the perfect way to keep things organised, also they can be defined at different, logical locations.
Nevertheless don't confuse classes and class instances. The instances of multiple classes can still be located in one class instance.

Though I still think "commandline" and "inifile" should be different instances since the commandline should be parsed before the other instances are created, but even then "config" could be a class containing commandline and inifile...

P.S. Shouldn't we discuss this in another topic ?

ann0see:
Yes. Probably in https://github.com/jamulussoftware/jamulus/issues/609

pljones:
Setting to draft as branch needs conflict resolution.

ann0see:
Closing for now. The ideas are still present.

PR #2618, opened 2022-04-26
Fixing memory leaks.

### LINKED ISSUE #2614: Memory leak when using Rpc server
**Describe the bug**
CServerRpc and CClientRpc/CServerRpc  are created with "new" in main.cpp, but are never deleted.

**To Reproduce**
Run Jamulus with and without valid --jsonrpcport and --jsonrpcsecretfile parameters.
Use Visual studio's memory usage diagnostic tool to check the difference in memory usage at exit.

**Expected behavior**
Cleanup on exit.

**Screenshots**

**Operating system**
n.a.

**Version of Jamulus**
master @ 953f2b8a1ccab70d715a54cc730e05059ae140d6

**Additional context**

pgScorpio:
Moved pointers to with new assigned objects outside main and added a
cleanup function that is called before every exit to delete those objects again.

CHANGELOG: 
Fixed some memory leaks.

**Context: Fixes an issue?**
Fixes #2614 

No documentation changes

Bugfix

Ready to merge.

## Checklist

-  [-] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -808,7 +808,7 @@ int main ( int argc, char** argv )
     // Application/GUI setup ---------------------------------------------------
     // Application object
 #ifdef HEADLESS
-    QCoreApplication* pApp = new QCoreApplication ( argc, argv );
+    pApp = new QCoreApplication ( argc, argv );
 #else
 #    if defined( Q_OS_IOS )
     bUseGUI        = true;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -59,8 +59,10 @@ extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
 //       should be declared here and checked and deleted in reverse order of creation in cleanup().
 //       cleanup() should be called before any exit() !
 
-QCoreApplication* pApp       = NULL;
-CRpcServer*       pRpcServer = NULL;
+QCoreApplication* pCoreApplicationInstance = NULL;
+QApplication*     pApplicationInstance     = NULL;
+
+CRpcServer* pRpcServer = NULL;
 #ifndef SERVER_ONLY
 CClientRpc* pClientRpc = NULL;
 #endif
@@ -88,10 +90,16 @@ void cleanup()
         pRpcServer = NULL;
     }
 
-    if ( pApp )
+    if ( pCoreApplicationInstance )
     {
-        delete pApp;
-        pApp = NULL;
+        delete pCoreApplicationInstance;
+        pCoreApplicationInstance = NULL;
+    }
+
+    if ( pApplicationInstance )
+    {
+        delete pApplicationInstance;
+        pApplicationInstance = NULL;
     }
 }
 
@@ -808,16 +816,24 @@ int main ( int argc, char** argv )
     // Application/GUI setup ---------------------------------------------------
     // Application object
 #ifdef HEADLESS
-    pApp = new QCoreApplication ( argc, argv );
+    QCoreApplication* pApp = pCoreApplicationInstance = new QCoreApplication ( argc, argv );
 #else
 #    if defined( Q_OS_IOS )
     bUseGUI        = true;
     bIsClient      = true; // Client only - TODO: maybe a switch in interface to change to server?
 
     // bUseMultithreading = true;
-    pApp = new QApplication ( argc, argv );
+    QApplication* pApp = pApplicationInstance = new QApplication ( argc, argv );
 #    else
-    pApp = bUseGUI ? new QApplication ( argc, argv ) : new QCoreApplication ( argc, argv );
+    QCoreApplication* pApp;
+    if ( bUseGUI )
+    {
+        pApp = pApplicationInstance = new QApplication ( argc, argv );
+    }
+    else
+    {
+        pApp = pCoreApplicationInstance = new QCoreApplication ( argc, argv );
+    }
 #    endif
 #endif
 
@@ -1048,17 +1064,13 @@ int main ( int argc, char** argv )
 #endif
         {
             qCritical() << qUtf8Printable ( QString ( "%1: %2" ).arg ( APP_NAME ).arg ( generr.GetErrorText() ) );
+        }
 
-            if ( exit_code == 0 )
-            {
-                exit_code = 1;
-            }
+        if ( exit_code == 0 )
+        {
+            exit_code = 1;
         }
     }
-
-#if defined( Q_OS_MACX )
-    activity.EndActivity();
-#endif
 
     cleanup();
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -61,8 +61,10 @@ extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
 
 QCoreApplication* pApp       = NULL;
 CRpcServer*       pRpcServer = NULL;
-CClientRpc*       pClientRpc = NULL;
-CServerRpc*       pServerRpc = NULL;
+#ifndef SERVER_ONLY
+CClientRpc* pClientRpc = NULL;
+#endif
+CServerRpc* pServerRpc = NULL;
 
 void cleanup()
 {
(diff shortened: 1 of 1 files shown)


ann0see:
@dtinth could you please Review?

ann0see:
@pgScorpio thank you again ;-)

src/main.cpp
```diff
@@ -999,15 +1044,21 @@ int main ( int argc, char** argv )
 #endif
         {
             qCritical() << qUtf8Printable ( QString ( "%1: %2" ).arg ( APP_NAME ).arg ( generr.GetErrorText() ) );
-            exit ( 1 );
+
+            if ( exit_code == 0 )
```
ann0see, line 1048: Looks a bit odd in my eyes...

ann0see (commented):
Not sure if that's the standard C++ clean up way? It looks a bit strange to me (but I have a Java and PHP background)

pgScorpio:
> Not sure if that's the standard C++ clean up way? It looks a bit strange to me (but I have a Java and PHP background)

Well.... what's the standard C++ cleanup way ?
The only "standard" is that every function should delete all allocated data before exiting. This is best achieved by using class instances, where the destructor cleans up everything when the instance goes out of scope. Unfortunately, because of the way things are implemented, there is no clear scope here.

src/main.cpp
pgScorpio: What looks strange to you ?

This is the "normal" way of exception handling in main.
catch the exception, set the exit code, and exit main normally with that exit code....

Though what looks strange to me is that the try block covers only part of the main() code. and that that are so many exit() calls. The "normal" way to exit main from main would be to throw an exception. In that case the cleanup can be done just before exiting main. And that would be the better solution, but that would also have a lot more code changed....

src/main.cpp
```diff
@@ -768,9 +811,9 @@ int main ( int argc, char** argv )
     bIsClient      = true; // Client only - TODO: maybe a switch in interface to change to server?
 
     // bUseMultithreading = true;
-    QApplication* pApp = new QApplication ( argc, argv );
+    pApp = new QApplication ( argc, argv );
```
hoffie, line 814: So, for iOS apps this used to be of type `QApplication` instead of `QCoreApplication`. Can you give a quick explanation why this is safe to do?

It's probably hard or artificial to put that into an individual commit, but at least the commit description should mention that this was intentional and why it is considered safe (I guess because iOS doesn't use any methods from the broader interface either?).

src/main.cpp
```diff
@@ -55,6 +55,42 @@ extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
...
+    {
+        delete pServerRpc;
+        pServerRpc = NULL;
+    }
+
+    if ( pClientRpc )
+    {
+        delete pClientRpc;
+        pClientRpc = NULL;
+    }
```
hoffie, line 79: Needs `#ifndef HEADLESS` (see Debian build failures).

src/main.cpp
```diff
@@ -55,6 +55,42 @@ extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
 
 // Implementation **************************************************************
 
+// NOTE: To prevent memory leaks any pointers assigned with new
+//       should be declared here and checked and deleted in reverse order of creation in cleanup().
+//       cleanup() should be called before any exit() !
+
+QCoreApplication* pApp       = NULL;
+CRpcServer*       pRpcServer = NULL;
+CClientRpc*       pClientRpc = NULL;
```
hoffie, line 64: Needs `#ifndef HEADLESS` (see Debian build failures).

src/main.cpp
```diff
@@ -55,6 +55,42 @@ extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
 
 // Implementation **************************************************************
 
+// NOTE: To prevent memory leaks any pointers assigned with new
+//       should be declared here and checked and deleted in reverse order of creation in cleanup().
+//       cleanup() should be called before any exit() !
```
hoffie, line 60: I'm wondering if it was harder to forget calling cleanup if we had a generic `CleanExit(int)` function?

src/main.cpp
ann0see: The whole handling here - as you said. Mostly the "if exit_code == 0 then set exit code to 1" thing. I'd really prefer the exit ( ) function to be extended with cleanup() but I assume exit() is not a method of a class, so you'd write a function calling exit instead.

src/main.cpp
```diff
@@ -63,6 +99,7 @@ int main ( int argc, char** argv )
     // Qt will not show these with underline characters in the GUI on MacOS. (#1873)
     qt_set_sequence_auto_mnemonic ( true );
 #endif
+    int exit_code = 0;
```
hoffie, line 102: This is about improving return code correctness, right? At least I don't see first-hand how it is related to the memory leak fix.

I guess it's halfway OK to do it in this PR, but it should be an individual commit.

src/main.cpp
pgScorpio: Correct. exit() is a C/C++ library function, and is not the best way to end the program when using exceptions. In that case throwing special exceptions to exit would be the better way.

src/main.cpp
pgScorpio: ???

I didn't change the type, I just moved the pointer declaration outside main() to be able to implement the cleanup() function.

src/main.cpp
pgScorpio: Yes! I missed that. already corrected....

src/main.cpp
pgScorpio: > I'm wondering if it was harder to forget calling cleanup if we had a generic `CleanExit(int)` function?

Yes I thought about that too, but then I could not use it for the normal exit.
I would prefer using an exception instead of using exit() even more...
(Note that calling exit() instead of normally returning from main with a return code may skip some cleanup in windows itself, and maybe it's the same on other OS's, so this increases the chance on memory leaks.)

src/main.cpp
hoffie: Yeah, I understand the reason for moving. I just wanted to hint at the side effect (maybe it's a noop, but I'm not sure):
`pApp` was previously either of type `QApplication` or `QCoreApplication` based on platform and/or GUI/non-GUI.
With the new code, it's always a `QCoreApplication` which is probably the lowest common denominator. It might be fine, but it's still a change for iOS, isn't it?

src/main.cpp
pgScorpio: Correct.
I added exit_code to be able to do the cleanup and a normal return on an exception.

You are correct I now also use the pApp->exec() result so when exiting via QApplication also the correct exit code is returned by main, and that is indeed not directly related to the problem.

src/main.cpp
pgScorpio: Same here: needed #ifdef SERVER_ONLY!

src/main.cpp
pgScorpio: I'm still wondering why you think anything changed for IOS, AFAIK It's was, and still is, always QApplication for IOS.

```
#    if defined( Q_OS_IOS )
    bUseGUI        = true;
    bIsClient      = true; // Client only - TODO: maybe a switch in interface to change to server?

    // bUseMultithreading = true;
    pApp = new QApplication ( argc, argv );
#    else

```

pgScorpio:
I opened a second PR #2620 on the same issue.
That one is the solution using exceptions.
Please pick your solution of choice.
As said I prefer the exception solution...

src/main.cpp
pgScorpio: OK, Now I see your point. Thanks for noting!

QCoreApplication is the base type of all QApplications, that's why I can use that type to store the pointer. 
But I expected yhe basetype to have virtual functions, but now I checked it and that isn't the case... 
Though all QApplications will call QCoreApplication::exec() eventually there might be a difference in behavior, so I will have to find a better solution for this.

pgScorpio:
I'm closing this PR in favor of #2620

Too many problems with this solution and it also won't work when `exit(1)` is used in other places, like in PR #2599
I'm also working on a CCommandlineParameters class and new settings handling which all need exception handling to solve this problem.

PR #2620, opened 2022-04-26
Fixing memory leaks using exceptions

### LINKED ISSUE #2614: Memory leak when using Rpc server

pgScorpio:

Fix some memory leaks.
Alternative for PR #2618 using exceptions to cleanly exit the application.
(#2618 won't solve the problem when still using exit() or on other unhandled exceptions.)

CHANGELOG: 
Fixed some memory leaks in main.

**Context: Fixes an issue?**

Fixes #2614

No documentation changes

Bugfix

Review

## Checklist

-  [-] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 

### DIFF
--- a/src/global.h
+++ b/src/global.h
@@ -343,6 +343,31 @@ protected:
     QString strErrorType;
 };
 
+class CInfoExit
+{
+public:
+    CInfoExit ( QString strMsg ) : strInfoMsg ( strMsg ) {}
+
+    QString GetInfoMessage() const { return strInfoMsg; }
+
+protected:
+    QString strInfoMsg;
+};
+
+class CErrorExit
+{
+public:
+    CErrorExit ( int exitcode, QString strNewErrorMsg ) : iExitcode ( exitcode ), strErrorMsg ( strNewErrorMsg ) {}
+
+    QString GetErrorMessage() const { return strErrorMsg; }
+
+    int GetExitCode() const { return iExitcode; }
+
+protected:
+    QString strErrorMsg;
+    int     iExitcode;
+};
+
 class CCustomEvent : public QEvent
 {
 public:
--- a/src/global.h
+++ b/src/global.h
@@ -319,29 +319,7 @@ typedef unsigned char      uint8_t;
 // definition for custom event
 #define MS_PACKET_RECEIVED 0
 
-/* Classes ********************************************************************/
-class CGenErr
-{
-public:
-    CGenErr ( QString strNewErrorMsg, QString strNewErrorType = "" ) : strErrorMsg ( strNewErrorMsg ), strErrorType ( strNewErrorType ) {}
-
-    QString GetErrorText() const
-    {
-        // return formatted error text
-        if ( strErrorType.isEmpty() )
-        {
-            return strErrorMsg;
-        }
-        else
-        {
-            return strErrorType + ": " + strErrorMsg;
-        }
-    }
-
-protected:
-    QString strErrorMsg;
-    QString strErrorType;
-};
+/* Exception Classes **********************************************************/
 
 class CInfoExit
 {
@@ -357,16 +335,48 @@ protected:
 class CErrorExit
 {
 public:
-    CErrorExit ( int exitcode, QString strNewErrorMsg ) : iExitcode ( exitcode ), strErrorMsg ( strNewErrorMsg ) {}
+    CErrorExit ( QString strNewErrorMsg, int iNewExitCode = 1 ) : strErrorMsg ( strNewErrorMsg ), iExitCode ( iNewExitCode ) {}
 
     QString GetErrorMessage() const { return strErrorMsg; }
 
-    int GetExitCode() const { return iExitcode; }
+    int GetExitCode() const { return iExitCode; }
 
 protected:
     QString strErrorMsg;
-    int     iExitcode;
+    int     iExitCode;
 };
+
+class CGenErr
+{
+public:
+    CGenErr ( QString strNewErrorMsg, QString strNewErrorType = "", int iNewExitCode = 1 ) :
+        strErrorMsg ( strNewErrorMsg ),
+        strErrorType ( strNewErrorType ),
+        iExitCode ( iNewExitCode )
+    {}
+
+    QString GetErrorText() const
+    {
+        // return formatted error text
+        if ( strErrorType.isEmpty() )
+        {
+            return strErrorMsg;
+        }
+        else
+        {
+            return strErrorType + ": " + strErrorMsg;
+        }
+    }
+
+    int GetExitCode() const { return iExitCode; }
+
+protected:
+    QString strErrorMsg;
+    QString strErrorType;
+    int     iExitCode;
+};
+
+/* Event Classes **************************************************************/
 
 class CCustomEvent : public QEvent
 {
@@ -384,6 +394,9 @@ public:
 };
 
 /* Prototypes for global functions ********************************************/
+
+extern const QString GetAppName();
+
 // command line parsing, TODO do not declare functions globally but in a class
 QString UsageArguments ( char** argv );
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1028,14 +1028,6 @@ int main ( int argc, char** argv )
         exit_code = 0;
     }
 
-#if defined( Q_OS_MACX )
-    // a mac activity is started after Application is created
-    if ( pApp )
-    {
-        activity.EndActivity();
-    }
-#endif
-
     if ( pServerRpc )
     {
         delete pServerRpc;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -808,7 +808,7 @@ int main ( int argc, char** argv )
     // Application/GUI setup ---------------------------------------------------
     // Application object
 #ifdef HEADLESS
-    QCoreApplication* pApp = new QCoreApplication ( argc, argv );
+    pApp = new QCoreApplication ( argc, argv );
 #else
 #    if defined( Q_OS_IOS )
     bUseGUI        = true;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -24,7 +24,7 @@
 
 #ifdef HEADLESS
 #    if defined( Q_OS_IOS )
-#        error HEADLES mode is not valid for IOS
+#        error HEADLESS mode is not supported on iOS
 #    endif
 #    if defined( ANDROID )
 #        error HEADLES mode is not valid for ANDROID
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -27,7 +27,7 @@
 #        error HEADLESS mode is not supported on iOS
 #    endif
 #    if defined( ANDROID )
-#        error HEADLES mode is not valid for ANDROID
+#        error HEADLESS mode is not supported on Android
 #    endif
 #endif
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -55,26 +55,26 @@ extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
 
 // Implementation **************************************************************
 
-// NOTE: To prevent memory leaks any pointers assigned with new
-//       should be declared here and checked and deleted in reverse order of creation in cleanup().
-//       cleanup() should be called before any exit() !
-
 int main ( int argc, char** argv )
 {
-
 #if defined( Q_OS_MACX )
     // Mnemonic keys are default disabled in Qt for MacOS. The following function enables them.
     // Qt will not show these with underline characters in the GUI on MacOS. (#1873)
     qt_set_sequence_auto_mnemonic ( true );
 #endif
-    int exit_code = 0;
+    // NOTE: To prevent memory leaks any pointers assigned with new in main should be
+    //       declared here and checked and deleted in reverse order of creation before main exit.
+    //       do NEVER call exit() but throw a CErrorExit, CInfoExit or a CGenErr !!
 
-    QCoreApplication* pApp       = NULL;
-    CRpcServer*       pRpcServer = NULL;
+    QCoreApplication* pCoreApplicationInstance = NULL;
+    QCoreApplication* pApplicationInstance     = NULL;
+    CRpcServer*       pRpcServer               = NULL;
 #ifndef SERVER_ONLY
     CClientRpc* pClientRpc = NULL;
 #endif
     CServerRpc* pServerRpc = NULL;
+
+    int exit_code = 0;
 
     QString        strArgument;
     double         rDbleArgument;
@@ -767,16 +767,24 @@ int main ( int argc, char** argv )
         // Application/GUI setup ---------------------------------------------------
         // Application object
 #ifdef HEADLESS
-        pApp = new QCoreApplication ( argc, argv );
+        QCoreApplication* pApp = pCoreApplicationInstance = new QCoreApplication ( argc, argv );
 #else
 #    if defined( Q_OS_IOS )
         bUseGUI   = true;
         bIsClient = true; // Client only - TODO: maybe a switch in interface to change to server?
 
         // bUseMultithreading = true;
-        pApp = new QApplication ( argc, argv );
+        QApplication* pApp = pApplicationInstance = new QApplication ( argc, argv );
 #    else
-        pApp = bUseGUI ? new QApplication ( argc, argv ) : new QCoreApplication ( argc, argv );
+        QCoreApplication* pApp;
+        if ( bUseGUI )
+        {
+            pApp = pApplicationInstance = new QApplication ( argc, argv );
+        }
+        else
+        {
+            pApp = pCoreApplicationInstance = new QCoreApplication ( argc, argv );
+        }
 #    endif
 #endif
 
@@ -1000,11 +1008,11 @@ int main ( int argc, char** argv )
 #endif
         {
             qCritical() << qUtf8Printable ( QString ( "%1: %2" ).arg ( APP_NAME ).arg ( generr.GetErrorText() ) );
+        }
 
-            if ( exit_code == 0 )
-            {
-                exit_code = 1;
-            }
+        if ( exit_code == 0 )
+        {
+            exit_code = 1;
         }
     }
 
@@ -1031,27 +1039,28 @@ int main ( int argc, char** argv )
     if ( pServerRpc )
     {
         delete pServerRpc;
-        pServerRpc = NULL;
     }
 
 #ifndef SERVER_ONLY
     if ( pClientRpc )
     {
         delete pClientRpc;
-        pClientRpc = NULL;
     }
 #endif
 
     if ( pRpcServer )
     {
         delete pRpcServer;
-        pRpcServer = NULL;
     }
 
-    if ( pApp )
+    if ( pCoreApplicationInstance )
     {
-        delete pApp;
-        pApp = NULL;
+        delete pCoreApplicationInstance;
+    }
+
+    if ( pApplicationInstance )
+    {
+        delete pApplicationInstance;
     }
 
     return exit_code;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -67,7 +67,7 @@ int main ( int argc, char** argv )
     //       do NEVER call exit() but throw a CErrorExit, CInfoExit or a CGenErr !!
 
     QCoreApplication* pCoreApplicationInstance = NULL;
-    QCoreApplication* pApplicationInstance     = NULL;
+    QApplication*     pApplicationInstance     = NULL;
     CRpcServer*       pRpcServer               = NULL;
 #ifndef SERVER_ONLY
     CClientRpc* pClientRpc = NULL;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -59,42 +59,6 @@ extern void qt_set_sequence_auto_mnemonic ( bool bEnable );
 //       should be declared here and checked and deleted in reverse order of creation in cleanup().
 //       cleanup() should be called before any exit() !
 
-QCoreApplication* pApp       = NULL;
-CRpcServer*       pRpcServer = NULL;
-#ifndef SERVER_ONLY
-CClientRpc* pClientRpc = NULL;
-#endif
-CServerRpc* pServerRpc = NULL;
-
-void cleanup()
-{
-    if ( pServerRpc )
-    {
-        delete pServerRpc;
-        pServerRpc = NULL;
-    }
-
-#ifndef SERVER_ONLY
-    if ( pClientRpc )
-    {
-        delete pClientRpc;
-        pClientRpc = NULL;
-    }
-#endif
-
-    if ( pRpcServer )
-    {
-        delete pRpcServer;
-        pRpcServer = NULL;
-    }
-
-    if ( pApp )
-    {
-        delete pApp;
-        pApp = NULL;
-    }
-}
-
 int main ( int argc, char** argv )
 {
 
@@ -104,6 +68,13 @@ int main ( int argc, char** argv )
     qt_set_sequence_auto_mnemonic ( true );
 #endif
     int exit_code = 0;
+
+    QCoreApplication* pApp       = NULL;
+    CRpcServer*       pRpcServer = NULL;
+#ifndef SERVER_ONLY
+    CClientRpc* pClientRpc = NULL;
+#endif
+    CServerRpc* pServerRpc = NULL;
 
     QString        strArgument;
     double         rDbleArgument;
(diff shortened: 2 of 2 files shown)


src/main.cpp
```diff
@@ -115,748 +127,737 @@ int main ( int argc, char** argv )
...
-#endif
-
-    // When adding new options, follow the same order as --help output
-
-    // QT docu: argv()[0] is the program name, argv()[1] is the first
-    // argument and argv()[argc()-1] is the last argument.
-    // Start with first argument, therefore "i = 1"
-    for ( int i = 1; i < argc; i++ )
+    try
     {
```
pgScorpio, line 132: Sorry, but because of moving this try, the indenting changed an a large part of the code and now the diff looks very confusing...

ann0see:
@dtinth could you please Review this PR?

src/main.cpp
npostavs: There's a "Hide whitespace" option for Github, and for git there is [`git diff -w`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt--w)

> `-w`
> `--ignore-all-space`
>
>    Ignore whitespace when comparing lines. This ignores differences even if one line has whitespace where the other line has none.


image

src/global.h
```diff
@@ -319,11 +319,41 @@ typedef unsigned char      uint8_t;
 // definition for custom event
 #define MS_PACKET_RECEIVED 0
 
-/* Classes ********************************************************************/
+/* Exception Classes **********************************************************/
+
+class CInfoExit
```
ann0see, line 324: Although not related - maybe more to the error message redesign, but wouldn't it be better to have a class
CException
with sub-classes CInfoExit, CGenErr etc?

src/main.cpp
```diff
@@ -22,10 +22,23 @@
  *
 \******************************************************************************/
 
+#ifdef HEADLESS
+#    if defined( Q_OS_IOS )
+#        error HEADLES mode is not valid for IOS
```
ann0see, line 27:
```suggestion
#        error HEADLESS mode is not supported on iOS
```

src/main.cpp
```diff
@@ -22,10 +22,23 @@
...
+#    endif
+#    if defined( ANDROID )
+#        error HEADLES mode is not valid for ANDROID
```
ann0see, line 30:
```suggestion
#        error HEADLESS mode is not supported on Android
```

src/main.cpp
```diff
@@ -22,10 +22,23 @@
...
+#    endif
+#endif
+
 #include <QCoreApplication>
 #include <QDir>
 #include <iostream>
 #include "global.h"
+#include "settings.h"
+#include "util.h"
+#include <memory>
```
ann0see, line 40: Did you just move code around here? Why?

src/main.cpp
```diff
@@ -34,36 +47,74 @@
...
+    }
+
+    return QString();
+}
 
 // Implementation **************************************************************
 
 int main ( int argc, char** argv )
 {
+    int exit_code = 0;
```
ann0see, line 93: Give a comment here why we need the variable. (I still don't like having an exit code variable here...)

src/main.cpp
```diff
@@ -115,748 +160,765 @@ int main ( int argc, char** argv )
...
+#if defined( Q_OS_IOS )
+        bIsClient = true; // Client only - TODO: maybe a switch in interface to change to server?
+        bUseGUI   = true;
+        // bUseMultithreading = true;
+#endif
+
+        // TODO create settings in default state, if loading from file do that next, then come back here to
+        //      override from command line options, then create client or server, letting them do the validation
+        //
+        // See https://github.com/pgScorpio/jamulus/tree/first-big-thing-for-settings for a possible solution on that
```
ann0see, line 642: Please remove this comment since it's just a short therm one. I hope in future we'd have a PR.

src/main.cpp
```diff
@@ -115,748 +160,765 @@ int main ( int argc, char** argv )
...
-            qCritical() << qUtf8Printable ( QString ( "JSON-RPC: Refusing to run with secret of length %1 (required: %2). Exiting." )
-                                                .arg ( strJsonRpcSecret.length() )
-                                                .arg ( JSON_RPC_MINIMUM_SECRET_LENGTH ) );
-            exit ( 1 );
-        }
+            QFile qfJsonRpcSecretFile ( strJsonRpcSecretFileName );
+            if ( !qfJsonRpcSecretFile.open ( QFile::OpenModeFlag::ReadOnly ) )
+            {
+                throw CErrorExit (
+                    qUtf8Printable ( QString ( "- JSON-RPC: Unable to open secret file %1. Exiting." ).arg ( strJsonRpcSecretFileName ) ) );
```
ann0see, line 900: Why is a new line needed here? Does clang-format complain otherwise?

src/main.cpp
```diff
@@ -903,15 +965,15 @@ int main ( int argc, char** argv )
 
                 // show dialog
```
ann0see, line 966: Why isn't the exit-code variable here?

ann0see (commented):
The diff isn't that clear on GitHub... But see some comments.

src/global.h
pgScorpio: @ann0see 

> "wouldn't it be better to have a class CException"

No, unfortunalely that's not the normal way exceptions are handled, but we already have the CGenErr class for Jamulus specific errors. (And CException often already exists as a standard class in libraries, i.e. in MFC)

Especially InfoExit should not be be a generic exception, since it's not an error and should be used with care (for now only in main/commandline parameter handling..)

Also ErrorExit should not be a generic exception, because we want that one always to propagate to main. So if catched anywhere else it should always be re-thrown. (So if it was a CGenErr we would have to adapt all code where a CGenErr is catched now.)

Proper exception handling in Jamulus would be a completely separate issue anyway since most exceptions are not handled at all in Jamulus, and that could be a big challenge since there probably will be different exceptions for the different OS's.

src/main.cpp
pgScorpio: To keep things more clear....
All those ifdef's are confusing enough, so whenever possible let's keep common things together...

src/main.cpp
pgScorpio: Hmmm Why not ? Its the return value of main, and it's quite common practice to declare the returned value as first in a function.

src/main.cpp
pgScorpio: The TODO is not my comment....
And I added the link to my solution in another attempt to get the settings discussion going... (After several failed attempts.)
Nevertheless, there are still a lot PR to merge in main, so I will remove it eventually... ;-)

src/main.cpp
pgScorpio: What new line ? (are you viewing the diff in split mode ?)
The newline after `CErrorExit (` is added by clang-format, the others are due to the split format diff...

src/main.cpp
pgScorpio: Since here main always returns 0 and we only did set an exit code with `exit(1)`, but in the new situation we wanna do the clean up, so we don't wanna use `exit()`, but we will return from main normally, so main must return the exit code.

atsampson:
Can I suggest a simpler solution that combines ideas from both of your approaches?

The only uses of `exit` outside of `main` are in the option-parsing functions, which are only used before any of the dynamic allocations happen anyway. So turning the `exit` calls in `main` into straightforward `return`s will guarantee that stack-allocated objects in main are destroyed properly.

You can then convert the pointer variables into smart pointers, adding new smart pointers for the RPC classes that didn't have them already as you've done above.

I've put a quick attempt at this on the [atsampson:scopedexit branch](https://github.com/jamulussoftware/jamulus/compare/master...atsampson:scopedexit). I haven't tested this with all the combinations of configuration options but it should give you an idea of what I mean...

pgScorpio:
@atsampson 

> Can I suggest a simpler solution that combines ideas from both of your approaches?

Thanks for the suggestion !
But...
For the current code that would be almost right. Though `exit(1)` is also used in the commandline parsing functions, and `return` can't be used there. Also there a several PR's pending that would also need an exception solution, that's why I changed to using exceptions.

atsampson:
> For the current code that would be almost right. Though exit(1) is also used in the commandline parsing functions, and return can't be used there.

That's fine, because the option-parsing functions are only used before any of the dynamic allocation happens. There's a comment noting this in my changes.

> BTW Where are the with new allocated instances deleted ???

[`QScopedPointer`](https://doc.qt.io/qt-5/qscopedpointer.html) is a smart pointer class; it's roughly Qt's equivalent of [`std::unique_ptr`](https://en.cppreference.com/w/cpp/memory/unique_ptr). It deletes the pointer it's holding automatically when it goes out of scope, so the object will be destroyed correctly when the function exits (either via `return` or through an exception). If you haven't encountered smart pointers before, I'd definitely recommend reading up on them - there's rarely a good reason to use `delete` in modern C++.

pgScorpio:
> [`QScopedPointer`](https://doc.qt.io/qt-5/qscopedpointer.html) is a smart pointer class; it's roughly Qt's equivalent of [`std::unique_ptr`](https://en.cppreference.com/w/cpp/memory/unique_ptr). It deletes the pointer it's holding automatically when it goes out of scope, so the object will be destroyed correctly when the function exits (either via `return` or through an exception)

Yes I know smart pointers, I even considered using them. But they have some disadvantages too.
1: They only work when there is a well defined scope. And the rpc classes don't have a well defined scope in the current implementation.
2: No garantee in wich order they are deleted if they are in the same scope (often same order as creation). And since the pointers are cross referenced we need to delete them in a specific order (reverse order of creation).

So I still think the exception solution is the simplest...

atsampson:
> No garantee in wich order they are deleted if they are in the same scope (often same order as creation).

No, that's not correct. In C++, variables with automatic storage duration (i.e. locals) are guaranteed to be destroyed in **reverse** order of construction - see [the language standard](https://timsong-cpp.github.io/cppwp/n4868/stmt.jump#general-2) or [the ISO C++ FAQ](https://isocpp.org/wiki/faq/dtors). If that wasn't the case, then lots of RAII idioms wouldn't work.

> And the rpc classes don't have a well defined scope in the current implementation.

... which means they do have a well-defined lifetime if you convert them to smart pointers - they will be deleted in reverse order of the declarations when `main` exits. You want to ensure that the RPC objects are deleted before the objects they're controlling, which matches the declaration order in the current code.

There's absolutely nothing wrong with using exceptions to cause a clean exit from elsewhere in the code - one of the reasons smart pointers were added to the standard library was to simplify memory management when exceptions are being used, and if you're adding more use of exceptions then you may well find that there are other places in the code where you can replace manual memory management with smart pointers to avoid complex cleanup code.

pgScorpio:
@atsampson 
Thanks for the information.
If this is true (though I encountered several occasions where it didn't work this way) we could use smart pointers here to do the cleanup (and so it also should be done on a lot of other places in the code), 
but see the note:

`Note [2] However, the program can be terminated (by calling std​::​exit() or std​::​abort() for example) without destroying objects with automatic storage duration — end note]`

So we still should use exceptions instead of exit. I don't like to rely on the fact that no pointers are assigned before using exit. This might (and will) change in the future, so it would be safer to state "NEVER use exit but use the CErrorExit exception.

src/global.h
ann0see: > could be a big challenge since there probably will be different exceptions for the different OS's.

Fair point. So "not part of this PR" then.

src/main.cpp
ann0see: > Its the return value of main, and it's quite common practice to declare the returned value as first in a function.

But I don't see any other places in Jamulus where functions have this?

src/main.cpp
ann0see: Still the link seems to be new. I'd rather NOT have future fixes in the code. Either fix them right away - that's out of scope of this PR - or document it on GitHub.

src/main.cpp
```diff
@@ -115,748 +160,765 @@ int main ( int argc, char** argv )
...
-#    endif
-#endif
+        if ( bUseGUI )
+        {
+            pApplication = new QApplication ( argc, argv );
+        }
+        else
+        {
+            pCoreApplication = new QCoreApplication ( argc, argv );
+        }
```
ann0see, line 835: Leaving this section up to @hoffie

src/main.cpp
ann0see: Yes. Probably that was GitHub...

src/main.cpp
pgScorpio: > But I don't see any other places in Jamulus where functions have this?

Perhaps you have to look better than ;=))

Just do a search on `return ret` , `return res` or `return result` and you will already find a bunch of them...
Or search for `return _id;` since also all qt_metacall moc_ functions use this coding pattern!
Or take a look look at:

`CChannel::SetSockBufNumFrames`
`CChannel::PutAudioData`
`CChannel::GetData`
`CProtocol::GetValFromStream`
`CNetBuf::Get`
`CNetBufWithStats::Get`
`CBuffer::GetAvailData()`
`CClient::GetAndResetbJitterBufferOKFlag()`
`CServer::CreateChannelList()`
`CServer::FindChannel`
`CServer::PutAudioData`
`CServer::CreateLevelsForAllConChannels`
`CSettings::GetIniSetting`
`CSettings::GetNumericIniSet`
`CSettings::GetFlagIniSet`
...
All the same pattern, just different variable names....

This really is a standard coding practice....

src/main.cpp
pgScorpio: > Still the link seems to be new. 

Yes. As I said "I added the link"... "in another attempt to get the settings discussion going"

But I get your point... I hope you get my point too. ;=))

ann0see:
@atsampson @pgScorpio what's the agreement on the other discussed solution I think @pgScorpio wanted to use exceptions?

pgScorpio:
> @atsampson @pgScorpio what's the agreement on the other discussed solution I think @pgScorpio wanted to use exceptions?

Correct. Since the exception solution will also work where normally `exit()` would be used elsewhere, while the other method only works when exiting from `main()`

atsampson:
@ann0see The only thing I'm really uncomfortable about here is the manual cleanup code using `delete` - it really should be using smart pointers in modern C++. I'm fine with adding an exception specifically to handle the option parsing errors - there's no advantage in terms of the original fault, but it does simplify the code in the parsing functions. If the exception rework goes beyond that, then I think it would be better to review it later when it's needed.

pgScorpio:
> The only thing I'm really uncomfortable about here is the manual cleanup code using `delete` - it really should be using smart pointers in modern C++. 

I agree that using smart pointers would be a better solution, but AFAIK, besides in CSignalHandler, it's nowhere else used in the code, but though not specifically the solution for the original problem, it could be implemented here as well, though I think that should better be a separate PR, replacing all pointers with smart pointers...


> I'm fine with adding an exception specifically to handle the option parsing errors - there's no advantage in terms of the original fault, but it does simplify the code in the parsing functions. 

> If the exception rework goes beyond that, then I think it would be better to review it later when it's needed.

And as you say yourself, it's already used in the option parsing functions... And this will be even more neccesay if we implement the commandline options class(es).

ann0see:
Ok. So @atsampson @pgScorpio could you please open an issue on that?
What's the state of this PR? Can @atsampson do a review so that we are a bit closer to a merge?

src/main.cpp
```diff
@@ -34,36 +47,74 @@
...
 {
+    int exit_code = 0;
 
 #if defined( Q_OS_MACX )
     // Mnemonic keys are default disabled in Qt for MacOS. The following function enables them.
     // Qt will not show these with underline characters in the GUI on MacOS. (#1873)
     qt_set_sequence_auto_mnemonic ( true );
 #endif
 
+#ifdef HEADLESS
```
atsampson, line 101: Remove this, and just replace the existing pointer declarations with `QScopedPtr` (or `std::unique_ptr`), keeping them in the same places as the original code. There is no need for the `#define`s - `pApplication` can just be a pointer to a `QCoreApplication`, as `QCoreApplication` has (as usual) a virtual destructor.

src/main.cpp
```diff
@@ -72,13 +123,7 @@ int main ( int argc, char** argv )
 
     // initialize all flags and string which might be changed by command line
     // arguments
-#if ( defined( SERVER_BUNDLE ) && defined( Q_OS_MACX ) ) || defined( SERVER_ONLY )
```
atsampson, line 75: Don't change this.

src/main.cpp
```diff
@@ -115,748 +160,765 @@ int main ( int argc, char** argv )
...
-        }
-
-        // Version number ------------------------------------------------------
-        if ( ( !strcmp ( argv[i], "--version" ) ) || ( !strcmp ( argv[i], "-v" ) ) )
+#if !defined( HEADLESS ) && defined( _WIN32 )
+        if ( AttachConsole ( ATTACH_PARENT_PROCESS ) )
         {
-            std::cout << qUtf8Printable ( GetVersionAndNameStr ( false ) );
-            exit ( 0 );
+            (void) freopen ( "CONOUT$", "w", stdout );
```
atsampson, line 169: Don't change this.

src/main.cpp
```diff
@@ -115,748 +160,765 @@ int main ( int argc, char** argv )
...
-            iPortNumber            = static_cast<quint16> ( rDbleArgument );
-            bCustomPortNumberGiven = true;
-            qInfo() << qUtf8Printable ( QString ( "- selected port number: %1" ).arg ( iPortNumber ) );
-            CommandLineOptions << "--port";
-            continue;
-        }
+            // Help (usage) flag ---------------------------------------------------
+            if ( ( !strcmp ( argv[i], "--help" ) ) || ( !strcmp ( argv[i], "-h" ) ) || ( !strcmp ( argv[i], "-?" ) ) )
+            {
+                throw CInfoExit ( UsageArguments ( argv ) );
```
atsampson, line 185: Just change `exit(0);` to `return 0;`.

src/main.cpp
```diff
@@ -115,748 +160,765 @@ int main ( int argc, char** argv )
...
-        {
-            iJsonRpcPortNumber = static_cast<quint16> ( rDbleArgument );
-            qInfo() << qUtf8Printable ( QString ( "- JSON-RPC port number: %1" ).arg ( iJsonRpcPortNumber ) );
-            CommandLineOptions << "--jsonrpcport";
-            continue;
-        }
+            // Version number ------------------------------------------------------
+            if ( ( !strcmp ( argv[i], "--version" ) ) || ( !strcmp ( argv[i], "-v" ) ) )
+            {
+                throw CInfoExit ( GetVersionAndNameStr ( false ) );
```
atsampson, line 191: Just change `exit(0);` to `return 0;`.

src/main.cpp
```diff
@@ -115,748 +160,765 @@ int main ( int argc, char** argv )
...
+            {
+                strMIDISetup = strArgument;
+                qInfo() << qUtf8Printable ( QString ( "- MIDI controller settings: %1" ).arg ( strMIDISetup ) );
+                CommandLineOptions << "--ctrlmidich";
+                ClientOnlyOptions << "--ctrlmidich";
+                continue;
+            }
 
-    // Dependencies ------------------------------------------------------------
-#ifdef HEADLESS
```
atsampson, line 571: Don't change this.

src/main.cpp
```diff
@@ -115,748 +160,765 @@ int main ( int argc, char** argv )
...
-        if ( ClientOnlyOptions.size() != 0 )
+#endif
+
+#ifdef HEADLESS
+        Q_UNUSED ( bStartMinimized )       // avoid compiler warnings
+        Q_UNUSED ( bShowComplRegConnList ) // avoid compiler warnings
+        Q_UNUSED ( bShowAnalyzerConsole )  // avoid compiler warnings
+        Q_UNUSED ( bMuteStream )           // avoid compiler warnings
+
+        if ( bUseGUI )
```
atsampson, line 626: Don't change this. In particular it's not a good idea to change error messages unless it's absolutely necessary - they show up in documentation.

src/main.cpp
atsampson: What ann0see said.

src/main.cpp
```diff
@@ -115,748 +160,765 @@ int main ( int argc, char** argv )
...
+                                     strArgument ) )
+            {
+                strClientName = strArgument;
+                qInfo() << qUtf8Printable ( QString ( "- client name: %1" ).arg ( strClientName ) );
+                CommandLineOptions << "--clientname";
+                ClientOnlyOptions << "--clientname";
+                continue;
+            }
 
-// clicking on the Mac application bundle, the actual application
```
atsampson, line 563: Explain in the commit message why removing this is safe (presumably it wasn't reached originally because of the `qCritical` - but I'd check the Git history to see why it was added).

src/main.cpp
```diff
@@ -115,748 +160,765 @@ int main ( int argc, char** argv )
...
+                    qWarning() << "Server Bind IP is invalid. Only plain IP addresses are supported.";
+                    strServerBindIP = "";
+                }
             }
         }
-    }
 
-    // Application/GUI setup ---------------------------------------------------
-    // Application object
-#ifdef HEADLESS
```
atsampson, line 763: This is where you should have `QScopedPointer<QCoreApplication> pApp;` - see my branch. You don't need it to be of different types in the headless/non-headless cases, because `QApplication` is a subclass of `QCoreApplication` and has a virtual destructor.

You can then remove all of the extra conditionals that you've added below because of splitting this into two variables.

src/main.cpp
```diff
@@ -115,748 +160,765 @@ int main ( int argc, char** argv )
...
 
 #ifndef SERVER_ONLY
-    // clang-format off
+        // clang-format off
 // TEST -> activate the following line to activate the test bench,
 //CTestbench Testbench ( "127.0.0.1", DEFAULT_PORT_NUMBER );
 // clang-format on
 #endif
 
-    CRpcServer* pRpcServer = nullptr;
```
atsampson, line 821: Smart pointer here - plus smart pointers for CServerRpc and CClientRpc - see my branch.

src/main.cpp
```diff
@@ -982,32 +1044,71 @@ int main ( int argc, char** argv )
...
+    }
+
+    catch ( ... )
+    {
+        // show all other unhandled (standard) exceptions
+        qCritical() << qUtf8Printable ( QString ( "%1: Unknown Exception, Exiting" ).arg ( strAppName ) );
+        exit_code = -1;
+    }
+
+    if ( pServerRpc )
```
atsampson, line 1080: Not needed with a smart pointer. (But even if it was, you never need to write `if (x) delete x;` - `delete NULL;` is [guaranteed to do nothing](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null).)

atsampson (changes requested):
Review added as requested.

ann0see:
@atsampson Thanks for the review ;-)

src/main.cpp
pgScorpio: > Better to avoid making unnecessary changes entirely.

In my opinion that is one of the main problems in the Jamulus development.
Just fixing bugs locally, without solving the real cause of the bug... often poor framework...
Improvements of the framework aren't "unnecessary changes"!
Real development is continuously improving the framework to prevent future similar bugs.
So you should strive for solutions that work in any case, not just for the bug you are fixing.

src/global.h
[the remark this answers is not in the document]
pgScorpio: > It's [bad style](https://isocpp.org/wiki/faq/exceptions#why-not-exceptions) in C++ to use exceptions for handling normal situations.

It's what you call a "normal" situation...
But I agree that, just for the --help and --version, it could be a return ( 0 ).

src/global.h
```diff
@@ -319,11 +319,41 @@ typedef unsigned char      uint8_t;
...
+};
+
+class CErrorExit
+{
+public:
+    CErrorExit ( QString strNewErrorMsg, int iNewExitCode = 1 ) : strErrorMsg ( strNewErrorMsg ), iExitCode ( iNewExitCode ) {}
+
+    QString GetErrorMessage() const { return strErrorMsg; }
+
+    int GetExitCode() const { return iExitCode; }
```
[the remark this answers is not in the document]
pgScorpio: > The exit code is always 1 on error in the code below, so `iExitCode` isn't needed. (You might want to add this in the future if there's really a good reason to have different exit codes, but I think that's very unlikely to be useful in Jamulus, and it's certainly not needed now.)

I agree the exit code is (strangely enough) always 1 in this code, but as said we should provide a framework with general solutions, not prohibiting any 'normal' behavior.
Prohibiting any 'normal' behavior now will possibly lead to unwanted 'solutions' in the future.

src/global.h
```diff
@@ -319,11 +319,41 @@ typedef unsigned char      uint8_t;
...
+public:
+    CInfoExit ( QString strMsg ) : strInfoMsg ( strMsg ) {}
+
+    QString GetInfoMessage() const { return strInfoMsg; }
+
+protected:
+    QString strInfoMsg;
+};
+
+class CErrorExit
```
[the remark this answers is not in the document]
pgScorpio: > I'd call this something like `CCmdLineErr`, to indicate what it's actually being used to report.

Again, this is a 'universal' solution, so not only usable for commandline errors, so why name it that way just because we now use it for commandline errors?

src/global.h
```diff
@@ -359,6 +394,9 @@ class CCustomEvent : public QEvent
 };
 
 /* Prototypes for global functions ********************************************/
+
```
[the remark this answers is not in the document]
pgScorpio: Again, Provide a Framework that should (and will) be used in the near future...

src/main.cpp
pgScorpio: Just a (good?) habit of me to always check for null pointers ;=)
But you are right, it's not necessary here....

src/main.cpp
pgScorpio: Yes, this was a strange exception....
Since 1. I tested it on macOS and didn't see any strange behavior, and 2. the same exit ( 1 ) is used in multiple places, so if it really was a problem it should have been implemented in multiple places...

But yeah, I should have mentioned it...

pgScorpio:
@ann0see 

> I think it's quite close to the KISS principle and "hard core Jamulus like".
       
 May I note that the KISS principle refers to the implementation of (unnecessary) functionality, NOT the way the code is implemented. "Simple and Stupid" code implementation will always lead to more and more problems and bugs....
That's the whole idea behind my 'improvement' PR's, improving the framework to make maintaining the code easier and less prone to bugs...

ann0see:
Yes. Probably. However the history of Jamulus was quite radical with code changes. It doesn't need to be like that - at least it's probably not up to me, but "the community" to decide.
The goal should be simple, extendable, unbloated code which doesn't do more than it needs to do. I think the refactoring of the Autobuild was a good example on simplification.

src/global.h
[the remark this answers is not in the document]
ann0see: I assume we can leave it as is - if we use it in future (and I think there is work coming up by @pgScorpio which uses it)

ann0see:
Not sure if that's a good idea, but I think for your PRs (especially the sound related redesign, I think it would be worth having npostavs and corrados as reviewers). I know corrados is officially no longer part of the Jamulus Team, however he has the most insight probably.

ann0see:
@jamulussoftware/maindevelopers how do we proceed here? I think the review by @atsampson has valid input, but I'm not fully sure how to continue (especially since - as discussed internally - some team members don't quite see the improvements).

I'd like someone like corrados to review sound redesign related changes (maybe not this one) since he's probably most familiar with the sound code.

pljones:
Moving out to 3.10.0 as no clear fix identified.

pljones:
Moving back to Triage until a way forward is proposed (and removing milestone).

pljones:
Setting to draft as branch needs conflict resolution.

pljones:
Closing as no activity in two years.

PR #2625, opened 2022-05-09
Automate build of iOS signed ipa file for App Store submission

danryu:
This PR adds automation to create a signed ipa (installer) file for direct submission to the iOS App Store.

CHANGELOG:  adds iOS signed pkg build automation

**Context: automates building of signed pkg file for iOS App Store**

### Required:
1) In Apple Developer Account, create the following resources in in https://developer.apple.com/account/resources/certificates/list

- Certificates:
  - iOS Distribution

- Identifier:
  - app ID (bundleID)

- Profile:
  - iOS Distribution (AppStore) - ProvisioningProfile

2) Add the certs and provisioning profile to Github Secrets as per https://docs.github.com/en/actions/deployment/deploying-xcode-applications/installing-an-apple-certificate-on-macos-runners-for-xcode-development

## Checklist


.github/autobuild/ios.sh
```diff
@@ -30,17 +30,76 @@ setup() {
...
+    #FIXME - maybe redundant?
+    mkdir -p ~/Library/MobileDevice/Provisioning\ Profiles
+    cp $IOS_PP_PATH ~/Library/MobileDevice/Provisioning\ Profiles
+
+    # for debug
+    echo "Checking found identities..."
+    security find-identity -v 
+
+    # Tell Github Workflow that we need notarization & stapling:
+    echo "::set-output name=macos_signed::true"
```
ann0see, line 77: should be renamed to apple_signed if we merge this.

.github/autobuild/ios.sh
```diff
@@ -30,17 +30,76 @@ setup() {
...
+    BUILD_ARGS=("")
+    if prepare_signing; then
+        BUILD_ARGS=("-s" "${IOSDIST_CERTIFICATE_ID}" "-k" "${KEYCHAIN_PASSWORD}")
+    fi
+    ./ios/deploy_ios.sh "${BUILD_ARGS[@]}"
 }
 
 pass_artifact_to_job() {
-    local artifact="jamulus_${JAMULUS_BUILD_VERSION}_iOSUnsigned${ARTIFACT_SUFFIX:-}.ipa"
+    local artifact="jamulus_${JAMULUS_BUILD_VERSION}_iOS_unsigned${ARTIFACT_SUFFIX:-}.ipa"
```
ann0see, line 94: Not sure how much a signed IPA on the web without the app store helps?

ann0see (commented):
Interesting! To me it seems as if this still needs manual work. In an ideal workflow a release push would upload the app to the app store automatically and just wait on someone to press a "publish" button there.

.github/autobuild/ios.sh
emlynmac: or maybe ios_signed

.github/autobuild/ios.sh
```diff
@@ -30,17 +30,76 @@ setup() {
...
     echo "Moving build artifact to deploy/${artifact}"
-    mv ./deploy/Jamulus.ipa "./deploy/${artifact}"
+    mv ./deploy/Jamulus_unsigned.ipa "./deploy/${artifact}"
```
emlynmac, line 96: Probably not worth keeping this around, only signed version would be useful past making sure the build works.

.github/workflows/autobuild.yml
```diff
@@ -149,13 +149,13 @@ jobs:
 
           - config_name:            iOS (artifacts)
             target_os:              ios
-            building_on_os:         macos-10.15
-            base_command:           QT_VERSION=5.15.2 ./.github/autobuild/ios.sh
+            building_on_os:         macos-11
+            base_command:           QT_VERSION=5.15.2 SIGN_IF_POSSIBLE=1 ./.github/autobuild/ios.sh
```
emlynmac, line 153: Could you use the presence of a certificate in the environment to determine whether to sign, so not needing another parameter?

emlynmac (commented):
Looks like this PR needs a bit more work to make it happy with the case of not having certificates around.

.github/autobuild/ios.sh
danryu: @ann0see it _is_ actually possible to upload the binary to the App Store, using "altool" as outlined here: https://help.apple.com/app-store-connect/#/devb1c185036

However, as the App Store submission process involves a number of manual setup steps _anyway_ (ie creating the actual submission on the App Store Connect website, adding screenshots, descriptions, compliance details and all sorts) and you likely only want to upload specific versions to the App Store (rather than having it run automatically), I thought the generation of the signed, verifiable IPA is a big enough win for this automation.

_Quick note: the only manual step remaining once you have the signed IPA is to drag and drop the IPA file into Transporter (on macOS), and click the Upload button._

If I have time today though I will experiment with adding the upload directly using xcrun altool.

.github/autobuild/ios.sh
danryu: I don't think there's any point to that, no - the signed IPA is purely for App Store submission. So it's really for @emlynmac or whoever is managing the Jamulus App Store account to have a ready-built file for submission.

I assume the unsigned IPA is useful for general testing purposes so obviously I've kept that in.

.github/autobuild/ios.sh
danryu: Ok, I left a note above, I wasn't sure if people were maybe using the unsigned IPA for local testing or somesuch.
No problem to remove it if you prefer.

.github/workflows/autobuild.yml
danryu: I would rather follow the lead as being done for macOS, which uses that flag. Or we can remove from both macOS and iOS but to have in just one feels weird.

ann0see:
@danryu The CI failed. We currently don't have a valid cert on this repo.

danryu:
> @danryu The CI failed. We currently don't have a valid cert on this repo.

Ok I've pushed a fix now to check for the cert non-zero value before attempting to move the signed IPA for download. This should resolve this CI failure where there is no valid cert in place.

Please note also that if signing is requested and dependencies are satisfied, the script will now attempt validate and upload to App Store using the existing NOTARIZATION_USER and NOTARIZATION_PASSWORD. We may need to fine-tune this behaviour if eg re-uploading the same version (eg 3.9.0) to the App Store repeatedly (I think it will fail rather than updating the existing similar version). Let me know what you think should happen in this case.

ann0see:
Hmm. I think the upload should only happen if we have a release build (there's a GitHub variable for that, I think). Look for the variable which enables/disabled uploading to GitHub releases.

danryu:
Ok, the Github release stuff switches on `create_release.outputs.publish_to_release == 'true'` - so maybe I should add a separate extra step in `autobuild.yml` for iOS App Store upload - after the Notarization and Staple steps perhaps?
Then I can split out the validate and upload stages from the main build run, which seems sensible.

ann0see:
Yes. I think so.

danryu:
I'm working on adding the extra step now. 

Also: I've noticed that the iOS App Store upload validation (to appear in TestFlight in App Connect) requires an images assets catalog in a directory entitled `Images.xcassets` - I have this working in branch CI as a subdir of `ios`. However the asset catalog is a collection of ~26 files, normally generated by XCode or another utility. I'm thinking of integrating a tool like https://github.com/alfianlosari/AppIconGenCLI, which would then mean we just have to include a single 1024x1024 png file to generate all the others, and saving us from putting all 26+ files in source control.

ann0see:
I think both approaches have advantages and disadvantages (another CI job vs more data in the repo). Any other comments on this @ngocdh @emlynmac

ann0see:
@danryu any updates here?

danryu:
Hi @ann0see yes, I have Store submission working as part of the release procedure (also for macOS for the other PR).
I just need to confirm what the desired behavior is before updating the PR. Do we still want to keep the unsigned IPA if the signed one is generated?

ann0see:
Can the signed ipa be re-signed for debugging? If yes, no need to have the unsigned ipa. Else just provide the undignified one a d upload the signed one to Apple.

danryu:
I'm actually not 100% sure on that point, so I will keep the unsigned IPA for now, to not break the existing pattern.

ann0see:
Strange. Do you use Qt5?

danryu:
> Strange. Do you use Qt5?

No, Qt 6.3.2 and actually 6.4.0 for iOS (for some touch-handling fixes),

It is indeed very strange - after all, if TestFlight validates the package, it should run! I suspect it may be related to something like Bitcode.

Incidentally, the result is the same when I use a dedicated Github Action for iOS building - yukiarrr/ios-build-action@v1.5.0 - which I think might be the simplest way of doing iOS builds going forward.

ann0see:
Yes, the iOS build crashes with Qt6 for me too (at least on GH actions). It's a known issue with Qt6

danryu:
> Yes, the iOS build crashes with Qt6 for me too (at least on GH actions). It's a known issue with Qt6
> 
> #2711

Hmmm, I think I may have fixed that particular issue by including this in .pro file for iOS:
`    QMAKE_LFLAGS += -Wl,-e,_qt_main_wrapper`

Though I don't have the reference for the fix to hand.
With this in place, I've been building and running the iOS app for several months, with Qt 6.3.2 and now 6.4.0.
However, within the last 4-6 weeks (roughly), TestFlight has been validating and accepting the CI iOS builds, only to have them crash with SIG11 on startup.
I spent considerable time building from historical commits, exactly replicating previous configurations - but couldn't shake the error.
To me it seems like an opaque change in how TestFlight is handling the builds in each case - very frustrating.
The workaround currently for me is to do the local XCode build as mentioned.

Please check and try the Koord app in the iOS App Store for proof that Qt6 works fine with iOS :)

ann0see:
> Koord app 

Great. That's a modified Jamulus client I suppose? If you plan to maintain your app and keep it compatible and up to date, we could link it as Jamulus compatible 3rd party App here in the KB: https://jamulus.io/kb/2022/03/04/Alternative-Install-Methods.html

danryu:
> Great. That's a modified Jamulus client I suppose? If you plan to maintain your app and keep it compatible and up to date, we could link it as Jamulus compatible 3rd party App here in the KB: https://jamulus.io/kb/2022/03/04/Alternative-Install-Methods.html

That would be great! We definitely intend on maintaining it :) and also keeping upstream compatibility!
https://github.com/koord-live/koord-app
We are a few months behind upstream master, and that will be merged in shortly.

There's a few points of interest in the app - but I might open a quick Discussion for that!

ann0see:
Adding this stackoverflow post as reference for the new flags: https://stackoverflow.com/questions/45508043/qt-ios-linker-error-entry-point-main-undefined

ann0see:
Needs a rebase. Not closing yet, but might in future.

.github/autobuild/ios.sh
ann0see: Should not be removed IMO.

ann0see:
@danryu I'd suggest you add a very brief note to compatibility at https://github.com/jamulussoftware/jamuluswebsite/edit/release/_posts/2022-03-04-Alternative-Install-Methods.md without any big links to Koord. It can and should not be an ad.

danryu:
> @danryu I'd suggest you add a very brief note to compatibility at https://github.com/jamulussoftware/jamuluswebsite/edit/release/_posts/2022-03-04-Alternative-Install-Methods.md without any big links to Koord. It can and should not be an ad.

Sorry I don't understand this request. Why would I mention Koord, and why is a specific note required?
This PR adds all the logic to be able to submit a working version of Jamulus to the App Store. If there's no interest in this, I'm happy to close the PR.

ann0see:
No. There is interest in this PR. This was only outstanding from above (https://github.com/jamulussoftware/jamulus/pull/2625#issuecomment-1276234487).

.github/autobuild/ios.sh
```diff
@@ -39,17 +39,75 @@ setup() {
...
+    # Set up a keychain for the build:
+    security create-keychain -p "${KEYCHAIN_PASSWORD}" build.keychain
+    security default-keychain -s build.keychain
+    security unlock-keychain -p "${KEYCHAIN_PASSWORD}" build.keychain
+    security import iosdist_certificate.p12 -k build.keychain -P "${IOSDIST_CERTIFICATE_PWD}" -A -T /usr/bin/codesign
+    security set-key-partition-list -S apple-tool:,apple: -s -k "${KEYCHAIN_PASSWORD}" build.keychain
+    # add notarization/validation/upload password to keychain
+    xcrun altool --store-password-in-keychain-item --keychain build.keychain APPCONNAUTH -u $NOTARIZATION_USER -p $NOTARIZATION_PASSWORD
+    # set lock timeout on keychain to 6 hours
+    security set-keychain-settings -lut 21600
```
ann0see, line 75: As in the macOS case - maybe this is no longer needed.

.github/autobuild/ios.sh
```diff
@@ -39,17 +39,75 @@ setup() {
...
+
+    # use this as filename for Provisioning Profile
+    IOS_PP_PATH="embedded.mobileprovision"
+
+    ## Put the cert to a file
+    # IOSDIST_CERTIFICATE - iOS Distribution
+    echo "${IOSDIST_CERTIFICATE}" | base64 --decode > iosdist_certificate.p12
+
+    ## Echo Provisioning Profile to file
+    echo -n "${IOS_PROV_PROFILE_B64}" | base64 --decode > $IOS_PP_PATH
```
ann0see, line 64: As in the macOS case, probably worth putting this in an if.

ann0see:
Why was this closed?

pljones:
> https://github.com/kormix-io closed this by deleting the head repository [8 hours ago](https://github.com/jamulussoftware/jamulus/pull/2625#event-26749684388)

What?!

danryu:
> Why was this closed?

I submitted it 4 years ago, and you guys didn't do anything with it. 
It's been usable and available to merge that entire time.
It's also at least partially obsolete now.
I'm no longer working on Jamulus code. I did have a working fork using CMake, JUCE (enabling VST/AU plugins), QML UI but it was no longer viable to maintain.

danryu:
On a personal note, I have to say that the Jamulus project presents as somewhat hostile to developers outside the core team.
As an example, when I submitted a working QML rewrite of the Jamulus client a couple of years ago, I expected a more positive response.
I had done 95% of the work in unlocking a new, simplified path of UI development for Jamulus, but only one person (ann0see - thanks! :) had the slightest positive thing to say! Not even acknowledgement from any other core devs.

My instinct was also to share the JUCE audio backend migration (which allows seamless building as VST3/AU plugin out of the box) with you upstream, but I figured that it would again be met either with indifference or perhaps even caustic remarks.

It seems to me that Jamulus development stultified and is merely being preserved by "gatekeepers" rather than people trying to advance the project.

ann0see:
> On a personal note, I have to say that the Jamulus project presents as somewhat hostile to developers outside the core team.

That's sad to hear - but I do understand the sentiment. I've raised this in our team channel and hope that we can find an official improvement plan. 
I think the most blocking problem is that many restructurings are large and thus very hard to review. Thus they linger. 

We've onboarded @dingodoppelt who sees the project from a more user centric perspective - this will of course change team dynamics a bit. Hopefully for the better. 

Jamulus 4 should allow for larger changes maybe this also shifts the mindset.

Issue #2628, opened 2022-05-15
linux/sound.h should not depend on src/server.h


ann0see:
One thing I noticed when I was doing my various changes... linux/sound.h depends on src/server.h.  It shouldn't - not sure if you want to go fixing this here, though.  (It means moving the high resolution timer from src/server.* to src/util.* - was a simple cut-n-paste.)

__Originally posted by @pljones in https://github.com/jamulussoftware/jamulus/pull/2575#issuecomment-1100919070__

pgScorpio:
I think we better create a new file i.e. timers.h.
util.h is already becomming a receptacle (or trash can) with a lot of code that is not really "utility".
Better to keep specific code in specific files.

Is there already someone working on this one?

ann0see:
Probably @pljones 

https://github.com/pljones/jamulus/compare/patch/move-high-precision-timer...pljones:patch/rethink-int-ext-serverlist-entry?expand=1

pljones:
> I think we better create a new file i.e. timers.h. 

I'm not sure it's "big" enough to warrant that -- too many files with tiny amounts of code in is a nightmare to manage.

> util.h is already becomming a receptacle (or trash can) with a lot of code that is not really "utility". Better to keep specific code in specific files.

Agreed on that point.  Something needs doing.

> Is there already someone working on this one?

Yep.

PR #2637, opened 2022-05-22
#2628 Move CHighPrecisionTimer to util

pljones:

Removes dependency of Linux sound code on Server code by moving High Precision Timer to join the other timer code in Util.

CHANGELOG: Internal: code restructuring to reduce cross-dependencies (CHighPrecisionTimer)

**Context: Fixes an issue?**
Fixes: #2628

No.

Running on all my clients and servers.

Nothing really but I'd wait until after 3.9.0, as moving to util might not be its final resting place.

## Checklist


src/sound/jack/sound.h
```diff
@@ -34,15 +34,15 @@
...
-
 #if WITH_JACK
 #    include <jack/jack.h>
 #    include <jack/midiport.h>
 #endif
 
+#include "../../global.h"
+#include "../../util.h"
+#include "../soundbase.h"
+
```
hoffie, line 45: Windows/Jack build reproducibly [fails](https://github.com/jamulussoftware/jamulus/runs/6544938563?check_suite_focus=true#step:9:278) due to some definition problem. Maybe this include reordering is the culprit?

src/util.h
```diff
@@ -369,7 +375,7 @@ public slots:
     void keyPressEvent ( QKeyEvent* pEvent )
     {
         // block escape key
-        if ( pEvent->key() != Qt::Key_Escape )
+        if ( ( this->windowFlags() & Qt::Dialog ) != 0 || pEvent->key() != Qt::Key_Escape )
```
hoffie, line 378: This seems unrelated at first glance?
If it was really intended to go in, it might make sense to put it into its own commit?

hoffie (commented):
>  Nothing really but I'd wait until after 3.9.0, as moving to util might not be its final resting place.

I've just tagged it for 3.9.1 based on this.
In general I'd say it should go into 3.9.0 if it is ready. We can always change it later. Otherwise, such a PR is prone to be a target for conflicts which might be annoying to resolve.
Also, 3.9.0 is currently in a state with lots of moving parts under the hood, so it should get proper "generic" testing anyway.

src/sound/jack/sound.h
pljones: `"../soundbase.h"` includes `"../../global.h"` and `"../../util.h"` (and in that order) -- this _should_ work.  Also, I can't see _any_ jack-ish dependency...  It's really weird but yeah, totally reproducible.  I'll try dropping the includes here...  Mm... better.

src/util.h
pljones: Uh...  No idea how that got in there.  Hm... Good spot.  I'd been toying with the idea - it was meant to have been reverted.

pljones: Strange.  Local diff from here to master doesn't show this as a change.

pljones: OK, after rebasing I get the change.  It must have been dropped on master and my rebase didn't pick it up for some reason.

pljones: Hm.  And then it didn't.  I've given in and reverted to just the needed change in the sound header, which seems to make it work.  It would be good to know why the reordering is fine on Linux but not Windows.

src/sound/jack/sound.h
pljones: Hm.  And then it didn't.  I've given in and reverted to just the needed change in the sound header, which seems to make it work.  It would be good to know why the reordering is fine on Linux but not Windows.

hoffie (approved):

ann0see (approved):
As we're moving a lot of files around, I think we can merge this...

I agree with @pgScorpio that the util file gets filled with a lot of stuff...

pljones:
> As we're moving a lot of files around, I think we can merge this...

Well, a lot of it seems to be translations - I'm not sure how they get managed so I just bundled them in as they seemed to have changed.

Merging.

pljones:
I should have changed the milestone when merging...

PR #2638, opened 2022-05-22
A11y: Combine client info into a group

chigkim:
Combine all the client info, so screen reader users don't have to track down multiple places.

CHANGELOG: Accessibility: Combine all the client info, so screen reader users don't have to track down multiple places.

**Context: Fixes an issue?**

## Checklist

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -865,6 +865,8 @@ void CChannelFader::SetChannelInfos ( const CChannelInfo& cChanInfo )
     pcbMute->setAccessibleName ( "Mute " + strAliasAccessible + ", " + strInstrumentAccessible );
     pcbSolo->setAccessibleName ( "Solo " + strAliasAccessible + ", " + strInstrumentAccessible );
     pcbGroup->setAccessibleName ( "Group " + strAliasAccessible + ", " + strInstrumentAccessible );
+    dynamic_cast<QWidget*> ( plblLabel->parent() )
+        ->setAccessibleName ( strAliasAccessible + ", " + strInstrumentAccessible + ", " + strLocationAccessible );
 }
 
 /******************************************************************************\


ann0see:
Thanks! Is it that easy?

chigkim:
Yes!!!

src/audiomixerboard.cpp
```diff
@@ -862,6 +862,8 @@ void CChannelFader::SetChannelInfos ( const CChannelInfo& cChanInfo )
     plblLabel->setToolTip ( strToolTip );
     plblLabel->setAccessibleName ( strAliasAccessible );
     plblLabel->setAccessibleDescription ( tr ( "Alias" ) );
+    dynamic_cast<QWidget*> ( plblLabel->parent() )
```
hoffie, line 865: Just trying to understand why the cast is needed:
- This is in `CChannelFader` which is a `QObject`. Therefore, `parent()` returns a `QObject` as well.
- Doing the dynamic_cast is safe and right to do because we know that `CChannelFader` is only used as a child in `CAudioMixerBoard`, which is a `QWidget`.

Is that correct? If so, would it make sense to add a short code comment about that?

hoffie (commented):
Thanks -- that sounds great! One inline question (see above).

chigkim:
In order to ->setAccessibleDescription, you need to cast to QWidget.

ann0see (commented):
Unfortunately, I couldn't verify this PR. How can I see it in action?

src/audiomixerboard.cpp
pljones: Most UI things are widgets.  But if parent (i.e. the form it's stuck to) is only _declared_ as a QObject, you'll get told off by the compiler trying to use methods QObject doesn't have -- _even though_ the QObject you're talking about does have them.  Hence the cast.  So long as it's viable (I've not checked).

chigkim:
You can check with a screen reader. VoiceOver on Mac, and NVDA with 
object navigation on Windows.

VoiceOver has a mini tutorial how to use it if you press command+f5 on 
your Mac.

ann0see:
Ok. Thanks. Could hear it now. The integrated one of Win 11 didn't find it

ann0see (changes requested):
This PR needs to be rebased to resolve the conflicts

pljones (approved):
It is a shame about that cast - I'd have though the parent _widget_ might be known somewhere.  It probably is....

ann0see (approved):
Thanks!

Issue #2403, opened 2022-02-19
Android: Duplicate permissions in Manifest


hoffie:
**Describe the bug**

When installing the Android `adb logcat` shows warnings regarding duplicate permissions in the Manifest.

**To Reproduce**

Install APK (or probably any other version)

**Expected behavior**

No warnings in the log

**Screenshots**

```
02-19 11:51:35.040 31937 31937 W PackageParser: Ignoring duplicate uses-permissions/uses-permissions-sdk-m: android.permission.INTERNET in package: com.github.jamulussoftware.jamulus at: Binary XML file line #19
02-19 11:51:35.041 31937 31937 W PackageParser: Ignoring duplicate uses-permissions/uses-permissions-sdk-m: android.permission.WRITE_EXTERNAL_STORAGE in package: com.github.jamulussoftware.jamulus at: Binary XML file line #21
```

**Operating system**

Android

**Version of Jamulus**

3.8.2rc1

pljones:
Anyone know how to get this resolved?  Is there a related fix raised?  It _sounds_ pretty simple if someone knows how to change the installer to check first.

ann0see:
Tagging for 3.9.1

hoffie:
There have been manifest changes via #2640. It should be re-checked, if this issue has been fixed as a side effect.

pljones:
Waiting for an update here.

pljones:
Moving to 3.10.0.  If there's no confirmation that it's still a problem there, it'll get closed.

ann0see:
I doubt it's fixed. @sthenos can you please have a look at Android? Do we know anyone else who knows more about Android?

pljones:
Moving out of 3.10.0 and back to Triage until someone can pick it up.

pljones:
Doesn't look like we have much Android interest.  Dropping milestone and moving to Triage.

pljones:
We can try to look into this at least for 3.11.0.  If not, it can probably be closed if it's just a warning.

mcfnord:
> [!NOTE]
> 📡 **STAND BY FOR AN LLM-AUTHORED MESSAGE.**

[...]

PR #2640, opened 2022-05-22
A11y Sound Alert for new person and new chat message

chigkim:
Added sound Alert for new person and new chat message

CHANGELOG: Added sound Alert for new person and new chat message

**Context: Fixes an issue?**
Screen reader users has no idea when someone joins or when receiving a new chat message. This plays a short sound alert for those events.

The sound alerts play through OS sound output. for this to work on Windows, your system sound output may need to be different than what Jamulus uses depending on the audio device/driver being used.
Please include the new GUI setting in the documentation.
The original sound files are from Freesound, and they were edited to fit the context  in Jamulus.
new_message.wav: https://freesound.org/people/zrrion_the_insect/sounds/469014/
new_user.wav: https://freesound.org/people/deleted_user_4772965/sounds/256513/
The licenses are CC0.

## Checklist


chigkim:
It compiled without a problem on both Windows and Mac machines, but it failed on autobuild on Github. Thoughts?

hoffie (commented):
> It compiled without a problem on both Windows and Mac machines, but it failed on autobuild on Github. Thoughts?

[`Project ERROR: Unknown module(s) in QT: multimedia`](https://github.com/jamulussoftware/jamulus/runs/6546156539?check_suite_focus=true#step:9:72)

We are only downloading the required modules in the build scripts, so those lists need to be adjusted. Those might be the proper places:
```
$ grep archive.*qtbase .github/ -rn
.github/autobuild/android.sh:73:            --archives qtbase qttools qttranslations qtandroidextras
.github/autobuild/ios.sh:23:        python3 -m aqt install-qt --outputdir "${QT_DIR}" mac ios "${QT_VERSION}" --archives qtbase qttools qttranslations
.github/autobuild/ios.sh:28:            python3 -m aqt install-qt --outputdir "${QT_DIR}" mac desktop "${QT_VERSION}" --archives qtbase
.github/autobuild/windows.ps1:45:        "--archives", "qtbase", "qttools", "qttranslations"
.github/autobuild/mac.sh:22:        python3 -m aqt install-qt --outputdir "${QT_DIR}" mac desktop "${QT_VERSION}" --archives qtbase qttools qttranslations
```

Minor general request: Your changelog suggestions sound good. Can you put them on the same line as the CHANGELOG: prefix in your PR descriptions? That way our script can pick the text up directly.

It seems like this PR introduces those sound effects unconditionally for all users. While I understand that some people might want or even need those, I think those need to have a way to be turned off (probably even by default in order to avoid changing the behavior for the existing user base?).

chigkim:
1. Thanks! I changed the auto build scripts and pushed a new commit to 
my ranch with the PR. Do I need to make another PR?
2. I updated the CHANGELOG: line.

3. I agree. If someone could implement this in the setting to toggle the 
future might be a good idea.

chigkim:
@[hoffie](https://github.com/hoffie) I just added qtmultimedia on the suggested lines, but all the builds except iOS failed.
I'd appreciate any help.

src/clientdlg.cpp
```diff
@@ -826,6 +829,9 @@ void CClientDlg::OnCLVersionAndOSReceived ( CHostAddress, COSUtil::EOpSystemType
 
 void CClientDlg::OnChatTextReceived ( QString strChatText )
 {
+    QSoundEffect* sf = new QSoundEffect();
```
pljones, line 832: How do I turn this off?

src/clientdlg.cpp
```diff
@@ -880,6 +886,13 @@ void CClientDlg::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo
 
 void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
 {
+    if ( iNewNumClients > iClients )
```
pljones, line 889: How do I turn this off?

chigkim:
Is there someone who could help with implementing a toggle in the setting dialog? I'm blind, and unfortunately qt designer is not accessible.
Thanks!

ann0see:
@henkdegroot @mulyaj I think you worked with the GUI?

henkdegroot:
> a toggle in the setting dialog?

@chigkim, perhaps I can help out with the setting. Are you sure a setting in a GUI dialog is the best/most convenient option?
Thinking that perhaps a command line option might be easier to use. What do you think?
If it should be in the GUI, what would be the best place for it? Perhaps the My Profile tab? Which contains settings to control the user interface.

henkdegroot:
> @[hoffie](https://github.com/hoffie) I just added qtmultimedia on the suggested lines, but all the builds except iOS failed.

@hoffie, could it be that the build fails due to the fact it is using cached versions (i.e. cached version without qtmultimedia)?

hoffie:
> @hoffie, could it be that the build fails due to the fact it is using cached versions (i.e. cached version without qtmultimedia)?

Umm, my first thought was "this should be covered by the hash key calculation", but the [log](https://github.com/jamulussoftware/jamulus/runs/6546694935?check_suite_focus=true#step:5:15) and the [code](https://github.com/jamulussoftware/jamulus/blob/481bfa7416c5a9e4fea2af726352e033c2b1661c/.github/workflows/autobuild.yml#L193) say otherwise. The code still refers to the old, no longer existing paths. Seems like I've missed that in autobuild refactoring cleanups.
@chigkim Sorry about that, will submit a PR shortly.

The Linux build failures are not related to that part though (no caching there). In that case I suspect that we need to install `qtmultimedia` explicitly as well (didn't expect that). It's slightly different there as we use the official Debian packages and no aqt.
@chigkim I suggest trying to add `qtmultimedia5-dev` to [sudo apt-get -qq --no-install-recommends -y install devscripts build-essential debhelper fakeroot libjack-jackd2-dev qtbase5-dev qttools5-dev-tools](https://github.com/jamulussoftware/jamulus/blob/481bfa7416c5a9e4fea2af726352e033c2b1661c/.github/autobuild/linux_deb.sh#L30) in `.github/autobuild/linux_deb.sh`.

henkdegroot:
@chigkim, I have cloned your branch and locally compiled without issues. However I don't hear any sound when a chat message arrives or when a new client joins the session. Any thoughts on this?

hoffie:
> @chigkim, I have cloned your branch and locally compiled without issues. However I don't hear any sound when a chat message arrives or when a new client joins the session. Any thoughts on this?

I suspect that by using Qt's audio code a totally different routing can apply. In other words, Jamulus may output to ASIO or Jack, but Qt will likely use the OS default instead, which might be a different device. :(

Another update regarding the build failures: Android seems to have a non-cache related cause as well (Manifest/permissions). Although the mssages seem logical, I can't really tell why they appear right now. Seems related to #2403.

pljones:
> > a toggle in the setting dialog?
> 
> @chigkim, perhaps I can help out with the setting. Are you sure a setting in a GUI dialog is the best/most convenient option? Thinking that perhaps a command line option might be easier to use. What do you think? If it should be in the GUI, what would be the best place for it? Perhaps the My Profile tab? Which contains settings to control the user interface.

1) CClient (in this case) should own the setting
2) CClient should have a socket to change the value of the setting
3) CClient should have a signal for when the value of the setting is changed
4) The setting should get read from and written to the inifile
5) Both the Client RPC code and Client GUI code should expose the setting and respond to updates (i.e. both could be active)
6) Any command line option given should overwrite a value from the inifile (but then get written back to the inifile)

The above is how _all_ Client (or Server) settings should work.

hoffie:
I've filed #2642 for the cache key fix and [I've started a CI run for this PR (#2642) plus the fix in my fork](https://github.com/hoffie/jamulus/actions/runs/2380708310). That should hopefully show that the qtmultimedia-based failure is gone on Mac/Windows.

chigkim:
@henkdegroot, thanks for your help! I think toggle in the setting dialog would be a better option than command line.
Hmm that's interesting. This feature has been in my accessible build for a long time, and it was one of the favorite features for screen reader users. I know it worked for other people. Only thing I could think of is that I changed from QSound to QSoundEffect for qt6 compatibility recently.
I assume git can handle binary wav files right?

hoffie:
>  I assume git can handle binary wav files right?

It should. The files in your PR play properly using `mplayer` for me.

I'm wondering whether it is possible and maybe even better to leave such possible notifications to the screen reader and just try to instruct the screen reader accordingly? Maybe [QAccessibleEvent](https://doc.qt.io/qt-5/qaccessibleevent.html) is of any help?

chigkim:
@hoffie, as far as I know unfortunately QT doesn't have a way to send a generic announcement with custom message to screen readers other than sending new value changed event for a specific UI element.
Also, would non-screen reader users find this useful/interesting at all? Maybe while playing and looking at  chart, you can find out when someone sends you a message?

henkdegroot:
@chigkim, still can't seem to get the sounds alerts working. I have also tried updating the code to use QSound. I am using a windows 10 device. I am testing my local build which I create using the powershell script provided (deploy_windows.ps1) and it uses Qt 5.15.2.

chigkim:
@[henkdegroot](https://github.com/henkdegroot), have you checked your Windows default sound card? As [hoffie](https://github.com/hoffie) mentioned, the sound alert should play through whatever your Windows default sound output is set to, not necessarily what Jamulus uses.

chigkim:
@[henkdegroot](https://github.com/henkdegroot), also, you could try my build. https://www.dropbox.com/s/1cevtv50e9077o9/Jamulus-a11y-sound-alert.zip?dl=0

henkdegroot:
@chigkim I will give your build a try and will also test with another windows laptop. I do get audio alerts from other programs so that is not the issue. I thought that perhaps the ::fromfile was trying to load a local file instead of the file from the resources. Hope to have some time in the weekend for this.

henkdegroot:
@chigkim, tested your build and tested with another windows laptop. Both cases I do not hear any notifications on new client and chat messages.

chigkim:
I asked some people to test the same build a couple of days ago, and 
they said it all worked fine for them. Are you using an external audio 
interface with Asio driver, or built-in sound card with asio4all?

henkdegroot:
> Are you using an external audio interface with Asio driver, or built-in sound card with asio4all?

I am using a built-on sound card with asio4all. I even tested with all other sound drivers disabled, and not difference. Not hearing sounds when a New Client joins or when a Chat Message is received. So not sure why I don't get these sound notifications.

ann0see:
@chigkim do you think this will make it up in the next few days to be able to get included into the next release? If not, I think it's worth to de-tag.

henkdegroot:
Just adding my view, as much I would like to assist creating a gui setting to be able to turn this on/off, I can only do so when I am able to get it working in the first place. If it fails for me, it probably fails for more people and we end up getting issues reported in that. I don't believe this not working is caused by my setup but don't know what is causing this.
@ann0see can you test this feature on a windows system using the link provided by @chigkim?

ann0see:
I assume that ASIO4ALL is the problem (Qt multimedia probably doesn't output the sound via ASIO)

henkdegroot:
> doesn't output the sound via ASIO

So this would only work when not using ASIO? I don't see why Jamulus would only output sound using ASIO when the Windows Sound module is loaded to support the QSoundEffect.
If correct, that it does not output through ASIO, then I would say the current proposed coding is isn't a correct solution.

ann0see:
I think the problem is that ASIO4ALL blocks the integrated soundcard. Some devices, like the Behringer UMC202HD can be accessed via WDM (or WASAPI; not sure???) and ASIO in parallel. So some devices would work, others wouldn't. I can test it on Windows 11 with ASIO4ALL (if I get it to work), but I doub't this would bring us anywhere.

ann0see:
Ok. I can confirm that - although ASIO4ALL doesn't seem to work, I can hear a sound.

chigkim:
Like I mentioned in previous thread, because Asio blocks sharing sounds 
with other apps, all screen reader users have to use two sound cards: 
one for Jamulus and One for system sound like screen reader and sound 
alerts from qt.

If you set your OS system sound output to use other sound card than what 
Jamulus uses, it should work.

pljones:
We still need to get the failing builds working.

henkdegroot:
> If you set your OS system sound output to use other sound card than what Jamulus uses, it should work.

Okay, so I installed another virtual driver and can hear sound indeed. Still strange that I had to do this, as my other applications are sending alerts without issues to my default sound card (even when that is also used as the output for ASIO4ALL) but guess my other applications do not have 2 sound output options.

Not sure if I will be able to get the setting added in the next days or so. @chigkim, can you fix the build issue first?

chigkim:
@hoffie, I added qtmultimedia5-dev to linux_deb.sh, but it still says "Project ERROR: Unknown module(s) in QT: multimedia."
Any thoughts? Thanks!

henkdegroot:
@chigkim, as an improvement I think the #include <QSoundEffect> should move from the .cpp into the .h file. This is how other files have it.

chigkim:
@henkdegroot, #include <QSoundEffect> is already in clientdlg.cpp.
Or, are you refering to something else?

chigkim:
Ok, I fixed all the auto builds except Android.
I added qtmultimedia at the end of the line so it reads:
```
        python3 -m aqt install-qt --outputdir "${QT_BASEDIR}" linux android "${QT_VERSION}" \
            --archives qtbase qttools qttranslations qtandroidextras qtmultimedia
```
Does anyone know how to fix Android?
Thanks for your help.

henkdegroot:
> #include is already in clientdlg.cpp.

That was my point, you have added the include in the .cpp file. All other files have the include in .h file (and of course the .h file is then included in the .cpp file).

chigkim:
@henkdegroot, sorry, I totally misread. lol I changed it now.

henkdegroot:
@chigkim, is it intentional that when I start Jamulus and I am the first user connecting to a server that I also get the New User Alert? 
FYI, I sort of have the setting working in the GUI but need to check and cleanup the code a bit, think I added too much code.

chigkim:
Yes, you can look at it as confirmation. Same for chat. When you send a 
messsage, you get a sound.

henkdegroot:
@chigkim, I believe I have got all changes required in place. I thought I could push my changes into your branch, but seems I cannot (permission issue). How can we best proceed with this? 

Maybe the easiest option is for you to include my changes? The zip file contains all the modified files. Note that I moved the <QSoundEffect> include statement from the client.h file into the clientdlg.h file.

[sound-alert-gui-update.zip](https://github.com/jamulussoftware/jamulus/files/8843013/sound-alert-gui-update.zip)

chigkim:
@henkdegroot, thanks so much for implementing the setting. It worked on my end.
Only fix for Android is left. @hoffie, do you have any suggestion on how to fix the failing android build?

henkdegroot:
@chigkim, I can get the Android working by modifying the AndroidManifest.xml. I have just pushed the fix... However, after doing that I found the application  is crashing on my android device. Not sure if that is caused by the implementation of the audio alert or not.

ann0see:
Commenting the microphone permission doesn't sound right. Did you see what exactly failed?

src/settings.cpp
```diff
@@ -634,6 +640,9 @@ void CClientSettings::WriteSettingsToXML ( QDomDocument& IniXMLDocument )
     // number of mixer panel rows
     SetNumericIniSet ( IniXMLDocument, "client", "numrowsmixpan", iNumMixerPanelRows );
 
+    // feedback detection
```
ann0see, line 643: The comment is probably wrong?

ann0see (dismissed):

henkdegroot:
> Commenting the microphone permission doesn't sound right. Did you see what exactly failed?

I found some articles on www indicating the android:required="true"can cause issues when the item is already loaded due to a dependency. Sounds like adding the qtmultimedia option on Android already include the microphone feature.

This is the error in the log:

```
> Task :extractDeepLinksDebug
/home/runner/work/jamulus/jamulus/build/AndroidManifest.xml:22:5-87 Error:
> Task :processDebugManifest FAILED
/home/runner/work/jamulus/jamulus/build/AndroidManifest.xml:12:5-66 Warning:
	Element uses-feature#android.hardware.microphone at AndroidManifest.xml:22:5-87 duplicated with element declared at AndroidManifest.xml:21:5-89
	Element uses-permission#android.permission.INTERNET at AndroidManifest.xml:12:5-66 duplicated with element declared at AndroidManifest.xml:7:5-67
/home/runner/work/jamulus/jamulus/build/AndroidManifest.xml Error:
	Validation failed, exiting
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':processDebugManifest'.
> Manifest merger failed with multiple errors, see logs
```
The INTERNET permission is not an issue. As that appears in other android autobuilds as well.
Perhaps we should try removing the "comment" line which auto inserts the required features?

src/settings.cpp
henkdegroot: Well spotted!, will update

henkdegroot:
> Perhaps we should try removing the "comment" line which auto inserts the required features?

I have tested this on my own branch and removing the "insert comment" for the features also resolves the build issue, however does not resolve the crash on app start. Qt should play wav files fine on Android when using QSoundEffect (at least according to the information I found).  An alternative approach would be to not implement this audio alert feature on Android (assuming that is causing the crash).

henkdegroot:
I did some tracing on the Android app and found the following:

```
06-07 07:36:11.893 14336 14352 I QtCore  : Start
06-07 07:36:11.962 14336 14352 E AndroidRuntime: FATAL EXCEPTION: qtMainLoopThread
06-07 07:36:11.962 14336 14352 E AndroidRuntime: Process: com.github.jamulussoftware.jamulus, PID: 14336
06-07 07:36:11.962 14336 14352 E AndroidRuntime: java.lang.UnsatisfiedLinkError: dlopen failed: library "libQt5Quick_armeabi-v7a.so" not found
```
So far the only solution I found is to remove the armeabi-v7a option for the Android ABIS. However when I do that the app is no longer compatible with my android device. Not sure how we can get this library included.

pljones:
>  Not sure how we can get this library included.

Is the library needed?  That is, it's failing to run because of an unsatisfied _declared_ dependency.  Will the _build_ complete without that dependency and will the app then run?

henkdegroot:
> Will the _build_ complete without that dependency and will the app then run?

The build currently completes. so does not seem to "miss" this library. Guessing adding the qtmultimedia is adding something into the mix which is not detected as required by the build process but when the app runs on the android device is still referenced. Don't know which reference though...

.github/autobuild/android.sh
```diff
@@ -70,7 +70,7 @@ setup_qt() {
         echo "Installing Qt..."
         python3 -m pip install "aqtinstall==${AQTINSTALL_VERSION}"
         python3 -m aqt install-qt --outputdir "${QT_BASEDIR}" linux android "${QT_VERSION}" \
-            --archives qtbase qttools qttranslations qtandroidextras
+            --archives qtbase qttools qttranslations qtandroidextras qtmultimedia
```
hoffie, line 73: I've had a quick look at `qtmultimedia` regarding the crash/dynamic linking problem. It looks like `qtmultimedia` contains a library (ibQt5MultimediaQuick_armeabi-v7a.so) which introduces a dependency on libQt5Quick_armeabi-v7a.so, which we do not download/package.

As a quick workaround, we could try to install that library as well. However, it sounds wrong to do so as it's QML-related (I think), which we do not use at all:

```suggestion
            --archives qtbase qttools qttranslations qtandroidextras qtmultimedia qtdeclarative
```

The real path forward should be getting rid of the unused dependency. I don't know if there are any magic flags to accomplish that. One thing we could try would be removing the unused libraries in order to avoid them being included in the APK:
```suggestion
            --archives qtbase qttools qttranslations qtandroidextras qtmultimedia
        # Delete libraries, which we don't use, but which bloat the resulting package and might introduce unwanted dependencies.
        find "${QT_BASEDIR}" -name 'libQt5*Quick*.so' -delete
        rm -r "${QT_BASEDIR}/${QT_VERSION}/android/qml/"
```
I haven't tried it, but I imagine that it could work.

Some logs from my testing:
```
# aqt install-qt --outputdir /tmp/x linux android 5.15.2 --archives qtmultimedia qtquick
# for x in /tmp/x/5.15.2/android/lib/libQt5*-v7a.so; do echo FILE $x; readelf -a $x | grep -P '^FILE|libQt5Quick_armeabi-v7a.so'; done
FILE /tmp/x/5.15.2/android/lib/libQt5MultimediaQuick_armeabi-v7a.so
 0x00000001 (NEEDED)                     Shared library: [libQt5Quick_armeabi-v7a.so]
FILE /tmp/x/5.15.2/android/lib/libQt5MultimediaWidgets_armeabi-v7a.so
FILE /tmp/x/5.15.2/android/lib/libQt5Multimedia_armeabi-v7a.so
```

pljones: Are there alternatives to `qtmultimedia` that aren't as bundled - i.e. omit the unwanted dependencies?

hoffie: I don't think so. The root cause might be in `androiddeployqt`'s heuristic for finding dependencies, which is documented to be prone to false positives:

https://doc.qt.io/qt-5/deployment-android.html#dependencies-detection
https://doc.qt.io/qt-5/qmake-variable-reference.html#android-deployment-dependencies
> However, since run-time usage of plugins cannot be detected, there could be false positives

Seems like one can disable auto-detection, but that would require us to list all libraries in the correct order ourselves (`ANDROID_DEPLOYMENT_DEPENDENCIES`).

I guess I'd still give the library removal a chance.

henkdegroot: @hoffie thank you so much for the hard work and details provided. Tested adding the qtdeclarative in my own branch and with that in place the android no longer crashes and the audio alerts are working as well. I have just update the android.sh script in this PR as well to get this updated as well.

> I guess I'd still give the library removal a chance.

Will look into this next.

henkdegroot:
@chigkim, perhaps you can update the PR description to include an update is required to the documentation. As the new GUI setting needs to be include in the documentation. Looks like we are getting close to get this working/completed.

chigkim:
Thanks everybody!!!

So should we say for this to work on Windows, your system sound output 
should be different than what Jamulus uses, and the sound alerts play 
through what's set in system sound output.

It works on other operating systems as expected right?

.github/autobuild/android.sh
henkdegroot: @hoffie, you are a star! The removal of these unwanted dependencies is also resolving the crash on Android. I have just updated and pushed the latest update. Qtdeclarative is now removed and the lib deletion in place. It worked fine on my own branch so sure it works here as well.

henkdegroot:
> So should we say for this to work on Windows, your system sound output should be different than what Jamulus uses, and the sound alerts play through what's set in system sound output. 

I have not tested with my audio device on Windows, but guess there might be scenario's or devices/drivers which do not require that. So instead of "should be different" I would propose to use "may need to be different depending on the audio device/driver being used".

> It works on other operating systems as expected right?

I tested on Android and it works correct.

BTW, you have added wav files to the solution. Where did you get them from? We may need to add some information to the about-libraries dialog.

chigkim:
@henkdegroot, does this sound good for Short description of changes?
Added sound Alert for new person and new chat message. The sound alerts play through OS sound output. for this to work on Windows, your system sound output may need to be different than what Jamulus uses depending on the audio device/driver being used. An update is required to the documentation, and the new GUI setting needs to be include in the documentation. The sound files come from Final Cut Pro Sound Effects library, and they are edited to be short.
Or, should I put this in CHANGELOG?

henkdegroot:
@chigkim, sounds like a good description to add in the "what need documented" section of the PR so that this can be handled.
I would keep the changelog as currently is. I checked the Finat Cut Pro and looks like the effects are rights-free, so should not be a problem to use these. I think we still need to add a reference to the about->libraries indicating they came from Final Cut Pro Sound Effects. @hoffie, what do you think? And can we add without a link to apple?

hoffie:
Does anyone have the source for the "we can use the files without problems"?
Is it https://www.apple.com/legal/sla/docs/FinalCutProX.pdf?

> F. Final Cut Sample Content. Except as otherwise provided, you may use the Apple and third party audio
file content (including, but not limited to, the built-in sound files, samples and impulse responses)
(collectively the “Sample Content”), contained in or otherwise included with the Apple Software, on a
royalty-free basis, to create your own original soundtracks for your film, video and audio projects. You
may broadcast and/or distribute your own soundtracks that were created using the Sample Content,
however, individual files may not be commercially or otherwise distributed on a standalone basis, nor
may they be repackaged in whole or in part as clipart, stock animation, audio samples, sound files or
music beds.

It doesn't sound to me as clear as I hoped (it seems to be mainly targeted at produced audio/video works, not at other software). We are not doing anything which is prohibited (publish as a collection) either, so I feel we are a bit in-between. IANAL though.

So, license-wise we should:
- be sure that we are allowed to use it.
- check whether we need to add copyright notices in the About dialog
- check whether we can add useful licensing metadata to the source file (I think @mirabilos once suggested that for other file types and I think it's a good idea; I think WAV files *can* contain metadata, but I don't see any license-related hints there yet, only "Logic Pro X")

mirabilos:
Christian Hoffmann dixit:

>however, individual files may not be commercially or otherwise distributed on a standalone basis, nor
>may they be repackaged in whole or in part as clipart, stock animation, audio samples, sound files or
>music beds.

Definitely nōn-free and unacceptable for FOSS.

bye,
//mirabilos (hats: BSD and Debian developer)
-- 
I believe no one can invent an algorithm. One just happens to hit upon it
when God enlightens him. Or only God invents algorithms, we merely copy them.
If you don't believe in God, just consider God as Nature if you won't deny
existence.		-- Coywolf Qi Hunt

mirabilos:
chigkim dixit:

>Should I just find something from freesounds? They're Creative Commons 
>Licensed.

More than half of all Creative Commons licences are nōn-free,
so that alone does not suffice. (Plus Debian doesn’t allow
any CC before their respective version 3.0 other than CC0-1.0
because they contain language considered nōn-free, but that
part I’m not personally an expert in.)

All but one CC licence are also very hard to honour correctly
(the attribution part is extremely difficult).

In addition, if the sound is mixed into something else (which
MAY include embedding into the executable) licence compatibility
needs to be considered…

pljones:
https://creativecommons.org/about/cclicenses/
CC-BY-SA and CC-0 are okay but I'd still like a CC-SA - no restrictions except that you can't put restrictions on others using the content.

mirabilos:
Peter L Jones dixit:

>https://creativecommons.org/about/cclicenses/
>CC-BY-SA and CC-0 are okay

and CC-BY

>but I'd still like a CC-SA - no restrictions except that you can't put
>restrictions on others using the content.

This is plain wrong, it has a huge amount of restrictions, including
the rather hard to get right BY clause (CC-*, except CC0, all have
that) and it cannot be mixed with other copyleft licences like (L)GPL.

bye,
//mirabilos
-- 
I believe no one can invent an algorithm. One just happens to hit upon it
when God enlightens him. Or only God invents algorithms, we merely copy them.
If you don't believe in God, just consider God as Nature if you won't deny
existence.		-- Coywolf Qi Hunt

pljones:
@chigkim @jamulussoftware/maindevelopers

Given the change as it stands isn't acceptable, should this be dropped from targetting 3.9.0 or is it still possible to complete?

hoffie:
Thanks @mirabilos.
As far as I can see, the "only" blocker would be the audio file licensing. I would really love to get this in for 3.9.0, but as it touches translations it must be done by 2022-06-30. @chigkim, @henkdegroot Would anyone of you be available to give this some further love?

I've tried to find some possible sounds:
- New user
  - https://freesound.org/people/WeakWero/sounds/584941/ (CC0-1.0)
  - https://freesound.org/people/stereostereo/sounds/124537/ (CC-BY-3.0)
  - https://freesound.org/people/deleted_user_4772965/sounds/256513/ (CC0-1.0)
- New message
  - https://freesound.org/people/Anaponce/sounds/535211/ (CC0-1.0)
  - https://freesound.org/people/TinyRay/sounds/624499/ (CC0-1.0)
  - https://freesound.org/people/zrrion_the_insect/sounds/469014/ (CC0-1.0)

Maybe there are even better candidates, but I guess some of those would work?

All of them need some editing (silence removal, some contain multiple knocks or just one although we want (?) two).

henkdegroot:
Believe this call should be made by @chigkim as the originator of the PR. Guessing two different sounds are required to be able to distinquish between the alerts.

chigkim:
If everyone is ok with the license from freesounds, I'd be happy to edit 
them to fit.

mirabilos:
chigkim dixit:

>If everyone is ok with the license from freesounds, I'd be happy to edit 
>them to fit.

Sure. CC0 is less work but CC-BY 3.0 or newer and probably CC-BY-SA will
also do, though getting attribution right is a chore with these.

henkdegroot:
@chigkim I think it will be good to add some text to the changelog message to indicate the sounds are coming from freesound and are CC0.

hoffie:
> @chigkim I think it will be good to add some text to the changelog message to indicate the sounds are coming from freesound and are CC0.

+ Git commit + WAV metadata (if not already existing). Maybe even the LICENSE file (although it isn't by-path right now).

chigkim:
@henkdegroot, I added the sound source information into the section "Does this change need documentation?". How does documentation work? Is there someone in charge of writing the documentation based on PR?
@hoffie, do you know how to add metadata into wave files? If so, do you mind putting them and commit? I tried to do it from Reaper, but meta data dialog doesn't seem to be accessible with a screen reader unfortunately.
Is there anything else do we need?

hoffie:
> @hoffie, do you know how to add metadata into wave files? If so, do you mind putting them and commit? I tried to do it from Reaper, but meta data dialog doesn't seem to be accessible with a screen reader unfortunately.

`ffmpeg -i src/res/sounds/new_....wav -metadata copyright=CC0 -metadata comment="Retrieved from https://freesound.org/people/... 2022-06-28, edited for Jamulus by chigkim under CC0" src/res/sounds/new_...-x.wav`
... worked for me and I've pushed the result (are you happy with the comment?).

> Is there anything else do we need?

I think this is good to go now.

hoffie (approved):
I've just added an additional freesound.org attribution to the About dialog.
We should probably clean up the git history a bit.
@chigkim Would you be ok with me doing that?

I'm approving the PR, but I guess it would be good to have two other maintainers to approve as I've touched this myself now.

henkdegroot:
> I've just added an additional freesound.org attribution to the About dialog.

Does that include the command line output as well? Just using my mobile phone at the moment, so cannot check easily.

henkdegroot:
> I've just added an additional freesound.org attribution to the About dialog.

Does that include the command line output as well? Just using my mobile phone at the moment, so cannot check easily.

henkdegroot:
> How does documentation work? Is

When the PR is merged, then this is usually picked by someone who maintains the website/docs. Usually this is @gilgongo....

henkdegroot:
> How does documentation work? Is

When the PR is merged, then this is usually picked by someone who maintains the website/docs. Usually this is @gilgongo....

hoffie:
Hrm, with #2672 merged, we seem to have a new issue with Qt multimedia. I haven't looked at the details yet. One issue might be [lack of `aqt` support for this version](https://github.com/jamulussoftware/jamulus/runs/7101459315?check_suite_focus=true#step:7:117) (should probably work nevertheless?).
Another potential cause is that [Qt6 changed the multimedia module](https://www.qt.io/blog/qt-multimedia-in-qt-6), although I don't see any class/function signature related failures (yet). I'll have another look tomorrow.

chigkim:
@hoffie, sure feel free to clean up the git history.
Also thanks for adding the metadata and letting me know about the ffmpeg command!
Hmm, not sure why Github failed to build for Mac now. It did build for mac before. For whatever worth, my Mac built without an error with qt 6.3.1.

hoffie:
> Hmm, not sure why Github failed to build for Mac now. It did build for mac before. For whatever worth, my Mac built without an error with qt 6.3.1.

This issue is now becoming apparent due to the merge of #2672 and a rebase of this PR (I guess?).
The cause seems to be around the Qt packaging. While `qtmultimedia` was an *archive* on Qt 5.x, it's a *module* on Qt 6.x. Therefore, it needs a different installation command. I'll push a potential fix in some minutes.

.github/autobuild/ios.sh
```diff
@@ -20,12 +20,12 @@ setup() {
...
         python3 -m pip install "aqtinstall==${AQTINSTALL_VERSION}"
         # Install actual ios Qt:
-        python3 -m aqt install-qt --outputdir "${QT_DIR}" mac ios "${QT_VERSION}" --archives qtbase qttools qttranslations
+        python3 -m aqt install-qt --outputdir "${QT_DIR}" mac ios "${QT_VERSION}" --archives qtbase qttools qttranslations qtmultimedia
         if [[ ! "${QT_VERSION}" =~ 5\..* ]]; then
             # Starting with Qt6, ios' qtbase install does no longer include a real qmake binary.
             # Instead, it is a script which invokes the mac desktop qmake.
             # As of aqtinstall 2.1.0 / 04/2022, desktop qtbase has to be installed manually:
-            python3 -m aqt install-qt --outputdir "${QT_DIR}" mac desktop "${QT_VERSION}" --archives qtbase
+            python3 -m aqt install-qt --outputdir "${QT_DIR}" mac desktop "${QT_VERSION}" --archives qtbase qtmultimedia
```
hoffie, line 28: I don't think we have to instsall the desktop qtmultimedia here. This call is just there to install the desktop `qmake` binary. I'm dropping this part of the change as part of the Qt6 fix.

hoffie:
@henkdegroot 
> Does that include the command line output as well?

Now it does. I've initially skipped it, as the feature is not usable without at least once activating it via the GUI (I think?), but as other GUI-only resources (e.g. flag icons) are mentioned in `--version` as well, I've added it there now, too.

mirabilos:
chigkim dixit:

>Also thanks for adding the metadata and letting me know about the ffmpeg command!

It’s usable but not perfect, see http://bugs.debian.org/1014027 ☺
(I didn’t know about it either before.)

bye,
//mirabilos
-- 
15:39⎜«mika:#grml» mira|AO: "mit XFree86® wär’ das nicht passiert" - muhaha
15:48⎜<thkoehler:#grml> also warum machen die xorg Jungs eigentlich alles
kaputt? :)    15:49⎜<novoid:#grml> thkoehler: weil sie als Kinder nie den
gebauten Turm selber umschmeissen durften?	-- ~/.Xmodmap wonders…

hoffie:
> As stated, we definitely should clean up the Git history, especially to avoid getting the intermediate non-free sound files into the history at all.

Done as well now.

pljones:
Sorry to be annoying...  Can the build script changes be split out, at least into a separate (initial) commit?

src/util.cpp
```diff
@@ -593,26 +593,27 @@ CAboutDlg::CAboutDlg ( QWidget* parent ) : CBaseDlg ( parent )
...
     // libraries used by this compilation
-    txvLibraries->setText ( tr ( "This app uses the following libraries, resources or code snippets:" ) + "<br><p>" +
-                            tr ( "Qt cross-platform application framework" ) +
-                            ", <i><a href=\"https://www.qt.io\">https://www.qt.io</a></i></p>"
-                            "<p>Opus Interactive Audio Codec"
-                            ", <i><a href=\"https://www.opus-codec.org\">https://www.opus-codec.org</a></i></p>"
+    txvLibraries->setText (
+        tr ( "This app uses the following libraries, resources or code snippets:" ) + "<br><p>" + tr ( "Qt cross-platform application framework" ) +
+        ", <i><a href=\"https://www.qt.io\">https://www.qt.io</a></i></p>"
+        "<p>Opus Interactive Audio Codec"
```
pljones, line 599: Mmm, should we get this translated?

hoffie:
> I thought the build script changes were needed first to get all the platforms working

The build scripts are only touched to add the newly required dependency. Yes, this has to be done first or at the same time as introducing the code which uses it. Not sure if you saw my question in the chat: I could split that part out into its own commit, but it wouldn't be meaningful on its own. We wouldn't add the `qtmultimedia` dependency if we weren't to add code which uses it. Therefore, I made that part a single commit. Let me know what you think.

> Being able to have the sound off needs to be available at the point the commit that turns the sounds on happens

Usability-wise I agree (i.e. it should be part of one PR and this is the case here). Technically, both states are buildable/usable on its own. If that's not acceptable, I would have to squash the GUI setting addition into the other commit as well. Do you want me to do that?

(Not sure when you've started reviewing -- your comment was posted only a few minutes after my rebasing/squash push; no code has changed, only git history did).

pljones:
Yes, I was checking the individual commits for sanity as a squash/rewrite had been done.  Based on what you've said, I think a single commit would make most sense.

pljones:
If I install this new version, does it require access to hardware that I might have dedicated to Jamulus and thus cause Jamulus to fail to load?

Say I have a single audio system with a sound drive that only allows exclusive access to the sound card and that's what I expect Jamulus to use.  Does _simply having this feature_ cause a problem under those circumstances (even when it's not switched on)?

One other fundmental issue I have with the approach is that it's adding non-visual code to ClientDlg.  That shouldn't happen.  Why would I not want the audio alerts if I were using the JSON RPC client without the graphic UI?

pljones (approved):
With comments as per conversation / chat.

ann0see:
> Say I have a single audio system with a sound drive that only allows exclusive access to the sound card and that's what I expect Jamulus to use. Does simply having this feature cause a problem under those circumstances (even when it's not switched on)?

That's exactly an/the issue which needs to be avoided.

android/AndroidManifest.xml
```diff
@@ -13,7 +13,7 @@
     <!-- The following comment will be replaced upon deployment with default features based on the dependencies of the application.
          Remove the comment if you do not require these default features. -->
     <!-- %%INSERT_FEATURES -->
-    <uses-feature android:name="android.hardware.microphone" android:required="true"/>
+    <!--<uses-feature android:name="android.hardware.microphone" android:required="true"/>-->
```
ann0see, line 16: That's not really good. Either remove it completely or leave it in.

src/clientdlg.cpp
```diff
@@ -253,6 +253,8 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     // set window title (with no clients connected -> "0")
     SetMyWindowTitle ( 0 );
+    // Set number of clients to track
```
ann0see, line 256: This comment should be a bit clearer. Why do we set clients to track? What do we track?

src/clientsettingsdlgbase.ui
```diff
@@ -242,6 +242,13 @@
                     </property>
                    </widget>
                   </item>
+                  <item>
+                   <widget class="QLabel" name="lblAudioAlerts">
+                    <property name="text">
+                     <string>Audio Alerts</string>
```
ann0see, line 248: --> Audio Alerts (might need secondary sound hardware)

hoffie: Hrm, not sure if there's space for so much text. Have you checked?

android/AndroidManifest.xml
hoffie: Agreed.
@henkdegroot, @chigkim  I've removed this completely now (it's still recoverable via git history). Leaving it in is not possible based on the history (it causes duplicates).

src/clientdlg.cpp
hoffie: Updated to `// track number of clients to detect joins/leaves for audio alerts`.

hoffie:
> Yes, I was checking the individual commits for sanity as a squash/fixup rebase had been done. Based on what you've said, I think a single commit would make most sense.

Ok. Squashed to a single commit now (with author information preserved as Co-authored-by).

> If I install this new version, does it require access to hardware that I might have dedicated to Jamulus and thus cause Jamulus to fail to load?
> Say I have a single audio system with a sound drive that only allows exclusive access to the sound card and that's what I expect Jamulus to use. Does simply having this feature cause a problem under those circumstances (even when it's not switched on)?

Based on my understanding, I would not expect such issues. Solely linking to qtmultimedia should not cause any audio devices to be considered "in use". All playback happens on demand, which means it happens after initializing Jamulus' main audio stack. So if exclusive access is a problem, I'd assume that the new feature breaks and not Jamulus' core functionality.

As far as I understand, this feature was part of @chigkim's a11y build which had already been in use by some people?

And as a last "Let's try it" argument: We are preparing a new minor release, which is not supposed to cause breakage, but is known to carry larger changes. We still have some time during betas to find possible issues.

> One other fundmental issue I have with the approach is that it's adding non-visual code to ClientDlg. That shouldn't happen. Why would I not want the audio alerts if I were using the JSON RPC client without the graphic UI?

Understandable, especially regarding future plans. I'll open that as a separate issue.

mirabilos:
Christian Hoffmann dixit:

>Based on my understanding, I would not expect such issues. Solely
>linking to qtmultimedia should not cause any audio devices to be
>considered "in use". All playback happens on demand, which means it
>happens after initializing Jamulus' main audio stack. So if exclusive
>access is a problem, I'd assume that the new feature breaks and not
>Jamulus' core functionality.

AIUI / in my experience (pure ALSA-based system, no Poettering stuff)
it is so that you have to start jackd (usually via qjackctl) before
starting Jamulus, and that fails if anything uses the sound device,
and afterwards any audio has to go via JACK anyway, so if anything
is playing once jackd is started, it either goes through JACK (and
thus is multiplexed) or is not audible (because jackd has exclusive
access to the soundcard) or goes through a different soundcard which
jackd does not use.

bye,
//mirabilos
-- 
21:12⎜<Vutral> sogar bei opensolaris haben die von der community so
ziemlich jeden mist eingebaut │ man sollte unices nich so machen das
desktopuser zuviel intresse kriegen │ das macht die code base kaputt
21:13⎜<Vutral:#MirBSD> linux war früher auch mal besser :D

src/clientsettingsdlgbase.ui
pljones: Maybe we could add a tooltip later - and something clear in the documentation?

src/clientdlg.cpp
```diff
@@ -254,6 +254,9 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // set window title (with no clients connected -> "0")
     SetMyWindowTitle ( 0 );
 
+    // track number of clients to detect joins/leaves for audio alerts
+    iClients = 0;
```
pljones, line 258: Shouldn't this happen every time I connect to a server?

hoffie: Wouldn't this be covered by OnNumClientsChanged? As it sets the absolute value, it doesn't necessarily have to be zeroed on connect, right?

pljones: OnNumClientsChanged tests for a change, doesn't it?  If when I was on server A iClients was 3 and I join server B - what should happen?  What if there are 3 clients there, too -- all different ones?

hoffie: > OnNumClientsChanged tests for an increase, doesn't it?

Yes, and afterwards (even if the test was false) it stores the new value.

I've added the following logging for testing locally:
```
diff --git a/src/clientdlg.cpp b/src/clientdlg.cpp
index a488365a..849564ae 100644
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -885,6 +885,7 @@ void CClientDlg::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo
 
 void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
 {
+    qInfo() << "OnNumClientsChanged iNewNumClients =" << iNewNumClients << ", iClients =" << iClients;
     if ( pSettings->bEnableAudioAlerts && iNewNumClients > iClients )
     {
         QSoundEffect* sf = new QSoundEffect();
```

When joining an empty server, this happens:
```
OnNumClientsChanged iNewNumClients = 0 , iClients = 0 <- this would additionally reset
OnNumClientsChanged iNewNumClients = 1 , iClients = 0 <- this triggers the sound
OnNumClientsChanged iNewNumClients = 1 , iClients = 1 <- this is just a duplicate and doesn't trigger the sound

OnNumClientsChanged iNewNumClients = 0 , iClients = 1 <- on disconnect, this properly resets
```

(Similar behavior on non-empty servers)

So, as far as I can see, it works properly and it can also be explained why it does. :)

This is the call chain which guarantees a proper reset on disconnect:
CClientDlg::Disconnect() -> MainMixerBoard->HideAll() -> emit NumClientsChanged ( 0 ); -> CClientDlg::OnNumClientsChanged()

pljones: Great, thanks for checking.  I think it's worth a comment in the code (in `OnNumClientsChanged` where it stores `iNewNumClients` into `iClients` that this will reset the value to zero when connecting to a new server.

src/clientdlg.cpp
```diff
@@ -876,6 +885,13 @@ void CClientDlg::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo
 
 void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
 {
+    if ( pSettings->bEnableAudioAlerts && iNewNumClients > iClients )
+    {
+        QSoundEffect* sf = new QSoundEffect();
+        sf->setSource ( QUrl::fromLocalFile ( ":sounds/res/sounds/new_user.wav" ) );
+        sf->play();
+    }
+    iClients = iNewNumClients;
```
pljones, line 894:
```suggestion
    // iNewNumClients will be zero on the first trigger of this signal handler when connecting to a
    // new server.  Subsequent triggers will thus sound the alert (if enabled).
    iClients = iNewNumClients;
```

src/clientsettingsdlgbase.ui
hoffie: The *What's this* text does have a hint regarding that already, btw.

src/clientdlg.cpp
hoffie: Added, thanks.

ann0see:
@hoffie this PR created a wrong CHANGELOG entry.

pljones:
Also, we need to get COMPILING.md updated to mention that `qtmultimedia5-dev` needs installing as a build time dependency (at least on Ubuntu 20 LTS).

henkdegroot:
> Also, we need to get COMPILING.md updated to mention that `qtmultimedia5-dev` needs installing

I believe I did this already as part of this PR....can you check please?

pljones:
Ah sorry, I meant to come back and say it was user error on my part.  I've noted elsewhere it needs highlighting as part of the release.

gilgongo:
Screenshot for website. Ignore.

jamulus-settings

chigkim:
r3_9_0beta2 didn't pick up:
CHANGELOG: Added sound Alert for new person and new chat message.
How should I format the line to be included?

ann0see:
I think @pljones forgot updating the changelog for this beta

chigkim:
Ah ok. Last time I checked, asio4all locks up and prevent other apps to 
play sound through the card it uses. I suspect asio4all is blocking QT 
to play sound with QSoundEffect.

If I use asio4all with built-in sound card, it even blocks screen 
reader. That's why screen reader users would use an external audio 
interface with its own Asio driver for music apps like Jamulus and use 
built-in sound card for OS system output and screen readers.

Do you have an external audio interface with Asio driver that you could 
test with?

Skip asio4all and use that audio interface with Jamulus. Then set your 
built-in soundcard as your OS system output.

Then hopefully you'll hear it.

PR #2643, opened 2022-05-28
Re-examine internal/external IP for serverlist

pljones:

I'm hoping this change at least _clarifies_ what's meant to happen with the more complicated client/server/directory trios, i.e. where one or other - or both - of the client and server are local to the directory itself.

CHANGELOG: Refactor: handling of requests for the server list.

**Context: Fixes an issue?**

Not that I can remember.

Should be noticeable in most situations and I don't think we document at this level of detail.

My directories run with it (at seem to get used).  It's not fixed anything or broken anything as far as I can tell.

Someone else reading it and understanding it...

## Checklist

### DIFF
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -23,6 +23,34 @@
 \******************************************************************************/
 
 #include "serverlist.h"
+/* *\
+
+--port           sets the port the server listens to locally
+--serverbindip   sets the IP   the server listens to locally
+--serverpublicip sets the public IP where server and directory are on the same LAN
+
+Manual port forwarding is not supported: the internal port must be open to external requests.
+Where a router modem does automatic port forwarding, directory pings MAY open the server to requests
+from clients not on the same LAN as the directory.
+
+*
+ PROTMESSID_CLM_SERVER_LIST
+ - SERVER internal to DIRECTORY (list entry external IP     same LAN)
+   - CLIENT internal to DIRECTORY (CLM msg     same LAN): use "external" fields (local LAN address)
+   - CLIENT external to DIRECTORY (CLM msg not same LAN): use "internal" fields (registered server public IP)
+ - SERVER external to DIRECTORY (list entry external IP not same LAN)
+   - CLIENT internal to SERVER (CLM     same IP as list entry external IP): use "internal" fields (local LAN address)
+   - CLIENT external to SERVER (CLM not same IP as list entry external IP): use "external" fields (public IP address from protocol)
+---
+*
+ PROTMESSID_CLM_REGISTER_SERVER
+ - As SERVER create CLM message using "internal" fields (IP layer is the "external" fields):
+   - DIRECTORY not on my LAN: self-determined  IP address and --port value
+   - DIRECTORY     on my LAN: --serverpublicip IP address and --port value
+ - As DIRECTORY store received CLM message as is (IP layer address is used to access the server list)
+*
+
+\* */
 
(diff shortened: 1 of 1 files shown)


gene96817:
@pljones    What is the internal/external IP issue that you are addressing? Since I am often supporting uses that trip over localIP/ExternalIP/NAT issues, I am interested in following the full range of issues.  THX

pljones:
Think of a LAN behind an external router.  (I'll not get into the router configuration for ports.)

Say the Directory is 10.0.1.23
Say the Server is 10.0.1.24

Say the external address of the router is 86.212.104.22

When an internal client wants the server IP, it wants 10.0.1.24

When an external client wants the server IP, it wants 86.212.104.22

Now say a server not on the LAN registers with an IP address of 37.141.86.185

When an internal (to the directory) client wants the server IP, it wants 37.141.86.185

When an external (to the directory) client wants the server IP, it wants ... what?  Well... 37.141.86.185 if it's not internal to that server but it wants the server LAN address if it is.

gene96817:
@pljones  If the person operating the servers is the only one using a client on the LAN, that person is technically savvy and they can choose whatever they like.

In my case, I want a solution for guest on my LAN. In this case, I want a solution were the guest (being less technically savvy)  has the same user experience on my LAN as when connected to the external network. For this goal, I put routing table entries so that the router routes the external IP address to the matching local address. 

In your example, an internal client would request 86.212.104.22 and the router will route the messages to 10.0.1.24.
Similar messages to 37.141.86.185 would be routed to 10.0.1.23.

Now we get to the inconsistency of your example.
Your router's external address is 86.212.104.22. All external messages to your private LAN needs to be sent to this router address. I expected the external address for your Directory to be 86.212.104.22:PortA and the Server to be reached at 86.212.104.22:PortB. You would have to set up the port mapping table.

Unfortunately this requires a discussion about routing tables and port mapping. Are you on a path for a different solution?
I believe the desired solution needs to be invisible to the client user (meaning the Jamulus client code doesn't need to know anything about the machinery).

pljones:
@gene96817 I'm not changing the intended current behaviour, just simplifying the code, as far as I'm aware.  If you want to start a discussion on the topic, the discussions area is where you should do it, not during code review.

gene96817:
@pljones    Thanks for your work. It just wasn't clear to me what you were trying to fix.

src/serverlist.cpp
```diff
@@ -23,6 +23,34 @@
 \******************************************************************************/
 
 #include "serverlist.h"
+/* *\
+
+--port           sets the port the server listens to locally
+--serverbindip   sets the IP   the server listens to locally
+--serverpublicip sets the public IP where server and directory are on the same LAN
```
hoffie, line 30:
```suggestion
--serverpublicip sets the public IP when server and directory are on the same LAN
```
?

hoffie (approved):
Looks sane to me.

softins (approved):
I _think_ this is ok. I've studied it for a while, and compiled it, but I don't currently have time to run any test cases.

ann0see:
I think the Changelog entry should be made a bit clearer (what would an interested user want to read? What has changed?)

pljones:
Hopefully nothing has changed.  It's to make the _code_ clearer.  So it's aimed purely at someone reading the code.  Really only smoeone looking at the diff would be interested.

ann0see:
Then the changelog should also say "clarify" and "Code" 

Refactoring: Clarify handling of requests for the server list in code

Issue #2480, opened 2022-03-09
Autobuild: Make Changelog generation an extra step


hoffie:
**What is the current behaviour and why should it be changed?**

The Changelog is currently generated by `analyze_git_references.py`. This is surprising...

**Describe possible approaches**

Changelog generation should be an individual step.

**Has this feature been discussed and generally agreed?**

https://github.com/jamulussoftware/jamulus/pull/2470/files#r822584276

pljones:
@jamulussoftware/maindevelopers Did this happen, does anyone know?

hoffie:
Meh, this was supposed to be fixed in #2584 already. Somehow it's missing most of the mentioned commits. I will try to re-submit a fixed up #2584 instead.

PR #2584, opened 2022-04-06
Autobuild: Improve and reorganize git/Changelog scripts

hoffie:
**Short description of changes**

- Autobuild: Don't hardcode python path in analyse_git_reference.py
  
- Autobuild: Move Changelog generation to the workflow
  Previously, the autobuild changelog (autoLatestChangelog.md) which is used for the Github release body, was generated as part of the "analyse_git_reference.py" script. It does not seem logical for an analysis script to have a side effect of producing a Changelog file. Also, the file name (autoLatestChangelog.md) was hardcoded in two places and had to match (workflow + script).
  With this commit, the changelog generation call (.github/actions_scripts/getChangelog.pl) is moved to the workflow. This follows the principle of least surprise and moves the output file name usage to a single file (the workflow).

- Autobuild: Improve variable naming/coding style in analyse_git_reference.py
  
- Autobuild: Move .github/action_scripts to .github/autobuild
  This also renames the two scripts to better reflect what they do.

- Autobuild: Drop redundant github.workspace variable reference in workflow

CHANGELOG: Autobuild: Improved and reorganized git/Changelog scripts
(to be condensed with the other entries)

**Context: Fixes an issue?**

I think it had been mentioned elsewhere
- that .github/action_scripts/* should be moved to .github/autobuild/ instead, as that's where they are used now
- that an `analyze_something` script is not supposed to have the side effect of writing a (Changelog) file

**Does this change need documentation? What needs to be documented and how?**

No.

**Status of this Pull Request**

Ready.

**What is missing until this pull request can be merged?**

Reviews.

## Checklist

-  [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
-  [x] I tested my code and it does what I want:
  - Build is still green
  - Release/changelog logic still works: https://github.com/hoffie/jamulus/releases/tag/r0_0_1devHoffieTest
-  [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
-  [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
-  [x] I've filled all the content above

PR #2655, opened 2022-06-18
Autobuild: Make release changelog generation explicit

hoffie:
**Short description of changes**
- Autobuild: Use a variable instead of hardcoding the release changelog path twice
- Autobuild: Make release changelog generation explicit
  This was previously done as part of the variable extraction step.
  It's rather surprising to have a script with the name "analyze*" write
  to some Changelog file.
  
  Therefore, move all of the generation to the same abstraction level (Github workflow).

CHANGELOG: (List along with the other autobuild improvements)

**Context: Fixes an issue?**
Fixes #2480

**Does this change need documentation? What needs to be documented and how?**
No.

**Status of this Pull Request**
Tested on my repo: https://github.com/hoffie/jamulus/runs/6948860313?check_suite_focus=true#step:5:1

**What is missing until this pull request can be merged?**
Reviews.

## Checklist

-  [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
-  [x] I tested my code and it does what I want
-  [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
-  [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
-  [x] I've filled all the content above

PR #2656, opened 2022-06-18
Autobuild: Improve and reorganize git/Changelog scripts (re-submission of #2584)

hoffie:
#2584 was already supposed and documented to contain these changes, but didn't (probably due to a bad rebase). Therefore, I'm re-submitting it and will edit the old PR to match what was actually merged.

- Autobuild: Don't hardcode python path in analyse_git_reference.py
  
- Autobuild: Move Changelog generation to the workflow
  Previously, the autobuild changelog (autoLatestChangelog.md) which is used for the Github release body, was generated as part of the "analyse_git_reference.py" script. It does not seem logical for an analysis script to have a side effect of producing a Changelog file. Also, the file name (autoLatestChangelog.md) was hardcoded in two places and had to match (workflow + script).

  With this commit, the changelog generation call (.github/actions_scripts/getChangelog.pl) is moved to the workflow. This follows the principle of least surprise and moves the output file name usage to a single file (the workflow).

  Fixes: #2480

- Autobuild: Improve variable naming/coding style in analyse_git_reference.py
  
- Autobuild: Move .github/action_scripts to .github/autobuild
  This also renames the two scripts to better reflect what they do.

- Autobuild: Use a variable instead of hardcoding the release changelog path twice
  Related: #2480

CHANGELOG: Autobuild: Improved and reorganized git/Changelog scripts
(to be condensed with the other entries)

**Context: Fixes an issue?**
Fixes: #2480

No.

Ready for review.

Reviews.

## Checklist

-  [x] I tested my code and it does what I want: Yes, on my fork: https://github.com/hoffie/jamulus/runs/6950719696?check_suite_focus=true#step:4:5 https://github.com/hoffie/jamulus/releases/tag/r0_0_1


.github/autobuild/get_build_vars.py
```diff
@@ -32,36 +32,24 @@ def get_git_hash():
...
-        version,
-    ])
-    with open(f'{REPO_PATH}/autoLatestChangelog.md', 'wb') as f:
-        f.write(changelog)
-
-
-def get_build_version(jamulus_pro_version):
+def get_build_version():
+    jamulus_pro_version = get_version_from_jamulus_pro()
     if "dev" in jamulus_pro_version:
```
pljones, line 37: I'm not keen on logging in this function.  I'd rather it returned a tuple of the type and version and the caller could choose to log or not.
```
type = "release"
version = jamulus_pro_version
if "dev" in jamulus_pro_version:
    type = "dev"
    version = "{}-{}".format(version, get_git_hash())
return type, version
```
```
type, version = get_build_version()
print(f"building a {type} version: {version}")
```

hoffie: I don't feel strongly about that and I think I've just taken what was there.
I have now (mostly) applied your suggestion.

pljones (approved):

.github/autobuild/get_build_vars.py
pljones: Yep wasn't fussed as it was already there.  But as it was potentially useful information (i.e. build type), it might as well be exposed to the caller.

.github/workflows/autobuild.yml
```diff
@@ -62,15 +62,21 @@ jobs:
...
 
-      - name:                       Determine release version, type and prerelease variables and generate Changelog
-        run:                        python3 ${{ github.workspace }}/.github/actions_scripts/analyse_git_reference.py
+      - name:                       Determine release version, type and prerelease variables
+        run:                        ./.github/autobuild/get_build_vars.py
         id:                         get-build-vars
 
+      - name:                       Extract Changelog for the Github release body
+        if:                         steps.get-build-vars.outputs.PUBLISH_TO_RELEASE == 'true'
+        run:                        ./.github/autobuild/extractVersionChangelog.pl ChangeLog $(grep -oP 'VERSION = \K\w[^\s\\]*' Jamulus.pro) > ${{ env.release_changelog_path }}
```
ann0see, line 78: Not that happy with a regex and grep of the version here. Don't we have the version somewhere else?

hoffie: The very version we need here is available in the script, but currently isn't output. We can change that, but it would be a more distributed change. See here:
https://github.com/jamulussoftware/jamulus/pull/2655/files#diff-54b7cee412550302a3e076162e611b9719407b264c79553452499450c54a07e5R52 https://github.com/jamulussoftware/jamulus/pull/2655/files#diff-55d631a22c9e350832273ee0abf8a6013435fa6051789cbe311b981b2ec0f348R88

Would you prefer that?

pljones: That's rather why I was wondering if returning the release type from `get_version_from_jamulus_pro()` might help elsewhere.

ann0see: Yes, personally, I'd prefer this since it feels cleaner.

.github/workflows/autobuild.yml
```diff
@@ -62,15 +62,21 @@ jobs:
       publish_to_release:           ${{ steps.get-build-vars.outputs.PUBLISH_TO_RELEASE }}
       upload_url:                   ${{ steps.create-release.outputs.upload_url }}
       build_version:                ${{ steps.get-build-vars.outputs.BUILD_VERSION }}
+    env:
+      release_changelog_path:       ./.github_release_changelog.md
```
ann0see, line 66: Is this file in the .gitignore?

ann0see (commented):
Thanks. Just another comment:

.github/workflows/autobuild.yml
hoffie: Ok, updated and re-tested on my fork (updated the link in the PR description).

.github/autobuild/get_build_vars.py
hoffie: > On which note - is the check for `dev` here the consistent check now?

Sorry, I missed that part -- Github doesn't send notifications for edits.
Can you elaborate on what you mean here?

.github/workflows/autobuild.yml
hoffie: Nice catch, thanks. The old name was and I had grepped for that, but missed hidden files. Updated .gitignore now.

.github/autobuild/get_build_vars.py
```diff
@@ -60,8 +47,9 @@ def set_github_variable(varname, varval):
...
 
 jamulus_pro_version = get_version_from_jamulus_pro()
-write_changelog(jamulus_pro_version)
-build_version = get_build_version(jamulus_pro_version)
+set_github_variable("JAMULUS_PRO_VERSION", jamulus_pro_version)
+build_type, build_version = get_build_version(jamulus_pro_version)
+print(f'building a version of type "{build_type}": {build_version}')
 
 fullref = os.environ['GITHUB_REF']
 publish_to_release = bool(re.match(r'^refs/tags/r\d+_\d+_\d+\S*$', fullref))
```
pljones, line 55: OK, can we _either_ do this or do what line 36 (`if "dev" in jamulus_pro_version:`) does, rather than have both ways in use.

I guess, ideally that means another function `get_release_type()` that just returns whether this is a `release` or not.  It can be boolean, to allow the caller to decide how to use the result - but if we're using the tag (as here) to drive what gets triggered in the automation, then the scripts (that get triggered) should also just use the tag.

So
- do we need to differentiate between `GITHUB_REF` version and `Jamulus.pro` version?
- if so, we should be consistent and use `GITHUB_REF` at least to drive whether a build is a release or not
- if not, we should be completely consistent and use one or other - everywhere

hoffie: > OK, can we _either_ do this or do what line 36 (`if "dev" in jamulus_pro_version:`) does, rather than have both ways in use.

I don't quite see yet how one could replace the other.

> I guess, ideally that means another function `get_release_type()` that just returns whether this is a `release` or not. It can be boolean, to allow the caller to decide how to use the result

That would be possible, for sure.

>  but if we're using the tag (as here) to drive what gets triggered in the automation, then the scripts (that get triggered) should also just use the tag.
> 
>  * do we need to differentiate between `GITHUB_REF` version and `Jamulus.pro` version?

With the current design, assumptions and constraints, we have to differentiate, I think. `GITHUB_REF`(= tag) might be different (e.g. for pre-releases) or not existing at all (branches/master).

>  * if so, we should be consistent and use `GITHUB_REF` at least to drive whether a build is a release or not

I think this should be the case already. We use the workflow variable `PUBLISH_TO_RELEASE` to drive the build. This variable is solely driven by `GITHUB_REF`.

>  * if not, we should be completely consistent and use one or other - everywhere

I think there's certainly room for improvement, but it needs thorough checking to avoid breaking existing use cases (such as local builds). In addition, I don't think this PR introduces any additional inconsistencies. It might make existing strangeness more obvious (which is a good thing, IMO). Therefore, I'm a bit hesitant to squeeze in additional larger changes here.

pljones: My concern is `get_build_version()` uses a different value.  The tag, `Jamulus.pro` and `ChangeLog` could all have different versions.

Given we have three sources for the version, we really need automation to make sure they're aligned, at least: i.e. error out if things look wrong.  But I agree, that's not in scope here.  For that, it sounds like (and I don't fully understand how this works):
* If the tag is set, we should use that to determine build type and version and that source should overrule others.
* ChangeLog and Jamulus.pro "target" versions should match - and, if set, match the tag build type and version.
* If the tag wasn't set, build type and version can then come from ChangeLog and Jamulus.pro, as needed.

pljones (approved):

.github/autobuild/get_build_vars.py
hoffie: Yep. I've split that out into #2662 to track that.

ann0see (approved):

Issue #2688, opened 2022-07-02
Improve Windows "headless" operation


pljones:
I just happened upon a StackOverflow thread I don't think I'd read before about handling the command line on Windows with Qt.

The basic problem we have now is that if you just go ahead and use the default approach, your app goes into the background and the console issues the next command prompt.  Which is a bit odd and probably not wanted.  Why it's like that is because _most_ Qt apps _aren't_ headless, of course...

What's suggested here is, on Windows, add `CONFIG += console` and then create your QtCoreApplication.  And it'll stay attached.  We currently have something like
```
#ifdef _WIN32
if (AttachConsole(ATTACH_PARENT_PROCESS) || AllocConsole()){
    freopen("CONOUT$", "w", stdout);
    freopen("CONOUT$", "w", stderr);
    freopen("CONIN$", "r", stdin);
}
#endif
```
to reattach - but, too late! We're in the background.  Instead, we'd want
```
#ifdef _WIN32
::ShowWindow( ::GetConsoleWindow(), SW_HIDE );
#endif
```
when running _with_ the GUI to close any launch console.  (Might need tuning - maybe check `::GetConsoleWindow()` is valid, etc.)

https://stackoverflow.com/questions/45500880/qt-console-application-with-optional-gui

I've not yet tried this out but it _could_ make Jamulus on Windows that bit slicker.

pljones:
As this is tiny - and done - I've added it to 3.9.0.

pljones:
Closing.  We've decided not to spend further effort on this annoying Windows-specific behaviour.

PR #2690, opened 2022-07-02
#2688 Make Windows console handling slick

pljones:

Rather than the Windows version of Jamulus being a GUI application that tries to attach a console when running "headless", this makes the program a command line program that cedes the console when running the GUI.  The console is only minimized.  Once Jamulus exits, the minimized console is restored.

If no console were in use, there is no effect (you don't get a console appearing and vanished or anything like that).

**EDIT** - **SEE COMMENTS BELOW** - Apparently this _does_ in fact happen.  Under investigation.

CHANGELOG: Improvement to Windows console handling

**Context: Fixes an issue?**

Fixes: #2688 

No.

Tested locally under Windows, both headless and GUI.  (No effect on other platforms.)

Should be good to go.

## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -130,14 +130,6 @@ int main ( int argc, char** argv )
     Q_UNUSED ( bCustomPortNumberGiven )
 #endif
 
-#if !defined( HEADLESS ) && defined( _WIN32 )
-    if ( AttachConsole ( ATTACH_PARENT_PROCESS ) )
-    {
-        freopen ( "CONOUT$", "w", stdout );
-        freopen ( "CONOUT$", "w", stderr );
-    }
-#endif
-
     // When adding new options, follow the same order as --help output
 
     // QT docu: argv()[0] is the program name, argv()[1] is the first
@@ -771,6 +763,14 @@ int main ( int argc, char** argv )
         }
     }
 
+#if defined( _WIN32 )
+    // If we are using the GUI, minimize any attached console
+    if ( bUseGUI && ::GetConsoleWindow() )
+    {
+        ::ShowWindow ( ::GetConsoleWindow(), SW_SHOWMINNOACTIVE );
+    }
+#endif
+
     // Application/GUI setup ---------------------------------------------------
     // Application object
 #ifdef HEADLESS
@@ -1031,6 +1031,13 @@ int main ( int argc, char** argv )
     activity.EndActivity();
 #endif
 
+#if defined( _WIN32 )
+    // If we are using the GUI, restore any attached console
+    if ( bUseGUI && ::GetConsoleWindow() )
+    {
+        ::ShowWindow ( ::GetConsoleWindow(), SW_SHOWNORMAL );
+    }
+#endif
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -1031,6 +1031,13 @@ int main ( int argc, char** argv )
     activity.EndActivity();
 #endif
 
+#if defined( _WIN32 )
+    // If we are using the GUI, minimize any attached console
```
pljones, line 1035: Should say "restore" here.

pljones: Is done :)

src/main.cpp
```diff
@@ -1031,6 +1031,13 @@ int main ( int argc, char** argv )
     activity.EndActivity();
 #endif
 
+#if defined( _WIN32 )
```
pljones, line 1034: Maybe both `activity.EndActivity()` and `::ShowWindow()` blocks should also happen inside the no GUI side of the `catch`...

hoffie (approved):
At first glance, the approach seems counter-intuitive to me (especially `CONFIG+=console`), but if it works, go for it. Should probably get some more validation by other Windows users.

> If no console were in use, there is no effect (you don't get a console appearing and vanished or anything like that).

I think that's important.

Does this PR improve or even fix #1281, btw?

pljones:
> Does this PR improve or even fix #1281, btw?

Thanks!  I couldn't find that.

Yeah, the `CONFIG += console` is a bit weird.  But then so is Qt on Windows deciding to go into the  background by default.  This just reverses that "decision" and the effects it has.

henkdegroot:
As a windows user, I really don't see the benefit in this and this is even annoying me. As when I start the application I briefly see a black window/dialog appearing which then disappears and I suddenly have two "windows" running for just one application. 

image

The minized window also contains text related to the allocated port number and perhaps other information which Windows users are not commonly interested in. 

image

Perhaps it is possible to have a "console" switch/option which can enable this output window when you really want to see that information. In general Windows users don't think about using the command-line or only display it when you actually start Jamulus from the command-line.

I appreciate the effort to make it more slick, but think this will not liked by "regular" windows users.

pljones:
Strange - I don't get that.  How are you running Jamulus?  If you're just using a shortcut, it shouldn't allocate a console at all.  If you run it from a console, it won't detach -- it's not meant to, you chose to run it from a console, after all.

henkdegroot:
@pljones, just run the app using the Jamulus shortcut (created by the installer) in the start menu. I am using Windows 11, maybe different windows versions re-act differently?

pljones:
Ugh, Win11... That _could_ be it.  The API call is marked "deprecated" with absolutely no indication of what the current method to achieve the same outcome is.  Standard approach for Microsoft.  "We've got this new toy that does 60% of what the old one did (and a whole load of stuff _you'll never use but we think is great_) - so start using the new toy now and never look back!  It's great!"  (OK, they keep the old stuff around for 25 or more years before turning it off sometimes...)

pljones:
Hm.  Strange.  It _does_ do it for me from the installer.  I'm on Qt 5.12 rather than Qt 5.15 but it's not going to be that...  OK, this can't go into 3.9.0 (and maybe not at all).  Moving to draft.

henkdegroot:
> Ugh, Win11... That _could_ be it.

Already found some more "odd" behaviour on Win11 with Qt. Hoping that a move to Qt6 may help out.

PR #2731, opened 2022-07-14
Autobuild: Avoid building release only builds on PR

ann0see:

This PR adds some logic to avoid building ARM, JACK etc. builds except on releases. This is still in draft stage.

CHANGELOG: Internal: Avoid building release only builds on every Pull Request. This is done to save resources. 
**Context: Fixes an issue?**

https://github.com/jamulussoftware/jamulus/discussions/2282#discussioncomment-2046262

No.

Still in Draft. Needs to be tested. Maybe the approach doesn't work at all.

Testing, review.

## Checklist

-  [ ] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -129,6 +129,7 @@ jobs:
             building_on_os:         ubuntu-18.04
             base_command:           TARGET_ARCH=armhf ./.github/autobuild/linux_deb.sh
             run_codeql:             false
+            build_release_only:     true
 
           - config_name:            MacOS (artifacts)
             target_os:              macos
@@ -146,6 +147,7 @@ jobs:
             # Disable CodeQL on mac as it interferes with signing the binaries (signing hangs, see #2563 and #2564)
             run_codeql:             false
             xcode_version:          13.4.1
+            build_release_only:     true
 
           # Reminder: If Legacy is removed, be sure to add a dedicated job for CodeQL again.
           - config_name:            MacOS Legacy (artifacts+CodeQL)
@@ -182,8 +184,14 @@ jobs:
             building_on_os:         windows-2022
             base_command:           powershell .\.github\autobuild\windows.ps1 -BuildOption jackonwindows -Stage
             run_codeql:             false
+            build_release_only:     true
 
     runs-on:                        ${{ matrix.config.building_on_os }}
+    # Only build release only artifacts on release
+    if:                             >-
+                                    matrix.config.build_release_only == 'true' &&
+                                    needs.create_release.outputs.publish_to_release == 'true'
+    needs:                          create_release
     steps:
       - name:                       Select Xcode version for Mac
         if:                         matrix.config.target_os == 'macos' || matrix.config.target_os == 'ios'


ann0see:
Actually, I think needs some engineering in our prepare release step: https://stackoverflow.com/questions/65384420/how-to-make-a-github-action-matrix-element-conditional

pljones:
Can this please be updated to make it clearer what the goal is.

Currently it says "to save resources".  What resources?  Disk?  Time?  What is the problem being fixed?

ann0see:
Ah. Found the discussion: https://github.com/jamulussoftware/jamulus/discussions/2282#discussioncomment-2046262

pljones:
OK, it still needs breaking down into something more specific.

- For each trigger of an action
    - and for each job within an action
        - and for each run within a job
        we should know what the costs of that action are (time, cpu, disk) and what the benefits of running them are.

With that information, we can decide which jobs are the most costly and whether they are worth the benefit, based on the trigger.

Any build against master (either a PR or merge) _should_ be able to ("wouldn't it be nice if...") the output of the last merge to master as its starting workspace (and yes, that would mean keeping state per build target).  Behaviour could then be based on
- what `make` would say is dirty
- if it's not a merge and `make` wouldn't do anything, then don't run the packaging step either

ann0see:
For the both ARM builds the benefits are low:
Debian ARM is very similar to debian amd64, macOS ARM is similar to macOS intel. So the benefit is just convenience of being able to run Linux arm builds on Raspberry Pi on each PR. 

macOS Intel builds work on ARM and Intel macs, so the benefit of the arm build is low on PRs.

Cost is high especially for macOS arm since it uses another macOS runner which is not available for other builds. Linux is not that critical and a bit faster.

ann0see:
So you'd recommend to sync the .o files during runs?

pljones:
> So you'd recommend to sync the .o files during runs?

For each build target, yeah, if that were possible.  Timestamping might be an issue - make has to believe the make target is newer than what it's looking as a source for that target.

We'd also need to make sure we were only keeping _one_ set of files for each target - and being very, very certain about that - to avoid storage accruing...  It might also not be a saving: it _could_ take longer to save and then restore (persistent storage, possibly over a congested network, to container storage) each time than rebuild everything (container storage only).

ann0see:
Hmm. I mean we could do something similar to the caching approach for the dependencies. I agree that this would probably hugely speed up the build

hoffie:
I don't think we should invest into sharing object files between PR runs:
- Header file changes often trigger rebuilds.
- Sometimes (e.g. everything build-related) explicit cleanups are required.
- We can't share object files between platforms or architectures.
- We can't share object files between different build types (headless vs. normal) due to #ifdefs.

I think the benefit is low, the implementation complexity is high and I'm not convinced that one of the many edge cases leads to so many extra CI runs that there is no net benefit anymore.

As an alternative to this PR regarding the default build matrix selection, I'm still planning to submit my already prepared, but not finally tested logic changes as hinted in https://github.com/jamulussoftware/jamulus/pull/2534#issuecomment-1092263931.

hoffie:
@ann0see With #2789 merged, this PR could be closed I guess?

ann0see:
Yes.

Issue #2730, opened 2022-07-12
Regression: Gain changes not applied in all cases (e.g. new client level = 0)


hoffie:
> I started up a fresh instance of 3.9.0beta1 (compiled from src on Ubuntu 18.04) with no existing ini file, set a username, stereo out, new client level at 0%, connected to a server with players on it (Jamsuckers 212.83.56.10 on Any Genre 1, running version 3.8.2dev-1646e905 on Linux).
> 
> I expected no audio as new client was at 0% and indeed all the sliders were at 0 yet there was audio. I double checked to make sure I hadn't accidently looped a jack port mapping but there wasn't one. I further tested this behaviour against other servers with players on them, deleting the ini file and starting from scratch with the same settings each time. Server versions tested against were 3.8.1 and 3.8.2dev-c5111db6-dirty. All had the same client behaviour, there was audio even though all players sliders were at 0%.
I repeated above procedures using 1% and got the reduced volumes expected, which were quite lower in volume than what I heard when new client level was at 0%.
> 
> I tested same procedure with 3.8.2 (compiled from src on Ubuntu 18.04) against the same servers and got the expected behaviour of no audio when new client level was set to 0%.
> 
> Bug?

_Originally posted by @rdica in https://github.com/jamulussoftware/jamulus/discussions/2725_

I can confirm this behavior.

To be checked:
- [x] Is the actual value relevant (i.e. is 0 or 1 problematic or is this just the most extreme effect?)
- [x] Is new client level completely broken or does it only affect initial connect?
- [x] The sliders look correct, but the mix is wrong. It seems like the levels aren't properly sent to the server. Should confirm this via Wireshark.
- [x] Was the level change rate limiting merged post-3.8.2 and could it be related?

pljones:
Having looked at the code... it's a nightmare.

The big, big problem is this is implemented in the mixer panel, which is a GUI component.  It should be in the client, either directly in CClient or in a non-GUI class handling the mixer.

The latter seems the better approach -- currently CClient doesn't know very much about what appears in the main client GUI.  Effectively splitting mixer into behaviour/state and a UI controller tied to it with signals/slots would clean things up and make them more consistent.

Unfortunately, it's not a quick fix, of course.  For 3.9.0, given I've no idea what change broke it, I'm not going to propose this is done.

hoffie:
Reproducing instructions:
- Start a server
- Connect a client with permanent audio
[...]

PR #2737, opened 2022-07-17
Client: Avoid losing gain changes

hoffie:
PR #2535 introduced rate limiting for gain change messages. The logic required storing the previously used gain value per channel. This logic had some flaws:

1. The previously used gain value defaulted to 0, despite the server-side view of the channel being set to 1 (as the default). Therefore, gain(0) changes during a series of gain changes would be lost. The most common scenario would be the initial connection, which always triggers the rate limit and therefore the faulty logic. This also affected New Client Level = 0.
2. The previously used gain values were not reset upon changing servers. This might have caused losing arbitrary gain change messages, e.g. stored fader values.
3. The previously used gain values were not reset upon a channel disconnect. This might have caused missing fader level restores.

This PR introduces a gain level memory reset to 1 (100%) on connect as well as on channel disconnects to fix these issues.

CHANGELOG: Client: Fix potential long delay in sending fader changes to the server.

**Context: Fixes an issue?**

Fixes: #2730

No.

Ready for review and testing.

Needs more testing / confirmation.
@rdica Can you please try this PR?

**Helpful debug patches to confirm proper working**
```diff
diff --git a/src/client.cpp b/src/client.cpp
index ad9beb62..509dbaf7 100644
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -314,6 +314,11 @@ void CClient::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo )
             // reset oldGain and newGain as this channel id is currently unused and will
             // start with a server-side gain at 1 (100%) again.
             oldGain[iId] = newGain[iId] = 1;
+            qInfo() << "OnConClientListMesReceived: resetting old/newGain[" << iId << "] as it is unused";
+        }
+        else
+        {
+            qInfo() << "OnConClientListMesReceived: not resetting old/newGain[" << iId << "] as it is still in use";
         }
     }
 }
@@ -428,6 +433,8 @@ void CClient::SetRemoteChanGain ( const int iId, const float fGain, const bool b
     // send the actual gain and reset the range of channel IDs to empty
     oldGain[iId] = newGain[iId] = fGain;
     Channel.SetRemoteChanGain ( iId, fGain );
+    qInfo() << "SetRemoteChanGain: directly sending gain changed message for iId =" << iId << ", oldGain =" << oldGain[iId]
+            << ", newGain =" << newGain[iId];
 
     StartDelayTimer();
 }
@@ -445,6 +452,13 @@ void CClient::OnTimerRemoteChanGain()
             float fGain = oldGain[iId] = newGain[iId];
             Channel.SetRemoteChanGain ( iId, fGain );
             bSent = true;
+            qInfo() << "OnTimerRemoteChainGain: sending gain changed message for iId =" << iId << ", oldGain =" << oldGain[iId]
+                    << ", newGain =" << newGain[iId];
+        }
+        else
+        {
+            qInfo() << "OnTimerRemoteChainGain: skipping unchanged gain changed message for iId =" << iId << ", oldGain =" << oldGain[iId]
+                    << ", newGain =" << newGain[iId];
         }
     }
 

```

## Checklist

-  [x] I tested my code and it does what I want: This PR fixes the mentioned bug for me and the debug log seems ok.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -118,6 +118,8 @@ CClient::CClient ( const quint16  iPortNumber,
 
     QObject::connect ( &Channel, &CChannel::ReqChanInfo, this, &CClient::OnReqChanInfo );
 
+    // The first ConClientListMesReceived handler performs the necessary cleanup and has to run first:
+    QObject::connect ( &Channel, &CChannel::ConClientListMesReceived, this, &CClient::OnConClientListMesReceived );
     QObject::connect ( &Channel, &CChannel::ConClientListMesReceived, this, &CClient::ConClientListMesReceived );
 
     QObject::connect ( &Channel, &CChannel::Disconnected, this, &CClient::Disconnected );
@@ -261,6 +263,15 @@ void CClient::OnJittBufSizeChanged ( int iNewJitBufSize )
 
 void CClient::OnNewConnection()
 {
+    // The oldGain and newGain arrays are used to avoid sending duplicate gain change messages.
+    // As these values depend on the channel IDs of a specific server, we have
+    // to reset those upon connect.
+    // We reset to 1 because this is what the server part sets by default.
+    for ( int iId = 0; iId < MAX_NUM_CHANNELS; iId++ )
+    {
+        oldGain[iId] = newGain[iId] = 1;
+    }
+
     // a new connection was successfully initiated, send infos and request
     // connected clients list
     Channel.SetRemoteInfo ( ChannelInfo );
@@ -278,6 +289,33 @@ void CClient::OnNewConnection()
     // needed for compatibility to old servers >= 3.4.6 and <= 3.5.12
     Channel.CreateReqChannelLevelListMes();
     //### TODO: END ###//
+}
+
+void CClient::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo )
+{
+    // Upon receiving a new client list, we have to reset oldGain and newGain
+    // entries for unused channels. This ensures that a disconnected channel
+    // does not leave behind wrong cached gain values which would leak into
+    // any new channel which reused this channel id.
+    int iNumConnectedClients = vecChanInfo.Size();
+
+    // Save what channel IDs are in use:
+    bool bChanIdInUse[MAX_NUM_CHANNELS] = {};
+    for ( int i = 0; i < iNumConnectedClients; i++ )
+    {
+        bChanIdInUse[vecChanInfo[i].iChanID] = true;
+    }
+
+    // Reset all gains for unused channel IDs:
+    for ( int iId = 0; iId < MAX_NUM_CHANNELS; iId++ )
+    {
+        if ( !bChanIdInUse[iId] )
+        {
+            // reset oldGain and newGain as this channel id is currently unused and will
+            // start with a server-side gain at 1 (100%) again.
+            oldGain[iId] = newGain[iId] = 1;
+        }
+    }
 }
 
 void CClient::CreateServerJitterBufferMessage()
--- a/src/client.h
+++ b/src/client.h
@@ -403,6 +403,7 @@ protected slots:
     void OnControllerInFaderIsMute ( int iChannelIdx, bool bIsMute );
     void OnControllerInMuteMyself ( bool bMute );
     void OnClientIDReceived ( int iChanID );
+    void OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo );
 
 signals:
     void ConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo );


src/client.cpp
```diff
@@ -119,6 +119,7 @@ CClient::CClient ( const quint16  iPortNumber,
     QObject::connect ( &Channel, &CChannel::ReqChanInfo, this, &CClient::OnReqChanInfo );
 
     QObject::connect ( &Channel, &CChannel::ConClientListMesReceived, this, &CClient::ConClientListMesReceived );
+    QObject::connect ( &Channel, &CChannel::ConClientListMesReceived, this, &CClient::OnConClientListMesReceived );
```
hoffie, line 122: I'm a bit unsure what is better here:
- Use a single method to handle both cases (would require to re-emit the signal though, I think?)
- or use two `connect()` calls for those unrelated jobs.

I chose the latter as the jobs are unrelated (besides sharing the same trigger).

src/client.cpp
```diff
@@ -280,6 +290,36 @@ void CClient::OnNewConnection()
...
+    // does not leave behind wrong cached gain values which would leak into
+    // any new channel which reused this channel id.
+    int   iNumConnectedClients = vecChanInfo.Size();
+    bool  bChanIdInUse;
+    for ( int iId = 0; iId < MAX_NUM_CHANNELS; iId++ )
+    {
+        bChanIdInUse = false;
+        for ( int j = 0; j < iNumConnectedClients; j++ )
+        {
+            if ( vecChanInfo[j].iChanID == iId )
```
hoffie, line 306: I don't quite like the performance characteristics here. This is `O(MAX_NUM_CHANNELS²)` as worst-case (and runs on every channel connect/disconnect). I can't think of a more performant way though and the GUI code uses a similar construct, so I guess it's fine...?

https://github.com/jamulussoftware/jamulus/blob/fa204ad4f3025b7b2a1a8c1130ad998784faeb4a/src/audiomixerboard.cpp#L1213-L1217

ann0see: Basically for each channel you're searching where in vecChanInfo it is and if it's in use (=found) you do not reset it?
Maybe we could do something with sorting, but haven't thought it through

ann0see: Note: I do not have a full understanding of the code, but would something like that work (probably we'd need to take care of the array sizes)

boolean[] inUse = new boolean[MAX_NUM_CHANNELS];
// get all channels which are in use from vacChanInfo[i]
for (int i = 0;i<inUse.length;i++) {
  inUse[vecChanInfo[i].iChanID] = true;
}

// set all channels which are not in use to gain 1
for (int i = 0;i<OldGain.length;i++) {
  if (!inUse[i]) {
    oldGain[i] = newGain[i] = 1;
  }
}

rdica:
@hoffie Compiled `3.9.0beta2` on Ubuntu 18.04 with these changes.
Confirmed expected behaviour.

src/client.cpp
hoffie: That sounds good, thanks! We could probably even skip the initialization loop. That'd make it `O(2n)=O(n)` again, I guess. Will update the PR later.

ann0see: Yeah. Not sure what the default values of a boolean array in C++ are...

ann0see: I'd still like to do a performance test.

ann0see: Ok. The O(n) approach is around 100 (ms?), the O(n²) is fluctuating around 300+ (ms) by a quick test with < 40 clients (but the running time should stay roughly constant, at least for my approach), it remains to test that it works as expected ;-) - and might also be implemented at the other place since I think the O(n) code is cleaner anyway.

src/client.cpp
```diff
@@ -280,6 +290,36 @@ void CClient::OnNewConnection()
...
+            }
+        }
+        if ( !bChanIdInUse )
+        {
+            // reset oldGain and newGain as this channel id is currently unused and will
+            // start with a server-side gain at 1 (100%) again.
+            oldGain[iId] = newGain[iId] = 1;
+        }
+    }
+}
```
ann0see, line 321:
```suggestion
void CClient::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo )
{
    // Upon receiving a new client list, we have to reset oldGain and newGain
    // entries for unused channels. This ensures that a disconnected channel
    // does not leave behind wrong cached gain values which would leak into
    // any new channel which reused this channel id.

    int iNumConnectedClients = vecChanInfo.Size();

    bool inUse[MAX_NUM_CHANNELS];

    // get all channels which are in use from vacChanInfo[]
    for ( int i = 0; i < iNumConnectedClients; i++ )
    {
        inUse[vecChanInfo[i].iChanID] = true;
    }

    // set all channels which are not in use to gain 1
    for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
    {
        if ( !inUse[i] )
        {
            // reset oldGain and newGain as this channel id is currently unused and will
            // start with a server-side gain at 1 (100%) again.
            oldGain[i] = newGain[i] = 1;
        }
    }
}
```

ann0see: I think it solves the bug. However, I'd like an independent test of the code from above.

src/client.cpp
ann0see: Actually, the code you linked looks overly complicated. Probably worth refactoring...

hoffie: > Yeah. Not sure what the default values of a boolean array in C++ are...

Um, I was wrong -- seems like one must initialize explicitly: https://stackoverflow.com/questions/1920430/c-array-initialization/1920481#1920481

src/client.cpp
hoffie: Thanks. Based on your earlier suggestions I came up with almost the same. I've just pushed that with proper attribution.

ann0see:
@rdica could you please re-test with the new approach?

rdica:
@ann0see Do I clone your repo/branch for this?

ann0see:
No. Just use the updated code from hoffie. (Clone his repo again and ensure you have the code from this PR

rdica:
@ann0see @hoffie Either I did something incorrectly, or the new code didn't work. I fresh cloned hoffies's repo, bulit, joined an active server with new client level set to 0, and heard audio.

ann0see:
Did you check out the correct branch?

rdica:
I grabbed the zip from https://github.com/hoffie/jamulus/tree/fix-lost-gain-changes

ann0see:
Ok. So probably my approach is wrong. But could you please again make clean and test it (I remember it worked as I tested it with the code in the suggestion)

ann0see:
@hoffie maybe The initialisation is wrong https://m.cplusplus.com/forum/general/62832/

rdica:
Leave a comment
I recloned the above URL, ran `git checkout fix-lost-gain-changes`, rebuilt. Decided to check the program ver and its `3.8.2dev-d3022da1`. I'm confused.

ann0see:
d3022da1 That’s what the version of this PR should show if you are using git

rdica:
Ok well I thought the problem was introduced in `3.9.0beta1`...`3.8.2` already behaves correctly, not sure what I'm testing here. Any way, I executed the binary created from above and got the expected behaviour with new client level set at 0, and joining a populated server.

hoffie:
> Decided to check the program ver and its 3.8.2dev-d3022da1. I'm confused.

@rdica It's just that the non-release versioning is confusing. 3.8.2dev-XXX describes a version *after* 3.8.2 (contrary to what everyone else in the world seems to be doing; also against semantic versioning practices).

As far as I understand, a fresh binary from this PR works well for you, so that would be good.

@ann0see The code part which we changed after initially opening this PR is probably not relevant for the "joining a fresh server" case, so it's unlikely that those modifications broke the fix. I've just re-checked the debug output and it continues to seem correct.

softins (approved):
I don't have any means of trying this out for a couple of weeks, but I have studied the changes, and they look good to me. Thanks to @hoffie and @ann0see for the work!

rdica:
@hoffie @ann0see Thanks for the explanations and assistance with getting the code to test with.

src/client.cpp
pljones: Two handlers is better.

pljones: This, basically, is the fundamental flaw in the current design.

The Client is simply reemitting the signal on line 123 without "knowing" about it.  I think it would be a better approach for the Client to _handle_ all signals, hold state for what they "mean" and have GUI/JSONRPC/etc send state updates as signals.  The client could them emit _specific_ state change messages (i.e. in this case, which channel changed, one emit per channel, along with the new state).

However, for now, I'd be quite happy with either approach - either the new handler emitting `CClient::ConClientListMesReceived` when complete or the one taken here.

pljones:
> CHANGELOG: (List together with #2535)

Please don't do this.  It makes editing the ChangeLog really hard.  Put the title of the issue being fixed.  Duplicates are easy to spot and the grouping mechanism works.  With this approach the grouping mechanism doesn't always get the two anywhere near each other.

pljones (approved):

ann0see:
@hoffie I think we can merge this.

hoffie:
> Put the title of the issue being fixed.

I've now put the same title as the PR which initially caused the issue, so it should be right next to each other.

PR #2738, opened 2022-07-17
Improve algorithm to find in use channels in O(n) instead O(n²)

ann0see:

This replaces the O(n²) approach to find channels in use with a simpler O(n) approach.

Probably there are other places which would benefit from a more efficient approach too.

CHANGELOG: Client: Improve performance of GUI when someone join or leave a server.

**Context: Fixes an issue?**

Related to: https://github.com/jamulussoftware/jamulus/pull/2737#discussion_r922843286

No.

Close performance evaluation and (in depth) testing:
- [ ] Does the ordering get messed up? (it seems ok for me for a small number of clients)
- [ ] Are the faders hidden correctly at all times (didn't test in real world scenario yet)

Testing

## Checklist

-  [ ] I tested my code and it does what I want (at least I don't see any bad changes on joins/leaves. It must however be tested)

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1208,7 +1208,7 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
             UpdateTitle();
         }
 
-        // search for channels with are already present and preserve their gain
+        // search for channels which are already present and preserve their gain
         // setting, for all other channels reset gain
 
         // get all channels which are in use/not in use. We use the array index of vecChanInfo
@@ -1225,12 +1225,12 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
             iFaderNumber[vecChanInfo[iFader].iChanID] = iFader;
         }
 
+        // Hide all unused faders and initialize used ones
         for ( int iChanID = 0; iChanID < MAX_NUM_CHANNELS; iChanID++ )
         {
             if ( iFaderNumber[iChanID] == INVALID_INDEX )
             {
-                // fader is not used --> hide it
-                // before hiding the fader, store its level (if some conditions are fulfilled)
+                // current fader is not used
                 StoreFaderSettings ( vecpChanFader[iChanID] );
 
                 vecpChanFader[iChanID]->Hide();
@@ -1238,22 +1238,22 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
             }
 
             // current fader is used
-            // check if fader was already in use -> preserve gain value
             if ( !vecpChanFader[iChanID]->IsVisible() )
             {
-                // the fader was not in use, reset everything for new client
+                // the fader was not in use,
+                // reset everything for new client
                 vecpChanFader[iChanID]->Reset();
                 vecAvgLevels[iChanID] = 0.0f;
 
-                // check if this is my own fader and set fader property
                 if ( iChanID == iMyChannelID )
                 {
+                    // this is my own fader --> set fader property
                     vecpChanFader[iChanID]->SetIsMyOwnFader();
                 }
 
-                // set and increment the running new client counter needed for sorting (per definition:
-                // a fader for a new client shall always be inserted at the right-hand-side if no other
-                // sorting type is selected (i.e. "no sorting" is active) (#673)
+                // keep track of each new client
+                // for "no sorting" channel sort order new clients are added
+                // to the right-hand side of the mixer (#673)
                 vecpChanFader[iChanID]->SetRunningNewClientCnt ( iRunningNewClientCnt++ );
 
                 // show fader
@@ -1274,7 +1274,6 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
                 }
             }
 
-            // restore gain (if new name is different from the current one)
             if ( vecpChanFader[iChanID]->GetReceivedName().compare ( vecChanInfo[iFaderNumber[iChanID]].strName ) )
             {
                 // the text has actually changed, search in the list of
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1210,92 +1210,98 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
 
         // search for channels with are already present and preserve their gain
         // setting, for all other channels reset gain
-        for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
+
+        // get all channels which are in use/not in use. We use the array index of vecChanInfo
+        // if the fader is in use, else INVALID_INDEX to specify it is not in use
+        int iFaderNumber[MAX_NUM_CHANNELS];
+
+        for ( int iChanID = 0; iChanID < MAX_NUM_CHANNELS; iChanID++ )
         {
-            bool bFaderIsUsed = false;
+            iFaderNumber[iChanID] = INVALID_INDEX;
+        }
 
-            for ( int j = 0; j < iNumConnectedClients; j++ )
+        for ( int iFader = 0; iFader < iNumConnectedClients; iFader++ )
+        {
+            iFaderNumber[vecChanInfo[iFader].iChanID] = iFader;
+        }
+
+        for ( int iChanID = 0; iChanID < MAX_NUM_CHANNELS; iChanID++ )
+        {
+            if ( iFaderNumber[iChanID] == INVALID_INDEX )
             {
-                // check if current fader is used
-                if ( vecChanInfo[j].iChanID == i )
+                // fader is not used --> hide it
+                // before hiding the fader, store its level (if some conditions are fulfilled)
+                StoreFaderSettings ( vecpChanFader[iChanID] );
+
+                vecpChanFader[iChanID]->Hide();
+                continue;
+            }
+
+            // current fader is used
+            // check if fader was already in use -> preserve gain value
+            if ( !vecpChanFader[iChanID]->IsVisible() )
+            {
+                // the fader was not in use, reset everything for new client
+                vecpChanFader[iChanID]->Reset();
+                vecAvgLevels[iChanID] = 0.0f;
+
+                // check if this is my own fader and set fader property
+                if ( iChanID == iMyChannelID )
                 {
-                    // check if fader was already in use -> preserve gain value
-                    if ( !vecpChanFader[i]->IsVisible() )
-                    {
-                        // the fader was not in use, reset everything for new client
-                        vecpChanFader[i]->Reset();
-                        vecAvgLevels[i] = 0.0f;
+                    vecpChanFader[iChanID]->SetIsMyOwnFader();
+                }
 
-                        // check if this is my own fader and set fader property
-                        if ( i == iMyChannelID )
-                        {
-                            vecpChanFader[i]->SetIsMyOwnFader();
-                        }
+                // set and increment the running new client counter needed for sorting (per definition:
+                // a fader for a new client shall always be inserted at the right-hand-side if no other
+                // sorting type is selected (i.e. "no sorting" is active) (#673)
+                vecpChanFader[iChanID]->SetRunningNewClientCnt ( iRunningNewClientCnt++ );
 
-                        // set and increment the running new client counter needed for sorting (per definition:
-                        // a fader for a new client shall always be inserted at the right-hand-side if no other
-                        // sorting type is selected (i.e. "no sorting" is active) (#673)
-                        vecpChanFader[i]->SetRunningNewClientCnt ( iRunningNewClientCnt++ );
+                // show fader
+                vecpChanFader[iChanID]->Show();
 
-                        // show fader
-                        vecpChanFader[i]->Show();
-
-                        // Set the default initial fader level. Check first that
-                        // this is not the initialization (i.e. previously there
-                        // were no faders visible) to avoid that our own level is
-                        // adjusted. If we have received our own channel ID, then
-                        // we can adjust the level even if no fader was visible.
-                        // The fader level of 100 % is the default in the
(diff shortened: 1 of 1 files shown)


src/audiomixerboard.cpp
```diff
@@ -1210,87 +1210,93 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
...
-            bool bFaderIsUsed = false;
+            iFaderNumber[i] = INVALID_INDEX;
+        }
 
-            for ( int j = 0; j < iNumConnectedClients; j++ )
+        for ( int i = 0; i < iNumConnectedClients; i++ )
+        {
+            iFaderNumber[vecChanInfo[i].iChanID] = i;
+        }
+
```
ann0see, line 1227: I think we need a comment here.

src/audiomixerboard.cpp
```diff
@@ -1210,87 +1210,93 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
...
-            }
 
-            // if current fader is not used, hide it
-            if ( !bFaderIsUsed )
+                // set the channel infos
+                vecpChanFader[i]->SetChannelInfos ( vecChanInfo[iFaderNumber[i]] );
+            }
+            else
             {
+                // if current fader is not used, hide it
```
ann0see, line 1299: Seems a bit redundant here

ann0see:
Own channel first doesn't seem to work correctly at the moment: If we set the own channel checkbox, nothing happes at first.,

src/audiomixerboard.cpp
```diff
@@ -1210,87 +1210,93 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
...
 
-            for ( int j = 0; j < iNumConnectedClients; j++ )
+        for ( int i = 0; i < iNumConnectedClients; i++ )
+        {
+            iFaderNumber[vecChanInfo[i].iChanID] = i;
+        }
+
+        for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
+        {
+            if ( iFaderNumber[i] != INVALID_INDEX )
```
hoffie, line 1230: If we're refactoring the control flow anyway, I'd suggest moving the non-negated and shorter part first.

```suggestion
            if ( iFaderNumber[i] == INVALID_INDEX )
            {
                // fader unused
                continue;
            }
            // fader used, more logic
```

ann0see: There's an else to hide the fader. (See below).

But it can be swapped for sure

pljones:
Note that a lot of the code in audiomixerboard.cpp (GUI only) needs moving into a non-GUI class, so that state is not being maintained in the GUI.  Any large scale restructuring should take that into account as one of its (main) goals.

src/audiomixerboard.cpp
hoffie: Yes, sorry, I wasn't clear. I didn't mean to reduce it to the comments but thought of them as placeholders. :)

pljones:
> Own channel first doesn't seem to work correctly at the moment: If we set the own channel checkbox, nothing happens at first. You need to set a ordering first.
> 
> It's not reproducible; At least it only appeared on the first launch of the app. Also setting ini files doesn't seem to get the issue back.

Doesn't surprise me.  A lot of the code is impentrable.  Hence I'd rather a full refactor than trying to "fix in place".

src/audiomixerboard.cpp
ann0see: Ok. Done. Please re-check

hoffie:
> Note that a lot of the code in audiomixerboard.cpp (GUI only) needs moving into a non-GUI class, so that state is not being maintained in the GUI. Any large scale restructuring should take that into account as one of its (main) goals.

Yep. In the mean time, we should merge any iterative improvements such as this PR (once proven correct... ;)) neverthless, in my opinion.

src/audiomixerboard.cpp
```diff
@@ -1210,92 +1210,98 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
 
         // search for channels with are already present and preserve their gain
         // setting, for all other channels reset gain
+
+        // get all channels which are in use/not in use. We use the array index of vecChanInfo
+        // if the fader is in use, else INVALID_INDEX to specify it is not in use
+        int iFaderNumber[MAX_NUM_CHANNELS];
+
         for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
```
hoffie, line 1218:
```suggestion
        for ( int iChanID = 0; iChanID < MAX_NUM_CHANNELS; iChanID++ )
        ...
```

src/audiomixerboard.cpp
```diff
@@ -1210,92 +1210,98 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
...
+            if ( !vecpChanFader[i]->IsVisible() )
+            {
+                // the fader was not in use, reset everything for new client
+                vecpChanFader[i]->Reset();
+                vecAvgLevels[i] = 0.0f;
+
+                // check if this is my own fader and set fader property
+                if ( i == iMyChannelID )
+                {
+                    vecpChanFader[i]->SetIsMyOwnFader();
```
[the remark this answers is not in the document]
hoffie: I don't see it either. The refactoring looks correct to me.

Are you sure that this PR broke it? I've just tested with 3.8.2 and it seems to be broken there as well, unless I'm testing the wrong thing? :o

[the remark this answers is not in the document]
ann0see: I haven't tested anything before that, but I don't remember that own fader first didn't work.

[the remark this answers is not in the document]
hoffie: I thought I had reproduced the own-fader-first-breakage with this PR and then with a clean 3.8.2 build. Now I can no longer reproduce it either way. Everything works fine and I'm not sure what I'm doing differently now.

@ann0see Can you share how you made it break?

src/audiomixerboard.cpp
hoffie: Looks good, thanks.

src/audiomixerboard.cpp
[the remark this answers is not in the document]
ann0see: I think it's related to a fresh start without sorting and then enabling own fader first.

Not sure, but the following could lead to the issue (untested):
1. Connect to a crowded server without sorting
2. Wait until someone joins
3. Set own fader first
4. Observe you aren't first

ann0see:
Now I seem to miss a new join (fader doesn't show up) with the same (empty) name. This clearly needs more testing

pljones:
> Now I seem to miss a new join (fader doesn't show up) with the same (empty) name. This clearly needs more testing Actually also 3.8.2dev-fa204ad4 doesn't show the fader, so it seems to be unrelated to my change...

```
* |   48168a67 2022-07-18 Merge pull request #2737 from hoffie/fix-lost-gain-changes [Christian Hoffmann]
```
Could it be related to that?

ann0see:
Probably they're not related to the change in this PR at all (but still to be investigated.) Can you reproduce any of them on Linux?

ann0see:
Line 1200: why is this not in the mutex? Can we guarantee the number of clients doesn't change before we acquire the lock?

pljones:
Actually, reading your "how to reproduce", I don't think own fader first has ever worked.  I generally don't use sort and I turned off the own fader first feature once I realised it didn't work.

ann0see:
Actually I can’t reproduce it anymore. That’s the issue. Sometimes it works, sometimes it doesn’t. Maybe it’s worth looking at the initial PR again.

pljones:
> Actually I can’t reproduce it anymore. That’s the issue. Sometimes it works, sometimes it doesn’t. Maybe it’s worth looking at the initial PR again.

Yes, that's the same behaviour I've always had - I see it, then when I try to get a reliable way to reproduce it, I can't.

ann0see:
I just re-checked, at least one of both bugs also shows up on the current 3.8.2 release, so it's unrelated.

ann0see:
@pljones @hoffie I think this PR is ok, the bugs are not related to it.

src/audiomixerboard.cpp
pljones: Do we?  If so, before the review 😄.

src/audiomixerboard.cpp
```diff
@@ -1210,92 +1210,98 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
...
         {
-            bool bFaderIsUsed = false;
+            iFaderNumber[vecChanInfo[iFader].iChanID] = iFader;
+        }
+
+        for ( int iChanID = 0; iChanID < MAX_NUM_CHANNELS; iChanID++ )
+        {
+            if ( iFaderNumber[iChanID] == INVALID_INDEX )
+            {
+                // fader is not used --> hide it
```
pljones, line 1232: 1232/1233 could be rewritten to flow better.  "Hide it."  "Oh wait!! Before that, just one thing - store the levels.  Oh!! Only of certain conditions are met."

So first we store the levels (I don't think the "if some..." is needed - that's the concern of the called method - though maybe the method name needs improving at some point), then we hide the fader.  The comment should reflect that.

However..!  I don't even think the comment is adding to the code here.  "Store...; ...Hide()" is fairly self-explanatory (as the vector name is good).

src/audiomixerboard.cpp
```diff
@@ -1210,92 +1210,98 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
...
+                // before hiding the fader, store its level (if some conditions are fulfilled)
+                StoreFaderSettings ( vecpChanFader[iChanID] );
 
-            for ( int j = 0; j < iNumConnectedClients; j++ )
+                vecpChanFader[iChanID]->Hide();
+                continue;
+            }
+
+            // current fader is used
+            // check if fader was already in use -> preserve gain value
```
pljones, line 1241: This comment (line 1241) should move inside the `if`.  Then the "check if" can be removed.  Also ... "-> preserve gain value" appears to refer to the code at line 1277 - so really shouldn't be here: there's too much else happening first.

src/audiomixerboard.cpp
```diff
@@ -1210,92 +1210,98 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
...
+            if ( !vecpChanFader[iChanID]->IsVisible() )
             {
-                // check if current fader is used
-                if ( vecChanInfo[j].iChanID == i )
+                // the fader was not in use, reset everything for new client
+                vecpChanFader[iChanID]->Reset();
+                vecAvgLevels[iChanID] = 0.0f;
+
+                // check if this is my own fader and set fader property
```
pljones, line 1248: Don't say "check if" on an if statement.  That's what an if statement _is_.  Inside the if -- if the test doesn't make it clear -- explain the state.  Here, a comment of "// this is my own fader" would be present.  However, it's pretty redundant.

src/audiomixerboard.cpp
```diff
@@ -1210,92 +1210,98 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
...
+                if ( iChanID == iMyChannelID )
+                {
+                    vecpChanFader[iChanID]->SetIsMyOwnFader();
+                }
+
+                // set and increment the running new client counter needed for sorting (per definition:
```
pljones, line 1254: This comment doesn't really help...

Something as brief as
```
// keep track of each new client (for "no sorting" channel sort order, new clients are added to the right-hand side of the mixer)
```
might do.

src/audiomixerboard.cpp
```diff
@@ -1210,92 +1210,98 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
...
+                // a fader for a new client shall always be inserted at the right-hand-side if no other
+                // sorting type is selected (i.e. "no sorting" is active) (#673)
+                vecpChanFader[iChanID]->SetRunningNewClientCnt ( iRunningNewClientCnt++ );
+
+                // show fader
+                vecpChanFader[iChanID]->Show();
+
+                // Set the default initial fader level. Check first that
```
pljones, line 1262: I'm not going to ask for a change here as I can make no sense of either the comment or the code... 😄

src/audiomixerboard.cpp
```diff
@@ -1210,92 +1210,98 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
...
-
-                    bFaderIsUsed = true;
+                    // the value is in percent -> convert range
+                    vecpChanFader[iChanID]->SetFaderLevel ( pSettings->iNewClientFaderLevel / 100.0 * AUD_MIX_FADER_MAX );
                 }
             }
 
-            // if current fader is not used, hide it
-            if ( !bFaderIsUsed )
+            // restore gain (if new name is different from the current one)
```
pljones, line 1277: This, again, is a duplicate comment -- the comment should be inside the if describing the state there.  There's already a comment which does that in the right place.

pljones:
Sorry, just went through in "picky" mode.  As there are a lot of changes happening, it might be nice to make the comments clear at the same time.

ann0see:
> Sorry, just went through in "picky" mode. As there are a lot of changes happening, it might be nice to make the comments clear at the same time.

No problem. I just added another commit.

src/audiomixerboard.cpp
```diff
@@ -1210,92 +1210,97 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
 
         // search for channels with are already present and preserve their gain
```
pljones, line 1211: ^with^which

pljones (approved):

pljones:
Approved but if you want to fix the existing typo, please do.

hoffie (approved):

Issue #2775, opened 2022-08-09
CI: Improve Windows dependency caching


hoffie:
**What is the current behaviour and why should it be changed?**

Some of the Windows dependency downloads (I'm suspecting SourceForge) increasingly fail.

```
Get-RedirectedUrl : Exception calling "GetResponse" with "0" argument(s): "The request was aborted: Could not create 
SSL/TLS secure channel."
```

https://github.com/jamulussoftware/jamulus/runs/7742832728?check_suite_focus=true#step:9:1

**Describe possible approaches**
~~More~~ Proper binary caching would reduce the impact.

**Has this feature been discussed and generally agreed?**

The increased build failures have been discussed via chat recently.

PR #2780, opened 2022-08-12
Build: Improve Windows dependency download resilience

hoffie:
- CI: Fix caching of NSIS & ASIO installations
  Previously, we were caching the wrong directories for NSIS/ASIO. Therefore, the cache never hit and wasn't effective against download problems.

- Build: Windows: Log when re-using NSIS/ASIO directories
  This makes it easier to spot problems in the caching logic.

- Build: Retry SourceForge (NSIS) downloads for improved build stability
  Previously, a single download failure made the whole build fail. Now we try three times in a row after a slight delay.

- Build: windows/deploy_windows: Make spacing/case more consistent
  ... while we are at it.

CHANGELOG: SKIP

**Context: Fixes an issue?**
Fixes: #2775

No.

Ready.

Reviews.

## Checklist

Test results:
- [x] Cache size increased from [179MB (master)](https://github.com/jamulussoftware/jamulus/runs/7755249088?check_suite_focus=true#step:5:12) to [186MB (a stripped down version of this PR)](https://github.com/hoffie/jamulus/runs/7808199877?check_suite_focus=true#step:5:12)
- [x] The cache hit is [logged as expected now](https://github.com/hoffie/jamulus/runs/7808199877?check_suite_focus=true#step:9:22)

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -213,8 +213,8 @@ jobs:
           path: |
             C:\Qt
             C:\ChocoCache
-            ~\windows\NSIS
-            ~\libs\ASIOSDK2
+            ${{ github.workspace }}\windows\NSIS
+            ${{ github.workspace }}\libs\ASIOSDK2
           key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/workflows/autobuild.yml', '.github/autobuild/windows.ps1', 'windows/deploy_windows.ps1') }}-${{ matrix.config.base_command }}
 
       - name:                       Cache Android dependencies
--- a/windows/deploy_windows.ps1
+++ b/windows/deploy_windows.ps1
@@ -61,11 +61,25 @@ Function Get-RedirectedUrl {
         [String]$URL
     )
 
-    $request = [System.Net.WebRequest]::Create($url)
-    $request.AllowAutoRedirect=$true
-    $response=$request.GetResponse()
-    $response.ResponseUri.AbsoluteUri
-    $response.Close()
+    foreach ($attempt in 1, 2, 3) {
+        try {
+            $request = [System.Net.WebRequest]::Create($url)
+            $request.AllowAutoRedirect=$true
+            $response=$request.GetResponse()
+            $response.ResponseUri.AbsoluteUri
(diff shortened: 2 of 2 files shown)


windows/deploy_windows.ps1
```diff
@@ -55,17 +55,30 @@ Function Clean-Build-Environment
...
-        [String]$URL
+        [string] $url
     )
 
-    $request = [System.Net.WebRequest]::Create($url)
-    $request.AllowAutoRedirect=$true
-    $response=$request.GetResponse()
-    $response.ResponseUri.AbsoluteUri
-    $response.Close()
+    foreach ($attempt in 1, 2, 3) {
```
ann0see, line 63: That’s a strange syntax. Is there any reason why that’s not a standard for loop (doesn’t that also exist in PS?)

.github/workflows/autobuild.yml
```diff
@@ -213,8 +213,8 @@ jobs:
           path: |
             C:\Qt
             C:\ChocoCache
-            ~\windows\NSIS
-            ~\libs\ASIOSDK2
+            ${{ github.workspace }}\windows\NSIS
```
ann0see, line 216: I probably didn’t take enough care about windows as I moved the SDK. Thanks for fixing

ann0see (approved):

windows/deploy_windows.ps1
hoffie: Probably depends where one comes from. Python is foreach-centric... ;)
The alternative would be:

```suggestion
    for ($attempt = 1; $attempt -le 3; $attempt++) {
```

Not sure if it's better though.

.github/workflows/autobuild.yml
hoffie: I'm not even sure what broke it, maybe it had never worked correctly, although I vaguely remember putting lots of effort into the testing. Anyway, now it should be good.

windows/deploy_windows.ps1
ann0see: I think it’s more standard, but it seems to be more verbose. 
I‘d go with the foreach loop then.

Issue #2800, opened 2022-08-22
Build: Debian -dev builds (non-release) show version as -dirty


hoffie:
**Describe the bug**

`jamulus-headless -v` outputs `-dirty`

**To Reproduce**

**Expected behavior**

Should not be `-dirty`.
Likely reason: The moving around of debian control files during build. Maybe symlinks would be an alternative.

**Operating system**

Debian oldstable

**Version of Jamulus**

master

**Additional context**

ann0see:
The release?

hoffie:
No, I think it only affects intermediate, non-release versions.

pljones:
`-dirty` is _correct_ if a build process doesn't use the binary from an unchanged source tree.  If debian is doing something to change the source tree _before generating the binary_, that's a bug in the build process.

(That is, I don't know how the `-dirty` gets applied: but if it only happens when using the debian flow, it's the debian flow that's wrong.  It doesn't happen with `make distclean; qmake && make -j3 && make clean`.)

PR #2802, opened 2022-08-22
Build: Debian/Ubuntu: Avoid incorrect -dirty version suffix

hoffie:
The debian build scripts will update config.{guess,sub} and will therefore cause the Jamulus.pro dev version detection to mark the build as dirty. We circumvent that by overriding the responsible debian helper call.

CHANGELOG: Debian/Ubuntu: Fixed displayed version for non-release builds to remove incorrect -dirty suffixes.

**Context: Fixes an issue?**

Fixes: #2800

No.

Ready.

Reviews.

## Checklist

### DIFF
--- a/distributions/debian/rules
+++ b/distributions/debian/rules
@@ -14,6 +14,14 @@ endif
 %:
 	dh $@
 
+override_dh_update_autotools_config:
+	# dh_update_autotools_config replaces libs/opus/config.{sub,guess}.
+	# This is unnecessary as we don't build opus via autotools at all
+	# (we use qmake). In addition, this would cause our -dev version generation
+	# logic to mark Debian builds as -dirty by default.
+	# Therefore, disable this behavior:
+	:
+
 override_dh_auto_configure:
 	mkdir -p build-gui && cd build-gui && $(QMAKE) "CONFIG+=noupcasename" PREFIX=/usr ../Jamulus.pro
 	mkdir -p build-nox && cd build-nox && $(QMAKE) "CONFIG+=headless serveronly" TARGET=jamulus-headless PREFIX=/usr ../Jamulus.pro


ann0see:
Did that happen starting from the introduction of the flag?

.github/autobuild/linux_deb.sh
```diff
@@ -29,6 +29,13 @@ setup() {
     sudo apt-get -qq update
     sudo apt-get -qq --no-install-recommends -y install devscripts build-essential debhelper fakeroot libjack-jackd2-dev qtbase5-dev qttools5-dev-tools qtmultimedia5-dev
 
+    # The debian build scripts will update config.{guess,sub} and will therefore cause the Jamulus.pro
```
pljones, line 32: Why does it do this?

And if they do need to do this _before_ building the binary, tough luck - it's dirty.

hoffie: > Why does it do this?

It's an optimization. Distributions commonly run autoreconf themselves or substitute those files with pre-generated content. I've found a different, better approach though and have updated this PR. I found a way to disable this debian-specific logic and this looks way cleaner.

hoffie:
> Did that happen starting from the introduction of the flag?

I'd guess so, but it's hard to tell without running new builds with the old code again. Not sure if it's worth it.
37f993aa7 first introduced the `-dirty` logic via #475 and #883.

hoffie:
PR logs look fine: https://github.com/jamulussoftware/jamulus/runs/7982212320?check_suite_focus=true#step:11:80
`Project MESSAGE: building version "3.9.0dev-47f507c" (intermediate in git repository)`

pljones (approved):

distributions/debian/rules
```diff
@@ -14,6 +14,14 @@ endif
 %:
 	dh $@
 
+override_dh_update_autotools_config:
+	# dh_update_autotools_config replaces libs/opus/config.{sub,guess}.
+	# This is unnecessary as we don't build opus via autotools at all
+	# (we use qmake). In addition, this causes our -dev version generation
+	# logic to mark Debian builds as -dirty by default.
```
ann0see, line 21: I'd add a link to this PR here

hoffie: Hm, why do that here? We don't add Github Issue/PR links for other code parts either. If you think that something is missing in the comment, I'm happy to add it. If someone wants to track that back to Github, this should be possible using `git blame` and `git log`.

ann0see: You describe an issue which is fixed. "In addition ... by default." Therefore I thought it would make sense to linke this PR for further information. 

You could also change the grammar: ... this causes --> This would have caused...

hoffie: Ok, I've added the *would* part in the hope that it makes it clearer. The point of the comment is to state that there are two reasons:
- First, the config.guess replacement behavior is unnecessary, as the file is not used at all.
- Second (therefore "in addition"), it causes the `-dirty` behavior.

ann0see (approved):
Thanks, I'm happy now

Issue #2791, opened 2022-08-20
Mac: Sign non-release builds with self-signed key


hoffie:
**What is the current behaviour and why should it be changed?**

Currently, only release builds are signed via @emlynmac. Now that we provide M1 builds (#2357), it would be interesting to sign non-release builds as well, as M1 Macs refuse to run unsigned binaries by default.

**Describe possible approaches**

It should be sufficient to use a self-signed key for that.

- [x] ~Key generation~ (turned out to be unneeded)
- [x] Build logic changes

**Has this feature been discussed and generally agreed?**

See #2357

ann0see:
I can create a cert. However, should we share it internally?

pljones:
Should it go into jamulussoftware/assets or jamulussoftware/settings/secrets/actions?

ann0see:
Since it's a cryptographic key, it needs to go to the secrets tab, not a repo.

hoffie:
> However, should we share it internally?

I think it should be ok to store it as a secret only. It will not be easily recoverable there though (but we shouldn't need it).
In addition to the key, we also need to adjust the build logic to avoid trying the online (apple.com) parts of the signing process. I guess it makes sense to do this via another boolean secret to make that decision (or derive it from the cert if possible?).

emlynmac:
Maybe have a look at the existing signing system and add the relevant gitHub secrets to enable signing.
It may require some further changes to make sure the notarization path is not followed, but should be simpler to do that way.

ann0see:
I think I followed https://www.simplified.guide/macos/keychain-cert-code-signing-create once.

ann0see:
I think I've now added everything unrelated to notarization from my side on this repo. So we'd need to skip MACOS_CERTIFICATE_ID and NOTARIZATION_PASSWORD

ann0see:
Ok. I've now added a fake password and for the MACOS_CERTIFICATE_ID the name of the certificate, but on a workflow_dispatch run nothing changes.

hoffie:
I think we need several more changes (which could probably be done on a fork first):
- https://github.com/jamulussoftware/jamulus/blob/e14f3dd9af84271c22575cf4ce7b10dbac0c2bdd/.github/autobuild/mac.sh#L44 needs dropping, I think
[...]

Issue #2790, opened 2022-08-20
Mac: Provide universal binaries for x86_64 & arm64e


hoffie:
**What is the current behaviour and why should it be changed?**

#2357 introduced native M1 releases. It would be more user-friendly if we packaged both x86_64 and arm64e builds into a single universal binary. This is possible using `lipo`, but the details have to be figured out.

https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary

**Describe possible approaches**

A first attempt of putting both archs into `QMAKE_APPLE_DEVICE_ARCHS` has been done in #2357, but it failed due to arch-specific code paths.

If we can't get that to work, it will either need moving both build runs into a single CI job or will require another CI job which takes the artifacts from the previous steps and combines them. The former sounds more realistical, especially considerung the .dmg generation.

**Has this feature been discussed and generally agreed?**

See #2357

ann0see:
> A first attempt via QMAKE_APPLE_DEVICE_ARCHS has been done in https://github.com/jamulussoftware/jamulus/issues/2357, but it failed due to arch-specific code paths.

So is opus the problem? We should understand if it's a Qt bug or opus doesn't support the config option.

hoffie:
opus doesn't know about `QMAKE_*` variables. The issue is that we sneak in the bundled opus compilation via our `qmake` build system. Part of this is the decision to include platform-specific performance improvements for opus. The issue is that this detection works once and decides to include certain things (i.e. the x86_64 opus optimizations). This makes the compiler fail when doing the arm compilation, as those headers are not supported there.

For testing if this path is usable at all, it might make sense to drop the opus performance improvements. If this works, we could look at finding a way to include those properly for such multi-arch builds. I don't think `qmake` has any features for that ("arm-only source files"), but we might be able to come up with something.

hoffie:
After having a closer look, I don't see a sane way of passing certain flags/defines for specific archs only (well, clang supports `-Xarch_foo`, but... we would have to sneak that in via `qmake`) or specific source or header files for specific archs only.

I suspect the better way forward would be running two standard builds. This could be done in a single CI job which would save all the resources for preparation steps.
I've manually (via SSH) done the following and the resulting binary looks correct:
- `/usr/local/opt/qt/6.3.1/macos/bin/qmake QMAKE_APPLE_DEVICE_ARCHS=x86_64 QT_ARCH=x86_64 TARGET=jamulus.x86_64`
[...]

PR #2808, opened 2022-08-25
Build: Mac: Add universal binary and -dev build ad-hoc signing support

hoffie:
- Build: Mac: Add universal binary support

  This PR combines the x86_64 and arm64e CI jobs into a single job in order to have both binary results at hand. It then uses `lipo` to combine those two binaries in a single universal binary which runs on both architectures.
  This increases user-friendliness as users no longer have to worry about downloading the proper build for their architecture.

- Build: Mac: Use ad-hoc signing when no developer certificate is available

  Previously, only (pre-)release builds were signed. This signing and notarization was only possible with a proper developer certificate. As the Jamulus project currently does not have one and has to rely on other developers doing the signing, we only go this route for (pre-)releases due to the manual effort required.
  With Mac M1 support, having a valid signature becomes more important as unsigned binaries can only be started after turning several knobs.

  This PR adds very basic ad-hoc signing which does not require any Apple account/certificate, but improves user experience for M1 users.

CHANGELOG: Build: Mac: Combined Intel & M1 builds into a single Universal binary and improved M1 -dev build user-friendliness by introducing ad-hoc signing support

**Context: Fixes an issue?**
Fixes: #2790
Fixes: #2791

- [x] Release announcement? Added reminder to #2813

Ready for review and testing.

- [x] M1/arm64 tests, see https://github.com/jamulussoftware/jamulus/issues/2790#issuecomment-1225618827
- [x] x86_64 tests (@ann0see?)

## Checklist

### DIFF
--- a/mac/deploy_mac.sh
+++ b/mac/deploy_mac.sh
@@ -36,19 +36,37 @@ cleanup() {
 
 build_app() {
     local client_or_server="${1}"
-
-    # Build Jamulus
-    declare -a BUILD_ARGS=("_UNUSED_DUMMY=''")  # old bash fails otherwise
-    if [[ "${TARGET_ARCH:-}" ]]; then
-        BUILD_ARGS=("QMAKE_APPLE_DEVICE_ARCHS=${TARGET_ARCH}" "QT_ARCH=${TARGET_ARCH}")
-    fi
-    qmake "${project_path}" -o "${build_path}/Makefile" "CONFIG+=release" "${BUILD_ARGS[@]}" "${@:2}"
-    local target_name
-    target_name=$(sed -nE 's/^QMAKE_TARGET *= *(.*)$/\1/p' "${build_path}/Makefile")
     local job_count
     job_count=$(sysctl -n hw.ncpu)
 
-    make -f "${build_path}/Makefile" -C "${build_path}" -j "${job_count}"
+    # Build Jamulus for all requested architectures, defaulting to x86_64 if none provided:
+    local target_name
+    local target_arch
+    local target_archs_array
+    IFS=' ' read -ra target_archs_array <<< "${TARGET_ARCHS:-x86_64}"
+    for target_arch in "${target_archs_array[@]}"; do
+        if [[ "${target_arch}" != "${target_archs_array[0]}" ]]; then
+            # This is the second (or a later) first pass of a multi-architecture build.
+            # We need to prune all leftovers from the previous pass here in order to force re-compilation now.
+            make -f "${build_path}/Makefile" -C "${build_path}" distclean
+        fi
+        qmake "${project_path}" -o "${build_path}/Makefile" \
+            "CONFIG+=release" \
+            "QMAKE_APPLE_DEVICE_ARCHS=${target_arch}" "QT_ARCH=${target_arch}" \
+            "${@:2}"
+        make -f "${build_path}/Makefile" -C "${build_path}" -j "${job_count}"
+        target_name=$(sed -nE 's/^QMAKE_TARGET *= *(.*)$/\1/p' "${build_path}/Makefile")
+        if [[ ${#target_archs_array[@]} -gt 1 ]]; then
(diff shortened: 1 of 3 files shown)


mac/deploy_mac.sh
```diff
@@ -36,23 +36,40 @@ cleanup() {
...
+        make -f "${build_path}/Makefile" -C "${build_path}" -j "${job_count}"
+        target_name=$(sed -nE 's/^QMAKE_TARGET *= *(.*)$/\1/p' "${build_path}/Makefile")
+        if [[ ${#target_archs_array[@]} -gt 1 ]]; then
+            # When building for multiple architectures, move the binary to a safe place to avoid overwriting by the other passes.
+            mv "${build_path}/${target_name}.app/Contents/MacOS/${target_name}" "${build_path}/${target_name}.app/Contents/MacOS/${target_name}.arch_${target_arch}"
+        fi
+        # We need to prune all object files here in order to force re-compilation in the next pass which builds
+        # for another architecture.
+        # We need to keep the ${build_path}/ contents though as we will use those artifacts later.
+        make -f "${build_path}/Makefile" -C "${build_path}" clean
```
pljones, line 61: I'd do `make distclean` - it's more vicious.

ann0see:
Intel Mac shows the App as "Universal" and works as expected.

mac/deploy_mac.sh
hoffie: I'll try, I was worried about it being extra-correct and removing intermediate artifacts, but maybe it could work. I'll give it a try when I have time.

ann0see (approved):
Approving based on my tests. @pljones comments should be addressed before a merge.

mac/deploy_mac.sh
hoffie: Ok, as I feared: Simply running `make distclean` here fails, because it also removes the artifact which we want to keep:
https://github.com/jamulussoftware/jamulus/runs/8058626894?check_suite_focus=true#step:10:1464
`fatal error: /Applications/Xcode_13.4.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: can't open input file: /Users/runner/work/jamulus/jamulus/build/Jamulus.app/Contents/MacOS/Jamulus.arch_* (No such file or directory)`

hoffie: > If you need to keep anything, move it somewhere safe like on line 56.

I've now more or less implemented that -- I'm running `make distclean` right before the second (or later) pass' `qmake` (running it after `qmake` would possibly be incorrect, as `qmake` would have generated new cleanup instructions based on the next architecture).

This seems to work properly now:
https://github.com/jamulussoftware/jamulus/runs/8058721581?check_suite_focus=true#step:10:753

@pljones Are you fine with that?

pljones (approved):

mac/deploy_mac.sh
pljones: Yes, my standard build is
```
make distclean; qmake {optional params} && make j3 && make clean
```
except during a dev cycle, where I just do `make` a lot :).  We don't need the `make clean` as the container will get dropped.  And the way you've done it, the `make distclean` should always find a `Makefile` to work from.

Nice.

ann0see:
@hoffie it seems as if the build created on master is not a universal build. It shows up as arm only, while the one I tested is universal.

ann0see:
Maybe the distclean thing broke it?

hoffie:
Meh, seems so. It's a universal build with a single architecture:
https://github.com/jamulussoftware/jamulus/runs/8071146815?check_suite_focus=true#step:10:1455

Will work on a fix shortly.

ann0see:
Great. Thanks jamulus_3.9.0dev-dead5c5_mac works as expected, that's the one I tested last.

pljones:
https://github.com/jamulussoftware/jamuluswebsite/pull/806 merged, so removing "Needs documentation".

PR #2812, opened 2022-08-28
Server: Correct handling of server list entry zero

pljones:

@corrados reported an issue with a change made for Jamulus 3.9.0 which changed the behaviour for this entry.

Previous, omitting `--serverpublicip` for a directory had resulted in the server list entry for the directory being determined correctly.

With the changes in 3.9.0, the wrong address could be determined, preventing use of the directory as a normal server.

For example, running jamulusdirectory8.drealm.info:22824 as follows (more or less):
`/opt/Jamulus/bin/Jamulus-Directory8 -n -s -p 22824 -u 1 -l /opt/Jamulus/log/Directory8.log -e localhost -f [3.8.0] --directoryfile /opt/Jamulus/run/Directory8.persistence -o Directory on 22824;London;224 -w <h3>Directory on 22824</h3><p>Server list hosted by drealm.info.</p>`
would result in
image
where `192.168.1.19` is the server address, not the public IP address, and is hence not reachable by Jamulus Explorer as a client.

3.9.0 made quite a bit of change to the server list handling and I'm hoping this small adjustment is all that's needed to restore the previous behaviour.  Server list entry zero is "self".  When a directory gets asked to return the server list, that means its own entry always comes first.  The change here returns a "default" (protocol determined) value for the host address that the client will use, regardless of whether it's LAN or internet.

CHANGELOG: Correct default "self" address for directory, enabling use as a server

**Context:**

Raised in email.

No.

jamulusdirectory8.drealm.info:22824 is running this patched version of the software.  private.drealm.info:56850 is registered with the directory (to prove registration still works).  Using Jamulus Explorer with the directory added, the display is correct.
image
Jamulus Explorer can see the welcome message and a LAN client can also connect to the directory as a server.

Review

## Checklist

### DIFF
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -181,10 +181,14 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
         iServInfoNumSplitItems = slServInfoSeparateParams.count();
     }
 
-    // Init server list entry (server info for this server) with defaults. Per
-    // definition the client substitutes the IP address of the directory server
-    // itself for his server list. If we are a directory server, we assume that
-    // we are a permanent server.
+    /*
+     * Init server list entry (server info for this server) with defaults.
+     *
+     * The client will use the built in or custom address when retrieving the server list from a directory.
+     * The values supplied here only apply when using the server as a server, not as a directory.
+     *
+     * If we are a directory, we assume that we are a permanent server.
+     */
     CServerListEntry ThisServerListEntry ( haServerAddr, ServerPublicIP, "", QLocale::system().country(), "", iNumChannels, bIsDirectoryServer );
 
     // parse the server info string according to definition:
@@ -631,6 +635,10 @@ void CServerListManager::Remove ( const CHostAddress& InetAddr )
  - SERVER external to DIRECTORY (list entry external IP same LAN)
    - CLIENT internal to SERVER (CLM     same IP as list entry external IP): use "internal" fields (local LAN address)
    - CLIENT external to SERVER (CLM not same IP as list entry external IP): use "external" fields (public IP address from protocol)
+
+ Finally, when retrieving the entry for the directory itself life is more complicated still.  It's easiest just to return "0"
+ and allow the client connect dialogue instead to use the IP and Port from which the list was received.
+
  */
 void CServerListManager::RetrieveAll ( const CHostAddress& InetAddr )
 {
@@ -646,10 +654,13 @@ void CServerListManager::RetrieveAll ( const CHostAddress& InetAddr )
         // allocate memory for the entire list
         CVector<CServerInfo> vecServerInfo ( iCurServerListSize );
 
-        // copy the list (we have to copy it since the message requires
-        // a vector but the list is actually stored in a QList object and
-        // not in a vector object
-        for ( int iIdx = 0; iIdx < iCurServerListSize; iIdx++ )
+        // copy list item for the directory and just let the protocol sort out the actual details
+        vecServerInfo[0]          = ServerList[0];
+        vecServerInfo[0].HostAddr = CHostAddress();
+
+        // copy the list (we have to copy it since the message requires a vector but the list is actually stored in a QList object
+        // and not in a vector object)
+        for ( int iIdx = 1; iIdx < iCurServerListSize; iIdx++ )
         {
             // copy list item
             CServerInfo& siCurListEntry = vecServerInfo[iIdx] = ServerList[iIdx];
@@ -657,7 +668,7 @@ void CServerListManager::RetrieveAll ( const CHostAddress& InetAddr )
             bool serverIsInternal = NetworkUtil::IsPrivateNetworkIP ( siCurListEntry.HostAddr.InetAddr );
 
             bool wantHostAddr = clientIsInternal /* HostAddr is local IP if local server else external IP, so do not replace */ ||
-                                ( !serverIsInternal && iIdx != 0 &&
+                                ( !serverIsInternal &&
                                   InetAddr.InetAddr != siCurListEntry.HostAddr.InetAddr /* external server and client have different public IPs */ );
 
             if ( !wantHostAddr )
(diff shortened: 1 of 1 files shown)


src/serverlist.cpp
```diff
@@ -631,6 +635,10 @@ void CServerListManager::Remove ( const CHostAddress& InetAddr )
  - SERVER external to DIRECTORY (list entry external IP same LAN)
    - CLIENT internal to SERVER (CLM     same IP as list entry external IP): use "internal" fields (local LAN address)
    - CLIENT external to SERVER (CLM not same IP as list entry external IP): use "external" fields (public IP address from protocol)
+
+ Finally, when retrieving the entry for the directory itself life is more complicated still.  It's easiest just to return "0"
+ and hope the IP and UDP fill in the gaps.
```
hoffie, line 640: That sounds a bit unfounded. Do we have better facts than hoping? ;)

hoffie (approved):
Sounds good to me, besides the comment which might benefit from a little bit more specifics.

src/serverlist.cpp
pljones: :)  Well, it side-step putting an actual definite value into the field.  It puts zero in and lets IP/UDP decide, _I think_.  Either that, or I'm missing something in `src/protocol.cpp`.

hoffie: Hm, I don't think IP/UDP can or should decide that. I think the protocol layer could decide, but [I don't see special casing for zero HostAddress fields](https://github.com/jamulussoftware/jamulus/blob/05f385c8019269369fb7fd5f0f6b038dd49ab84c/src/protocol.cpp#L2036-L2042). The [receiving side does not show special-casing either](https://github.com/jamulussoftware/jamulus/blob/master/src/protocol.cpp#L2080-L2081).

It seems like it is once again the **GUI** which contains relevant logic (and shouldn't...):
https://github.com/jamulussoftware/jamulus/blob/05f385c8019269369fb7fd5f0f6b038dd49ab84c/src/connectdlg.cpp#L346-L354

After reading the logic, that the client always special-cases index 0, I'm no longer sure why the server-set address is relevant at all.

Did someone confirm that it affects Jamulus?
Could it be that it's "just" Jamulus Explorer which takes what the directory sends?

(I'm not saying we shouldn't fix it, as it is obviously an unintended change of the previous release; but it would be less critical if it didn't affect vanilla Jamulus usage)

hoffie: Jamulus Explorer works different: It does not special-case index 0, instead [it lets 0 IP addresses default to the directory address](https://github.com/softins/jamulus-php/blob/d3b295a594ad35a3891033db31fcd2e77dbde905/servers.php#L753-L756). Therefore, this is consistent with how 3.9.0 behaves and should be remedied with this PR.

ann0see:
I think the commit message should be similar to the PR title as the current one needs more background knowledge and doesn't add that much IMHO.

ann0see (commented):
I'd like someone else to review this PR, either Volker or Tony. 

It looks a bit like a workaround for an issue caused elsewhere.

src/serverlist.cpp
pljones: Hm...  I don't like it that it's in `src/connectdlg.cpp` -- _no_ decision making processing should be done in the dialogue layer, only display/formatting processing.  This should either be in the protocol layer (when receiving the list) or in the client layer (when returning the retrieved list).

Admittedly, the server list is only used by the connect dialogue...

pljones:
> It looks a bit like a workaround for an issue caused elsewhere.

It was a workaround for an issue caused elsewhere before.  I took it away in 3.9.0.  This restores the workaround.

src/serverlist.cpp
softins, line 640:
```suggestion
 and allow the client instead to use the IP and Port from which the list was received.
```

softins (approved):
I haven't tried the code, but the change looks like it should do what's intended, and I see Volker has validated its operation in practice.
So happy to approve.

src/serverlist.cpp
pljones: Done... more or less.

src/serverlist.cpp
```diff
@@ -631,6 +635,10 @@ void CServerListManager::Remove ( const CHostAddress& InetAddr )
  - SERVER external to DIRECTORY (list entry external IP same LAN)
    - CLIENT internal to SERVER (CLM     same IP as list entry external IP): use "internal" fields (local LAN address)
    - CLIENT external to SERVER (CLM not same IP as list entry external IP): use "external" fields (public IP address from protocol)
+
+ Finally, when retrieving the entry for the directory itself life is more complicated still.  It's easiest just to return "0"
+ and allow the client connect dialogue instead to use the IP and Port from which the list was received.
```
hoffie, line 640: I haven't tried it, but I think the connect dialog should work independent of this PR as it works based on index 0 without looking at the address at all.

It's other protocol consumers (such as Jamulus Explorer) which break without this change.

pljones: That means it's open to question where the fix should be made.

If Jamulus Explorer gets the address/port right with this fix, I'd rather leave the change in place and remove the work around in the client connect dialogue.  The above is meant as a reminder that there's code there needing investigation...  Maybe it says the wrong thing, though.

hoffie: I think this PR is correct and should be merged because 3.9.0 unintentionally changed protocol details. We don't have a formal specification of the protocol, so the Jamulus implementation effectively dictates the details.

Removing the logic in the client is not a trivial thing, IMO. If we remove it, *all* servers would have to generate a real, non-empty index 0, which is close to impossible... :)
Moving the logic from the GUI to the client makes sense, of course, and should be transparent to do.

pljones: > Removing the logic in the client is not a trivial thing, IMO.

I didn't say the _client_ shouldn't do it.  Actually, I think it should be in `src/protocol.cpp` when receiving the list.  That way any use would get the right value.  However, it _definitely_ shouldn't be in the UI.  That layer should _NOT_ be required for correct behaviour.  The RPC client, for example, _should_ be able to retrieve the server list and connect to any entry.

Issue #2818, opened 2022-08-29
Some Servers are inaccessible from a Client on the LAN of the Directory


pljones:
(Raised so I've an issue to fix.)

**Describe the bug**

### Discussed in https://github.com/orgs/jamulussoftware/discussions/2811
See thread at https://github.com/orgs/jamulussoftware/discussions/2811#discussioncomment-3499166 onwards.

![image](https://user-images.githubusercontent.com/1549463/187063801-405247d1-ba61-4276-9ae4-d5a2f70a78e7.png)

* The image on the left is `mtr -f 2 93.42.88.44` - that is, it's a progressive ICMP ping trace route.
* The image on the right is `mtr -f 2 -u -L 55859 -P 22124 93.42.88.44` - that is, much closer to the Jamulus protocol.

The IP and target port chosen here are visible and work on Jamulus Explorer
![image](https://user-images.githubusercontent.com/1549463/187063892-10ba8f04-0605-4c7d-81b1-fdc548894cb6.png)
and the source port is open at my end.  It's not just that host - if I run with `showallservers`, there's a lot of servers I see with empty ping:
![image](https://user-images.githubusercontent.com/1549463/187064259-52137256-0f03-418b-9cf4-718b2b79db24.png).

**To Reproduce**

See above.

**Expected behavior**

Jamulus Client should be able to see all servers where the server has been configured following the server setup guide.

**Screenshots**

See above.

**Operating system**

Various.

**Version of Jamulus**

Various.

**Additional context**

See the thread.

PR #2819, opened 2022-08-29
Directory: Use configured Directory Public IP for Client when hole-punching

pljones:

Some Servers are inaccessible from a Client on the LAN of the Directory.  This is because of how the "directory hole punch" works.  We advertise this as a way of avoiding interfering with a router, to ease running a server.  It works by sending a Jamulus message claiming to come from the client requesting the server list from the directory, to each server in the list.

However, where the client is on the same LAN as the directory, the source address for the client is the LAN address, not the public IP address which the server will subsequently receive.  Hence the "hole punch" fails.

This change tries to make use of the directory public IP address, where that is explicitly known.  (`CHostAddress().InetAddr` isn't enough - it needs to be set through `--serverpublicip`.)  It avoids _requiring_ a known value, however - it will simply continue failing as currently.

CHANGELOG: Directory: Use configured Directory Public IP for Client when hole-punching

**Context: Fixes an issue?**

Fixes: 2818 (#2818)

I would strongly suggest we review the documentation regarding the directory hole punch to make it clearer that there are known limitations in its effectiveness.

Genre Jazz is currently running with it.  I can see most of the servers that were previously not working.  The only one not working in my client is also not working in Jamulus Explorer.  So I consider than 100% effective.

Reviews.

## Checklist

### DIFF
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -649,6 +649,15 @@ void CServerListManager::RetrieveAll ( const CHostAddress& InetAddr )
         // if the client IP address is a private one, it's on the same LAN as the directory
         bool clientIsInternal = NetworkUtil::IsPrivateNetworkIP ( InetAddr.InetAddr );
 
+        CHostAddress clientPublicAddr = InetAddr;
+        if ( clientIsInternal && CHostAddress().InetAddr != ServerList[0].LHostAddr.InetAddr &&
+             !NetworkUtil::IsPrivateNetworkIP ( ServerList[0].LHostAddr.InetAddr ) )
+        {
+            // client and directory on same LAN, directory has public IP set, that should be suitable for the
+            // client, too (i.e. same router with same public IP will be used for both), so use it for client public IP
+            clientPublicAddr.InetAddr = ServerList[0].LHostAddr.InetAddr;
+        }
+
         const ushort iCurServerListSize = static_cast<ushort> ( ServerList.size() );
 
         // allocate memory for the entire list
@@ -676,14 +685,14 @@ void CServerListManager::RetrieveAll ( const CHostAddress& InetAddr )
                 vecServerInfo[iIdx].HostAddr = siCurListEntry.LHostAddr;
             }
 
-            // do not send a "ping" either to a server local to the directory (no need)
-            // or for a client local to the directory (the address will be wrong)
-            if ( !serverIsInternal && !clientIsInternal )
+            // do not send a "ping" to a server local to the directory (no need)
+            if ( !serverIsInternal )
             {
                 // create "send empty message" for all other registered servers
                 // this causes the server (vecServerInfo[iIdx].HostAddr)
-                // to send a "reply" to the client (InetAddr) - with the intent of opening its firewall for the client
-                pConnLessProtocol->CreateCLSendEmptyMesMes ( siCurListEntry.HostAddr, InetAddr );
+                // to send a "reply" to the client (InetAddr or best guess public IP address if internal to directory)
+                // - with the intent of opening the server firewall for the client
+                pConnLessProtocol->CreateCLSendEmptyMesMes ( siCurListEntry.HostAddr, clientPublicAddr );
             }
         }
 


src/serverlist.cpp
```diff
@@ -649,6 +649,14 @@ void CServerListManager::RetrieveAll ( const CHostAddress& InetAddr )
         // if the client IP address is a private one, it's on the same LAN as the directory
         bool clientIsInternal = NetworkUtil::IsPrivateNetworkIP ( InetAddr.InetAddr );
 
+        CHostAddress clientPublicAddr = CHostAddress ( InetAddr.InetAddr, InetAddr.iPort );
+        if ( clientIsInternal && CHostAddress().InetAddr != ServerList[0].LHostAddr.InetAddr &&
+             !NetworkUtil::IsPrivateNetworkIP ( ServerList[0].LHostAddr.InetAddr ) )
+        {
+            // client and directory on same LAN, directory has public IP set, use that for client public IP
```
pljones, line 656: For some mad crazy reason, in this version of serverlist.cpp, LHostAddr is not the local host address, it's the Public IP for the directory.  This is because the directory doesn't swap it's HostAddr and LHostAddr when creating its server list entry...

softins:
Will look at the code later, but just wanted to pick up on one part of the description!

> It works by sending a Jamulus message claiming to come from the client requesting the server list from the directory, to each server in the list.

This is not accurate. The directory sends a Jamulus message to each listed server _from itself_ (not claiming to be from the client). It is the content of the message the says "please send a message to  <client>". When a server acts on that message by sending a message to the specified client, that has the effect of priming the server's NAT/firewall to accept subsequent messages from that client.

pljones:
> > It works by sending a Jamulus message claiming to come from the client requesting the server list from the directory, to each server in the list.
> 
> This is not accurate. The directory sends a Jamulus message to each listed server _from itself_ (not claiming to be from the client). It is the content of the message that says "please send a message to `<client>`". When a server acts on that message by sending a message to the specified client, that has the effect of priming the server's NAT/firewall to accept subsequent messages from that client.

Yep, I realised that on a re-read of the code and changed an "as" to back to "to".  So there's a protocol message actually from the directory (and source address says so), actually to the server (dest address says so) carrying the address of the client requesting the server list.  The server is then expected to send (i.e. "reply" but not a reply as such at all) a message to that address, opening its firewall to a reply (which may never happen but woud likely be the client pinging the server).

src/serverlist.cpp
```diff
@@ -649,6 +649,14 @@ void CServerListManager::RetrieveAll ( const CHostAddress& InetAddr )
...
+            clientPublicAddr.InetAddr = ServerList[0].LHostAddr.InetAddr;
```
ann0see, line 657: Is this always safe? Could we get the public address from the client directly?

pljones: The point here is that the client doesn't know its public IP address - it's on the same LAN as the directory.  The client _never_ knows its public IP address.

And hence, where the `if` test is true, yes, this is the best we can manage: the client is on the same LAN, the LHostAddr isn't "default" (unusable) and it's also not on the same LAN (unuseable).  It's set in the class constructor, either from `--serverpublicip` or from autodetection (which will get a value `IsPrivateNetworkIP` says "yes" to, unless it's actually useable).

At the worst case, we end up using a value that's no use and the client can't connect -- i.e. exactly the same as it is now.

ann0see: Ok. Maybe the comment should then say: ... use that for client public IP as it is most likely the same as the client.

pljones: Done.

ann0see: Thanks.

src/serverlist.cpp
hoffie: `LHostAddr` should probably be renamed to something more generic such as `AlternateHostAddr`?

src/serverlist.cpp
```diff
@@ -676,14 +685,14 @@ void CServerListManager::RetrieveAll ( const CHostAddress& InetAddr )
...
-            // or for a client local to the directory (the address will be wrong)
-            if ( !serverIsInternal && !clientIsInternal )
+            // do not send a "ping" to a server local to the directory (no need)
+            if ( !serverIsInternal )
             {
                 // create "send empty message" for all other registered servers
                 // this causes the server (vecServerInfo[iIdx].HostAddr)
-                // to send a "reply" to the client (InetAddr) - with the intent of opening its firewall for the client
-                pConnLessProtocol->CreateCLSendEmptyMesMes ( siCurListEntry.HostAddr, InetAddr );
+                // to send a "reply" to the client (InetAddr or best guess public IP address if interal to directory)
```
hoffie, line 693:
```suggestion
                // to send a "reply" to the client (InetAddr or best guess public IP address if internal to directory)
```

src/serverlist.cpp
```diff
@@ -649,6 +649,14 @@ void CServerListManager::RetrieveAll ( const CHostAddress& InetAddr )
         // if the client IP address is a private one, it's on the same LAN as the directory
         bool clientIsInternal = NetworkUtil::IsPrivateNetworkIP ( InetAddr.InetAddr );
 
+        CHostAddress clientPublicAddr = CHostAddress ( InetAddr.InetAddr, InetAddr.iPort );
```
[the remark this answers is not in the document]
hoffie: Probably ok here:
https://github.com/jamulussoftware/jamulus/blob/6f10b9850de001062a62e3a8d75c089b8fa61072/src/util.h#L750-L755

hoffie (approved):
Logic seems good to me, but haven't run any tests. Should certainly wait for Tony's review now that he has offered it! :)

@pljones PR title/Changelog/Commit title could even more specific. Maybe
`Directory: Use configured Public IP for Client when hole-punching`?

src/serverlist.cpp
pljones: Not really - it holds the local host address when sending the protocol message.  (Also, it's easy enough to fix the directory to store its own details the "right" way around.  And, indeed, I have done that... I could submit a patch... I've rather a lot of code changes stacked up in a single commit currently...  I should break them up...)

pljones:
https://jamulus.io/wiki/Custom-Directories (to be renamed, IIRC) should have an additional note added under "Points to note about Directories":

* When using a client on the same LAN as a directory, the directory itself needs to be run with the correct `--serverpublicip` value, so that any (non-LAN) servers registering with the directory that require the "hole punch" can be accessed by the LAN client(s).  This is because otherwise the directory would provide the LAN address of the client to the server and the "hole punch" would fail.

ann0see:
@gilgongo should judge that

pljones:
And probably get confirmation I've got it clear from @softins, too 😄.

softins (approved):
I can't really test this, as it is particular to a rare scenario, such as that of @pljones. But the logic looks sound.

As I understand it, it only affects the unusual situation where a directory server is on the same LAN as the client, but the directory server has external servers registered to it via port forwarding. In this situation, it ensures that the IP address specified in the body of the message request sent to the remote servers is actually the public IP (assumed shared by the client and the directory), and not the LAN IP of the local client.

pljones:
Running on all my directories (and servers) apart from updatecheck2.jamulus.io, which remains on latest.

pljones:
https://github.com/jamulussoftware/jamuluswebsite/pull/815 raised, so removing "Needs documentation" label.

PR #2823, opened 2022-08-29
Update celt_encoder.c

38github:
Let those sweet sharp transients from rides, hi-hats, western guitar strings etc be there for us to enjoy! Until we have lossless audio this will probably make some people happy.

I changed int isTransient=0 to the value of 1.

Nothing that I know of. I have compiled it and used it as server and client on a couple of computer and no issues at all.

## Checklist

### DIFF
--- a/libs/opus/celt/celt_encoder.c
+++ b/libs/opus/celt/celt_encoder.c
@@ -1429,7 +1429,7 @@ int celt_encode_with_ec(CELTEncoder * OPUS_RESTRICT st, const opus_val16 * pcm, 
    celt_sig *prefilter_mem;
    opus_val16 *oldBandE, *oldLogE, *oldLogE2, *energyError;
    int shortBlocks=0;
-   int isTransient=0;
+   int isTransient=1;
    const int CC = st->channels;
    const int C = st->stream_channels;
    int LM, M;


libs/opus/celt/celt_encoder.c
```diff
@@ -1429,7 +1429,7 @@ int celt_encode_with_ec(CELTEncoder * OPUS_RESTRICT st, const opus_val16 * pcm,
    celt_sig *prefilter_mem;
    opus_val16 *oldBandE, *oldLogE, *oldLogE2, *energyError;
    int shortBlocks=0;
-   int isTransient=0;
+   int isTransient=1;
```
pljones, line 1432: Does this have any effect?  A quick grep shows:
```
1432:   int isTransient=0;
1691:   isTransient = 0;
1699:      isTransient = transient_analysis(in, N+overlap, CC,
```

38github: Here is samples that I have encoded where *transient0 is without modification  and *transient1 is using what I posted. For some it is subtle but for example listen to how sharp the ride hits are. 
[NOFX-Shes_Gone.zip](https://github.com/jamulussoftware/jamulus/files/9447196/NOFX-Shes_Gone.zip)

hoffie: Thanks for the samples. I do see that the files have different checksums (but even running `opusenc` twice produces two different checksums), but I don't hear a difference and don't see one in the wave form either.
After @pljones's hint I'm not even sure if I should expect to hear one.

@38github Can you explain
1. what the flag does,
2. why it is safe to modify,
3. why it is supposed to change anything at all despite @pljones's remark that the change is effectively overwritten later in the code?

Also, has there been any public discussion with the opus devs?

Issue #2680, opened 2022-06-28
Option ctrlmidich writes fader tags in .ini file as Channel':'Name


pcar75:
**Describe the bug**
When using the `--ctrmidich` option, the jamulus.ini file is rewritten with fader tags identified by `ChannelNumber:Name` which defeats the purpose of conserving fader info since the same user only has a 1/250 (I think) chance of being assigned the same channel number in a future session.

**To Reproduce**

1. Provided a previous Jamulus usage was without this option, the jamulus.ini  fader tags are identified with the Base64 hash as `Name` only.
2. While Jamulus is not executing , make a backup copy of jamulus.ini file
3. Launch the Jamulus client with the --ctrlmidich option (ex.: `--ctrlmidich "1;o2"` ; will work even if you don't have a midi device)
4. Connect to a site with participants : notice that if one was muted in previous session, it will not be so now and you will have to mute it again (should apply to any other fader settings)
5. For each participant's fader, the diplayed information is now identified with `ChannelNumber:Name` along with instrument and country if provided.
6. Disconnect and quit Jamulus
7. The rewritten Jamulus .ini file will now contain fader tags identified as `ChannelNumber:Name`
 

Note: the Base64 encryption/decryption can be verified with [http://www.unit-conversion.info/texttools/base64/](url)

**Expected behavior**
The usage of fader info in .ini file should always be written and used as identification as `Name` , not `ChannelNumber:Name` .

**Screenshots**

![Jamulus Client_display with Channel numbers_02](https://user-images.githubusercontent.com/64435162/176254912-d612e436-8e17-40bb-9323-115d043dc13c.png)
_Names modified_

jamulus.ini file when using the --ctrlmidich option.

```
 <storedfadertag5_base64>OTpObyBOYW1l</storedfadertag5_base64>
 <storedfaderlevel5>2</storedfaderlevel5>
 <storedpanvalue5>50</storedpanvalue5>
 <storedfaderissolo5>0</storedfaderissolo5>
 <storedfaderismute5>1</storedfaderismute5>
 <storedgroupid5>-1</storedgroupid5>

```
In Base64 "OTpObyBOYW1l" equals "9:No Name"  i.e. channel number, ":" and name.
It should stay as "Tm8gTmFtZQ==" equals "No Name" i.e name only

**Operating system**
This was observed on up-to-date Windows 10.

**Version of Jamulus**
Jamulus standard 3.8.2 from the Jamulus site.

**Additional context**
[...]

Issue #2813, opened 2022-08-28
Prepare Release 3.9.1


hoffie:
**Target timeline**

| Phase | Target date | Original | Actual |
| :--- | ---: | ---: | ---: |
| Scheduled feature freeze / Start of translation process | **2022-09-24** | **2022-09-24** | **2022-09-21** (app translations) |
| Targeted translation completion date | **2022-10-08** | **2022-10-08** | **2022-10-09** (app translations, forced) |
| Approximate release date | **2022-10-15** | **2022-10-15** | **2022-10-17** (announcement 2022-10-21) |

Current state: **Released** 

**Checklist**
- [x] Assign this issue to the release shepherd who is in charge of managing this checklist.
- [x] Pin this issue
- [x] Ensure that all issues/PR targeted for this release are done by checking the Project board with [the appropriate filter for this release](https://github.com/orgs/jamulussoftware/projects/2?card_filter_query=milestone%3A%22Release%203.9.1%22).
- [x] Remind main developers to review entries in *Waiting on team* state.
- [x] De-tag unfinished Issues/PRs.
- [x] Declare a freeze for code and website by updating this Issue and adding a comment. PRs can still be worked on and may get reviewed, but must not be merged unless agreed explicitly.
- [x] Check the `needs documentation` label for any outstanding PRs flagged for this release and remove that label if done.
- [x] Check ./Jamulus -h output against the [Include-Client](https://github.com/jamulussoftware/jamuluswebsite/blob/next-release/wiki/en/Include-Client-Commands.md)/[Server-Commands](https://github.com/jamulussoftware/jamuluswebsite/blob/next-release/wiki/en/Include-Server-Commands.md) pages and [man page](https://github.com/jamulussoftware/jamulus/blob/master/distributions/Jamulus.1). Update if necessary.
- [x] Start Website translations
  - [x] [Check for broken links](https://github.com/jamulussoftware/jamuluswebsite/blob/release/contribute/en/Release-Process.md#website-check-links)
  - [x] Open a Pull Request from `next-release` to release, set it as "Draft", sanity check for conflicts and any obvious problems.
  - [x] Declare a full freeze of the `next-release` and `release` branch. No changes should be made from now on to ensure translators don't have to work twice.
  - [x] Check if the list of translators in `tools/create-translation-issues.sh`. Make sure issue text is up-to-date. Add any URLs that will need localisation into the "New/Changed screenshots" section.
[...]

PR #2839, opened 2022-09-05
Client: Pass through flag for ctrlmidich used

pljones:

Rather than changing the name of the connecting client (and losing the actual name), based on whether `--ctrlmidich` is used, pass that information to the fader itself.

Also, to partly undo the damage, attempt to recover fader details previously saved with the channel number prefix.

CHANGELOG: Client: Bugfix: Remove channel number from saved fader tag when `--ctrlmidich` is in use.  In this release `--cleanuplegacyfadersettings` can be used on any saved fader settings that have been corrupted to attempt recovery.

**Context: Fixes an issue?**

Fixes: #2680 

No.

I've tested it local.  The following should be borne in mind:
- This does not remove entries from Jamulus.ini (it _should_ match the old ones, but they'll still be there, cleaned values will get written).
- I've not tested fader sort options.

User test: @pcar75 and @rdica - can you take a look, please?

## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -572,6 +572,20 @@ int main ( int argc, char** argv )
             continue;
         }
 
+        // Clean up legacy fader settings --------------------------------------
+        // Undocumented temporary command line argument: Clean up fader settings
+        // corrupted by bug #2680.  Only needs to be used once (per file).
+        if ( GetFlagArgument ( argv,
+                               i,
+                               "--cleanuplegacyfadersettings", // no short form
+                               "--cleanuplegacyfadersettings" ) )
+        {
+            qInfo() << "- will clean up legacy fader settings on load";
+            CommandLineOptions << "--cleanuplegacyfadersettings";
+            ClientOnlyOptions << "--cleanuplegacyfadersettings";
+            continue;
+        }
+
         // Unknown option ------------------------------------------------------
         qCritical() << qUtf8Printable ( QString ( "%1: Unknown option '%2' -- use '--help' for help" ).arg ( argv[0] ).arg ( argv[i] ) );
 
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -226,11 +226,13 @@ void CClientSettings::SaveFaderSettings ( const QString& strCurFileName )
     WriteToFile ( strCurFileName, IniXMLDocument );
 }
 
-void CClientSettings::ReadSettingsFromXML ( const QDomDocument& IniXMLDocument, const QList<QString>& )
+void CClientSettings::ReadSettingsFromXML ( const QDomDocument& IniXMLDocument, const QList<QString>& CommandLineOptions )
 {
     int  iIdx;
     int  iValue;
     bool bValue;
+
+    bCleanUpLegacyFaderSettings = CommandLineOptions.contains ( "--cleanuplegacyfadersettings" );
 
     // IP addresses
     for ( iIdx = 0; iIdx < MAX_NUM_SERVER_ADDR_ITEMS; iIdx++ )
@@ -557,6 +559,53 @@ void CClientSettings::ReadSettingsFromXML ( const QDomDocument& IniXMLDocument, 
     ReadFaderSettingsFromXML ( IniXMLDocument );
 }
 
+QString CClientSettings::CleanUpLegacyFaderSetting ( QString strFaderTag, int iIdx )
+{
+    bool ok;
+    int  iIdy;
+    bool bDup;
+
+    if ( !bCleanUpLegacyFaderSettings || strFaderTag.isEmpty() )
+    {
+        return strFaderTag;
+    }
+
+    QStringList slChanFaderTag = strFaderTag.split ( ":" );
+    if ( slChanFaderTag.size() != 2 )
+    {
+        return strFaderTag;
+    }
+
+    const int iChan = slChanFaderTag[0].toInt ( &ok );
+    if ( ok && iChan >= 0 && iChan <= MAX_NUM_CHANNELS )
+    {
+        // *assumption*: legacy tag that needs cleaning up
+        strFaderTag = slChanFaderTag[1];
+    }
+
+    // duplicate detection
+    // this assumes the first entry into the vector is the newest one and skips any later ones.
+    // the alternative is to use iIdy for the vector entry, so overwriting the duplicate.
+    // (in both cases, this currently leaves holes in the vector.)
+    bDup = false;
+    for ( iIdy = 0; iIdy < iIdx; iIdy++ )
+    {
+        if ( strFaderTag == vecStoredFaderTags[iIdy] )
+        {
+            // duplicate entry
+            bDup = true;
+            break;
+        }
+    }
+    if ( bDup )
+    {
+        // so skip all settings for this iIdx (use iIdx here even if using iIdy and not doing continue below)
+        return QString();
+    }
+
+    return strFaderTag;
+}
+
 void CClientSettings::ReadFaderSettingsFromXML ( const QDomDocument& IniXMLDocument )
 {
     int  iIdx;
@@ -566,8 +615,17 @@ void CClientSettings::ReadFaderSettingsFromXML ( const QDomDocument& IniXMLDocum
     for ( iIdx = 0; iIdx < MAX_NUM_STORED_FADER_SETTINGS; iIdx++ )
     {
         // stored fader tags
-        vecStoredFaderTags[iIdx] =
-            FromBase64ToString ( GetIniSetting ( IniXMLDocument, "client", QString ( "storedfadertag%1_base64" ).arg ( iIdx ), "" ) );
+        QString strFaderTag = CleanUpLegacyFaderSetting (
+            FromBase64ToString ( GetIniSetting ( IniXMLDocument, "client", QString ( "storedfadertag%1_base64" ).arg ( iIdx ), "" ) ),
+            iIdx );
+
+        if ( strFaderTag.isEmpty() )
+        {
+            // duplicate from clean up code
+            continue;
+        }
+
+        vecStoredFaderTags[iIdx] = strFaderTag;
 
         // stored fader levels
         if ( GetNumericIniSet ( IniXMLDocument, "client", QString ( "storedfaderlevel%1" ).arg ( iIdx ), 0, AUD_MIX_FADER_MAX, iValue ) )
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -29,7 +29,8 @@
 \******************************************************************************/
 CChannelFader::CChannelFader ( QWidget* pNW ) :
     eDesign ( GD_STANDARD ),
-    BitmapMutedIcon ( QString::fromUtf8 ( ":/png/fader/res/mutediconorange.png" ) )
+    BitmapMutedIcon ( QString::fromUtf8 ( ":/png/fader/res/mutediconorange.png" ) ),
+    bMIDICtrlUsed ( false )
 {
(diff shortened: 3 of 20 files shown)


ignotus666:
Won't be able to test until tomorrow but will do so.

pcar75:
Bonjour,

> What is missing until this pull request can be merged?
User test: @pcar75 and @rdica - can you take a look, please?

Sorry I missed @pljones  message above ... first time i participate in a project (since retirement he he ).

Before any tests , I need to know that I could - and how I would - revert to my current Jamulus version ?
Given that , I can test in my Windows 10 env ... not my new Ubuntu Studio env as it is still somewhat unstable.
- Where do I get the modified version and how to install it ?
- Any specifics tests/procedures you would like me to run ?
- Where and how do I report results ?

Thanks

pljones:
Hi @pcar75

> Where do I get the modified version and how to install it ?

Open the "Checks" tab on this pull request in a new browser tab then select the "Auto-Build" group.  You'll see a number of "Artifacts".  Pick whichever you need -- it will be a ZIP containing the normal install file for that platform.

When installing, I'd recommend using a separate folder, then "reverting" isn't a problem (the windows installer will remember this new location as the default for your next install but nothing else will care).

Never use one of your "real" inifiles for test purposes - back them up (just in case) and use copies (not the backups, fresh copies...).

For test purposes:
- run 3.9.0 with a clean `--inifile` (either take a copy and manually just delete all the fader details or start afresh)
  - save to a new save file
  - check the bug is there
  - check how fader sorting works
  - exit the app and see the inifile has the bug, too
- run the build from this pull request (again, either a copy of your 3.9.0 inifile with the fader details removed or a fresh one)
  - save to a new save file
  - check the bug fixed
  - check how fader sorting works - it should be the same... hopefully...
  - exit the app and see the inifile is also good
- run the build from this pull request (again, ideally a copy of your 3.9.0 inifile with the buggy fader details)
  - take a copy of a buggy save file
  - load the settings from the copy
  - save to a new save file
  - check the bug fixed

pljones:
You haven't followed the instructions.
Click: https://github.com/jamulussoftware/jamulus/pull/2839/checks
Click on image
Scroll down and find the build you want:
image
The download will be a ZIP.

ignotus666:
Just did some testing and it seems to work as advertised. An old .ini file with the bug gets written over with correct fader tags (no prepending with a number) and sorting seems to be correct too.

ignotus666 (approved):

pcar75:
> run the build from this pull request (again, either a copy of your 3.9.0 inifile with the fader details removed or a fresh one) 

Test Ok : --ctrlmidich does not make Jamulus write ini file with ChannelNumber,
ImportExport file also OK.
Sort by Name does not sort by ChannelNumber.

> run the build from this pull request (ideally a copy of your 3.9.0 inifile with the buggy fader details) 

Test Ok : buggy ini with Jamulus no --ctrlmidich is rewritten without ChannelNumber.
ImportExport file also OK
Sort by Name Ok

pljones:
Thanks both!

src/settings.cpp
```diff
@@ -566,8 +566,20 @@ void CClientSettings::ReadFaderSettingsFromXML ( const QDomDocument& IniXMLDocum
     for ( iIdx = 0; iIdx < MAX_NUM_STORED_FADER_SETTINGS; iIdx++ )
     {
         // stored fader tags
-        vecStoredFaderTags[iIdx] =
+        QString strFaderTag =
             FromBase64ToString ( GetIniSetting ( IniXMLDocument, "client", QString ( "storedfadertag%1_base64" ).arg ( iIdx ), "" ) );
+        bool        ok;
```
pljones, line 571: Move into the correct scope - inside the `if`.

src/settings.cpp
```diff
@@ -566,8 +566,20 @@ void CClientSettings::ReadFaderSettingsFromXML ( const QDomDocument& IniXMLDocum
...
+        QStringList slChanFaderTag = strFaderTag.split ( ":" );
+        if ( slChanFaderTag.size() == 2 )
+        {
+            const int iChan = slChanFaderTag[0].toInt ( &ok );
+            if ( ok && iChan >= 0 && iChan <= MAX_NUM_CHANNELS )
+            {
+                // *assumption*: legacy tag that needs cleaning up
+                strFaderTag = slChanFaderTag[1];
```
pljones, line 579: This could potentially cause duplicate entries in `vecStoredFaderTags` - check 0..iIdx-1 for matches?

pcar75:
Bonjour,
Shouldn't there be a test with a real MIDI device ? (I don't have one.)

ignotus666:
I tested it with a MIDI device - works fine.

src/settings.cpp
pljones: Added duplicate detection.

hoffie (commented):
The code looks good to me. I've got one behavioral question though:

Before this change, MIDI channel ifs were hacked into the fader name and neither the fader nor the settings logic knew about that, right?

With this change, the fader learns about that logic and keeps the information properly separated. This has the consequence that a user who used the feature previously would lose the stored fader values assignments as the fader tags would not match, right?

Therefore, this PR contains compatibility code to make those old entries fit into the new (better) scheme, right?

My worries are that this cleanup logic also applies to matching fader tags which did not stem from midi channel usage at all.

If my Jamulus name was hoffie:1, my name would be reduced to just hoffie and th entry would no longer match. Is that correct or is something preventing this?

If this is the case, do we really want that? It would help MIDI users but might introduce such bugs for non-MIDI users. Should we limit this conversion behavior to MIDI users maybe?
We should also add a note when this logic was introduced (version) and when it would be safe to drop.

(I'm writing this from mobile so I might have missed a relevant detail)

pljones:
> If my Jamulus name was hoffie:1, my name would be reduced to just hoffie and th entry would no longer match. Is that correct or is something preventing this?

No.  The logic would "clean up" `1:hoffie` to `hoffie` but leave `any non-numeric:anything at all` alone.

henkdegroot:
> I'm hoping the "clean up" code can be removed in 3.10.0 anyway.

Hmm... so if a user would skip the 3.9.1 version and go straight to 3.10.0, it won't "clean up"?
Perhaps do as @hoffie suggested to only perform this actions when you have started the client with the ctrlmidich option enabled?
As you considered it was important enough to "convert" incorrect entries.....it should not go into the waste bin that soon.

pljones:
> Should we limit this conversion behavior to MIDI users maybe?

> Perhaps do as @hoffie suggested to only perform this actions when you have started the client with the ctrlmidich option enabled?

This means _even more_ invasive code just to pass around the flag with all the additional overhead of then removing it again...

"Old" entries will, over time, get dropped anyway.  So the "clean up" is more so that existing entries get used - regardless of whether in this session `--ctrlmidich` is in use or not.  If it's seen as "too dangerous" because of the outside chance of some users naming themselves "<valid integer><colon><anything>", then the safe option is to just drop the code and let the aging take care of unused entries.

pljones:
OK, because I didn't like using `--ctrlmidich` itself, what I've done is add a new ("undocumented") command line option: `--cleanuplegacyfadersettings`.  You should only need to run a client with this once - any fader settings loaded whilst the option is active should get cleaned.  Without it, the client shouldn't interfere with the tags (i.e. it won't even try to clean them up).

@pcar75, @rdica and all, if you could run another few tests:

1. Test without the option to show that it doesn't clean up the fader tags.
2. Test the `--cleanuplegacyfadersettings` by running the previous tests, using this option when testing the new code.  Check all tags that should be cleaned are cleaned.
3. After cleaning up, run without the option and check that the fader tags remain clean.

ann0see:
> You should only need to run a client with this once - any fader settings loaded whilst the option is active should get cleaned. 

I had a thought of all this compatibility code and am not sure how bad (read: inconvenient) re-setting the fader levels really is in practice. Is it worth Introducing an option like that or the compatibility code at all? Personally, I'd just set the fader levels etc. again once this person shows up again --> no compatibility code needed

pljones:
> Is it worth Introducing an option like that or the compatibility code at all?

I'm still of the opinion that if it stays in 3.9.1, then it goes in 3.10.0.  Like I say, I'd be as happy leaving it out.

ann0see:
Agree. 

However, I'm a bit more in favor of not adding any compatibility code or removing it quickly. I assume every compatibility code will not be able to solve 100% of the issue. 

--cleanuplegacyfadersettings: We could also only add this flag and tell users via release announcement and on the website that they should start Jamulus with this once. After 2 major releases we'd remove the option.

src/main.cpp
```diff
@@ -572,6 +572,20 @@ int main ( int argc, char** argv )
             continue;
         }
 
+        // Clean up legacy fader settings --------------------------------------
+        // Undocumented debugging command line argument: Clean up fader settings
```
hoffie, line 576:
```suggestion
        // Undocumented temporary command line argument: Clean up fader settings
```

hoffie (approved):
I like the updated approach with the temporary cleanup command line flag. Approving with a minor comment update suggestion.

What should be done when about to merge:
- [ ] Create cleanup issue and assign it to the appropriate milestone (I'm fine with either 3.10.0 or 3.11.0, slight tendency to the latter) and link to this diff.
- [ ] Create a reminder in #2813 to list this in the announcement.

I have a hard time gaining a proper gut feeling for the usefulness if the cleanup code in this case. Had we decided beforehand, I might have voted to skip it. Now that it has been developed I'm all for making it available for a limited amount of time (i.e. a clear removal plan).

src/main.cpp
pljones: Bah, cut-n-paste error :)

pljones: Fixed.

pljones:
> * Maybe also document the special flag as part of the Changelog with a note saying that this flag is only required once and is about to be removed in an upcoming version again.

I've added it to the "changelog" tag on the PR - I've tried to keep it short but it can be edited/formatted better when we get there.

src/main.cpp
```diff
@@ -572,6 +572,20 @@ int main ( int argc, char** argv )
             continue;
         }
 
+        // Clean up legacy fader settings --------------------------------------
+        // Undocumented temporary command line argument: Clean up fader settings
```
ann0see, line 576: Since we agreed to remove this soon, I'd prefer to document it in the help text.

pljones: It's going to be totally meaningless to most users.  To those for whom it's applicable, they'll use it once and never again.  I can't see the point of making the help text get even larger than it is now for this.

pljones: > If it shows up under --help, the version supports it,

If running the app with the option doesn't complain it's not recognised, it's supported, too.

Also, if it goes into help, that means we should update the website User Guide, too.

ann0see: > If running the app with the option doesn't complain it's not recognised, that means it's supported, too.

Fair point. 

I'd still like to have it prominently documented somewhere besides the release announcement. I assume most users don't read the announcement at all.

pljones: True, true.  We could add it to the hints and tips about `--ctrlmidich`?  It's less prominent but it's more relevant.

ann0see: Probably? 
What about printing it to stdout if the --ctrlmidich option is set and to the website as long as it's in?

ann0see: We could write a forum post with a how to?

pljones: This all arose in https://github.com/orgs/jamulussoftware/discussions/2678 -- that could have a comment added with how to fix it.

pljones: I've noted on the _Reminders for the release announcement_ to comment on the original discussion, so going ahead with the merge here.

Issue #2362, opened 2022-02-08
CLI --serverinfo: Add ISO country code support in addition to Qt Country IDs


hoffie:
Currently, `--serverinfo` requires users to specify Qt Country IDs. This is not user-friendly.
With Qt6, `--serverinfo` will require users to specify Qt**5** Country IDs, even if they are on Qt6. This is really not user-friendly.

**Has this feature been discussed and generally agreed?**

The ugliness of the status quo, especially when adding support for Qt6, has been officially stated by @pljones here: https://github.com/jamulussoftware/jamulus/pull/2299#pullrequestreview-876366907 :)

**Describe the solution you'd like**
 
`--serverinfo` should support ISO country codes, e.g. `--serverinfo 'FooServer;FooCity;de`
We should still keep support for the numerical IDs for compatibility.

pljones:
The current Locale -> Country letter method could possibly be used in reverse: take a two letter country code (`de` or `zw`), then look up the Locale based on the language (`EN-us`, `DE-de`, etc) to get the Qt Locale and hence the related country code for the "current" Qt.  Then magic mapping gets that into the wire format, as per #2299.

If I've understood it correctly.  I could be wrong...

softins:
> `--serverinfo` should support named countries, e.g. `--serverinfo 'FooServer;FooCity;FooCountry` We should still keep support for the numerical IDs for compatibility.

For FooCountry, I think we should use the international standard abbreviations, rather than names, as defined in [ISO-3316](https://www.iso.org/iso-3166-country-codes.html). Either the 2-letter or 3-letter variants (we could accept both). It then becomes independent of language, which could complicate things if we tried to interpret country names (e.g. Germany/Deutschland/Allemagne, Holland/Netherlands/Nederland, Belgium/Belgien/Belgie)

acerix:
After updating to 3.9.0 and Qt6, the country of my server changed without any changes to the config, apparently it is using Qt6 Country IDs now.  I fixed it by updating to the new code in this list: https://doc.qt.io/qt-6/qlocale.html#Country-enum

Agree it would be more user friendly to specify country names/codes.

hoffie:
> After updating to 3.9.0 and Qt6, the country of my server changed without any changes to the config, apparently it is using Qt6 Country IDs now.

Thanks for the report! This is unintended, but I can see what's missing. I've opened #2809 to fix this with rather high priority.

hoffie:
[...]

PR #2841, opened 2022-09-07
Server: Add --serverinfo ISO country code support

hoffie:
With this change, --serverinfo gains support two-letter ISO country codes    such as "de", "gb", "nl", "fr". Numeric Qt5-style codes keep working as expected.

To simplify user-side debugging, the `--serverinfo` parsing now contains logging for the final values (or a warning in case of parsing problems).

CHANGELOG: Server: Added support for ISO country codes (de, gb, nl, ...) in --serverinfo.

**Context: Fixes an issue?**
Fixes: #2362
Related: https://github.com/jamulussoftware/jamulus/pull/2829#discussion_r965316253

As this is a (tiny) new feature, docs would be useful, but not mandatory. This is not a breaking change. Delaying the docs update would be no problem.

Ready

- [x] #2829 must be merged first

## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1075,7 +1075,7 @@ QString UsageArguments ( char** argv )
            "  -L, --licence         show an agreement window before users can connect\n"
            "  -m, --htmlstatus      enable HTML status file, set file name\n"
            "  -o, --serverinfo      registration info for this Server.  Format:\n"
-           "                        [name];[city];[country as Qt5 QLocale ID]\n"
+           "                        [name];[city];[country as two-letter ISO country code or Qt5 QLocale ID]\n"
            "      --serverpublicip  public IP address for this Server.  Needed when\n"
            "                        registering with a server list hosted\n"
            "                        behind the same NAT\n"
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -205,15 +205,26 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
         // [this server country as QLocale ID]
         bool      ok;
         const int iCountry = slServInfoSeparateParams[2].toInt ( &ok );
-        if ( ok && iCountry >= 0 && CLocale::IsCountryCodeSupported ( iCountry ) )
+        if ( ok )
         {
-            // Convert from externally-supplied format ("wire format", Qt5 codes) to
-            // native format. On Qt5 builds, this is a noop, on Qt6 builds, a conversion
-            // takes place.
-            // We try to do such conversions at the outer-most interface which is capable of doing it.
-            // Although the value comes from src/main -> src/server, this very place is
-            // the first where we have access to the parsed country code:
-            ThisServerListEntry.eCountry = CLocale::WireFormatCountryCodeToQtCountry ( iCountry );
+            if ( iCountry >= 0 && CLocale::IsCountryCodeSupported ( iCountry ) )
+            {
+                // Convert from externally-supplied format ("wire format", Qt5 codes) to
+                // native format. On Qt5 builds, this is a noop, on Qt6 builds, a conversion
+                // takes place.
+                // We try to do such conversions at the outer-most interface which is capable of doing it.
+                // Although the value comes from src/main -> src/server, this very place is
+                // the first where we have access to the parsed country code:
+                ThisServerListEntry.eCountry = CLocale::WireFormatCountryCodeToQtCountry ( iCountry );
+            }
+        }
+        else
+        {
+            QLocale::Country qlCountry = CLocale::GetCountryCodeByTwoLetterCode ( slServInfoSeparateParams[2] );
+            if ( qlCountry != QLocale::AnyCountry )
+            {
+                ThisServerListEntry.eCountry = qlCountry;
+            }
         }
     }
 
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -226,6 +226,15 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
                 ThisServerListEntry.eCountry = qlCountry;
             }
         }
+        qInfo() << qUtf8Printable ( QString ( "Using server info: name = \"%1\", city = \"%2\", country/region = \"%3\" (%4)" )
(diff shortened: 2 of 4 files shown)


hoffie:
Marking it as Draft to prevent accidential merges before #2829 is merged. Feel free to un-mask afterwards.
As to whether this should go into 3.9.1 or not: In theory, it's a (tiny) small feature and would therefore rather belong to 3.10.0. If you (mainly @pljones and @ann0see) deem it ready to merge, feel free to include it in 3.9.1. Otherwise, please move the milestone.

pljones:
I'm hoping to get this in to 3.9.1 - mostly because I'd love to get away from the numbers in the server info :).

hoffie:
Rebased and un-draft-ed as #2829 has been merged.
(I'm planning to work on #2845 in a seperate PR)

src/util.cpp
```diff
@@ -1436,6 +1436,27 @@ bool CLocale::IsCountryCodeSupported ( unsigned short iCountryCode )
 #endif
 }
 
+QLocale::Country CLocale::GetCountryCodeByTwoLetterCode ( QString sTwoLetterCode )
+{
+    QList<QLocale> vLocaleList = QLocale::matchingLocales ( QLocale::AnyLanguage, QLocale::AnyScript, QLocale::AnyCountry );
+    QStringList    vstrLocParts;
+
+    // Qt < 6.2 does not support lookups from two-letter iso codes to
```
pljones, line 1444: We have Qt6 code that does `#if` on the version, don't we?  Why not here?

hoffie: Yes, we do. At most places it's about different *required* logic per Qt version though (and we don't have minor version dependent `#if`s for Qt6, IIRC).

So, the downside is that we have different implementations for Qt5 and Qt6 although we could live with a single one (less complexity, risk for differences between that two implementations).
The upside is that we would have the "better" code in place right now, therefore making it easier to drop the legacy path in X years.

I can add a 6.2+ code path via `QLocale::codeToTerritory()` as part of this PR.
Would you prefer that?

(Additionally, we could even have a 6.1 code path via `QLocale::codeToCountry`, which was introduced and deprecated right away due to the renaming to `QLocale::codeToTerritory` in 6.2+. I guess this step would really not be worth it as we don't use 6.1 anywhere in official releases. It would therefore release zero testing.)

ann0see: I think there was a place with commented Qt5 code and a Qt4 implementation with the country codes. We could do the same here.

hoffie: > I think there was a place with commented Qt5 code and a Qt4 implementation with the country codes.

Yep.

> We could do the same here.

I guess we would drop the whole helper method `GetCountryCodeByTwoLetterCode` in favor of the native Qt6.2+ function. So this would probably boil down to just updating the comment with the proper Qt native method name?

pljones: So the two paths would either be "call the native method" or "call the compatibiltity method", depending on the Qt version?

hoffie: > So the two paths would either be "call the native method" or "call the compatibiltity method", depending on the Qt version?

Exactly.

hoffie: I've added that now.
I've also added the suggested logging in a seperate commit (initially proposed in https://github.com/jamulussoftware/jamulus/pull/2829#discussion_r965316253)

src/main.cpp
```diff
@@ -1075,7 +1075,7 @@ QString UsageArguments ( char** argv )
            "  -L, --licence         show an agreement window before users can connect\n"
            "  -m, --htmlstatus      enable HTML status file, set file name\n"
            "  -o, --serverinfo      registration info for this Server.  Format:\n"
-           "                        [name];[city];[country as Qt5 QLocale ID]\n"
+           "                        [name];[city];[country as two-letter code or Qt5 QLocale ID]\n"
```
pljones, line 1078: Should we say "two letter ISO code"?  Otherwise people still might not know whether to use "UK" or "gb" or whatever.

src/util.cpp
```diff
@@ -1436,6 +1436,31 @@ bool CLocale::IsCountryCodeSupported ( unsigned short iCountryCode )
 #endif
 }
 
+QLocale::Country CLocale::GetCountryCodeByTwoLetterCode ( QString sTwoLetterCode )
+{
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 2, 0 )
+    return QLocale::codeToTerritory ( sTwoLetterCode );
```
pljones, line 1442: Relying on the compiler to optimise this "call" to inline?  Or have the .h declare inline the 6.2.0 version?

hoffie: This function is called once on server start. I would rely on the compiler to do its thing... :)

src/main.cpp
hoffie: Done.

pljones (approved):

src/serverlist.cpp
```diff
@@ -205,16 +205,36 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
...
+        }
+        else
+        {
+            QLocale::Country qlCountry = CLocale::GetCountryCodeByTwoLetterCode ( slServInfoSeparateParams[2] );
+            if ( qlCountry != QLocale::AnyCountry )
+            {
+                ThisServerListEntry.eCountry = qlCountry;
+            }
         }
+        qInfo() << qUtf8Printable ( QString ( "Using server info: name = \"%1\", city = \"%2\", country = \"%3\" (%4)" )
```
ann0see, line 229: As mentioned elsewhere by BLumina, this should be country/region as in the GUI.

ann0see (approved):
Besides my comment, happy to approve.

src/serverlist.cpp
hoffie: Done. Will merge on CI green.

hoffie:
Opened https://github.com/jamulussoftware/jamuluswebsite/issues/822 for the documentation. Therefore removing the needs-documentation label here. As stated, this is totally optional and doesn't have to be integrated at 3.9.1 release time.

ann0see:
As it's small, I've merged it.

Issue #2758, opened 2022-07-26
Using Weblate for translations


ann0see:
<div type='discussions-op-text'>

<sup>Originally posted by **ignotus666** July 26, 2022</sup>
@jamulussoftware/maindevelopers @jamulussoftware/translators 

In #2675 the use of [Weblate](https://weblate.org/en-gb/) for translations was proposed. It basically means moving the translation process to an online tool where translators can login, do their work, and need not worry about using GitHub, PRs, forks, etc. as Weblate can be configured to take care of that, removing what is often a big barrier. If you’ve never used GH before and are not technologically minded, just reading through the TRANSLATING.md file is enough to make you faint. With Weblate, that document could practically be reduced to “Log in to Weblate, click on a doc and translate it. The End”. If for whatever reason a translator prefers to continue using the current workflow, there is nothing stopping them – no one would be forced to use Weblate.

So far the response by translators to the proposal has been positive. However, such a move would involve some changes on the admin side, mainly:

- PRs would arrive by doc, not by author/language. For the app, this would mean all translations would go into the same PR (unless they are merged as they arrive, though this could still mean more than one language in the same PR).
- Language reviews should preferably happen on Weblate. They could theoretically take place on the PR here, but it'd be better if the language could be corrected before getting to that stage - and less complicated for translators.
- As @BLumia commented: 
> we'll need to pay more attention to the source strings that contains arguments (e.g. "About %1") and make use of the 2nd disambiguation argument in QObject::tr() to tell translator what's the argument means, since unlike using QtLinguist with full source code fetched, the translation platform won't show the related source code in the translation page.
- Preferably, the .nsi files should also be included, but that involves creating .po files for them. The question is, where should these .po files live and what process do we follow to reconvert them into .nsi files? There is already the po4a script 'infrastructure' for doing these conversions on the website repo. Could they live there and we push the converted files from there to the main repo?
- … possibly other stuff I can't think of now.

All in all, this would vastly reduce complexity for (in particular new) translators, but would require some changes, mainly on the main repo. For the website repo, aside from how PRs are submitted, it would be quite seamless. An option could be to first use it for the website docs and over time include the app stuff (and meanwhile explore in more depth the best way to go about it).

[...]

PR #2849, opened 2022-09-12
Translation: Add empty nb_NO translation files

ann0see:

Adds nb_NO translation to Jamulus as requested by @comradekingu

CHANGELOG: Translation: Add nb_NO App translation

**Context: Fixes an issue?**

Raised on weblate. 

Yes. Once the translation arrives the translator needs to be added to src/util.h

Ready for merge

/

## Checklist


ann0see:
We also need to add him to the org and multiple other places.

hoffie:
This would make nb_NO show up as a selectable language despite having no translations right now, correct?
Would it make sense to leave .qrc addition out of this PR and only add it once some amount of content has been translated?

comradekingu:
@hoffie 
I'll get right on it as soon as it is possible.
To require of people to do this, and add translations directly via Git makes the software look less polished than what is at worst
a selectable option with no translations.

nb_NO will be given by the locale if available anyway, the selection is secondary.

pljones:
I'm proposing 3.10.0 for this.

ann0see:
> I'm proposing 3.10.0 for this.

Why? I think translations are non blocking and having a translation of the most important parts is better than having none.

ann0see:
> This would make nb_NO show up as a selectable language despite having no translations right now, correct?

Yes.

hoffie:
I'd be ok with adding everything what is necessary to make nb_NO show up for translation on Weblate, but keeping the language unlisted as long as there are no translations at all.

Once there are translations, I'm fine with making it usable from within Jamulus. I'm also fine with this going into 3.9.1 depending on the timing. I guess we wouldn't delay 3.9.1 for that though.

([We use *Addition of translations which were too late to be included into the last release.*](https://jamulus.io/contribute/Administration) as one acceptable type of change for patch-level releases such as 3.9.1; while this was initially meant to target the addition of translation strings of existing translated languages, I see no reason to "block" completely new translations based on that)

pljones:
OK, seeing as it doesn't touch source and it can't break things, I guess there's no reason to delay it.

ann0see:
@comradekingu Could you please comment on https://github.com/jamulussoftware/jamulus/issues/2758 with your suggestions about The Weblate process? Since we're quite new to it, it's probably worth getting insight from you.

ann0see:
> Unless it is the Weblate bot, I could add myself if there is a "Translator credit" string wherein it is possible to also write down an e-mail, so people can reach out.

No. That's another step we need to do via git

src/resources.qrc
```diff
@@ -14,6 +14,9 @@
     <qresource prefix="/translations">
         <file alias="translation_es">translation/translation_es_ES.qm</file>
     </qresource>
+    <qresource prefix="/translations">
```
ann0see, line 17: So we'd remove this to not let the translation show up?

hoffie: I think this would work, yes (at least that was the effect when we forgot adding it for Korean, I think?).

So, yes, either that. Or go with the PR as-is and add a reminder to the 3.9.1 checklist to remove that part again if no translations have been integrated at release candidate time.

ann0see: > Or go with the PR as-is and add a reminder to the 3.9.1 checklist to remove that part again if no translations have been integrated at release candidate time

I'd go with that since I understood @comradekingu that he'll start translating right away.

hoffie (approved):
Approving based on the previous comments.
@ann0see Please add a reminder to #2813 wrt removal of nb_NO if no translations have been integrated at RC stage.

ann0see:
@pljones feel free to merge this.

pljones (approved):

PR #2893, opened 2022-10-01
audiomixerboard size_t clean-up

pljones:

Small patch to `audiomixerboard.cpp` to clean up warnings in Qt Creator.  Changes `int` to `size_t` where appropriate and handles any casts required.

CHANGELOG: Refactor: use `size_t` for vector and array indexes that must not be negative

**Context: Fixes an issue?**

Fixes warnings in Qt Creator in one file.  Warnings of a similar nature are also available.

No.

I use it in the client I run when jamming and it seems okay.

Needs a review.  

## Checklist

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -913,7 +913,7 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent ) :
 
     vecAvgLevels.Init ( MAX_NUM_CHANNELS, 0.0f );
 
-    for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
+    for ( size_t i = 0; i < MAX_NUM_CHANNELS; i++ )
     {
         vecpChanFader[i] = new CChannelFader ( this );
         vecpChanFader[i]->Hide();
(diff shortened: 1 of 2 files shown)


ann0see (commented):
Is there any valid reason why size_t is prefered by Qt? I doubt we get over the positive integer range anywhere.

ann0see (approved):
Approving based on reading b

hoffie (approved):
I don't see anything wrong here, but I can't say that I'm able to assess that everything is right... ;)

> Fixes warnings in Qt Creator in one file. Warnings of a similar nature are also available.

What are those warnings?

pljones:
> What are those warnings?

I wish cut-n-paste from the warnings was simple.  They seemed to be Qt IDE rather than Compiler warnings (so the compile output didn't show them).  Whilst correct (i.e. size_t can hold a larger number than int as it's unsigned), the int usage is also correct in places (we use -1 as an invalid index into an array or list, which has an index _type_ of size_t, so doing something like
```
void methodName(SomeArrayType someArray) {
  int i = sizeof(someArray) - 1;
  if (i == INVALID_INDEX) {
    return;
  }
  ...
}
```
will get a warning, as someArray _could_ be so big that even taking one off the size is still too large a _positive_ number for the int to hold.  In practice, we generally use no extent over about 256 (sound card buffer size is probably the worst - maybe 4K? - still well under the number an int can hold).

Issue #2528, opened 2022-03-19
Qt6: Android compatibility


hoffie:
**What is the current behaviour and why should it be changed?**

Jamulus will fail to build with Qt6 even when #2299 is merged as it uses androidextras for permission handling, which is no longer available.

**Describe possible approaches**

1. Stay on unsupported Qt 5.15.2
2. Use the private implementation when building against Qt6 as described in the relevant [Qt ticket](https://bugreports.qt.io/browse/QTBUG-84382?focusedCommentId=566019&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-566019)

I'd go with (2), although it is not nice either.

**Has this feature been discussed and generally agreed?**

No.

pljones:
I've no desire to move off 5.15.2.

ann0see:
There's probably mostly a need for other OS which have potential (security) issues.

pljones:
Moving to 3.9.1 along with the other Android ticket.

pljones:
Do we have a deadline for completing this?  As it's not labelled as such, I'm moving it as notabug.

hoffie:
No deadline -- it's done when it's done. Fine to move.

hoffie:
Lot's of relevant changes can be taken from #2909 (contributed by @danryu).

pljones:
Moving out of 3.10.0 and back to Triage until someone can pick this up.

pljones:
Also, for reference
https://doc.qt.io/qt-6/extras-changes-qt6.html#changes-to-qt-android-extras

pljones:
https://github.com/jamulussoftware/jamulus/issues/2907
And bringing this back into 3.11.0, too.

pljones:
Moved out again.

PR #2909, opened 2022-10-12
Draft: Android CI production build and Play Store auto-submission

danryu:
## Description

This PR implements a best-practices CI build for submission of the client to Google Play Store, following the advice at 
https://doc.qt.io/qt-6/android-publishing-to-googleplay.html

This includes additional small changes for recording permission with Qt6 on Android, and does not support Qt5.

Note: I have found from experience that it is better to use the Github-image-provided Android SDK/NDK  tools, rather than create custom logic to do custom installations of those. See the notes in `android.sh`.

This build creates 4 x separate .aab files (one for each available Android ABI) and submits them as part of the same package to the Play Store, where the developer can then add it to a Testing track as they please.

CHANGELOG:  Build Android release for Play Store and auto-submit

## Documentation

Requires additional GITHUB SECRETS:

`${{ secrets.GOOGLE_SERVICE_ACCOUNT_JSON }}`
- Google Play - Service Account JSON setup:
  - Google Play Console -> Setup -> API Access -> Create/Link Google Cloud Project 
  - Google Cloud console -> IAM & Admin -> Service Accounts -> Create (Wizard). Then create JSON key and export/save

`${{ secrets.GOOGLE_KEYSTORE }}`
`${{ secrets.GOOGLE_KEYSTORE_PASS }}`
- Google Play - Keystore setup - refer to https://doc.qt.io/qt-6/android-publishing-to-googleplay.html#building-the-app

## Checklist

### DIFF
--- a/.github/autobuild/android.sh
+++ b/.github/autobuild/android.sh
@@ -52,15 +52,18 @@ setup_qt() {
         # icu needs explicit installation 
         # otherwise: "qmake: error while loading shared libraries: libicui18n.so.56: cannot open shared object file: No such file or directory"
         python3 -m aqt install-qt --outputdir "${QT_BASEDIR}" linux desktop "${QT_VERSION}" \
-            --archives qtbase qtdeclarative qtsvg qttools icu
+            --archives qtbase qtdeclarative qtsvg qttools icu \
+            --modules qtmultimedia
 
         # - 64bit required for Play Store
         python3 -m aqt install-qt --outputdir "${QT_BASEDIR}" linux android "${QT_VERSION}" android_arm64_v8a \
-            --archives qtbase qtdeclarative qtsvg qttools
+            --archives qtbase qtdeclarative qtsvg qttools \
+            --modules qtmultimedia
         
         # Also install for arm_v7 to build for 32bit devices
         python3 -m aqt install-qt --outputdir "${QT_BASEDIR}" linux android "${QT_VERSION}" android_armv7 \
-            --archives qtbase qtdeclarative qtsvg qttools
+            --archives qtbase qtdeclarative qtsvg qttools \
+            --modules qtmultimedia
         
     fi
 }
--- a/.github/autobuild/android.sh
+++ b/.github/autobuild/android.sh
@@ -23,6 +23,7 @@ QT_VERSION=6.3.2
 QT_BASEDIR="/opt/Qt"
 BUILD_DIR=build
 ANDROID_NDK_HOST="linux-x86_64"
+# Use Github CI image-provided env var for ANDROID_NDK_ROOT, but explicitly re-assign to placate shellcheck
 ANDROID_NDK_ROOT="${ANDROID_NDK_ROOT}"
 # Only variables which are really needed by sub-commands are exported.
 export JAVA_HOME=${JAVA_HOME_11_X64}
--- a/.github/autobuild/android.sh
+++ b/.github/autobuild/android.sh
@@ -39,7 +39,7 @@ setup_ubuntu_dependencies() {
 
     sudo apt-get -qq update
     sudo apt-get -qq --no-install-recommends -y install \
-        build-essential zip unzip bzip2 p7zip-full curl chrpath openjdk-11-jdk-headless
+        build-essential zip unzip bzip2 p7zip-full curl chrpath
 }
 
 
--- a/.github/autobuild/android.sh
+++ b/.github/autobuild/android.sh
@@ -77,15 +77,24 @@ build_app() {
     echo "${GOOGLE_RELEASE_KEYSTORE}" | base64 --decode > android/android_release.keystore
 
     echo ">>> Compiling for ${ARCH_ABI} ..."
-    # if ARCH_ABI=android_armv7 we need to override ANDROID_ABIS for qmake 
+
+    # Override ANDROID_ABIS according to build target 
     # note: seems ANDROID_ABIS can be set here at cmdline, but ANDROID_VERSION_CODE cannot - must be in qmake file
     if [ "${ARCH_ABI}" == "android_armv7" ]; then
         echo ">>> Running qmake with ANDROID_ABIS=armeabi-v7a ..."
         ANDROID_ABIS=armeabi-v7a \
             "${QT_BASEDIR}/${QT_VERSION}/${ARCH_ABI}/bin/qmake" -spec android-clang
-    else
+    elif [ "${ARCH_ABI}" == "android_arm64_v8a" ]; then
         echo ">>> Running qmake with ANDROID_ABIS=arm64-v8a ..."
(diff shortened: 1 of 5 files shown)


pljones:
Hi @danryu,

Thanks for your PR.  Could you review the check failures and fix them, please.

-- Peter

danryu:
Pushed all the fixes that I can see.
This build is now set for Qt 6.3.2 and will fail with Qt5.
I've also updated the recording permission handling to work with Qt6.
From personal experience I highly recommend 6.3.2+ for Android.

.github/autobuild/android.sh
```diff
@@ -23,6 +23,7 @@ QT_VERSION=6.3.2
 QT_BASEDIR="/opt/Qt"
 BUILD_DIR=build
 ANDROID_NDK_HOST="linux-x86_64"
+ANDROID_NDK_ROOT="${ANDROID_NDK_ROOT}"
```
pljones, line 26: Why do this?

danryu: Purely to get past the shellcheck error. It's otherwise totally unnecessary.

pljones: @hoffie any idea why it's complaining?

danryu: It complains because shellcheck is somewhat brittle and expects all environment variables to be explicitly assigned a value within the context of the script. 
However it is already set by the Github CI, so I re-assign the value in the script to side-step the error. That's it.

Jamulus.pro
```diff
@@ -229,18 +229,42 @@ win32 {
...
+    # !defined(ANDROID_ABIS, var):ANDROID_ABIS = arm64-v8a
+
+    # Experiments show likely better results with Android 10+ devices
+    ANDROID_MIN_SDK_VERSION = 29
+    ANDROID_TARGET_SDK_VERSION = 32
     ANDROID_VERSION_NAME = $$VERSION
-    ANDROID_VERSION_CODE = $$system(git log --oneline | wc -l)
+
+    # date-based unique value - prevent clashes in Play Store build numbering
+    !defined(ANDROID_VERSION_CODE, var):ANDROID_VERSION_CODE = $$system(date +%s | cut -c 2-)
```
ann0see, line 247: This will not work on windows I suppose? Wouldn't the git commit hash or something else be better?

Jamulus.pro
```diff
@@ -229,18 +229,42 @@ win32 {
     LIBS += -framework AVFoundation \
         -framework AudioToolbox
 } else:android {
-    ANDROID_ABIS = armeabi-v7a arm64-v8a x86 x86_64
+    # ANDROID_ABIS = armeabi-v7a arm64-v8a x86 x86_64
+    # note: only armeabi-v7a arm64-v8a are targeted, others are dead/deprecated
```
ann0see, line 233: x86_64 should be supported.

.github/autobuild/android.sh
danryu: Added explanatory comment for ANDROID_NDK_ROOT.
Also pushed some clang-format fixes.

ann0see (commented):
Thank you very much

Jamulus.pro
danryu: The CI is obviously targeted purely for Linux.
For local building on Windows we can just hardcode any int value.
`ANDROID_VERSION_CODE = 123456`

Android Version Code requires an integer <=2100000000.
So you'd need some custom logic to convert the shorthash to int first.

I don't like relying on git log because you need a separate version code for armeabi7 and armv8 builds.
The above `date` output guarantees a unique, valid value for each Play Store submission.

Jamulus.pro
danryu: Why do you need x86_64? There are almost zero x86_64 Android devices.
This CI is to build a package for submission to the Play Store, for installation on real-world devices.

And even for local building and testing, I have found it infinitely superior to test with armv8 images in the simulator, or to connect Android devices directly - it reduces the error "surface area" a lot.

danryu: Ok, I've thought of a good possible reason to build x86_84 ... :)

The Play Store submission will also work on Chromebooks - even if Intel-based (this is tested and working). But that will involve ARM->Intel translation which will likely slow performance.

I'll look at adding x86_64 support in ASAP.

danryu: In fact, since Chromebook support is basically "out-of-the-box" with the Play Store submission process, there's an argument for including plain old x86 as well - at least according to https://developer.android.com/topic/arc/device-support ...

ann0see: Yes, and also WSA (at least for testing purposes). I still own an Intel powered tablet...

Jamulus.pro
ann0see: > The CI is obviously targeted purely for Linux.

But as it's in Jamulus.pro we should ensure cross compatibility. Even if it's a noop.

Jamulus.pro
danryu: Good point on WSA!
I've now updated the build to target all the 4 x ABIs (including x86 and x86_64), and all are submitted to Play Store as part of the same release. (Tested and working - see the Koord app in Play Store for example.)

danryu:
If the additional Github Secrets are added according to the updated Documentation section in this PR, the Auto-Build should complete successfully.

There are some Coding Style failures and I'm not sure how valid or useful the remaining shfmt errors are - unfortunately I don't have more time to devote to this PR.

If desired, you can merge the changes to a future feature branch and I can close this PR.

ann0see:
Ok. Squashed and merged to a feature branch: https://github.com/jamulussoftware/jamulus/tree/feature/koord/CIPlaystore. Hopefully, in future work can be continued.

Closing.

hoffie:
@danryu Thanks a lot of upstreaming this! I think this PR contains some very valuable contributions to update the build environment and enable Qt6 compatibility (#2528). Not sure if we should integrate the playstore-related parts right now, might still be a bit early with little to no testing happening in this area.

Sorry for not chiming in earlier (time is limited...). I certainly plan to integrate large parts of this and will submit new PR(s) with the relevant parts (with proper attribution of @danryu, of course).

danryu:
@hoffie Happy to contribute!

PS if you want to get a sense of the end result of this PR, a reasonable indicator is to try Koord, as installed from the Play Store.
While the Koord app client build differs visually/presentation-wise to Jamulus, the core business logic is basically untouched.
By the way, once the transition to QML is done, mobile UI stuff will of course get a lot easier.

pljones:
For clarity, we're looking at splitting this into:
- Qt6 for Android
- Google Play Store for Android

Currently in progress is https://github.com/jamulussoftware/jamulus/issues/2528

ann0see:
I think a lot of work is already done on the feature branch. I'd personally focus on iOS as I don't own an up to date Android device.

Issue #2773, opened 2022-08-08
Build: Update macOS runner (legacy) (macos-10.15)


ann0see:
**Describe the bug**

GitHub is deprecating the use of macOS Catalina: https://github.com/actions/virtual-environments/issues/5583

We'll need to update soon to still support the legacy build. 

**To Reproduce**

/

**Expected behavior**

/

**Screenshots**

/

**Operating system**
macOS

hoffie:
We currently use Xcode 12.1.1 for the legacy build. This version is no longer available on macos-11 images.
We either need to [downgrade to 11.7 or upgrade to 12.4](https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md).

Based on the comments, we can't upgrade to [12.4 as it contains SDK 11.x](https://xcodereleases.com/) which is not supported by Qt5:
https://github.com/jamulussoftware/jamulus/blob/340ea33649b0b102e71dcb31a19bd40c5a21bede/.github/workflows/autobuild.yml#L157-L161
Therefore, it seems like the only valid options would be downgrading to 11.7 even if it sounds totally wrong.

hoffie:
Started a build here:
https://github.com/hoffie/jamulus/runs/7814643184?check_suite_focus=true

https://github.com/hoffie/jamulus/pull/new/autobuild-macos-11-legacy

ann0see:
The question is if not supported = produces broken build or just untested.

hoffie:
This issue is almost certainly going to affect us this year. We should somehow figure out how to continue with legacy builds.
1. Do we now which macOS versions require the legacy version at all?
2. Are those macOS versions still supported?
3. Do we know of any users who use such software?

I'd propose to ask at least the latter question in the Discussion section (maybe ping some known Mac users). If they are still on those old releases and have good reasons, we might keep it. Otherwise I'd argue for dropping support for that build.

@ann0see Could you do that?
Pinging @softins here as well in case he has opinions on that... :)

ann0see:
@hoffie macOS legacy is being used: https://somsubhra.github.io/github-release-stats/?username=jamulussoftware&repository=jamulus&page=1&per_page=5

Ofc. Not that many people in comparison to the normal Mac version, but still a considerable amount. 
I do know quite some (not so tech savvy) musicians who use outdated macs/PCs for music applications like my second choir leader.

ann0see:
[...]

PR #2917, opened 2022-10-17
JSON-RPC: Add option to specify bind address for server

Rob-NY:

Adds a command line switch to specify the bind address for the RPC server.  Today this address is hard-coded to the loopback interface.  The loopback remains the default, but can be overridden with this new command line option.

CHANGELOG:  New command line switch (--jsonrpcbindip) to specify the bind address for the RPC server.

**Context: Fixes an issue?**

Does not fix a reported issue, but does address that there is no way to change the bind address without recompiling.

If accepted, will update doc repo.

Tested on Linux only.

## Checklist

### DIFF
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -17,6 +17,9 @@ It can be generated like this:
 $ openssl rand -base64 10 > /file/with/a/secret.txt
 ```
 
+The JSON-RPC server defaults to listening on the local loopback network interface (127.0.0.1).  This can be optionally changed by using the `--jsonrpcbindip <ip address>` command line option. **IPv4 only. IPv6 support has not been tested.**
+
+
 ## Wire protocol
 
 The JSON-RPC server is based on the [JSON-RPC 2.0](https://www.jsonrpc.org/specification) protocol, using [streaming newline-delimited JSON over TCP](https://clue.engineering/2018/introducing-reactphp-ndjson) as the transport. There are three main types of messages being exchanged:
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -99,6 +99,7 @@ int main ( int argc, char** argv )
     int          iNumServerChannels          = DEFAULT_USED_NUM_CHANNELS;
     quint16      iPortNumber                 = DEFAULT_PORT_NUMBER;
     int          iJsonRpcPortNumber          = INVALID_PORT;
+    QString      strJsonRpcBindIP            = DEFAULT_JSON_RPC_LISTEN_ADDRESS;
     quint16      iQosNumber                  = DEFAULT_QOS_NUMBER;
     ELicenceType eLicenceType                = LT_NO_LICENCE;
     QString      strMIDISetup                = "";
@@ -206,6 +207,15 @@ int main ( int argc, char** argv )
             strJsonRpcSecretFileName = strArgument;
             qInfo() << qUtf8Printable ( QString ( "- JSON-RPC secret file: %1" ).arg ( strJsonRpcSecretFileName ) );
             CommandLineOptions << "--jsonrpcsecretfile";
+            continue;
+        }
+
+        // JSON-RPC bind address ------------------------------------------------
+        if ( GetStringArgument ( argc, argv, i, "--jsonrpcbindip", "--jsonrpcbindip", strArgument ) )
+        {
+            strJsonRpcBindIP = QString ( strArgument );
+            qInfo() << qUtf8Printable ( QString ( "- JSON-RPC will bind to: %1 if enabled" ).arg ( strJsonRpcBindIP ) );
+            CommandLineOptions << "--jsonrpcbindip";
             continue;
         }
 
@@ -783,6 +793,31 @@ int main ( int argc, char** argv )
                 strServerBindIP = "";
             }
         }
+
+#ifndef NO_JSON_RPC
+        //
+        // strJsonRpcBind address defaults to loopback and should not be empty, but
+        // in the odd chance that an empty IP is passed, we'll check for it here.
+
+        if ( strJsonRpcBindIP.trimmed().isEmpty() )
+        {
+            qCritical() << qUtf8Printable ( QString ( "JSON-RPC is enabled but no bind address was provided, exiting." ) );
+            exit ( 1 );
+        }
+
+        // This means of testing the validity of IP addresses is far from perfect but
+        // we do it here as an upfront check.  The downstream network calls will error
+        // out on malformed addresses not caught here.
+        {
+            QHostAddress InetAddr;
+            if ( !InetAddr.setAddress ( strJsonRpcBindIP ) )
+            {
+                qCritical() << qUtf8Printable ( QString ( "The JSON-RPC address specified is not valid, exiting. " ) );
+                exit ( 1 );
+            }
+        }
+
+#endif
     }
 
     // Application/GUI setup ---------------------------------------------------
@@ -880,7 +915,7 @@ int main ( int argc, char** argv )
         qWarning() << "- JSON-RPC: This interface is experimental and is subject to breaking changes even on patch versions "
                       "(not subject to semantic versioning) during the initial phase.";
 
-        pRpcServer = new CRpcServer ( pApp, iJsonRpcPortNumber, strJsonRpcSecret );
+        pRpcServer = new CRpcServer ( pApp, strJsonRpcBindIP, iJsonRpcPortNumber, strJsonRpcSecret );
         if ( !pRpcServer->Start() )
         {
             qCritical() << qUtf8Printable ( QString ( "- JSON-RPC: Server failed to start. Exiting." ) );
@@ -1072,6 +1107,7 @@ QString UsageArguments ( char** argv )
            "      --jsonrpcsecretfile\n"
            "                        path to a single-line file which contains a freely\n"
            "                        chosen secret to authenticate JSON-RPC users.\n"
+           "      --jsonrpcbindip   optional network address to bind RPC server. Defaults to 127.0.0.1 (IPv4 only, IPv6 not tested).\n"
            "  -Q, --qos             set the QoS value. Default is 128. Disable with 0\n"
            "                        (see the Jamulus website to enable QoS on Windows)\n"
            "  -t, --notranslation   disable translation (use English language)\n"
--- a/src/rpcserver.cpp
+++ b/src/rpcserver.cpp
@@ -26,8 +26,9 @@
 #include "global.h"
 #include "rpcserver.h"
 
-CRpcServer::CRpcServer ( QObject* parent, int iPort, QString strSecret ) :
+CRpcServer::CRpcServer ( QObject* parent, QString strBindIP, int iPort, QString strSecret ) :
     QObject ( parent ),
+    strBindIP ( strBindIP ),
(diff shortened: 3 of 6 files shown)


Rob-NY:
> This could also make `--jsonrpcport` redundant, if it supports a port too.

It does not support a port at this time.  Moreover, I did not change the fact that specifying the port is the **switch** that turns on/off RPC.  This is how it was before, just defaulting to the loopback.

src/main.cpp
```diff
@@ -191,6 +192,15 @@ int main ( int argc, char** argv )
             continue;
         }
 
+        // JSON-RPC bind address ------------------------------------------------
+        if ( GetStringArgument ( argc, argv, i, "--jsonrpcbind", "--jsonrpcbind", strArgument ) )
```
hoffie, line 196: Hmm, thinking of `--serverbindip` this should be `--jsonrpcbindip` for consistency, I guess?
(Other places would need to be adapted as well, of course)

hoffie (commented):
Minor nit, looks good to me in general and works for me. Thanks!

I know that we might not be consistent regarding this in the existing code base, but it might make sense to error out if `--jsonrpcbind(ip)` is given, but `--jsonrpcport` isn't (i.e., JSON-RPC is not enabled at all). Otherwise the option is ignored silently.

pljones:
Same approach as `--port` and `--serverbindip`.  One sets the listening port (but doesn't interfere with the listening address), the other sets the listening address (but doesn't touch the listening port).  Maybe have it as `--jsonrpcbindip` (i.e. with the `ip`) to match the server one.

hoffie:
> This could also make `--jsonrpcport` redundant, if it supports a port too.

I thought about it as well, but considering that it does break CLI flag compatibility (which would be ok, as this is still experimental), it would not be consistent with the existing flags for handling ports/bind addresses either (`--serverbindip`). Therefore, I'd be in favor of doing what this PR does already.

Rob-NY:
> Same approach as `--port` and `--serverbindip`. One sets the listening port (but doesn't interfere with the listening address), the other sets the listening address (but doesn't touch the listening port). Maybe have it as `--jsonrpcbindip` (i.e. with the `ip`) to match the server one.

Pushed an update to have the RPC IP and Port match the naming convention for server.

src/rpcserver.cpp
```diff
@@ -26,8 +26,9 @@
 #include "global.h"
 #include "rpcserver.h"
 
-CRpcServer::CRpcServer ( QObject* parent, int iPort, QString strSecret ) :
+CRpcServer::CRpcServer ( QObject* parent, QString strBindIP, int iPort, QString strSecret ) :
     QObject ( parent ),
+    strBindIP ( strBindIP ),
```
pljones, line 31: Can you validate this and issue an error if it's syntactically rubbish, please.

Rob-NY: The way legacy code has been "validating" IP's (primarily in main.cpp) is by creating an instance of QHostAddress and then using setAddress() with the candidate string.   However, while it catches range errors such as 300.200.100.50, it does NOT catch string-related issues like "192.168. 1.00" (note spaces), and these fall through to the downstream network calls where they eventually throw an error.   I can certainly add parity code in main.cpp to be consistent with other provided addresses, but it doesn't do anything that isn't also caught by the network calls later.

pljones: OK, I was pretty sure it wasn't too good, if it had any effect at all.  Not to worry then, I think.

pljones (approved):

hoffie:
I'm mostly happy with this PR, but unless someone opposes, I'd prefer to have the validation implemented which is mentioned here:

> I know that we might not be consistent regarding this in the existing code base, but it might make sense to error out if `--jsonrpcbind(ip)` is given, but `--jsonrpcport` isn't (i.e., JSON-RPC is not enabled at all). Otherwise the option is ignored silently.

Rob-NY:
> I'm mostly happy with this PR, but unless someone opposes, I'd prefer to have the validation implemented which is mentioned here:
> 
> > I know that we might not be consistent regarding this in the existing code base, but it might make sense to error out if `--jsonrpcbind(ip)` is given, but `--jsonrpcport` isn't (i.e., JSON-RPC is not enabled at all). Otherwise the option is ignored silently.

Pushed an update with additional error checking both for "empty" addresses and using QHostAddress as is done consistently elsewhere.  This should wrap up this PR, let me know if otherwise.

Rob-NY:
Docs updated in the code base (python doc-generation script) and regenerated the rpc md file.

src/main.cpp
```diff
@@ -783,6 +793,31 @@ int main ( int argc, char** argv )
...
+            exit ( 1 );
+        }
+        else
+        {
+            // This means of testing the validity of IP addresses is far from perfect but
+            // we do it here as an upfront check.  The downstream network calls will error
+            // out on malformed addresses not caught here.
+
+            QHostAddress InetAddr;
+            if ( !InetAddr.setAddress ( strJsonRpcBindIP ) )
```
ann0see, line 814: We should introduce a method which checks IP validity e.g via Regex

docs/JSON-RPC.md
```diff
@@ -17,6 +17,9 @@ It can be generated like this:
 $ openssl rand -base64 10 > /file/with/a/secret.txt
 ```
 
+The JSON-RPC server defaults to listening on the local loopback network interface (127.0.0.1).  This can be optionally changed by using the `--jsonrpcbindip <ip address>` command line option.
```
ann0see, line 20: Please clarify if IPv6 is also available.

ann0see (commented):
Thanks. Some minor comments

docs/JSON-RPC.md
Rob-NY: I personally cannot test IPv6 at this time.  @dtinth might want to comment, but RPC uses QTcpServer, which in turn gets its bind address from QHostAddress( string ) which, according to the docs, can be either IPv4 or 6.  So, in theory, this should work but I do not want to make that representation as I cannot test it.

src/main.cpp
Rob-NY: You're probably right -- an IP address validation function used through the codebase.  I think that is out of scope for this particular PR, however.

docs/JSON-RPC.md
dtinth: I’m not using IPv6 either, so I cannot say… May suffice to say that only IPv4 is supported for now.

Rob-NY: pushed updated docs

Rob-NY:
@ann0see  @hoffie  - I think something happened with the auto checks.  The Mac legacy "was cancelled".  Let me know if there's something I need to do.

ann0see:
Probably that's an outrage due to the old macOS version it's using

docs/JSON-RPC.md
ann0see: Ok. Tested ipv6 on Windows 11 (binding to ::1) and I could see a connection via ipv6 but not ipv4 as expected. So you can say that ipv6 is also ok.

src/main.cpp
```diff
@@ -783,6 +793,31 @@ int main ( int argc, char** argv )
...
+            {
+                qCritical() << qUtf8Printable ( QString ( "The JSON-RPC address specified is not valid, exiting. " ) );
+                exit ( 1 );
+            }
+        }
```
hoffie, line 819: We can avoid deep nesting here by dropping `else` (we'll never reach that code due to the `exit` before.

hoffie:
> > I'm mostly happy with this PR, but unless someone opposes, I'd prefer to have the validation implemented which is mentioned here:
> > > I know that we might not be consistent regarding this in the existing code base, but it might make sense to error out if `--jsonrpcbind(ip)` is given, but `--jsonrpcport` isn't (i.e., JSON-RPC is not enabled at all). Otherwise the option is ignored silently.
> 
> Pushed an update with additional error checking both for "empty" addresses and using QHostAddress as is done consistently elsewhere. This should wrap up this PR, let me know if otherwise.

I think there might have been a misunderstanding. We're now checking if the user has provided an empty IP argument. That's nice to have, but not what I meant with the above suggestion.

What I meant was:
```
$ ./Jamulus -s -n --jsonrpcbindip 127.0.0.1
[...]
- JSON-RPC binding to: 127.0.0.1
```

This might be confusing as it seems like JSON-RPC is enabled, while it isn't (as no `--jsonrpcport` was given, which controls to enabling).

I was suggesting something like:
```
$ ./Jamulus -s -n --jsonrpcbindip 127.0.0.1
[...]
Ignoring --jsonrpcbindip as no --jsonrpcport was given.
```

As this is a new feature, we could be even more harsh and log it as an error and `exit` to make such usage errors obvious (there is no risk of breaking anything at this stage).

> The Mac legacy "was cancelled".

I've restarted that job. It was indeed due to the MacOS 10.15 "brown out" phase by Github (see #2773).

src/main.cpp
Rob-NY: I'll refactor taking into account @pljones comment about the port being specified as well.

Rob-NY: FWIW, the reason this was in an if..else block was to manage the variable scope for InetAddr.  Even if I pull it out and make the "else" its own block, I need to (or should) bracket it to protect that scope.

hoffie (approved):
@Rob-NY Fair points, I'm fine with that. Thanks again for this PR!

src/main.cpp
```diff
@@ -783,6 +793,31 @@ int main ( int argc, char** argv )
...
         }
+
+#ifndef NO_JSON_RPC
+        //
+        // strJsonRpcBind address defaults to loopback and should not be empty, but
+        // in the odd chance that an empty IP is passed, we'll check for it here.
+
+        if ( strJsonRpcBindIP.trimmed().isEmpty() )
+        {
+            qCritical() << qUtf8Printable ( QString ( "JSON-RPC is enabled but no bind address was provided, exiting." ) );
```
pljones, line 804: I think the wording of the "objection" is slightly misleading here.  `--serverbindip` was passed with a parameter, otherwise `GetStringArgument` would have failed earlier.  So
```suggestion
            qCritical() << qUtf8Printable ( QString ( "JSON-RPC is enabled but bind address provided is invalid, exiting." ) );
```
It might even be useful to do `if ( !InetAddr.setAddress ( strJsonRpcBindIP ) )`, like for Server Bind IP.

I'd also question whether "exiting" is in keeping with the other checks we do... it's a bit hit and miss, I guess.  I may even have said do it like this :).

src/main.cpp
pljones: I think using the same methods as for Server Bind IP will be okay for now.

src/main.cpp
hoffie: > I think the wording of the "objection" is slightly misleading here.

I'll squeeze that improvement into #2958 and rename it to be more generic.

> I'd also question whether "exiting" is in keeping with the other checks we do... it's a bit hit and miss, I guess. I may even have said do it like this :).

I would have certainly supported that (maybe I did ;)). I think we should fail early for new features, even if we didn't do so previously and don't want to make those existing checks more strict for compatibility reasons for now.

ann0see:
Dropped needs documentation as https://github.com/jamulussoftware/jamuluswebsite/pull/871 was merged a while ago.

Issue #2488, opened 2022-03-10
JSON-RPC - Extend jamulusserver/getClients to include additional client attributes


Rob-NY:
**What is the current behaviour and why should it be changed?**

The current getClients RPC method does not return skill level, country, city, or instrument for connected clients.  In order to close the gap between desired functionality in (now closed) PR #2006, getClients can be extended to provide these additional properties.

**Describe possible approaches**

@dtinth had as specific goal of not touching server.cpp when initially authoring JSON-RPC.  This limited the information available to the RPC interface.  The most direct solution I see now is to modify GetConCliParam() to add an additional vector that can be populated with Channel Information (core).   **GetConCliParam() is already used by JSON-RPC within the getClients method.**
```
void CServer::GetConCliParam ( CVector<CHostAddress>& vecHostAddresses,
                               CVector<QString>&      vecsName,
                               CVector<int>&          veciJitBufNumFrames,
                               CVector<int>&          veciNetwFrameSizeFact,
                               CVector<CChannelCoreInfo>& vecChanInfo )
```
Changing this method impacts server.cpp/h and serverdlg.cpp only.    While serverdlg would not make use of this additional information, having connection attributes such as country, city, instrument, and skill would allow the serverdlg to make use of this information in the future within the UI's "connections" widget.

The modified GetConCliParam method proposed is:
```
void CServer::GetConCliParam ( CVector<CHostAddress>& vecHostAddresses,
                               CVector<QString>&      vecsName,
                               CVector<int>&          veciJitBufNumFrames,
                               CVector<int>&          veciNetwFrameSizeFact,
                               CVector<CChannelCoreInfo>& vecChanInfo )
{
    // init return values
    vecHostAddresses.Init ( iMaxNumChannels );
    vecsName.Init ( iMaxNumChannels );
    veciJitBufNumFrames.Init ( iMaxNumChannels );
    veciNetwFrameSizeFact.Init ( iMaxNumChannels );
    vecChanInfo.Init ( iMaxNumChannels );

    // check all possible channels
    for ( int i = 0; i < iMaxNumChannels; i++ )
    {
        if ( vecChannels[i].IsConnected() )
        {
            // get requested data
            vecHostAddresses[i]      = vecChannels[i].GetAddress();
            vecsName[i]              = vecChannels[i].GetName();
            veciJitBufNumFrames[i]   = vecChannels[i].GetSockBufNumFrames();
[...]

PR #2918, opened 2022-10-17
JSON-RPC: Extend jamulusserver/getClients method

Rob-NY:

Expands jamulusserver/getClients method to include all elements of a user profile as well as the number of active client connections.  Changes were made so they are backward compatible with the prior message format.

CHANGELOG:  jamulusserver/getClients method expanded to include all elements of a user profile as well as the number of active client connections.  

**Context: Fixes an issue?**

Addresses issue #2488 and comments in #2890 

JSON-RPC documentation was updated via python tool and is included in this PR.

Compile tested on Linux only.

## Checklist

### DIFF
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -234,12 +234,17 @@ Results:
 
 | Name | Type | Description |
 | --- | --- | --- |
+| result.connections | number | The number of active connections. |
 | result.clients | array | The list of connected clients. |
 | result.clients[*].id | number | The client’s channel id. |
 | result.clients[*].address | string | The client’s address (ip:port). |
 | result.clients[*].name | string | The client’s name. |
 | result.clients[*].jitterBufferSize | number | The client’s jitter buffer size. |
 | result.clients[*].channels | number | The number of audio channels of the client. |
+| result.clients[*].instrumentCode | number | The id of the instrument for this channel. |
+| result.clients[*].city | string | The city name provided by the user for this channel. |
+| result.clients[*].countryName | number | The text name of the country specified by the user for this channel (see QLocale::Country). |
+| result.clients[*].skillLevelCode | number | The skill level id provided by the user for this channel. |
 
 
 ### jamulusserver/getRecorderStatus
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1469,16 +1469,18 @@ bool CServer::PutAudioData ( const CVector<uint8_t>& vecbyRecBuf, const int iNum
     return bNewConnection;
 }
 
-void CServer::GetConCliParam ( CVector<CHostAddress>& vecHostAddresses,
-                               CVector<QString>&      vecsName,
-                               CVector<int>&          veciJitBufNumFrames,
-                               CVector<int>&          veciNetwFrameSizeFact )
+void CServer::GetConCliParam ( CVector<CHostAddress>&     vecHostAddresses,
+                               CVector<QString>&          vecsName,
+                               CVector<int>&              veciJitBufNumFrames,
+                               CVector<int>&              veciNetwFrameSizeFact,
+                               CVector<CChannelCoreInfo>& vecChanInfo )
 {
     // init return values
     vecHostAddresses.Init ( iMaxNumChannels );
     vecsName.Init ( iMaxNumChannels );
     veciJitBufNumFrames.Init ( iMaxNumChannels );
     veciNetwFrameSizeFact.Init ( iMaxNumChannels );
+    vecChanInfo.Init ( iMaxNumChannels );
 
     // check all possible channels
     for ( int i = 0; i < iMaxNumChannels; i++ )
@@ -1490,6 +1492,7 @@ void CServer::GetConCliParam ( CVector<CHostAddress>& vecHostAddresses,
             vecsName[i]              = vecChannels[i].GetName();
             veciJitBufNumFrames[i]   = vecChannels[i].GetSockBufNumFrames();
             veciNetwFrameSizeFact[i] = vecChannels[i].GetNetwFrameSizeFact();
+            vecChanInfo[i]           = vecChannels[i].GetChanInfo();
         }
     }
 }
--- a/src/serverdlg.cpp
+++ b/src/serverdlg.cpp
@@ -679,14 +679,15 @@ void CServerDlg::OnCLVersionAndOSReceived ( CHostAddress, COSUtil::EOpSystemType
 
 void CServerDlg::OnTimer()
 {
-    CVector<CHostAddress> vecHostAddresses;
-    CVector<QString>      vecsName;
-    CVector<int>          veciJitBufNumFrames;
-    CVector<int>          veciNetwFrameSizeFact;
+    CVector<CHostAddress>     vecHostAddresses;
+    CVector<QString>          vecsName;
+    CVector<int>              veciJitBufNumFrames;
+    CVector<int>              veciNetwFrameSizeFact;
+    CVector<CChannelCoreInfo> vecChanInfo;
 
     ListViewMutex.lock();
     {
-        pServer->GetConCliParam ( vecHostAddresses, vecsName, veciJitBufNumFrames, veciNetwFrameSizeFact );
+        pServer->GetConCliParam ( vecHostAddresses, vecsName, veciJitBufNumFrames, veciNetwFrameSizeFact, vecChanInfo );
 
         // we assume that all vectors have the same length
         const int iNumChannels = vecHostAddresses.Size();
--- a/src/serverrpc.cpp
+++ b/src/serverrpc.cpp
@@ -56,20 +56,28 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
     /// @rpc_method jamulusserver/getClients
     /// @brief Returns the list of connected clients along with details about them.
     /// @param {object} params - No parameters (empty object).
-    /// @result {array} result.clients - The list of connected clients.
+    /// @result {number} result.connections - The number of active connections.
+    /// @result {array}  result.clients - The list of connected clients.
     /// @result {number} result.clients[*].id - The client’s channel id.
     /// @result {string} result.clients[*].address - The client’s address (ip:port).
     /// @result {string} result.clients[*].name - The client’s name.
     /// @result {number} result.clients[*].jitterBufferSize - The client’s jitter buffer size.
     /// @result {number} result.clients[*].channels - The number of audio channels of the client.
+    /// @result {number} result.clients[*].instrumentCode - The id of the instrument for this channel.
+    /// @result {string} result.clients[*].city - The city name provided by the user for this channel.
+    /// @result {number} result.clients[*].countryName - The text name of the country specified by the user for this channel (see QLocale::Country).
+    /// @result {number} result.clients[*].skillLevelCode - The skill level id provided by the user for this channel.
     pRpcServer->HandleMethod ( "jamulusserver/getClients", [=] ( const QJsonObject& params, QJsonObject& response ) {
-        QJsonArray            clients;
-        CVector<CHostAddress> vecHostAddresses;
-        CVector<QString>      vecsName;
-        CVector<int>          veciJitBufNumFrames;
-        CVector<int>          veciNetwFrameSizeFact;
+        QJsonArray                clients;
+        CVector<CHostAddress>     vecHostAddresses;
+        CVector<QString>          vecsName;
+        CVector<int>              veciJitBufNumFrames;
+        CVector<int>              veciNetwFrameSizeFact;
+        CVector<CChannelCoreInfo> vecChanInfo;
 
-        pServer->GetConCliParam ( vecHostAddresses, vecsName, veciJitBufNumFrames, veciNetwFrameSizeFact );
+        int connections = 0;
+
+        pServer->GetConCliParam ( vecHostAddresses, vecsName, veciJitBufNumFrames, veciNetwFrameSizeFact, vecChanInfo );
 
         // we assume that all vectors have the same length
         const int iNumChannels = vecHostAddresses.Size();
@@ -81,18 +89,26 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
             {
                 continue;
             }
+
             QJsonObject client{
                 { "id", i },
                 { "address", vecHostAddresses[i].toString ( CHostAddress::SM_IP_PORT ) },
                 { "name", vecsName[i] },
                 { "jitterBufferSize", veciJitBufNumFrames[i] },
                 { "channels", pServer->GetClientNumAudioChannels ( i ) },
+                { "instrumentCode", vecChanInfo[i].iInstrument },
+                { "city", vecChanInfo[i].strCity },
+                { "countryName", QLocale::countryToString ( vecChanInfo[i].eCountry ) },
+                { "skillLevelCode", vecChanInfo[i].eSkillLevel },
             };
             clients.append ( client );
+
+            ++connections;
         }
 
         // create result object
         QJsonObject result{
+            { "connections", connections },
             { "clients", clients },
         };
         response["result"] = result;
--- a/src/server.h
+++ b/src/server.h
@@ -116,10 +116,11 @@ public:
 
     int GetNumberOfConnectedClients();
 
-    void GetConCliParam ( CVector<CHostAddress>& vecHostAddresses,
-                          CVector<QString>&      vecsName,
-                          CVector<int>&          veciJitBufNumFrames,
-                          CVector<int>&          veciNetwFrameSizeFact );
+    void GetConCliParam ( CVector<CHostAddress>&     vecHostAddresses,
+                          CVector<QString>&          vecsName,
+                          CVector<int>&              veciJitBufNumFrames,
+                          CVector<int>&              veciNetwFrameSizeFact,
+                          CVector<CChannelCoreInfo>& vecChanInfo );
 
     void CreateCLServerListReqVerAndOSMes ( const CHostAddress& InetAddr ) { ConnLessProtocol.CreateCLReqVersionAndOSMes ( InetAddr ); }
 


src/util.h
```diff
@@ -677,6 +677,27 @@ enum ESkillLevel
     SL_PROFESSIONAL = 3
 };
 
+inline QString SkillLevelToString ( const int iSkill )
```
pljones, line 680: Why is  this bound to `"CServerDlg"`?  Why not just re-use the existing look-up method?

pljones: As a reminder - **_NO_** functionality - and _this_ counts as functionality - should exist "in the UI".  So explicitly making a second copy of something that should not exist really is bad...

pljones: Ideally `server.cpp` would have a method to do the lookup that delegated to a shared utility method also used by `client.cpp`... although, in both cases, likely delegated via `channel.cpp`.

Rob-NY: > Why is this bound to `"CServerDlg"`? Why not just re-use the existing look-up method?

Perhaps I missed where this exists today... I'm happy to change it.

src/serverrpc.cpp
```diff
@@ -81,18 +90,27 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
...
+                { "countryName", QLocale::countryToString ( vecChanInfo[i].eCountry ) },
+                { "skillLevelCode", vecChanInfo[i].eSkillLevel },
+                { "skillLevelName", SkillLevelToString ( vecChanInfo[i].eSkillLevel ) },
             };
             clients.append ( client );
         }
 
         // create result object
         QJsonObject result{
+            { "connections", pServer->GetNumberOfConnectedClients() },
```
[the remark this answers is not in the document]
Rob-NY: I believe vecHostAdddresses.Size() returns the number of slots, not the number of active connections.

src/util.h
Rob-NY: > As a reminder - **_NO_** functionality - and _this_ counts as functionality - should exist "in the UI". So explicitly making a second copy of something that should not exist really is bad...

What is in the UI that you object to?   The change to CServerDlg is required because a method prototype was changed in CServer that is called **from** CServerDlg.   

And, again, if I missed an existing method for converting skill number to skill literal I will happily refactor -- but I didn't find one (perhaps I missed it?)

pljones: UI is any user interface component.  That means the graphical user interface, the RPC interface and MIDI controls for the Jamulus core code.

pljones: > And, again, if I missed an existing method for converting skill number to skill literal I will happily refactor -- but I didn't find one (perhaps I missed it?)

Like I say, there's a _lot_ of code in the UI that shouldn't be there.  Currently the mapping is done _literally_ in the UI: src/clientsettingsdlg.cpp lines 611 to 629.

I'm happy adding the method here, as you've done.  The first argument in the `QCoreApplication::translate` call (`"CServerDlg"` or whatever) doesn't seem used - I can't see anything in the *.ts files referencing it.  The *.ts files _do_ reference the containing file (`util.h` here).  I'd rather it just said "util", and we moved away from trying to guess where utility methods got called from...

Rob-NY: Sorry, but I do not know how to respond to this.  I don't see where there is "functionality" in the RPC code that is any different from the first release of RPC.  It merely calls the relevant methods elsewhere (just as the first version of RPC did).   And if you're only referring to skill code translation, even that is in util.* where other core code is -- not in the RPC code.   So I'm sorry @pljones, I'm not following your comments enough to know how to effect a change.

src/server.cpp
```diff
@@ -1469,16 +1469,18 @@ bool CServer::PutAudioData ( const CVector<uint8_t>& vecbyRecBuf, const int iNum
...
+                               CVector<int>&              veciJitBufNumFrames,
+                               CVector<int>&              veciNetwFrameSizeFact,
+                               CVector<CChannelCoreInfo>& vecChanInfo )
 {
     // init return values
     vecHostAddresses.Init ( iMaxNumChannels );
     vecsName.Init ( iMaxNumChannels );
     veciJitBufNumFrames.Init ( iMaxNumChannels );
     veciNetwFrameSizeFact.Init ( iMaxNumChannels );
+    vecChanInfo.Init ( iMaxNumChannels );
```
pljones, line 1483: This isn't used by `server.cpp` then?  Why set it here?

src/util.h
pljones: `src/clientsettingsdlg.cpp` isn't the RPC code.  You're duplicating code here that's literally in the graphical user interface.  It shouldn't be.  That's what I'm complaining about.  The trigger for my complaint was the mention of the Service Graphical User Interface.  Which is irrelevant to what you're doing and hence _wrong_ and needs changing.

I've said that this is the _right_ place to do it.  Ideally the code in `src/clientsettingsdlg.cpp` would get fixed to call this method, too, and the translation would only happen in one place, consistently.

hoffie: Heh, I was confused about the subject of complaint as well as the primary indicator for the comment thread is... well, the new code, which was *not* the subject. ;)

src/serverrpc.cpp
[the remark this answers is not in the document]
pljones: But I see this at line 86:
```
// fill list with connected clients
for ( int i = 0; i < iNumChannels; i++ )
```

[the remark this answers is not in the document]
pljones: Oh and at line 60
```
/// @result {array} result.clients - The list of connected clients.
```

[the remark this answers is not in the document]
pljones: Rather than call GetNumberOfConnectedClients, having actively filtered for the connected clients, just keep your own local count and use it.  You know you'll get the right number, that way.  Nothing will have changed between one call and the next.

src/serverdlg.cpp
```diff
@@ -682,14 +682,15 @@ void CServerDlg::OnCLVersionAndOSReceived ( CHostAddress, COSUtil::EOpSystemType
...
+    CVector<CHostAddress>     vecHostAddresses;
+    CVector<QString>          vecsName;
+    CVector<int>              veciJitBufNumFrames;
+    CVector<int>              veciNetwFrameSizeFact;
+    CVector<CChannelCoreInfo> vecChanInfo;
 
     ListViewMutex.lock();
     {
-        pServer->GetConCliParam ( vecHostAddresses, vecsName, veciJitBufNumFrames, veciNetwFrameSizeFact );
+        pServer->GetConCliParam ( vecHostAddresses, vecsName, veciJitBufNumFrames, veciNetwFrameSizeFact, vecChanInfo );
```
pljones, line 693: What does the Graphical user interface want to do this for?

src/util.h
```diff
@@ -677,6 +677,27 @@ enum ESkillLevel
...
+    case SL_INTERMEDIATE:
+        return QCoreApplication::translate ( "CServerDlg", "Intermediate" );
+        break;
+
+    case SL_PROFESSIONAL:
+        return QCoreApplication::translate ( "CServerDlg", "Expert" );
+        break;
+
+    default:
+        return QString ( "" );
```
pljones, line 697: There shouldn't be a default - all the options should be covered and "None" should be returned for SL_NOT_SET

src/serverdlg.cpp
Rob-NY: While not my code, the return values of GetConCliParam are used to populate the connection box at the top of the server dialog.  This method call was already here;  I just added the additional parameter.   This _exact_ discussion was had in another thread that I cannot find now -- but will keep looking.

pljones: I'm not keen on unused variables (neither is the IDE).  If there are two different use cases, there should be two supporting methods.  One might call the other with a dummy parameter - or pass a dummy value here and ignore it in `GetConCliParam`.  At the moment, extra work is being done unnecessarily.

src/util.h
Rob-NY: > The trigger for my complaint was the mention of the Server Graphical User Interface. Which is irrelevant to what you're doing and hence _wrong_ and needs changing.

we covered this here: https://github.com/jamulussoftware/jamulus/issues/2488#issuecomment-1065865193

The challenge with GetConCliParam is that it entirely uses pass-by-reference parameters.  I’m not sure it is reasonable to expect every caller to this method to have to use every reference returned.   Before this PR, the only place this method was called was from the server dialog.   I thought it was more efficient to modify and re-use this method instead of creating an entirely new one for RPC.   In fact, the server dialog might benefit now from the additional channel info that was never available to it before; I’ll look into this.

src/serverrpc.cpp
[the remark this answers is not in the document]
Rob-NY: iNumChannels is the size of vecHostAddresses which 150 with the standard compile -- representing the max size of the all these vectors which is hard-coded as 15 (global.h:209)

It is true that I could count the loop through, but with large # of active connections would represent 'x' number of operations as opposed to a single call to a method that already knows the answer.

src/server.cpp
Rob-NY: This method was **never called** by CServer and is not used by CServer.  Prior to this PR, it was only called by server dialog which, as discussed below, now will also has access to the chanInfo that it didn't have before (albeit not yet used).  See discussion #2488

src/util.h
Rob-NY: I will update with an SL_NOT_SET case -- but I believe having a default is prudent;  perhaps "Not defined" (untranslated).

Rob-NY: It is worth noting from a semantics perspective that saying "Skill -> None" can mean both "having no skill" and "having not made a selection"  

image

I'm not fighting the suggestion, just making the point that an empty string is more reflective of "not making a choice" than the word "None" is.

src/serverrpc.cpp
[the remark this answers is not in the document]
pljones: i++ no more than 150 times is probably a _lot_ cheaper than the method call (allocate memory for the stack, put return address on stack, allocate memory for any returned values, make the call, copy local variable to stack, return from method, copy stack to local variable, free stack... depends how expensive getting the memory turns out to be...).

But imagine now you're in a separate thread.  And the JSONRPC code _should_ be running in a separate thread from the real-time server code.  You're now not only doing that in process call, you're doing a cross-thread context switch.

src/util.h
pljones: It's about consistency.  If the "None" in the English UI is changed to "Not specified", for example, the JSONRPC response should change, if you're expecting to get the human-readable value.

ann0see:
I'd like to have this in the next release...

Rob-NY:
> I'd like to have this in the next release...

Me too.  I've fixed what I believe are the issues raised (as opposed to philosophical questions), squashed and pushed.  Once this is merged I will resubmit the depended PR.

src/serverrpc.cpp
```diff
@@ -81,18 +92,29 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
...
             QJsonObject client{
                 { "id", i },
                 { "address", vecHostAddresses[i].toString ( CHostAddress::SM_IP_PORT ) },
                 { "name", vecsName[i] },
                 { "jitterBufferSize", veciJitBufNumFrames[i] },
                 { "channels", pServer->GetClientNumAudioChannels ( i ) },
+                { "instrumentCode", vecChanInfo[i].iInstrument },
+                { "instrumentName", CInstPictures::GetName ( vecChanInfo[i].iInstrument ) },
+                { "city", vecChanInfo[i].strCity },
+                { "countryCode", vecChanInfo[i].eCountry },
```
hoffie, line 105: Do we have a specific use case for exposing the numeric country code?
Unless there is a very good reason, we should not expose it here. IIRC the country code is Qt-native here. This means, the interpretation of this code would be influenced by whether a server is running a Qt5 or a Qt6 build.

If we do want to expose something besides the country name, we should expose the ISO code, I think.

src/util.h
hoffie: @pljones Sorry to join this conversation that late, but do we really want to expose the translated string here? Wouldn't this mean that the API client would need to handle different server languages...?

hoffie (commented):
Thanks for the PR and sorry for not commenting earlier. Looks good to me in general despite two questions (see inline comments).

src/util.h
pljones: If I'm writing a web site module that interacts over the JSONRPC interface, I don't want to have to worry about the API response getting translated because of the country Qt thinks the software is running in.  The API responses should always be the same _value_ for the same _meaning_.  How the web site then displays that in _its_ user interface is a different issue and outside the scope of JSONRPC.

Rob-NY:
> I'd like to have this in the next release...

@hoffie - is there anything I still need to do to move this along?  I'm asking because I'd like to resubmit the dependent PR which builds upon this basic one as a candidate as well.  I believe I've addressed all the refactor requests and questions.

src/util.h
pljones: @Rob-NY This is still outstanding.

src/serverrpc.cpp
ann0see: @Rob-NY This is still outstanding I think

ann0see: @Rob-NY ?

pljones:
@Rob-NY there are a number of unanswered requests.

Rob-NY:
> @Rob-NY there are a number of unanswered requests.

ISO Country code standardization:  I have no idea how to approach this at this time, so I asked to pass on this for the time being - no response.

In November, @hoffie asked you if we really wanted to expose the translated strings in some areas of this api - no response.

I made the changes for the "Skillset: NONE" vs. "not set" request as well as all the others raised here.   If I missed something, please point it out to me and I'll address right away.

ann0see:
Ok. So then there was a misunderstanding on both sides about the state of this PR. 

I‘ll look where I can help if I have time.

src/serverrpc.cpp
ann0see: > If we do want to expose something besides the country name, we should expose the ISO code, I think.


This would probably be https://doc.qt.io/qt-6.2/qlocale.html#codeToScript or https://doc.qt.io/qt-6.2/qlocale.html#bcp47Name or https://doc.qt.io/qt-5/qlocale.html#name then.

src/util.h
ann0see: Why can't we just return the int? If the API is well documented, the number should be sufficient.

docs/JSON-RPC.md
```diff
@@ -234,12 +234,20 @@ Results:
...
+| result.connections | number | The number of active connections. |
 | result.clients | array | The list of connected clients. |
 | result.clients[*].id | number | The client’s channel id. |
 | result.clients[*].address | string | The client’s address (ip:port). |
 | result.clients[*].name | string | The client’s name. |
 | result.clients[*].jitterBufferSize | number | The client’s jitter buffer size. |
 | result.clients[*].channels | number | The number of audio channels of the client. |
+| result.clients[*].instrumentCode | number | The id of the instrument for this channel. |
+| result.clients[*].instrumentName | number | The text name of the instrument for this channel. |
+| result.clients[*].city | string | The city name provided by the user for this channel. |
```
ann0see, line 246: Could someone please check if this returns the correct value?

I set the city to "...." but JSON-RPC just gives me:

```
{"id":1,"jsonrpc":"2.0","result":{"clients":[{"address":"127.0.0.1:22217","channels":2,"city":"","countryCode":0,"countryName":"Default","id":0,"instrumentCode":0,"instrumentName":"Nichts","jitterBufferSize":2,"name":"eto22","skillLevelCode":0,"skillLevelName":"Keins"}],"connections":1}}
```

ann0see: @Rob-NY could you please test this on your machine?

src/util.h
```diff
@@ -682,6 +682,31 @@ enum ESkillLevel
...
+    }
```
ann0see, line 707:
```suggestion
    return iSkill;
```

I'd rather remove this.

src/serverrpc.cpp
ann0see: After further thinking, probably the countryName is sufficient.

src/serverrpc.cpp
```diff
@@ -56,20 +56,31 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
...
-    /// @result {array} result.clients - The list of connected clients.
+    /// @result {number} result.connections - The number of active connections.
+    /// @result {array}  result.clients - The list of connected clients.
     /// @result {number} result.clients[*].id - The client’s channel id.
     /// @result {string} result.clients[*].address - The client’s address (ip:port).
     /// @result {string} result.clients[*].name - The client’s name.
     /// @result {number} result.clients[*].jitterBufferSize - The client’s jitter buffer size.
     /// @result {number} result.clients[*].channels - The number of audio channels of the client.
+    /// @result {number} result.clients[*].instrumentCode - The id of the instrument for this channel.
+    /// @result {number} result.clients[*].instrumentName - The text name of the instrument for this channel.
```
ann0see, line 67: I think we could argue to also remove this, as it's a Jamulus protocol ID. But to get this PR going, I'd accept it for now.

src/serverrpc.cpp
ann0see: The issue here is the QT5/QT6 change and inconsistency.

ann0see (changes requested):
Ok. See https://github.com/ann0see/jamulus/commit/f3d6bea9a89eb71a081a17a5f7ece7111ad5300f 
I hope we can agree on removing these methods as:
1. The country code is inconsistent between QT5 and QT6 (we already expose the country name which is not optimal but probably ok)
2. The skill level is translated and thus inconsistent. Alternatively, we'd find a way to only use the English translation. However, I think using just IDs is fine in this case, as the Jamulus protocol specifies the meaning. I believe that if we remove this - to stay consistent we should also remove the text translations of other translated names.

docs/JSON-RPC.md
ann0see: Also I believe that instrumentName is not a number but rather a String. As mentioned below, I think - as this is translated - we should think of removing it.

ann0see:
@pljones @Rob-NY would you be ok with my proposed changes?

src/util.h
pljones: Still outstanding.

src/util.h
pljones: Please, this is better.

Rob-NY:
> @pljones @Rob-NY would you be ok with my proposed changes?

No issue.

ann0see:
@Rob-NY @pljones pushed my commit. Waiting on further review now.

ann0see:
Ok. Rebased. Will do a new local test now.

ann0see (approved):
Local testing says this is ok. still, we could remove the instrument name - as it is translated and thus not reliable. Qt doesn't seem to support any short name for countries, so the country name is probably the best we can get.

ann0see:
@pljones (and maybe @dtinth) Could you please review this (especially from a C++ dev viewpoint)? I'd like to have this in. The legacy and iOS build failures are unrelated.

dtinth (approved):
Looks good to me too.

`vecChanInfo` in `CServerDlg` seems unused but that's fine I guess.

docs/JSON-RPC.md
```diff
@@ -234,12 +234,18 @@ Results:
...
 | --- | --- | --- |
+| result.connections | number | The number of active connections. |
 | result.clients | array | The list of connected clients. |
 | result.clients[*].id | number | The client’s channel id. |
 | result.clients[*].address | string | The client’s address (ip:port). |
 | result.clients[*].name | string | The client’s name. |
 | result.clients[*].jitterBufferSize | number | The client’s jitter buffer size. |
 | result.clients[*].channels | number | The number of audio channels of the client. |
+| result.clients[*].instrumentCode | number | The id of the instrument for this channel. |
+| result.clients[*].instrumentName | number | The text name of the instrument for this channel. |
```
pljones, line 245: `instrumentName | number`?  That sounds strange.  What is it meant to represent?  The position in the enum?  In which case it's redundant.

pljones:
I'd like a new issue raised with the changes requested here but not made.  Once that's done, I'm happy to approve it.

docs/JSON-RPC.md
ann0see: >  we could remove the instrument name

I think this is wrongly documented and should be a string (but I still believe it should be removed.)

ann0see:
@pljones 

For fixable stuff, I'd only have the removal of translated content (= instrument name). 
I believe there is more, but I think others have raised these issues:
* https://github.com/jamulussoftware/jamulus/pull/2918#discussion_r997291702
* https://github.com/jamulussoftware/jamulus/pull/2918#discussion_r997459985
* https://github.com/jamulussoftware/jamulus/pull/2918#pullrequestreview-1419847382
(feel free to add your comments to this list and then open an issue - it seems that most open stuff is related to design issues in the original code you found.)

pljones:
> @pljones
> 
> For fixable stuff, I'd only have the removal of translated content (= instrument name). I believe there is more, but I think others have raised these issues:

Yes, it's probably mostly the use of strings where the number would be better that should be fixed (at some point):
https://github.com/jamulussoftware/jamulus/pull/2918#discussion_r1191588195

ann0see:
Ok. I've now removed the instrument name too. 
@pljones could you please go through the comments (especially your comments) and open the issue? I think you have more context.

pljones (approved):
I'll approve anyway on the basis nothing here is breaking as far as I can see.  The `countryName` issue is complicated because of Qt 5/6 rather than anything else, as far as I remember and I wasn't involved much in the solution there.

Issue #2925, opened 2022-10-19
Mac: Signed release build process gets stuck during signing


hoffie:
**Describe the bug**

The 3.9.1 Mac builds could not be completed as all attempts to run the job in @emlynmac's repo with the appropriate signing environment in place failed. The builds got stuck. Further debugging showed that the builds got stuck during signing. Replacing `macdeployqt`'s signing part with `codesign` directly showed the same issue.
It is suspected that cccb21934c9162c0da28c0776d213e6fd8beb8c4 will fix the issue.

It's suspected that this issue only became apparent due to #2790, which made the builds take longer (roughly twice as long). That might have made them take long enough for them to trigger the keychain re-lock timeout.

**To Reproduce**
Run a 3.9.1 release build in a repo with signing environment.

**Expected behavior**
Builds should complete.

**Screenshots**

**Operating system**
Mac

**Version of Jamulus**
3.9.1 (3.9.0 did not show this issue)

**Additional context**

PR #2927, opened 2022-10-19
Build: Mac: Keep the signing-related keychain unlocked

hoffie:
This fixes hanging Mac build runs in CI which got stuck as they required interactive unlocking of the keychain.
This issue only became appareant after the switch to universal builds as that made the build take longer than the re-lock interval.

CHANGELOG: Build: Mac: Keep the signing-related keychain unlocked

**Context: Fixes an issue?**
Fixes: #2925

No.

The fix has already been committed and tested successfully on @emlynmac's repo where I've manually modified the r3_9_1 tag to do so.

Reviews.
(CI on this PR must not file, but it being successful does not prove anything either ;))

Please merge once reviews are completed.

## Checklist

### DIFF
--- a/.github/autobuild/mac.sh
+++ b/.github/autobuild/mac.sh
@@ -52,6 +52,8 @@ prepare_signing() {
     # Set up a keychain for the build:
     security create-keychain -p "${KEYCHAIN_PASSWORD}" build.keychain
     security default-keychain -s build.keychain
+    # Remove default re-lock timeout to avoid codesign hangs:
+    security set-keychain-settings build.keychain
     security unlock-keychain -p "${KEYCHAIN_PASSWORD}" build.keychain
     security import certificate.p12 -k build.keychain -P "${MACOS_CERTIFICATE_PWD}" -T /usr/bin/codesign
     security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k "${KEYCHAIN_PASSWORD}" build.keychain


emlynmac (approved):
Nice; LGTM

ann0see:
@daryu maybe you experienced something like this too? Could you please comment?

pljones:
One other comment: why did the build _hang_ rather than fail?  Do we know?  Is that something that could have been prevented and made to fail with a useful error?

pljones (approved):
If it works, it works.  Subject to my other comment that's out of scope for this hot fix.

hoffie:
> One other comment: why did the build _hang_ rather than fail? Do we know? Is that something that could have been prevented and made to fail with a useful error?

As far as I have found, it's likely not a real hang, but rather waiting for activity in a graphical dialog box, which we'll never see in the CI which is only CI-based.

pljones:
> > One other comment: why did the build _hang_ rather than fail? Do we know? Is that something that could have been prevented and made to fail with a useful error?
> 
> As far as I have found, it's likely not a real hang, but rather waiting for activity in a graphical dialog box, which we'll never see in the CI which is only CI-based.

That's what I was guessing, too.  Is there any way to prevent it, or a tool that better understands automated CI?

hoffie:
> That's what I was guessing, too. Is there any way to prevent it, or a tool that better understands automated CI?

I don't think so. The most official way seems to be Xcode, which is purely graphical. `codesign` is the fall back and is implicitly used by `macdeployqt`. It's probably not `codesign` itself who generates the dialog box, but rather the system keychain integration. I don't know of any other tool.
https://stackoverflow.com/questions/20205162/user-interaction-is-not-allowed-trying-to-sign-an-osx-app-using-codesign

ann0see:
I remember suspecting something like that some time ago too. That was the Apple script invocation in create-dmg (GitHub fixed it on their side a while ago.)

ann0see (approved):

pljones:
I _think_ this actually didn't make the 3.9.1 cut.  This is just looking at the merges since the r3_9_1 tag.

Issue #1945, opened 2021-08-08
Do not fail silently if IP in connection window is invalid


ann0see:
Follow up #1938 by @softins 

If an invalid IP is entered, the connection window closes and Jamulus does not give an error message. 

Proposed behaviour:

1. Check IP/Domain on entry and give a red warning "The IP you entered is invalid. Make sure you enter it in a correct format" 
2. Same as above, but after clicking on connect as error message

gilgongo:
It's really a bug, so I'll flag for next release in hope.

pljones:
#1938 is rather long.  Was there any particular comment?

By "Invalid", does this mean "Syntactically incorrect", rather than "Cannot connect to a Jamulus server at that address"?

If it's the former, it should be before closing the Connect dialog.  If it's the latter, the existing warning banner should suffice.

gilgongo:
@pljones I think invalid as in syntactically incorrect (so 192.168.o.10, and hopefully also a DNS address that doesn't resolve) as the message is "Make sure you enter it in a correct format".  Should probably be "The **address** you entered is invalid..." I think.

pljones:
It's got to cope with non-IP addresses, of course.  "I0.10.123.I0" is a perfectly valid address syntactically (it probably doesn't resolve, so you won't connect to anything).  The best check is to do just that - attempt to resolve the input text to an IP address.  If it fails, you error: "The address entered is not valid.  Make sure you typed it correctly."

ann0see:
De-Tagged. I think it’s a thing for the backloh

pljones:
I'm going to put this on the 3.11.0 backlog as it _shouldn't_ be too hard.

ann0see:
Probably we can check with a regular expression. I still believe the problem lies deeper: what if we enter an which doesn't have a Jamulus server running? 
But maybe that's already handled correctly with a timeout.

ann0see:
I believe https://github.com/jamulussoftware/jamulus/pull/3372 may fix this as side effect - to some extent at least.

PR #2930, opened 2022-10-20
Client: Add SRV-based virtual hosting support

hoffie:
- Internal: Refactor NetworkUtil::ParseNetworkAddress
  This moves the host elaborate and reusable host name lookup logic to its own ParseNetworkAddressName method.

- Client: Add SRV-based virtual hosting support
  When connecting to host names without port (e.g. example.org), the client will now perform an SRV lookup to _jamulus._udp.example.org. If this lookup returns exactly one result, the result is used to select the actual target address/port.

  If the lookup does not return a usable result, the regular connect logic kicks in (A lookup, default port).

CHANGELOG: Client: Add SRV-based virtual hosting support

**Context: Fixes an issue?**
Related: https://github.com/orgs/jamulussoftware/discussions/1772

Website, Server administration

Ready for testing/discussion.

Ready for testing/discussion.

## Checklist

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -914,6 +914,75 @@ bool NetworkUtil::ParseNetworkAddressString ( QString strAddress, QHostAddress& 
     return false;
 }
 
+#ifndef CLIENT_NO_SRV_CONNECT
+bool NetworkUtil::ParseNetworkAddressSrv ( QString strAddress, CHostAddress& HostAddress, bool bEnableIPv6 )
+{
+    // init requested host address with invalid address first
+    HostAddress = CHostAddress();
+
+    QRegularExpression plainHostRegex ( "^([^\\[:0-9.][^:]*)$" );
+    if ( plainHostRegex.match ( strAddress ).capturedStart() != 0 )
+    {
+        // not a plain hostname? then don't attempt SRV lookup and fail
+        // immediately.
+        return false;
+    }
+
+    QDnsLookup* dns = new QDnsLookup();
+    dns->setType ( QDnsLookup::SRV );
+    dns->setName ( QString ( "_jamulus._udp.%1" ).arg ( strAddress ) );
+    dns->lookup();
+    // QDnsLookup::lookup() works asynchronously. Therefore, wait for
+    // it to complete here by resuming the main loop here.
+    // This is not nice and blocks the UI, but is similar to what
+    // the regular resolve function does as well.
+    QTime dieTime = QTime::currentTime().addMSecs ( DNS_SRV_RESOLVE_TIMEOUT_MS );
+    while ( QTime::currentTime() < dieTime && !dns->isFinished() )
+    {
+        QCoreApplication::processEvents ( QEventLoop::ExcludeUserInputEvents, 100 );
+    }
+    QList<QDnsServiceRecord> records = dns->serviceRecords();
+    dns->deleteLater();
+    if ( records.length() != 1 )
+    {
+        return false;
+    }
+    QDnsServiceRecord record = records.first();
+    if ( record.target() == "." || record.target() == "" )
+    {
+        // RFC2782 says that "." indicates that the service is not available.
+        // Qt strips the trailing dot, which is why we check for empty string
+        // as well. Therefore, the "." part might be redundant, but this would
+        // need further testing to confirm.
+        // End processing here (= return true), but pass back an
+        // invalid HostAddress to let the connect logic fail properly.
+        HostAddress = CHostAddress ( QHostAddress ( "." ), 0 );
+        return true;
+    }
+    qDebug() << qUtf8Printable (
+        QString ( "resolved %1 to a single SRV record: %2:%3" ).arg ( strAddress ).arg ( record.target() ).arg ( record.port() ) );
+
+    QHostAddress InetAddr;
+    if ( ParseNetworkAddressString ( record.target(), InetAddr, bEnableIPv6 ) )
+    {
+        HostAddress = CHostAddress ( InetAddr, record.port() );
+        return true;
+    }
+    return false;
+}
+
+bool NetworkUtil::ParseNetworkAddressWithSrvDiscovery ( QString strAddress, CHostAddress& HostAddress, bool bEnableIPv6 )
+{
+    // Try SRV-based discovery first:
+    if ( ParseNetworkAddressSrv ( strAddress, HostAddress, bEnableIPv6 ) )
+    {
+        return true;
+    }
+    // Try regular connect via plain IP or host name lookup (A/AAAA):
+    return ParseNetworkAddress ( strAddress, HostAddress, bEnableIPv6 );
+}
+#endif
+
 bool NetworkUtil::ParseNetworkAddress ( QString strAddress, CHostAddress& HostAddress, bool bEnableIPv6 )
 {
     QHostAddress InetAddr;
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -889,6 +889,31 @@ QSize CMinimumStackedLayout::sizeHint() const
 * Other Classes                                                                *
 \******************************************************************************/
 // Network utility functions ---------------------------------------------------
+bool NetworkUtil::ParseNetworkAddressString ( QString strAddress, QHostAddress& InetAddr, bool bEnableIPv6 )
+{
+    // try to get host by name, assuming
+    // that the string contains a valid host name string or IP address
+    const QHostInfo HostInfo = QHostInfo::fromName ( strAddress );
+
+    if ( HostInfo.error() != QHostInfo::NoError )
+    {
+        // qInfo() << qUtf8Printable ( QString ( "Invalid hostname" ) );
+        return false; // invalid address
+    }
+
+    foreach ( const QHostAddress HostAddr, HostInfo.addresses() )
+    {
+        // qInfo() << qUtf8Printable ( QString ( "Resolved network address to %1 for proto %2" ) .arg ( HostAddr.toString() ) .arg (
+        // HostAddr.protocol() ) );
+        if ( HostAddr.protocol() == QAbstractSocket::IPv4Protocol || ( bEnableIPv6 && HostAddr.protocol() == QAbstractSocket::IPv6Protocol ) )
+        {
+            InetAddr = HostAddr;
+            return true;
+        }
+    }
+    return false;
+}
+
 bool NetworkUtil::ParseNetworkAddress ( QString strAddress, CHostAddress& HostAddress, bool bEnableIPv6 )
 {
     QHostAddress InetAddr;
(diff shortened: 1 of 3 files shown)


src/util.cpp
```diff
@@ -889,6 +889,97 @@ QSize CMinimumStackedLayout::sizeHint() const
...
+    {
+        HostAddress = CHostAddress ( InetAddr, record.port() );
+        return true;
+    }
+    return false;
+}
+
+bool NetworkUtil::ParseNetworkAddressWithSrvDiscovery ( QString strAddress, CHostAddress& HostAddress, bool bEnableIPv6 )
+{
+    // Try SRV-based discovery first:
```
hoffie, line 974: Trying SRV first is the saner behavior. It comes at the cost that everyone has to do the SRV lookup and wait for the (possibly empty) reply.

pljones: Agreed.  The DNS server should fail fairly fast if the entry isn't there.  At worst it should double the lookup time.

src/util.cpp
```diff
@@ -889,6 +889,97 @@ QSize CMinimumStackedLayout::sizeHint() const
 * Other Classes                                                                *
 \******************************************************************************/
 // Network utility functions ---------------------------------------------------
+bool NetworkUtil::ParseNetworkAddressName ( QString strAddress, QHostAddress& InetAddr, bool bEnableIPv6 )
+{
+    // try to get host by name, assuming
+    // that the string contains a valid host name string
```
pljones, line 895:
```suggestion
    // that the string contains a valid host name string or IP address
```

src/util.cpp
```diff
@@ -889,6 +889,97 @@ QSize CMinimumStackedLayout::sizeHint() const
 * Other Classes                                                                *
 \******************************************************************************/
 // Network utility functions ---------------------------------------------------
+bool NetworkUtil::ParseNetworkAddressName ( QString strAddress, QHostAddress& InetAddr, bool bEnableIPv6 )
```
pljones, line 892: I'm not keen on this method name.  It doesn't explain that it's the one place we turn any string into an IP address.  I think I'd prefer `ParseNetworkAddressString`.

src/util.cpp
```diff
@@ -889,6 +889,97 @@ QSize CMinimumStackedLayout::sizeHint() const
...
+{
+    // Try SRV-based discovery first:
+    if ( ParseNetworkAddressSrv ( strAddress, HostAddress, bEnableIPv6 ) )
+    {
+        return true;
+    }
+    // Try regular connect:
+    return ParseNetworkAddress ( strAddress, HostAddress, bEnableIPv6 );
+}
+
```
pljones, line 982: Could you add a new comment to say something like this expects to find the address either from a valid IP numeric value or a name via DNS lookup through A or AAAA records.

pljones: Oh - note that, given the RFC says the SRV entry can point to an AAAA record, `ParseNetworkAddress` might not quite work.  I think it expects IPv6 addresses to be numeric? `QRegularExpression rx1 ( "^\\[([^]]*)\\](?::(\\d+))?$" );`

src/util.cpp
hoffie: Thansk, updated.

src/util.cpp
hoffie: Ok, changed.

src/util.cpp
hoffie: > Could you add a new comment to say something like this expects to find the address either from a valid IP numeric value or a name via DNS lookup through A or AAAA records.

Added.

> Oh - note that, given the RFC says the SRV entry can point to an AAAA record, `ParseNetworkAddress` might not quite work. I think it expects IPv6 addresses to be numeric? `QRegularExpression rx1 ( "^\\[([^]]*)\\](?::(\\d+))?$" );`

Not sure I understand. I think plain IPs in SRV records aren't supposed to be supported. SRV records return names which can be looked up to A or AAAA records. Shouldn't ParseNetworkAddressString handle that properly already?

ann0see (commented):
Could you please #ifndef the code? I think some "raw jamulus" people might not want this feature (although it's small)

hoffie:
> Could you please #ifndef the code? I think some "raw jamulus" people might not want this feature (although it's small)

Done now (`QMAKE_CXXFLAGS+="-DCLIENT_NO_SRV_CONNECT"`). I don't think it warrants its own qmake flag though. I've also skipped #ifdef'ing one constant, which will be optimized away by the compiler anyway.

pljones (approved):

ann0see:
As soon as the DNS records are propagated (and I didn't make a mistake in the format), I can test this.

ann0see:
Ok. Tested ok.

src/util.cpp
```diff
@@ -889,6 +889,99 @@ QSize CMinimumStackedLayout::sizeHint() const
...
+    if ( HostInfo.error() != QHostInfo::NoError )
+    {
+        // qInfo() << qUtf8Printable ( QString ( "Invalid hostname" ) );
+        return false; // invalid address
+    }
+
+    foreach ( const QHostAddress HostAddr, HostInfo.addresses() )
+    {
+        // qInfo() << qUtf8Printable ( QString ( "Resolved network address to %1 for proto %2" ) .arg ( HostAddr.toString() ) .arg (
```
ann0see, line 906: Although unrelated, how do we handle debug output in other places at the moment? Commented code, QDebug()?

src/util.cpp
```diff
@@ -889,6 +889,99 @@ QSize CMinimumStackedLayout::sizeHint() const
...
+    QDnsLookup* dns = new QDnsLookup();
+    dns->setType ( QDnsLookup::SRV );
+    dns->setName ( QString ( "_jamulus._udp.%1" ).arg ( strAddress ) );
+    dns->lookup();
+    // QDnsLookup::lookup() works asynchronously. Therefore, wait for
+    // it to complete here by resuming the main loop here.
+    // This is not nice and blocks the UI, but is similar to what
+    // the regular resolve function does as well.
+    QTime dieTime = QTime::currentTime().addMSecs ( DNS_SRV_RESOLVE_TIMEOUT_MS );
+    while ( QTime::currentTime() < dieTime && !dns->isFinished() )
```
ann0see, line 940: Not a fan of busy waiting (even if it's limited). Isn't there any blocking function or something like a signal?

src/util.cpp
```diff
@@ -889,6 +889,99 @@ QSize CMinimumStackedLayout::sizeHint() const
...
+    if ( records.length() != 1 )
+    {
+        return false;
+    }
+    QDnsServiceRecord record = records.first();
+    if ( record.target() == "." || record.target() == "" )
+    {
+        // RFC2782 says that "." indicates that the service is not available.
+        // (Qt strips the trailing dot, which is why we check for empty string
+        // as well)
```
ann0see, line 955: Does this mean we might have two dots? (When and why) is this valid? Why don't we check just for the empty string?

src/util.cpp
```diff
@@ -889,6 +889,99 @@ QSize CMinimumStackedLayout::sizeHint() const
...
+        // Therefore, end processing here (= return true), but pass back an
+        // invalid HostAddress to let the connect logic fail properly.
```
ann0see, line 957: It doesn't ever fail properly in my opinion as there's no user output if we have an invalid IP/Hostname. But that's out of scope.

src/util.cpp
pljones: You're asking for a Qt equivalent of AA/AAAA lookup for SRV names?

src/util.cpp
hoffie: Right. I've noticed that as well and chose to implement it as close to the existing code path as possible.

src/util.cpp
hoffie: Note: This code is unchanged -- it only got moved around and therefore appears in this PR.

I don't see a clear logging policy in the code base, at least not as clear as other (server) software. We shouldn't have commented out code. Instead, we should make proper use of different debug levels and allow controlling them via command line. But yeah, out of scope here...

src/util.cpp
hoffie: > Isn't there any blocking function

The existing code path uses [QHostInfo::fromName](https://doc.qt.io/qt-6/qhostinfo.html#fromName), which is a blocking function. I haven't found any equivalent for SRV lookups.

> or something like a signal?

The `QDnsLookup`, which I've used here, is supposed to be used via signals. However, this requires -- as far as I can see -- an inversion of the control flow as the existing logic relies on blocking. Even when that refactoring is done, new questions come up: How should the UI behave? Currently it blocks. If it continues to stay responsive during lookup, we would need some other kind of indicator, so that's kind of a rabbit hole. Not saying that it shouldn't be done, but rather that it's a larger refactoring which I didn't want to do as part of this PR.

In short: That's the best and least intrusive way I've found. If someone finds a better way, I'm certainly open to it. :)

src/util.cpp
hoffie: The code basically covers both what the RFC states plus what I've practically observed during testing on Linux. As I haven't tested all other platforms or different DNS implementations, I'd go with this duplicate check -- it's about the failure path anyway. It might be sufficient to check for the empty string, but it would be some effort to prove, I guess?

src/util.cpp
ann0see: We can leave it like this for now and look at it in a later refactoring.

src/util.cpp
ann0see: Do we have an issue for that? I think I once opened one...

hoffie: #1945 is about improving the connect window error handling. Do you mean that?

ann0see: Yes.

src/util.cpp
ann0see: At least add a comment here that the dot check might not be needed.

hoffie: Done

ann0see:
I still don't feel comfortable with the dot/empty string workaround. The Qt docs don't seem to give much insight. Why does the dot get removed?

hoffie:
> I still don't feel comfortable with the dot/empty string workaround. The Qt docs don't seem to give much insight. Why does the dot get removed?

Qt [uses the OS' libresolv](https://github.com/mburakov/qt5/blob/93bfa3874c10f6cb5aa376f24363513ba8264117/qtbase/src/network/kernel/qdnslookup_unix.cpp#L298) (usually glibc). glibc [has logic to remove a leading dot, which also covers names consisting only of a dot](https://github.com/bminor/glibc/blob/b92a49359f33a461db080a33940d73f47c756126/resolv/dn_expand.c#L99).

ann0see:
Hmm ok. Approving and merging then.

ann0see (approved):

ann0see:
Dropped needs documentation as the next-release page is up to date: https://github.com/jamulussoftware/jamuluswebsite/blob/next-release/wiki/en/Unregistered-Servers.md#dns-srv-record-support

Issue #2933, opened 2022-10-22
Two letter country codes need to be in lower case to work in 3.9.1


Rob-NY:
Upper or mixed-case ISO country codes are not recognized per #2932 but lower case works.  Comparison statement in util.h:1463 needs to be slightly refactored.

PR #2934, opened 2022-10-22
Fix case-related ISO country code issue

Rob-NY:
Fixes issue relating to ISO country codes only working if provided in lower case.  Now support any case combination.

CHANGELOG: ISO country codes can now be supplied in upper or lower case

**Context: Fixes an issue?**
Fixes issue #2933 

Probably not as I think the general expectation is that the country code should not be case sensitive anyway.

## Checklist

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -1460,7 +1460,7 @@ QLocale::Country CLocale::GetCountryCodeByTwoLetterCode ( QString sTwoLetterCode
     {
         QStringList vstrLocParts = qLocale.name().split ( "_" );
 
-        if ( vstrLocParts.size() >= 2 && vstrLocParts.at ( 1 ).toLower() == sTwoLetterCode )
+        if ( vstrLocParts.size() >= 2 && vstrLocParts.at ( 1 ).toLower() == sTwoLetterCode.toLower() )
         {
             return qLocale.country();
         }


hoffie:
CHANGELOG: Server: ISO country codes can now be supplied in upper or lower case

hoffie (approved):
Thanks!

src/util.cpp
```diff
@@ -1460,7 +1460,7 @@ QLocale::Country CLocale::GetCountryCodeByTwoLetterCode ( QString sTwoLetterCode
     {
         QStringList vstrLocParts = qLocale.name().split ( "_" );
 
-        if ( vstrLocParts.size() >= 2 && vstrLocParts.at ( 1 ).toLower() == sTwoLetterCode )
+        if ( vstrLocParts.size() >= 2 && vstrLocParts.at ( 1 ).toLower() == sTwoLetterCode.toLower() )
```
pljones, line 1463: https://doc.qt.io/qt-6/qlocale.html#codeToTerritory
https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements
From what I can make out, I'd be expecting upper case?

Rob-NY: Best I can tell, the QT6 code handles the case conversion so it will accept upper/lower/mixed.  See https://codebrowser.dev/qt6/qtbase/src/corelib/text/qlocale.cpp.html line 157.

For QT < 6, my contribution here was to fix an immediate bug.  I added exactly 10 characters to the codebase.  Since the original ISO contribution forced a toLower(), I simply balanced it to solve the problem. 

I have not walked up the QT code tree to see what the QT5 data for locales looks like, so unless someone wants to do that, I think it is safer to normalize the strings (upper or lower) as done here.

pljones (approved):

Issue #2937, opened 2022-10-25
Build: Use Github environment files instead of deprecated set-output before 2023-05-31


hoffie:
> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Many build steps show this warning during autobuild jobs. Need to check other jobs as well. Some warnings might be related to third-party repos. #2936 fixes one such instance.

```
.github/workflows/autobuild.yml:            echo "::set-output name=build_all_targets::${1}"
.github/autobuild/android.sh:    echo "::set-output name=artifact_1::${artifact}"
.github/autobuild/ios.sh:    echo "::set-output name=artifact_1::${artifact}"
.github/autobuild/linux_deb.sh:    echo "::set-output name=artifact_1::${artifact_1}"
.github/autobuild/linux_deb.sh:    echo "::set-output name=artifact_2::${artifact_2}"
.github/autobuild/windows.ps1:    echo "::set-output name=artifact_1::${artifact}"
.github/autobuild/get_build_vars.py:    print("::set-output name={}::{}".format(varname, varval))
.github/autobuild/mac.sh:    echo "::set-output name=macos_signed::true"
.github/autobuild/mac.sh:    echo "::set-output name=artifact_1::${artifact}"
```

Adding to 3.10.0 milestone for now, but feel free to drop it if it is not done at release done.

PR #2938, opened 2022-10-25
Remove deprecated "set-output" from Autobuild

ann0see:
GitHub is deprecating "set-output", thus changing it to new approac, thus changing it to new approach

Needs to be tested. Especially the Windows/Python parts

CHANGELOG: SKIP

**Context: Fixes an issue?**

Fixes: https://github.com/jamulussoftware/jamulus/issues/2937

No

Test (check if all output variables are still ok)

CI run

## Checklist

-  [] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)


.github/autobuild/get_build_vars.py
```diff
@@ -43,8 +43,9 @@ def get_build_version(jamulus_pro_version):
 
 def set_github_variable(varname, varval):
     print("{}='{}'".format(varname, varval))  # console output
-    print("::set-output name={}::{}".format(varname, varval))
-
+    output = os.getenv('GITHUB_OUTPUT')
+    with open(output, "a") as ghout:
+        ghout.write(f"{varname}={varval}")
```
hoffie: I think this needs a line break, otherwise it fails as soon as a second value is written.

```suggestion
        ghout.write(f"{varname}={varval}\n")
```

.github/autobuild/windows.ps1
```diff
@@ -150,8 +150,8 @@ Function Pass-Artifact-to-Job
     {
         throw "move failed with exit code $LastExitCode"
     }
-    echo "Setting Github step output name=artifact_1::${artifact}"
-    echo "::set-output name=artifact_1::${artifact}"
+    echo "Setting Github step output name=artifact_1 to ${artifact}"
+    $Env:GITHUB_OUTPUT="artifact_1=${artifact}"
```
hoffie, line 154: This looks unexpected as it assigns the output to an environment variable instead of writing to the file referenced by that variable. Shouldn't this be some kind of redirection?

```suggestion
    echo "{artifact_1}={${artifact}}" >> "$Env:GITHUB_OUTPUT"
```

.github/autobuild/get_build_vars.py
```diff
@@ -43,7 +43,8 @@ def get_build_version(jamulus_pro_version):
 
 def set_github_variable(varname, varval):
     print("{}='{}'".format(varname, varval))  # console output
-    print("::set-output name={}::{}".format(varname, varval))
+    os.environ['GITHUB_OUTPUT'] = f"varname=varval"
```
hoffie, line 46:
```suggestion
    os.environ['GITHUB_OUTPUT'] = f"varname=varval"
```

This sounds wrong, similar to the powershell part. The previous attempt was better, but lacked a new line, I think.

ann0see:
@hoffie thanks for the quick intervention here ;-)

I used 
https://bitcoden.com/answers/new-format-of-output-in-custom-github-actions-since-set-output-is-going-to-be-deprecated

Will check...

.github/autobuild/windows.ps1
ann0see: Probably yes. Overwriting the envionment variable would be fatal ;-).

.github/autobuild/get_build_vars.py
```diff
@@ -43,8 +43,9 @@ def get_build_version(jamulus_pro_version):
 
 def set_github_variable(varname, varval):
     print("{}='{}'".format(varname, varval))  # console output
-    print("::set-output name={}::{}".format(varname, varval))
-
+    outputfile = os.getenv('GITHUB_OUTPUT')
+    with open(outputfile, "a") as ghout:
+        ghout.write(f"varname=varval\n")
```
hoffie, line 48: Try this:
```suggestion
        ghout.write(f"{{{varname}}}={{{varval}}}\n")
```

ann0see: Does this let python interpret varname as variable?

hoffie: Yep, it should.
```
>>> foo = 1
>>> f'{{{foo}}}'
'{1}'
```

ann0see: Hmm. I think we both initially read the docs in a wrong way ;-). There's no need for "{"

ann0see: It seems to be building now. But I'll change it to the f syntax

ann0see:
iOS is probably unrelated. Seems ok now.

hoffie:
> iOS is probably unrelated. Seems ok now.

Second attempt already got further.

hoffie (approved):
Looks good to me. Let's hope iOS doesn't have surprises for us (CI is unfinished at the time of writing). ;)

pljones (approved):

Issue #2947, opened 2022-10-30
Logging: Don't complain about serverinfo if no --serverinfo is given


hoffie:
**Describe the bug**

`Jamulus -s -n` logs `Ignoring invalid serverinfo, please verify the parameter syntax.`.

**To Reproduce**

**Expected behavior**
No error should be logged if no `--serverinfo` is given.

**Screenshots**

**Operating system**
Linux

**Version of Jamulus**
master

**Additional context**
Issue got introduced in 5b2c2de6d0d2ee2363957b87a5ee984b2cce3e9f.

pljones:
@hoffie how is progress on this?

PR #2951, opened 2022-10-31
Server: Don't warn when no --serverinfo is provided

hoffie:
The intention of 5b2c2de was to introduce logging for bad --serverinfo parameters. However, it also warns about bad serverinfo if no --serverinfo is given at all. This is confusing.
Therefore, this commit removes the error when no --serverinfo has been given.

CHANGELOG: Server: Removed spurious log message when not using --serverinfo

**Context: Fixes an issue?**
Fixes: #2947

No.

Ready.

Reviews.

## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -650,6 +650,13 @@ int main ( int argc, char** argv )
             exit ( 1 );
         }
 
+        if ( ServerOnlyOptions.contains ( "--serverinfo" ) && strServerInfo.split ( ";" ).count() < 3 )
+        {
+            qWarning() << qUtf8Printable (
+                QString ( "\"--serverinfo '%1'\": must contain [name];[city];[country] - value will be ignored" ).arg ( strServerInfo ) );
+            strServerInfo = "";
+        }
+
 #ifndef HEADLESS
         if ( bUseGUI )
         {
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -232,13 +232,6 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
                                         .arg ( slServInfoSeparateParams[2] )
                                         .arg ( QLocale::countryToString ( ThisServerListEntry.eCountry ) ) );
     }
-    else
-    {
-        if ( !strServerInfo.isEmpty() )
-        {
-            qWarning() << "Ignoring invalid serverinfo, please verify the parameter syntax.";
-        }
-    }
 
     // per definition, the very first entry is this server and this entry will
     // never be deleted
--- a/src/serverlist.cpp
+++ b/src/serverlist.cpp
@@ -234,7 +234,10 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
     }
     else
     {
-        qWarning() << "Ignoring invalid serverinfo, please verify the parameter syntax.";
+        if ( !strServerInfo.isEmpty() )
+        {
+            qWarning() << "Ignoring invalid serverinfo, please verify the parameter syntax.";
+        }
     }
 
     // per definition, the very first entry is this server and this entry will


ann0see (approved):

pljones:
When was there a change to make `--serverinfo ""` valid?

pljones:
Can we at least change the PR title to match what the problem is (i.e. error on missing `--serverinfo` param) rather than on empty param argument.

hoffie:
> When was there a change to make `--serverinfo ""` valid?

I think there was no change.

> Can we at least change the PR title to match what the problem is (i.e. error on missing `--serverinfo` param) rather than on empty param argument.

I agree that the title was confusing and I've changed the title and the commit message accordingly

> What I'm saying above is that I don't think this check _should_ be needed in CServer - the check for empty argument should be done in main.cpp and start up shouldn't happen.

Yeah, right, but that's a different, rather hypothetical issue, I'd say?

ann0see:
@pljones Do you think this is ready?

src/serverlist.cpp
```diff
@@ -234,7 +234,10 @@ CServerListManager::CServerListManager ( const quint16  iNPortNum,
     }
     else
     {
-        qWarning() << "Ignoring invalid serverinfo, please verify the parameter syntax.";
+        if ( !strServerInfo.isEmpty() )
```
pljones, line 237: Why is this repeating the check done on line 175?  This `else` block is only entered `if ( !strServerInfo.isEmpty() )`.

pljones:
Reviewed.  I still don't think the PR is valid.

ann0see:
> Can we at least change the PR title to match what the problem is (i.e. error on missing `--serverinfo` param) rather than on empty param argument.
> 
> 
> 
> What I'm saying above is that I don't think this check _should_ be needed in CServer - the check for empty argument should be done in main.cpp and start up shouldn't happen.

Ok. Due to this comment?

src/serverlist.cpp
ann0see: Then the issue is somewhere else probably.

pljones:
> Ok. Due to this comment?

Yes.

pljones:
We need to agree what happens next here.  I'm proposing to drop from 3.10.0 until that's done.

ann0see:
I think it's a regression which should be fixed. 

At *least* the message should say that serverinfo is be empty *or* invalid as intermediate fix.

pljones:
> I think it's a regression which should be fixed.
> 
> At _least_ the message should say that serverinfo is be empty _or_ invalid as intermediate fix.

Yes, in `main.cpp`.  It shouldn't be possible to pass syntactically invalid values through to `server.cpp` and `serverlist.cpp`, so no changes should be needed there.

ann0see (commented):
@pljones actually, I think, the else is related to line 197 

```    if ( iServInfoNumSplitItems >= 3 )```

**Comment by ann0see** [comment]
Ah yes, and empty serverinfo is valid for unregistered servers. And I'm quite sure that this isn't checked. Do you think that registered servers *must* provide a server info?

**Comment by pljones** [comment]
1. It's valid not to supply `--serverinfo` on the command line whether registering or not - `serverlist.cpp`:
```c
    // prepare the server info information
    QStringList slServInfoSeparateParams;
    int         iServInfoNumSplitItems = 0;

    if ( !strServerInfo.isEmpty() )
    {
        // split the different parameter strings
        slServInfoSeparateParams = strServerInfo.split ( ";" );

        // get the number of items in the split list
        iServInfoNumSplitItems = slServInfoSeparateParams.count();
    }
...
    // parse the server info string according to definition:
    // [this server name];[this server city];[this server country as QLocale ID] (; ... ignored)
    // per definition, we expect at least three parameters
    if ( iServInfoNumSplitItems >= 3 )
    {
...
    }
    else
    {
        qWarning() << "Ignoring invalid serverinfo, please verify the parameter syntax.";
    }
```

There's nothing _invalid_ about it.  It simply isn't used.  The else clause should never have been added.

Once upon a time, a long long time ago, before the pandemic, `--serverinfo` was used to pass the list of servers that the central server knew about (rather than have the few hard coded ones that preceeded that version).

Ideally, now, we'd pass a parsed "ServerInfo" object to the server on start up, to avoid the parsing deeper inside.

However, for now, I'd be happy just to see the parameter actually validated in `main.cpp`.

ann0see:
>

pljones:
The problem with `main.cpp` currently is around settings handling.  We parse the command line _fully_ before considering settings.  That means the content of settings aren't taken into account (and we have dirty hacks in place to take the command line into account when processing settings).

Currently, the "keep it small and simple" approach would _just_ be to validate the parameter when we're passed `--serverinfo` on the command line.  The change that added the warning in `serverlist.cpp` when no `--serverinfo` is passed could also be reverted.

But I don't think we should _just_ revert that `serverlist.cpp` change alone.

ann0see:
@pljones could you please have a look at this PR then: https://github.com/jamulussoftware/jamulus/pull/2599 it changes a lot in this area, I suppose?

pljones:
Commented on #2599.

ann0see:
@pljones I know you think this is wrong to merge, but can you update this PR to have it in an acceptable state? I believe this is a regression and should be fixed in the next release.

pljones:
As a minimum, I think that if `--serverinfo` is provided on the command line then the required parameter should be checked to ensure it has two separator characters in (at least).  No further check at that stage is required.

ann0see:
Yes. Sounds good:
* Get rid of the else
* Verify syntax in main.cpp ?

pljones:
I've taken the liberty of pushing to this branch to avoid raising a new PR.

pljones:
@ann0see are you happy with this approach?

ann0see:
Not quite. If it is empty, I think it should be skipped without any output.

src/main.cpp
```diff
@@ -650,6 +650,13 @@ int main ( int argc, char** argv )
             exit ( 1 );
         }
 
+        if ( strServerInfo.split ( ";" ).count() < 3 )
```
ann0see, line 653: if no cli argument is given, I think we shouldn't output anything. I'd argue that's the point of this PR.

ann0see (changes requested):
See comment

src/main.cpp
pljones: No - it's an error.  The argument is _mandatory_, it's a command line syntax error.  It won't get this far.

pljones: The problem comes if you swap `-s` and `--serverinfo`:
```
$ ./Jamulus -p 50000 --serverinfo -s
- Starting in server mode by default (due to compile time option)
- selected port number: 50000
- server info: -s
No GUI support compiled. Running in headless mode.
"--serverinfo '-s'": must contain [name];[city];[country] - value will be ignored
... continues to run ...
```
It could error out.

pljones: Of course, the above is when running compiled server only.  Without that, it's still an error:
````
p$ ./Jamulus -p 50000 --serverinfo -s
- selected port number: 50000
- server info: -s
./Jamulus: Server only option(s) '--serverinfo' used.  Did you omit '--server'?
```

pljones: This is why I say the `main.cpp` validation needs tightening up.  But not in this PR -- it's a much bigger problem than this one parameter.

ann0see: Hmm. So you say `jamulus -s` is invalid? 

The point of this PR is that jamulus -s doesn't give an error. `jamulus -s --serverinfo` should give an error.

pljones: No, `Jamulus -s` is perfectly valid.  The issue is that the parser incorrectly accepts `-s` as the argument to `--serverinfo` and swallows it, then spits out the `--serverinfo` as it's a server-only parameter (and it's not running as a server).  And _that's_ inconsistent, depending on whether you compiled server-only or not.  So it's a mess bigger than the scope of this PR.

ann0see: Ah ok. That's a separate issue. However I get a warning of invalid serverinfo for `jamulus -s` that should not happen.

pljones: Give the full command line, otherwise I can't comment.

pljones: Ah, right, I see...  Oops.

pljones: Patch applied (and one that should probably be used when validating all the other parameters).

ann0see (approved):
I'm happy now. Thanks

pljones (approved):

PR #2944, opened 2022-10-29
Mac: Enable signing with self signed cert

ann0see:
**Short description of changes**
Enables the CI to use a code signing certificate signed by a non Apple CA to sign development builds. 

grafik
grafik

CHANGELOG: 

**Context: Fixes an issue?**

Fixes: https://github.com/jamulussoftware/jamulus/issues/2924

**Does this change need documentation? What needs to be documented and how?**

Probably not. 

**Status of this Pull Request**

Ready for review (and test on Apple Silicon, a repo without the respective secrets set and a repo with the real apple secrets set (@emlynmac 's repo))

**What is missing until this pull request can be merged?**

Still needs some (external testing). Artifacts are building on my repo: https://github.com/ann0see/jamulus/actions/runs/3352547263/jobs/5554760442
## Checklist

-  [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
-  [x] I tested my code and it does what I want (on my repo)
-  [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
-  [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
-  [x] I've filled all the content above

AUTOBUILD: Please build all targets

PR #3009, opened 2023-02-13
Remove request for non native recording directory

ann0see:

Removes the request for a non native directory selection box on all OS for the recording directory and the persistence file. I hope this doesn't introduce any bad side effects. 

This is done to allow write access to a recording directory outside of the macOS sandbox for the Server.

CHANGELOG: Server: Use native file selection dialog for recording directory and persistence file to allow read and write access outside of the macOS sandbox.

**Context: Fixes an issue?**

Related to #2131 

Probably yes?

To be tested with a signed build. I'd like to have an ok from @emlynmac to build on his repo and a confirmation why we chose to use the Qt file dialog instead of the native OS one by @pljones 

Internal clarification and testing.

## Checklist

-   [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)


pljones:
> why we chose to use the Qt file dialog instead of the native OS one

One reason could have been documentation.  Using the Qt dialog means that each OS gets the same Look and Feel when choosing a file/directory from Jamulus.  Such usage in the documentation is then clear to the user.  Using the OS-native one means we need to document each OS separately, so the user is hand-held (noting that GUI server users are likely less technical).  The up-side is that they're also likely more familiar with the OS-native widget.

Another reason could have been testing: one can make assumptions like "if I've tested the workflow on Linux with the Qt widget, it'll work on Windows with the Qt widget".  That may be an invalid assumption using the OS-native widget.  For example, choosing a file and choosing a directory may use different widgets - I think I hit some issue there at some point.  Qt uses the same widget for both.

ann0see:
Ok. Probably we'll only know more if we test it on all OS

pljones:
Code looks okay, I'll wait for test feedback to approve.

ann0see:
macOS unsigned artifact from this build works.

ann0see:
@emlynmac can you/me push the changes to your repo and test if the recording directory can now be set outside the sandbox?

ann0see:
I've now signed the build with a self signed certificate via #2944 here: https://github.com/ann0see/jamulus/actions/runs/4193064214 and I can save files (= Recordings) outside of the Sandbox (I tested the Downloads folder).

emlynmac:
> @emlynmac can you/me push the changes to your repo and test if the recording directory can now be set outside the sandbox?

I've sync'd. 
Can we also set up github to have main as the main branch, rather than a disclaimer that master is gone?

ann0see:
@emlynmac Thanks. Unfortunately, I only see that the main branch got built. Could you please git checkout my changes from this PR and build them?

emlynmac:
@ann0see https://github.com/emlynmac/jamulus/actions/runs/4198106890

ann0see:
The build seems to do what it should.

pljones:
Do we have example screen shots yet?

ann0see:
It's just the normal OS file/directory dialog. 

I can take some pictures tomorrow

pljones:
I've no idea what that looks like :).  Also, we need examples from all supported platforms for "Open file" and "Open directory" (also, in case there are differences, with different permission requests).

ann0see:
Open Directory in macOS:

image

pljones:
> Sorry for the delay. @pljones here are the screenshots:

Thanks!  Really useful as I've currently only access to a Chromebook, so couldn't even do Windows ones easily.

ann0see:
@pljones do you consider this as mergable?

pljones (approved):

pljones:
Yes, feel free to merge.

Issue #3019, opened 2023-02-25
The Jamulus build system runs with an unsupported clang format version


pljones:
**Describe the bug**

The Jamulus build system runs with an unsupported clang format version:
```
Run DoozyX/clang-format-lint-action@c3b2c943e924028b93a707a5b1b017976ab8d50c
  with:
    clangFormatVersion: 10
```
Ubuntu only provides `apt` with 11 to 15:
```
$ sudo apt install clang-format<TAB><TAB>
clang-format     clang-format-11  clang-format-12  clang-format-13  clang-format-14  clang-format-15
```
(https://apt.llvm.org/ only provides 15 and newer.)
```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.2 LTS
Release:        22.04
Codename:       jammy
```

**To Reproduce**

Install and use the earliest supported clang-format version:
```
$ clang-format --version
Ubuntu clang-format version 11.1.0-6
$ make clang_format
...
$ git diff
(nothing changed)
$ git status
Your branch is up-to-date with 'origin/bugfix/#2723-self-label-checkboxes-forms1'.
```

Run "Verify coding style" on https://github.com/pljones/jamulus/tree/bugfix/remove-remaining-directory-server-mentions
https://github.com/jamulussoftware/jamulus/actions/runs/4270698669/jobs/7435420804
https://github.com/jamulussoftware/jamulus/actions/runs/4270698669/attempts/1

**Expected behavior**

Consistent clang format behaviour with _at least_ clang-format-11 (but, given https://apt.llvm.org/, I'd suggest getting onto 15 as soon as we can).

**Screenshots**

n/a

**Operating system**

Github build system on Linux.

**Version of Jamulus**

n/a

**Additional context**

Blocking https://github.com/jamulussoftware/jamulus/pull/3017

ann0see:
I'm not sure if there's still a reason for the old clang format version. 
Even Debian Stable uses clang format 11.

ann0see:
Related to: https://github.com/jamulussoftware/jamulus/issues/2552

PR #3020, opened 2023-03-01
[Fix] #3019 Unsupported clang version

pljones:

Updates clang version to Ubuntu LTS version.

CHANGELOG: Update clang version to Ubuntu LTS version

**Context: Fixes an issue?**

Fixes: 3019

No.

Needed a couple of whitespace updates in line with formatter rules... I _think_ it's right...  But someone else should confirm.

Just check it's not now doing the wrong thing.

## Checklist

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -934,20 +934,20 @@ void CServerSettings::ReadSettingsFromXML ( const QDomDocument& IniXMLDocument, 
                                     static_cast<int> ( AT_DEFAULT ),
                                     static_cast<int> ( AT_CUSTOM ),
                                     iValue ) )
-        {
-            directoryType = static_cast<EDirectoryType> ( iValue );
-        }
-        //### TODO: END ###//
+            {
+                directoryType = static_cast<EDirectoryType> ( iValue );
+            }
+            //### TODO: END ###//
 
-        else if ( GetNumericIniSet ( IniXMLDocument,
-                                     "server",
-                                     "directorytype",
-                                     static_cast<int> ( AT_NONE ),
-                                     static_cast<int> ( AT_CUSTOM ),
-                                     iValue ) )
-        {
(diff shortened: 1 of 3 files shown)


src/settings.cpp
```diff
@@ -934,20 +934,20 @@ void CServerSettings::ReadSettingsFromXML ( const QDomDocument& IniXMLDocument,
                                     static_cast<int> ( AT_DEFAULT ),
                                     static_cast<int> ( AT_CUSTOM ),
                                     iValue ) )
-        {
-            directoryType = static_cast<EDirectoryType> ( iValue );
-        }
-        //### TODO: END ###//
+            {
```
pljones, line 937: So this is the one I think is questionable...  I think.

ann0see: Probably. I'd just add a `{` after the else - also for clarity.

In my own code, I only omit `{` if it's possible to write the statement in the same line as the else but still consider it "lazy styling":
```c
if (something()) call_true();
```

pljones: The blocks here are
- 911-915 (open/close align)
- 916-960 (open/close align)
    - 918-925 (`//### TODO: BEGIN ###//` and `//### TODO: END ###//` do _not_ align.  This is because the END is in the middle of an `else ... if`)
    - 927-940 (note line 942 should be split as 924 and 931 are, causing the same issue; open/close align)
    - 942-950 (open/close align)
    - 952-959 (open/close align)

pljones: So I don't think this is a problem now, resolving.

ann0see: It still looks confusing. Did you open an issue on this?

pljones:
Unless anyone's got objections, I'll force-merge to unblock other changes.

ann0see (approved):

PR #3046, opened 2023-04-15
Define JACK headers and sources globally and add some documentation 

ann0see:
Instead of using hardcoded paths to the sound files @pgScorpio suggested to use variables. This PR uses global definitions for JACK sources and headers and adds some small comments.

CHANGELOG: Internal: Improved maintainability of Jamulus.pro by using global definitions.

**Context: Fixes an issue?**
Fixes: #2573

No

Ready for testing

Needs whole CI to pass

## Checklist

-  [x] I tested my code and it does what I want (Linux (JACK gives no warning and sound is OK), Windows too. I can't test macOS at the moment - but build is OK.)

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -125,7 +125,7 @@ win32 {
                 libjackname = "libjack64.lib"
             }
             !exists("$${programfilesdir}/JACK2/include/jack/jack.h") {
-                error("Error: jack.h was not found in the expected location ($${programfilesdir}). Ensure that the right JACK2 variant is installed (32bit vs. 64bit).")
+                error("Error: jack.h was not found in the expected location ($${programfilesdir}). Ensure that the right JACK2 variant is installed (32 Bit vs. 64 Bit).")
             }
 
             HEADERS += src/sound/jack/sound.h
@@ -178,6 +178,7 @@ win32 {
     QMAKE_BUNDLE_DATA += OSX_ENTITLEMENTS
 
     macx-xcode {
+        # As of 2023-04-15 the macOS build with Xcode only fails. This is tracked in #1841
         QMAKE_INFO_PLIST = mac/Info-xcode.plist
         XCODE_ENTITLEMENTS.name = CODE_SIGN_ENTITLEMENTS
         XCODE_ENTITLEMENTS.value = mac/Jamulus.entitlements
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -68,6 +68,10 @@ INCLUDEPATH_OPUS = libs/opus/include \
     libs/opus/silk/fixed \
     libs/opus
 
+# As JACK is used in multiple OS, we declare it globally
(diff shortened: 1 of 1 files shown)


Jamulus.pro
```diff
@@ -68,6 +68,35 @@ INCLUDEPATH_OPUS = libs/opus/include \
     libs/opus/silk/fixed \
     libs/opus
 
+# Add include paths and headers for new audio APIs here
+
+# ASIO
```
pljones, line 73: I'd split this into two parts:
- provided by the SDK
- "part of Jamulus"

pljones: I'd also make _all_ the variable definitions be either
```
VARNAME =
```
or
```
VARNAME = value
```
depending on the config that makes them relevant.

ann0see:
Hmm. Then I misunderstood you. I thought you were referring to splitting the ASIO part out to another variable? Can you please provide a suggestion what you'd like to see?

Jamulus.pro
```diff
@@ -68,6 +68,34 @@ INCLUDEPATH_OPUS = libs/opus/include \
...
+SOURCES_COREAUDIO_MAC = src/sound/coreaudio-mac/sound.cpp
+
+# CoreAudio iOS
+HEADERS_COREAUDIO_IOS = src/sound/coreaudio-ios/sound.h
+SOURCES_COREAUDIO_IOS = src/sound/coreaudio-ios/sound.mm
+
+# Oboe
+HEADERS_OBOE = src/sound/oboe/sound.h
+SOURCES_OBOE = src/sound/oboe/sound.cpp \
+        src/android/androiddebug.cpp # TODO: Remove debugging code if no longer needed
```
pljones, line 97: Should line 97 be done more like
```
SOURCES_OBOE = src/sound/oboe/sound.cpp
# TODO: Remove debugging code if no longer needed else move to src/sound if part of sound
SOURCES_OBOE += src/android/androiddebug.cpp
```

pljones:
I'm starting to think this is creating a lot of variables for no benefit.

The variables should only be added to the overall HEADERS, etc, variables conditionally - so wouldn't it be better to do _that_ at the time these are being set?

Jamulus.pro
ann0see: Ok. Yes. Done.

ann0see:
@pljones so you mean basically that we should revert this? I think @pgScorpio mainly wanted to get rid of the duplication introduced by the JACK headers.

pljones:
> @pljones so you mean basically that we should revert this? I think @pgScorpio mainly wanted to get rid of the duplication introduced by the JACK headers.

Are the variables used in multiple places?  Otherwise there's no duplication being removed.

ann0see:
The JACK headers are used for all three main OS.

ann0see:
@pljones I've now updated my PR and split it into two commits. I hope that's ok for all.

pljones (approved):

PR #3075, opened 2023-05-28
Build: Move release handling to GH cli

ann0see:

Makes more use of native gh command to get rid of third party actions. This has the benefit of less complexity and the need of trust on external dependencies. 

CHANGELOG: Internal: Use native gh command for release management and uploading artefacts 

**Context: Fixes an issue?**

Related to: #3069 

No

Ready for review. Tested Prerelease and release on my repo. We should decide if the current way of uploading two files on the linux build is ok or two steps/moving the whole upload logic into the autobuild scripts is preferred. This would add more complex logic. 

Agreement on how to handle multiple artefacts. 

## Checklist

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -65,7 +65,7 @@ jobs:
     runs-on:                        ubuntu-20.04
     outputs:
       publish_to_release:           ${{ steps.get-build-vars.outputs.PUBLISH_TO_RELEASE }}
-      upload_url:                   ${{ steps.create-release.outputs.upload_url }}
+      tag_name:                     ${{ steps.get-build-vars.outputs.RELEASE_TAG }}
       build_version:                ${{ steps.get-build-vars.outputs.BUILD_VERSION }}
       build_all_targets:            ${{ steps.decide-build-targets.outputs.build_all_targets }}
     env:
@@ -85,28 +85,28 @@ jobs:
         if:                         steps.get-build-vars.outputs.PUBLISH_TO_RELEASE == 'true'
         run:                        ./.github/autobuild/extractVersionChangelog.pl ChangeLog ${{ steps.get-build-vars.outputs.JAMULUS_PRO_VERSION }} > ${{ env.release_changelog_path }}
 
-      - name:                       Remove release ${{steps.get-build-vars.outputs.RELEASE_TAG}}, if existing
-        if:                         steps.get-build-vars.outputs.PUBLISH_TO_RELEASE == 'true'
-        continue-on-error:          true
-        uses:                       dev-drprasad/delete-tag-and-release@0885ecc4123e071dd1a8be04d3a0578f5390581c
-        with:
-          delete_release:           true
-          tag_name:                 ${{ steps.get-build-vars.outputs.RELEASE_TAG }}
-        env:
-          GITHUB_TOKEN:             ${{ secrets.GITHUB_TOKEN }}
-
-      - name:                       Create Release ${{steps.get-build-vars.outputs.RELEASE_TAG}}  ${{steps.get-build-vars.outputs.RELEASE_TITLE}}
+      - name:                       Create/Recreate Release ${{steps.get-build-vars.outputs.RELEASE_TAG}}  ${{steps.get-build-vars.outputs.RELEASE_TITLE}}
         if:                         steps.get-build-vars.outputs.PUBLISH_TO_RELEASE == 'true'
         id:                         create-release
-        uses:                       actions/create-release@v1
+        run:                        |
+          set -eu
+
+          # delete release if existing
+          gh release delete "${tag_name}" || true
+
+          RELEASE_PRERELEASE_ARG=""
+          if [[ "${prerelease}" == 'true' ]]; then
+            RELEASE_PRERELEASE_ARG="--prerelease"
+          fi
+          # Actually create the release. This will print the release url if successful
+          gh release create "${tag_name}" --title "${release_name}" --notes-file "${body_path}" ${RELEASE_PRERELEASE_ARG}
+
         env:
           GITHUB_TOKEN:             ${{ secrets.GITHUB_TOKEN }}
-        with:
           tag_name:                 ${{ steps.get-build-vars.outputs.RELEASE_TAG }}
           release_name:             ${{ steps.get-build-vars.outputs.RELEASE_TITLE }}
           body_path:                ${{ env.release_changelog_path }}
           prerelease:               ${{ steps.get-build-vars.outputs.IS_PRERELEASE }}
-          draft:                    false
 
       - name:                       Decide which targets to build for
         id:                         decide-build-targets
@@ -296,10 +296,13 @@ jobs:
         run:                        |
           set -eu
           apt-get -qq update
-          apt-get install -y software-properties-common sudo
+          apt-get install -y software-properties-common sudo curl
           add-apt-repository ppa:git-core/ppa
+          curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg
+          sudo chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg
+          echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null
           apt-get -qq update
-          apt-get -qq -y install git-core
+          apt-get -qq -y install git-core gh
           # Github Runner environments usually whitelist the default checkout directory for git.
           # However, when using containers, we have to do that manually in order for git commands to work
           # (e.g. in Jamulus.pro's VERSION logic):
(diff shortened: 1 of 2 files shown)


.github/autobuild/android.sh
```diff
@@ -100,10 +100,10 @@ build_app_as_apk() {
 }
 
 pass_artifact_to_job() {
-    mkdir deploy
+    mkdir publish
     local artifact="jamulus_${JAMULUS_BUILD_VERSION}_android.apk"
```
ann0see, line 104: uploading everything from the deploy directory doesn’t work as
1. GH on Windows doesn‘t support wildcards
2. MacOS moves the binary (.app files) to the deploy folder

If instead we upload the files by name - as we do in this state - we could also use the deploy directory directly.

.github/workflows/autobuild.yml
```diff
@@ -410,36 +413,16 @@ jobs:
...
-          asset_content_type:       application/octet-stream
-
-      - name:                       Upload Artifact 2 to Release
-        if:                         >-
-                                    steps.get-artifacts.outputs.artifact_2 != '' &&
-                                    needs.create_release.outputs.publish_to_release == 'true'
-        id:                         upload-release-asset2
-        uses:                       actions/upload-release-asset@v1
+        id:                         upload-release-assets
+        run:                        cd publish; gh release upload "${{ needs.create_release.outputs.tag_name }}" ${{ steps.get-artifacts.outputs.artifact_1 }} ${{ steps.get-artifacts.outputs.artifact_2 }} --clobber
```
ann0see, line 421: Not sure if omitting quotes is the best way to "overload" this command. Basically if quotes are omitted we can have an arbitrary amount of files to upload

.github/workflows/autobuild.yml
```diff
@@ -85,28 +85,28 @@ jobs:
...
-        if:                         steps.get-build-vars.outputs.PUBLISH_TO_RELEASE == 'true'
-        continue-on-error:          true
-        uses:                       dev-drprasad/delete-tag-and-release@0885ecc4123e071dd1a8be04d3a0578f5390581c
-        with:
-          delete_release:           true
-          tag_name:                 ${{ steps.get-build-vars.outputs.RELEASE_TAG }}
-        env:
-          GITHUB_TOKEN:             ${{ secrets.GITHUB_TOKEN }}
-
       - name:                       Create Release ${{steps.get-build-vars.outputs.RELEASE_TAG}}  ${{steps.get-build-vars.outputs.RELEASE_TITLE}}
```
pljones, line 88: Maybe change the step name to make it clear any existing release also gets removed?

.github/workflows/autobuild.yml
pljones, line 421: Can this be split onto multiple lines for readability?
```
  run:
    cd publish
    gh release upload "${{ needs.create_release.outputs.tag_name }}" ${{ steps.get-artifacts.outputs.artifact_1 }} ${{ steps.get-artifacts.outputs.artifact_2 }} --clobber
```

pljones: I also wish we built an array of artifacts created by the build and passed that, rather than `artifact_1` and `artifact_2`... it feels wrong.

.github/workflows/autobuild.yml
```diff
@@ -85,28 +85,28 @@ jobs:
...
         if:                         steps.get-build-vars.outputs.PUBLISH_TO_RELEASE == 'true'
         id:                         create-release
```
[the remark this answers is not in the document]
ann0see: Deleting the release should not happen that often. "create-release" is good enough in my opinion.

.github/workflows/autobuild.yml
ann0see: I'm trying something else now.

ann0see: See the "or" approach. We still need to check if PowerShell is ok with that. What do you think?

.github/workflows/autobuild.yml
```diff
@@ -412,34 +415,15 @@ jobs:
...
-
-      - name:                       Upload Artifact 2 to Release
-        if:                         >-
-                                    steps.get-artifacts.outputs.artifact_2 != '' &&
-                                    needs.create_release.outputs.publish_to_release == 'true'
-        id:                         upload-release-asset2
-        uses:                       actions/upload-release-asset@v1
+        id:                         upload-release-assets
+        run:                        |
+          gh release upload "${{ needs.create_release.outputs.tag_name }}" "deploy/${{ steps.get-artifacts.outputs.artifact_1 }}" "deploy/${{ steps.get-artifacts.outputs.artifact_2 }}" --clobber || gh release upload "${{ needs.create_release.outputs.tag_name }}" "deploy/${{ steps.get-artifacts.outputs.artifact_1 }}" --clobber
```
pljones, line 422: https://www.baeldung.com/yaml-multi-line#folded-style
This might be nicer as
```yaml
  run: >
    gh release upload "${{ needs.create_release.outputs.tag_name }}" "deploy/${{ steps.get-artifacts.outputs.artifact_1 }}" "deploy/${{ steps.get-artifacts.outputs.artifact_2 }}" --clobber
    ||
    gh release upload "${{ needs.create_release.outputs.tag_name }}" "deploy/${{ steps.get-artifacts.outputs.artifact_1 }}" --clobber
```
(I think that'll fold right....)

.github/workflows/autobuild.yml
```diff
@@ -412,34 +415,17 @@ jobs:
...
-        if:                         >-
-                                    steps.get-artifacts.outputs.artifact_2 != '' &&
-                                    needs.create_release.outputs.publish_to_release == 'true'
-        id:                         upload-release-asset2
-        uses:                       actions/upload-release-asset@v1
+        id:                         upload-release-assets
+        run:                        >
+          gh release upload "${{ needs.create_release.outputs.tag_name }}" "deploy/${{ steps.get-artifacts.outputs.artifact_1 }}" "deploy/${{ steps.get-artifacts.outputs.artifact_2 }}" --clobber
+          ||
+          gh release upload "${{ needs.create_release.outputs.tag_name }}" "deploy/${{ steps.get-artifacts.outputs.artifact_1 }}" --clobber
```
pljones, line 424: 🤦 I just thought... why not test for the artifact presence before trying to upload it?  (It's two separate operations under the covers and this avoids succeeding to upload artifact_1, partly uploading artifact_2 then failing to upload the already existing artifact_1:
```yaml
run: -
  gh release upload "${{ needs.create_release.outputs.tag_name }}" "deploy/${{ steps.get-artifacts.outputs.artifact_1 }}" --clobber
  [[ -f "deploy/${{ steps.get-artifacts.outputs.artifact_2 }}" ]] && gh release upload "${{ needs.create_release.outputs.tag_name }}" "deploy/${{ steps.get-artifacts.outputs.artifact_2 }}" --clobber
```

ann0see: Because Powershell doesn't support this syntax as far as I know.

pljones: OK, the yaml's just fed to host shell, of course.  But this is the artifact deploy, not the build, isn't it..?  Does it run on top of PowerShell?

ann0see: I think so. Or CMD but I'm quite sure it's not bash.

pljones: Looks like it must be `cmd`, then - Powershell doesn't like the syntax when I try it.  (`echo hello || echo goodbye` for example.)

pljones: Anyway, at the moment, this works out as
> Upload the first file
> If that was okay (script still running), try uploading the second file but don't worry if that fails

So
```
run: -
  gh release upload "${{ needs.create_release.outputs.tag_name }}" "deploy/${{ steps.get-artifacts.outputs.artifact_1 }}" --clobber
  gh release upload "${{ needs.create_release.outputs.tag_name }}" "deploy/${{ steps.get-artifacts.outputs.artifact_2 }}" --clobber || echo "Ignoring failed artifact_2 upload"
```

ann0see: > Powershell doesn't like the syntax when I try it. (echo hello || echo goodbye for example.)

Strange. I read that it was the other way around. I'll fire up a test build.

ann0see: See: https://stackoverflow.com/questions/2416662/what-are-the-powershell-equivalents-of-bashs-and-operators/41816341#41816341 which I used as reference

ann0see: And that's the relevant issue: https://github.com/cli/cli/issues/5099

ann0see: @pljones see https://github.com/ann0see/jamulus/releases/tag/r3_9_1devTestPlainGHRelease - so it works with the current hacky syntax.

pljones: Given all the preceeding steps (e.g. use of `apt-get` to install `gh`), I'd expect it to be on bash...  Things like `set -eu` aren't very portable :).

ann0see: Yes. But we have a conditional there. I wanted to avoid them as it's a simple invocation of gh which *should* behave the same on all OS

pljones: We've got platform-dependent code even in `Jamulus.pro`, which isn't nice.  (Means building for Android on Windows doesn't work -- there's a `wc -l` call...)

But as I say, here we might as well do the first artifact, then, if that's worked, do the second artifact, then, if that fails, continue.  As that appears to be the only option for checking whether artifact two is present in the same step.

ann0see: Not the only one. We could also augment the building scripts as in the initial state of this PR -> Each OS has a custom "upload artifact to release" step comparable to the build.

.github/workflows/autobuild.yml
```diff
@@ -494,49 +479,14 @@ jobs:
...
-        id:                         deb-upload-keyascfile
-        uses:                       actions/upload-release-asset@v1
         env:
           GITHUB_TOKEN:             ${{ secrets.GITHUB_TOKEN }}
-        with:
-          upload_url:               ${{ needs.create_release.outputs.upload_url }}
-          asset_path:               repo/key.asc
-          asset_name:               key.asc
-          asset_content_type:       application/pgp-keys
+        run:                        |
```
ann0see, line 490: Since this is Bash/Linux only wildcards should work fine.

ann0see:
I've now hard-coded the repo files. Hopefully that will work now.

ann0see (commented):
Dismiss

.github/workflows/autobuild.yml
ann0see: @pljones I've now fixed the PR for the repo too. What do you think?

ann0see: > Anyway, at the moment, this works out as
> 
> > Upload the first file
> 
> > If that was okay (script still running), try uploading the second file but don't worry if that fails
> 
> 
> 
> So
> 
> ```
> 
> run: -
> 
>   gh release upload "${{ needs.create_release.outputs.tag_name }}" "deploy/${{ steps.get-artifacts.outputs.artifact_1 }}" --clobber
> 
>   gh release upload "${{ needs.create_release.outputs.tag_name }}" "deploy/${{ steps.get-artifacts.outputs.artifact_2 }}" --clobber || echo "Ignoring failed artifact_2 upload"
> 
> ```

Probably that's more readable, true. Hopefully PowerShell supports that

pljones (approved):

ann0see:
I've removed the release upload and release create and delete actions now from the actions whitelist as they're no longer used.

PR #1975, opened 2021-08-25
JSON-RPC interface to control client and server

dtinth:
**Short description of changes**

Allow Jamulus client and server to be programmatically controlled through a JSON-RPC-based protocol.

**Context: Fixes an issue?**

This PR addresses the following items:

- https://github.com/jamulussoftware/jamulus/discussions/1958
  - `jamulusserver/setWelcomeMessage` to dynamically change welcome message.
- https://github.com/jamulussoftware/jamulus/discussions/1863
  - `jamulusserver/getConnectedClients` to get clients connected w/ IP info.
- https://github.com/jamulussoftware/jamulus/discussions/1847
  - `jamulusserver/getServerInfo` to get server status

**Does this change need documentation? What needs to be documented and how?**

- `--jsonrpcport <port>`
- The JSON-RPC protocol.

**Status of this Pull Request**

~~I did this PR mainly to solve my own issues, so I see this more like an RFC with a working implementation. So PR status is more like a proof of concept, although I am already using it in my own systems.~~ Also I am C++ noob, so there may be major issues with the code, esp with memory management.

Update: All PR comments have been address and all checks have passed, I believe the PR is in a mergeable state.

**What is missing until this pull request can be merged?**

- [x] Consensus on whether or not, and how, to proceed forward? [Assumed through lazy consensus.](https://github.com/jamulussoftware/jamulus/pull/1975#issuecomment-913170767)
- [x] Updating this PR on top of `master` branch (right now it is on top of `r3_8_0`).
- [x] Some exploratory testing to ensure that Jamulus does not crash.
- [x] Test running server with constant API usage to ensure no memory leak
- [x] API documentation for jamulus/ and jamulusserver/
- [x] Check for other fixes/extensions on https://github.com/dtinth/jamulus/commits/dtinth
- [x] Final test of all API calls
- [x] Squash commits in a history friendly way
## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

## Background
I would like to programmatically control Jamulus server and client. My use cases include:

- **Server**
[...]

PR #3089, opened 2023-06-22
steer people away from curl and toward nc or ncat.

mcfnord:
I tried curl and failed. With this fix, the next guy won't make the same mistake, and might even be on a platform that supports nc or ncat.

### DIFF
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -29,6 +29,8 @@ The JSON-RPC server is based on the [JSON-RPC 2.0](https://www.jsonrpc.org/speci
 - A **response** from Jamulus to the consumer.
 - A **notification** from Jamulus to the consumer.
 
+**Tip:** You can't use `curl` to connect to this server. On Linux, you can use the `nc` CLI tool. On Windows, you can download and use the `ncat` CLI tool.
+
 ## Example
 
 After opening a TCP connection to the JSON-RPC server, the connection must be authenticated:


docs/JSON-RPC.md
```diff
@@ -29,6 +29,8 @@ The JSON-RPC server is based on the [JSON-RPC 2.0](https://www.jsonrpc.org/speci
 - A **response** from Jamulus to the consumer.
 - A **notification** from Jamulus to the consumer.
 
+**Tip:** You can't use `curl` to connect to this server. On Linux, you can use the `nc` CLI tool. On Windows, you can download and use the `ncat` CLI tool.
```
ann0see, line 32: I wouldn't say "can't use curl" but rather say something positive like "use netcat if you want to control the API manually via CLI"

ann0see (changes requested):

docs/JSON-RPC.md
mcfnord: Is this because curl is based on wget? Has anyone ever used curl in this scenario?

ann0see: I think I once tried curl, but as you found out it's the wrong tool (you might be able to work around it [by misusing http headers](https://stackoverflow.com/questions/12661159/curl-read-headers-from-file) which I didn't try). Curl and wget are both standard tools for getting web (HTTP) access and I often use them interchangeably.

https://github.com/jamulussoftware/jamulus/pull/1975 is the initial PR and I agree that a quick and dirty start is missing. 
I'd suggest not to even mention curl or wget as this would just cause confusion. Direct the user to the correct tool right away.

mcfnord: I also tried curl, [as did others](https://github.com/orgs/jamulussoftware/discussions/2691). Standard tools seem uniquely qualified for explicit guidance when they aren't fit.

Can we link to #1975 from here? It even has a code snippet. "See this page for connection examples using nc and Node.js."

ann0see: I'd rather include the code snippet here.
Maybe even add a shell example with authentication. 

Or we could see this as the start to write an actual JSON-RPC page on the website?

mcfnord: If a squash becomes required, I'm not into it. So now I'm reluctant to add commits, because the wrong one can trigger a squash request, and stall the work. Would it be controversial to merge this now, and perhaps do more later? I bet that's unrealistic too, due to translation?

ann0see: The app isn't a problem at all. If you're really that reluctant of squashing, one of the maintainers can do it. We try to merge every PR with at least one (better two) reviews which means that contributors might need to add commits. The last step is a rebase and squash.

mcfnord: Would it be controversial to merge this now, and perhaps do more later?

I will figure squash out, but it's just part of my problem. Writing docs professionally has made me opinionated about workflows. In the workplace, sometimes I will say, "We might do more later. But can this be merged?" It's a way to ask if an acceptability bar has been met. It can circumvent apparent disagreement about the ideal options. I caused some guy to flip out when I asked this once. He had (unrealistic) demands and I wasn't dancing like a monkey for him. But a simple yes/no was appropriate! There can be perfectionism in the engineering mindset.

What are the implications of adequate but non-ideal documentation changes? The same content might involve multiple PR's, which probably implies more commits. I suppose it results in a publication that could include sub-optimal tips. Adequate but not optimal. Any other concerns?

Finally, have you used the **Add a suggestion** interface? Give it a try here if you can. I guess it would generate a commit that might need squashing. But there's something about the "documentation philosophy" discourses that make me itch, and maybe they wouldn't if we all tried writing our ideal improvement, rather than writing about it.

ann0see:
Your repo is gone... Not sure

mcfnord:
> Your repo is gone... Not sure

I've been making a pull request and then destroying my fork to make another fork. I thought this gave me a "clean slate" but didn't know this caused the PR to close. I'll stick with different branches on the same fork from now on.

PR #3098, opened 2023-07-09
Add tip for connecting to JSON-RPC endpoint

mcfnord:
Discussed in https://github.com/jamulussoftware/jamulus/pull/3089/

PR #3101, opened 2023-07-11
ip server connection tips

mcfnord:
This script includes additional tips for connecting to a JSON-RPC server, [as discussed here](https://github.com/jamulussoftware/jamulus/pull/3098/).


mcfnord:
Maybe these are tips, rather than just one tip.

tools/generate_json_rpc_docs.py
```diff
@@ -211,6 +211,17 @@ def to_markdown(self):
 - A **response** from Jamulus to the consumer.
 - A **notification** from Jamulus to the consumer.
 
+**Tip:** On Linux, you can connect to a JSON-RPC server using the `nc` CLI tool. On Windows, you can download and use the `ncat` CLI tool.  This snippet uses [jayson](https://www.npmjs.com/package/jayson) to connect using Node.js:
```
ann0see, line 214:
```suggestion
## Connecting to the JSON-RPC server

On **Linux**, you can use the `nc` CLI tool. 

On **Windows**, you can download and use the `ncat` CLI tool. 

To control the JSON-RPC server programmatically, you can use a package like [jayson](https://www.npmjs.com/package/jayson) for Node.js line this:
```

I'd give a link to ncat if it isn't installed by default. Maybe even give examples for the nc commands.

ann0see (commented):
See comment.

docs/JSON-RPC.md
```diff
@@ -29,6 +29,17 @@ The JSON-RPC server is based on the [JSON-RPC 2.0](https://www.jsonrpc.org/speci
 - A **response** from Jamulus to the consumer.
 - A **notification** from Jamulus to the consumer.
 
+**Tip:** On Linux, you can connect to a JSON-RPC server using the `nc` CLI tool. On Windows, you can download and use the `ncat` CLI tool.  This snippet uses [jayson](https://www.npmjs.com/package/jayson) to connect using Node.js:
```
mcfnord, line 32:
```suggestion
**Connection Tips:** On Linux, you can connect to a JSON-RPC server using the `nc` CLI tool. On Windows, [you can download](https://nmap.org/ncat/) and use the `ncat` CLI tool.  This snippet uses [jayson](https://www.npmjs.com/package/jayson) to connect using Node.js:
```

mcfnord:
What about Mac? I bet it has nc ?

ann0see:
Probably? But IDK.

ann0see (approved):
We'll squash this on merge

docs/JSON-RPC.md
```diff
@@ -1,14 +1,14 @@
...
 
 <!--
 This file is automatically generated from the source code.
 Do not edit this file manually.
 See `tools/generate_json_rpc_docs.py` for details.
 -->
 
-A JSON-RPC interface is available for both Jamulus client and server to allow programmatic access.
-To use the JSON-RPC interface, run Jamulus with the `--jsonrpcport <port> --jsonrpcsecretfile /file/with/a/secret.txt` options.
+A JSON-RPC server can be added to both Jamulus client and Jamulus server to allow programmatic access.
```
pljones, line 10: Style: "Jamulus client" and "Jamulus server" should be "Jamulus Client" and "Jamulus Server" to comply with our terminology guide.

tools/generate_json_rpc_docs.py
```diff
@@ -181,16 +181,16 @@ def to_markdown(self):
...
 
 <!--
 This file is automatically generated from the source code.
 Do not edit this file manually.
 See `tools/generate_json_rpc_docs.py` for details.
 -->
 
-A JSON-RPC interface is available for both Jamulus client and server to allow programmatic access.
-To use the JSON-RPC interface, run Jamulus with the `--jsonrpcport <port> --jsonrpcsecretfile /file/with/a/secret.txt` options.
+A JSON-RPC server can be added to both Jamulus client and Jamulus server to allow programmatic access.
```
pljones, line 192: At _runtime_, it's nothing is being "adding" - the service is there.  The `--jsonrpc...` options _enable_ the service.

docs/JSON-RPC.md
```diff
@@ -1,14 +1,14 @@
 
-# Jamulus JSON-RPC Documentation
+# Jamulus JSON-RPC Server Documentation
 
 <!--
 This file is automatically generated from the source code.
 Do not edit this file manually.
 See `tools/generate_json_rpc_docs.py` for details.
 -->
```
pljones, line 8: Given the above, why is it added to git?

tools/generate_json_rpc_docs.py
pljones: Also, a style comment: "Jamulus client" and "Jamulus server" should be "Jamulus Client" and "Jamulus Server" to comply with our terminology guide.  (However, "JSON-RPC server" would use title case - only capitalising the "server" in titles.  I've a marginal preference for "service", as "server" suggests it's separate.)

I guess this guide hasn't been through style guide compliance check yet, though.

docs/JSON-RPC.md
mcfnord: I want the finished page to be found at a fixed location on the internet.

tools/generate_json_rpc_docs.py
mcfnord: I defer to the style of the document I'm in. 

Does the JSON-RPC server reveal itself in any way when no port-password params are present? Add those, get it?

My bias against **enable** started about 20 years ago. Now I frequently prefer **adds**, **lets**, or **turns on**.  I often ask myself, "What would a 4-year-old understand?" Alright, let's examine your claim: The service _is there_, yet it can be _enabled_? Is it there and disabled? If userspace can't see its enabled-disabled state, then I don't find its there-ness noteworthy. Google's style guide relegates _enabled_ to UI states, where a disabled control has a visual indication in userspace. Is there any such state indication here? What can I do with a disabled JSON-RPC server? Can I enable it in some way other than providing these parameters? Are you concerned that a systems-level developer would be misled by this oversimplification presented to value-add developers?

docs/JSON-RPC.md
pljones: Maybe it should be handled under jamulus/jamuluswebsite?  I don't think we've anything else quite like this...

@gilgongo ?

ann0see: I think it was considered as "developer only" but if people think JSON RPC is ok to become a bit more official, we can surely add it to the website

gilgongo: I'm not sure if there's a convention for this. If the majority of those who would use JSON RPC would naturally go to the repo to find the docs on it (as they would for compiling, or licensing, say), then that's an argument for keeping it in the repo. Otherwise it would make sense to have as much info as possible in one place (the website), I think?

ann0see: I don't think most of them would go to the repo by default. But I'd say they can be directed to the repo - but that isn't needed. 

However - this file is automatically generated and updated if the code changes. So that's something to consider.

pljones: > However - this file is automatically generated and updated if the code changes. So that's something to consider.

And that's what I don't like - the generated files shouldn't be under git but deployed some other way as part of the build.  The Jamulus executable is generated as part of the build and people expect to come to the Github repo to get hold of it (some people who know about Github and the build/release process it has).  I'd like the same process to trigger deployment of these files, which can be referenced from a fixed place that _is_ held in git.

mcfnord: If you're saying it's dodgy to rely on someone running this script to update this output file when code changes, I agree.

pljones: If it's getting to Github, then it's going through the PR process.

That means it goes through the deployment process.

Which means the generated files don't need to be committed to git, just published, like the Jamulus executable gets published when it's built.

mcfnord: There's inherent duplication now, which caused me some confusion, but I'm not sure what the change would look like.

gilgongo: > We have developer docs at a fixed address that we can link to. What's wrong with that?

Yes - seems a bit too cosmetic to move the API docs to the website if the "journey" for anyone interested in using it is that they browse the website, see there's `--jsonrpcport` and then follow the link. I don't know if there's some SEO advantage in having it under jamulus.io, but that's likely to be very small.

The question of the update script and how that gets run is a separate matter though.

gilgongo: > There's inherent duplication now, 

Is there?

pljones: I was thinking of Github pages.  When the build runs, you can configure the workflow run to publish the file (see [here](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site)).  We could publish under `http://jamulussoftware.github.io/jamulus/`, to differentiate from the "official" `https://jamulus.io/` site, using GitHub workflows to render the file(s).  That could be linked from the main GitHub README.md, if desired...

ann0see: I think that's confusing. It should in the best case open a PR with the changes on the website.

mcfnord: > > There's inherent duplication now,
> 
> Is there?

Yes. There's the finished page (which I initially modified) and there's the script that creates that page (which I modify in this PR).

pljones: > the finished page

That's like the Jamulus executable.  That's why it shouldn't be held under source control - it's _not_ source material, it should be output from the build process.

We just need to ensure the build process builds and publishes it correctly.

mcfnord: I understand, and even agree, in principle. In reality, owning the "[care and feeding](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md)" of a PR is unrealistic when the implications become so far-reaching.

pljones: Agreed - if the existing process _already_ works this way, it's not for this PR to change it.  It's just sad it does.  It needs an issue raised, at least, before the PR can be signed off, though, so the existing issue is tracked.

mcfnord: https://github.com/jamulussoftware/jamulus/issues/3108

tools/generate_json_rpc_docs.py
pljones: > Is there any such state indication here?

Assuming the build ran with the JSONRPC code, then there _should_ be a control in client and server UI allow control over use.  That there isn't is part of the JSONRPC implementation not being completed, rather than design.  Basing design decisions on an incomplete implementation isn't good practice...

pljones (approved):

tools/generate_json_rpc_docs.py
mcfnord: Technical writing is not design decisions. It's communication decisions. Public-facing content should reflect the product shipping, in the most easily understood terms. I like to update it when new features are finally live, as a last and savory step. In general, futureware should have nearly no impact on live documentation. At least that's a rule of thumb I try to use. We don't have to worry about confusing our readers when we switch later to coverage of a more complex featureset. Our time with readers is incredibly brief, and we often have just one chance to be understood adequately. What may seem like oversimplification is often a service to readers. (I also simplify to an extreme degree when coding UI, because I think the same cognition principles apply there. I'm a little obsessed with how closely I can say the bare minimum or simplest thing without impeding outcomes.) Another term I've heard for straddling this racy oversimplification line is _marketecture_.

ann0see: Fair point. Did you already read our style and tone guide? https://jamulus.io/contribute/Style-and-Tone

I think that's the place we should remember to stay close to. Probably worth discussing that in a GH discussion or issue.

PR #3115, opened 2023-07-29
Refactor: Pass "CommandLineOptions" to CSettings::Load by reference

pljones:

Pass "CommandLineOptions" to CSettings::Load by reference

CHANGELOG: SKIP

**Context: Fixes an issue?**

N/A

No

I was using it in my main build but I split it out to make it easier to raise the PR.  I'll use it in my (GUI) client and (headless) server for today.

Review.

## Checklist

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -25,7 +25,7 @@
 #include "settings.h"
 
 /* Implementation *************************************************************/
-void CSettings::Load ( const QList<QString> CommandLineOptions )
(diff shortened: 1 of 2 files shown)


ann0see (commented):
We should check that this doesn't cause side effects.  In the code later down (needs code reading).

pljones:
It shouldn't have any effect: the value is (well, should be) used only read-only in the code.

ann0see (approved):
Looking at the code, we do just use it with .contains, so everything is fine - as it should be.

PR #3151, opened 2023-08-23
#1420 Fader Settings improvements

### LINKED ISSUE #1420: "Set all faders to new client level" only affects current users
**Describe the bug**

When using "Set all faders to new client level" this only resets the faders of current users, if a previous user signs in it's level is still set to the previous level.

**To Reproduce**

Sign in to a server with multiple users.
Set all levels to i.e. max.
Sign out one user.
use "Set all faders to new client level".
sign in the previous user again and it's level will still be set to max. not the "New Client Level".

**Expected behavior**

Just like when using "Clear All Stored Solo and Mute Settings" also the stored levels should be reset to new client level when using "Set all faders to new client level"

**Screenshots**

**Operating system**

n.a.

**Version of Jamulus**

3.6.x, 3.7.0
**Additional context**

Maybe also an added option to delete all stored users?

pljones:

Two changes relating to #1420:

* Edit -> Set All Faders to New Client Level
This retained the last known level for any client, so when that client rejoined the server, their old level was restored
image
This has been reworded to make its behaviour clear
image
"Set Current Faders to New Client Level" will be the behaviour that is intended - it will not take any stored fader levels into account.  (Of course, these fader levels will then apply should a "current" client leave and rejoin.)

* Edit -> Stored Fader Settings
This new sub-menu replaces "Clear All Stored Solo and Mute Settings" and adds finer-grained control, including allowing all settings to be purged:
image
image

CHANGELOG: Client: Fader Settings improvements

**Context: Fixes an issue?**

Fixes: #1420 

Yes!
- App translations are affected
- Site documentation is affected

It runs and looks okay.  I don't generally worry about stored fader settings, so I've not got much to test on.  It could do with several people who use the feature heavily looking at it.

@pgScorpio, hopefully this is what you were after.

* [ ] Code review.
* [ ] More testing by people who use the feature.
* [ ] Usability feedback.

## Checklist

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -298,9 +298,17 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // Edit menu  --------------------------------------------------------------
     QMenu* pEditMenu = new QMenu ( tr ( "&Edit" ), this );
 
-    pEditMenu->addAction ( tr ( "Clear &All Stored Solo and Mute Settings" ), this, SLOT ( OnClearAllStoredSoloMuteSettings() ) );
+    QMenu* pFaderMenu = new QMenu ( tr ( "Stored &Fader Settings" ), this );
 
-    pEditMenu->addAction ( tr ( "Set All Faders to New Client &Level" ),
+    pEditMenu->addMenu ( pFaderMenu );
+
+    pFaderMenu->addAction ( tr ( "Clear &All Stored Fader Settings" ), this, SLOT ( OnClearFaderAllSettings() ) );
+    pFaderMenu->addAction ( tr ( "Clear Stored &Level Settings" ), this, SLOT ( OnClearFaderLevelSettings() ) );
+    pFaderMenu->addAction ( tr ( "Clear Stored &Solo Settings" ), this, SLOT ( OnClearFaderSoloSettings() ) );
+    pFaderMenu->addAction ( tr ( "Clear Stored &Mute Settings" ), this, SLOT ( OnClearFaderMuteSettings() ) );
+    pFaderMenu->addAction ( tr ( "Clear Stored &Group ID Settings" ), this, SLOT ( OnClearFaderGroupIdSettings() ) );
+
+    pEditMenu->addAction ( tr ( "Set Current Faders to New Client &Level" ),
                            this,
                            SLOT ( OnSetAllFadersToNewClientLevel() ),
                            QKeySequence ( Qt::CTRL + Qt::Key_L ) );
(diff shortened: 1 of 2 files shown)


pgScorpio:
Unfortunately  this is NOT what I intendended.

Changing text does not change the behaviour.
See wanted behaviour at #1420.

Though adding the "clear stored" options can overcome our problem it will take two actions instead of one.
Also splitting the (very often used) reset mute and solo seams a bit odd to me, since this now, most of the time, will take two actions instead of one also. So maybe we could (re-)add a "clear (current) mute and solo settings" too ?

pljones:
OK, I must be not understanding what you're after.

Could you write it out long-hand for me, please.

ann0see:
I believe we might have two options then? Reset all client levels to default (= clear ini file = reset all faders) and reset currently connected clients to default?

gilgongo:
> I believe we might have two options then? Reset all client levels to default (= clear ini file = reset all faders) and reset currently connected clients to default?

What's the scenario in which you'd set the level, but then opt to apply that to current but **not** to new clients (or vice versa), or perhaps not at all? 

Also feels like these functions would be better on the same screen as the level setting, assuming the majority case is to set the level then apply it to clients in some way?

ann0see:
@pgScorpio should comment on that.

pgScorpio:
> > I believe we might have two options then? Reset all client levels to default (= clear ini file = reset all faders) and reset currently connected clients to default?
> 
> What's the scenario in which you'd set the level, but then opt to apply that to current but **not** to new clients (or vice versa, but I suppose that's when you don't want to change the current mix)? Maybe I'm confusing myself...

My idea too... That's why I just opted that "Set all levels to new client level" should clear stored levels too, nothing more than that. (And I really don't know why somebody decided otherwise and added all kinds of other functionality too.)

> 
> Also feels like these functions would be better on the same screen as the level setting, assuming the majority case is to set the level then apply it to clients in some way?

I can understand your feeling/assumption, though this is certainly not always the case... This function is often used to reset the mix between songs, without changing new client level.
I think, to be consistent, the reset function is quite ok at the current location... (Always accessible within two mouse clicks, and no need to open settings.)

pljones:
MacOS Legacy build failing for `dmg` build.
```
==> Downloading https://ghcr.io/v2/homebrew/core/create-dmg/manifests/1.2.1
==> Fetching create-dmg
==> Downloading https://ghcr.io/v2/homebrew/core/create-dmg/blobs/sha256:a1d0ce1a3da65fb140ef2740dd778066c9f26122f5fb58fd64f4200bb168fc85
==> Pouring create-dmg--1.2.1.all.bottle.tar.gz
🍺  /usr/local/Cellar/create-dmg/1.2.1: 12 files, 67.9KB
Creating disk image...
hdiutil: create failed - No child processes
Error: Process completed with exit code 1.
```

hoffie:
> MacOS Legacy build failing for dmg build.

unrelated to this PR. I've tried tracking it down with no success so farm

src/clientdlg.cpp
```diff
@@ -298,9 +298,17 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // Edit menu  --------------------------------------------------------------
     QMenu* pEditMenu = new QMenu ( tr ( "&Edit" ), this );
 
-    pEditMenu->addAction ( tr ( "Clear &All Stored Solo and Mute Settings" ), this, SLOT ( OnClearAllStoredSoloMuteSettings() ) );
+    QMenu* pFaderMenu = new QMenu ( tr ( "Stored &Fader Settings" ), this );
```
softins, line 301: `&F` is already used in the same menu for `Auto-Adjust all &Faders`

src/clientdlg.cpp
```diff
@@ -298,9 +298,17 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
...
 
-    pEditMenu->addAction ( tr ( "Set All Faders to New Client &Level" ),
+    pEditMenu->addMenu ( pFaderMenu );
```
softins, line 303: This appears to be the first time Jamulus has used a cascading menu. Works fine on desktop devices, but do we know if it is friendly to Android and iOS?

pljones:
Here's a screenshot from my phone
Screenshot_20240328-175036

src/clientdlg.cpp
pljones: Here's a screenshot from my phone
Screenshot_20240328-175036

pljones: The iconography isn't clear -- the higher level menu doesn't indicate that the first item is a sub-menu.  But that's par for the course with Qt on Android.

ann0see:
This has been lingering for quite some time...

PR #3159, opened 2023-08-28
Fixes #1421: Add "delete server" button to connect dialog

### LINKED ISSUE #1421: Add option to delete stored servers from combo lists

pljones:

Adds a "delete server" tool button to the connect dialog, in the same style as the server dialog, to delete the currently selected server.  There's no "undo".
image

Selected server gets deleted.  Next server in list is then current in field.  On exit, the updated list is saved.

CHANGELOG: Add "delete server" button to connect dialog

**Context: Fixes an issue?**

Fixes #1421

Yes - screenshots and translation strings need updating.

Tested locally and appears to work as expected.

Code and UI review.

## Checklist

### DIFF
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -96,6 +96,11 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     cbxServerAddr->setAccessibleName ( tr ( "Server address edit box" ) );
     cbxServerAddr->setAccessibleDescription ( tr ( "Holds the current server address. It also stores old addresses in the combo box list." ) );
 
+    butDeleteServerAddr->setAccessibleName ( tr ( "Delete server address button" ) );
+    butDeleteServerAddr->setWhatsThis ( "<b>" + tr ( "Delete Server Address" ) + ":</b> " +
+                                        tr ( "Click the button to clear the currently selected server address "
+                                             "and delete it from the list of stored servers." ) );
+
     UpdateDirectoryComboBox();
 
     // init server address combo box (max MAX_NUM_SERVER_ADDR_ITEMS entries)
@@ -178,6 +183,9 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     QObject::connect ( butCancel, &QPushButton::clicked, this, &CConnectDlg::close );
 
     QObject::connect ( butConnect, &QPushButton::clicked, this, &CConnectDlg::OnConnectClicked );
+
+    // tool buttons
+    QObject::connect ( butDeleteServerAddr, &QPushButton::clicked, this, &CConnectDlg::OnDeleteServerAddrClicked );
 
     // timers
     QObject::connect ( &TimerPing, &QTimer::timeout, this, &CConnectDlg::OnTimerPing );
@@ -729,6 +737,31 @@ void CConnectDlg::OnConnectClicked()
 
     // tell the parent window that the connection shall be initiated
     done ( QDialog::Accepted );
+}
+
+void CConnectDlg::OnDeleteServerAddrClicked()
+{
+    if ( cbxServerAddr->currentText().isEmpty() )
+    {
+        return;
+    }
+
+    // move later items down one
+    for ( int iLEIdx = 0; iLEIdx < MAX_NUM_SERVER_ADDR_ITEMS - 1; iLEIdx++ )
+    {
+        while ( pSettings->vstrIPAddress[iLEIdx].compare ( cbxServerAddr->currentText() ) == 0 )
+        {
+            for ( int jLEIdx = iLEIdx + 1; jLEIdx < MAX_NUM_SERVER_ADDR_ITEMS; jLEIdx++ )
+            {
+                pSettings->vstrIPAddress[jLEIdx - 1] = pSettings->vstrIPAddress[jLEIdx];
+            }
+        }
+    }
+    // empty last entry
+    pSettings->vstrIPAddress[MAX_NUM_SERVER_ADDR_ITEMS - 1] = QString();
+
+    // redisplay to pick up updated list
+    showEvent ( nullptr );
 }
 
 void CConnectDlg::OnTimerPing()
--- a/src/connectdlgbase.ui
+++ b/src/connectdlgbase.ui
@@ -107,6 +107,30 @@
        </property>
       </widget>
      </item>
+     <item>
+      <widget class="QPushButton" name="butDeleteServerAddr">
+       <property name="sizePolicy">
+        <sizepolicy hsizetype="Maximum" vsizetype="Ignored">
+         <horstretch>0</horstretch>
+         <verstretch>0</verstretch>
+        </sizepolicy>
+       </property>
+       <property name="maximumSize">
+        <size>
+         <width>24</width>
+         <height>16777215</height>
+        </size>
+       </property>
+       <property name="font">
+        <font>
+         <bold>true</bold>
+        </font>
+       </property>
+       <property name="text">
+        <string notr="true">⌫</string>
+       </property>
+      </widget>
+     </item>
     </layout>
    </item>
    <item>
--- a/src/connectdlg.h
+++ b/src/connectdlg.h
@@ -101,6 +101,7 @@ public slots:
     void OnExpandAllStateChanged ( int value ) { ShowAllMusicians ( value == Qt::Checked ); }
     void OnCustomDirectoriesChanged();
     void OnConnectClicked();
+    void OnDeleteServerAddrClicked();
     void OnTimerPing();
     void OnTimerReRequestServList();
 


ann0see:
@gilgongo Could you please comment on the UX here? How can we avoid confusion with the connect button?

src/connectdlg.cpp
```diff
@@ -731,6 +740,12 @@ void CConnectDlg::OnConnectClicked()
     done ( QDialog::Accepted );
 }
 
+void CConnectDlg::OnDeleteServerAddrClicked()
+{
+    pSettings->vstrIPAddress.erase ( pSettings->vstrIPAddress.begin() + cbxServerAddr->currentIndex() );
```
softins, line 745: Having built on my pi, clicking the Delete button crashes the app, reliably. If I comment out this line and rebuild, it doesn't crash (but of course doesn't delete). Will look more closely.

pljones: What Qt version are you building with?  I tested with Qt6 only.

softins: 5.15.2 on the Pi. I don't think it's the Qt version. My suspicion is that `vstrIPAddress.erase` is changing the size of `vstrIPAddress`, which the rest of the code doesn't expect. Haven't proved or disproved it yet. I see there are no calls to an `erase()` method on a vector anywhere else in the code.

pljones: Mmmm, reading the C++ definition now, I don't see why I used that operation.

softins: Just about to test an alternative.

pljones: OK, replaced with a different approach which seems to work.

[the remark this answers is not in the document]
pljones: Let's assume
`int iIdx = cbxServerAddr->currentIndex();` returns `MAX_NUM_SERVER_ADDR_ITEMS - 1` (which is the highest it can get to).
```
pSettings->vstrIPAddress[iIdx] = pSettings->vstrIPAddress[iIdx + 1];
```
This will then be indexing off the end of the list.
```
pSettings->vstrIPAddress[MAX_NUM_SERVER_ADDR_ITEMS - 1] = "";
```
This would then correctly set the last entry to empty.

softins:
> @gilgongo Could you please comment on the UX here? How can we avoid confusion with the connect button?

Personally, I would prefer something like a red X instead of a black backspace symbol.

src/connectdlg.cpp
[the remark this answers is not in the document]
softins: > Let's assume `int iIdx = cbxServerAddr->currentIndex();` returns `MAX_NUM_SERVER_ADDR_ITEMS - 1` (which is the highest it can get to).
> 
> ```
> pSettings->vstrIPAddress[iIdx] = pSettings->vstrIPAddress[iIdx + 1];
> ```
> 
> This will then be indexing off the end of the list.

No it won't, because the `for` condition will fail first. Note that the condition both specifies `MAX_NUM_SERVER_ADDR_ITEMS - 1` *and* uses `<` not `<=`. For that very reason.

> ```
> pSettings->vstrIPAddress[MAX_NUM_SERVER_ADDR_ITEMS - 1] = "";
> ```
> 
> This would then correctly set the last entry to empty.

pljones:
> Personally, I would prefer something like a red X instead of a black backspace symbol.

Should we change the Server UI, too?  (I didn't find a button with a red X on any more meaningful - though I admit the symbol chosen isn't particularly obvious.)

ann0see:
If it's changed here, it should be changed everywhere

src/connectdlg.cpp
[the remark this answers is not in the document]
pljones: 🤦 yes...

Anyway - the other issue I had was with `cbxServerAddr->currentIndex()` - a couple of times it seemed to point to the wrong item but I couldn't reproduce the problem reliably.  The string compare method I've pushed should
- find all multiple matches (just in case some have crept in)
- have no problems with entering a new value, _not_ hitting enter but hitting delete entry (I think this may have caused deleting the wrong entry...)
- have no problems with indexing the list

I've not yet checked but I'm hoping on write, all entries (including empty get written) but on read, only the non-empty ones get into current settings - so on next write, only the non-empty ones should be stored...  Ideally, on write, only non-empty ones would get written.

softins:
> > Personally, I would prefer something like a red X instead of a black backspace symbol.
> 
> Should we change the Server UI, too? (I didn't find a button with a red X on any more meaningful (I think "stop" or "wrong" rather than "delete") - though I admit the symbol chosen isn't particularly obvious.)

Ah, I hadn't remembered it is already used in the Server UI (I so seldom use a server with a GUI!)

src/connectdlg.cpp
```diff
@@ -731,6 +740,21 @@ void CConnectDlg::OnConnectClicked()
...
+    if ( !cbxServerAddr->currentText().isEmpty() )
+    {
+        for ( int iLEIdx = 0; iLEIdx < MAX_NUM_SERVER_ADDR_ITEMS; iLEIdx++ )
+        {
+            if ( pSettings->vstrIPAddress[iLEIdx].compare ( cbxServerAddr->currentText() ) == 0 )
+            {
+                pSettings->vstrIPAddress[iLEIdx] = QString();
+            }
+        }
+    }
```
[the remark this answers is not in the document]
pljones: We shouldn't write empty entries, then.

pljones:
> Ah, I hadn't remembered it is already used in the Server UI (I so seldom use a server with a GUI!)

Yeah, it means "clear field" there (where the user can't directly type into the field).

src/connectdlg.cpp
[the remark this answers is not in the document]
pljones: I guess something similar to the entry squash could be done using the value...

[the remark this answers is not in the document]
softins: Or else stop at an empty entry when bumping them up to make room for the new entry.

I can't answer the question about `currentIndex()`, but I preferred my approach in general. But it's no big deal so long as it works without undesirable side-effects.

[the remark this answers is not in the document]
pljones: I've updated again.

The thing with `currentIndex()` is it's invalid if you type a prospective new entry but it doesn't get into the list.  It's still on the most recently selected index.  So select the third entry, type a new entry, click delete - did you mean to delete that third entry?  No.  So you need to check the text in the box matches the entry in the list - which might be another entry earlier or later than the most recently selected.

ann0see:
✗ https://www.compart.com/en/unicode/U+2717

⊗

Could be options

pljones:
As it says on the last reference, "X" tends to imply cancel - rather than remove or delete.  So I'm not keen on anything looking _just_ like an "X".  The character I've used is called "erase to the left", which is - essentially - what it does.  https://www.fileformat.info/info/unicode/char/232b/index.htm

ann0see:
Agree. You just don't see that often.

ann0see:
Just a behavioral discussion point: If the entry is emptied with backspace, and then the new button is clicked, we jump to another entry from the list. I wouldn't expect that. Maybe the button should be grayed out then? Or just do nothing?

ann0see:
(Also if we add this, the UX should be the same for custom directories, I'd say.)

pljones:
> (Also if we add this, the UX should be the same for custom directories, I'd say.)

That would want a separate issue raised... though I agree.  The existing solution is inelegant.

pljones:
> Just a behavioral discussion point: If the entry is emptied with backspace, and then the new button is clicked, we jump to another entry from the list. I wouldn't expect that. Maybe the button should be grayed out then? Or just do nothing?

"Do nothing" is possible -- currently the redisplay happens anyway (as if the connect window had been closed then opened again).

ann0see:
Ok. Then I think the do nothing should be implemented.

softins (approved):
Seems to work fine now.

pljones:
> Ok. Then I think the do nothing should be implemented.

Done.

pljones:
Here's a comparison with the "Clear text" button in Calibre:
image
which, I think, is close enough to the glyph used here (inverse colours, bold text, maybe - and overlayed in the input box, which could be fun to replicate in Qt).

pljones:
Oh, and apparently Qt Designer uses the same glyph.  I wonder where it gets it...
image

ann0see:
So there's maybe a different text input field then?

pljones:
> So there's maybe a different text input field then?

It could be a custom widget in Calibre.

src/connectdlgbase.ui
```diff
@@ -107,6 +107,9 @@
        </property>
       </widget>
      </item>
+     <item>
```
ann0see, line 110: The button looks a bit odd on macOS
macOS connect dialog with backspace deletion button, 

can you use another one, like the cancel button?

pljones: I've looked through the Qt options and didn't find anything else.

ann0see: So you can't just use the same button?

pljones: OK, now you've lost me.

ann0see: I meant that Qt creator should show the button (Cancel, Connect below the ip/server input field) in the gui. Can't you use the same button style for the button you added?

gilgongo: Do you mean there would be three buttons? "Cancel", "Delete", "Connect"? I'd be slightly worried about fat-fingering the delete my mistake perhaps.

ann0see: Sorry for being unclear. But I meant that the button "Cancel" has some styling. This styling should be applied to the new button which is added here (QPushButton not QToolButton)

pljones: OK, so that's what it looks like as a QToolButton
image
If I literally replace "QToolButton" with "QPushButton" in the `.ui` file, it looks like this:
image
(Don't ask me why it's that big, I've no idea...)

With a bit of a struggle I can get it looking like this in Qt Creator
image
but at run time, it still ends up looking like this
image

In all cases, it's taller than the QPushButtons.

pljones:
Aha!
image
That looks better.

src/connectdlgbase.ui
pljones: Aha!
image
That looks better.

ann0see: Will check on macOS.

ann0see: Seems as if nothing changed on macOS

pljones: Can you post another screenshot, please?

pljones: @ann0see if you get a chance to look at this, the screenshot would help me understand what you're seeing on MacOS.

ann0see: Sorry for the delay, but it looks basically the same:
Jamulus delete button screenshot

pljones: Well, at least it's centred vertically...  I can try dropping the vertical max size.
```
       <property name="sizePolicy">
        <sizepolicy hsizetype="Maximum" vsizetype="Ignored">
         <horstretch>0</horstretch>
         <verstretch>0</verstretch>
        </sizepolicy>
       </property>
       <property name="maximumSize">
        <size>
         <width>32</width>
         <height>16777215</height>
        </size>
```
`<sizepolicy hsizetype="Maximum" vsizetype="Ignored">` to `<sizepolicy hsizetype="Maximum" vsizetype="Maximum">` and `<property name="maximumSize">`->`<size>`->`<height>16777215</height>` to 24.

(I really just want to tell it "no smaller than the glyph, fill the height of the line that's already there".  But nothing I've read tells me how to do that.)

pljones:
Reverting to https://github.com/jamulussoftware/jamulus/commit/10c5931f5b714996a1282d207259cc854b9aa678, pretty much, but with width 24, bold and using the unicode glyph directly.

**Linux local builds:** QMake version 3.1, Qt version 6.6.0 and Qt version 5.15.2
and
**Linux github artifact:** Qt version 5.9.5
image
image
Server address drop down and new button same height.  The text colour doesn't match the drop down arrow colour, unfortunately -- and, from what I can see in the manual, you can't refer to the widget colours, so I'd have to hard code it, which would break Qt theming.



**Windows github artifact:**
(My Windows local builds for Qt 6.6.0 crash out.)
image
image
image
Not perfect but close.

----

Given that Qt seems unable to provide a consistent, cross-platform experience for this, we're going to have to compromise.

ann0see:
BTW: can you look at your git configuration? I believe something is wrong with your GPG key.

pljones:
> BTW: can you look at your git configuration? I believe something is wrong with your GPG key.

I've been trying to fix it... I know _why_ it broke, just too long since I first got it working to remember what needs doing.

ann0see:
I believe it's in GitHubs' documentation about GPG keys.

pljones:
> I believe it's in GitHubs' documentation about GPG keys.

Not the bit I needed... I don't know if I've fixed it until I create a new commit, though...

softins:
@pljones what's the current status of this? I've just built it on my Pi and it seems to work fine.

I noticed that the branch `pljones/feature/add-tooltips-to-connectdlg` contains additions to this PR that haven't been pushed to it yet. Is it planned to add them before merging?

I also happened to notice that in *Settings / Advanced Setup* there is a combi-box with drop-down for multiple Custom Directories that could usefully have a similar button. I currently have two entries in that list, with no way I can see to remove one of them.

ann0see:
I believe it's ok except for the button. But it's not too important.

ann0see:
Probably going to approve it like this. I'm a bit unsure about the UX though the backspace icon is used in the server too, but maybe there's something clearer.

ann0see (approved):

ann0see:
> However, the same trick didn't work in the Connect dialog.

@softins @pljones could you please open an issue on that?

pljones:
> @softins @pljones could you please open an issue on that?

Done.

pljones:
https://github.com/jamulussoftware/jamuluswebsite/issues/1002 raised for documentation.

PR #3187, opened 2023-10-06
Update jamulus-headless.service

rdica:
This should be sufficient but feel free to improve if I've missed something.

The program was being executed indirectly via a call from a bash shell. This shell was also intercepting SIGUSR2 signals used to start/stop recording from the command line per https://github.com/jamulussoftware/jamulus/issues/1515#issuecomment-819577918
This change updates the systemd service file to execute the program directly instead of from inside a shell.

CHANGELOG: Replace old program initialization that broke start/stop recording via command line with a direct call to the executable.

**Context: Fixes an issue?**

## Checklist

### DIFF
--- a/linux/debian/jamulus-headless.service
+++ b/linux/debian/jamulus-headless.service
@@ -16,8 +16,7 @@ IOSchedulingPriority=0
 
 #### Change this to publish this server, set genre, location and other parameters.
 #### See https://jamulus.io/wiki/Command-Line-Options ####
-ExecStart=/bin/sh -c 'exec /usr/bin/jamulus-headless -s -n'
-
+ExecStart=/usr/bin/jamulus-headless -s -n
 
 Restart=on-failure
 RestartSec=30


linux/debian/jamulus-headless.service
```diff
@@ -16,8 +16,7 @@ IOSchedulingPriority=0
 
 #### Change this to publish this server, set genre, location and other parameters.
 #### See https://jamulus.io/wiki/Command-Line-Options ####
-ExecStart=/bin/sh -c 'exec /usr/bin/jamulus-headless -s -n'
-
+ExecStart=/usr/bin/jamulus-headless -s -n
```
mcfnord, line 19: Should we ship code that won't run?

[My PR](https://github.com/jamulussoftware/jamulus/pull/3188/files) says:
ExecStart=/usr/bin/jamulus-headless -s -n -e anygenre2.jamulus.io:22224 -o "MyServer;MyLocation;MyTwoLetterCountry"

ann0see: No. The reason we don't is that there are already many public directories and the user should make a choice where/if they register.

pljones (approved):

ann0see:
I will test this hopefully soon on my server.

ann0see:
Thanks. 
In the meantime, please think of a changelog entry and edit your PR description with it.

rdica:
> Thanks. In the meantime, please think of a changelog entry and edit your PR description with it.

Done. Let me know if it's missing anything, or needs rewording, thanks.

linux/debian/jamulus-headless.service
mcfnord: I might be in agreement with that. But let's look at the user experience a bit more:

I just set up a headless server by following our instructions.

Now I have a running server that nobody can reach. I face a decision about listing on a directory (which opens the port), or opening the port manually. Following the instructions, I learn that [running on a directory is a diagnostic step](https://jamulus.io/wiki/Unregistered-Servers) even if I want an unregistered server. This is one reason I advocate for _[Minimum setup to run in Registered mode](https://jamulus.io/wiki/Running-a-Server)_, and am glad it's still here. I also continue believing the Installation section should conclude with a test step where the server is listed on a public directory using this CLI technique. It would be a 7th step, or a new 4th step, and could conclude with ctrl-C. This is a 'quickstart' approach where user success is decoupled from user learning. I am not sure exactly what this would look like. It just comes up when I look at this process with fresh eyes. If it's a test step buried rather deeply, perhaps it should be listed inline here.

rdica: The server administrator will decide when/where to publish their server entry. As @ann0see already pointed out.
Server operators will customize their configurations before going public, thats when they decide to publish via directories, or not.
Are you really advocating publishing a potentially insecure server on the internet _by default_?

mcfnord: This page advocates doing so, in its first sentences:

https://jamulus.io/wiki/Unregistered-Servers

Are we really advocating for this? Yes.

But if that server is potentially insecure, should we change that page? And could you tell me what that insecurity looks like?

rdica: Only the server administrator can make that kind of risk assessment, its definitely not our place to do that for them. Hence why they don't automatically register on install.

The documentation does mention some means of troubleshooting, but they are by no means the only ways. Feel feel free to update our documentation to include more ways, thanks.

mcfnord: The very first sentence to every person installing an unregistered server advises activity you claim is dangerous. I've written guidance for server administrators as my sole profession for 25 years. It often is our place to warn and advise. But sure, skip it. Wave some hands, make some claims, and endanger our users. All in a day's typey-typey, right?

ann0see (approved):
Ok. Sorry for the delay. But it works fine - as expected.

PR #3191, opened 2023-10-15
Minor refactoring of main cpp

pljones:

There are four very small changes to tidy up `src/main.ccp` in this pull request:
* Minor formatting changes
  This commit literally adjusts spacing for consistency.
* Adjust #ifdef block
  To avoid strange handling of a `{ ... }` block, I've adjusted a `#ifdef` position and duplicated a line of code.
* Adjust "use translations"
  At the point `bUseTranslation` is used, we will only use the result in the GUI, so code adjusted to avoid duplication.
* Use single application exec
  We always declare and assign `pApp` to something with an `exec()` method and we only create one instance that can be executed, so duplicate code removed for clarity.

CHANGELOG: Refactor src/main.cpp

**Context: Fixes an issue?**

Refactoring for good code hygiene.

No change to application behaviour.

Can be merged.

Can be merged after review.

## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -932,13 +932,6 @@ int main ( int argc, char** argv )
             CClientSettings Settings ( &Client, strIniFileName );
             Settings.Load ( CommandLineOptions );
 
-            // load translation
-            if ( bUseGUI && bUseTranslation )
-            {
-                CLocale::LoadTranslation ( Settings.strLanguage, pApp );
-                CInstPictures::UpdateTableOnLanguageChange();
-            }
-
 #    ifndef NO_JSON_RPC
             if ( pRpcServer )
             {
@@ -949,6 +942,13 @@ int main ( int argc, char** argv )
 #    ifndef HEADLESS
             if ( bUseGUI )
             {
+                // load translation
+                if ( bUseTranslation )
+                {
+                    CLocale::LoadTranslation ( Settings.strLanguage, pApp );
+                    CInstPictures::UpdateTableOnLanguageChange();
+                }
+
                 // GUI object
                 CClientDlg ClientDlg ( &Client,
                                        &Settings,
@@ -1015,7 +1015,7 @@ int main ( int argc, char** argv )
                 Settings.Load ( CommandLineOptions );
 
                 // load translation
-                if ( bUseGUI && bUseTranslation )
+                if ( bUseTranslation )
                 {
                     CLocale::LoadTranslation ( Settings.strLanguage, pApp );
                 }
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -599,20 +597,19 @@ int main ( int argc, char** argv )
         bUseGUI = false;
         qWarning() << "No GUI support compiled. Running in headless mode.";
     }
-#endif
 
-    if ( bIsClient )
+#endif
 #ifdef SERVER_ONLY
+    if ( bIsClient )
```
ann0see, line 603: Probably the initial idea was to avoid duplicating the if, but It should be ok either way.

src/main.cpp
```diff
@@ -954,6 +942,13 @@ int main ( int argc, char** argv )
 #    ifndef HEADLESS
             if ( bUseGUI )
             {
+                // load translation
+                if ( bUseTranslation )
```
ann0see, line 946: This will in theory remove the ability to translate CLI output, even if we might not use it yet/it doesn't work currently.

src/main.cpp
```diff
@@ -1019,7 +1015,7 @@ int main ( int argc, char** argv )
                 Settings.Load ( CommandLineOptions );
 
                 // load translation
-                if ( bUseGUI && bUseTranslation )
+                if ( bUseTranslation )
```
ann0see, line 1018: We might want to check how much of the CLI text is translated already.

ann0see (commented):
Main open question: 
State of translation of CLI

src/main.cpp
pljones: Yep, I can see why it was done.  I prefer that the code "looks right" without the pre-processor commands, though - there's less problem with IDEs, too.

src/main.cpp
pljones: It's for consistency, really.  Also, I don't think it was intentional to have it outside the `bUseGUI` block.

ann0see: I think that we once discussed this, but I'd probably be ok with it as is.

src/main.cpp
pljones: #3195 raised for this.

pljones:
> Main open question: State of translation of CLI

See #3195

ann0see (approved):
Ok. Otherwise it looks good. Thanks

pljones:
https://github.com/jamulussoftware/jamulus/actions/runs/6736061602

The MacOS Legacy build has failed twice... I'll kick it again.

PR #3229, opened 2024-02-10
Build CI on ubuntu 20 for Linux

softins:

These are the changes needed for the CI to build the three Linux artefacts using Ubuntu 20.04 instead of the EOL 18.04

CHANGELOG: Autobuild: Build Linux releases on Ubuntu 20.04 instead of 18.04

**Context: Fixes an issue?**

This is needed before dependabot #3168 can be applied successfully. `checkout@v4` will not run on Ubuntu 18.
It also has the advantage that Ubuntu 20.04 uses Qt 5.12 (U18.04 uses Qt 5.9), which will make it easier to progress #2393

The release notes should make it clear what versions of OS are compatible: Ubuntu >= 20, Debian >= 11

Ready to merge

Review

## Checklist

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -205,7 +205,7 @@ jobs:
           - config_name:            Linux .deb amd64 (artifacts+codeQL)
             target_os:              linux
             building_on_os:         ubuntu-22.04
-            building_container:     ubuntu:18.04
+            building_container:     ubuntu:20.04
             base_command:           ./.github/autobuild/linux_deb.sh
             run_codeql:             true
             is_main_build_target:   true
@@ -213,14 +213,14 @@ jobs:
           - config_name:            Linux .deb armhf (artifacts)
             target_os:              linux
             building_on_os:         ubuntu-22.04
-            building_container:     ubuntu:18.04
+            building_container:     ubuntu:20.04
             base_command:           TARGET_ARCH=armhf ./.github/autobuild/linux_deb.sh
             run_codeql:             false
 
           - config_name:            Linux .deb arm64 (artifacts)
             target_os:              linux
             building_on_os:         ubuntu-22.04
-            building_container:     ubuntu:18.04
+            building_container:     ubuntu:20.04
             base_command:           TARGET_ARCH=arm64 ./.github/autobuild/linux_deb.sh
             run_codeql:             false
 
@@ -293,7 +293,7 @@ jobs:
           xcode-version:            ${{ matrix.config.xcode_version }}
 
       - name:                       Set up base dependencies in container environment
-        if:                         matrix.config.building_container == 'ubuntu:18.04'
+        if:                         matrix.config.building_container == 'ubuntu:20.04'
         run:                        |
           set -eu
           apt-get -qq update
--- a/.github/autobuild/linux_deb.sh
+++ b/.github/autobuild/linux_deb.sh
@@ -75,7 +75,7 @@ setup_cross_compiler() {
     if [[ "${TARGET_ARCH}" == amd64 ]]; then
         return
     fi
-    local GCC_VERSION=7  # 7 is the default on 18.04, there is no reason not to update once 18.04 is out of support
+    local GCC_VERSION=9  # 9 is the default on 20.04, there is no reason not to update once 20.04 is out of support
     sudo apt install -qq -y --no-install-recommends "g++-${GCC_VERSION}-${ABI_NAME}" "qt5-qmake:${TARGET_ARCH}" "qtbase5-dev:${TARGET_ARCH}" "libjack-jackd2-dev:${TARGET_ARCH}" "qtmultimedia5-dev:${TARGET_ARCH}"
     sudo update-alternatives --install "/usr/bin/${ABI_NAME}-g++" g++ "/usr/bin/${ABI_NAME}-g++-${GCC_VERSION}" 10
     sudo update-alternatives --install "/usr/bin/${ABI_NAME}-gcc" gcc "/usr/bin/${ABI_NAME}-gcc-${GCC_VERSION}" 10
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,7 +60,7 @@ We support the following platforms and versions:
 
 - **Windows 10** or later
 - **macOS 10.10** or later
-- **Ubuntu 18.04** or later, **Debian 10** or later, most Linux flavors with recent enough Qt versions
+- **Ubuntu 20.04** or later, **Debian 11** or later, most Linux flavors with recent enough Qt versions. (Currently, it may still build on Ubuntu 18.04 and Debian 10, but the binaries built by Github will not run on those versions)
 
 _While Android and iOS aren't officially supported, please don't break their builds._
 
--- a/linux/debian/control
+++ b/linux/debian/control
@@ -14,10 +14,9 @@ Vcs-Browser: https://github.com/jamulussoftware/jamulus
 
 Package: jamulus
 Architecture: any
-# Define dependencies explicitly for the best compatibility across all supported Ubuntu/Debian versions.
-# The automatism would otherwise select package names or versions which are not available on some systems,
-# especially when run in the Github Ubuntu 18.04 build environment.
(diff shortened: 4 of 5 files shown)


.github/workflows/autobuild.yml
```diff
@@ -205,22 +205,22 @@ jobs:
...
-            building_container:     ubuntu:18.04
+            building_container:     ubuntu:20.04
             base_command:           TARGET_ARCH=armhf ./.github/autobuild/linux_deb.sh
             run_codeql:             false
 
           - config_name:            Linux .deb arm64 (artifacts)
             target_os:              linux
             building_on_os:         ubuntu-22.04
-            building_container:     ubuntu:18.04
+            building_container:     ubuntu:20.04
```
ann0see, line 223: I think this should be tested on older Ubuntu versions. We had problems with pan pots not showing up due to something being broken with the dependencies. 

@hoffie do you remember?

softins: I think we would be justified in specifying that the next version requires Ubuntu 20.04 or newer. I can easily check that the CI build runs on 20.04. I will need to create a 22.04 VM on VirtualBox before I can check it on that version.

I don't think we need to continue supporting 18.04.

softins: Hmmm, it looks like VirtualBox only provides 44100Hz audio to the guest OS, and not 48000Hz. So it might need a physical PC to test it. Maybe USB passthrough from host (Windows) to guest (Ubuntu) might allow a 48kHz audio device?

softins: OK, I managed to set up jack with a dummy device at 48000Hz, which allowed me to run the Jamulus client.

On Ubuntu 22, it does indeed fail to display the pan pot. Will try 20 as well.

softins: My mistake. I had not set the audio to Stereo. Once I had done so, the pan pot showed up fine on Ubuntu 22. Will try 20 too.

ann0see:
> It also has the advantage that Ubuntu 20.04 uses Qt 5.12 (U18.04 uses Qt 5.9)

I believe that we will need to drop official support for < 5.9 in the future. macOS legacy builds on 5.9, I think.

softins:
> > It also has the advantage that Ubuntu 20.04 uses Qt 5.12 (U18.04 uses Qt 5.9)
> 
> I believe that we will need to drop official support for < 5.9 in the future. macOS legacy builds on 5.9, I think.

Yes, I also have a branch waiting for me to raise a PR to build Mac Legacy with Qt 5.15.2 instead of 5.9.5. That will still support MacOS High Sierra 10.13 or newer, but will no longer support 10.10, 10.11 or 10.12. I think that will be fine.

.github/workflows/autobuild.yml
softins: OK, the artifact built on Ubuntu 20 from this PR's action runs fine on Ubuntu 20 and Ubuntu 22, and shows the pan pots correctly in both cases.

It does not run on Ubuntu 18, giving the following errors on invocation:
```
tony@ubuntu18desktop:~$ jamulus 
jamulus: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by jamulus)
jamulus: /usr/lib/x86_64-linux-gnu/libQt5Core.so.5: version `Qt_5.12' not found (required by jamulus)
```

I think this is fine. but we should make it clear somewhere that our future release builds need to run on Ubuntu 20 or newer.

softins: Done some further tests with the same artifact on different versions of Debian:
- Debian 12 (stable) (Qt 5.15.8): runs fine, pan pots show properly
- Debian 11 (oldstable) (Qt 5.15.2): runs fine, pan pots show properly
- Debian 10 (obsolete) (Qt 5.11.3): fails to run with the same errors as Ubuntu 18 above.

softins:
Given that the executable build on Ubuntu 20 with Qt 5.12.8 will not run on systems with a lower version of Qt than 5.12, we will also need to update the dependencies in `linux/debian/control` to prevent the .deb being installed on such a system:

https://github.com/jamulussoftware/jamulus/blob/e462529320537cef2881d61cc596820aef7d8694/linux/debian/control#L20

https://github.com/jamulussoftware/jamulus/blob/e462529320537cef2881d61cc596820aef7d8694/linux/debian/control#L34

softins:
I'll check the new .deb on the same systems above once it is built.

.github/workflows/autobuild.yml
ann0see: Ok. So this sounds good.

ann0see:
Yes. Maybe also have a look at the dependencies again. They might also want to be updated.

softins:
After adjusting the dependency versions for `libstdc++6` and `libgcc1` from 10.5 to 10.2 (see b7a84a8), the .deb correctly refuses to install on Debian 10 and Ubuntu 18, due to unsatisfied Qt version dependency. It installs and runs correctly on Debian 11 and 12, and on Ubuntu 20 and 22.

softins:
I did a test build with the auto-dependency variables reinstated:

```
Depends:
 ${shlibs:Depends},
 ${misc:Depends},
```

The .deb produced results in different dependencies, which I would assume are correct according to the package rules:

```
tony@pi:/tmp $ dpkg-deb -I jamulus_3.10.0dev-b6edd46_ubuntu_amd64.deb
 new Debian package, version 2.0.
 size 1385404 bytes: control archive=1132 bytes.
    1042 bytes,    19 lines      control
     628 bytes,     8 lines      md5sums
 Package: jamulus
 Version: 3.10.0dev
 Architecture: amd64
 Maintainer: "The Jamulus Development Team" <team@jamulus.io>
 Installed-Size: 3135
 Depends: libc6 (>= 2.29), libgcc-s1 (>= 3.0), libjack-jackd2-0 (>= 1.9.10+20150825) | libjack-0.125, libqt5core5a (>= 5.12.2), libqt5gui5 (>= 5.6.0~beta) | libqt5gui5-gles (>= 5.6.0~beta), libqt5multimedia5 (>= 5.6.0~beta), libqt5network5 (>= 5.8.0), libqt5widgets5 (>= 5.12.2), libqt5xml5 (>= 5.0.2), libstdc++6 (>= 7)
...
```

and

```
tony@pi:/tmp $ dpkg-deb -I jamulus-headless_3.10.0dev-b6edd46_ubuntu_amd64.deb
 new Debian package, version 2.0.
 size 1158968 bytes: control archive=1336 bytes.
    1024 bytes,    21 lines      control
     290 bytes,     4 lines      md5sums
     508 bytes,    17 lines   *  postinst             #!/bin/sh
     184 bytes,     7 lines   *  postrm               #!/bin/sh
     199 bytes,     7 lines   *  prerm                #!/bin/sh
 Package: jamulus-headless
 Source: jamulus
 Version: 3.10.0dev
 Architecture: amd64
 Maintainer: "The Jamulus Development Team" <team@jamulus.io>
 Installed-Size: 2292
 Depends: libc6 (>= 2.29), libgcc-s1 (>= 3.0), libqt5core5a (>= 5.12.2), libqt5network5 (>= 5.8.0), libqt5xml5 (>= 5.0.2), libstdc++6 (>= 7), adduser
...
```

I then did a test install of these packages on the three versions of Debian and three of Ubuntu. Debian 10 and Ubuntu 18 correctly refused to install due to dependencies. Debian 11/12 and Ubuntu 20/22 all installed and ran correctly.

So I will update `linux/debian/control` to use the automatic dependency variables again instead of manual versions.

softins:
Are there still concerns with this, or can we move forward with approval and merging?

ann0see:
Same issue as with macOS, but I think upgrading Linux is much easier. 
I'm going to check it works on Debian stable.

ann0see (approved):
Maybe also check that COMPILING.md is up to date.

ann0see:
CONTRIBUTING.md needs to be updated.

softins:
> Maybe also check that COMPILING.md is up to date.

I can't see anything in that file that would be affected by this change.

ann0see:
It's the minimum version in CONTRIBUTING.md

softins:
> It's the minimum version in CONTRIBUTING.md

Yes, indeed, I've just updated that. It was in COMPILING.md that I couldn't see any changes needed.

pljones (approved):

ann0see (approved):

Issue #3230, opened 2024-02-10
jamulus --version should show Qt version from dynamic libraries


softins:
When running `jamulus --version`, it displays the Qt version of the host on which it was compiled. It should display the Qt version of the dynamic Qt libraries being used on the running system, the same as **About Qt...** does.

The version displayed by `jamulus --version` is currently taken from `QT_VERSION_STR`, which was set at compile time.

_Originally posted by @softins in https://github.com/jamulussoftware/jamulus/issues/3228#issuecomment-1937056540_

PR #3231, opened 2024-02-10
Show both build and runtime versions of Qt

softins:

Displays both the version of Qt the executable was built with, and the version currently running via the dynamic libraries. These are available from `QT_VERSION_STR` and `qVersion()` respectively.

`jamulus --version` now shows something like:
```
 *** This app uses the following libraries, resources or code snippets:
 *** 
 *** Qt cross-platform application framework 5.9.5 (build), 5.15.2 (runtime)
 *** <https://www.qt.io>
```
**About Jamulus.. / Libraries** shows the same information.

CHANGELOG: Client/Server: Display Qt versions for both build and runtime.

**Context: Fixes an issue?**

Fixes: #3230 

No

Tested and ready to merge.

Just review

## Checklist

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -594,7 +594,8 @@ CAboutDlg::CAboutDlg ( QWidget* parent ) : CBaseDlg ( parent )
 
     // libraries used by this compilation
     txvLibraries->setText ( tr ( "This app uses the following libraries, resources or code snippets:" ) + "<p>" +
-                            tr ( "Qt cross-platform application framework" ) + QString ( " %1" ).arg ( QT_VERSION_STR ) +
+                            tr ( "Qt cross-platform application framework" ) + QString ( " %1 " ).arg ( QT_VERSION_STR ) + tr ( "(build)" ) +
+                            QString ( ", %1 " ).arg ( qVersion() ) + tr ( "(runtime)" ) +
                             ", <i><a href=\"https://www.qt.io\">https://www.qt.io</a></i>"
                             "</p>"
                             "<p>"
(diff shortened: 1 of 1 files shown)


src/util.cpp
```diff
@@ -594,7 +594,8 @@ CAboutDlg::CAboutDlg ( QWidget* parent ) : CBaseDlg ( parent )
 
     // libraries used by this compilation
     txvLibraries->setText ( tr ( "This app uses the following libraries, resources or code snippets:" ) + "<p>" +
-                            tr ( "Qt cross-platform application framework" ) + QString ( " %1" ).arg ( QT_VERSION_STR ) +
+                            tr ( "Qt cross-platform application framework" ) + QString ( " %1 " ).arg ( QT_VERSION_STR ) + tr ( "(build)" ) +
```
ann0see, line 597: Can we think of another name then running? Maybe runtime?

Running is a bit ambiguous

softins: Happy with "runtime", unless anyone else has a better suggestion?

softins: Or "current"? Or "now"?

pljones: I'd prefer "runtime" (as that's the general term -- runtime libraries).  End user is only expected to use this to cut-n-paste into a bug report, not understand the terminology or implications, after all.

pljones: What happens with a static build -- does it correctly report the same for both values?

softins: > What happens with a static build -- does it correctly report the same for both values?

Yes, it should, as it uses standard methods to retrieve the two values.

Although I think our binaries for Mac and Windows are not static builds, but they dynamically link to private libraries provided by the Jamulus installer. I'm not sure how to make a static build to test.

> (Ideally, the message could indicate whether the binary was running against dynamic libraries or was a static build. If dynamic libraries, we'd also want to know where they were, of course -- are they ones our installer installed, or are they supplied by the platform?)

I don't know how this would be achieved, nor whether it is really worth working out how?

ann0see (approved):
Seems ok now to me.

src/util.cpp
pljones: > > What happens with a static build -- does it correctly report the same for both values?
> 
> Yes, it should, as it uses standard methods to retrieve the two values.

OK

> Although I think our binaries for Mac and Windows are not static builds, but they dynamically link to private libraries provided by the Jamulus installer. I'm not sure how to make a static build to test.

You can download the source for Qt and create the object files and then do a static build in Qt Creator.  (I used to do this when brewing Jamulus before the current build system was in place.  With the dynamic build, using a locally built version isn't as easy -- much easier to push to a branch and run the GitHub builder - but much slower.)

> (Ideally, the message could indicate whether the binary was running against dynamic libraries or was a static build. If dynamic libraries, we'd also want to know where they were, of course -- are they ones our installer installed, or are they supplied by the platform?)
> 
> I don't know how this would be achieved, nor whether it is really worth working out how?

If it was trivial, it would be worth it as it would be useful for diagnostics, which is all any of this is _useful_ for.  (Quite apart from any licence compliance issues...)  But I've no idea how to do it, either 😄 .

softins: > But I've no idea how to do it, either 😄 .

OK, so are you able to approve it as it is?

PR #3249, opened 2024-03-22
Add JSON-RPC access to directories and server lists.

riban-bw:

Adds JSON-RPC notification of server lists from directories. Allows polling a directory for this info.

CHANGELOG: Added jamulusclient/pollServerList methods and jamulusclient/receivedServerList notification to JSON-RPC interface.

**Context: Fixes an issue?**

As discussed [here](https://github.com/orgs/jamulussoftware/discussions/2949#discussioncomment-8882899), it may be advantageous to increase the control and monitoring provided by the JSON-RPC interface. This provides access to server lists on directories.

This PR adds the associated documentation of the new method and notification in the JSON-RPC md page.

## Checklist

### DIFF
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -200,7 +200,7 @@ Results:
 | result.countryId | string | The musician’s country ID (see QLocale::Country). |
 | result.country | string | The musician’s country. |
 | result.city | string | The musician’s city. |
-| result.instrumentId | number | The musician’s instrument ID (see QLocale::Country). |
+| result.instrumentId | number | The musician’s instrument ID (see CInstPictures::GetTable). |
 | result.instrument | number | The musician’s instrument. |
 | result.skillLevel | string | Your skill level (beginner, intermediate, expert, or null). |
 
@@ -500,7 +500,7 @@ Parameters:
 | params.clients[*].countryId | number | The musician’s country ID (see QLocale::Country). |
 | params.clients[*].country | string | The musician’s country. |
 | params.clients[*].city | string | The musician’s city. |
-| params.clients[*].instrumentId | string | The musician’s instrument ID (see QLocale::Country).. |
+| params.clients[*].instrumentId | string | The musician’s instrument ID (see CInstPictures::GetTable). |
 | params.clients[*].instrument | string | The musician’s instrument. |
 
 
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -129,6 +129,57 @@ Results:
 | result.version | string | The Jamulus version. |
 
 
+### jamulus/pollServerList
+
+Request list of servers in a directory
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.directory | string | socket address of directory to query, e.g. anygenre1.jamulus.io:22124. |
+
+Results:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| result | string | "ok" or "error" if bad arguments. |
+
+
+### jamulusclient/connect
+
+Disconnect client from server
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.address | string | Server socket address (ip_addr:port). |
+
+Results:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| result | string | Always "ok". |
+
+
+### jamulusclient/disconnect
+
+Disconnect client from server
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params | object | No parameters (empty object). |
+
+Results:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| result | string | Always "ok". |
+
+
 ### jamulusclient/getChannelInfo
 
 Returns the client's profile information.
@@ -146,9 +197,9 @@ Results:
 | result.id | number | The channel ID. |
 | result.name | string | The musician’s name. |
 | result.skillLevel | string | The musician’s skill level (beginner, intermediate, expert, or null). |
-| result.countryId | number | The musician’s country ID (see QLocale::Country). |
+| result.country | string | The musician’s country. |
 | result.city | string | The musician’s city. |
-| result.instrument | string | The musician’s instrument. |
+| result.instrument | number | The musician’s instrument. |
 | result.skillLevel | string | Your skill level (beginner, intermediate, expert, or null). |
 
 
@@ -186,22 +237,6 @@ Results:
 | result.clients | array | The client list. See jamulusclient/clientListReceived for the format. |
 
 
-### jamulusclient/pollServerList
-
-Requests the server list from a specified directory.
-
-Parameters:
-
-| Name | Type | Description |
-| --- | --- | --- |
-| params.directory | string | Socket address of directory (hostname:port), e.g. anygenre1.jamulus.io:22124 |
-
-Results:
-
-| Name | Type | Description |
-| --- | --- | --- |
-| result | string | "ok" or "error" if invalid socket address |
-
 ### jamulusclient/sendChatText
 
 Sends a chat text message.
@@ -217,40 +252,6 @@ Results:
 | Name | Type | Description |
 | --- | --- | --- |
 | result | string | Always "ok". |
-
-
-### jamulusclient/connect
-
-Connect client to a server.
-
-Parameters:
-
-| Name | Type | Description |
-| --- | --- | --- |
-| params.address | string | Server socket address (hostname:port) |
-
-Results:
-
-| Name | Type | Description |
-| --- | --- | --- |
-| result | string | "ok" or error if invalid address.
-
-
-### jamulusclient/disconnect
-
-Disconnect client from server.
-
-Parameters:
-
-| Name | Type | Description |
-| --- | --- | --- |
-| params | object | No parameters (empty object). |
-
-Results:
-
-| Name | Type | Description |
-| --- | --- | --- |
-| result | string | Always "ok".
 
 
 ### jamulusclient/setName
@@ -308,9 +309,9 @@ Results:
 | result.clients[*].name | string | The client’s name. |
 | result.clients[*].jitterBufferSize | number | The client’s jitter buffer size. |
 | result.clients[*].channels | number | The number of audio channels of the client. |
-| result.clients[*].instrument | string | The instrument name provided by the user for this channel. |
+| result.clients[*].instrumentCode | number | The id of the instrument for this channel. |
 | result.clients[*].city | string | The city name provided by the user for this channel. |
-| result.clients[*].countryName | string | The country name provided by the user for this channel. |
+| result.clients[*].countryName | number | The text name of the country specified by the user for this channel (see QLocale::Country). |
 | result.clients[*].skillLevelCode | number | The skill level id provided by the user for this channel. |
 
 
@@ -499,34 +500,6 @@ Parameters:
 | params.clients[*].instrument | string | The musician’s instrument. |
 
 
-### jamulusclient/serverListReceived
-
-Emitted when the server list is received.
-
-Parameters:
-
-| Name | Type | Description |
-| --- | --- | --- |
-| params.servers | array | The server list. |
-| params.servers[*].address | string | The server's socket address (hostname:port). |
-| params.servers[*].name | string | The server’s name. |
-| params.servers[*].country | string | The servers’s country. |
-| params.servers[*].city | string | The server’s city. |
-
-
-### jamulusclient/serverInfoReceived
-
-Emitted when a server info is received.
-
-Parameters:
-
-| Name | Type | Description |
-| --- | --- | --- |
-| params.servers[*].address | string | The server's socket address (hostname:port). |
-| params.servers[*].pingTime | number | The round-trip ping time in ms. |
-| params.servers[*].numClients | number | The quantity of clients connected to the server . |
-
-
 ### jamulusclient/connected
 
 Emitted when the client is connected to the server.
@@ -557,6 +530,34 @@ Parameters:
 
 | Name | Type | Description |
 | --- | --- | --- |
-| params.state | number | The recorder state (see ERecorderState). |
+| params.state | number | The recorder state |
+
+
+### jamulusclient/serverInfoReceived
+
+Emitted when a server info is received.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.address | string | The server socket address |
+| params.pingtime | number | The round-trip ping time in ms |
+| params.numClients | number | The quantity of clients connected to the server |
+
+
+### jamulusclient/serverListReceived
+
+Emitted when the server list is received.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.servers | array | The server list. |
+| params.servers[*].address | string | Socket address (ip_address:port) |
+| params.servers[*].name | string | Server name |
+| params.servers[*].country | string | Server country |
+| params.servers[*].city | string | Server city |
 
 
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -186,6 +186,21 @@ Results:
 | result.clients | array | The client list. See jamulusclient/clientListReceived for the format. |
 
 
+### jamulusclient/pollServerList
+
+Requests server list.
+
+Parameters:
+
+| Name | Type | Description
+| --- | --- | --- |
+| params.directory | string | URL of directory, e.g. anygenre1.jamulus.io:22124 |
+
+Results:
+
+| Name | Type | Description |
+| result | string | "ok" or "error" if invalid URL |
+
 ### jamulusclient/sendChatText
 
 Sends a chat text message.
@@ -447,6 +462,21 @@ Parameters:
 | params.clients[*].countryId | number | The musician’s country ID (see QLocale::Country). |
 | params.clients[*].city | string | The musician’s city. |
 | params.clients[*].instrumentId | number | The musician’s instrument ID (see CInstPictures::GetTable). |
+
+
+### jamulusclient/serverListReceived
+
+Emitted when the server list is received.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.servers | array | The server list. |
+| params.servers[*].id | url | The server's URL. |
+| params.servers[*].name | string | The server’s name. |
+| params.servers[*].countryId | number | The servers’s country ID (see QLocale::Country). |
+| params.servers[*].city | string | The server’s city. |
 
 
 ### jamulusclient/connected
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -197,8 +197,10 @@ Results:
 | result.id | number | The channel ID. |
 | result.name | string | The musician’s name. |
 | result.skillLevel | string | The musician’s skill level (beginner, intermediate, expert, or null). |
+| result.countryId | string | The musician’s country ID (see QLocale::Country). |
 | result.country | string | The musician’s country. |
 | result.city | string | The musician’s city. |
+| result.instrumentId | number | The musician’s instrument ID (see QLocale::Country). |
 | result.instrument | number | The musician’s instrument. |
 | result.skillLevel | string | Your skill level (beginner, intermediate, expert, or null). |
 
@@ -495,8 +497,10 @@ Parameters:
 | params.clients[*].id | number | The channel ID. |
 | params.clients[*].name | string | The musician’s name. |
 | params.clients[*].skillLevel | string | The musician’s skill level (beginner, intermediate, expert, or null). |
+| params.clients[*].countryId | number | The musician’s country ID (see QLocale::Country). |
 | params.clients[*].country | string | The musician’s country. |
 | params.clients[*].city | string | The musician’s city. |
+| params.clients[*].instrumentId | string | The musician’s instrument ID (see QLocale::Country).. |
 | params.clients[*].instrument | string | The musician’s instrument. |
 
 
@@ -557,6 +561,7 @@ Parameters:
 | params.servers | array | The server list. |
 | params.servers[*].address | string | Socket address (ip_address:port) |
 | params.servers[*].name | string | Server name |
+| params.servers[*].countryId | string | Server country ID (see QLocale::Country). |
 | params.servers[*].country | string | Server country |
 | params.servers[*].city | string | Server city |
 
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -194,7 +194,7 @@ Parameters:
 
 | Name | Type | Description |
 | --- | --- | --- |
-| params.directory | string | Socket address of directory, e.g. anygenre1.jamulus.io:22124 |
+| params.directory | string | Socket address of directory (hostname:port), e.g. anygenre1.jamulus.io:22124 |
 
 Results:
 
@@ -217,6 +217,40 @@ Results:
 | Name | Type | Description |
 | --- | --- | --- |
 | result | string | Always "ok". |
+
+
+### jamulusclient/connect
+
+Connect client to a server.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.address | string | Server socket address (hostname:port) |
+
+Results:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| result | string | "ok" or error if invalid address.
+
+
+### jamulusclient/disconnect
+
+Disconnect client from server.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params | object | No parameters (empty object). |
+
+Results:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| result | string | Always "ok".
 
 
 ### jamulusclient/setName
@@ -474,10 +508,23 @@ Parameters:
 | Name | Type | Description |
 | --- | --- | --- |
 | params.servers | array | The server list. |
-| params.servers[*].address | string | The server's socket address. |
+| params.servers[*].address | string | The server's socket address (hostname:port). |
 | params.servers[*].name | string | The server’s name. |
-| params.servers[*].countryId | number | The servers’s country ID (see QLocale::Country). |
+| params.servers[*].country | string | The servers’s country. |
 | params.servers[*].city | string | The server’s city. |
+
+
+### jamulusclient/serverInfoReceived
+
+Emitted when a server info is received.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.servers[*].address | string | The server's socket address (hostname:port). |
+| params.servers[*].pingTime | number | The round-trip ping time in ms. |
+| params.servers[*].numClients | number | The quantity of clients connected to the server . |
 
 
 ### jamulusclient/connected
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -528,7 +528,7 @@ Parameters:
 
 ### jamulusclient/recorderState
 
-Emitted when the client is connected to a server who's recorder state changes.
+Emitted when a server's recorder state changes.
 
 Parameters:
 
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -192,7 +192,7 @@ Requests server list.
 
 Parameters:
 
-| Name | Type | Description
+| Name | Type | Description |
 | --- | --- | --- |
 | params.directory | string | URL of directory, e.g. anygenre1.jamulus.io:22124 |
 
@@ -473,7 +473,7 @@ Parameters:
 | Name | Type | Description |
 | --- | --- | --- |
 | params.servers | array | The server list. |
-| params.servers[*].id | url | The server's URL. |
+| params.servers[*].url | url | The server's URL. |
 | params.servers[*].name | string | The server’s name. |
 | params.servers[*].countryId | number | The servers’s country ID (see QLocale::Country). |
 | params.servers[*].city | string | The server’s city. |
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -199,6 +199,7 @@ Parameters:
 Results:
 
 | Name | Type | Description |
+| --- | --- | --- |
 | result | string | "ok" or "error" if invalid URL |
 
 ### jamulusclient/sendChatText
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -197,11 +197,11 @@ Results:
 | result.id | number | The channel ID. |
 | result.name | string | The musician’s name. |
 | result.skillLevel | string | The musician’s skill level (beginner, intermediate, expert, or null). |
-| result.countryId | string | The musician’s country ID (see QLocale::Country). |
+| result.countryId | number | The musician’s country ID (see QLocale::Country). |
 | result.country | string | The musician’s country. |
 | result.city | string | The musician’s city. |
 | result.instrumentId | number | The musician’s instrument ID (see CInstPictures::GetTable). |
-| result.instrument | number | The musician’s instrument. |
+| result.instrument | string | The musician’s instrument. |
 | result.skillLevel | string | Your skill level (beginner, intermediate, expert, or null). |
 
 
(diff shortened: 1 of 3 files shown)


docs/JSON-RPC.md
```diff
@@ -186,6 +186,22 @@ Results:
 | result.clients | array | The client list. See jamulusclient/clientListReceived for the format. |
 
 
+### jamulusclient/pollServerList
+
+Requests server list.
```
ann0see, line 191: Which Server List? Server list from directory.

docs/JSON-RPC.md
```diff
@@ -186,6 +186,22 @@ Results:
...
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.directory | string | URL of directory, e.g. anygenre1.jamulus.io:22124 |
```
ann0see, line 197: URL maybe call it host and port. Not really what I'd use understand d as URL

src/clientrpc.cpp
```diff
@@ -94,11 +94,67 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
...
+    /// @param {number} params.servers[*].countryId - Server country code
+    /// @param {string} params.servers[*].city - Server city
+    connect ( pClient->getConnLessProtocol(), &CProtocol::CLServerListReceived, [=] ( CHostAddress /* unused */, CVector<CServerInfo> vecServerInfo ) {
+        QJsonArray arrServerInfo;
+        for ( const auto& serverInfo : vecServerInfo )
+        {
+            QJsonObject objServerInfo{
+                { "url", serverInfo.HostAddr.toString() },
+                { "name", serverInfo.strName },
+                { "countryI", serverInfo.eCountry },
```
ann0see, line 111: Is that consistent how we handle countries elsewhere? Also why is it countryI? Not countryId?

src/clientrpc.cpp
```diff
@@ -94,11 +94,67 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
...
 
+    /// @rpc_method jamulus/pollServerList
+    /// @brief Request list of servers in a directory
+    /// @param {string} params.directory - URL of directory to query, e.g. anygenre1.jamulus.io:22124.
+    /// @result {string} result - "ok" or "error" if bad arguments.
+    pRpcServer->HandleMethod ( "jamulus/pollServerList", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        auto jsonDirectoryIp = params["directory"];
+        if ( !jsonDirectoryIp.isString() )
+        {
+            response["error"] = CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: directory is not a string" );
```
ann0see, line 135: The error wording needs some more thought to be less technical and more solution oriented.

ann0see (commented):
All in all good. Thanks. See my comments. Also @dtinth could you please have a look at this?

src/clientrpc.cpp
riban-bw: countryId is how country codes are handled elsewhere, e.g. in clientList.
"countryI" is a type - I will fix.

docs/JSON-RPC.md
riban-bw: I will add a more descriptive sentence.

docs/JSON-RPC.md
riban-bw: Okay, it does not include the protocol prefix. I will change this.

src/clientrpc.cpp
riban-bw: This is similar wording to the error given for existing method `jamulusclient/setName`. I tried to be consistent. Any suggestions for improvement welcome.

docs/JSON-RPC.md
riban-bw: I have used "address" for the parameter (which is what I had originally used but changed it!) and "socket address" for descriptions. According to [Wikipedia](https://en.m.wikipedia.org/wiki/Network_socket) a combination of IP address and port number is known as a socket address.

riban-bw:
Pushed enhancements in response to PR feedback.

src/clientrpc.cpp
riban-bw: On further consideration (and attempt to use in the field) I think this should be a string holding the country name as text. QLocale is QT specific and JSON-RPC clients may not be QT applications hence this id would be meaningless. I will change it in the PR. I think the country code used in client list should also be changed but that is a breaking change that is not directly related to this PR.

riban-bw:
I have changed from countryID (which is QT specific an not particulary useful to a JSON-RPC client that may not have access to the QLocale class) to country string. I have done this throughout including clientList, etc. for the same reason (it is of more use to a remote client.)

I have added connect / disconnect methods but am not triggering a signal to update the GUI. Any help here would be appreciated.

I have added a method to provide server ping time and quantity of connected clients. I poll for this from each server after receiving a serverList. I wonder if this is desirable behaviour. (It works well for my use case.)

I am aware that I added a function to the client class to expose its ConnLessProtocol. That feels wrong to me but I couldn't see an alternative method to receive CLServerListReceived signal. Maybe someone can validate this.

FYI these changes are working well when integrating Jamulus client into Zynthian.

ann0see:
> I have changed from countryID (which is QT specific an not particulary useful to a JSON-RPC client

I think we had a discussion about that in the past. I'm not sure what the consensus was. 

> have added connect / disconnect methods but am not triggering a signal to update the GUI. Any help here would be appreciated.

Seems not trivial. Does the UI look like it's not connected? I believe there's a Boolean flag to show the faders. But not sure.

ann0see:
Ah. I think the problem was translation. Therefore we used the ID over a string for e.g instruments. Otherwise there's a chance that someone who set the client to non English gets other values than someone who uses the English version

riban-bw:
> > I have changed from countryID (which is QT specific an not particulary useful to a JSON-RPC client
> 
> I think we had a discussion about that in the past. I'm not sure what the consensus was.

I could not find this discussion. In  a bid to be consistent and provide the most useful information to any client I have changed countryID to country text and instrumentID to instrument text in the JSON-RPC messages. We can't expect all clients to have access to these id/string maps. A client needs to be able to show the correct information. There may be an issue with i18n/l10n in which case we should use an internation standard for country code, e.g. ISO 3166-1. For instrument names we may prefer to change back to jamulus table to allow localisation within client (with associated local tables) although that could be done with strings too, i.e. a default English instrument name with clients able to map to local language.
 
> > have added connect / disconnect methods but am not triggering a signal to update the GUI. Any help here would be appreciated.
> 
> Seems not trivial. Does the UI look like it's not connected? I believe there's a Boolean flag to show the faders. But not sure.

I think the issue here is that there isn't a signal sent from the client to the client GUI when a connect occurs. There is an assumption that connect will be done only from the connect dialog which explicitly updates the client GUI. I am not sufficiently familiar with QT and jamulus code to best know how to refactor to implement such a signal.

I have added `jamulusclient/recorderState` notification so that JSON-RPC connected clients can indicate the recorer status of the connected server which seems important for privacy.

ann0see:
> There may be an issue with i18n/l10n in which case we should use an internation standard for country code, e.g. ISO 3166-1. For instrument names we may prefer to change back to jamulus table to allow localisation within client (with associated local tables) although that could be done with strings too

Yes, we need to be consistent and I think the ID was the non optimal agreement. please check what the string is if you change the la gauge to verify it gives back something useful.

softins:
> > There may be an issue with i18n/l10n in which case we should use an internation standard for country code, e.g. ISO 3166-1. For instrument names we may prefer to change back to jamulus table to allow localisation within client (with associated local tables) although that could be done with strings too
> 
> Yes, we need to be consistent and I think the ID was the non optimal agreement. please check what the string is if you change the la gauge to verify it gives back something useful.

I can't remember the discussion; it was probably while I was inactive for a time. But as mentioned above, using the ID allows for consistency between different language settings.

We already require a server operator to look up the correspondence between ID and Country, so they can specify the correct ID. I think so long as we document the list of IDs for Countries (as per Qt5) and Instruments (as per [util.cpp](https://github.com/jamulussoftware/jamulus/blob/2c19b5ede52db20ad79314026eee011908f49c0d/src/util.cpp#L1277)), then we can expect a JSON-RPC client developer to have reference to this list and include a suitable mapping in their client.

Maybe it would also be useful to output the Country name and Instrument name in any case, according to the current language that is set, but we should still include the ID as well, so the client developer can choose which to use. There's no harm having both.

softins:
> Maybe it would also be useful to output the Country name and Instrument name in any case, according to the current language that is set, but we should still include the ID as well, so the client developer can choose which to use. There's no harm having both.

On a side note: Having made the above point, I checked my [jamulus-php](https://github.com/softins/jamulus-php) to see what I had done there. I was not including the IDs, so I've just updated it additionally to include the IDs for Country, Instrument and Skill Level.

riban-bw:
I was away for a few days which delayed my response. I just tested and the instrument name (string) is sent in the language selected / configured in the client, e.g. changing from English to Spanish, (after a client restart) the client sends "oyente" instead of "listener". I think this works well. The JSON-RPC client will have the correct instrument name without requring a lookup table (with data redundency and associated risk of error and bloat).

The country is always in English, the same as the GUI client so this looks like it might be a compilation issue, i.e. not using the locale / lookup at runtime. So this looks like an issue with the client's core code, not with the JSON-RPC interface. I suggest this be corrected in the client core code so that users see the country name in their native / selected language.

We could send both numeric and text representation of the data (country, instrument) as suggested above but this is _data redundancy_ (generally considered a bad thing) within a rather verbose protocol (JSON) so I would prefer not. If the decision makers here decide this is desirable I will change the PR to include the codes too. I recommend keeping the strings as this provides the JSON-RPC client with useful data without the need for manually maintained look-up tables which adds undesirable and unnecessary code.

ann0see:
I'd argue that the API should not change language. 
I think we didn't want duplication and thus decided on the ID.

riban-bw:
If the API uses codes then the map from code to string should also be available via the API, e.g. `jamulus/getCountryCodes`. This would allow a client to create the correct text without needing to store a static map. The JSON-RPC client should not need to have detailed knowledge / tables of the server's internals. It should be able to control and monitor the remote device through function calls.

I could add these extra methods: `jamulus/getCountryNames`, `jamulus/getInstrumentNames`. Maybe `jamulus/getCountryName` with parameter `countryId` would return the name for the requested country.

ann0see:
yes. That makes sense. @dtinth any concerns?

ann0see:
@riban-bw could you please have a look at the code style starting here: https://github.com/jamulussoftware/jamulus/actions/runs/8416630554/job/23043980690?pr=3249#step:4:20 this needs some adjustments. You can most likely copy the patch directly from the output.

Otherwise, add the proposed messages. Usually, we'd prefer to have them in a separate PR or at least a clearly separated commit. This PR would then only focus on the initial changes you requested.

ann0see:
Also to make the JSON-RPC documentation check happy

> Please run ./tools/generate_json_rpc_docs.py to regenerate docs/JSON-RPC.md

docs/JSON-RPC.md
```diff
@@ -186,6 +186,22 @@ Results:
 | result.clients | array | The client list. See jamulusclient/clientListReceived for the format. |
 
 
+### jamulusclient/pollServerList
+
+Requests the server list from a specified directory.
```
ann0see, line 191: Sorry for my forgetfulness here - you need to update the python script I mentioned below and not edit the file directly

ann0see (changes requested):

docs/JSON-RPC.md
ann0see: Thanks!

src/clientrpc.cpp
ann0see: Something in terms of:

```suggestion
            response["error"] = CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "params has an invalid format: directory is not a string. Change the datatype to string." );
```

ann0see:
> I have added connect / disconnect methods but am not triggering a signal to update the GUI. Any help here would be appreciated.

I'm not sure how to do that unfortunately, but I think the code has "emit" somewhere. 

Otherwise, I'd probably look in the documentation (e.g. https://doc.qt.io/qt-6/signalsandslots.html). Maybe @pljones knows more about the GUI?

docs/JSON-RPC.md
riban-bw: I have run the script and pushed the changes.

ann0see:
Thanks. There's still a spacing error (remove space before the *) in client.h and a missing one in clientrpc.cpp

+        if ( pClient->SetServerAddr ( jsonAddr.toString() ) )

is required.

riban-bw:
> Also the ID should still be there. Then in a later PR you'd change them if needed.

I'm on it. Just checking compile then will push fixes.

src/clientrpc.cpp
```diff
@@ -125,17 +245,21 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
...
     /// @result {string} result.skillLevel - Your skill level (beginner, intermediate, expert, or null).
     pRpcServer->HandleMethod ( "jamulusclient/getChannelInfo", [=] ( const QJsonObject& params, QJsonObject& response ) {
         QJsonObject result{
             // TODO: We cannot include "id" here is pClient->ChannelInfo is a CChannelCoreInfo which lacks that field.
             { "name", pClient->ChannelInfo.strName },
             { "countryId", pClient->ChannelInfo.eCountry },
+            { "country", QLocale::countryToString ( pClient->ChannelInfo.eCountry ) },
             { "city", pClient->ChannelInfo.strCity },
-            { "instrumentId", pClient->ChannelInfo.iInstrument },
+            { "instrumentId",  pClient->ChannelInfo.iInstrument },
```
ann0see, line 261: There's a space too much.

src/clientrpc.cpp
```diff
@@ -125,17 +245,21 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
     /// @result {number} result.id - The channel ID.
     /// @result {string} result.name - The musician’s name.
     /// @result {string} result.skillLevel - The musician’s skill level (beginner, intermediate, expert, or null).
-    /// @result {number} result.countryId - The musician’s country ID (see QLocale::Country).
+    /// @result {string} result.countryId - The musician’s country ID (see QLocale::Country).
+    /// @result {string} result.country - The musician’s country.
     /// @result {string} result.city - The musician’s city.
-    /// @result {number} result.instrumentId - The musician’s instrument ID (see CInstPictures::GetTable).
+    /// @result {number} result.instrumentId - The musician’s instrument ID (see QLocale::Country).
```
ann0see, line 251: Is this correct?

src/clientrpc.cpp
```diff
@@ -54,8 +54,10 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
     /// @param {string} params.clients[*].name - The musician’s name.
     /// @param {string} params.clients[*].skillLevel - The musician’s skill level (beginner, intermediate, expert, or null).
     /// @param {number} params.clients[*].countryId - The musician’s country ID (see QLocale::Country).
+    /// @param {string} params.clients[*].country - The musician’s country.
     /// @param {string} params.clients[*].city - The musician’s city.
-    /// @param {number} params.clients[*].instrumentId - The musician’s instrument ID (see CInstPictures::GetTable).
+    /// @param {string} params.clients[*].instrumentId - The musician’s instrument ID (see QLocale::Country)..
```
ann0see, line 59: Please re check this. Probably it's still the table.

ann0see (changes requested):

src/clientrpc.cpp
riban-bw: This should be consistent. If I change the error description for this parameter (`pollServerList`) should the other, similarly described parameters also be changed, including pre-existing ones, like `setName`?

src/clientrpc.cpp
riban-bw: Fixed!

src/clientrpc.cpp
riban-bw: Fixed!

src/clientrpc.cpp
ann0see: Yes. True. But probably not in this PR then.

src/clientrpc.cpp
```diff
@@ -54,8 +54,10 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
     /// @param {string} params.clients[*].name - The musician’s name.
     /// @param {string} params.clients[*].skillLevel - The musician’s skill level (beginner, intermediate, expert, or null).
     /// @param {number} params.clients[*].countryId - The musician’s country ID (see QLocale::Country).
+    /// @param {string} params.clients[*].country - The musician’s country.
     /// @param {string} params.clients[*].city - The musician’s city.
-    /// @param {number} params.clients[*].instrumentId - The musician’s instrument ID (see CInstPictures::GetTable).
+    /// @param {string} params.clients[*].instrumentId - The musician’s instrument ID (see CInstPictures::GetTable).
```
ann0see, line 59:
```suggestion
    /// @param {number} params.clients[*].instrumentId - The musician’s instrument ID (see CInstPictures::GetTable).
```

src/clientrpc.cpp
```diff
@@ -94,11 +98,127 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
                                             } );
     } );
 
+    /// @rpc_notification jamulusclient/serverListReceived
+    /// @brief Emitted when the server list is received.
+    /// @param {array} params.servers - The server list.
+    /// @param {string} params.servers[*].address - Socket address (ip_address:port)
+    /// @param {string} params.servers[*].name - Server name
+    /// @param {string} params.servers[*].countryId - Server country ID (see QLocale::Country).
```
ann0see, line 106:
```suggestion
    /// @param {number} params.servers[*].countryId - Server country ID (see QLocale::Country).
```

src/clientrpc.cpp
```diff
@@ -125,17 +245,21 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
     /// @result {number} result.id - The channel ID.
     /// @result {string} result.name - The musician’s name.
     /// @result {string} result.skillLevel - The musician’s skill level (beginner, intermediate, expert, or null).
-    /// @result {number} result.countryId - The musician’s country ID (see QLocale::Country).
+    /// @result {string} result.countryId - The musician’s country ID (see QLocale::Country).
```
ann0see, line 248:
```suggestion
    /// @result {number} result.countryId - The musician’s country ID (see QLocale::Country).
```

ann0see (changes requested):

src/clientrpc.cpp
riban-bw: Doh! Fixed!

src/clientrpc.cpp
riban-bw: Fixed!

src/clientrpc.cpp
riban-bw: Fixed!

src/clientrpc.cpp
riban-bw: I will leave it as-is then. It is consistent. Someone can change the wording of the errors in a separate PR if they desire.

src/clientrpc.cpp
```diff
@@ -94,11 +98,127 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
...
+    /// @rpc_notification jamulusclient/serverListReceived
+    /// @brief Emitted when the server list is received.
+    /// @param {array} params.servers - The server list.
+    /// @param {string} params.servers[*].address - Socket address (ip_address:port)
+    /// @param {string} params.servers[*].name - Server name
+    /// @param {number} params.servers[*].countryId - Server country ID (see QLocale::Country).
+    /// @param {string} params.servers[*].country - Server country
+    /// @param {string} params.servers[*].city - Server city
+    connect ( pClient->getConnLessProtocol(),
+              &CProtocol::CLServerListReceived,
```
pljones, line 110: How often is an RPC client going to receive this message?

Given the nature of the server list (i.e. it's not guaranteed to be in any particular order or even complete), in what use cases would this message be consumed?  (Just thinking the JSON RPC doc might be expanded to explain - it's a bit brief currently.)

riban-bw: The list seems to be sent only after being requested by the poll. I am using it to populate a list as required on the basis that the list will be accurate at that time. It seems to work well.

ann0see (commented):
I will have a look at the functionality in the next few days (= local testing). Then most likely do the last review.

src/clientrpc.cpp
```diff
@@ -94,11 +98,127 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
...
+        {
+            response["error"] =
+                CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: directory is not a valid socket address" );
+        }
+
+        response["result"] = "ok";
+    } );
+
+    /// @rpc_method jamulusclient/connect
+    /// @brief Disconnect client from server
```
ann0see, line 183: The documentation is wrong.

```suggestion
    /// @brief Connect client to server
```

src/clientrpc.cpp
```diff
@@ -94,11 +98,127 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
...
+    /// @param {object} params - No parameters (empty object).
+    /// @result {string} result - Always "ok".
+    pRpcServer->HandleMethod ( "jamulusclient/disconnect", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        if ( pClient->IsRunning() )
+        {
+            pClient->Stop();
+        }
+
+        response["result"] = "ok";
```
ann0see, line 218: Why is this always ok and not an error if this is called on an already disconnected client?

src/clientrpc.cpp
```diff
@@ -126,16 +246,20 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
...
     /// @result {string} result.skillLevel - The musician’s skill level (beginner, intermediate, expert, or null).
     /// @result {number} result.countryId - The musician’s country ID (see QLocale::Country).
+    /// @result {string} result.country - The musician’s country.
     /// @result {string} result.city - The musician’s city.
     /// @result {number} result.instrumentId - The musician’s instrument ID (see CInstPictures::GetTable).
+    /// @result {string} result.instrument - The musician’s instrument.
     /// @result {string} result.skillLevel - Your skill level (beginner, intermediate, expert, or null).
     pRpcServer->HandleMethod ( "jamulusclient/getChannelInfo", [=] ( const QJsonObject& params, QJsonObject& response ) {
         QJsonObject result{
             // TODO: We cannot include "id" here is pClient->ChannelInfo is a CChannelCoreInfo which lacks that field.
```
ann0see, line 256: Typo, but not your issue.
```suggestion
            // TODO: We cannot include "id" here as pClient->ChannelInfo is a CChannelCoreInfo which lacks that field.
```

src/clientrpc.cpp
ann0see: It also gets issued if the connect dialog is open.

src/clientrpc.cpp
```diff
@@ -94,11 +98,127 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
...
+    } );
+
+    /// @rpc_method jamulusclient/disconnect
+    /// @brief Disconnect client from server
+    /// @param {object} params - No parameters (empty object).
+    /// @result {string} result - Always "ok".
+    pRpcServer->HandleMethod ( "jamulusclient/disconnect", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        if ( pClient->IsRunning() )
+        {
+            pClient->Stop();
```
ann0see, line 215: I believe that this should call Disconnect() here instead https://github.com/jamulussoftware/jamulus/blob/e12cb71abc2de9477095ebfdfe62f31f15e09513/src/clientdlg.cpp#L1244 if the UI is running.

src/clientrpc.cpp
```diff
@@ -94,11 +98,127 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
...
+    /// @param {string} params.address - Server socket address (ip_addr:port).
+    /// @result {string} result - Always "ok".
+    pRpcServer->HandleMethod ( "jamulusclient/connect", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        auto jsonAddr = params["address"];
+        if ( !jsonAddr.isString() )
+        {
+            response["error"] = CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: address is not a string" );
+            return;
+        }
+
```
ann0see, line 193: See https://github.com/jamulussoftware/jamulus/blob/e12cb71abc2de9477095ebfdfe62f31f15e09513/src/clientdlg.cpp#L1196 as below.

ann0see (changes requested):
The GUI stuff is a blocker in my opinion, but methods seem to be available.
We should probably check if the gui is running and then call different code.
Otherwise it's the right direction.
Sorry for the delay.

docs/JSON-RPC.md
```diff
@@ -471,3 +526,43 @@ Parameters:
 | params | object | No parameters (empty object). |
 
 
+### jamulusclient/recorderState
+
+Emitted when the client is connected to a server who's recorder state changes.
```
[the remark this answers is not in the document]
pljones: It only gets emitted when the client is connect to a server and that server's recorder state changes.  I think the first part is probably redundant -- i.e. it can be assumed that a client won't be informed of state changes to servers to which it isn't connected.

src/clientrpc.cpp
pljones: Nothing should call Dialog methods.  Ever.  The Dialog UI might not exist if th RPC interface is in use.

src/clientrpc.cpp
pljones: As above.

pljones:
If there's code in the Dialog UI that's needed by the RPC interface, it should be moved to the client itself.  There's a lot of code that's been put in the wrong place.  The Dialog UI layer should be dumb - just responding to UI events and updating state in the Client (or Server for the Server Dialog UI).

src/clientrpc.cpp
ann0see: We need to update the UI somehow. -> refactoring needed.

ann0see:
The only thing I can think of is to factor out the actual disconnection process and introduce a GUI method to set the UI to connected and disconnected. 
We might also use signals?

src/clientrpc.cpp
pljones: Yes.  If an a change to state happens to the Client (`CClient`), it should emit an event.  The UI (`CClientDlg`) can respond to that.  But the state should be in the Client itself, not the UI.  That way, it's accessible to both UI and RPC-I.

pljones:
> We might also use signals?

That's the only correct way to do it with Qt, in my view.

The UI is meant to respond to signals - either from the widget set or from CClient.  The response to a widget signal should be to update CClient.  The response to a CClient signal should be to update the widget layer (including shutting down CClientDlg, if appropriate and not otherwise handled).

ann0see:
Ok. Then the way to go is to also issue a signal for updating the UI which means that the UI update code needs to be refactored into a new method.

src/clientrpc.cpp
riban-bw: I was only following similar code elsewhere in the interface. A common question is whether we want an error when we ask for an action and the action gives the desired effect, e.g. disconnect from a device to which we are not connected results in being disconnected. Is this erroneous? I would guess other places this same,  alway "ok" result is used here allows such an action that will never fail to give an acknowledgement that the request was recevied and actioned.

src/clientrpc.cpp
riban-bw: Fixed in next commit.

src/clientrpc.cpp
riban-bw: Fixed in next commit.

docs/JSON-RPC.md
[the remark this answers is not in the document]
riban-bw: Fixed in next commit.

riban-bw:
This PR is to provide improved JSON-RPC interface. I may not be the right person to refactor the core code to improve UI/backend integration with better use of signals. I appreciate and agree with the plan but does that belong in this PR? I feel like there should be a separate bit of work to abstract the core client functionality from the UP to the client code. That may impact (improve) the remote control but feels wrong to bind it to this PR.

I am not clear what is outstanding on this PR. I think I have reviewed and fixed all suggestions (thanks) apart from refactoring the client and UI to implement better use of signals.

ann0see:
The issue is that the UI thing is a blocker for me.

pljones:
So we're suggesting:
- mixer UI and/or RPC interface issues signal to CClient to update fader state
- CClient updates fader state in CChannel and sends appropriate protocol message
- CClient receives protocol messages, updates CChannel state, which issues signal to indicate appropriate state change
- mixer UI and/or RPC interface listen for state change signals and respond appropriately

src/clientrpc.cpp
ann0see: Ok. Then we can leave it as is.

ann0see (commented):
I'm sorry to say that, but in my opinion we need to get the GUI stuff solved. I would suggest to revert the connection and disconnect behavior- which I know is **substential** out of this PR and get the other changes merged. Then open a new PR with only the connect/disconnect changes. Then we can work on the slot/signal idea there.

ann0see:
Ok. Can we get everything which doesn't break the UI in?

pljones:
If this can be split into one PR per change it would probably be easier to get it moving.  The changes where there are problems would then not block the changes where there aren't.

riban-bw:
> If this can be split into one PR per change it would probably be easier to get it moving. The changes where there are problems would then not block the changes where there aren't.

Many of those changes were added in response to reviewer's comments. Do you mean that each reviewer observation should be a seperate PR until the codebase is sufficiently clean to allow the actual PR? Many of the additional changes requested by reviewers relate to historical issues that I was requested to resolve so we do seem to be some way beyond the original PR. Some of these I do not have sufficient understanding of the whole codebase and I may not be the ideal canidate to resolve.

docs/JSON-RPC.md
```diff
@@ -129,6 +129,57 @@ Results:
...
+Results:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| result | string | "ok" or "error" if bad arguments. |
+
+
+### jamulusclient/connect
+
+Disconnect client from server
```
pljones, line 151: Surely this is
```suggestion
Connect from client to server
```

pljones:
I'm not sure what this means from the auto build
image

I'm actually not too unhappy with getting this merged as is, given:
- fix the above error
- fix the doc for connect

I'd consider whether or not leaving the ClientUI Connect dialog open is a problem for the JSON-RPC client a separate issue, to be looked at when thinking about properly refactoring the ClientUI so it doesn't hold state.

ann0see:
Please get rid of the connect/disconnect features then and put them in a separate PR.

ann0see:
Otherwise: do we find a compromise that the connect/disconnect methods only work headless?

ann0see:
@pljones do you think it makes sense extracting everything except for the connect and disconnect functionality out of this PR and then open a separate PR for just connection an disconnection? Then we do not depend on refactoring the connection/disconnection methods yet.

pljones:
I'd rather do no more work on the RPC client that didn't depend on _signals_ from CClient itself -- depending on signals from elsewhere (`CProtocol::CLServerListReceived` in this case) isn't something I'm keen on.  CClient should be exposing a cleaner API and we should be working on that _first_, in my view, at this stage.  The more we add workarounds using the existing APIs, the more work we create to undo in the future.

ann0see:
@pljones, @riban-bw I'd really like to have parts of this in the upcoming release, not 4.0. I will open a PR without the disconnect and connect methods and tag it for 3.12.

riban-bw:
It looks like https://github.com/jamulussoftware/jamulus/pull/3479 has superceeded this PR. Should this one (3249) be closed?

ann0see:
Yes.

Done.

PR #3252, opened 2024-03-28
Add tooltips to connectdlg

pljones:

Add tool tips to the Connect Dialog to improve useability.  I've aimed to keep the same translation strings as "What's this?" uses, as with the other dialogs.

CHANGELOG: Add tooltips to connectdlg

**Context: Fixes an issue?**

No.

No

Will need translations updated.

Ready to merge.

## Checklist

### DIFF
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -45,13 +45,13 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     QString strDirectoryWT = "<b>" + tr ( "Directory" ) + ":</b> " +
                              tr ( "Shows the servers listed by the selected directory. "
                                   "You can add custom directories in Advanced Settings." );
-    QString strDirectoryAN = tr ( "Directory combo box" );
-
-    lblList->setWhatsThis ( strDirectoryWT );
-    lblList->setToolTip ( strDirectoryWT );
-    lblList->setAccessibleName ( strDirectoryAN );
+    lblDirectory->setWhatsThis ( strDirectoryWT );
+    lblDirectory->setToolTip ( strDirectoryWT );
     cbxDirectory->setWhatsThis ( strDirectoryWT );
     cbxDirectory->setToolTip ( strDirectoryWT );
+
+    QString strDirectoryAN = tr ( "Directory combo box" );
+    lblDirectory->setAccessibleName ( strDirectoryAN );
     cbxDirectory->setAccessibleName ( strDirectoryAN );
 
     // filter
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -42,9 +42,9 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
 
     // Add help text to controls -----------------------------------------------
     // directory
-    QString strDirectoryWT = "<b>" + tr ( "Directory" ) + ":</b> " +
-                             tr ( "Shows the servers listed by the selected directory. "
-                                  "You can add custom directories in Advanced Settings." );
+    QString strDirectoryWT = "<b>" + tr ( "Directory" ) + ":</b> " + tr ( "The list of available directories." ) + "<br/>" +
+                             tr ( "Select a directory to list the servers registered with it." ) + "<br/>" +
+                             tr ( "You can add custom directories in Advanced Settings." );
     lblDirectory->setWhatsThis ( strDirectoryWT );
     lblDirectory->setToolTip ( strDirectoryWT );
     cbxDirectory->setWhatsThis ( strDirectoryWT );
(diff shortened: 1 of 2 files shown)


src/connectdlg.cpp
```diff
@@ -42,43 +42,51 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
...
     // server list view
-    lvwServers->setWhatsThis ( "<b>" + tr ( "Server List" ) + ":</b> " +
-                               tr ( "The Connection Setup window lists the available servers registered with "
-                                    "the selected directory. Use the Directory dropdown to change the directory, "
-                                    "find the server you want to join in the server list, click on it, and "
-                                    "then click the Connect button to connect. Alternatively, double click on "
-                                    "the server name to connect." ) +
-                               "<br>" + tr ( "Permanent servers (those that have been listed for longer than 48 hours) are shown in bold." ) +
-                               "<br>" + tr ( "You can add custom directories in Advanced Settings." ) );
+    QString strServersWT = "<b>" + tr ( "Server List" ) + ":</b> " +
```
ann0see, line 80: I'm a bit worried about the server list as it could be annoying to see a long text. Otherwise all is good.

ann0see: I'd suggest to shorten it drastically for the hover menu while leaving the what's this text as is.

pljones: OK, shorter version done.

src/connectdlg.cpp
```diff
@@ -80,6 +86,11 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
                                "<br>" + tr ( "Permanent servers (those that have been listed for longer than 48 hours) are shown in bold." ) +
                                "<br>" + tr ( "You can add custom directories in Advanced Settings." ) );
 
+    lvwServers->setToolTip ( "<b>" + tr ( "Server List" ) + ":</b> " +
+                             tr ( "The list of servers registered with the selected directory. "
```
ann0see, line 90: Maybe even shorter: Just mention the double click to join a server. The other option doesn't really need to be explained and is slower.

softins:
Hmm, I've just tried this out, compiling on my RPi. I must admit I found the tooltips continually popping up in the connect window rather obtrusive, particularly the tooltip within the list of servers, which would pop up anywhere I stopped moving the pointer.

Not sure what to suggest, but to my mind, such tooltips are hand-holding for the unfamiliar user, but obtrusive to anyone who has become familiar with the program, which doesn't take long.

ann0see:
Agree. Other places are maybe not that obstructive. Would you be ok with a shorter version or would you still be bothered by it?

pljones:
The reason I did this is we have tooltips on all the other dialogs, with the text the same as the "What's this?" text - so I felt we should be _consistent_ in our approach.  If it's wrong / unhelpful / distracting here, why is it working elsewhere?  Just because those other places aren't used so much?  Is it distracting there when they _are_ used?

Do people even think of the "What's this?" function?  (It's not particularly common.)

softins:
I haven't yet explored to see where else tooltips are used in Jamulus (I was going to do so as part of this review). But to my mind, "What's this?" and tooltips should provide different functions:

* "What's this?" is a user-initiated action, and it's very appropriate for the pop-up to provide some detailed explanation or description, over several lines.
* A tooltip pops up uninvited, and is intended to provide a short handy hint as to the identity or function of something that is otherwise non-obvious, such as an unlabelled icon. I don't feel that a box of explanation is appropriate for a tooltip, and it would rapidly become annoying.

I also wouldn't use a tooltip to describe a large area, such as the list of servers in the connect dialog. That was what I found the most annoying: the tooltip popping up every time I moved the mouse within the server list. And the tooltips for the Directory selector and filter box were also more explanatory than tooltippy to my mind, although definitely appropriate for "What's This?"

I understand the desire for consistency, but don't think it should trump benefit and usability. I've been struggling to understand who would benefit from the changes in this particular PR, and for how long.

pljones:
> A tooltip pops up uninvited

To me, it's meant to be invited - only if I hover long enough to be obvious I'm unsure what the control _does_ it I use it.

I'd agree they serve different purposes, though.  I'd expect "What's this?" to redirect the user to the appropriate paragraph in the manual, to be honest, to avoid redundancy in documentation.  The tooltip should be a "hint" as to usage.  We don't follow this approach, though.

ann0see:
One of the users I work with really liked the hover feature to explain the usage so it's valuable. Most likely what's this is much less often used.

src/connectdlg.cpp
ann0see: Server list: Double click a Server to connect.

softins:
> One of the users I work with really liked the hover feature to explain the usage - so it's valuable.

I can understand this for someone who is new and learning to use Jamulus. But as an experienced user, I would at least want the ability to turn it off, particularly for the wordy ones like in this PR.

src/connectdlg.cpp
```diff
@@ -42,31 +42,37 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
...
-    QString strDirectoryWT = "<b>" + tr ( "Directory" ) + ":</b> " +
-                             tr ( "Shows the servers listed by the selected directory. "
-                                  "You can add custom directories in Advanced Settings." );
-    QString strDirectoryAN = tr ( "Directory combo box" );
-
-    lblList->setWhatsThis ( strDirectoryWT );
-    lblList->setAccessibleName ( strDirectoryAN );
+    QString strDirectoryWT = "<b>" + tr ( "Directory" ) + ":</b> " + tr ( "The list of available directories." ) + "<br/>" +
+                             tr ( "Select a directory to list the servers registered with it." ) + "<br/>" +
+                             tr ( "You can add custom directories in Advanced Settings." );
```
mcfnord, line 47: I don't see **Advanced Settings**. I see **Advanced Setup**, a tab in the **Settings** dialog.

softins:
I've just added #3446 to allow the user to turn tooltips on or off. With that addition, I would be happy to approve this PR.

ann0see:
I'm inclined to reject this as it seems annoying and solvable with good website documentation.

mcfnord:
Update: the tooltip on/off toggle softins conditioned his approval on (#3446) is now rebased onto current `main` as #3807.

One gap I noticed while testing the two together: the toggle works via per-widget event filters, which #3446 installs only in `CClientDlg` and `CClientSettingsDlg` — so the connect-dialog tooltips this PR adds (the ones originally found obtrusive, especially on the server list) would *not* be governed by the checkbox. The fix is small: install the same filter in `CConnectDlg` and add its `eventFilter`. I've built and smoke-tested this PR merged with #3807 plus that ~25-line addition here: https://github.com/mcfnord/jamulus/commit/6a4617e1 — @pljones feel free to cherry-pick it onto this branch (or I can open it as a follow-up once both PRs land).

With the toggle in place, @ann0see's annoyance concern has a user-controlled answer: anyone who finds the tooltips obtrusive can turn them off.

pljones:
At some point - and it's low priority as we've discovered the Connect Dialog UI accessibility problems aren't easily fixed in it's current state - all of the text needs reworking.  I should really close this - it's just a placeholder to remind me "something must be done" currently.

ann0see:
Then please open an issue/add a todo in your repo.

PR #3259, opened 2024-04-02
Add some missing const refs to CHostAddress params

softins:

Refactoring: pass `CHostAddress` read-only parameters as const references instead of by value, except for signals and slots.

CHANGELOG: Internal: changed passing of some `CHostAddress` parameters to be const references.

**Context: Fixes an issue?**

In the current code base, there are many instances where a `CHostAddress` is already passed as `const CHostAddress&`, but some instances where a `CHostAddress` object is passed by value instead. Depending on the platform, passing a reference may be more efficient than passing a whole object by value. This PR catches a few pass-by-value instances that can better be passed as const refs. However, when passing to signal or slot functions, pass-by-value is needed for thread safety, so those occurrences have not been changed.

The only functions that need to receive a non-const `CHostAddress&` are the ones for parsing addresses, as they pass the results of the parsing back to the caller via the reference.

Probably not.

Compiled fine and under test.

Review, and running for a while on a server.

## Checklist

### DIFF
--- a/src/channel.h
+++ b/src/channel.h
@@ -70,7 +70,7 @@ public:
 
     void PutProtocolData ( const int iRecCounter, const int iRecID, const CVector<uint8_t>& vecbyMesBodyData, const CHostAddress& RecHostAddr );
 
-    EPutDataStat PutAudioData ( const CVector<uint8_t>& vecbyData, const int iNumBytes, CHostAddress RecHostAddr );
+    EPutDataStat PutAudioData ( const CVector<uint8_t>& vecbyData, const int iNumBytes, const CHostAddress& RecHostAddr );
 
     EGetDataStat GetData ( CVector<uint8_t>& vecbyData, const int iNumBytes );
 
@@ -83,7 +83,7 @@ public:
     void SetEnable ( const bool bNEnStat );
     bool IsEnabled() { return bIsEnabled; }
 
-    void                SetAddress ( const CHostAddress NAddr ) { InetAddr = NAddr; }
+    void                SetAddress ( const CHostAddress& NAddr ) { InetAddr = NAddr; }
     const CHostAddress& GetAddress() const { return InetAddr; }
 
     void ResetInfo()
@@ -250,12 +250,12 @@ public slots:
         Protocol.ParseMessageBody ( vecbyMesBodyData, iRecCounter, iRecID );
     }
 
-    void OnProtocolMessageReceived ( int iRecCounter, int iRecID, CVector<uint8_t> vecbyMesBodyData, CHostAddress RecHostAddr )
+    void OnProtocolMessageReceived ( int iRecCounter, int iRecID, CVector<uint8_t> vecbyMesBodyData, const CHostAddress& RecHostAddr )
     {
         PutProtocolData ( iRecCounter, iRecID, vecbyMesBodyData, RecHostAddr );
     }
 
-    void OnProtocolCLMessageReceived ( int iRecID, CVector<uint8_t> vecbyMesBodyData, CHostAddress RecHostAddr )
+    void OnProtocolCLMessageReceived ( int iRecID, CVector<uint8_t> vecbyMesBodyData, const CHostAddress& RecHostAddr )
     {
         emit DetectedCLMessage ( vecbyMesBodyData, iRecID, RecHostAddr );
     }
@@ -282,7 +282,7 @@ signals:
     void RecorderStateReceived ( ERecorderState eRecorderState );
     void Disconnected();
 
-    void DetectedCLMessage ( CVector<uint8_t> vecbyMesBodyData, int iRecID, CHostAddress RecHostAddr );
+    void DetectedCLMessage ( CVector<uint8_t> vecbyMesBodyData, int iRecID, const CHostAddress& RecHostAddr );
 
     void ParseMessageBody ( CVector<uint8_t> vecbyMesBodyData, int iRecCounter, int iRecID );
 };
--- a/src/channel.h
+++ b/src/channel.h
@@ -250,12 +250,12 @@ public slots:
         Protocol.ParseMessageBody ( vecbyMesBodyData, iRecCounter, iRecID );
     }
 
-    void OnProtocolMessageReceived ( int iRecCounter, int iRecID, CVector<uint8_t> vecbyMesBodyData, const CHostAddress& RecHostAddr )
+    void OnProtocolMessageReceived ( int iRecCounter, int iRecID, CVector<uint8_t> vecbyMesBodyData, CHostAddress RecHostAddr )
     {
         PutProtocolData ( iRecCounter, iRecID, vecbyMesBodyData, RecHostAddr );
     }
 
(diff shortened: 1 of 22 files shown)


src/channel.h
```diff
@@ -83,7 +83,7 @@ class CChannel : public QObject
     void SetEnable ( const bool bNEnStat );
     bool IsEnabled() { return bIsEnabled; }
 
-    void                SetAddress ( const CHostAddress NAddr ) { InetAddr = NAddr; }
+    void                SetAddress ( const CHostAddress& NAddr ) { InetAddr = NAddr; }
```
pljones, line 86: Is this one OK?  If the caller changes the content of the reference, the value on the channel would change.  Is that safe or should this actually be a "saved copy"?

src/socket.cpp
```diff
@@ -70,7 +70,7 @@ CSocket::CSocket ( CServer* pNServP, const quint16 iPortNumber, const quint16 iQ
     QObject::connect ( this, &CSocket::ProtocolCLMessageReceived, pServer, &CServer::OnProtocolCLMessageReceived );
 
     QObject::connect ( this,
-                       static_cast<void ( CSocket::* ) ( int, int, CHostAddress )> ( &CSocket::NewConnection ),
+                       static_cast<void ( CSocket::* ) ( int, int, const CHostAddress& )> ( &CSocket::NewConnection ),
```
pljones, line 73: Mmm, not sure on this one, either.  Aren't we creating a new entity here and throwing away the source?  Is it okay to have a dangling reference?

pljones:
Yep, I guess my only concern is if _everything_ is a reference, then _somewhere_ (untracked?) there's the original data... which could get freed and thus the references are no longer valid.  There must be _somewhere_ the actual value gets stored...

src/channel.h
softins: This is fine, it's not like a pointer. The reference is dereferenced as part of fetching `NAddr`, and it is the dereferenced value that is stored in `InetAddr`.

src/socket.cpp
softins: I will check, but I believe it's ok.

src/channel.h
pljones: Ah right.

ann0see (commented):
Not sure how much I can say here, but if it runs correctly, it should be fine?

softins:
I'm off on holiday for a week. I'll pick it up again when I get back.

softins:
I've reverted most of the changes, specifically the ones relating to signals and slots. Connections from signals to slots can cross thread boundaries, so the receiver doesn't want to be referring to data belonging to the sender. So these should remain as call-by-value, ensuring the receiver has its own copy of the parameter. My original changes probably worked ok for a client, but would be liable to breakage in a server with multiple connections, particularly with multi-threading.

That leaves only a few changes remaining. I still need to double-check `CChannel::PutAudioData()`, but I think its ok. The other changes that remain are just immediate uses for initialisation, which is fine.

softins:
Also removed a couple of unused members from `CSocket` and moved `RecHostAddr` from the class to `OnDataReceived()`, as it is only used in that function and doesn't need to persist.

softins:
I need to update the original PR description above, as it is now rather out of date.

src/socket.cpp
```diff
@@ -365,6 +365,8 @@ void CSocket::OnDataReceived()
         use the signal/slot mechanism (i.e. we use messages for that).
     */
 
+    CHostAddress RecHostAddr;
```
pljones, line 368: If we're moving this to a local variable, can we rename it to `haReceivedFrom`?  (Local variables shouldn't have leading capitals...  I know this is just to cut down on the changes, so it's non-blocking.)

src/socket.cpp
```diff
@@ -365,6 +365,8 @@ void CSocket::OnDataReceived()
...
+
```
pljones, line 369: There's a comment down on line 401 I think should be against the whole block starting at 386, if I'm not mistaken.

src/socket.cpp
```diff
@@ -365,6 +365,8 @@ void CSocket::OnDataReceived()
...
     // read block from network interface and query address of sender
     uSockAddr UdpSocketAddr;
```
pljones, line 371: Should really be named `saReceivedFrom`?  Maybe better if both the `CHostAddr` and `uSockAddr` variables had a same-ish name with Jamulus-ish type specifier as prefix.  (I prefer the concept of "sender address" over "received from".)

softins:
>  I still need to double-check `CChannel::PutAudioData()`, but I think that one is ok.

Yes, `CChannel::PutAudioData()` is fine being changed to take a const ref. `CServer::PutAudioData()` already does so.

pljones:
Overall, there's no problems now, I think.

src/socket.cpp
softins: I'll have a think about that before unmarking as draft.

softins: Only thing is as you go down the calls, the receiving parameters are still often called `RecHostAddr` or even just `HostAddr`. Improving the consistency of variable naming conventions is a bigger project!

pljones: Yeah - I started thinking that too.  We say "HostAddr" without the variable saying _what it is_ a lot...  I guess the same for other data types but that seems to stand out.  In some places, where the method name makes it clear why it's consuming a host address, it might just about be acceptable.

src/socket.cpp
softins: I changed that one back to call by value already, as it is an implicit signal-slot connection.

src/socket.cpp
softins: Yes, I think you're right; I'll move it.

pljones (approved):

pljones:
Could you squash before the merge, please.

ann0see:
I may run it quickly ASAP to check for something obvious too but otherwise it's probably ok then.

src/recorder/jamrecorder.cpp
```diff
@@ -41,7 +41,7 @@ using namespace recorder;
  * Creates a file for the raw PCM data and sets up a QDataStream to which to write received frames.
  * The data is stored Little Endian.
  */
-CJamClient::CJamClient ( const qint64 frame, const int _numChannels, const QString name, const CHostAddress address, const QDir recordBaseDir ) :
```
ann0see, line 44: Please check for bugs here too. 

I tested running the server on macOS and it needed multiple tries to show up as directory. The directory did not show up for a local client if a path for recordings is set. It may not be related and needs more investigation (test with a path that is not writable)

ann0see: Hmm. Ok. Current main also shows this issue. Seems to be unrelated.

ann0see (approved):
Otherwise everything seems ok (testing to connect to server, listening, audio goes both ways)

src/recorder/jamrecorder.cpp
softins: OK, if this problem behaviour is repeatable post-3.10.0 but not in 3.10.0, it needs to be raised as a new issue. Maybe need to bisect to find where it was introduced?

ann0see: Issue opened

Issue #3241, opened 2024-02-27
"Delete server" button doesn't show properly on macOS


softins:
**Describe the bug**

The "Delete server" button that was added in #3159 doesn't show properly on macOS. At least on my Mac running Catalina 10.15. The button is there, but it is too narrow to show the whole icon.

The button acts correctly when clicked on.

**To Reproduce**

Download and install a recent build artefact, e.g. from https://github.com/jamulussoftware/jamulus/actions/runs/8031934048. In this case, I installed the Mac Legacy, which had been built with Qt 5.15.2

**Expected behavior**

The button should be wide enough to show the backspace icon.

**Screenshots**

![image](https://github.com/jamulussoftware/jamulus/assets/3224952/0edcd0ef-3a48-4987-880f-8771f6811572)

**Operating system**

MacOS 10.15 Catalina

**Version of Jamulus**

3.10.0dev-dbd6e61

**Additional context**

The button shows correctly on Windows, and when compiled on my Raspberry Pi. I haven't been able to try a more recent MacOS or the non-legacy build.

ann0see:
That's bad. Does the same happen in the server where a comparable button is?

softins:
> That's bad. Does the same happen in the server where a comparable button is?

I will look more closely. Checking the various dev versions I have installed, it seems the delete button in the connect dialog shows up ok with Qt 5.9.9, but is truncated as shown above with Qt 5.12 or Qt 5.15. However, in all three cases the delete buttons in the Server Options page show up fine. I'll have to see what the subtle differences are.

softins:
It appears ok in the artifact I downloaded from #3260. I'll check again once that has been merged and built with Qt 5.15.2.

pljones:
![image](https://github.com/jamulussoftware/jamulus/assets/1549463/38b30788-c2de-491a-808d-2617b3e04ca9)
![image](https://github.com/jamulussoftware/jamulus/assets/1549463/14f210f6-005b-4f57-9192-3a113f63af2d)
Hmm... they're not identical.  The button in the server is larger and the glyph in the client is bold...

softins:
> Hmm... they're not identical. The button in the server is larger and the glyph in the client is bold... (And, of course, different fonts between Linux and Windows. Probably some of the styling is set by the OS theming...)

Yes, the button in `connectdlgbase.ui` has a `<bold>true</bold>` in the `font` setting. We can make sure all is consistent in the PR to fix this issue.

[...]

Issue #3233, opened 2024-02-12
Ability to delete Server and Directory from a list should be the same


pljones:
**What is the current behaviour and why should it be changed?**

In the Client:
* In the Settings Dialog, Advanced tab, the Custom Directories list allows adding and removing of Directories.  To remove a Directory from the list, the edit box must be cleared and enter pressed or focus moved to another field with the box empty.
* In the Connect Dialog, the Server list (near the "Connect" button) allows adding and, recently, removing of Servers.  To remove a Server, from the list, the `<[X]` button is pressed.

In the Server, on the Options tab:
* The recording directory can be set or cleared.
* The server list filename (when running as a Directory) can be set or cleared.

In both cases, to clear the value, the `<[X]` button is pressed.

The Settings Dialog, Advanced tab, Custom Directories list isn't as easy to use or obvious in how to remove an entry as the others.

**Describe possible approaches**

The fields in the Server are read-only, with a dialog to select a new value for the field.  The dialog itself cannot be used to clear the value.  So having the extra "clear" button is necessary.

The Client Connect dialog has used the same approach for the Server list.

The options are:
* Add a `<[X]` button to the Advanced, Custom Directories list to allow deleting the selected value the same was as the Connect dialog and remove the existing method.
* Add a `<[X]` button to the Advanced, Custom Directories list to allow deleting the selected value the same was as the Connect dialog, whilst retaining the existing method.
* Add the same "clear the field and move focus" method to the Connect dialog (hitting enter won't work here as it would attempt to connect, thus connecting to the current Genre Directory as no server was selected)

I don't think the Server buttons should be changed.

I think the best method is to switch the Advanced, Custom Directories list to use the `<[X]` button.

**Has this feature been discussed and generally agreed?**

See https://github.com/jamulussoftware/jamulus/pull/3159#issuecomment-1939406156 for earlier discussion.

ann0see:
@gilgongo @mcfnord what do you think? (Please stay on topic)

AdamGLIN:
Hello !

I planned on implementing the first option given by @pljones. Is there anybody assigned yet ?

Thanks in advance

softins:
Hi @AdamGLIN, thanks for your comment and offer! 

> I planned on implementing the first option given by @pljones. Is there anybody assigned yet ?

I'm not aware of anyone having been assigned, and I can't see any branch in @pljones' fork that relates to this.

So I'd say go for it, submit the PR, and we'll review it. Thanks again!

[...]

PR #3260, opened 2024-04-03
#3233 UI: Common method to delete server and custom directory entries

AdamGLIN:
I added a horizontal box layout which include the box and the delete button. I wrote a fonction to implement the delete button.

It works well when we change the focus with the mouse to add a new directory but I can't make the enter key work : it delete the input and nothing happend. I tried to add the QLineEdit::returnPressed signal but it changed nothing. I struggle to understand the former use of the QComboBox::activated signal since I commented it and nothing seems to have changed.

TODO :
Allowing to add directory via enter key.
Improve spacers in the ui.

CHANGELOG: UI: Add "Delete Entry" button to Advanced Settings, Custom Directories

**Context: Fixes an issue?**

#3233 

## Checklist

### DIFF
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -360,13 +360,15 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     // custom directories
     QString strCustomDirectories = "<b>" + tr ( "Custom Directories" ) + ":</b> " +
                                    tr ( "If you need to add additional directories to the Connect dialog Directory drop down, "
-                                        "you can enter the addresses here.<br>"
-                                        "To remove a value, select it, delete the text in the input box, "
-                                        "then move focus out of the control." );
+                                        "you can enter the addresses here.<br>" );
 
     lblCustomDirectories->setWhatsThis ( strCustomDirectories );
     cbxCustomDirectories->setWhatsThis ( strCustomDirectories );
     cbxCustomDirectories->setAccessibleName ( tr ( "Custom Directories combo box" ) );
+
+    butDeleteCustomDirectory->setAccessibleName ( tr ( "Delete custom directory button" ) );
+    butDeleteCustomDirectory->setWhatsThis ( "<b>" + tr ( "Delete Custom Directory" ) + ":</b> " +
+                                             tr ( "Click the button to delete the currently selected custom directory." ) );
 
     // current connection status parameter
     QString strConnStats = "<b>" + tr ( "Audio Upstream Rate" ) + ":</b> " +
@@ -691,12 +693,9 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                        this,
                        &CClientSettingsDlg::OnMeterStyleActivated );
 
-    QObject::connect ( cbxCustomDirectories->lineEdit(), &QLineEdit::editingFinished, this, &CClientSettingsDlg::OnCustomDirectoriesEditingFinished );
-
-    QObject::connect ( cbxCustomDirectories,
-                       static_cast<void ( QComboBox::* ) ( int )> ( &QComboBox::activated ),
-                       this,
-                       &CClientSettingsDlg::OnCustomDirectoriesEditingFinished );
+    QObject::connect ( cbxCustomDirectories->lineEdit(), &QLineEdit::editingFinished, this, [this] {
+        CClientSettingsDlg::OnCustomDirectoriesChanged ( false );
+    } );
 
     QObject::connect ( cbxLanguage, &CLanguageComboBox::LanguageChanged, this, &CClientSettingsDlg::OnLanguageChanged );
 
@@ -710,6 +709,8 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
     // Driver Setup button is only available for Windows when JACK is not used
     QObject::connect ( butDriverSetup, &QPushButton::clicked, this, &CClientSettingsDlg::OnDriverSetupClicked );
 #endif
+
+    QObject::connect ( butDeleteCustomDirectory, &QPushButton::clicked, this, [this] { CClientSettingsDlg::OnCustomDirectoriesChanged ( true ); } );
 
     // misc
     // sliders
@@ -1009,29 +1010,32 @@ void CClientSettingsDlg::OnEnableOPUS64StateChanged ( int value )
 
 void CClientSettingsDlg::OnFeedbackDetectionChanged ( int value ) { pSettings->bEnableFeedbackDetection = value == Qt::Checked; }
 
-void CClientSettingsDlg::OnCustomDirectoriesEditingFinished()
+void CClientSettingsDlg::OnCustomDirectoriesChanged ( bool bDelete )
 {
-    if ( cbxCustomDirectories->currentText().isEmpty() && cbxCustomDirectories->currentData().isValid() )
+    if ( bDelete )
     {
-        // if the user has selected an entry in the combo box list and deleted the text in the input field,
-        // and then focus moves off the control without selecting a new entry,
-        // we delete the corresponding entry in the vector
-        pSettings->vstrDirectoryAddress[cbxCustomDirectories->currentData().toInt()] = "";
-    }
-    else if ( cbxCustomDirectories->currentData().isValid() && pSettings->vstrDirectoryAddress[cbxCustomDirectories->currentData().toInt()].compare (
-                                                                   NetworkUtil::FixAddress ( cbxCustomDirectories->currentText() ) ) == 0 )
-    {
-        // if the user has selected another entry in the combo box list without changing anything,
-        // there is no need to update any list
-        return;
+        if ( !cbxCustomDirectories->currentData().isValid() )
+        {
+            // no selected directory to delete
+            return;
+        }
+        // delete the selected directory
+        pSettings->vstrDirectoryAddress[cbxCustomDirectories->currentData().toInt()] = QString();
     }
     else
     {
+
+        if ( cbxCustomDirectories->currentText().isEmpty() ||
+             ( cbxCustomDirectories->currentData().isValid() && pSettings->vstrDirectoryAddress[cbxCustomDirectories->currentData().toInt()].compare (
+                                                                    NetworkUtil::FixAddress ( cbxCustomDirectories->currentText() ) ) == 0 ) )
+        {
+            // no need to add a already added directory
+            return;
+        }
         // store new address at the top of the list, if the list was already
         // full, the last element is thrown out
         pSettings->vstrDirectoryAddress.StringFiFoWithCompare ( NetworkUtil::FixAddress ( cbxCustomDirectories->currentText() ) );
     }
-
     // update combo box list and inform connect dialog about the new address
     UpdateDirectoryComboBox();
     emit CustomDirectoriesChanged();
--- a/src/clientsettingsdlgbase.ui
+++ b/src/clientsettingsdlgbase.ui
@@ -1039,11 +1039,40 @@
               </widget>
              </item>
              <item>
-              <widget class="QComboBox" name="cbxCustomDirectories">
-               <property name="editable">
-                <bool>true</bool>
-               </property>
-              </widget>
+              <layout class="QHBoxLayout">
+               <item>
+                <widget class="QComboBox" name="cbxCustomDirectories">
+                 <property name="editable">
+                  <bool>true</bool>
+                 </property>
+                </widget>
+               </item>
+               <item>
+                <widget class="QPushButton" name="butDeleteCustomDirectory">
+                 <property name="sizePolicy">
+                  <sizepolicy hsizetype="Maximum" vsizetype="Ignored">
+                   <horstretch>0</horstretch>
+                   <verstretch>0</verstretch>
+                  </sizepolicy>
+                 </property>
+                 <property name="maximumSize">
+                  <size>
+                   <width>24</width>
+                   <height>16777215</height>
+                  </size>
+                 </property>
+                 <property name="font">
+                  <font>
+                   <weight>75</weight>
+                   <bold>true</bold>
+                  </font>
+                 </property>
+                 <property name="text">
+                  <string notr="true">⌫</string>
+                 </property>
+                </widget>
+               </item>
+              </layout>
              </item>
              <item>
               <spacer name="verticalSpacer_4">
@@ -1053,7 +1082,7 @@
                <property name="sizeHint" stdset="0">
                 <size>
                  <width>20</width>
-                 <height>13</height>
+                 <height>40</height>
                 </size>
                </property>
               </spacer>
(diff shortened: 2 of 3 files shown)


ann0see:
Thanks for opening this PR. It's a draft/WIP so it needs discussion. @pljones Should be most familiar with this.

ann0see:
Could you please also post your questions here? Best right in the code tab close to where the question arises.

softins:
Thanks Adam! I started to look at your commit and questions earlier today, but got diverted onto other things. I will have a look soon, hopefully tomorrow!

src/clientsettingsdlg.cpp
```diff
@@ -360,14 +360,17 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
...
-                                        "then move focus out of the control." );
+                                        "you can enter the addresses here.<br>" );
 
     lblCustomDirectories->setWhatsThis ( strCustomDirectories );
     cbxCustomDirectories->setWhatsThis ( strCustomDirectories );
     cbxCustomDirectories->setAccessibleName ( tr ( "Custom Directories combo box" ) );
 
+    butDeleteCustomDirectories->setAccessibleName ( tr ( "Delete custom directory button" ) );
+    butDeleteCustomDirectories->setWhatsThis ( "<b>" + tr ( "Delete Custom Directory" ) + ":</b> " +
+                                               tr ( "Click the button to clear the currently selected custom directory "
```
pljones, line 371: There's a specific usage difference between the Connect dialog and the Settings dialog.

On the Connect dialog, if no server is selected in the server list and the server combo box field is empty, hitting connect will default to the server hosting the current directory.

As you'll understand, that makes no sense for the Settings screen.  So the phrasing of the "What's This" needs to be a bit simpler: just refer to deleting the entry.

src/clientsettingsdlg.cpp
```diff
@@ -691,12 +694,16 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                        this,
                        &CClientSettingsDlg::OnMeterStyleActivated );
 
+    QObject::connect ( butDeleteCustomDirectories, &QPushButton::clicked, this, &CClientSettingsDlg::OnDeleteCustomDirectoriesClicked );
```
pljones, line 697: This new field surely comes after the combobox, so the code here should, logically, come after the connects fpr the combobox.

src/clientsettingsdlg.cpp
```diff
@@ -1009,17 +1016,37 @@ void CClientSettingsDlg::OnEnableOPUS64StateChanged ( int value )
 
 void CClientSettingsDlg::OnFeedbackDetectionChanged ( int value ) { pSettings->bEnableFeedbackDetection = value == Qt::Checked; }
 
-void CClientSettingsDlg::OnCustomDirectoriesEditingFinished()
+void CClientSettingsDlg::OnDeleteCustomDirectoriesClicked()
 {
-    if ( cbxCustomDirectories->currentText().isEmpty() && cbxCustomDirectories->currentData().isValid() )
+    if ( cbxCustomDirectories->currentText().isEmpty() )
```
pljones, line 1021: The ` && cbxCustomDirectories->currentData().isValid()` check is probably still worth having.

src/clientsettingsdlg.cpp
```diff
@@ -1009,17 +1016,37 @@ void CClientSettingsDlg::OnEnableOPUS64StateChanged ( int value )
 
 void CClientSettingsDlg::OnFeedbackDetectionChanged ( int value ) { pSettings->bEnableFeedbackDetection = value == Qt::Checked; }
 
-void CClientSettingsDlg::OnCustomDirectoriesEditingFinished()
+void CClientSettingsDlg::OnDeleteCustomDirectoriesClicked()
```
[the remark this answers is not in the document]
pljones: (In fact, both signal handlers could call a common method passing an argument to indicate "delete" or "add", so that the only change would be working out whether an entry was being deleted by seeing if the field is empty.  Adding would have to prevent adding empty values, of course.)

src/clientsettingsdlgbase.ui
```diff
@@ -1038,13 +1038,41 @@
...
+             <layout class="QHBoxLayout">
              <item>
               <widget class="QComboBox" name="cbxCustomDirectories">
                <property name="editable">
                 <bool>true</bool>
                </property>
               </widget>
              </item>
+             <item>
+             <widget class="QPushButton" name="butDeleteCustomDirectories">
```
pljones, line 1051: The button only deletes a single entry so the name should be singular.

src/clientsettingsdlg.cpp
pljones, line 1019: This should be a one line wrapper around the common code, passing in a "delete current entry" flag.

src/clientsettingsdlg.cpp
```diff
@@ -1009,17 +1016,37 @@ void CClientSettingsDlg::OnEnableOPUS64StateChanged ( int value )
...
+    // empty last entry
+    pSettings->vstrDirectoryAddress[MAX_NUM_SERVER_ADDR_ITEMS - 1] = QString();
+
+    // redisplay to pick up updated list
+    UpdateDisplay();
+    UpdateDirectoryComboBox();
+    emit CustomDirectoriesChanged();
+}
+
+void CClientSettingsDlg::OnCustomDirectoriesEditingFinished()
```
pljones, line 1046: This should be a one line wrapper around the current code, passing in an "update current entry" flag.

src/clientsettingsdlg.cpp
```diff
@@ -1009,17 +1016,37 @@ void CClientSettingsDlg::OnEnableOPUS64StateChanged ( int value )
 
 void CClientSettingsDlg::OnFeedbackDetectionChanged ( int value ) { pSettings->bEnableFeedbackDetection = value == Qt::Checked; }
 
-void CClientSettingsDlg::OnCustomDirectoriesEditingFinished()
+void CClientSettingsDlg::OnDeleteCustomDirectoriesClicked()
 {
     if ( cbxCustomDirectories->currentText().isEmpty() && cbxCustomDirectories->currentData().isValid() )
```
pljones, line 1021: This needs to change to check whether the call was made for "delete current entry".

If it was, it then needs the existing check that current entry is valid and then it can delete the entry.  The old line 1019 should be all that's needed to do that.  It shouldn't need the `for ... while ...` loops.

It also doesn't need to return.  The `else` (i.e. "not the delete button") case will prevent anything unintended happening.

src/clientsettingsdlg.cpp
```diff
@@ -1009,17 +1016,37 @@ void CClientSettingsDlg::OnEnableOPUS64StateChanged ( int value )
...
+    {
+        while ( pSettings->vstrDirectoryAddress[iLEIdx].compare ( NetworkUtil::FixAddress ( cbxCustomDirectories->currentText() ) ) == 0 )
+        {
+            for ( int jLEIdx = iLEIdx + 1; jLEIdx < MAX_NUM_SERVER_ADDR_ITEMS; jLEIdx++ )
+            {
+                pSettings->vstrDirectoryAddress[jLEIdx - 1] = pSettings->vstrDirectoryAddress[jLEIdx];
+            }
+        }
     }
-    else if ( cbxCustomDirectories->currentData().isValid() && pSettings->vstrDirectoryAddress[cbxCustomDirectories->currentData().toInt()].compare (
```
pljones, line 1021: The `else` here will be for "not the delete button".  The whole of the `if ... else` block that follows should be inside that `else`.

So "if they didn't click the delete button, we got here because editing finished".  Check the data is valid and do nothing if nothing changed.  Else update the entry.

src/clientsettingsdlg.cpp
```diff
@@ -1009,17 +1016,37 @@ void CClientSettingsDlg::OnEnableOPUS64StateChanged ( int value )
...
+    emit CustomDirectoriesChanged();
+}
+
+void CClientSettingsDlg::OnCustomDirectoriesEditingFinished()
+{
+    if ( cbxCustomDirectories->currentData().isValid() && pSettings->vstrDirectoryAddress[cbxCustomDirectories->currentData().toInt()].compare (
+                                                              NetworkUtil::FixAddress ( cbxCustomDirectories->currentText() ) ) == 0 )
     {
         // if the user has selected another entry in the combo box list without changing anything,
         // there is no need to update any list
```
pljones, line 1052: At (old) line 1035, the outer "if delete or not" block should have closed and the method can continue to update the dialog box and emit the signal.

src/clientsettingsdlg.cpp
```diff
@@ -1009,29 +1010,19 @@ void CClientSettingsDlg::OnEnableOPUS64StateChanged ( int value )
 
 void CClientSettingsDlg::OnFeedbackDetectionChanged ( int value ) { pSettings->bEnableFeedbackDetection = value == Qt::Checked; }
 
-void CClientSettingsDlg::OnCustomDirectoriesEditingFinished()
+void CClientSettingsDlg::OnCustomDirectoriesChanged ( bool bDelete )
 {
```
pljones, line 1014: `cbxCustomDirectories->currentData().isValid()` looks common to both legs of the `if` -- so the method could just return (doing nothing) if it's false.

src/clientsettingsdlg.cpp
```diff
@@ -1009,29 +1010,19 @@ void CClientSettingsDlg::OnEnableOPUS64StateChanged ( int value )
...
     }
-    else if ( cbxCustomDirectories->currentData().isValid() && pSettings->vstrDirectoryAddress[cbxCustomDirectories->currentData().toInt()].compare (
-                                                                   NetworkUtil::FixAddress ( cbxCustomDirectories->currentText() ) ) == 0 )
-    {
-        // if the user has selected another entry in the combo box list without changing anything,
-        // there is no need to update any list
-        return;
-    }
-    else
+    else if ( !cbxCustomDirectories->currentText().isEmpty() )
```
pljones, line 1020: As above - still needs to be _valid_ - otherwise the index isn't correct for the delete.

pljones, line 1020: The old test is the right one:
```suggestion
    if ( cbxCustomDirectories->currentData().isValid() && pSettings->vstrDirectoryAddress[cbxCustomDirectories->currentData().toInt()].compare ( cbxCustomDirectories->currentText() ) ) == 0 )
```

src/clientsettingsdlg.cpp
```diff
@@ -1009,29 +1010,19 @@ void CClientSettingsDlg::OnEnableOPUS64StateChanged ( int value )
 
 void CClientSettingsDlg::OnFeedbackDetectionChanged ( int value ) { pSettings->bEnableFeedbackDetection = value == Qt::Checked; }
 
-void CClientSettingsDlg::OnCustomDirectoriesEditingFinished()
+void CClientSettingsDlg::OnCustomDirectoriesChanged ( bool bDelete )
 {
-    if ( cbxCustomDirectories->currentText().isEmpty() && cbxCustomDirectories->currentData().isValid() )
+    if ( bDelete && cbxCustomDirectories->currentData().isValid() )
```
pljones, line 1015: This should be:
```pseudocode
IF the user clicked delete
    IF the entry is not valid THEN return ENDIF
    delete the entry
ELSE ...

File: src/clientsettingsdlgbase.ui [hunk 13]
```diff
@@ -1031,19 +1031,48 @@
              <item>
               <widget class="QLabel" name="lblCustomDirectories">
                <property name="text">
-                <string>Custom Directories:</string>
+                <string>Custom Directories</string>
```
pljones, line 1034: Why this change?  This affects translations.

pljones:
Looks good to me, now.  I've kicked off the build.

AdamGLIN:
Thank you for your patience and your advices :)

pljones:
Just tried the Windows version from the build artifacts and it looks okay to me.

pljones (approved):

ann0see:
This needs some squashing and rebasing, I believe. I'll look into it once I have time.

softins (approved):
I've tried this on my Pi and my Mac. Both ok, so happy to approve.

There don't seem to be any merge conflicts, so a squash merge could be done without needing to rebase. Or else I'd be happy to rebase and squash manually first.

pljones:
I prefer everything neat and tidy, so if you could rebase and squash (happy to do it, too).

softins:
OK, I'll rebase it onto main, which now has the Mac legacy Qt update, let it build, and then check it on my Mac.

softins (changes requested):
I've just loaded the Mac legacy version built with Qt 5.15 and noticed a couple of issues which need to be addressed first. I'm happy to look at these.

1. The button is too narrow for the delete character (it's ok on other platforms and was ok with Qt 5.9). This is the same as I reported in #3241 
image

2. If I type in a custom directory address and finish with "Enter", the field goes blank and the address isn't added. I need to use Tab, or to click outside the field, to make it lose focus. I did see some comments about pressing Enter, but thought they had been resolved. Evidently not, at least on some platforms.

pljones:
Yep Qt on MacOS certainly seems to come with its own UI layout peculiarities that aren't present on other platforms.

The action/focus trigger for updating the list has also been a bit hard to pin down: it works _right_ sometimes on both Windows and Linux for me.  It also works _wrong_ sometimes, too...

softins:
I fixed the issue with Enter. The problem was that pressing Enter within the combi box was causing a button push to delete, because the default setting for `autoDefault` appears to be `true`. Having set `autoDefault` explicitly to `false` for the delete button in `clientsettingsdlgbase.ui`, the Enter key works correctly for just finishing editing.

Will investigate the button appearance on Mac later.

softins:
For the delete buttons, I needed to change the horizontal size policy from "Maximum" to "Fixed", so that the buttons were definitely wide enough, but not allowed to grow. Need to check they are ok on all platforms.

softins:
I've tried the latest build above on Mac Legacy, Windows and Pi Linux. Not current Mac though.

The button width is now fine on Mac Legacy, but it's a bit wider than ideal on Windows and Linux. Must be scaling differences. It may be acceptable, but if not, we would need to add code to adjust the width of those buttons at runtime for Mac only. Thoughts?

softins (approved):

pljones:
Maybe set the size policy to maximum only for Mac if that's where it's needed?

softins:
> Maybe set the size policy to maximum only for Mac if that's where it's needed?

It's the other way around. It was previously maximum only, but that allows the button to be shrunk as much as needed. But on Mac from 5.12 on, the button was shrinking right down. My guess is that it couldn't correctly calculate the size of the contained text when it was just a Unicode character for backspace, so took the text width as 0. I found I needed to specify Fixed so that both maximum and minimum were constrained.

Either way, to fix it we would need to add conditional runtime code for Mac only, as there is no way to put architecture conditionals in a `.ui` file. The question is, is it worth doing?

I have wondered whether it might be worth putting an image in the button instead of a Unicode character? e.g. a wastebasket or trashcan?

softins:
A further thought: maybe merge this PR without the button width fix, since that applies to both the settings and connect dialogs? The connect dialog is arguably outside the scope of this PR. Then raise a separate PR to fix #3241 in both places for Mac?

pljones:
OK by me.

pljones (approved):

pljones:
image
image
Hmm... they're not identical.  The button in the server is larger and the glyph in the client is bold...

ann0see:
Ok. So this also needs new screenshots for the docs potentially.

pljones:
Linux:
image

Windows:
image

ann0see:
Yes. But that should show up

pljones:
https://github.com/jamulussoftware/jamuluswebsite/issues/1002 raised for documentation.

PR #3264, opened 2024-04-09
Added translation of errors into Russian

IgneInNocte:
I added part of the translation in Russian, with plans to make a full translation.


src/translation/wininstaller/ru_RU.nsi
jujudusud, line 1: Change "ENGLISH" by "RUSSIAN" in lang_english expression.

ann0see:
Thanks! 
This might need some changes on weblate too (not sure). But you should then also be added as translator.

jujudusud:
> Thanks! This might need some changes on weblate too (not sure). But you should then also be added as translator.

It is in :  src/utils.h

Issue #3092, opened 2023-07-01
Place build generated files in separate folder


ann0see:
**Describe the bug**

Currently, the build generated files e.g. `.moc` are scattered in the whole root directory. This is the case especially on manual builds. 
https://github.com/jamulussoftware/jamulus/pull/2588 tried to clean them up by moving them into a separate folder however it was not successful.

This is annoying as there are a lot of files.

Please note that if the files are moved, the `.gitignore` file needs to be updated too.

**To Reproduce**

Run a build on e.g. Linux and see the files in the root.

**Expected behavior**

Have a clean root directory.

**Questions/Comments**

* Why is this a problem? I think overview is the main issue
* Cross compilation is more difficult
* Make clean removes them

pljones:
QtCreator does out-of-tree builds by default.

`make distclean; qmake && make && make clean` keeps the project root clean and leaves the user with the option of _not_ running `make clean`.

`make` needs to remain re-runnable from a point of failure with any proposed solution.

softins:
I have a simple solution for this, which is in my queue for raising a PR soon.

PR #3288, opened 2024-06-10
Build: Generate qm files and embedded resource during build

softins:

Generate the translation .qm files during build and remove them from Github.

CHANGELOG: Build: Generate qm files and embedded resource during build

**Context: Fixes an issue?**

This replaces previous closed PRs #2393 and #1303

The qm files are binary dependants of the ts text files. Historically they have been generated manually and stored in github. Qt versions from 5.12 include qmake rules for generating and embedding them. After a couple of years of intermittent experimentation, this PR is completely working.

The breakthrough was recently discovered in a [Qt bug report](https://bugreports.qt.io/browse/QTBUG-102070). Builds for multiple architectures (e.g. on Android) or targets (Windows defaults to `debug_and_release`) only work if `RCC_DIR` and `LRELEASE_DIR` are allowed to default and not specified in `Jamulus.pro`.

It needs to be made conspicuous that Jamulus will now only build on Qt 5.12 or later. `Jamulus.pro` will cause `qmake` to exit with an error on earlier (obsolete) versions of Qt.

Working and ready for review.

Just review and general testing.

## Checklist


.github/workflows/autobuild.yml
```diff
@@ -314,7 +314,8 @@ jobs:
       - name:                       Checkout code
         uses:                       actions/checkout@v4
         with:
-          submodules:               true
+          # only Android needs the oboe submodule, so don't fetch it for other builds
```
pljones, line 317: Being picky, this should be part of a separate change, really.

src/util.cpp
```diff
@@ -1612,8 +1612,9 @@ QMap<QString, QString> CLocale::GetAvailableTranslations()
         // get alias of translation file
         const QString strCurFileName = DirIter.next();
 
-        // extract only language code (must be at the end, separated with a "_")
-        const QString strLoc = strCurFileName.right ( strCurFileName.length() - strCurFileName.indexOf ( "_" ) - 1 );
+        // extract only language code "xx_XX" from "translation_xx_XX.qm"
```
pljones, line 1615: Somehow I feel this is an ideal candidate for using a regexp to extract the value.  It should be code that's only run through once on execution, so any overhead can be ignored.  It would be more readable and verifiable that way, I think, than trying to work out what's going on here.

.gitignore
pljones, line 1: Should `*.qm` go into the list?

.github/workflows/autobuild.yml
softins: Yes, I did wonder whether to bite the bullet and raise a whole new PR for the one line. Can do if you prefer.

src/util.cpp
softins: I think a regex is just slightly overkill for this. All we are doing is extracting the part between the first `'_'` and the first `'.'`, so two simple `indexOf()` calls and a `mid()` are fine. The change was needed because the resource file created on the fly by `qmake` includes the `.qm` string in the generated `.qrc` file, when we didn't include it manually in `resources.qrc`.

.gitignore
softins: The `.qm` files go into either `.qm/`, `debug/` or `release/`, which all have entries already. Adding `*.qm` isn't necessary, but wouldn't hurt either. Can do if you prefer.

.github/workflows/autobuild.yml
softins: OK, I've moved it to #3289

.gitignore
softins: Just added and pushed

pljones (approved):

Jamulus.pro
```diff
@@ -1,5 +1,10 @@
 VERSION = 3.10.0dev
 
+# Using lrelease and embed_translations only works for Qt 5.12 or later
```
ann0see, line 3: Please link this PR here. This would users allow to see that they could in theory build using an older Qt version if they revert this.

Jamulus.pro
```diff
@@ -59,6 +64,10 @@ TRANSLATIONS = src/translation/translation_de_DE.ts \
     src/translation/translation_sv_SE.ts \
     src/translation/translation_zh_CN.ts
 
+# This value must match the value for DirIter in the function
```
ann0see, line 67: Please write briefly what the value means and not just say "this must equal that"

pljones: Also, an explanation of why that function _cannot_ use the variable here would be good.

softins: Not sure how much detailed explanation belongs in the `Jamulus.pro` file, but I'll explain the issue here and you guys can advise.

In `src/util.cpp` is the following function excerpt:

```
QMap<QString, QString> CLocale::GetAvailableTranslations()
{
    QMap<QString, QString> TranslMap;
    QDirIterator           DirIter ( ":/translations" );
    ......
```

The string `:/translations` is a reference to internal resources, currently defined in `src/resources.qrc`:

```
<RCC>
    <qresource prefix="/translations">
        <file alias="translation_de">translation/translation_de_DE.qm</file>
    </qresource>
    <qresource prefix="/translations">
        <file alias="translation_fr">translation/translation_fr_FR.qm</file>
    </qresource>
    ......
```

It is the value of `prefix` in this file that is referred to by `DirIter` and must therefore match.

Now, when we move to `embed_translations`, the translations are put in a new resource file generated by `qmake`, viz. `qmake_qmake_qm_files.qrc`. By default, `qmake` gives them the prefix `i18n`:

```
<!DOCTYPE RCC><RCC version="1.0">
<qresource prefix="i18n">
<file alias="translation_de_DE.qm">/home/tony/jamulus/.qm/translation_de_DE.qm</file>
<file alias="translation_fr_FR.qm">/home/tony/jamulus/.qm/translation_fr_FR.qm</file>
......
```

So, to be able to find them, exactly one of two things needs to be done:
1. Modify the value of `DirIter` in `CLocale::GetAvailableTranslations()` to contain `:/i18n` instead of `:/translations`.
2. Tell `qmake` to use a prefix of `translations` instead of the default, by setting `QM_FILES_RESOURCE_PREFIX` in `Jamulus.pro`, to match the existing value for `DirIter`.

I did the latter, and added a comment to alert the user of the consistency needed between the two places. But I didn't feel it needed an essay at that point.

Maybe instead we revert to option 1 above instead? Maybe with a short comment about the default value used by `qmake` for the resource. Or else stick with option 2 and in `Jamulus.pro` add a link to this comment for further details? Opinions? After typing this, I'm now leaning towards option 1. Especially since that function already has another change within it.

Jamulus.pro
softins: OK, done. Will be in my next push.

Jamulus.pro
pljones: If there's a variable that can affect what qmake does that affects what the code needs to do, then the _code_ should be sensitive to that value, too.  So I'd still say we should use `QM_FILES_RESOURCE_PREFIX` in the code, defaulting to the qmake default if it's not set.

softins: This would require passing in the value of `QM_FILES_RESOURCE_PREFIX` to the compiler as a define. I disagree with the need in this case. For something that just needs to be got right once and there is never going to be a need to change, I think that is overkill. I would agree if it were something the user needs the ability to choose or change, but that's not the case here.

So my preferred option is to take out the `QM_FILES_RESOURCE_PREFIX` change from `Jamulus.pro` and just accept the default used by `qmake`, and to update the function in `util.cpp` to use the default prefix value, with an explanatory comment. I'll add a commit to that effect for review, which I can squash into the original commit before merging.

softins:
Once this update is merged, I can follow it up with a PR to easily solve #3092

pljones (approved):

Jamulus.pro
```diff
@@ -1,5 +1,11 @@
 VERSION = 3.10.0dev
 
+# Using lrelease and embed_translations only works for Qt 5.12 or later.
+# See https://github.com/jamulussoftware/jamulus/pull/3288 for these changes.
+lessThan(QT_MAJOR_VERSION, 5) | equals(QT_MAJOR_VERSION, 5) : lessThan(QT_MINOR_VERSION, 12) {
+    error(Jamulus requires at least Qt5.12)
```
ann0see, line 6:
```suggestion
    error(Jamulus requires at least Qt5.12. Check PR #3288)
```

ann0see (commented):
Seems ok. I will test on windows

Jamulus.pro
pljones: I'd suggest the full URL in case the source ever moves off GitHub (the PR reference alone could become meaningless).

softins: Done, see 97306e7

ann0see (approved):
Thanks

PR #3348, opened 2024-08-19
Add changable bundle id and docs for macOS signing

ann0see:
**Short description of changes**

Adds variable for a changable Bundle ID for notarization for macOS. One can add the bundle id here: https://github.com/jamulussoftware/jamulus/settings/variables/actions/new

Also adds some more (basic) documentation which needs improvement as soon as signing works, and some clean up.

CHANGELOG: Autobuild: Use variable for Bundle ID on macOS. This allows to change the bundle ID for macOS notarization on GitHub

**Context: Fixes an issue?**

No. Related to: https://github.com/jamulussoftware/jamulus/issues/3255#issuecomment-2295387824

**Does this change need documentation? What needs to be documented and how?**

Maybe the signing - once it works.

**Status of this Pull Request**

Ready for review

**What is missing until this pull request can be merged?**
Review (language wise)

## Checklist

-  [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
-  [ ] I tested my code and it does what I want (should be merged if it doesn't break. I will try signing on my own repo)
-  [ ] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
-  [ ] I've filled all the content above

PR #3350, opened 2024-08-22
Refactor macOS App Store upload

ann0see:

Part of a series of refactoring the macOS AppStore upload procedure. This PR includes mostly moving code around and decouples the app store upload from the normal signing process.
Not tested without the other changes, but it should be ok.

CHANGELOG: Refactoring: Decouple App Store upload from dmg build
**Context: Fixes an issue?**

Related to: #3348 - at least somewhat.

No

Ready for review. Especially the moving around between multiple folders is somewhat not so nice.

Review

## Checklist

-  [x] I tested my code and it does what I want (together with other PRs)

### DIFF
--- a/mac/deploy_mac.sh
+++ b/mac/deploy_mac.sh
@@ -4,8 +4,11 @@ set -eu -o pipefail
 root_path=$(pwd)
 project_path="${root_path}/Jamulus.pro"
 resources_path="${root_path}/src/res"
+raw_path="${root_path}/rawbuild" # Path for raw, not yet runnable or signed binaries.
 build_path="${root_path}/build"
 deploy_path="${root_path}/deploy"
+macapp_deploy_path="${root_path}/storedeploy"
+
 cert_name=""
 macapp_cert_name=""
 macinst_cert_name=""
@@ -51,8 +54,12 @@ cleanup() {
     # Clean up previous deployments
     rm -rf "${build_path}"
     rm -rf "${deploy_path}"
+    rm -rf "${macapp_deploy_path}"
+    rm -rf "${raw_path}"
     mkdir -p "${build_path}"
     mkdir -p "${deploy_path}"
+    mkdir -p "${macapp_deploy_path}"
+    mkdir -p "${raw_path}"
 }
 
 build_app() {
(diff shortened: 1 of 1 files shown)


mac/deploy_mac.sh
```diff
@@ -181,6 +172,33 @@ build_installer_image() {
         "${deploy_path}/"
 }
 
+build_storesign_pkg() {
+    # Build installer pkg file - for submission to App Store
+    if [[ -z "$macapp_cert_name" ]]; then
+        echo "No macapp certificate specified. Thus skipping pkg creation for App Store"
+        return
+    else
```
pljones, line 180: This can end the `if` as there's been a return - no need for an intented `else` block.
```suggestion
    fi
```

mac/deploy_mac.sh
```diff
@@ -220,3 +238,5 @@ build_app server_app "CONFIG+=server_bundle"
 
 # Create versioned installer image
 build_installer_image "${CLIENT_TARGET_NAME}" "${SERVER_TARGET_NAME}"
+# Create pkg file for App Store submission if certificate is given
```
pljones, line 241: This is unconditional, so the comment should be clearer.

pljones, line 241: This is unconditional, so the comment should be clearer.
```suggestion
# Create pkg file and, if certificate is given, prepare for App Store submission 
```

mac/deploy_mac.sh
pljones: Actually, move the whole `if` block out and wrap the call -- then the comment on the call will be accurate and the proc will just do its thing unconditionally.

mac/deploy_mac.sh
```diff
@@ -181,6 +172,27 @@ build_installer_image() {
...
+build_storesign_pkg() {
+    # Build installer pkg file - for submission to App Store
+    # Note: We do not upload the server to the app store for now. This could be changed easily by uncommenting the respective lines below
+    local client_target_name="${1}"
+    # local server_target_name="${2}"
+    echo "Cert signing for App Store started..."
+
+    # Copy binaries to separate temporary deploy directory leave the (adhoc) signed app untouched
+    cp -a "${raw_path}/${client_target_name}.app" "${macapp_deploy_path}/${client_target_name}.app"
+    # cp -a "${deploy_path}/${server_target_name}.app" "${macapp_deploy_path}/${server_target_name}.app"
```
softins, line 184: Is it correct that client uses `raw_path` but server uses `deploy_path`? Or should they both be the same?

I can see from line 110 above that one contains a copy of the other, but I'm not clear whether at this point they are still the same.

ann0see: No. That's an oversight from me.

pljones (approved):
Subject to fixing https://github.com/jamulussoftware/jamulus/pull/3350/files#r1727831460 I'm okay with it.

mac/deploy_mac.sh
ann0see: Done.

softins (approved):
LGTM

PR #3364, opened 2024-09-08
Fix connection status issue (#2519)

### LINKED ISSUE #2519: Connection status client and gui can get out of sync.

ann0see:
This is a manual port of https://github.com/jamulussoftware/jamulus/pull/2550 by @pgScorpio

First step to get to know the codebase of #2550 

CHANGELOG: DO NOT MERGE

**Context: Fixes an issue?**

Fixes: #2519 
Supersedes: #2550 

No

Ready for first review and decision on how to move forward.

Review.
It compiles but not every part of the code is fully understood and commented.

## Checklist


ann0see:
Coding style check also fails on main. Seems to be an unrelated issue?

ann0see:
Pasting from previous PR here:

> Moved Connect/Disconnect code from CClientdlg to CClient.
> Now using the proper connected checks in several places.
> Added bDisconnectAndDisable to CChannel. (For a Client now
Channel.Disconnect() will block audio data and auto disable the channel on disconnected)

src/client.cpp
```diff
@@ -474,26 +466,6 @@ void CClient::StartDelayTimer()
     }
 }
 
-bool CClient::SetServerAddr ( QString strNAddr )
```
ann0see, line 477: The SRV part may need to be re implemented elsewhere.

src/client.cpp
```diff
@@ -758,8 +734,18 @@ void CClient::OnSndCrdReinitRequest ( int iSndCrdResetType )
     }
     MutexDriverReinit.unlock();
 
+    if ( !strError.isEmpty() )
+    {
+#ifndef HEADLESS
+        QMessageBox::critical ( 0, APP_NAME, strError, tr ( "Ok" ) );
+#else
+        qCritical() << qUtf8Printable ( strError );
```
ann0see, line 742: Not great. We may want to use the old error message handling for now and than change this in future.

src/main.cpp
```diff
@@ -963,6 +961,12 @@ int main ( int argc, char** argv )
                 // show dialog
                 ClientDlg.show();
 
+                // Connect on startup
+                if ( !strConnOnStartupAddress.isEmpty() )
```
ann0see, line 965: I believe we'd want to have some more validity checking here

src/client.cpp
```diff
@@ -862,50 +845,88 @@ void CClient::OnClientIDReceived ( int iChanID )
     emit ClientIDReceived ( iChanID );
 }
 
-void CClient::Start()
+bool CClient::Connect ( QString strServerAddress, QString strServerName )
```
ann0see, line 848: I believe we may want to still have the split into two functions for modularity, but provide an API for a direct connection.

ann0see:
Ok. 

https://github.com/jamulussoftware/jamulus/pull/3364/commits/a5af8ccd478e405e0ccbc760315c366f12c97367 is for having a smaller diff in this PR. The renaming of methods can happen in another PR.

Client.Start() and Client.Stop() have been re-introduced in https://github.com/jamulussoftware/jamulus/pull/3364/commits/535f781eece9a2a20598dcdf40df6a44d1c1148f, but can be - of course - merged again if wanted.

I think, this PR should now get an in depth review.

src/client.h
```diff
@@ -121,8 +120,11 @@ class CClient : public QObject
 
     void Start();
     void Stop();
+    bool Connect ( QString strServerAddress, QString strServerName );
```
ann0see, line 123: It does make sense to introduce an easy to call Connect() and Disconnect() method IMO, so this is good.

src/client.h
```diff
@@ -121,8 +120,11 @@ class CClient : public QObject
...
+    bool Disconnect();
+
     bool IsRunning() { return Sound.IsRunning(); }
-    bool IsCallbackEntered() const { return Sound.IsCallbackEntered(); }
+    bool IsCallbackEntered() const { return Sound.IsCallbackEntered(); } // For OnTimerCheckAudioDeviceOk only
```
ann0see, line 127: Reverted naming. May need some in depth review and discussion.

src/client.h
```diff
@@ -427,8 +429,9 @@ protected slots:
 
     void CLChannelLevelListReceived ( CHostAddress InetAddr, CVector<uint16_t> vecLevelList );
 
+    void Connecting ( QString strServerName );
```
ann0see, line 432: Probably a missing slot.

src/client.h
```diff
@@ -427,8 +429,9 @@ protected slots:
...
     void Disconnected();
-    void SoundDeviceChanged ( QString strError );
+    void SoundDeviceChanged();
```
ann0see, line 434: I don't feel strongly about removing the error message handling. Could be reverted.

src/clientdlg.cpp
```diff
@@ -730,15 +719,12 @@ void CClientDlg::OnConnectDlgAccepted()
...
+        // initiate connection
+        if ( pClient->Connect ( strSelectedAddress, strMixerBoardLabel ) )
         {
-            Disconnect();
+            OnConnect ( strMixerBoardLabel );
         }
-
-        // initiate connection
-        Connect ( strSelectedAddress, strMixerBoardLabel );
+        // TODO: investigate and add error handling on failed Connect() call.
```
ann0see, line 727: TODO as commented in code.

src/clientdlg.cpp
```diff
@@ -748,10 +734,10 @@ void CClientDlg::OnConnectDlgAccepted()
 void CClientDlg::OnConnectDisconBut()
 {
     // the connect/disconnect button implements a toggle functionality
-    if ( pClient->IsRunning() )
+    if ( pClient->Disconnect() )
```
ann0see, line 737: To be refactored to not have an empty if() -- just style.

src/clientdlg.cpp
```diff
@@ -1193,65 +1167,36 @@ void CClientDlg::OnCLPingTimeWithNumClientsReceived ( CHostAddress InetAddr, int
     ConnectDlg.SetPingTimeAndNumClientsResult ( InetAddr, iPingTime, iNumClients );
 }
 
-void CClientDlg::Connect ( const QString& strSelectedAddress, const QString& strMixerBoardLabel )
+void CClientDlg::OnConnect ( QString strServerName )
```
ann0see, line 1170: Add documentation: precondition: ServerAddr() must have been set.

src/clientdlg.cpp
```diff
@@ -1193,65 +1167,36 @@ void CClientDlg::OnCLPingTimeWithNumClientsReceived ( CHostAddress InetAddr, int
...
 
-        // change connect button text to "disconnect"
-        butConnect->setText ( tr ( "&Disconnect" ) );
+    // change connect button text to "disconnect"
+    butConnect->setText ( tr ( "&Disconnect" ) );
 
-        // set server name in audio mixer group box title
-        MainMixerBoard->SetServerName ( strMixerBoardLabel );
+    // set server name in audio mixer group box title
+    MainMixerBoard->SetServerName ( strServerName );
```
ann0see, line 1182: I believe this needs to be a label, since it could include something about recording?
@pljones ?

src/clientdlg.cpp
```diff
@@ -1275,11 +1220,7 @@ void CClientDlg::Disconnect()
     TimerDetectFeedback.stop();
     bDetectFeedback = false;
 
-    //### TODO: BEGIN ###//
-    // is this still required???
-    // immediately update status bar
-    OnTimerStatus();
-    //### TODO: END ###//
+    UpdateDisplay();
```
ann0see, line 1223: TODO was removed and replaced by OnTimerStatus()
clientdlg.h shows that OnTimerStatus() just calls UpdateDisplay().

src/clientdlg.cpp
```diff
@@ -1293,6 +1234,8 @@ void CClientDlg::Disconnect()
 
     // clear mixer board (remove all faders)
     MainMixerBoard->HideAll();
+
+    SetMixerBoardDeco ( RS_UNDEFINED, pClient->GetGUIDesign() );
```
ann0see, line 1238: Is this needed?

src/connectdlg.cpp
ann0see, line 1: Changed to non executable file type. Probably this shouldn't be part of this PR.

ann0see:
To decide:
* What to keep
* What to split into another PR
* How and what to document.

pljones:
I've no idea what this change is trying to do.  It appears to be making changes to both client and server.  There's no explanation given as to whether one relies on the other or what happens when different versions interact.

ann0see:
I think it **should** only do changes on the client side. 
The PR is still huge and potentially we need to extract the most important points. See the linked PR for old discussions. 
My goal is to use this to lay foundations for the JSON-RPC connect/disconnect calls.

ann0see:
So we need to try understanding what gets changed.

src/channel.cpp
```diff
@@ -622,6 +637,11 @@ EGetDataStat CChannel::GetData ( CVector<uint8_t>& vecbyData, const int iNumByte
                 eGetStatus  = GS_CHAN_NOW_DISCONNECTED;
                 iConTimeOut = 0; // make sure we do not have negative values
 
+                if ( bDisconnectAndDisable )
```
ann0see, line 640:
```suggestion
                if ( bDisconnectAndDisable && !bIsServer )
```

?

src/channel.cpp
```diff
@@ -643,6 +663,13 @@ EGetDataStat CChannel::GetData ( CVector<uint8_t>& vecbyData, const int iNumByte
         {
             // channel is disconnected
             eGetStatus = GS_CHAN_NOT_CONNECTED;
+
+            if ( bDisconnectAndDisable )
```
ann0see, line 667:
```suggestion
            if ( bDisconnectAndDisable && !bIsServer )
```

src/channel.h
```diff
@@ -76,7 +76,7 @@ class CChannel : public QObject
 
     void PrepAndSendPacket ( CHighPrioSocket* pSocket, const CVector<uint8_t>& vecbyNPacket, const int iNPacketLen );
 
-    void ResetTimeOutCounter() { iConTimeOut = iConTimeOutStartVal; }
+    void ResetTimeOutCounter() { iConTimeOut = bDisconnectAndDisable ? 1 : iConTimeOutStartVal; }
```
ann0see, line 79: I don't like this. Should probably be expanded into an assignment and boolean comparison.

ann0see:
Maybe we can reconstruct something from https://github.com/pgScorpio/jamulus/commit/648880cbfec48ab10b3eaac06c846c9d6f365c20

src/clientdlg.cpp
ann0see, line 1170: @pljones Potentially we only need this signal for the JSON-RPC PR.

softins:
I looked quickly through the changes the other day and didn't notice anything of concern, but I haven't yet had time to study in detail.

src/clientdlg.cpp
pljones: If it's a private slot (with an `On` prefix it should be a signal handling slot), the precondition _should_ be guaranteed.  Otherwise there's an error in where the signal is raised.  Of course, as it's potentially passed across threads, thread scope would need considering.

src/clientdlg.cpp
pljones: When you disconnect, you're no longer being recorded.  As far as I know, the status would have been reset and the banner already removed.  But then, the client dialog should already be handling disconnection robustly.  If there some issue with it I'm not aware of, then maybe it's needed.

src/channel.cpp
```diff
@@ -511,6 +513,24 @@ void CChannel::Disconnect()
...
+    {
+        if ( IsConnected() )
+        {
+            // for a Client, block further audio data and disable the channel as soon as Disconnect() is called
+            // TODO: Add reasoning from #2550
+            bDisconnectAndDisable = true;
+        }
+        else
+        {
+            // For disconnected clients set defaults
```
pljones, line 527: Why isn't this done when the disconnect _actually_ happens, rather than here?  Like in line 645?  Is this just duplicated?  Or is there a reason for it?

src/channel.cpp
```diff
@@ -643,6 +668,13 @@ EGetDataStat CChannel::GetData ( CVector<uint8_t>& vecbyData, const int iNumByte
         {
             // channel is disconnected
             eGetStatus = GS_CHAN_NOT_CONNECTED;
+
+            if ( bDisconnectAndDisable && !bIsServer )
+            {
+                bDisconnectAndDisable = false;
```
pljones, line 674: ??! So these get set to the same values whether `eGetStatus` is `GS_CHAN_NOW_DISCONNECTED` or `GS_CHAN_NOT_CONNECTED`?  Now I'm really confused.

src/client.cpp
```diff
@@ -184,13 +183,6 @@ CClient::CClient ( const quint16  iPortNumber,
     // start the socket (it is important to start the socket after all
     // initializations and connections)
     Socket.Start();
-
-    // do an immediate start if a server address is given
-    if ( !strConnOnStartupAddress.isEmpty() )
-    {
-        SetServerAddr ( strConnOnStartupAddress );
-        Start();
-    }
```
pljones, line 193: Why has this code been removed?  It's used for headless clients.  Restore it.

src/client.cpp
```diff
@@ -337,7 +329,7 @@ void CClient::CreateServerJitterBufferMessage()
 void CClient::OnCLPingReceived ( CHostAddress InetAddr, int iMs )
 {
     // make sure we are running and the server address is correct
-    if ( IsRunning() && ( InetAddr == Channel.GetAddress() ) )
+    if ( Channel.IsEnabled() && ( InetAddr == Channel.GetAddress() ) )
```
pljones, line 332: This isn't about the channel being enabled. It's about whether the client is running.  It's a different intent.

src/client.cpp
```diff
@@ -626,12 +618,16 @@ QString CClient::SetSndCrdDev ( const QString strNewDev )
         Sound.Start();
     }
 
-    // in case of an error inform the GUI about it
     if ( !strError.isEmpty() )
     {
-        emit SoundDeviceChanged ( strError );
+        // due to error, disconnect
+        Disconnect();
```
pljones, line 624: Why hide information like this?  Is it any way related to the rest of the PR?

src/client.cpp
```diff
@@ -626,12 +618,16 @@ QString CClient::SetSndCrdDev ( const QString strNewDev )
...
     }
 
+    // in case of an error, this will inform the GUI about it
+
+    emit SoundDeviceChanged();
```
pljones, line 629: Why _not_ pass the error?  What's that got to do with this PR?

src/client.cpp
```diff
@@ -758,8 +754,18 @@ void CClient::OnSndCrdReinitRequest ( int iSndCrdResetType )
...
+        QMessageBox::critical ( 0, APP_NAME, strError, tr ( "Ok" ) );
+#else
+        qCritical() << qUtf8Printable ( strError );
+        exit ( 1 );
+#endif
+    }
+
     // inform GUI about the sound card device change
-    emit SoundDeviceChanged ( strError );
+    emit SoundDeviceChanged();
```
pljones, line 768: Why not leave it in the signal?

Ideally we'd have two slots - one for GUI mode and one for headless mode.

src/client.cpp
pljones: I'd like all this removed from the PR.  I can't see how it relates.

src/client.cpp
```diff
@@ -876,18 +879,13 @@ void CClient::Start()
 
 void CClient::Stop()
 {
-    // stop audio interface
-    Sound.Stop();
+    // start disconnection
+    // Channel.Disconnect() should automatically disable Channel as soon as disconnected.
+    // Note that this only works if sound is active!
```
pljones, line 884: So why make the changes here?

src/client.h
pljones: Having these in `client.cpp` rather than `clientdlg.cpp` is important.  It's important enough to be done in a single PR on its own. I think we should start with that.

src/client.h
pljones: As above:
- split this out to a separate PR
- have two slots, one for GUI, one for headless, and pass the error to each

src/clientdlg.cpp
pljones: This should just call `pClient->Disconnect()`, I think.  When that emits "Disconnected", the client dialog can clean up the mixer board.

src/clientdlg.cpp
pljones: Yeah, we've mangled it not to _just_ be the server name, haven't we.

src/clientdlg.cpp
```diff
@@ -1193,65 +1167,36 @@ void CClientDlg::OnCLPingTimeWithNumClientsReceived ( CHostAddress InetAddr, int
...
+    // audio feedback detection
+    if ( pSettings->bEnableFeedbackDetection )
+    {
+        TimerDetectFeedback.start ( DETECT_FEEDBACK_TIME_MS ); // single shot timer
+        bDetectFeedback = true;
     }
 }
 
-void CClientDlg::Disconnect()
+void CClientDlg::OnDisconnect()
```
pljones, line 1198: Tidying up the mixer board should go in here, I guess.

ann0see:
Thanks for the review! I'll go through what I can answer later.

src/clientdlg.cpp
pljones: Um, depends on what `MainMixerBoard->HideAll();` does, I guess.  I'd expect it not to be, as the deco is inside what `HideAll` hides.  Ideall `HideAll` would know to tidy things up, too...

src/clientsettingsdlg.cpp
```diff
@@ -1066,22 +1066,24 @@ void CClientSettingsDlg::OnSndCrdBufferDelayButtonGroupClicked ( QAbstractButton
...
     UpdateJitterBufferFrame();
     UpdateSoundCardFrame();
-
-    if ( !pClient->IsRunning() )
-    {
-        // clear text labels with client parameters
-        lblUpstreamValue->setText ( "---" );
-        lblUpstreamUnit->setText ( "" );
-    }
+    UpdateUploadRate();
```
pljones, line 1086: Yeah, makes sense.  But completely unrelated the connect/disconnect stuff.

src/client.cpp
ann0see: It's probably just because of a change in error handling with the message box class he wanted to implement. I don't think it's needed here -> revert it

src/client.cpp
ann0see: Nothing, I think.

src/main.cpp
```diff
@@ -963,6 +961,12 @@ int main ( int argc, char** argv )
                 // show dialog
                 ClientDlg.show();
 
+                // Connect on startup
```
pljones, line 964: These shouldn't be in `main.cpp`.

src/channel.h
pljones: Yeah, I not enthused by the whole `bDisconnectAndDisable` -- I can't see the problem it's trying to fix.

src/client.cpp
pljones: Oh, it got (wrongly) moved to `main.cpp`.

src/client.h
ann0see: Agree. That's issue https://github.com/jamulussoftware/jamulus/issues/3367

src/channel.h
ann0see: I think they wanted to introduce an intermediate state somehow to fix the button being able to get out of sync.
From https://github.com/jamulussoftware/jamulus/pull/2550#issue-1181299521

> Added bDisconnectAndDisable to CChannel. (For a Client now
Channel.Disconnect() will block audio data and auto disable the channel on disconnected)

src/client.cpp
ann0see: I think the initial PR confused active channel and active client: https://github.com/jamulussoftware/jamulus/pull/2550#discussion_r835643759

This difference needs documentation. IMO: Add (like for JSON-RPC) respective technical documentation before the respective methods. As end goal: Write them for every function, for now: Write docs while changing.

src/clientsettingsdlg.cpp
ann0see: -> New small PR.

src/main.cpp
ann0see: I think that's up to discussion. I do understand that one could put this functionality here. How it's handled currently also seems somewhat messy.

src/channel.cpp
ann0see: I think there's a reason for it. 
It's probably trying to ensure that bDisconnectAndDisable is representing the client being in process of disconnecting?

src/client.cpp
ann0see: Should probably be reverted...

src/channel.cpp
pljones: I don't see what job it's doing that setting `iConTimeOut` to 1 doesn't already do.

src/clientsettingsdlg.cpp
ann0see: Factored out to https://github.com/jamulussoftware/jamulus/pull/3369

src/connectdlg.cpp
ann0see: https://github.com/jamulussoftware/jamulus/pull/3370

src/clientdlg.cpp
ann0see: Reverted.

src/client.cpp
ann0see: Reverted

src/channel.h
```diff
@@ -76,7 +76,7 @@ class CChannel : public QObject
 
     void PrepAndSendPacket ( CHighPrioSocket* pSocket, const CVector<uint8_t>& vecbyNPacket, const int iNPacketLen );
 
-    void ResetTimeOutCounter() { iConTimeOut = iConTimeOutStartVal; }
+    void ResetTimeOutCounter() { iConTimeOut = (bDisconnectAndDisable && !bIsServer) ? 1 : iConTimeOutStartVal; }
```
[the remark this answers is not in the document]
ann0see: > On manual disconnect, it should get set to 1, which is less than iAudioFrameSizeSamples, which means on the next time around, the client will disconnect as above.

Which I believe may not happen? Probably the fix is not here though.

src/client.cpp
ann0see: I believe it's mainly moving around the Channel.Disconnect() and Sound.Stop() code around.

src/clientdlg.cpp
```diff
@@ -730,15 +719,12 @@ void CClientDlg::OnConnectDlgAccepted()
             }
         }
 
-        // first check if we are already connected, if this is the case we have to
-        // disconnect the old server first
-        if ( pClient->IsRunning() )
+        // initiate connection
+        if ( pClient->Connect ( strSelectedAddress, strMixerBoardLabel ) )
```
ann0see, line 723: This should be refactored imo. Feels very clumsy.

src/client.h
ann0see: Factored out into https://github.com/jamulussoftware/jamulus/pull/3372

src/clientdlg.cpp
pljones: UI should simply pass any captured details to Client and await feedback.  There shouldn't be any testing of the response.  (Any input validation should be factored out of Client if it needs sharing between Client and ClientDlg.)

ann0see:
Tagged for Jamulus 4. This PR should NOT be merged. However, the comments are still valuable.

pljones:
@mcfnord did you have a reworking of this?  (I've lost a few PRs somewhere.)

mcfnord:
> [!NOTE]

@pljones — yes. The reworking is **#3805** ("Refactor Connect/Disconnect out to CClient with an explicit connection state"), open and awaiting review. It's the same line of work as this PR: both descend from @pgScorpio's #2550. This PR ports #2550 directly onto `main`; #3805 goes through @ann0see's #3372 extract, finishes it, and adds a state machine on top. It's the one to carry forward.

Structurally #3805 does what #3367 asks for: `Connect()` / `Disconnect()` move out of `CClientDlg` into `CClient`, and the dialog only reacts to signals (`OnConnecting` / `OnConnectingFailed` / `OnDisconnect`). It adds an explicit `EConnectionState` (`CS_DISCONNECTED` / `CS_CONNECTING` / `CS_CONNECTED`) owned by `CClient` as the single source of truth, with every transition emitting `ConnectionStateChanged`. Connect-on-startup also moves into `main.cpp`, so it behaves the same under `--nogui`.

That structural change is what actually resolves #2519, and it answers the doubt @ann0see raised in the review here — *"check if the changed conditions in the if statements are really correct or if the bug lays elsewhere."* It does lie elsewhere. The GUI's "connected" is `CClient::IsRunning()`, which is `Sound.IsRunning()` — audio-device state, not network state (`src/client.h:162`). The real network predicate `IsConnected()` = `Channel.IsConnected()` exists (`src/client.h:174`) but the connect/disconnect UI never reads it: `OnConnectDisconBut()` (`src/clientdlg.cpp:786`), `OnConnectDlgAccepted()`, and `closeEvent()` all key off `IsRunning()`. So a device error while connected flips the button to "Connect" over a live connection, and audio starting before the server answers shows "Disconnect" while nothing is connected — both reported directions. (Full trace in [#2519](https://github.com/jamulussoftware/jamulus/issues/2519#issuecomment-5017527114).)

This PR refines the conditions inside that dialog-centric model; #3805 replaces the model with the single `EConnectionState` source of truth, which is the durable version. Once #3805 lands, I'd suggest this one can be closed in its favor — your call.

pljones:
@ann0see are you okay we can close this in favour of https://github.com/jamulussoftware/jamulus/pull/3372 and https://github.com/jamulussoftware/jamulus/pull/3805 ?  I'd like to pick both up together in a single release (i.e. 4.1.0 now).

ann0see:
Yes.

PR #3369, opened 2024-09-13
Refactor UpdateDisplay() to call UpdateUploadRate()

ann0see:

Refactor UpdateDispay() method. This was extracted from https://github.com/jamulussoftware/jamulus/pull/2550
and is related to: https://github.com/jamulussoftware/jamulus/pull/3364/files/21815c1a0708979fe0414c30e0294feaa7632be3#r1759185427

It makes sense to call UpdateUploadRate() from UpdateDisplay() since it gets same functionality to the same place.

Not sure, if the condition is correct, as we do not check if pClient is running, but rather it is connected now.

CHANGELOG: SKIP

**Context: Fixes an issue?**

No.

No. This even adds documentation

Ready for review

Review. Tested locally by disconnecting and connecting via GUI and changing Audio quality and enabling/disabling small network buffers. Network rate shows correctly. Also checked by starting Jamulus with `-c` flag and seems to do what it should do: Label shows correct data if connected and `---` if not connected.

## Checklist

### DIFF
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -1063,25 +1063,31 @@ void CClientSettingsDlg::OnSndCrdBufferDelayButtonGroupClicked ( QAbstractButton
     UpdateDisplay();
 }
 
+/// @method
+/// @brief Sets upstream rate label to current upload rate if the client is connected, else resets label
 void CClientSettingsDlg::UpdateUploadRate()
 {
-    // update upstream rate information label
-    lblUpstreamValue->setText ( QString().setNum ( pClient->GetUploadRateKbps() ) );
-    lblUpstreamUnit->setText ( "kbps" );
-}
-
-void CClientSettingsDlg::UpdateDisplay()
-{
-    // update slider controls (settings might have been changed)
-    UpdateJitterBufferFrame();
-    UpdateSoundCardFrame();
-
-    if ( !pClient->IsRunning() )
+    // update upstream rate information label if needed
+    if ( pClient->IsConnected() )
+    {
+        lblUpstreamValue->setText ( QString().setNum ( pClient->GetUploadRateKbps() ) );
+        lblUpstreamUnit->setText ( "kbps" );
+    }
+    else
     {
         // clear text labels with client parameters
         lblUpstreamValue->setText ( "---" );
         lblUpstreamUnit->setText ( "" );
     }
+}
+
+/// @method
+/// @brief Updates slider controls (settings might have been changed) and upstream rate information label
+void CClientSettingsDlg::UpdateDisplay()
+{
+    UpdateJitterBufferFrame();
+    UpdateSoundCardFrame();
+    UpdateUploadRate();
 }
 
 void CClientSettingsDlg::UpdateDirectoryComboBox()


src/clientsettingsdlg.cpp
```diff
@@ -1063,25 +1063,31 @@ void CClientSettingsDlg::OnSndCrdBufferDelayButtonGroupClicked ( QAbstractButton
     UpdateDisplay();
 }
 
+/// @method
```
ann0see, line 1066: Probably no need to add \@method here.

src/clientsettingsdlg.cpp
```diff
@@ -1063,25 +1063,31 @@ void CClientSettingsDlg::OnSndCrdBufferDelayButtonGroupClicked ( QAbstractButton
...
+/// @brief Updates slider controls (settings might have been changed) and upstream rate information label
 void CClientSettingsDlg::UpdateDisplay()
 {
-    // update slider controls (settings might have been changed)
     UpdateJitterBufferFrame();
     UpdateSoundCardFrame();
-
-    if ( !pClient->IsRunning() )
```
pljones, line 1079: This is
```
```

src/clientsettingsdlg.cpp
```diff
@@ -1063,25 +1063,31 @@ void CClientSettingsDlg::OnSndCrdBufferDelayButtonGroupClicked ( QAbstractButton
...
 
+/// @method
+/// @brief Sets upstream rate label to current upload rate if the client is connected, else resets label
 void CClientSettingsDlg::UpdateUploadRate()
 {
-    // update upstream rate information label
-    lblUpstreamValue->setText ( QString().setNum ( pClient->GetUploadRateKbps() ) );
-    lblUpstreamUnit->setText ( "kbps" );
+    // update upstream rate information label if needed
+    if ( pClient->IsConnected() )
```
pljones, line 1071: This is
```
bool IsConnected() { return Channel.IsConnected(); }
```

src/clientsettingsdlg.cpp
```diff
@@ -1063,25 +1063,31 @@ void CClientSettingsDlg::OnSndCrdBufferDelayButtonGroupClicked ( QAbstractButton
...
+    {
+        lblUpstreamValue->setText ( QString().setNum ( pClient->GetUploadRateKbps() ) );
```
pljones, line 1073: This is
```
int GetUploadRateKbps() { return Channel.GetUploadRateKbps(); }
```
So it makes far more sense to check the `Channel.IsConnected()`.

But really, it's down to the audio settings, nothing to do with whether a sound card is connected or a server is connected:
image
It can be computed from those values, I think.  (But it's probably the Channel that gets passed the values to do the computation - it gets passed uncompressed audio and does the number crunching.)

pljones (approved):

src/clientsettingsdlg.cpp
ann0see: Yes. That's from https://github.com/jamulussoftware/jamulus/pull/2550/files#diff-d5cde46c72d0c085d1d2c0ca59957407b17931c2f194a33268b3535169006629R1042

A potential semantic change. Probably worth reverting to IsRunning() unless proven to be incorrect.

ann0see: I'd like to hear a second OK/not OK from someone nevertheless.

pljones:
Currently we show the rate only on the settings page.

The factor that connecting to a server affects in the calculation is the "Small Buffers" overhead, I think.  I can't think of anything else that would need to check for server support before performing the calculation.

Originally - before "Small Buffers" - I _think_ it was calculated just from the audio settings, as the network overheads were fixed.

So I guess this change makes sense and is, therefore, long overdue.

ann0see:
@softins could you please have a look at this?

softins (approved):
Code looks sensible. Compiled and tested ok on a Pi.

PR #3372, opened 2024-09-13
Refactor Connect and Disconnect functionality out to CClient

### LINKED ISSUE #3367: Refactor Connect() and Disconnect() functions in GUI to allow JSON RPC to update UI

Besides multiple other coupling issues, we should refactor the Connect() and Disconnect() functions in

https://github.com/jamulussoftware/jamulus/blob/main/src/clientdlg.cpp#L1196-L1296

to be able to call a method on the client to update the GUI state if necessary via JSON-RPC.

#3364 moves them to client.cpp and introduces signals and is probably a good starting point. 

No. However it’s a known issue that the GUI holds state and cannot be controlled that well in its current state programmatically.

ann0see:
This is an extract from https://github.com/jamulussoftware/jamulus/pull/2550

Refactor Connect() and disconnect functionality.
Introduces a new Connect() method for quick connection and adds signals to Start() and Stop() of client.

CHANGELOG: Refactoring: Move Connect() functionality to CClient.

**Context: Fixes an issue?**

Fixes: #3367 

No

Ready for review. Start for refactoring. Most likely overlaps and to a large extent relates somehow to https://github.com/jamulussoftware/jamulus/pull/3364

Review, clarification, probably documentation and refactoring.

## Checklist

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -27,7 +27,6 @@
 /* Implementation *************************************************************/
 CClient::CClient ( const quint16  iPortNumber,
                    const quint16  iQosNumber,
-                   const QString& strConnOnStartupAddress,
                    const QString& strMIDISetup,
                    const bool     bNoAutoJackConnect,
                    const QString& strNClientName,
@@ -121,7 +120,7 @@ CClient::CClient ( const quint16  iPortNumber,
     // The first ConClientListMesReceived handler performs the necessary cleanup and has to run first:
     QObject::connect ( &Channel, &CChannel::ConClientListMesReceived, this, &CClient::OnConClientListMesReceived );
 
-    QObject::connect ( &Channel, &CChannel::Disconnected, this, &CClient::Disconnected );
+    QObject::connect ( &Channel, &CChannel::Disconnected, this, &CClient::Stop );
 
     QObject::connect ( &Channel, &CChannel::NewConnection, this, &CClient::OnNewConnection );
 
@@ -183,13 +182,6 @@ CClient::CClient ( const quint16  iPortNumber,
     // start the socket (it is important to start the socket after all
     // initializations and connections)
     Socket.Start();
-
-    // do an immediate start if a server address is given
-    if ( !strConnOnStartupAddress.isEmpty() )
-    {
-        SetServerAddr ( strConnOnStartupAddress );
-        Start();
-    }
 }
 
 CClient::~CClient()
@@ -578,6 +570,9 @@ bool CClient::SetServerAddr ( QString strNAddr )
         // apply address to the channel
         Channel.SetAddress ( HostAddress );
 
+        // By default, set server name to HostAddress. If using the Connect() method, this may be overwritten
+        SetConnectedServerName ( HostAddress.toString() );
+
         return true;
     }
     else
@@ -865,11 +860,8 @@ void CClient::OnHandledSignal ( int sigNum )
     {
     case SIGINT:
     case SIGTERM:
-        // if connected, terminate connection (needed for headless mode)
-        if ( IsRunning() )
-        {
-            Stop();
-        }
+        // if connected, Stop client (needed for headless mode)
+        Stop();
 
         // this should trigger OnAboutToQuit
         QCoreApplication::instance()->exit();
@@ -979,8 +971,14 @@ void CClient::Start()
 
     // start audio interface
     Sound.Start();
+
+    emit Connected ( GetConnectedServerName() );
 }
 
+/// @method
+/// @brief Stops client and disconnects from server
+/// @emit Disconnected
+///       Use to set CClientDlg to show not being connected
 void CClient::Stop()
 {
     // stop audio interface
@@ -1013,6 +1011,53 @@ void CClient::Stop()
     // reset current signal level and LEDs
     bJitterBufferOK = true;
     SignalLevelMeter.Reset();
+
+    // emit Disconnected() to inform UI of disconnection
+    emit Disconnected();
+}
+
+/// @method
+/// @brief Stops the client if the client is running
+/// @emit Disconnected
+void CClient::Disconnect()
+{
+    if ( IsRunning() )
+    {
+        Stop();
+    }
+}
+
+/// @method
+/// @brief Connects to strServerAddress
+/// @emit Connected (strServerName) if the client wasn't running and SetServerAddr was valid. emit happens through Start().
+///       Use to set CClientDlg to show being connected
+/// @emit ConnectingFailed (error) if an error occurred
+///       Use to display error message in CClientDlg
+/// @param strServerAddress - the server address to connect to
+/// @param strServerName - the String argument to be passed to Connecting()
+void CClient::Connect ( QString strServerAddress, QString strServerName )
+{
+    try
+    {
+        if ( !IsRunning() )
+        {
+            // Set server address and connect if valid address was supplied
+            if ( SetServerAddr ( strServerAddress ) )
+            {
+                SetConnectedServerName ( strServerName );
+                Start();
+            }
+            else
+            {
+                throw CGenErr ( tr ( "Received invalid server address. Please check for typos in the provided server address." ) );
+            }
+        }
+    }
+    catch ( const CGenErr& generr )
+    {
+        Stop();
+        emit ConnectingFailed ( generr.GetErrorText() );
+    }
 }
 
 void CClient::Init()
--- a/src/client.h
+++ b/src/client.h
@@ -126,7 +126,6 @@ class CClient : public QObject
 public:
     CClient ( const quint16  iPortNumber,
               const quint16  iQosNumber,
-              const QString& strConnOnStartupAddress,
               const QString& strMIDISetup,
               const bool     bNoAutoJackConnect,
               const QString& strNClientName,
@@ -137,6 +136,13 @@ public:
 
     void Start();
     void Stop();
+    void Disconnect();
+    void Connect ( QString strServerAddress, QString strServerName );
+
+    // The ConnectedServerName is emitted by Connected() to update the UI with a human readable server name
+    void    SetConnectedServerName ( const QString strServerName ) { strConnectedServerName = strServerName; };
+    QString GetConnectedServerName() const { return strConnectedServerName; };
+
     bool IsRunning() { return Sound.IsRunning(); }
     bool IsCallbackEntered() const { return Sound.IsCallbackEntered(); }
     bool SetServerAddr ( QString strNAddr );
@@ -306,6 +312,9 @@ protected:
     void FreeClientChannel ( const int iServerChannelID );
     int  FindClientChannel ( const int iServerChannelID, const bool bCreateIfNew ); // returns a client channel ID or INVALID_INDEX
     bool ReorderLevelList ( CVector<uint16_t>& vecLevelList );                      // modifies vecLevelList, passed by reference
+    // information for the connected server
+
+    QString strConnectedServerName;
 
     // only one channel is needed for client application
     CChannel  Channel;
@@ -417,7 +426,8 @@ protected slots:
     {
         if ( InetAddr == Channel.GetAddress() )
         {
-            emit Disconnected();
+            // Stop client in case it received a Disconnection request
+            Stop();
         }
     }
     void OnCLPingReceived ( CHostAddress InetAddr, int iMs );
@@ -459,7 +469,11 @@ signals:
 
     void CLChannelLevelListReceived ( CHostAddress InetAddr, CVector<uint16_t> vecLevelList );
 
+    void Connected ( QString strServerName );
+    void ConnectingFailed ( QString errorMessage );
+    void DisconnectClient();
     void Disconnected();
+
     void SoundDeviceChanged ( QString strError );
     void ControllerInFaderLevel ( int iChannelIdx, int iValue );
     void ControllerInPanValue ( int iChannelIdx, int iValue );
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -27,7 +27,6 @@
 /* Implementation *************************************************************/
 CClientDlg::CClientDlg ( CClient*         pNCliP,
                          CClientSettings* pNSetP,
-                         const QString&   strConnOnStartupAddress,
                          const QString&   strMIDISetup,
                          const bool       bNewShowComplRegConnList,
                          const bool       bShowAnalyzerConsole,
@@ -272,14 +271,6 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     TimerCheckAudioDeviceOk.setSingleShot ( true ); // only check once after connection
     TimerDetectFeedback.setSingleShot ( true );
 
-    // Connect on startup ------------------------------------------------------
-    if ( !strConnOnStartupAddress.isEmpty() )
-    {
-        // initiate connection (always show the address in the mixer board
-        // (no alias))
-        Connect ( strConnOnStartupAddress, strConnOnStartupAddress );
-    }
-
     // File menu  --------------------------------------------------------------
     QMenu* pFileMenu = new QMenu ( tr ( "&File" ), this );
 
@@ -481,7 +472,11 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // other
     QObject::connect ( pClient, &CClient::ConClientListMesReceived, this, &CClientDlg::OnConClientListMesReceived );
 
-    QObject::connect ( pClient, &CClient::Disconnected, this, &CClientDlg::OnDisconnected );
+    QObject::connect ( pClient, &CClient::Connected, this, &CClientDlg::OnConnect );
+
+    QObject::connect ( pClient, &CClient::ConnectingFailed, this, &CClientDlg::OnConnectingFailed );
+
+    QObject::connect ( pClient, &CClient::Disconnected, this, &CClientDlg::OnDisconnect );
 
     QObject::connect ( pClient, &CClient::ChatTextReceived, this, &CClientDlg::OnChatTextReceived );
 
@@ -616,11 +611,8 @@ void CClientDlg::closeEvent ( QCloseEvent* Event )
     ConnectDlg.close();
     AnalyzerConsole.close();
 
-    // if connected, terminate connection
-    if ( pClient->IsRunning() )
-    {
-        pClient->Stop();
-    }
+    // Disconnect if needed
+    pClient->Disconnect();
 
     // make sure all current fader settings are applied to the settings struct
     MainMixerBoard->StoreAllFaderSettings();
@@ -738,15 +730,11 @@ void CClientDlg::OnConnectDlgAccepted()
             }
         }
 
-        // first check if we are already connected, if this is the case we have to
-        // disconnect the old server first
-        if ( pClient->IsRunning() )
-        {
-            Disconnect();
-        }
+        // Disconnect the client. We could be currently connected.
+        pClient->Disconnect();
 
         // initiate connection
-        Connect ( strSelectedAddress, strMixerBoardLabel );
+        pClient->Connect ( strSelectedAddress, strMixerBoardLabel );
 
         // reset flag
         bConnectDlgWasShown = false;
@@ -758,11 +746,12 @@ void CClientDlg::OnConnectDisconBut()
     // the connect/disconnect button implements a toggle functionality
     if ( pClient->IsRunning() )
     {
-        Disconnect();
-        SetMixerBoardDeco ( RS_UNDEFINED, pClient->GetGUIDesign() );
+        pClient->Stop();
     }
     else
     {
+        // If the client isn't running, we assume that we weren't connected. Thus show the connect dialog
+        // TODO: Refactor to have robust error handling
         ShowConnectionSetupDialog();
     }
 }
@@ -870,7 +859,7 @@ void CClientDlg::OnLicenceRequired ( ELicenceType eLicenceType )
         // disconnect from that server.
         if ( !LicenceDlg.exec() )
         {
-            Disconnect();
+            pClient->Disconnect();
         }
 
         // unmute the client output stream if local mute button is not pressed
@@ -1173,10 +1162,7 @@ void CClientDlg::OnSoundDeviceChanged ( QString strError )
     if ( !strError.isEmpty() )
     {
         // the sound device setup has a problem, disconnect any active connection
-        if ( pClient->IsRunning() )
-        {
-            Disconnect();
-        }
+        pClient->Disconnect();
 
         // show the error message of the device setup
         QMessageBox::critical ( this, APP_NAME, strError, tr ( "Ok" ), nullptr );
@@ -1204,65 +1190,38 @@ void CClientDlg::OnCLPingTimeWithNumClientsReceived ( CHostAddress InetAddr, int
     ConnectDlg.SetPingTimeAndNumClientsResult ( InetAddr, iPingTime, iNumClients );
 }
 
-void CClientDlg::Connect ( const QString& strSelectedAddress, const QString& strMixerBoardLabel )
+void CClientDlg::OnConnect ( const QString& strMixerBoardLabel )
 {
-    // set address and check if address is valid
-    if ( pClient->SetServerAddr ( strSelectedAddress ) )
+
+    // hide label connect to server
+    lblConnectToServer->hide();
+    lbrInputLevelL->setEnabled ( true );
+    lbrInputLevelR->setEnabled ( true );
+
+    // change connect button text to "disconnect"
+    butConnect->setText ( tr ( "&Disconnect" ) );
+
+    // set server name in audio mixer group box title
+    MainMixerBoard->SetServerName ( strMixerBoardLabel );
+
+    // start timer for level meter bar and ping time measurement
+    TimerSigMet.start ( LEVELMETER_UPDATE_TIME_MS );
+    TimerBuffersLED.start ( BUFFER_LED_UPDATE_TIME_MS );
+    TimerPing.start ( PING_UPDATE_TIME_MS );
+    TimerCheckAudioDeviceOk.start ( CHECK_AUDIO_DEV_OK_TIME_MS ); // is single shot timer
+
+    // audio feedback detection
+    if ( pSettings->bEnableFeedbackDetection )
     {
-        // try to start client, if error occurred, do not go in
-        // running state but show error message
-        try
-        {
-            if ( !pClient->IsRunning() )
-            {
-                pClient->Start();
-            }
-        }
-
-        catch ( const CGenErr& generr )
-        {
-            // show error message and return the function
-            QMessageBox::critical ( this, APP_NAME, generr.GetErrorText(), "Close", nullptr );
-            return;
-        }
-
-        // hide label connect to server
-        lblConnectToServer->hide();
-        lbrInputLevelL->setEnabled ( true );
-        lbrInputLevelR->setEnabled ( true );
-
-        // change connect button text to "disconnect"
-        butConnect->setText ( tr ( "&Disconnect" ) );
-
-        // set server name in audio mixer group box title
-        MainMixerBoard->SetServerName ( strMixerBoardLabel );
-
-        // start timer for level meter bar and ping time measurement
-        TimerSigMet.start ( LEVELMETER_UPDATE_TIME_MS );
-        TimerBuffersLED.start ( BUFFER_LED_UPDATE_TIME_MS );
-        TimerPing.start ( PING_UPDATE_TIME_MS );
-        TimerCheckAudioDeviceOk.start ( CHECK_AUDIO_DEV_OK_TIME_MS ); // is single shot timer
-
-        // audio feedback detection
-        if ( pSettings->bEnableFeedbackDetection )
-        {
-            TimerDetectFeedback.start ( DETECT_FEEDBACK_TIME_MS ); // single shot timer
-            bDetectFeedback = true;
-        }
+        TimerDetectFeedback.start ( DETECT_FEEDBACK_TIME_MS ); // single shot timer
+        bDetectFeedback = true;
     }
 }
 
-void CClientDlg::Disconnect()
-{
-    // only stop client if currently running, in case we received
-    // the stopped message, the client is already stopped but the
-    // connect/disconnect button and other GUI controls must be
-    // updated
-    if ( pClient->IsRunning() )
-    {
-        pClient->Stop();
-    }
+void CClientDlg::OnConnectingFailed ( const QString& strError ) { QMessageBox::critical ( this, APP_NAME, strError, tr ( "Close" ), nullptr ); }
 
+void CClientDlg::OnDisconnect()
+{
     // change connect button text to "connect"
     butConnect->setText ( tr ( "C&onnect" ) );
 
@@ -1304,6 +1263,9 @@ void CClientDlg::Disconnect()
 
     // clear mixer board (remove all faders)
     MainMixerBoard->HideAll();
+
+    // Reset the deco
+    SetMixerBoardDeco ( RS_UNDEFINED, pClient->GetGUIDesign() );
 }
 
 void CClientDlg::UpdateDisplay()
--- a/src/clientdlg.h
+++ b/src/clientdlg.h
@@ -76,7 +76,6 @@ class CClientDlg : public CBaseDlg, private Ui_CClientDlgBase
 public:
     CClientDlg ( CClient*         pNCliP,
                  CClientSettings* pNSetP,
-                 const QString&   strConnOnStartupAddress,
                  const QString&   strMIDISetup,
                  const bool       bNewShowComplRegConnList,
                  const bool       bShowAnalyzerConsole,
@@ -94,8 +93,6 @@ protected:
     void ShowAnalyzerConsole();
     void UpdateAudioFaderSlider();
     void UpdateRevSelection();
-    void Connect ( const QString& strSelectedAddress, const QString& strMixerBoardLabel );
-    void Disconnect();
     void ManageDragNDrop ( QDropEvent* Event, const bool bCheckAccept );
     void SetPingTime ( const int iPingTime, const int iOverallDelayMs, const CMultiColorLED::ELightColor eOverallDelayLEDColor );
 
@@ -127,6 +124,9 @@ protected:
     CAnalyzerConsole   AnalyzerConsole;
 
 public slots:
+    void OnConnect ( const QString& strServerName );
+    void OnConnectingFailed ( const QString& strErrorText );
+    void OnDisconnect();
     void OnConnectDisconBut();
     void OnTimerSigMet();
     void OnTimerBuffersLED();
@@ -233,7 +233,6 @@ public slots:
     }
 
     void OnConnectDlgAccepted();
-    void OnDisconnected() { Disconnect(); }
     void OnGUIDesignChanged();
     void OnMeterStyleChanged();
     void OnRecorderStateReceived ( ERecorderState eRecorderState );
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -922,14 +922,7 @@ int main ( int argc, char** argv )
         {
             // Client:
             // actual client object
-            CClient Client ( iPortNumber,
-                             iQosNumber,
-                             strConnOnStartupAddress,
-                             strMIDISetup,
-                             bNoAutoJackConnect,
-                             strClientName,
-                             bEnableIPv6,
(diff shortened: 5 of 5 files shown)


src/client.cpp
```diff
@@ -908,6 +905,56 @@ void CClient::Stop()
...
+    return false;
+}
+
+/// @method
+/// @brief Disconnects client
+/// @emit Disconnected
+/// @result true if client wasn't running, false otherwise
+bool CClient::Disconnect()
+{
+    if ( IsRunning() )
```
ann0see, line 939: I have the feeling that we MUST clarrify this. 
The code in #3364 is different.

src/clientdlg.cpp
```diff
@@ -730,16 +735,14 @@ void CClientDlg::OnConnectDlgAccepted()
             }
         }
 
-        // first check if we are already connected, if this is the case we have to
-        // disconnect the old server first
-        if ( pClient->IsRunning() )
+        // initiate connection
+        // TODO: Refactor this for failing call on Connect()
+
+        if ( pClient->Connect ( strSelectedAddress, strMixerBoardLabel ) )
```
ann0see, line 741: Doesn't seem to call OnConnect() via signal.

src/client.cpp
```diff
@@ -908,6 +905,56 @@ void CClient::Stop()
     SignalLevelMeter.Reset();
 }
 
+/// @method
+/// @brief Connects to strServerAddress
+/// @emit Connecting (strServerName) if the client wasn't running and SetServerAddr returned true.
+/// @param strServerAddress - the server address to connect to
+/// @param strServerName - the String argument to be passed to Connecting()
+/// @result true if client wasn't running and SetServerAddr returned true, false otherwise
```
pljones, line 913: I don't like this.  `Connect` says "Go off and connect".  It should either emit "Connected" or "ConnectionFailed" signals, not _return_ a value.  The point is that ClientDlg and Client can be on separate threads and you need to use the slots/signals to communicate.

src/client.cpp
```diff
@@ -908,6 +905,56 @@ void CClient::Stop()
...
+    {
+        // Set server address and connect if valid address was supplied
+        if ( SetServerAddr ( strServerAddress ) )
+        {
+
+            Start();
+
+            emit Connecting ( strServerName );
+
+            return true;
```
pljones, line 926: So delete the return.

src/client.cpp
```diff
@@ -908,6 +905,56 @@ void CClient::Stop()
...
+        }
+    }
+
+    return false;
```
pljones, line 930: And replace this with an `else emit ConnectionFailed`.

src/client.cpp
```diff
@@ -908,6 +905,56 @@ void CClient::Stop()
...
+/// @emit Connecting (strServerName) if the client wasn't running and SetServerAddr returned true.
+/// @param strServerAddress - the server address to connect to
+/// @param strServerName - the String argument to be passed to Connecting()
+/// @result true if client wasn't running and SetServerAddr returned true, false otherwise
+bool CClient::Connect ( QString strServerAddress, QString strServerName )
+{
+    if ( !IsRunning() )
+    {
+        // Set server address and connect if valid address was supplied
+        if ( SetServerAddr ( strServerAddress ) )
```
pljones, line 919: I don't like this style at all.  If there's some validation to happen, it should have happened before we get this far.  We shouldn't _try_ to connect to an address that's not invalid.

Perhaps this whole routine is redundant and `SetServerAddr` should be called `Connect` and do the emits?

src/client.cpp
```diff
@@ -908,6 +905,56 @@ void CClient::Stop()
     SignalLevelMeter.Reset();
 }
 
+/// @method
```
pljones, line 908: This should also document what signals are emitted.  The signals should be documented as to how they're intended to be used, too.

src/client.cpp
pljones: I've no idea what this is trying to encapsulate.

src/clientdlg.cpp
```diff
@@ -277,7 +277,13 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
...
         // initiate connection (always show the address in the mixer board
         // (no alias))
-        Connect ( strConnOnStartupAddress, strConnOnStartupAddress );
+
+        // initiate connection
+        // TODO: Refactor this for failing call on Connect()
+
+        pClient->Connect ( strConnOnStartupAddress, strConnOnStartupAddress );
+        // TODO: Find out why without this the mixer status issue still occurs. 
+        OnConnect ( strConnOnStartupAddress );
```
pljones, line 286: Something called `OnAnything` should only be used as a connected signal handler, not called directly like this.

src/clientdlg.cpp
```diff
@@ -277,7 +277,13 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     {
         // initiate connection (always show the address in the mixer board
         // (no alias))
-        Connect ( strConnOnStartupAddress, strConnOnStartupAddress );
+
```
pljones, line 280: I don't think ClientDlg should be responsible for any of this.

Client should handle it.  It should get passed a ClientDlg instance (or null if headless), initialise the ClientDlg (if present), initialise audio and then initiate any "on start up" server connection.  That should be enough to get everything to "just work".

src/clientdlg.cpp
```diff
@@ -1193,65 +1193,36 @@ void CClientDlg::OnCLPingTimeWithNumClientsReceived ( CHostAddress InetAddr, int
     ConnectDlg.SetPingTimeAndNumClientsResult ( InetAddr, iPingTime, iNumClients );
 }
 
-void CClientDlg::Connect ( const QString& strSelectedAddress, const QString& strMixerBoardLabel )
+void CClientDlg::OnConnect ( const QString& strMixerBoardLabel )
```
pljones, line 1196: Ideally, this signal would get issued only on successful connection.

As we're going cross-thread here, we need Client to supply _any_ listener -- ClientDlg or the Client JSON-RPC -- with full details of the newly established connection.

We should follow this rule for _all_ state changes in the Client -- that should be the _only_ way for ClientDlg and the Client JSON-RPC to acquire knowledge of Client state, no calls to functions across threads.

(The same is true on the server side, too.)

Similarly, to change the state of the Client, the Client should expose slots and ClientDlg or Client JSON-RPC should signal those slots to effect the change in state.  Not call methods across threads.

This does demand a major restructure but I think this could be a good place to start.  Set out _what state changes_ when the Client successfully establishes a connection to a server, create an object instance containing that information, and issue a signal passing that object.  This gives a clearly defined API to that change in state.

Either the message on the signal could be minimal, covering _only_ the state that changes on any connect, or extensive, supplying the _whole_ Client state.  In the latter case, every signal from the Client would pass a new state instance with the full current Client state, making the listener's job much easier.  (I don't like it as much as the minimal approach but it's likely much easier to explain and maintain.  It smacks of "global variables" a bit too much, though...)

src/clientdlg.cpp
```diff
@@ -1193,65 +1193,36 @@ void CClientDlg::OnCLPingTimeWithNumClientsReceived ( CHostAddress InetAddr, int
...
 {
-    // only stop client if currently running, in case we received
-    // the stopped message, the client is already stopped but the
-    // connect/disconnect button and other GUI controls must be
-    // updated
-    if ( pClient->IsRunning() )
-    {
-        pClient->Stop();
-    }
-
```
pljones, line 1254: It's good to see this gone.  Direct method calls like this don't look good.  (Qt says the pointers are thread-safe, I still don't like it -- I much prefer using only signal/slot calls.)

src/client.cpp
```diff
@@ -773,11 +773,8 @@ void CClient::OnHandledSignal ( int sigNum )
     {
     case SIGINT:
     case SIGTERM:
-        // if connected, terminate connection (needed for headless mode)
-        if ( IsRunning() )
-        {
-            Stop();
-        }
+        // if connected, disconnect (needed for headless mode)
+        Disconnect();
```
pljones, line 777: Maybe `OnAboutToQuit` should handle all this.  It feels cleaner to me to have only the `QCoreApplication::instance()->exit()` call here and deal with everything else in the handlers.

Again, if _different handlers_ are needed for UI/RPC/headless, they can be wired up appropriately on start-up.

Start up would be something like:
- Create a Client instance
- If requested, create a ClientDlg instance, passing only the pointer to the Client instance
  - Wire up Client public signals to ClientDlg private slots
  - Wire up ClientDlg private signals to Client public slots
- If requested, create a Client JSON-RPC instance, passing only the pointer to the Client instance
  - Wire up Client public signals to Client JSON-RPC private slots
  - Wire up Client JSON-RPC private signals to Client public slots
- Pass Client the start up parameters
  - effect Client start up, emitting any such signals as are needed when state changes

Thus, when exiting, it's the Client instance that's in control.  It says "OK, time to quit".  It should have itself wired up any internal Audio and (network) Channel slots to clean up the connection.  All the ClientDlg and Client JSON-RPC need to know is that exit is happening and to shut themselves down cleanly.

src/clientdlg.cpp
pljones: Changed my mind a bit about this...

That's the problem with "small changes" that affect the way the application works.  They're not really "small".

src/client.cpp
ann0see: > SetServerAddr

Should just do exactly what the name implies IMO - Otherwise we could introduce mixing of functionality.
Connect() would then emit the signals. That's cleaner and closer to what we do now.

src/client.cpp
ann0see: @ emit does this. But Ofc, we can rewrite what we want to document.

src/client.cpp
ann0see: Agree. Potentially a larger rewrite.

src/clientdlg.cpp
ann0see: See https://github.com/jamulussoftware/jamulus/pull/3372/files/0b6a2bb3c0c70329c85b48f0a6ae52c7e226c2f4#r1759507277

src/clientdlg.cpp
ann0see: The global approach could also be wasteful while the small approach could be more difficult. If we restructure correctly, we should aim at the small approach.

src/client.cpp
ann0see: > that exit is happening and to shut themselves down cleanly.

Which should then in turn also fix the connection status issue if implemented correctly.

I'll have a thought about this.

src/clientdlg.cpp
```diff
@@ -1193,65 +1193,36 @@ void CClientDlg::OnCLPingTimeWithNumClientsReceived ( CHostAddress InetAddr, int
...
 
-void CClientDlg::Connect ( const QString& strSelectedAddress, const QString& strMixerBoardLabel )
+void CClientDlg::OnConnect ( const QString& strMixerBoardLabel )
 {
-    // set address and check if address is valid
-    if ( pClient->SetServerAddr ( strSelectedAddress ) )
-    {
-        // try to start client, if error occurred, do not go in
-        // running state but show error message
-        try
```
ann0see, line 1203: This exception handling should probably also be in the client and issue a ConnectionFailed(message) signal.

src/clientdlg.cpp
ann0see: Potentially, we could just emit pClient->Connect(...)?

src/client.h
```diff
@@ -121,6 +121,9 @@ class CClient : public QObject
 
     void Start();
     void Stop();
+    bool Connect ( QString strServerAddress, QString strServerName );
+    bool Disconnect();
+
     bool IsRunning() { return Sound.IsRunning(); }
     bool IsCallbackEntered() const { return Sound.IsCallbackEntered(); }
```
ann0see, line 128: IMO this is flawed. Sound should emit a signal if it started correctly, else emit an error signal. However, implementing this opens another pandora box.

ann0see:
* channel.cpp emits Disconnected() in GetData --> may need some refactoring

src/clientdlg.h
```diff
@@ -232,7 +233,6 @@ public slots:
     }
 
     void OnConnectDlgAccepted();
-    void OnDisconnected() { Disconnect(); }
```
ann0see, line 235: This introduces a bug: disconnecting server side yields to a state where the UI is updated, but the client isn't.

src/clientdlg.cpp
```diff
@@ -1193,64 +1196,42 @@ void CClientDlg::OnCLPingTimeWithNumClientsReceived ( CHostAddress InetAddr, int
...
-    if ( pClient->IsRunning() )
-    {
-        pClient->Stop();
-    }
+    QMessageBox::critical ( this, APP_NAME, strError, "Close", nullptr );
+}
+
+void CClientDlg::OnDisconnect()
+{
+    // channel.cpp also issues Disconnected()
```
ann0see, line 1234: Now DisconnectClient()

src/clientdlg.h
ann0see: Ok. I believe it was https://github.com/jamulussoftware/jamulus/pull/3372/files#diff-2ee628bb1cdcbf04f484b57ea20ddf172db2dfaeb22214ebfb24ff36718a7509R393

src/channel.cpp
```diff
@@ -654,7 +654,7 @@ EGetDataStat CChannel::GetData ( CVector<uint8_t>& vecbyData, const int iNumByte
         Protocol.Reset();
 
         // emit message
-        emit Disconnected();
+        emit DisconnectClient();
```
ann0see, line 657: Could also be reverted...

src/clientdlg.cpp
ann0see: Probably the issue is that the connection happens in the constructor.
It would make sense to remove Connecting from the constructor and do it as in the other PR https://github.com/jamulussoftware/jamulus/pull/3364#discussion_r1759236825

src/client.cpp
ann0see: I think it's not part of this PR though. Could you please open an issue?

src/clientdlg.h
ann0see: Should be fixed now.

src/clientdlg.cpp
ann0see: Ok. See https://forum.qt.io/topic/111772/signals-and-slots-within-same-class/3

src/client.cpp
ann0see: I mean we could probably emit in the Stop() method already...

src/client.cpp
```diff
@@ -908,6 +905,56 @@ void CClient::Stop()
...
+/// @method
+/// @brief Disconnects client. If the client is not running, it just stops Sound
+/// @emit Disconnected
+///       Use to set CClientDlg to show not being connected
+void CClient::Disconnect()
+{
+    if ( IsRunning() )
+    {
+        Stop();
+        emit Disconnected();
```
ann0see, line 948: Anyway, put this after the if

src/clientdlg.cpp
pljones: Yup: I said somewhere else, I think -- create the client, clientdlg and client json-rpc, wire them up, then tell the client to start running (passing in the start up details).  It would be in the "start running" that the network connection was initiated and signals start getting emitted from the network channel.

src/clientdlg.cpp
```diff
@@ -750,11 +750,12 @@ void CClientDlg::OnConnectDisconBut()
     // the connect/disconnect button implements a toggle functionality
```
pljones, line 750: This should be implemented by switching the `connect` setting for the signal handler for the button, rather than having one signal handler with an if that consults the state of the client.

(Button life-cycle:
- default: onClickConnect (connects ConnectDlg onClick handler to onConnectDlgConnect and) launches ConnectDlg
- onConnectDlgConnect _disables_ the button and asks the client to connect
- onConnect switches the onClick handler to onClickDisconnect and enables the button
- onConnectFailed just enables the button (keeping the onClickConnect handler)
- onClickDisconnect _disables_ the button and asks the client to disconnect

pljones: For JSON-RPC, the equivalent is
- (get a list of servers from somewhere)
- Send a Client JSON-RPC request to connect to a server - get an "ACK" back to say message received
- At some asynchronous time in the future, receive a Client JSON-RPC push of either a successful connect or failed connect
- (assuming connected) Send a Client JSON-RPC request to disconnect the current connection - get an "ACK" back to say message received
- At some asynchronous time in the future, receive a Client JSON-RPC push when the client decides it has disconnected

Note that in both the UI and JSON-RPC cases, the actual change in client state is asynchronous.  For disconnect, it's possible (given that the client might not actually be connected), that _nothing will happen_.

Now, for that "nothing happened" case, we've a potential problem: the UI/JSON-RPC "thought" we were connected and asked to disconnect.  Is asking the client to disconnect all that's needed or should the UI/JSON-RPC _wait_ for confirmation of the disconnect?  If it _should wait_, we then _MUST_ send that "disconnected" message back, whether or not we were connected.

src/client.cpp
```diff
@@ -908,6 +905,56 @@ void CClient::Stop()
...
+    }
+    else
+    {
+        // make sure sound is stopped in any case
+        Sound.Stop();
```
pljones, line 953: Maybe "Sound" should be listening to Client.Disconnected and its OnDisconnected handler should Stop?

ann0see: Sound.Stop() is driver specific, I think.
For ASIO, for example it calls ASIOStop(), iOS calls AudioOutputUnitStop(), JACK nothing, Android closeStreams(). But the important part is that all call CSoundBase::Stop().

In any way, CClient::Stop adds the 100ms timer additionally and sends a disconnect message to the server. Not sure if that's a problem (I'd rather think it isn't)

ann0see: I've for now still included a small method.

src/clientdlg.cpp
ann0see: You rejected connecting in main.cpp here: https://github.com/jamulussoftware/jamulus/pull/3364/files#r1759236825 ?

So how would you do it?

1. Create client
2. Create GUI
3. Create json rpc
4. Connect here by calling pClient->Connect() or what?

src/main.cpp
```diff
@@ -963,6 +961,12 @@ int main ( int argc, char** argv )
                 // show dialog
                 ClientDlg.show();
 
+                // Connect on startup if requested
+                if ( !strConnOnStartupAddress.isEmpty() )
+                {
+                    Client.Connect ( strConnOnStartupAddress, strConnOnStartupAddress );
+                }
+
```
ann0see, line 969: This is not nice, but it seems to crash the app if I factor out the if and pApp->exec() after the else

src/client.cpp
ann0see: ... And removed it...

src/clientdlg.cpp
ann0see: See the changes.

src/clientdlg.cpp
ann0see: Yes, this should be changed. But in a follow up PR

src/client.cpp
```diff
@@ -773,11 +765,8 @@ void CClient::OnHandledSignal ( int sigNum )
     {
     case SIGINT:
     case SIGTERM:
-        // if connected, terminate connection (needed for headless mode)
-        if ( IsRunning() )
-        {
-            Stop();
-        }
+        // if connected, Stop client (needed for headless mode)
+        Stop();
```
pljones, line 769: Client::Stop is a slot which looks like it's meant for handling the CChannel::Disconnected signal.

We shouldn't conflate concepts like this.  Having the (command line interface) send a "stop" signal isn't the same as the network saying it's lost the connection.  `QCoreApplication::instance()->exit()` _should_ be enough here, with appropriate `OnAboutToQuit` handlers.

pljones: So
- Sound.OnAboutToQuit would call Sound.Stop.
- Channel.OnAboutToQuit would call Channel.SetEnable ( false )
- Client.OnAboutToQuit would call Client.Stop -- which wouldn't call Sound.Stop or Channel.SetEnable ( false )

src/clientdlg.cpp
```diff
@@ -608,11 +603,8 @@ void CClientDlg::closeEvent ( QCloseEvent* Event )
...
-    // if connected, terminate connection
-    if ( pClient->IsRunning() )
-    {
-        pClient->Stop();
-    }
+    // Disconnect if needed
+    pClient->Stop();
 
     // make sure all current fader settings are applied to the settings struct
     MainMixerBoard->StoreAllFaderSettings();
```
pljones, line 610: Maybe `closeEvent` should just call `QCoreApplication::instance()->exit();` and then do the work in `OnAboutToQuit` handlers?  That should stop needing to call `XYZ.Stop()`.

ann0see:
Now the coding style check seems to work. Not so sure about the style change in main.cpp:
https://github.com/jamulussoftware/jamulus/actions/runs/10982471650/job/30490757878?pr=3372#step:4:54

src/main.cpp
ann0see: Maybe we could introduce a method or signal which informs the client that the object is allocated and the connection can be established. Then the emits in the connect method would become redundant.

src/clientdlg.cpp
ann0see: https://github.com/jamulussoftware/jamulus/issues/3401

src/client.cpp
ann0see: https://github.com/jamulussoftware/jamulus/issues/3402

src/client.cpp
ann0see: https://github.com/jamulussoftware/jamulus/issues/3402

src/clientdlg.cpp
ann0see: https://github.com/jamulussoftware/jamulus/issues/3402

ann0see:
Please recheck if the important parts are in the two issues I opened.

src/client.cpp
```diff
@@ -906,6 +904,42 @@ void CClient::Stop()
...
+            if ( SetServerAddr ( strServerAddress ) )
+            {
+                SetConnectedServerName ( strServerName );
+                Start();
+            }
+            else
+            {
+                throw CGenErr ( tr ( "Received invalid server address. Please check for typos in the provided server address." ) );
+            }
+        }
```
pljones, line 936: Should this throw some indication it ignored the request?

ann0see: It emits the error message further down

ann0see:
I do not understand why clang format complains about the list style here.

src/clientdlg.cpp
```diff
@@ -608,11 +603,8 @@ void CClientDlg::closeEvent ( QCloseEvent* Event )
     ConnectDlg.close();
     AnalyzerConsole.close();
 
-    // if connected, terminate connection
-    if ( pClient->IsRunning() )
-    {
-        pClient->Stop();
-    }
+    // Disconnect if needed
```
pljones, line 606: I'd do this as the last thing before `Event->accept();`

And do we want to ignore the triggered `CClient::Disconnected` signal that will happen when `CClient::Stop` gets called?

ann0see: > And do we want to ignore the triggered CClient::Disconnected signal that will happen when CClient::Stop gets called?

Probably not.

ann0see: > I'd do this as the last thing before Event->accept();

This gives a small timeframe (in theory) where the fader settings are not stored, but the client disconnected. I don't think that's a problem.

src/clientdlg.cpp
```diff
@@ -730,15 +722,9 @@ void CClientDlg::OnConnectDlgAccepted()
             }
         }
 
-        // first check if we are already connected, if this is the case we have to
-        // disconnect the old server first
-        if ( pClient->IsRunning() )
-        {
-            Disconnect();
-        }
-
```
softins, line 739: I don't think this code should be removed. Or else equivalent code should be added somewhere according to the new structure (which I don't yet fully understand).

It is possible, while connected to a server, to go `File`->`Connection Setup...`, or `Ctrl-C`, to bring up the connect dialog, and then select a new server to connect to. In existing versions of Jamulus, this disconnects from the connected server, and then immediately connects to the new server. This was used to great effect during WorldJam to quickly move between rooms.

In the version built from this PR, this doesn't work. The existing connection continues instead of connecting to the new server.

softins (changes requested):

src/clientdlg.cpp
ann0see: We could probably call something like pClient->Disconnect()/Stop() and then pClient->Connect().
The main idea of this PR is to:
1. Emit signals if the client disconnects or connects
2. Introduce convenient methods for programmatic (JSON-RPC) disconnection and connection.

ann0see: Ok. Fixed. Not sure if we need the if is running then stop.
I'd like to have this in the client then.

src/clientdlg.cpp
```diff
@@ -740,15 +732,12 @@ void CClientDlg::OnConnectDlgAccepted()
...
-        if ( pClient->IsRunning() )
-        {
-            Disconnect();
-        }
+        // Disconnect the client. We could be currently connected.
+        pClient->Stop();
 
         // initiate connection
-        Connect ( strSelectedAddress, strMixerBoardLabel );
+
```
pljones, line 739: Unneeded (and inconsistent) blank line.

pljones (changes requested):

src/client.cpp
```diff
@@ -906,6 +904,53 @@ void CClient::Stop()
...
     SignalLevelMeter.Reset();
+
+    // emit Disconnected() to inform UI of disconnection
+    emit Disconnected();
+}
+
+/// @method
+/// @brief Stops the client if the client is running
+/// @emit Disconnected
+void CClient::Disconnect()
```
ann0see, line 915: Introduced again. Mainly an alias. But now also with the check

src/client.cpp
```diff
@@ -122,7 +121,7 @@ CClient::CClient ( const quint16  iPortNumber,
     QObject::connect ( &Channel, &CChannel::ConClientListMesReceived, this, &CClient::OnConClientListMesReceived );
     QObject::connect ( &Channel, &CChannel::ConClientListMesReceived, this, &CClient::ConClientListMesReceived );
 
-    QObject::connect ( &Channel, &CChannel::Disconnected, this, &CClient::Disconnected );
+    QObject::connect ( &Channel, &CChannel::Disconnected, this, &CClient::Stop );
```
pljones, line 124: If this _only_ gets used _as a slot_ by `CChannel::Disconnected` could we use a name like `OnChannelDisconnected` rather than the somewhat ambiguous `Stop`?

pljones: Oh, I see it's also used just before exiting the client in the `SIGINT`/`SIGTERM` handler...  Hm.

`SIGINT`/`SIGTERM` and "File -> Exit (Ctrl-Q)" should both really do the same thing...

I'd rather we had a lambda here, `(whatever) -> Stop()`, so it was clear that `Stop` wasn't a slot but a normal method.

pljones, line 124: I'm not happy with `Stop` as a name.  What does it _do_?  It doesn't exit the application...

It's used (at least) here and in the `SIGINT`/`SIGTERM` handler.  The latter makes me wonder if it's also used in the `OnAboutToQuit` handling (which should be invoked by the `SIGINT`/`SIGTERM` handler...).

ann0see: Stop() means it stops the client. 
But this concept is not well defined IMO.

pljones:
OK...  so what I'm expecting to see...
- sort out "exit client" so that there's only "one way out"
  - there should be one CClient method that tells Qt the client is exiting and does _nothing_ else.
    - this would be called from GUI (File->Exit) and SIGINT/SIGTERM handlers to make the client close and those would do nothing else.
  - _any and all_ other cleaning up should happen when Qt sends the "OnAboutToQuit" signal.
- work out what "disconnecting from a server" means
  - have a CClient method that performs disconnection
    - does nothing if not connected
    - tells the server the client is disconnecting (?? do we do this ??)
    - calls another CClient method to handle disconnecting
  - have CClient respond to the server disconnecting by calling another CClient method to handle disconnecting (lambda)
  - have a CClient method that handles disconnecting by
    - cleaning up any CClient structures
    - emitting a (blocking) Disconnected signal
  - amend other components to listen for Disconnected and take appropriate clean up actions
- work out what "connecting to a server" means
  - have a CClient method that
    - calls the CClient method to perform disconnection
    - _initiates_ the connection (which could fail)
    - emits a signal indicating a connection has been requested but not yet granted
  - have a CClient method that handles a successful connection
  - have a CClient method that handles an unsuccessful connection

I think most of the Connect stuff is there.  I'm less clear what's happening around Disconnect.  And the exiting the client currently appears very messy - at least, it's confusing me...

ann0see:
Yes, I agree. 
This may imply that we may need to restructure the UI functionality a bit. 
I'll have a look in a few weeks. Currently I have a higher load with another development project.

ann0see:
I would like to define the methods as follows (not sure if this is everything)

CClient is connected iff all of those hold
1. CClient.Init() was called
2. CClient.Connect() was called
3. Channel.IsEnabled() is true
4. Sound.IsRunning() is true
5. CClient.IsRunning() is true (which is to be redefined)
6. There is no soundcard error (to be defined)

CClient will disconnect if at least one of these conditions holds
1. A soundcard error occurs
2. The server disconnects
3. Disconnect() was called

CClient is running iff
1. CClient.Init() was called
2. CClient.Start() was called
3. CClient.IsRunning() returns true
4. Potentially: SetServerAddress() was called and succeeded

This means, there may be a sound card error while CClient is running and the client may not be connected to a server

To transition from CClient being running to CClient being connected, we need to call Connect() and the connection to the server must stand (within the timeout).
To transition from CClient being connected to just being running, we have to:
1. Either call Disconnect()
2. Have any kind of network or audio issue
3. The Server requests a disconnect

Does this make sense?

pljones:
#### Background

(Apologies for writing this out in a PR where you know it as well as I do but I'm trying to make sure my thoughts are in order...)

We need states:
- disconnected
- connection requested but not connected
- connection established and current

"Disconnected" means no connection has been requested or established
"Connection requested but not connected" means a connection has been requested but audio data is not being exchanged yet
"Connection established and current" means a server connection is in progress

Now, these states should apply equally to the Client's view of it's single connection to a Server and Server's view of each of its connections to a Client.  This is important (and a core concept to Jamulus).

The Client Channel owns that state.  The Client has one Channel but the Server has many Channels.  So any changes here must not break the Server.

A Channel is considered disconnected when no more audio data has been exchanged for a certain period (or something, I don't remember this bit) or an explicit disconnect is received (maybe, I'm not sure on that, either).  Again, that goes equally for both ends of the Channel.

(And finally...  The Channel carries Protocol messages that can either be audio or other messages.  The audio messages go into the jitter buffer, triggering audio processing.  The other protocol messages get processed, quite often by the Channel itself but sometimes needing either Client or Server (as they know more about the world) to get involved.  The main difference between Client and Server, of course, is audio processing -- the Client exchanges CSound audio data with the jitter buffer but the Server pushes the audio through the mixer.)

----

> CClient is connected iff all of those hold

OK, let's start with "CClient has a CChannel" and it's "CChannel" that holds state regarding whether an active connection has been established.  Given this is the case (and it has to be for the Client and Server to work using the same code):

> CClient is "connected" or "running" or "started" or whatever, if and only if
> - its CChannel.ConnectionState() is Connected (or "CChannel.IsConnected()")
> - its CSound.IsRunning() is true

Neither of those alone is useful for components outside CClient (other than CChannel and CSound maintaining their own state).  Nothing else needs worrying about.

> CClient.Connect() was called

Maybe we rename CClient.Connect(...) to CClient.RequestConnection(QString server) because that's where all the Client-specific logic and it doesn't end up by connecting.  It would:
- always call a CClient.EndConnection() method that
  - calls CChannel.Disconnect()
    - (somewhere drains the out-going and in-coming audio packets in the jitter buffer - I guess it's here)
  - calls CSound.Stop()
    - (by this point it should have had one packet of silence to output)
  - does any other Client-specific clean up (e.g. clear connected server name)
  - emit a ConnectionEnded signal
- calls CSound.Start()
- calls CChannel.RequestConnection(QString server)
- emits a ConnectionRequested(QString server) signal

CChannel.Connect() is Client-only.  It would
- set connection state of ConnectionRequested (so CChannel.ConnectionState() returns that -- IsConnected() would be false)
- set up the connection to the server
  - fills the jitter buffers with silence and keeps them that way until the connection is established
  - open the socket and set up the signal handlers
  - kill the connection if a timer expires before the timer is stopped

(The Client is responsible for putting audio from CSound into the channel jitter buffer, which should trigger CChannel to take audio from the jitter buffer and send it to the Server, and for reading the jitter buffer and passing the audio to CSound.  I don't clearly remember the structure here.  There's a timer that expires if processing doesn't start.  There's a timer that expires for each audio cycle, too.  I think both are in CChannel as it's very similar in Client and Server.)

CChannel will handle audio being received by marking the CChannel connection state as ConnectionEstablished (IsConnected() is true).  Again, same at Client and Server for this.  It also emits a ConnectionEstablished signal.

----

> CClient will disconnect if at least one of these conditions holds

The first two should cause CClient.Disconnect() to be called and nothing else - lambda slot handlers in their respective classes calling the client method for signals.  It'll get called during CClient.OnAboutToQuit, I expect.

Actually, I'm proposing CClient.EndConnection() in place of CClient.Disconnect() to make it clear it's a non-trivial operation.  It should also _always_, without condition:
- call CChannel.Disconnect() (which may itself check whether it's in a ConnectionEstablished state)
  - (somewhere drains the out-going and in-coming audio packets in the jitter buffer - I guess it's here)
- call CSound.Stop() (which may itself check whether it's running)
  - (by this point it should have had one packet of silence to output)
- do any other Client-specific clean up (e.g. clear connected server name)
- emit a ConnectionEnded signal

----

> CClient is running iff

We need two things:
- CChannel.ConnectionState() should be ConnectionEstablished
- CSound.IsRunning() should be true

We could also check the wiring up of the signals from those two classes to the slots in the Client are in place so the sound goes to/from the channel... because CChannel and CSound do not know CClient is handling their signals.  However, that would be wired up once during creation of CClient, so can be taken as a given.

Nothing else matters because, so long as we have sound going in and out and packets going in and out - that means we set up okay and we're running fine.

> This means, there may be a sound card error while CClient is running and the client may not be connected to a server

Yes - but by CChannel changing state from ConnectionEstablished to Disconnected and emitting a ConnectionEnded signal or by CSound changing state from Running to Stopped and emitting an SoundStopped signal, where both are handled by CClient.  In either case, asking CClient whether it's "connected" or "running" or "started" or whatever, it'll say false.

In fact, CChannel::ConnectionEnded and CSound::SoundStopped would get handled the same way -- calling CClient::EndConnection.  (Or StopConnection or something... it needs a better name...)

----

> To transition from CClient being connected to just being running, we have to:

OK, I don't understand your concept of "just being running".  The Client just being there isn't the Client being in any particular state in and of itself.  It's just there, handling signals.  That doesn't need a state in and of itself.
- If CChannel.ConnectionState() should - unexpectedly - transition from ConnectionEstablished to anything else, we terminate the connection and stop sound -- because it means we lost the connection (timer expired).
- If CSound.IsRunning() should - unexpectedly - transition from true to anything else, we terminate the connection and stop sound -- because we got an audio error.

Neither of the above changes the state of the Client itself -- it would change the result of the method that combines the result of CChannel.ConnectionState() and CSound.IsRunning(), though.  But that's not a state, it's a computation.

If CChannel gets a server disconnect protocol message, it'll need to pass that to CClient to deal with - not simply disconnect, as that would be seen as an error happening.

Any _request_ to Disconnect would be handled the same way -- call CClient.EndConnection().

Exiting the client can happen multiple ways -- CLI through SIGINT/SIGTERM, GUI through menu emitting a signal, or a JSON-RPC command.  These should all trigger the Qt OnAboutToQuit by asking Qt to exit the application.  The Window Manager can also trigger Qt OnAboutToQuit by asking Qt to exit the application.  In all cases, when the client terminates, part of handling OnAboutToQuit in CClient would be to call CClient.EndConnection().

----

Or maybe I'm missing something?

src/client.cpp
softins: > Certainly possible, but I'd like to hear @softins opinion here too.

Yes, sorry. This PR is on my list of things to look at again, but just hadn't got to the top of the pile yet! I would certainly like to understand and comment.

ann0see:
Tagging this for 4.0 then...

ann0see:
Closing in favour of #3805 then

PR #3381, opened 2024-09-24
Disable memory swapping on Linux servers

dtinth:

Updated systemd unit file to disable swap on Jamulus process.

CHANGELOG: Server: Disabled swap memory usage on Linux.

**Context: Fixes an issue?**

I noticed that Jamulus server randomly (intermittently) become unresponsive for 0.3 – 1 seconds every once in a while. (This issue has happened for a long time prior.) Last month I installed [Netdata](https://www.netdata.cloud/) on my servers to collect detailed metrics. It seems that Linux has been paging some memory to/from disk every once in a while.

So systemd has the ability to prevent an app from using swap space. This might help with this issue. From my preliminary testing, it seems to have helped.

As memory footprint of Jamulus server is very minimal (19.7M used after 8 months of uptime) I think this config change should not have any negative impact unless the server is run on a potato.

No.

Ready

N/A

## Checklist

### DIFF
--- a/linux/debian/jamulus-headless.service
+++ b/linux/debian/jamulus-headless.service
@@ -13,6 +13,7 @@ ProtectHome=true
 Nice=-20
 IOSchedulingClass=realtime
 IOSchedulingPriority=0
+MemorySwapMax=0
 
 #### Change this to publish this server, set genre, location and other parameters.
 #### See https://jamulus.io/wiki/Running-a-Server#server-mode-related-options ####


ann0see:
Thanks. But why would the OS swap in the first place? The app is active and ram should be available for such a small footprint

dtinth:
Yeah, the server runs other stuff too (such as a system that broadcasts MP3 streams, some web servers), so that’s a possible cause. I did `cat /proc/<pid>/smaps` before the change and some memory is indeed in the swap area.

The system I was running happens to have `vm.swappiness = 60` (default value), so I think there is always a chance that the system moves the memory to swap area despite Jamulus having a low memory footprint and the system having plenty of free RAM.

softins:
I've added this change to my own servers, but I can't actually test properly until I'm back home after my holiday.

pljones:
Added to my servers and directories.  I've a few other bits:
```
[Service]
CPUSchedulingPolicy=fifo
CPUSchedulingPriority=20
IOSchedulingClass=realtime
IOSchedulingPriority=3
MemorySwapMax=0
```
No idea if the differences help.

pljones (approved):

ann0see:
@softins do you think this is ok?

softins:
> @softins do you think this is ok?

I added it to my server on AWS some time ago. It didn't make anything worse, but it was hard to tell if it made any improvement.

I've since moved my server to Vultr, and forgot to transfer that setting. I've just added it back and restarted the jamulus services, so we'll see. I've just discovered that `top` has a `SWAP` column that can be added to the display, to show the swap usage of each process.

I'm on Debian 12, which has systemd 252. I've read that this setting is only available in systemd 239 and newer, but I don't know what versions older distros use.

I've no objection to merging this.

softins (approved):

ann0see:
Ok. Let's merge this. I don't think there are any disadvantages unless it's run on low RAM machines.

PR #3394, opened 2024-10-02
Add ctrlmidich option for "own channel" regardless of number

pljones:

From @AndersGoran.

This adds a `;z` option to `--ctrlmidich` parsing that adds control of the client channel, regardless of its channel number.  This applies to all the channel-specific controls used.  It takes up one extra MIDI controller number for each.

CHANGELOG: Add ctrlmidich option for "own channel" regardless of number

**Context: Fixes an issue?**

Raised in https://github.com/orgs/jamulussoftware/discussions/2220#discussioncomment-10811813

Yes, but it needs reviewing to see if this is how we want to do it first.

Proof of concept.

Needs reviewing and considering whether "own client id" concept should apply to all controls (e.g. use `;z` as a switch to bump lock the first CC number for each control to "own client id").

## Checklist

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1351,6 +1351,16 @@ void CAudioMixerBoard::SetFaderLevel ( const int iChannelIdx, const int iValue )
             vecpChanFader[static_cast<size_t> ( iChannelIdx )]->SetFaderLevel ( iValue );
         }
     }
+    // Proposed change: if iChannelIdx is -1 and our own channel ID is a valid index
+    // then we adjust our own fader level:
+    if((iChannelIdx == -1) && iMyChannelID != INVALID_INDEX)
+    {
+        if ( vecpChanFader[static_cast<size_t> ( iMyChannelID )]->IsVisible() )
+        {
+            //printf("debug: set our own fader(%d) level to %d\n", iMyChannelID, iValue);
+            vecpChanFader[static_cast<size_t> ( iMyChannelID )]->SetFaderLevel ( iValue );
+        }
+    }
 }
 
 void CAudioMixerBoard::SetPanValue ( const int iChannelIdx, const int iValue )
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1343,22 +1343,16 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
 
 void CAudioMixerBoard::SetFaderLevel ( const int iChannelIdx, const int iValue )
 {
+    // If iChannelIdx is I_MY_CHANNEL and our own channel ID is a valid index
+    // then we adjust our own fader level
+    const int iTheChannelIdx = ( iChannelIdx == I_MY_CHANNEL ) ? iMyChannelID : iChannelIdx;
+
     // only apply new fader level if channel index is valid and the fader is visible
-    if ( ( iChannelIdx >= 0 ) && ( iChannelIdx < MAX_NUM_CHANNELS ) )
+    if ( ( iTheChannelIdx >= 0 ) && ( iTheChannelIdx < MAX_NUM_CHANNELS ) )
     {
-        if ( vecpChanFader[static_cast<size_t> ( iChannelIdx )]->IsVisible() )
+        if ( vecpChanFader[static_cast<size_t> ( iTheChannelIdx )]->IsVisible() )
         {
-            vecpChanFader[static_cast<size_t> ( iChannelIdx )]->SetFaderLevel ( iValue );
-        }
-    }
-    // Proposed change: if iChannelIdx is -1 and our own channel ID is a valid index
-    // then we adjust our own fader level:
-    if((iChannelIdx == -1) && iMyChannelID != INVALID_INDEX)
-    {
-        if ( vecpChanFader[static_cast<size_t> ( iMyChannelID )]->IsVisible() )
-        {
-            //printf("debug: set our own fader(%d) level to %d\n", iMyChannelID, iValue);
-            vecpChanFader[static_cast<size_t> ( iMyChannelID )]->SetFaderLevel ( iValue );
+            vecpChanFader[static_cast<size_t> ( iTheChannelIdx )]->SetFaderLevel ( iValue );
         }
     }
 }
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1343,53 +1343,56 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
 
 void CAudioMixerBoard::SetFaderLevel ( const int iChannelIdx, const int iValue )
 {
-    // If iChannelIdx is I_MY_CHANNEL and our own channel ID is a valid index
-    // then we adjust our own fader level
-    const int iTheChannelIdx = ( iChannelIdx == I_MY_CHANNEL ) ? iMyChannelID : iChannelIdx;
+    const int iRealChannelIdx = getRealChannelIdx ( iChannelIdx );
 
     // only apply new fader level if channel index is valid and the fader is visible
-    if ( ( iTheChannelIdx >= 0 ) && ( iTheChannelIdx < MAX_NUM_CHANNELS ) )
+    if ( ( iRealChannelIdx >= 0 ) && ( iRealChannelIdx < MAX_NUM_CHANNELS ) )
     {
-        if ( vecpChanFader[static_cast<size_t> ( iTheChannelIdx )]->IsVisible() )
+        if ( vecpChanFader[static_cast<size_t> ( iRealChannelIdx )]->IsVisible() )
         {
-            vecpChanFader[static_cast<size_t> ( iTheChannelIdx )]->SetFaderLevel ( iValue );
+            vecpChanFader[static_cast<size_t> ( iRealChannelIdx )]->SetFaderLevel ( iValue );
         }
     }
 }
 
 void CAudioMixerBoard::SetPanValue ( const int iChannelIdx, const int iValue )
 {
+    const int iRealChannelIdx = getRealChannelIdx ( iChannelIdx );
+
     // only apply new pan value if channel index is valid and the panner is visible
-    if ( ( iChannelIdx >= 0 ) && ( iChannelIdx < MAX_NUM_CHANNELS ) && bDisplayPans )
+    if ( ( iRealChannelIdx >= 0 ) && ( iRealChannelIdx < MAX_NUM_CHANNELS ) && bDisplayPans )
     {
-        if ( vecpChanFader[static_cast<size_t> ( iChannelIdx )]->IsVisible() )
+        if ( vecpChanFader[static_cast<size_t> ( iRealChannelIdx )]->IsVisible() )
         {
-            vecpChanFader[static_cast<size_t> ( iChannelIdx )]->SetPanValue ( iValue );
+            vecpChanFader[static_cast<size_t> ( iRealChannelIdx )]->SetPanValue ( iValue );
         }
     }
 }
 
 void CAudioMixerBoard::SetFaderIsSolo ( const int iChannelIdx, const bool bIsSolo )
 {
-    // only apply solo if channel index is valid and the fader is visible
-    if ( ( iChannelIdx >= 0 ) && ( iChannelIdx < MAX_NUM_CHANNELS ) )
+    const int iRealChannelIdx = getRealChannelIdx ( iChannelIdx );
 
+    // only apply solo if channel index is valid and the fader is visible
+    if ( ( iRealChannelIdx >= 0 ) && ( iRealChannelIdx < MAX_NUM_CHANNELS ) )
     {
-        if ( vecpChanFader[static_cast<size_t> ( iChannelIdx )]->IsVisible() )
+        if ( vecpChanFader[static_cast<size_t> ( iRealChannelIdx )]->IsVisible() )
         {
-            vecpChanFader[static_cast<size_t> ( iChannelIdx )]->SetFaderIsSolo ( bIsSolo );
+            vecpChanFader[static_cast<size_t> ( iRealChannelIdx )]->SetFaderIsSolo ( bIsSolo );
         }
     }
 }
 
 void CAudioMixerBoard::SetFaderIsMute ( const int iChannelIdx, const bool bIsMute )
 {
+    const int iRealChannelIdx = getRealChannelIdx ( iChannelIdx );
+
     // only apply mute if channel index is valid and the fader is visible
-    if ( ( iChannelIdx >= 0 ) && ( iChannelIdx < MAX_NUM_CHANNELS ) )
+    if ( ( iRealChannelIdx >= 0 ) && ( iRealChannelIdx < MAX_NUM_CHANNELS ) )
     {
-        if ( vecpChanFader[static_cast<size_t> ( iChannelIdx )]->IsVisible() )
+        if ( vecpChanFader[static_cast<size_t> ( iRealChannelIdx )]->IsVisible() )
         {
-            vecpChanFader[static_cast<size_t> ( iChannelIdx )]->SetFaderIsMute ( bIsMute );
+            vecpChanFader[static_cast<size_t> ( iRealChannelIdx )]->SetFaderIsMute ( bIsMute );
         }
     }
 }
--- a/src/audiomixerboard.h
+++ b/src/audiomixerboard.h
@@ -288,6 +288,15 @@ protected:
     template<unsigned int slotId>
     inline void connectFaderSignalsToMixerBoardSlots();
 
+    // When handling MIDI controllers for adjusting Jamulus channel controls,
+    // each channel is assigned by the server.  As a user's Jamulus channel
+    // will vary on each server connection, to enable the assignment of MIDI
+    // controllers to the user's own Jamulus channel, the constant I_MY_CHANNEL
+    // is passed into the methods used to handle the requests.  This method then
+    // maps I_MY_CHANNEL to the current value of the user's Jamulus channel,
+    // held in iMyChannel.
+    inline int getRealChannelIdx ( const int iChannelIdx ) const { return iChannelIdx == I_MY_CHANNEL ? iMyChannelID : iChannelIdx; }
+
 signals:
     void ChangeChanGain ( int iId, float fGain, bool bIsMyOwnFader );
     void ChangeChanPan ( int iId, float fPan );
--- a/src/sound/soundbase.cpp
+++ b/src/sound/soundbase.cpp
@@ -33,6 +33,7 @@ char const sMidiCtlChar[] = {
     /* [EMidiCtlType::Solo]        = */ 's',
     /* [EMidiCtlType::Mute]        = */ 'm',
     /* [EMidiCtlType::MuteMyself]  = */ 'o',
+    /* [EMidiCtlType::OurFader]    = */ 'z', // Proposed addition: a new enum value for "our fader"
     /* [EMidiCtlType::None]        = */ '\0' };
 
 /* Implementation *************************************************************/
@@ -370,6 +371,13 @@ void CSoundBase::ParseMIDIMessage ( const CVector<uint8_t>& vMIDIPaketBytes )
                             // consider offset for the faders
 
                             emit ControllerInFaderLevel ( cCtrl.iChannel, iFaderLevel );
+                        }
+                        break;
+                        case OurFader:
+                        {
+                            // special message about our own fader - emit a fader level whatever-it-is-called with channel id -1
+                            const int iFaderLevel = static_cast<int> ( static_cast<double> ( iValue ) / 127*AUD_MIX_FADER_MAX );
+                            emit ControllerInFaderLevel ( -1, iFaderLevel);
                         }
                         break;
                         case Pan:
--- a/src/sound/soundbase.cpp
+++ b/src/sound/soundbase.cpp
@@ -24,18 +24,6 @@
 
 #include "soundbase.h"
 
-// This is used as a lookup table for parsing option letters, mapping
-// a single character to an EMidiCtlType
-char const sMidiCtlChar[] = {
-    // Has to follow order of EMidiCtlType
-    /* [EMidiCtlType::Fader]       = */ 'f',
-    /* [EMidiCtlType::Pan]         = */ 'p',
-    /* [EMidiCtlType::Solo]        = */ 's',
-    /* [EMidiCtlType::Mute]        = */ 'm',
-    /* [EMidiCtlType::MuteMyself]  = */ 'o',
-    /* [EMidiCtlType::OurFader]    = */ 'z', // Proposed addition: a new enum value for "our fader"
-    /* [EMidiCtlType::None]        = */ '\0' };
-
 /* Implementation *************************************************************/
 CSoundBase::CSoundBase ( const QString& strNewSystemDriverTechniqueName,
                          void ( *fpNewProcessCallback ) ( CVector<int16_t>& psData, void* pParg ),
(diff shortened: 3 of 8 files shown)


pljones:
There may currently be a problem with `CSoundBase::ParseCommandLineArgument` in that it doesn't necessarily set every element of `aMidiCtls` for each `sMidiCtlChar` but it doesn't check in `CSoundBase::ParseMIDIMessage` that the element exists.

So I'd like the `ParseCommandLineArgument` changed to ensure each possible option gets set to some "unset" default value and then overridden by the command line argument.

Separately, I'd like `;z<n>` changed to just `;z` and have that toggle "Use Own Channel Id In First CC For Control" mode -- so it's not just Fader that gets the benefit, it's Pand and Mute, too.  Or have `;z<n>` with `<n>` being the offset into the usual range of CC numbers for each control.

src/sound/soundbase.h
```diff
@@ -49,6 +49,7 @@ enum EMidiCtlType
     Solo,
     Mute,
     MuteMyself,
+    OurFader, // Proposed addition: a MidiCtrlType for our own fader level
```
ann0see, line 52:
```suggestion
    MyFader, // Proposed addition: a MidiCtrlType for our own fader level
```

I think that's more in line with the code otherwise.

src/audiomixerboard.cpp
```diff
@@ -1335,12 +1335,16 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
 
 void CAudioMixerBoard::SetFaderLevel ( const int iChannelIdx, const int iValue )
 {
+    // Proposed change: if iChannelIdx is -1 and our own channel ID is a valid index
+    // then we adjust our own fader level:
+    const int iTheChannelIdx = ( iChannelIdx == INVALID_INDEX ) ? iMyChannelID : iChannelIdx;
```
ann0see, line 1340: I don't like this implementation. Why abuse invalid index?

AndersGoran: Not sure if you need my involvement at this stage but my patch used `-1` explicitly at both the `emit` statement and here at `SetFaderLevel`; at first I didn't even know that is was the same value as INVALID_INDEX. I can definitely see your point.

pljones: Yes.  And `INVALID_INDEX` is used to mean "the appropriate channel" in other places, too, already.  The only problem with `-1` for me is that it wasn't clear what it meant, whereas `INVALID_INDEX` in both places means "magic out of range value" quite clearly.

AndersGoran: Well, if one knows about this magic double meaning of "invalid", fine. I'm happy to leave it to you guys to decide how it's best done, I'm just happy you are considering adding this feature at all. Much appreciated.

pljones:
OK, so I decided to make the change I suggested.

```
The legacy definition has just one or two numbers that only provides a definition for the controller offset of the fader controllers (default 70 for the sake of Behringer X-Touch)
   [MIDI channel];[offset for first fader]

The more verbose new form is a sequence of offsets and counts for various controllers:
   [MIDI channel];[control letter][offset]*[count](;...)

Currently, the following control letters are defined:
   Fader       - 'f'
   Pan         - 'p'
   Solo        - 's'
   Mute        - 'm'
'offset' is the base MIDI CC number for the control.
'count'  is the number of CC values for the control (Jamulus channels).

Additionally, 'o' has a single offset and controls Mute Myself.

In addition, 'z' reserves the first CC number for a control
to mean "my Jamulus channel".  (For example
   1;f0*9;z
would mean fader CC0 controlled "my" channel, with CC1 to CC8 for normal channels 1 to 8 (even if you were channel 6 - so you get two faders).

An example for a Korg nanoKONTROL2 with 8 fader controllers starting at offset 0 and 8 pan controllers starting at offset 16
would be
   [MIDI channel];f0*8;p16*8

However, at the current point of time only 'f' and 'p' controllers are actually implemented. // may not be true...
```

src/sound/soundbase.h
```diff
@@ -158,7 +159,20 @@ class CSoundBase : public QThread
...
+
+    // This is used as a lookup table for parsing option letters, mapping
+    // a single character to an EMidiCtlType
+    static char constexpr sMidiCtlChar[] = { // Has to follow order of EMidiCtlType
+                                             /* [EMidiCtlType::Fader]       = */ 'f',
+                                             /* [EMidiCtlType::Pan]         = */ 'p',
+                                             /* [EMidiCtlType::Solo]        = */ 's',
+                                             /* [EMidiCtlType::Mute]        = */ 'm',
+                                             /* [EMidiCtlType::MuteMyself]  = */ 'o',
+                                             /* [EMidiCtlType::MyFader]     = */ 'z',
```
pljones, line 172: Fixed comment.

src/sound/soundbase.cpp
```diff
@@ -232,98 +221,176 @@ QVector<QString> CSoundBase::LoadAndInitializeFirstValidDriver ( const bool bOpe
...
-    // the legacy definition with just one or two numbers that only
-    // provides a definition for the controller offset of the level
+    // default to the legacy kind of specifying
+    // the fader controller offset without an
+    // indication of the count of controllers
+    bool bSimple = true;
+
+    // Parse the --ctrlmidich string.  There are two formats.
+    //
+    // The legacy definition has just one or two numbers that only
```
pljones, line 240: I vote this comment is removed.  It doesn't explain the code, only provides a "user guide" or "how-to" for someone using the feature.  So it's better in user-facing documentation.

pljones:
Ugh, failure due to qt5 -> qt6.  Investigating.

pljones:
OK, I can't get JACK to run without crashing on Windows any more for some reason.  So I can't actually test this (as Jamulus ASIO doesn't support MIDI, as far as I know, and I've no MIDI - or audio - on my Linux box).

src/sound/soundbase.cpp
pljones: Done.

https://github.com/jamulussoftware/jamuluswebsite/pull/1053 raised to update the documentation.

pljones:
@AndersGoran could you check whether the latest build is useable?

It should work like this:

`--ctrlmidich 0;z;f0*9` gives you MIDI CC 0 controlling "my" fader, regardless of server-assigned Jamulus channel number, MIDI CC 1 controlling Jamulus channel 0 through to MIDI CC 8 controlling Jamulus channel 7.

It should work for _all_ the Jamulus controls.

`--ctrlmidich 0;z;f0;p1;s2;m3;o4` should give you control only over your own channel for fader, pan, mute (um, no, don't) and mute myself (better).  `--ctrlmidich 0;z;f0*9;p1*9;s2*9;m3*9;o4` gives you your own channel, plus additional controls for up to eight Jamulus channels, which could include your own channel.  (If you were on Jamulus channel 3, you'd get MIDI CC 0 and MIDI CC 4 controlling your fader, for example; if you were on Jamulus channel 17, you'd still get control over your own channel and the first eight Jamulus channels.)

**_Note_**: remember, having control over your own channel only affects what you hear, not what others hear.

AndersGoran:
I cloned https://github.com/pljones/jamulus.git, current HEAD is 3f0983e2, compiles fine (I'm on macOS) but I'm afraid the "z" flag makes no difference. As before, my controller sends CC 100-103 from four separate pedals, so I use "0;z;f100*4" and the 100 pedal controls channel 0, 101 control 1, and so on. When I'm on e.g. 1, I have to use the pedal that sends 101 to move my fader. 

Is there a pre-built executable somewhere that I should be using?

ann0see:
Yes. Checkout the Checks tab of this PR. There you'll find the macOS artifact for this PR.

AndersGoran:
So I suppose it is the DMG linked here that I want: https://github.com/jamulussoftware/jamulus/actions/runs/11194714268/artifacts/2019359055 

I'm afraid this one does nothing at all related to MIDI. Not with the usual string without "z", nor with "z". 

Here's how I invoke it: 

```
[~/Desktop]$ exec /Applications/Jamulus.app/Contents/MacOS/Jamulus --ctrlmidich "0;z;f100*4" "$@"
- MIDI controller settings: 0;z;f100*4
- allocated port number: 22134
QString::arg: Argument missing: fpsmoz,
```

That last `QString::arg` looks a bit off - `fpsmoz` - aren't those exactly the char's we allow in the midi ctrl argument?

pljones:
Okay, thanks, let me take a look.

pljones:
OK, new build there.

(I can't get Jamulus for Linux to see Pipewire -- PW is running, `qpwgraph` and both `pavucontrol` are fine but Jamulus runs Jack and doesn't appear as a PW app.)

AndersGoran:
Sorry, same with this one. The "fpsmoz" thing is gone, though.
No MIDI action at all. Official 3.11 client is fine (but of course not with "z" flag), but this one does nothing with or without "z". Installed from https://github.com/jamulussoftware/jamulus/actions/runs/11202345633/artifacts/2020757377

pljones:
Thanks.  Hm.  Clearly I'm not understanding the code...  I'll back out my changes for now.

pljones:
OK, once that's built, can you check it works?

pljones:
Extra debugging logging in my version:
```
$ ./Jamulus --ctrlmidich "0;z;p1;f2;m3"
- MIDI controller settings: 0;z;p1;f2;m3
- allocated port number: 22134
"sParm: 1*1*1"
"eTyp: 1; iFirst: 1; iNum: 1; bMyChannel: 1"
"aMidiCtls[1 + 0] = { (eType) 1, (iChannel) -1 }"
"sParm: 0*2*1"
"eTyp: 0; iFirst: 0; iNum: 2; bMyChannel: 1"
"aMidiCtls[0 + 0] = { (eType) 0, (iChannel) -1 }"
"aMidiCtls[0 + 1] = { (eType) 0, (iChannel) 0 }"
"sParm: 3*3*1"
"eTyp: 3; iFirst: 3; iNum: 3; bMyChannel: 1"
"aMidiCtls[3 + 0] = { (eType) 3, (iChannel) -1 }"
"aMidiCtls[3 + 1] = { (eType) 3, (iChannel) 0 }"
"aMidiCtls[3 + 2] = { (eType) 3, (iChannel) 1 }"
```
Well, that's wrong, isn't it!  It's getting the type, offset and count right in `sParm` and then my loop populates `aMidiCtls` wrongly.
```
"aMidiCtls[0 + 1] = { (eType) 0, (iChannel) 0 }"
```
comes along and overwrites
```
"aMidiCtls[1 + 0] = { (eType) 1, (iChannel) -1 }"
```
for example.  At least now I can see something is wrong...

pljones:
OK:

| | |
|---|---|
| Branch (includes this PR) | https://github.com/jamulussoftware/jamulus/compare/main...pljones:jamulus:ctrlmidich-for-own-channel-all-controls |
| Just the extra diff | https://github.com/pljones/jamulus/commit/3ba65c611c1692c61ef3b6b4ab4fb77cb9e50c4a |
| Builds | https://github.com/pljones/jamulus/actions/runs/11221638882 |

AndersGoran:
Now we're talking! I've tested briefly on a server where "0:" is someone else, and I am "1:".

If I start this latest build (macOS non-legacy is what I use) with `0;f100*4` - without `z` - then everything is as it used to be, my first pedal CC100 controls "someone else" and CC101 controls my fader. As expected. 

When I add `z`, i.e. `0;f100*4;z`, then the first pedal, CC100, controls my fader, CC101 controls "0:", i.e. the first non-me channel. But I also notice that CC102 controls my fader too. Maybe that's expected, maybe not. 

But at least, now the z flag is recognised and it does what I believe we want it to do. Great!

AndersGoran:
I think we should test this a bit more on sessions with many participants. Now I connected to a server where I got channel 0 -  there had been lots of people on it earlier but many had disconnected, so the few ones left were 6, 7, 8 or something like that. As I have 4 CC controllers configured, all I could control was my own fader because I had the "z" flag. 

But that's not really relevant for this PR, is it? 

I can certainly do some more digging, but there's only so many hours in a day... it seems I should be able to clone the branch mentioned above and do some testing, right? 

Anyway, thanks for the effort of getting it to where it is now!

pljones:
Yes, you'll have your own channel number.  Start out setting something like `0;f100*2;z`, that should give you MIDI Ctrl 100 for your channel, whatever number it is and 101 for Jamulus channel 0.  If you're not channel 0, you'll still be able to control your fader.  If you are channel 0, you'll have 100 as the "sticky" number and 101 just because you're on Jamulus channel 0 and that's what 101 controls.

Are you trying the build here or from the link I posted, by the way?

AndersGoran:
I was trying the macOS non-legacy build from https://github.com/pljones/jamulus/actions/runs/11221638882

I think the main purpose of the PR is fulfilled, that one can be sure to have control over one's own fader.

src/audiomixerboard.cpp
ann0see: I'd rather use something else. Abusing invalid index doesn't feel right at all.

src/audiomixerboard.cpp
```diff
@@ -1335,12 +1335,16 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
...
+
     // only apply new fader level if channel index is valid and the fader is visible
-    if ( ( iChannelIdx >= 0 ) && ( iChannelIdx < MAX_NUM_CHANNELS ) )
+    if ( ( iTheChannelIdx >= 0 ) && ( iTheChannelIdx < MAX_NUM_CHANNELS ) )
```
ann0see, line 1343: TheChannelIdx is also not expressive.

src/audiomixerboard.cpp
pljones: We'll need to go through the whole application finding where it's been used like this, then.

Alternatively, I could bump all the numbers up, so they're "1-based" and reserve 0 for "my channel", then subtract one...

src/audiomixerboard.cpp
pljones: It's the channel index that will be used.  `iChannelIdx` is already taken.  I could use `i`?  Any other suggestions?

src/audiomixerboard.cpp
ann0see: Semantically speaking, it's just not INVALID_INDEX in my opinion.

Can we introduce another DEFINE saying something better than INVALID_INDEX? it could still be -1 internally.

src/audiomixerboard.cpp
ann0see: iChannelIdxToModify?

src/audiomixerboard.cpp
pljones: OK, changed to `I_MY_CHANNEL`.

src/sound/soundbase.h
```diff
@@ -32,6 +32,7 @@
 #endif
 #include "../global.h"
 #include "../util.h"
+#define I_MY_CHANNEL -1
```
ann0see, line 35: Yes. That's better.

src/audiomixerboard.cpp
```diff
@@ -1335,12 +1335,16 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
 
 void CAudioMixerBoard::SetFaderLevel ( const int iChannelIdx, const int iValue )
 {
+    // Proposed change: if iChannelIdx is I_MY_CHANNEL and our own channel ID is a valid index
```
ann0see, line 1338:
```suggestion
    // If iChannelIdx is I_MY_CHANNEL and our own channel ID is a valid index
```

src/sound/soundbase.cpp
```diff
@@ -232,98 +221,140 @@ QVector<QString> CSoundBase::LoadAndInitializeFirstValidDriver ( const bool bOpe
...
+* Command Line Handling                                                        *
 \******************************************************************************/
 void CSoundBase::ParseCommandLineArgument ( const QString& strMIDISetup )
 {
+    if ( strMIDISetup.isEmpty() )
+        return;
+
     int iMIDIOffsetFader = 70; // Behringer X-TOUCH: offset of 0x46
 
-    // parse the server info string according to definition: there is
```
ann0see, line 241: Where did this go?

ann0see (changes requested):

src/sound/soundbase.cpp
pljones: (new) Line 233.

pljones: (`parse the server info string` was obviously copied from the parsing of the server info string - nothing to do with parsing the `ctrlmidich` string.)

pljones:
> Coding Style Check / Verify Python coding style (pull_request)

Why did that fail?

src/audiomixerboard.cpp
pljones: `iRealChannelIdx` chosen.

src/audiomixerboard.cpp
pljones: Moved into header as a comment on the new function to return the real channel index.

src/sound/soundbase.cpp
pljones: Oh, and [more documentation](https://github.com/jamulussoftware/jamuluswebsite/pull/1053) is being added.

src/audiomixerboard.h
```diff
@@ -287,6 +287,9 @@ class CAudioMixerBoard : public QGroupBox, public CAudioMixerBoardSlots<MAX_NUM_
     template<unsigned int slotId>
     inline void connectFaderSignalsToMixerBoardSlots();
 
+    // If iChannelIdx is I_MY_CHANNEL it means iMyChannelID should be used instead
```
ann0see, line 290: Clarify _why_ I_MY_Channel must be used instead.

pljones: It doesn't _have to be_ used.  _By definition_, if you use `I_MY_CHANNEL` for `iChannel`, it means you want `iMyChannel` used instead.  That's just what it _means_.  I'm not sure what else I could say.

AndersGoran (approved):
I was requested to review, and while I'm not sure if you really want or need my approval regarding code details, I just tried the latest macOS build artifact `jamulus_3.11.0dev-9cb1e39_mac.dmg` and it still feels solid regarding the "z" flag.

src/audiomixerboard.h
ann0see: I meant more of an explanation why this function exists. So say that this constant is used for the new flag introduced here.

src/sound/soundbase.cpp
```diff
@@ -232,98 +221,142 @@ QVector<QString> CSoundBase::LoadAndInitializeFirstValidDriver ( const bool bOpe
...
-    // nanoKONTROL2 with 8 fader controllers starting at offset 0 and
-    // 8 pan controllers starting at offset 16 would be
-    //
-    // [MIDI channel];f0*8;p16*8
-    //
-    // Namely a sequence of letters indicating the kind of controller,
-    // followed by the offset of the first such controller, followed
-    // by * and a count for number of controllers (if more than 1)
-    if ( !strMIDISetup.isEmpty() )
+    if ( strMIDISetup.isEmpty() )
```
ann0see, line 228: I'd prefer curly braces...
The code in general doesn't - I know

src/sound/soundbase.cpp
```diff
@@ -232,98 +221,142 @@ QVector<QString> CSoundBase::LoadAndInitializeFirstValidDriver ( const bool bOpe
...
 
-        bool bSimple = true; // Indicates the legacy kind of specifying
-                             // the fader controller offset without an
-                             // indication of the count of controllers
+    bool        bMyChannel = false;
+    QStringList slValid; // keep track of valid entries to make later processing simple
 
-        // [offset for level]
-        if ( slMIDIParams.count() >= 2 )
+    for ( int i = 0; i < slMIDIParams.count(); i++ )
```
ann0see, line 288: Do we need to call count() on every iteration?

ann0see (changes requested):

src/sound/soundbase.cpp
pljones: Existing code but I agree.  However, the style guide says it's accepted.

src/audiomixerboard.h
pljones: I've added a long explanation.

src/audiomixerboard.h
```diff
@@ -287,6 +287,15 @@ class CAudioMixerBoard : public QGroupBox, public CAudioMixerBoardSlots<MAX_NUM_
     template<unsigned int slotId>
     inline void connectFaderSignalsToMixerBoardSlots();
 
+    // When handling MIDI controllers for adjusting Jamulus channel controls,
```
ann0see, line 290: Yes. Much better. Thanks

src/sound/soundbase.cpp
```diff
@@ -232,98 +221,156 @@ QVector<QString> CSoundBase::LoadAndInitializeFirstValidDriver ( const bool bOpe
...
-        bool bSimple = true; // Indicates the legacy kind of specifying
-                             // the fader controller offset without an
-                             // indication of the count of controllers
+    // Use Behringer X-TOUCH as default offset of 0x46
+    int iMIDIOffsetFader = 70;
+    if ( iNumParams >= 2 )
+    {
+        // if there is a second parameter that can be parsed as a number,
+        // we have the legacy specification of controllers.
+        int i = slMIDIParams[1].toUInt ( &bSimple );
```
ann0see, line 265: Can toUInt fail? What happens then?

ann0see: Also UInt != int?

pljones: `toUInt` sets the boolean argument passed by reference to true or false, depending on success.

`i` should be a number from 0 to 127.  Of course, someone _could_ type a really stupid value in that would give a really weird result...  The `toUInt` is to prevent parsing of a valid negative value, though, as that could cause issues.

pljones:
> I have the feeling that documentation is lacking on explaining the code. I feel unsure about the exact functionality - especially the cli parsing.

Which documentation?

What are you unsure about?

ann0see:
I think I need to go over the code in my editor. The diff on GitHub doesn't show me enough.

softins:
I know I'm rather late to the party here, due to limited time, but have today been studying and thinking about this solution to @AndersGoran's original issue.

I don't like the fact that when using the `z` option, there will be _two_ faders that control the user's own channel: the first fader (by virtue of the `z` option), and the fader assigned to the user's actual channel. This seems user-unfriendly, wasteful of a physical fader, and with the potential for confusion, if the two faders, pans or mute/solo buttons are set to different states.

I would like to propose a different solution (apologies for not doing so until now!):

* The user's own channel _always_ gets given the internal fader channel 0, whatever client ID it is on. So my fader would always look like `0: TonyM`.
* Any other channels with a client ID _less_ than my own client ID will be given the internal fader channel of ID+1.
* Any other channels with a client ID _greater_ than my own client ID will be given the internal fader channel equal to their ID as at present.

So a session where I am **TonyM** might look like this:

| Client ID on server | Name | Fader ID | Tag on mixer |
|--:|--|--:|--|
| 0 | Tom | 1 | 1:Tom |
| 1 | Dick | 2 | 2:Dick |
| 2 | **TonyM** | 0 | **0:TonyM** |
| 3 | Harry | 3 | 3:Harry |
| 4 | Jim | 4 | 4:Jim |

That way, there is only one fader channel per client, and my own fader is always the first one. These fader IDs are local to my own Jamulus client, and don't need to be the same as those on any other user's client.

Since the user never has any control over which client IDs are assigned by the server, it doesn't matter that the other users fader number doesn't necessarily equal their client ID, and the above technique would ensure the first fader is always the user's own, since they always get fader channel 0. This wouldn't even need the extra `z` flag in the `--ctrlmidich` string; there would be no downside in making the above _always_ the case.

This would also work well with @pljones recent "Sort by Channel" option in #3418 (where Channel is the fader ID).

Comments?

pljones:
It's probably best if @AndersGoran comments.

ann0see:
@softins agree. 

> The user's own channel always gets given the internal fader channel 0, whatever client ID it is on. So my fader would always look like 0:TonyM.

This could potentially also simply the own fader first implementation

AndersGoran:
Thank you @softins, for looking and thinking about this. Your suggestion sounds good to me. 

My use case has always been such that I want a predictable way to assign one single MIDI CC to always control _my_ fader. I don't really care about any other faders than my own, in any of the groups I play in on Jamulus. 

I just want to be able to mute myself in my own mix with MIDI CC. If I could trust that the first CC given in `--ctrlmidich` always controls my fader, that's all I need.

softins:
I'm in the middle of doing a proof of concept for my suggestion, which we will be able to try out in a few days. It's quite complicated working out the best way, due to the onion-like many layers in the software structure: Client; ClientDlg; MainMixerBoard; Faders; with many signal/slot connections between them.

One thing we can't do is overwrite the iChanID, since that by definition must match the channel ID on the server. So the MIDI channel numbers must be stored or calculated independently on the client side, and the mapping between them and channel IDs done either algorithmically or via lookup tables. I'm currently trying the former approach as per my suggestion further up.

ann0see:
Lookup table sounds more performant and less messy.

pljones:
> Lookup table sounds more performant and less messy.

Yup - the current version uses a lookup table.

pljones:
OK, got to test this on Windows (`develop` branch build of `jackd` needed to fix MIDI on Windows).  The current build is doing what I'd intended.

pljones:
Set to draft as this will get reworked / dropped after https://github.com/jamulussoftware/jamulus/pull/3426 merges.

pljones:
OK, closing this in favour of https://github.com/jamulussoftware/jamulus/pull/3426

Issue #3406, opened 2024-10-10
Make Jamulus iOS compatible with Qt6


ann0see:
**Describe the bug**

Even though Jamulus can be compiled by Qt6 and it runs, there are multiple UI issues. 
Also the app hangs and crashes in multiple places. 

1. The chat window doesn't have the correct proportions. Especially if turning the device around the chat window overflows the screen (screenshot will follow)
2. Tapping the chat window Close button hangs and then crashes the app on the next tap. 
3. Changing the language in settings doesn't apply and the app hangs. 
4. Further bugs to be investigated

Related: https://github.com/jamulussoftware/jamulus/issues/2711

**Operating system**

iOS 17

**Version of Jamulus**

Latest version from main

**Additional context**

Try compiling Jamulus with Qt6 for iOS.

ann0see:
Pasting https://forum.qt.io/topic/15326/solved-qlayout-items-not-expanding-on-resize/4 here. Maybe related to the crash of the languages change.

ann0see:
@softins do you own an iPad?
I'd like to get the app on TestFlight to find further bugs.

softins:
Yes, and have Testflight on it. iOS 16.7, which I think is the latest it will take (2016 model).

ann0see:
Thanks. I'll try to work out how to get the app on TestFlight in the following days. It would be great if you could confirm if the mentioned issues all also occur on iPadOS

ann0see:
Hotfix for 2 is probably not using the close() signal but hide() instead.

ann0see:
Hotfix for 3 is conditionally getting rid of the warning message on iOS.

ann0see:
Ok. Also the Qt5 build hangs with the language change. So that's not a new bug.
Qt5 doesn't even open the about dialog.
However, there's no need to turn the screen to get a working, non overflowing screen on Qt5.

ann0see:
Changing the skin to compact, Bar (narrow) makes the app somewhat usable.

ann0see:
Ok. The mixer board issue comes from the "Setup your audio,... message". I'll change it to be brief on iOS.

ann0see:
We could force the GD_SLIMFADER design on iOS...
Hiding the text is not so easy. The issue still remains, if .hide() is applied on the text. However, if it is removed from the Qt created code, it's ok. 
Potentially also the stylesheet for the fancy design could be adapted.

ann0see:
[...]

PR #3397, opened 2024-10-03
Make about chat show up as fullscreen on iOS

ann0see:
Fixes an overflow bug for the about dialog on iOS.

Related to: https://github.com/jamulussoftware/jamulus/pull/3343

**Short description of changes**

Attempt to show chat dialog as full screen. Doesn't work yet - especially if rotating the screen.

CHANGELOG: CONDENSE WITH https://github.com/jamulussoftware/jamulus/pull/3343
**Context: Fixes an issue?**

Fixes: #3383

**Does this change need documentation? What needs to be documented and how?**

No

**Status of this Pull Request**

To be tested. Unfortunately, also the close button on the chat on iOS crashes the application. I think this was documented somewhere and is probably due to a null pointer de-reference.

**What is missing until this pull request can be merged?**

Not sure

## Checklist

-  [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
-  [ ] I tested my code and it does what I want
-  [ ] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#source-code-consistency) 
-  [ ] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
-  [ ] I've filled all the content above

PR #3413, opened 2024-10-14
Fix crash on closing iOS chat dialog

ann0see:

Using close() instead of hide() would crash the app as soon as one taps at the screen after closing the chat dialog. Due to a null pointer dereference (?).
I do **not** understand why the app crashes. Probably it's a qt bug.

```
ASSERT failure in QCoreApplication::sendEvent: "Unexpected null receiver", file /Users/qt/work/qt/qtbase/src/corelib/kernel/qcoreapplication.cpp, line 1587

Can't show file for stack frame : <DBGLLDBStackFrame: 0x7fcf61f51120> - stackNumber:3 - name:qAbort. The file path does not exist on the file system: /Users/qt/work/qt/qtbase/src/corelib/global/qglobal.cpp
```

https://qtcentre.org/threads/5572-QCoreApplication-postEvent-Unexpected-null-receiver (?)

CHANGELOG: iOS: Fix crash on Qt6 after closing the chat window.

**Context: Fixes an issue?**

Related to: https://github.com/jamulussoftware/jamulus/issues/3406

Ready for review.

Discussion. This is a hot fix...
## Checklist

### DIFF
--- a/src/chatdlg.cpp
+++ b/src/chatdlg.cpp
@@ -61,12 +61,11 @@ CChatDlg::CChatDlg ( QWidget* parent ) : CBaseDlg ( parent, Qt::Window ) // use 
 
     pMenu->addMenu ( pEditMenu );
 #if defined( Q_OS_IOS )
-    QAction* action = pMenu->addAction ( tr ( "&Close" ) );
-    connect ( action, SIGNAL ( triggered() ), this, SLOT ( close() ) );
+    QAction* closeAction = pMenu->addAction ( tr ( "&Close" ) );
 #endif
 
 #if defined( ANDROID ) || defined( Q_OS_ANDROID )
-    pEditMenu->addAction ( tr ( "&Close" ), this, SLOT ( close() ), QKeySequence ( Qt::CTRL + Qt::Key_W ) );
+    pEditMenu->addAction ( tr ( "&Close" ), this, SLOT ( OnCloseClicked() ), QKeySequence ( Qt::CTRL + Qt::Key_W ) );
 #endif
 
     // Now tell the layout about the menu
@@ -78,6 +77,10 @@ CChatDlg::CChatDlg ( QWidget* parent ) : CBaseDlg ( parent, Qt::Window ) // use 
     QObject::connect ( butSend, &QPushButton::clicked, this, &CChatDlg::OnSendText );
 
     QObject::connect ( txvChatWindow, &QTextBrowser::anchorClicked, this, &CChatDlg::OnAnchorClicked );
+
+#if defined( Q_OS_IOS )
+    QObject::connect ( closeAction, &QAction::triggered, this, &CChatDlg::OnCloseClicked );
+#endif
 }
 
 void CChatDlg::OnLocalInputTextTextChanged ( const QString& strNewText )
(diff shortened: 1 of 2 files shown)


src/chatdlg.cpp
```diff
@@ -62,7 +62,7 @@ CChatDlg::CChatDlg ( QWidget* parent ) : CBaseDlg ( parent, Qt::Window ) // use
     pMenu->addMenu ( pEditMenu );
 #if defined( Q_OS_IOS )
     QAction* action = pMenu->addAction ( tr ( "&Close" ) );
-    connect ( action, SIGNAL ( triggered() ), this, SLOT ( close() ) );
```
pljones, line 65: Whilst we're doing this, I would like the two lines changed to match the `addAction`s before and after that both include the action handler.

I'll also ask _why_ this should be `hide()` for iOS and `close()` for Android -- I'm going to hazard a guess it should be the same request to Qt on both platforms and that Qt should do the right thing.

ann0see: As far as I understand hide() is close() minus deleting the dialog. 
Somehow on iOS under Qt6 the dialog already gets deleted and close() may attempt to do it twice.

Potentially it's the same on android with Qt6 - but we don't know for sure yet. 

Yes, we could add the close menu entry on all OS - on desktop it seems redundant however.

ann0see:
@softins I've pushed a new build with this + the icon PR on TestFlight.

src/chatdlg.cpp
pljones: I'd argue that UI consistency across platforms isn't "redundant" but makes for a more consistent user experience.

ann0see: The close entry on desktop OS is redundant since there's a close button in the window manager. 

It does not make sense to hide the close button in a menu on mobile OS (at least iOS, where there's no back button).

softins:
Thanks for the updated Testflight. I've tried it out, and can confirm it doesn't crash, and the icon looks fine.

But I have found some oddities regarding the Chat window, which while fairly minor, would be worth investigating.

When opening the Chat window, it does not cover the main menu bar, but instead adds its own menu bar below it:



Closing and re-opening it makes no difference.

This is in contrast to the Settings window, which replaces the main menu bar with its own, and the Connect window, which doesn't have a menu bar at all (it is exited with either Cancel or Connect).

Having connected to a server with a welcome message, the Chat window automatically displayed as usual, but still with the main menu bar showing:



I closed it, disconnected and connected again to check, and it was the same.

I then tapped in the text box to type a message, and the on-screen keyboard appeared, correctly moving the window upwards. After typing the message and pressing Send, the keyboard disappeared again as normal, but now the Chat window was correctly showing _without_ the main menu bar:



The very first time I tried this Testflight, and also when I was trying the previous one, the Chat window showed with the message entry box and Send button half off the bottom of the screen. But I have failed to reproduce this behaviour since, so maybe it was a hangover from the older version?

softins:
Also, at one point after this, I tapped on Connect in the main window, and the Connect dialog shows at the wrong size:



Tipping the iPad to rotate to portrait mode and back to landscape mode reset the dialog size to full screen correctly.

I'm not sure what I did to provoke that behaviour.

ann0see:
I did notice problems like this on the main mixer board too. All can be fixed by rotating the device. 

I believe it's due to an overflow introduced by some widget.

ann0see:
Maybe https://github.com/jamulussoftware/jamulus/pull/3397
Could fix this. I'd like to investigate the UI issues separately.

softins (approved):

ann0see:
@gilgongo @pljones Not sure if you can test this, but I'd like this to be merged. 

The TestFlight link is in the chat.

pljones:
I've no way to test it.

gilgongo:
I'm unable to test on iOS either.

src/chatdlg.cpp
```diff
@@ -61,12 +61,11 @@ CChatDlg::CChatDlg ( QWidget* parent ) : CBaseDlg ( parent, Qt::Window ) // use
 
     pMenu->addMenu ( pEditMenu );
 #if defined( Q_OS_IOS )
-    QAction* action = pMenu->addAction ( tr ( "&Close" ) );
-    connect ( action, SIGNAL ( triggered() ), this, SLOT ( close() ) );
+    QAction* closeAction = pMenu->addAction ( tr ( "&Close" ) );
```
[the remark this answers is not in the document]
ann0see: Using the button in the menu bar still shows the crash on iOS.

pljones:
Can we get this merged and cut a development release so any iOS users have a release they can install and test?  If it breaks, we can revert it.

ann0see:
Yes, sure

PR #3416, opened 2024-10-25
Display server versions in Connect dialog

softins:

Adds a Version column to the server list in the Connect dialog, and populates this once for each server in the list.

CHANGELOG: Client: Display version for each server in the Connect dialog

**Context: Fixes an issue?**

This is a proof of concept in response to the discussion at https://github.com/orgs/jamulussoftware/discussions/3382#discussioncomment-10941423, although not itself related to access control.

On the website, any screenshots of the Connect dialog should be updated.

Ready for review. We can discuss whether to put it in 3.12.0 (if there is one) or 4.0.0

Only a decision about when. We should also check the column width is ok on all platforms.

## Checklist

### DIFF
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -808,13 +808,13 @@ void CConnectDlg::EmitCLServerListPingMes ( const CHostAddress& haServerAddress,
     // block the GUI).
     QThread::msleep ( 11 );
 
-    emit CreateCLServerListPingMes ( haServerAddress );
-
-    // also request the server version if we have not already received it
+    // first request the server version if we have not already received it
     if ( bNeedVersion )
     {
         emit CreateCLServerListReqVerAndOSMes ( haServerAddress );
     }
+
+    emit CreateCLServerListPingMes ( haServerAddress );
 }
 
 void CConnectDlg::SetPingTimeAndNumClientsResult ( const CHostAddress& InetAddr, const int iPingTime, const int iNumClients )
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -113,12 +113,12 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
 #ifdef ANDROID
     // for Android we need larger numbers because of the default font size
     lvwServers->setColumnWidth ( 0, 200 );
-    lvwServers->setColumnWidth ( 1, 130 );
+    lvwServers->setColumnWidth ( 1, 110 );
     lvwServers->setColumnWidth ( 2, 130 );
     lvwServers->setColumnWidth ( 3, 100 );
 #else
     lvwServers->setColumnWidth ( 0, 180 );
-    lvwServers->setColumnWidth ( 1, 75 );
+    lvwServers->setColumnWidth ( 1, 65 );
     lvwServers->setColumnWidth ( 2, 75 );
     lvwServers->setColumnWidth ( 3, 70 );
     lvwServers->setColumnWidth ( 4, 220 );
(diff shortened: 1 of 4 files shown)


softins:
Example screen shot:

image

softins:
If we add this, I think we should also consider again adding #1940

softins:
Well we could rearrange the columns (should probably create some defines or an enum for column numbers), but I don't think hover is a good idea, especially if we want to target tablets and phones too, which don't have hover.

softins:
Personally, I prefer it where I showed above.

ann0see:
It's not essential. Ping is more important for example. So it should not be first in my opinion

softins:
Updated view. I don't like it as much, but am not interested in arguing about it. But at least with an enum for column values it is easier to try different ways.

image

ann0see:
The linux standard window size seems to hide the version in the connect dialog. Not really a problem, but just so you know.

src/connectdlg.h
```diff
@@ -59,12 +59,27 @@ class CConnectDlg : public CBaseDlg, private Ui_CConnectDlgBase
...
+    enum EConnectListViewColumns
+    {
+        LVC_NAME,               // server name
+        LVC_PING,               // ping time
+        LVC_CLIENTS,            // number of connected clients (including additional strings like " (full)")
+        LVC_LOCATION,           // location
+        LVC_VERSION,            // server version
+        LVC_PING_MIN_HIDDEN,    // minimum ping time (invisible)
+        LVC_CLIENTS_MAX_HIDDEN, // maximum number of clients (invisible)
+        LVC_COLUMNS             // total number of columns
```
ann0see, line 80: Alternatively we could most likely also calculate this programmatically. It could go out of sync in future due to a mistake?

ann0see (approved):
Tested and despite some small comments, it looks ok.

src/connectdlg.h
softins: I think a better approach would be to remove the headings from the widget in the `.ui` file and to set them in the code in the same section as we set the widths. I did consider this, but didn't get around to looking up the correct way to do so. Then the enum would be the one source of truth for column order.

softins:
> The linux standard window size seems to hide the version in the connect dialog. Not really a problem, but just so you know.

That's interesting, I didn't notice that issue under Linux on my Pi; the connect window was wide enough to show the version column. Is the connect window size remembered in the `.ini` file?

ann0see:
Yes. The ini file has the correct size. I deleted the ini file to test the size and that's too small.

softins:
> The linux standard window size seems to hide the version in the connect dialog. Not really a problem, but just so you know.

I've updated the default width in `connectdlgbase.ui`. Once there is an ini file it will just get overwritten by the saved value.

ann0see (approved):

mcfnord:
Maybe we could prepare a documentation page that's just a table that lists the advanced features someone might enjoy starting with each particular server version. Presumably this would involve condensing release notes for each version into a single at-a-glance reference.

ann0see:
Yes. It's currently in the changelog. If we do a good job with the changelog we could probably even generate this table automatically

src/connectdlg.cpp
```diff
@@ -112,14 +112,16 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     // must not be specified since this column takes all the remaining space)
 #ifdef ANDROID
```
pljones, line 113: Qt _should_ offer a platform independent way to get how much pixel space is taken up by a string.  If "that's too big generally", then we should then scale _all platforms_ back by the same multiplier.  (OK, not part of this PR, just noticing the nasty `#ifdef`.)

src/connectdlg.cpp
```diff
@@ -148,7 +153,7 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     if ( bShowCompleteRegList )
     {
         lvwServers->setSortingEnabled ( true );
-        lvwServers->sortItems ( 0, Qt::AscendingOrder );
+        lvwServers->sortItems ( LVC_NAME, Qt::AscendingOrder );
```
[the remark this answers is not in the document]
softins: The code you quoted is only used when the undocumented option `--showallservers` is given. In this case, no automatic ping sorting is done, and instead the display can be sorted by clicking on the column headers. The default sort is by name, set by the line above.

If `--showallservers` is not given, then column header sorting is not enabled, and the sorting by ping is done within `CConnectDlg::SetPingTimeAndNumClientsResult()` in these lines:

```
        // Update sorting. Note that the sorting must be the last action for the
        // current item since the topLevelItem(iIdx) is then no longer valid.
        // To avoid that the list is sorted shortly before a double click (which
        // could lead to connecting an incorrect server) the sorting is disabled
        // as long as the mouse is over the list (but it is not disabled for the
        // initial timer of about 2s, see TimerInitialSort) (#293).
        if ( bDoSorting && !bShowCompleteRegList &&
             ( TimerInitialSort.isActive() || !lvwServers->underMouse() ) ) // do not sort if "show all servers"
        {
            lvwServers->sortByColumn ( LVC_PING_MIN_HIDDEN, Qt::AscendingOrder );
        }
```
`LVC_PING_MIN_HIDDEN` is a hidden column used to store the minimum ping time received for each client since the connect dialog was opened.

src/connectdlg.cpp
pljones: I'll try to keep things like this in mind for the general Qt overhaul...

pljones (approved):

ann0see:
Needs updated screenshots.

pljones:
Without Show All Musicians, without `--showallservers`, so not sortable:
image
github.com user-attachments/assets/fafa44dd-3134-4189-bef8-4e350129ad55

With Show All Musicians, with `--showallservers`, sorted descending version
image
github.com user-attachments/assets/a8ee495c-2527-416f-a748-7b0ae19e27bc

Hm.  Not sure the sort is working as expected now I look at it.  I'd expect
```
* | | | |   c12cd6bb 2025-12-21 Merge pull request #3562 from pljones/feature/sort-commits-by-timestamp [Peter L Jones]
```
to mean that cluster of r3_11_0dev-<commit_id> servers with these commit IDs:
```
db481521 2025-10-04 Merge pull request #3544 from pljones/jamuluswebsite-1097-change-email [Peter L Jones]
a624b90e 2026-03-29 Make .github/instructions ignore more focussed [Peter L Jones]
8d4c1557 2026-02-18 Merge pull request #3622 from rdica/IPv6-QoS-fix [Tony Mountifield]
703da09e 2025-05-07 Merge pull request #3498 from ann0see/autobuild/addWindowsSleep [ann0see]
```
would at least have separated those with a timestamp from those without... but looking at the code, I can see why it doesn't.  It's not important enough to fix.  The screenshots above will do for now.  It'll work better after r3_12_0 is out...

PR #3418, opened 2024-10-25
Add "Sort by Channel" to mixer view

### LINKED ISSUE #3417: Add "Sort by Channel" to mixer view

Currently, when "No user sort" is specified, channels appear in the order they're received from the server -- and sometimes this isn't the same as the channel number:
image

Because `--ctrlmidich` uses the channel numbers to assign Jamulus channels to MIDI controllers, it improves user experience to be able to sort the faders to match.

The existing "No user sort" could be replaced with or have its meaning changed to "Sort by channel".

However, this might _not_ be what all users want, so I'm proposing that "Sort by channel" is a new option.

This arises from the changes recently to `--ctrlmidich` - just something I thought of.

pljones:

Adds "Sort by Channel" to mixer view

CHANGELOG: Client: Add "Sort by Channel" to mixer view

**Context: Fixes an issue?**

Fixes #3417 

Yes.  The new option needs documenting.  The entry should explain its relevance to `--ctrlmidich`.

Test working on Ubuntu, Qt 6.8.0

Review, testing on other platforms.

## Checklist


src/audiomixerboard.cpp
```diff
@@ -1092,39 +1092,47 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
             iMyFader = static_cast<int> ( i );
         }
 
-        if ( eChSortType == ST_BY_NAME )
+        switch ( eChSortType )
```
ann0see, line 1095: Makes sense.

src/audiomixerboard.cpp
```diff
@@ -1092,39 +1092,47 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
...
-        else if ( eChSortType == ST_BY_GROUPID )
-        {
+            break;
+        case ST_BY_GROUPID:
+            // sort first "by group" and second "by name" by adding the name after the group
             if ( vecpChanFader[i]->GetGroupID() == INVALID_INDEX )
             {
                 // put channels without a group at the end
-                PairList << QPair<QString, size_t> ( "z", i ); // group IDs are numbers, use letter to put it at the end
+                PairList << QPair<QString, size_t> ( "999" + vecpChanFader[i]->GetReceivedName().toLower(),
```
ann0see, line 1115: Seems hacky to offset by 999

src/clientdlg.cpp
```diff
@@ -333,6 +333,9 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     QAction* ByCityAction =
         pViewMenu->addAction ( tr ( "Sort Users by &City" ), this, SLOT ( OnSortChannelsByCity() ), QKeySequence ( Qt::CTRL + Qt::Key_T ) );
 
+    QAction* ByChannelAction =
+        pViewMenu->addAction ( tr ( "Sort Users by Ch&annel" ), this, SLOT ( OnSortChannelsByChannel() ), QKeySequence ( Qt::CTRL + Qt::Key_A ) );
```
ann0see, line 337: Maybe this is not expressive enough and too technical. 
We should make sure that the user knows it's mostly "something todo with MIDI" and not random.

ann0see: @gilgongo Any ideas to make this concise but not too technical?
The ordering seems "random" unless midi is enabled.

gilgongo: Until now, we've always used "Channel" to mean the audio signal as part of a mix - so does this introduce a new meaning of "MIDI channel"? 

If so, I think we'd need to say, "Sort Users by MIDI Channel" (although I'd prefer we said just "Sort by MIDI Channel" really, because it should be obvious that you're sorting the UI Faders/Users).

src/audiomixerboard.cpp
pljones: Far less hacky than using "z" as bigger than "8", though.  The groups will be "001" to "008" and "not a group" will be "999".

src/clientdlg.cpp
pljones: No, this is nothing to do with MIDI Channel.  This is the Jamulus Server-assigned Channel ID.  It's inherent to how `--ctrlmidich` works, hence my statement in the related issue that the documentation needs to make the meaning of the menu option clear.

gilgongo: Oh OK. "Sort by Channel" does feel a bit elliptical somehow (like a numeric list being "sorted by value"). Would "Sort by assigned Channel" be any clearer, assuming the curious would then read the docs on it? An alternative could be "Default sort", which again would be expanded upon in the docs.

pljones: It's not the default sort, as explained above.  The default sort, explicitly, puts new channels on the left of the mixer - that's a definition and it's intended to work that way so that the existing channels, when default sort is selected, don't keep moving around.  Someone leaves, their fader is removed and everyone moves left.  Someone joins, they're added at the right-hand end.  I don't think that should be changed.

"Assigned" begs the question of "by whom?".  As the `--ctrlmidich` docs answer that -- it's the Server -- then can I suggest "Sort by Server Channel"?

src/audiomixerboard.cpp
pljones: (And it allows for groups expanding to be one per server channel up to more than max server channels...)

src/clientdlg.cpp
pljones: By the way, I've added a change to the `--ctrlmidich` Pull Request that would mean the "Sort by Server Channel" could either be invisible (which would potentially be confusing) or disabled if MIDI controls were not enabled.  I'm not convinced it would help anyone, though.

gilgongo: "Sort by Server Channel" is perhaps marginally less "recursive" from the POV of someone who knows nothing about `--ctrlmidich`. Although I'd still expect somebody who was curious to simply look it up in the docs, because how the Server applies the sort would otherwise be mysterious to them. 

I don't think there's an easy way of wording it :-)

ann0see: Sort by Server channel is ok I think.

src/util.h
```diff
@@ -558,11 +558,12 @@ enum ERecorderState
 enum EChSortType
```
ann0see, line 558: Not part of this PR, but having this enum in utils looks unorganized.

ann0see (commented):
I'll test it on my device soon.

gilgongo (approved):

ann0see (approved):
Works ok on Debian

ann0see:
Should still have a quick look by @softins, but otherwise I think it's good to be merged. @pljones, please tag it needs documentation too

softins:
> Should still have a quick look by @softins, but otherwise I think it's good to be merged.

Will look soon!

src/util.h
pljones: Yeah.  I'll look to move it.

pljones: OK it's in util.h because it's used in CSettings as well as the mixer board, I think -- all of the `static_cast<ENUM>` enums that CSettings does get shared with somewhere else and CSettings doesn't want to be fishing through lots of headers, so they get dumped into util.h.

(Ideally, CSettings wouldn't work like that: each component would have its own settings and read or write them to the XML that CSettings simply read from a file or wrote to a file...  But _that_ is definitely a different PR!)

softins (approved):
Looks good to me - nice code!

Tested on a Pi with Qt 5.15.8

pljones:
https://github.com/jamulussoftware/jamuluswebsite/issues/1059 raised to get the documentation change done.

PR #3419, opened 2024-11-06
Always give own channel MIDI number 0

softins:

Ensure the user's own channel is always given MIDI channel 0, so that the user is always the leftmost fader when using an external MIDI fader bank.

1. Moved my channel ID from the Mixer to the Client, so it is available when headless too.
2. Client's own channel, sent by the server, is always assigned to MIDI number 0, for left-most fader. Server channel numbers less than the user's own channel are bumped up by 1 to get the MIDI channel.
3. Translate between server channel and MIDI number at the client level.

CHANGELOG: Client: always assign MIDI channel 0 to the user's own channel.

**Context: Fixes an issue?**

Raised in https://github.com/orgs/jamulussoftware/discussions/2220#discussioncomment-10811813

This supersedes the solution proposed in #3394, and in particular avoids the use of a duplicate fader. It also needs no change to the `--ctrlmidich` parameter.

It is an implementation of my proposal in https://github.com/jamulussoftware/jamulus/pull/3394#issuecomment-2447827384

Only to document that the user's own channel will always be given fader channel 0, and that this is only relevant
when using `--ctrlmidich` to assign an external fader bank.

Tested and ready for review

Consensus on the approach and testing on other platforms. Tested on Pi Linux.

## Checklist

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1246,6 +1246,8 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
             iFaderNumber[vecChanInfo[iFader].iChanID] = static_cast<int> ( iFader );
         }
 
+        const int iMyChannelID = pClient->GetMyChannelID();
+
         // Hide all unused faders and initialize used ones
         for ( size_t iChanID = 0; iChanID < MAX_NUM_CHANNELS; iChanID++ )
         {
@@ -1267,7 +1269,7 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
                 vecpChanFader[iChanID]->Reset();
                 vecAvgLevels[iChanID] = 0.0f;
 
-                if ( static_cast<int> ( iChanID ) == pClient->GetMyChannelID() )
+                if ( static_cast<int> ( iChanID ) == iMyChannelID )
                 {
                     // this is my own fader --> set fader property
                     vecpChanFader[iChanID]->SetIsMyOwnFader();
@@ -1288,8 +1290,7 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
                 // we can adjust the level even if no fader was visible.
                 // The fader level of 100 % is the default in the
                 // server, in that case we do not have to do anything here.
-                if ( ( !bNoFaderVisible ||
-                       ( ( pClient->GetMyChannelID() != INVALID_INDEX ) && ( pClient->GetMyChannelID() != static_cast<int> ( iChanID ) ) ) ) &&
+                if ( ( !bNoFaderVisible || ( ( iMyChannelID != INVALID_INDEX ) && ( iMyChannelID != static_cast<int> ( iChanID ) ) ) ) &&
                      ( pSettings->iNewClientFaderLevel != 100 ) )
                 {
                     // the value is in percent -> convert range
@@ -1395,10 +1396,12 @@ void CAudioMixerBoard::SetAllFaderLevelsToNewClientLevel()
 {
     QMutexLocker locker ( &Mutex );
 
+    const int iMyChannelID = pClient->GetMyChannelID();
+
     for ( size_t i = 0; i < MAX_NUM_CHANNELS; i++ )
     {
         // only apply to visible faders and not to my own channel fader
-        if ( vecpChanFader[i]->IsVisible() && ( static_cast<int> ( i ) != pClient->GetMyChannelID() ) )
+        if ( vecpChanFader[i]->IsVisible() && ( static_cast<int> ( i ) != iMyChannelID ) )
         {
             // the value is in percent -> convert range, also use the group
             // update flag to make sure the group values are all set to the
@@ -1423,11 +1426,13 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
     CVector<CVector<float>> levels;
     levels.resize ( MAX_NUM_FADER_GROUPS + 1 );
 
+    const int iMyChannelID = pClient->GetMyChannelID();
+
     // compute min/max level per group and number of channels per group
     for ( size_t i = 0; i < MAX_NUM_CHANNELS; ++i )
     {
         // only apply to visible faders (and not to my own channel fader)
-        if ( vecpChanFader[i]->IsVisible() && ( static_cast<int> ( i ) != pClient->GetMyChannelID() ) )
+        if ( vecpChanFader[i]->IsVisible() && ( static_cast<int> ( i ) != iMyChannelID ) )
         {
             // map averaged meter output level to decibels
             // (invert CStereoSignalLevelMeter::CalcLogResultForMeter)
@@ -1517,7 +1522,7 @@ void CAudioMixerBoard::AutoAdjustAllFaderLevels()
     for ( size_t i = 0; i < MAX_NUM_CHANNELS; ++i )
     {
         // only apply to visible faders (and not to my own channel fader)
-        if ( vecpChanFader[i]->IsVisible() && ( static_cast<int> ( i ) != pClient->GetMyChannelID() ) )
+        if ( vecpChanFader[i]->IsVisible() && ( static_cast<int> ( i ) != iMyChannelID ) )
         {
             // map averaged meter output level to decibels
             // (invert CStereoSignalLevelMeter::CalcLogResultForMeter)
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -260,7 +260,7 @@ void CChannelFader::SetGUIDesign ( const EGUIDesign eNewDesign )
     UpdateGroupIDDependencies();
 
     // the instrument picture might need scaling after a style change
-    SetChannelInfos ( cReceivedChanInfo );
+    SetChannelInfos ( cReceivedChanInfo, cReceivedMIDIID );
 }
 
 void CChannelFader::SetMeterStyle ( const EMeterStyle eNewMeterStyle )
@@ -436,6 +436,7 @@ void CChannelFader::Reset()
     plblCountryFlag->setVisible ( false );
     plblCountryFlag->setToolTip ( "" );
     cReceivedChanInfo = CChannelInfo();
+    cReceivedMIDIID   = INVALID_INDEX;
     SetupFaderTag ( SL_NOT_SET );
 
     // set a defined tool tip time out
@@ -666,10 +667,11 @@ void CChannelFader::UpdateSoloState ( const bool bNewOtherSoloState )
 
 void CChannelFader::SetChannelLevel ( const uint16_t iLevel ) { plbrChannelLevel->SetValue ( iLevel ); }
 
-void CChannelFader::SetChannelInfos ( const CChannelInfo& cChanInfo )
+void CChannelFader::SetChannelInfos ( const CChannelInfo& cChanInfo, int iMIDIID )
 {
     // store received channel info
     cReceivedChanInfo = cChanInfo;
+    cReceivedMIDIID   = iMIDIID;
 
     // init properties for the tool tip
     int              iTTInstrument = CInstPictures::GetNotUsedInstrument();
@@ -682,7 +684,7 @@ void CChannelFader::SetChannelInfos ( const CChannelInfo& cChanInfo )
     // show channel numbers if --ctrlmidich is used (#241, #95)
     if ( bMIDICtrlUsed )
     {
-        strModText.prepend ( QString().setNum ( cChanInfo.iChanID ) + ":" );
+        strModText.prepend ( QString().setNum ( iMIDIID ) + ":" );
     }
 
     QTextBoundaryFinder tbfName ( QTextBoundaryFinder::Grapheme, cChanInfo.strName );
@@ -886,7 +888,6 @@ CAudioMixerBoard::CAudioMixerBoard ( QWidget* parent ) :
     bDisplayPans ( false ),
     bIsPanSupported ( false ),
     bNoFaderVisible ( true ),
-    iMyChannelID ( INVALID_INDEX ),
     iRunningNewClientCnt ( 0 ),
     iNumMixerPanelRows ( 1 ), // pSettings->iNumMixerPanelRows is not yet available
     strServerName ( "" ),
@@ -1049,7 +1050,6 @@ void CAudioMixerBoard::HideAll()
     bIsPanSupported      = false;
     bNoFaderVisible      = true;
     eRecorderState       = RS_UNDEFINED;
-    iMyChannelID         = INVALID_INDEX;
     iRunningNewClientCnt = 0; // reset running counter on new server connection
 
     // use original order of channel (by server ID)
@@ -1123,7 +1123,7 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
             }
             break;
         case ST_BY_SERVER_CHANNEL:
-            PairList << QPair<QString, size_t> ( QString ( "%1" ).arg ( vecpChanFader[i]->GetReceivedChID(), 3, 10, QLatin1Char ( '0' ) ) +
+            PairList << QPair<QString, size_t> ( QString ( "%1" ).arg ( vecpChanFader[i]->GetReceivedMIDIID(), 3, 10, QLatin1Char ( '0' ) ) +
                                                      vecpChanFader[i]->GetReceivedName().toLower(),
                                                  i );
             break;
@@ -1267,7 +1267,7 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
                 vecpChanFader[iChanID]->Reset();
                 vecAvgLevels[iChanID] = 0.0f;
 
-                if ( static_cast<int> ( iChanID ) == iMyChannelID )
+                if ( static_cast<int> ( iChanID ) == pClient->GetMyChannelID() )
                 {
                     // this is my own fader --> set fader property
                     vecpChanFader[iChanID]->SetIsMyOwnFader();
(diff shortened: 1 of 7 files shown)


softins:
@AndersGoran - I would be interested in your feedback on this solution. Thanks!

src/audiomixerboard.cpp
```diff
@@ -436,6 +436,7 @@ void CChannelFader::Reset()
     plblCountryFlag->setVisible ( false );
     plblCountryFlag->setToolTip ( "" );
     cReceivedChanInfo = CChannelInfo();
+    cReceivedMIDIID   = INVALID_INDEX;
```
pljones, line 439: Should we be using `INVALID_INDEX` here?

src/audiomixerboard.cpp
```diff
@@ -1123,7 +1123,7 @@ void CAudioMixerBoard::ChangeFaderOrder ( const EChSortType eChSortType )
             }
             break;
         case ST_BY_SERVER_CHANNEL:
```
pljones, line 1125: Should this get renamed if it's always the received MIDI ID?

src/audiomixerboard.cpp
```diff
@@ -1267,7 +1267,7 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
                 vecpChanFader[iChanID]->Reset();
                 vecAvgLevels[iChanID] = 0.0f;
 
-                if ( static_cast<int> ( iChanID ) == iMyChannelID )
+                if ( static_cast<int> ( iChanID ) == pClient->GetMyChannelID() )
```
pljones, line 1270: Maybe stick `pClient->GetMyChannelID()` into a variable as it's used in a few places here.

src/client.cpp
```diff
@@ -789,8 +790,44 @@ void CClient::OnHandledSignal ( int sigNum )
 #endif
 }
 
-void CClient::OnControllerInFaderLevel ( int iChannelIdx, int iValue )
+// Ensure user's own channel is always given MIDI offset 0, so it is the first physical fader
+// All channels with a client ID less than user's own will use ID+1 as their MIDI offset
+
+int CClient::ChanToMIDI ( const int iChannelIdx )
+{
+    if ( iMyChannelID == INVALID_INDEX )
```
pljones, line 798: As above, it was suggested using `INVALID_INDEX` for something that isn't an index into a vector wasn't a good idea.

src/client.h
```diff
@@ -272,6 +274,10 @@ class CClient : public QObject
         Channel.GetBufErrorRates ( vecErrRates, dLimit, dMaxUpLimit );
     }
 
+    // convert between MIDI index and real channel index
```
pljones, line 277: Shouldn't this be done my the MIDI-handling code?  Otherwise, it could do with a longer comment explaining why it's here.

src/client.cpp
```diff
@@ -789,8 +790,44 @@ void CClient::OnHandledSignal ( int sigNum )
...
+
+    return iMIDIIdx;
+}
+
+// Handle received MIDI controls
+
+void CClient::OnControllerInFaderLevel ( int iMIDIIdx, int iValue )
 {
+    // map the MIDI index to the real channel number
+    const int iChannelIdx = MIDIToChan ( iMIDIIdx );
```
pljones, line 829: The client shouldn't need to worry about this.  It should have been handled in the MIDI code.

src/sound/soundbase.h
```diff
@@ -168,9 +168,9 @@ class CSoundBase : public QThread
 
 signals:
     void ReinitRequest ( int iSndCrdResetType );
-    void ControllerInFaderLevel ( int iChannelIdx, int iValue );
-    void ControllerInPanValue ( int iChannelIdx, int iValue );
-    void ControllerInFaderIsSolo ( int iChannelIdx, bool bIsSolo );
-    void ControllerInFaderIsMute ( int iChannelIdx, bool bIsMute );
+    void ControllerInFaderLevel ( int iMIDIIdx, int iValue );
```
pljones, line 171: `CMidiCtlEntry` should have the first entry renamed to match if it's not `iChannel` any more.  Or else rename it to `iChannel` -- it's the value of `CMidiCtlEntry::iChannel`, after all...

src/sound/soundbase.h
pljones, line 1: I still think `CSoundBase::ParseMIDIMessage` is where the mapping from MIDI Controller Number to Jamulus Channel Number should happen.  Keep the MIDI out of CClient entirely.

src/audiomixerboard.cpp
softins: Well it just needs to be something like `-1` that couldn't match a real MIDI Controller offset from `0` upwards. I don't mind whether we use that symbol or define another symbol for it, although I don't see the point of changing it.

src/audiomixerboard.cpp
softins: I did wonder about that, and even started to change the name, but then had second thoughts, as it implied the menu option should change to "Sort by MIDI Channel". I think this could be more confusing to some users.

The actual server channel numbers and the channel assigned to the particular client are never actually visible to the user, except when used for MIDI channels.

What we are doing here by mapping is presenting to the user a view that _always_ uses `0` for the user's own channel, and maps the rest of the channels up by one. As far as the user is concerned, they _are_ server channels (even if virtual); we just guarantee that the user themselves is channel zero. So whether or not a MIDI controller is used, a "Sort by Server Channel" will always put the user's own channel first, whether or not "Show Own Fader First" is enabled. I think that is a worthwhile improvement.

src/audiomixerboard.cpp
softins: The value doesn't change while connected, and it is just a simple Get method, so I don't think threads are relevant in this case. However, I see most of these uses are within a loop, so I'm happy to fetch once to a local variable within each function and use that within the loop. I'll add a commit for that shortly.

softins:
> I'm not particularly happy with putting MIDI knowledge into CClient

I can see where you're coming from, but I'm not convinced. At the moment, Soundbase doesn't know anything about Jamulus channels, and particularly not about which is _our_ channel. All it does is pass MIDI numbers (CC offsets) to the client.

It's the client that know what our channel number is, and decides how to use and map those MIDI numbers.

In fact before this PR, it was only the Audio Mixer Board that knew about our channel. That meant decisions based on it could only be made at the GUI level, and precluded headless operation. I needed to move `iMyChannelID` down to the `CClient` so that it was also available for the mapping in headless mode.

You mentioned both MIDI and JSON-RPC. Should they all be responsible for knowing what our current channel number is, so that they can each do their own mapping? I'm not convinced. They can both just rest in the knowledge that virtual channel number 0 will always control our own channel, and rely on `CClient` to to the right thing. That isn't specific to MIDI.

So I would still argue in favour of what I put in this PR.

src/sound/soundbase.h
softins: It's really a "virtual channel", where 0 is _our_ fader, and the others are allocated in server order.

In MIDI-speak it's not a channel anyway, it's a controller offset. We are just using it to access a Jamulus channel.

I'm happy to rename items if you have suggestions, but I still think the architecture is ok.

src/sound/soundbase.h
softins: See my general comment in the main thread.

src/audiomixerboard.cpp
softins: Done 072b2af

src/audiomixerboard.cpp
pljones: That's what I said.  I was told it was confusing so I changed it.  I'm raising here because it was one of the first objections on my pull request, so I'm assuming it was something important.

src/audiomixerboard.cpp
pljones: I think that's a definite source of potential confusion, to be honest.  I'd _only_ want my fader sorted first if I asked for it, if there's an option in the menu.  I'd be wondering why the option was there, otherwise.

Certainly for "No sorting", always putting own fader first is wrong unless requested.

src/client.cpp
pljones: See my PR.

pljones:
If you look at my implementation, you'll see that soundbase.cpp continues to know nothing about Jamulus channels - all that changes is the parsing of the `--ctrlmidich` argument.  The determination of channel remains localised in audiomixerboard.cpp, as it is now.

I think you're pulling into CClient concerns that shouldn't be brought in.

My implementation also removes CClientDlg needing to be involved in knowing about MIDI controls, so improving the separation rather than making it worse, like this.

softins:
> If you look at my implementation, you'll see that soundbase.cpp continues to know nothing about Jamulus channels - all that changes is the parsing of the `--ctrlmidich` argument. The determination of channel remains localised in audiomixerboard.cpp, as it is now.

Except that you end up with two separate faders controlling the "own" channel, rather than having the "own" channel only being fader 0 and the others being arranged to suit.

And audiomixerboard is way to high up the hierarchy, as I found, which means the headless code can't use it.

> I think you're pulling into CClient concerns that shouldn't be brought in.

I still disagree. I'm effectively virtualising the channel numbers so that own channel is always 0.

> My implementation also removes CClientDlg needing to be involved in knowing about MIDI controls, so improving the separation rather than making it worse, like this.

`CClientDlg` is not involved. If you look at my changes, the only changes to it are to pass the client pointer up to the mixer board, and the removal of a couple of unneeded signal/slot connections.

softins:
This shouldn't be such hard work :(

src/sound/soundbase.h
softins: Well, whatever. My problem with the `z` solution is the duplicate fader. That's a show-stopper for me, and why I put in a lot of work to come up with an improved solution.

I'm going to leave this for a while.

pljones:
> Except that you end up with two separate faders controlling the "own" channel, rather than having the "own" channel only being fader 0 and the others being arranged to suit.

Yes, that's to be expected:
- assign self + four faders
- join a server with no one on, you get self (first fader), plus self (second fader)
- join a server with four people on, you get self (first fader) and four other people (assuming the server doesn't give you a lower fader number than four)
- join a server with 150 other people on, same deal
- join a server with three other people on, you get self (first fader), those three in 1 to 3 (probably) and self again

I don't see that as confusing or a problem.

pljones:
> And audiomixerboard is way to high up the hierarchy, as I found, which means the headless code can't use it.

Agreed.  The "model-view-controller" concept torn into shreds.  Only CClient should know about state (whether transient or persistent).  The UI controls, JSON-RPC controls, MIDI controls, etc, shouldn't have to know anything - just send signals for "what just happened".

Maybe that should be what Jamulus 4 does...

pljones:
> I still disagree. I'm effectively virtualising the channel numbers so that own channel is always 0.

If that's the intent, discussing "MIDI" throughout the approach is highly misleading.

softins:
> > I still disagree. I'm effectively virtualising the channel numbers so that own channel is always 0.
> 
> If that's the intent, discussing "MIDI" throughout the approach is highly misleading.

Well, yes, but that concept only came to mind when you started also relating JSON-RPC to it.

pljones:
OK -- maybe the special `z` `--ctrlmidich` flag should _force_ "Own Channel First" in the UI sort order.  Maybe it should also _force_ "Sort by Jamulus Channel".  This way the relationship between UI and MIDI hardware controls would be fixed.  (I was thinking of doing this as a follow up to my original PR.)

softins:
I still think there is a better approach. I'm just writing up some notes to post in a while.

softins:
> > > I still disagree. I'm effectively virtualising the channel numbers so that own channel is always 0.
> >
> >
> > If that's the intent, discussing "MIDI" throughout the approach is highly misleading.
>
> Well, yes, but that concept only came to mind when you started also relating JSON-RPC to it.

OK, let's take this concept and run with it, although without the "virtual" term.

Currently, there is no distinction between "client-side" channel and "server-side" channel. The `iChannelID` or `iChannelIdx`, and also the `iMyChannelID`, currently refer to the channel index allocated within the connected server and sent to the client. The Jamulus protocol by definition uses the server-side channel number to identify channels.

Let's decouple the client-side and server-side channel numbers, and make "client-side channel" a concept local to an individual client, mapping it within the client at the lowest level possible to the "server-side channel". Something like this within `client.h`:


```
class CClientChannel {
  ...
  int iServerChannelID;
  ...
};

CClientChannel clientChannels[MAX_NUM_CHANNELS];
```

Maybe we put other items in there that are currently just arrays indexed by server channel index. Or maybe not, but they would now be indexed by client-side channel ID.

- When the client connects to a server (unless the server is ancient), it receives a `CLIENT_ID` message before anything else, containing the server-side channel allocated to that client. We assign that server channel to client channel 0, always, and put the received ID into `clientChannels[0].iServerChannelID`.

- When we receive a list of connected clients from the server, for any new server channel in the list, we allocated the lowest free client channel to it and again store the server channel in `iServerChannelID`. For existing known server channels, we just look up the associated client channel to update its data if necessary.

- If connected to an ancient server that doesn't send `CLIENT_ID`, the channels will just be allocated as received, so the "own" channel could be anywhere. That's unavoidable, but hopefully very rare. In that case, we don't reserve client channel 0, and it will get used by the first server channel to be seen.

- Any time we pass an `iChannelIdx` or `iChannelID` around within the client hierarchy, it is referring _only_ to a client channel.

- Server channel numbers _only_ get used when processing or sending protocol messages. If we have a server channel ID and need to find the associated client channel, we can either do a quick linear search for matching `iServerChannelID` or can maintain a lookup table of `CClientChannel*` pointers indexed by server channel ID. Obviously, to convert from a client channel ID to a server channel ID for sending to the server, we just index the `CClientChannel` array/vector by client channel ID and fetch `iServerChannelID`. Both of these operations should be local to the `CClient`.

- Once a server channel has been assigned to a specific client channel, that assignment will not be changed until that server channel disappears again. This will avoid things like fader assignments jumping unexpectedly. When a server-side channel disappears due a remote client disconnecting, our client-side channel that it was associated with gets marked as free for re-use (`iServerChannelID = INVALID_INDEX;`), but other channels do not get moved to occupy the freed client channel.

- Midi controller number offsets can then be used unchanged to index into client-side channels, and channel 0 will _always_ be the client's own channel (except with an ancient server). There will be no need for the kludgy `z` option in `--ctrlmidich`, nor for a duplicate fader.

- Other potential operations such as a JSON-RPC method would also identify channels only by client-side channel ID. As would the GUI Audio Mixer Board.

- "No sorting" will by default be by client-side channel ID, so "Sort by Server Channel" would no longer be required. I can't see any reason to sort by actual server-side channel number.

- This technique would also overcome the observation of @AndersGoran that when he connected to a server where the only other user had a high-numbered server channel, he could not control it with a MIDI fader.

In view of recent experience, I'm not going to waste a lot of time implementing the above until there is agreement about the approach.

pljones:
OK, from a GUI perspective - all the existing sort options would continue to work as is, except "Sort by Channel", where "My Channel" would automatically be first (and "Own channel first" would have no visible effect).  `--ctrlmidich` would show _"client channel"_ and those would always _start_ in range.  The client would no long "care" what server channel was used.  So:

A server with one person on where they got server channel 10 allocated when they joined.
I join:
- the server allocates me server channel 0 (as it's now free) and I store that in `clientChannels[0].iServerChannelID`
- the server tells me there's someone else on server channel 10 and I store that in `clientChannels[1].iServerChannelID` (because that's my next free channel as I'm just joining)
- someone else joins and gets server channel 1 (that's also now free) and I store that in `clientChannels[2].iServerChannelID`

Now two different scenarios - not problems, just to think through:
- the person who was on leaves, freeing server channel 10 and I free `clientChannels[1]`.
- Scenario 1:
  - someone else new joins, gets server channel 2 and I store that in `clientChannels[1].iServerChannelID` because it's the lowest free
  - the person who was on and left rejoins almost immediately and gets server channel 3, that in `clientChannels[3].iServerChannelID`
- Scenario 2:
  - the person who was on and left rejoins almost immediately and gets server channel 2, that goes in `clientChannels[1].iServerChannelID`
  - someone else new joins, gets server channel 3 and I store that in `clientChannels[3].iServerChannelID` because it's the lowest free

OK, I think that makes sense, too -- if you leave and rejoin, the likelihood is you'll get the same `clientChannels` index, right?  If you leave and someone else joins before you get back, hard luck, you're in a different place on the desk now.

pljones:
From there, GUI, JSON-RPC and MIDI control of CClient messaging to the server to adjust the mix:
- I think CClient needs to have the channel volume level stored (so the GUI fader just tells CClient to adjust that, and CClient knows to tell the server to adjust the mix, JSON-RPC and MIDI make the same call); if the value is changed, CClient emits and "volume changed" signal (the GUI fader tracks this to follow changes from JSON-RPC and MIDI but ignores it whilst doing its own update).
- Same for mute and solo settings -- currently these are embedded in the GUI but should be managed in CClient to allow JSON-RPC and MIDI control not rely on the GUI.

So I think this is a good first step (and a discrete change from the above).

pljones:
I'm wondering if we need to avoid making `CClient` (or, rather, `client.cpp`) too complicated -- should the concept of the mixer (which spans audio I/O and network I/O conceptually) be handled by `CClient`?  Should `CClient` get pared down to managing the client protocol message requests and responses?

That would mean the existing "GUI-only" mixing desk would be replaced by a class that was independent of `CClient` and the GUI that just managed all the interaction between `CClient` and the `CClientChannel[]`.  The GUI would be a visual presentation layer over _that_, and it would also have JSON-RPC and MIDI "presentation" layers.

ann0see:
> That would mean the existing "GUI-only" mixing desk would be split into a class that was independent of CClient and the GUI that just managed all the interaction between CClient and the CClientChannel[]. The GUI would then be a visual presentation layer over that, and there would also be JSON-RPC and MIDI "presentation" layers.


I was honestly also thinking about this for a while. From the perspective of good separation it would make sense:
1. MIDI handling gets less messy
2. JSON-RPC can get feature parity with the GUI
3. Mobile GUI could be another "GUI Client" which is native to the mobile OS

The biggest downside would be that this may be hard to get through a review.

softins:
These are all good thoughts. But I think we should go incrementally, rather than for a big bang which might take ages to finish or not even get around to being completed.

So I would think of implementing what I suggested above within `CClient` and not yet making any changes to the GUI and Audio Mixer. That could then be done separately afterwards. I don't actually think the mixer is within `CClient` anyway. When running with a GUI, we have `CClientDlg` above it, and `CAudioMixerBoard` above that, with signals from below being passed to slots above. It should all work unchanged and won't care that `iChannelIdx` is now a client-side channel instead of a server-side channel.

I'm aware too that it will be necessary to translate from server channel IDs to client channel IDs within `CChanInfo`, when passing them up to the mixer.

pljones:
Yep, not suggesting doing it in one go.  But having these goals laid out clearly as incremental planned steps would be useful, I think, so we can assess progress to a better implementation.

softins:
Please continue any discussion in #3423, the result of which will be a PR in due course that will supersede the one here (and also supersede #3394)

PR #3426, opened 2024-11-20
Manage client-side channels locally, own channel always 0

### LINKED ISSUE #3423: Make client-side channels independent of server-side channels
I'm opening this issue so that implementation can be discussed independently of #3419, which will be superseded by this approach. I'll copy the relevant comments across with attribution.

_Originally posted by @softins in https://github.com/jamulussoftware/jamulus/issues/3419#issuecomment-2466251470_

> > > I still disagree. I'm effectively virtualising the channel numbers so that own channel is always 0.
> >
> >
> > If that's the intent, discussing "MIDI" throughout the approach is highly misleading.
>
> Well, yes, but that concept only came to mind when you started also relating JSON-RPC to it.

OK, let's take this concept and run with it, although without the "virtual" term.

Currently, there is no distinction between "client-side" channel and "server-side" channel. The `iChannelID` or `iChannelIdx`, and also the `iMyChannelID`, currently refer to the channel index allocated within the connected server and sent to the client. The Jamulus protocol by definition uses the server-side channel number to identify channels.

Let's decouple the client-side and server-side channel numbers, and make "client-side channel" a concept local to an individual client, mapping it within the client at the lowest level possible to the "server-side channel". Something like this within `client.h`:

```
class CClientChannel {
  ...
  int iServerChannelID;
  ...
};

CClientChannel clientChannels[MAX_NUM_CHANNELS];
```

Maybe we put other items in there that are currently just arrays indexed by server channel index. Or maybe not, but they would now be indexed by client-side channel ID.

- When the client connects to a server (unless the server is ancient), it receives a `CLIENT_ID` message before anything else, containing the server-side channel allocated to that client. We assign that server channel to client channel 0, always, and put the received ID into `clientChannels[0].iServerChannelID`.

- When we receive a list of connected clients from the server, for any new server channel in the list, we allocated the lowest free client channel to it and again store the server channel in `iServerChannelID`. For existing known server channels, we just look up the associated client channel to update its data if necessary.

- If connected to an ancient server that doesn't send `CLIENT_ID`, the channels will just be allocated as received, so the "own" channel could be anywhere. That's unavoidable, but hopefully very rare. In that case, we don't reserve client channel 0, and it will get used by the first server channel to be seen.

- Any time we pass an `iChannelIdx` or `iChannelID` around within the client hierarchy, it is referring _only_ to a client channel.

- Server channel numbers _only_ get used when processing or sending protocol messages. If we have a server channel ID and need to find the associated client channel, we can either do a quick linear search for matching `iServerChannelID` or can maintain a lookup table of ~`CClientChannel*` pointers~ integer client channel IDs, indexed by server channel ID. Obviously, to convert from a client channel ID to a server channel ID for sending to the server, we just index the `CClientChannel` array/vector by client channel ID and fetch `iServerChannelID`. Both of these operations should be local to the `CClient`.

- Once a server channel has been assigned to a specific client channel, that assignment will not be changed until that server channel disappears again. This will avoid things like fader assignments jumping unexpectedly. When a server-side channel disappears due a remote client disconnecting, our client-side channel that it was associated with gets marked as free for re-use (`iServerChannelID = INVALID_INDEX;`), but other channels do not get moved to occupy the freed client channel.

- Midi controller number offsets can then be used unchanged to index into client-side channels, and channel 0 will _always_ be the client's own channel (except with an ancient server). There will be no need for the kludgy `z` option in `--ctrlmidich`, nor for a duplicate fader.

- Other potential operations such as a JSON-RPC method would also identify channels only by client-side channel ID. As would the GUI Audio Mixer Board.

- ~"No sorting" will by default be by client-side channel ID, so "Sort by Server Channel" would no longer be required. I can't see any reason to sort by actual server-side channel number.~ Actually on second thoughts, it might still be required for the same reason it was introduced, due to the current allocation strategy of the faders in the GUI. But it could just be renamed to “Sort by Channel”, as it would be sorting on the client-side channel number.

- This technique would also overcome the observation of @AndersGoran that when he connected to a server where the only other user had a high-numbered server channel, he could not control it with a MIDI fader.

softins:

Manages the allocation of client-side channels independently of server-side channels, and maps between
the two at the boundary between CClient and the lower-level CChannel and CProtocol.

All layers above (e.g. CClientDlg and the audio mixer board) will use client-side channels transparently.

When talking to a server version 3.5.5 or newer, the client is told its own server channel, and will always map
this to client channel 0. This allows the user of a MIDI device or JSON-RPC client to know that channel 0 will always control
the user's own audio channel.

In addition, when connecting to a server that only has a small number of high-numbered channels remaining
after many users have left, those channels will automatically be mapped to lower client channel numbers and
so will be controllable using local MIDI faders.

CHANGELOG: Client: allocate channel numbers locally and always give user own channel of 0

**Context: Fixes an issue?**

Resolves #3423 and supersedes both #3419 and #3394

Most operation is unchanged from before, but it is worth mentioning that the user's own channel will always be
number 0 unless connected to a server older than 3.5.5

Tested and working, ready for review.

Testing on multiple platforms.

## Checklist

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -361,8 +361,7 @@ void CClient::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo )
         Q_ASSERT ( iActiveChannels == iNumConnectedClients );
     }
 
-    // TODO Does vecChanInfo need to be ordered by client channel ID instead of server channel ID?
-    // check if this is actually necessary, and whether it affects the level meters
+    // pass the received list onwards, now containing client channel IDs
 
     emit ConClientListMesReceived ( vecChanInfo );
 }
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -286,7 +286,10 @@ void CClient::OnMuteStateHasChangedReceived ( int iServerChanID, bool bIsMuted )
     // map iChanID from server channel ID to client channel ID
     int iChanID = FindClientChannel ( iServerChanID, false );
 
-    emit MuteStateHasChangedReceived ( iChanID, bIsMuted );
+    if ( iChanID != INVALID_INDEX )
+    {
+        emit MuteStateHasChangedReceived ( iChanID, bIsMuted );
+    }
 }
 
 void CClient::OnCLChannelLevelListReceived ( CHostAddress InetAddr, CVector<uint16_t> vecLevelList )
@@ -321,8 +324,9 @@ void CClient::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo )
             // server channel ID of this entry
             const int iServerChannelID = vecChanInfo[i].iChanID;
 
-            // find matching client channel ID, creating new if necessary
-            const int iClientChannelID = FindClientChannel ( iServerChannelID, true );
+            // find matching client channel ID, creating new if necessary,
+            // update channel number to be client-side
+            vecChanInfo[i].iChanID = FindClientChannel ( iServerChannelID, true );
 
             // discard any lower server channels that are no longer in our local list
             while ( iSrvIdx < iServerChannelID )
@@ -337,10 +341,8 @@ void CClient::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo )
                 iSrvIdx++;
             }
 
-            // now should have matching server channel IDs
-            vecChanInfo[i].iChanID = iClientChannelID; // update channel number to be client-side
-            i++;                                       // next list entry
-            iSrvIdx++;                                 // next local server channel
+            i++;       // next list entry
+            iSrvIdx++; // next local server channel
         }
 
         // have now run out of active channels, discard any remaining from our local list
@@ -1476,7 +1478,7 @@ void CClient::ClearClientChannels()
         clientChannelIDs[i] = INVALID_INDEX;
     }
 
-    qInfo() << "> Client channel list cleared";
+    // qInfo() << "> Client channel list cleared";
 }
 
 void CClient::FreeClientChannel ( const int iServerChannelID )
@@ -1497,10 +1499,12 @@ void CClient::FreeClientChannel ( const int iServerChannelID )
 
     iActiveChannels -= 1;
 
+    /*
     qInfo() << qUtf8Printable ( QString ( "> Freed client ch %1 for server ch %2; chan count = %3" )
                                     .arg ( iClientChannelID )
                                     .arg ( iServerChannelID )
                                     .arg ( iActiveChannels ) );
+     */
 }
 
 // find, and optionally create, a client channel for the supplied server channel ID
@@ -1544,10 +1548,12 @@ int CClient::FindClientChannel ( const int iServerChannelID, const bool bCreateI
 
                 iActiveChannels += 1;
 
+                /*
                 qInfo() << qUtf8Printable ( QString ( "> Alloc client ch %1 for server ch %2; chan count = %3" )
                                                 .arg ( iClientChannelID )
                                                 .arg ( iServerChannelID )
                                                 .arg ( iActiveChannels ) );
+                 */
 
                 return iClientChannelID; // new client channel ID
             }
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -300,9 +300,12 @@ void CClient::OnMuteStateHasChangedReceived ( int iServerChanID, bool bIsMuted )
 
 void CClient::OnCLChannelLevelListReceived ( CHostAddress InetAddr, CVector<uint16_t> vecLevelList )
 {
-    // TODO reorder levels from server channel order to client channel order
+    // reorder levels from server channel order to client channel order
 
-    emit CLChannelLevelListReceived ( InetAddr, vecLevelList );
+    if ( ReorderLevelList ( vecLevelList ) )
+    {
+        emit CLChannelLevelListReceived ( InetAddr, vecLevelList );
+    }
 }
 
 void CClient::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo )
@@ -1562,4 +1565,61 @@ int CClient::FindClientChannel ( const int iServerChannelID, const bool bCreateI
     }
 
     return INVALID_INDEX;
+}
+
+bool CClient::ReorderLevelList ( CVector<uint16_t>& vecLevelList )
+{
+    QMutexLocker locker ( &MutexChannels );
+
+    // vecLevelList is sent from server ordered by server channel ID
+    // re-order it by client channel ID before passing up to the GUI
+    // the list is passed in by reference and modified in situ
+
+    // check it is the right length
+    if ( vecLevelList.Size() != iActiveChannels )
+    {
+        return false; // tell caller to ignore it
+    }
+
+    int iClientCh;
+    int iServerCh = 0;
+
+    // fetch levels by server channel ID
+
+    for ( int i = 0; i < iActiveChannels; i++ )
+    {
+        // find next active server channel
+        while ( iServerCh < MAX_NUM_CHANNELS )
+        {
+            iClientCh = clientChannelIDs[iServerCh++];
+
+            if ( iClientCh != INVALID_INDEX )
+            {
+                clientChannels[iClientCh].level = vecLevelList[i];
+                break;
+            }
+        }
+    }
+
+    // store levels by client channel ID
+
+    iClientCh = 0;
+
+    for ( int i = 0; i < iActiveChannels; i++ )
+    {
+        while ( iClientCh < MAX_NUM_CHANNELS )
+        {
+            uint16_t level = clientChannels[iClientCh].level;
+
+            iServerCh = clientChannels[iClientCh++].iServerChannelID;
+
+            if ( iServerCh != INVALID_INDEX )
+            {
+                vecLevelList[i] = level;
+                break;
+            }
+        }
+    }
+
+    return true; // tell caller to emit signal with new list
 }
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -1574,6 +1574,9 @@ int CClient::FindClientChannel ( const int iServerChannelID, const bool bCreateI
 //
 // The list size is checked against the current number of active channels to guard against
 // any unexpected temporary mismatch in size due to potential out-of-order message delivery.
+//
+// This function returns true if the list has been processed and should be passed on,
+// or false if it was the wrong size and should be discarded.
 
 bool CClient::ReorderLevelList ( CVector<uint16_t>& vecLevelList )
 {
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -307,6 +307,7 @@ void CClient::OnCLChannelLevelListReceived ( CHostAddress InetAddr, CVector<uint
 void CClient::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo )
 {
     // TODO translate from server channel IDs to client channel IDs
+    // ALSO here is where we allocate and free client channels as required
 
     // Upon receiving a new client list, we have to reset oldGain and newGain
     // entries for unused channels. This ensures that a disconnected channel
@@ -331,6 +332,8 @@ void CClient::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo )
             oldGain[iId] = newGain[iId] = 1;
         }
     }
+
+    // TODO vecChanInfo needs to be ordered by client channel ID instead of server channel ID
 
     emit ConClientListMesReceived ( vecChanInfo );
 }
@@ -444,7 +447,7 @@ void CClient::SetRemoteChanGain ( const int iId, const float fGain, const bool b
     // here the timer was not active:
     // send the actual gain and reset the range of channel IDs to empty
     oldGain[iId] = newGain[iId] = fGain;
-    Channel.SetRemoteChanGain ( iId, fGain );
+    Channel.SetRemoteChanGain ( iId, fGain );   // TODO translate client channel to server channel ID
 
     StartDelayTimer();
 }
@@ -460,7 +463,7 @@ void CClient::OnTimerRemoteChanGain()
         {
             // send new gain and record as old gain
             float fGain = oldGain[iId] = newGain[iId];
-            Channel.SetRemoteChanGain ( iId, fGain );
+            Channel.SetRemoteChanGain ( iId, fGain );   // TODO translate client channel to server channel ID
             bSent = true;
         }
     }
@@ -489,6 +492,11 @@ void CClient::StartDelayTimer()
     {
         TimerGain.start ( iCurPingTime * 2 );
     }
+}
+
+void CClient::SetRemoteChanPan ( const int iId, const float fPan )
+{
+    Channel.SetRemoteChanPan ( iId, fPan ); // TODO translate client channel to server channel ID
 }
 
 bool CClient::SetServerAddr ( QString strNAddr )
@@ -868,12 +876,14 @@ void CClient::OnControllerInMuteMyself ( bool bMute )
 
 void CClient::OnClientIDReceived ( int iChanID )
 {
+    // TODO allocate and map client-side channel 0
+
     // for headless mode we support to mute our own signal in the personal mix
     // (note that the check for headless is done in the main.cpp and must not
     // be checked here)
     if ( bMuteMeInPersonalMix )
     {
-        SetRemoteChanGain ( iChanID, 0, false );
+        SetRemoteChanGain ( iChanID, 0, false );    // TODO this will need client channel ID
     }
 
     emit ClientIDReceived ( iChanID );
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -894,6 +894,9 @@ void CClient::Start()
     // init object
     Init();
 
+    // initialise client channels
+    ClearClientChannels();
+
     // enable channel
     Channel.SetEnable ( true );
 
@@ -1421,10 +1424,12 @@ void CClient::ClearClientChannels()
     QMutexLocker locker ( &MutexChannels );
 
     iActiveChannels = 0;
+    iJoinSequence   = 0;
 
     for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
     {
         clientChannels[i].iServerChannelID = INVALID_INDEX;
+        clientChannels[i].iJoinSequence    = 0;
         // TODO reset any other fields in CClientChannel
 
         clientChannelIDs[i] = INVALID_INDEX;
@@ -1448,6 +1453,11 @@ void CClient::FreeClientChannel ( const int iServerChannelID )
     clientChannels[iClientChannelID].iServerChannelID = INVALID_INDEX;
 
     iActiveChannels -= 1;
+
+    qInfo() << qUtf8Printable ( QString ( "> Freed client ch %1 for server ch %2; chan count = %3" )
+                                    .arg ( iClientChannelID )
+                                    .arg ( iServerChannelID )
+                                    .arg ( iActiveChannels ) );
 }
 
 // find, and optionally create, a client channel for the supplied server channel ID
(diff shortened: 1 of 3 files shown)


src/client.cpp
```diff
@@ -291,31 +281,90 @@ void CClient::OnNewConnection()
...
+                FreeClientChannel ( iSrvIdx );
+            }
+            iSrvIdx++;
+        }
+
+        Q_ASSERT ( iActiveChannels == iNumConnectedClients );
     }
+
+    // TODO Does vecChanInfo need to be ordered by client channel ID instead of server channel ID?
+    // check if this is actually necessary, and whether it affects the level meters
```
softins, line 365: It doesn't actually appear to be necessary. The updating of channel IDs in-place seems to be sufficient. But I'd be interested in others' opinions.

src/client.cpp
```diff
@@ -1386,3 +1459,158 @@ int CClient::EstimatedOverallDelay ( const int iPingTimeMs )
...
+
+    int iClientCh;
+    int iServerCh = 0;
+
+    // fetch levels by server channel ID
+
+    for ( int i = 0; i < iActiveChannels; i++ )
+    {
+        // find next active server channel
+        while ( iServerCh < MAX_NUM_CHANNELS )
```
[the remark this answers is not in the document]
ann0see: Maybe we could find all "holes" first and then fill them one by one.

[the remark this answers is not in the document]
softins: I've put a lot of thought into this. The level list is received in ascending order of server channel. So when processing it, we first need a single pass through the list, skipping unused server channels, and placing the level value in the corresponding client channel. The we go through the client channels in order, fetching the level values from the active channels and placing them back in the list in order, so it can be passed up to the mixer board in the order it expects. So just two linear passes, and only as far as the number of active channels, and I can't think of any way that could be made more efficient. In practice we only need to skip a few channels, and never need to go all the way up to MAX_NUM_CHANNELS, as we will run out of list entries as soon as we have done all the active channels.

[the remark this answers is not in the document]
softins: Bear in mind that we are not resetting `iServerCh` each time round the outer loop. As the levels are in channel order, we just keep going from where we left off, looking for the next active channel, which will probably be next, or only a small number of steps away.

src/client.cpp
```diff
@@ -291,31 +281,90 @@ void CClient::OnNewConnection()
...
-    for ( int iId = 0; iId < MAX_NUM_CHANNELS; iId++ )
+    const int iNumConnectedClients = vecChanInfo.Size();
+    int       i, iSrvIdx;
+
+    // on a new connection, a server sends an empty channel list before sending
+    // the real channel list (see #1010). To avoid this discarding "our" channel
+    // that we have just created, we skip this processing and just pass the empty
+    // list to the emitted signal.
+
+    if ( iNumConnectedClients != 0 )
```
pljones, line 315: I'm not keen on a long indented block like this with two loops.  It's code complexity itches :).

Can we return early?  Or refactor to two called methods (inline maybe)?

src/client.cpp
pljones: So long as the "No fader sort" continues to add newly joining clients to the right-hand end of the mixer GUI, it's working okay.  If a newly joining client jolts anyone already on rightwards and slots in to the left, that's not okay.

src/client.cpp
```diff
@@ -849,8 +907,20 @@ void CClient::OnControllerInMuteMyself ( bool bMute )
     emit ControllerInMuteMyself ( bMute );
 }
 
-void CClient::OnClientIDReceived ( int iChanID )
+void CClient::OnClientIDReceived ( int iServerChanID )
 {
+    // if we have just connected to a running server, iActiveChannels will be 0
+    // if iActiveChannels is not 0, the server must have been restarted on the fly
+    // in that case, channels might have changed, so clear our list to get it afresh.
```
pljones, line 914: Does this new code prevent the old trashing of "No fader sort" by this event?  Given the server channel id isn't used, the "no sort order" should remain stable, too?

src/client.cpp
```diff
@@ -1386,3 +1459,158 @@ int CClient::EstimatedOverallDelay ( const int iPingTimeMs )
...
+
+                return iClientChannelID; // new client channel ID
+            }
+        }
+    }
+
+    return INVALID_INDEX;
+}
+
+bool CClient::ReorderLevelList ( CVector<uint16_t>& vecLevelList )
```
pljones, line 1561: Could do with a comment here saying why this exists, I think.

src/client.cpp
softins: I think it's short enough that it can still be comprehended easily.

We could do an early return, but then we'd have to duplicate `emit ConClientListMesReceived ( vecChanInfo );` before the early return, which smells to me worse than the conditional block.

Refactoring to multiple methods means having to jump about between functions to follow the code. There is already too much of that within Jamulus!

src/client.cpp
softins: I don't think it will be guaranteed to be any better. It will depend on the order that the other clients appear back on the server, and thereby appear in a client list sent to us by that server. If we receive them in a different order from what they were, they might get recreated in a different order.

I don't think it will be any worse though.

src/client.cpp
pljones: And we don't want to expose the `if` logic to the caller.  OK, leave as is.

src/client.cpp
```diff
@@ -1558,6 +1558,24 @@ int CClient::FindClientChannel ( const int iServerChannelID, const bool bCreateI
...
+// passing the client list up to the mixer board, it is also necessary to re-order the values
+// in the level list so that they are in order of mapped client channel ID.
+// This function performs that re-ordering by scanning the server channels in order, once,
+// for active channels, and storing the level value in the corresponding client channel.
+// Then the function scans the client channels in order, fetching the level values and putting
+// them back into vecLevelList in order of client channel. The mixer board will then display
+// the levels against the correct channels.
+//
+// The list size is checked against the current number of active channels to guard against
+// any unexpected temporary mismatch in size due to potential out-of-order message delivery.
```
pljones, line 1577: Plus why is it boolean?

src/client.cpp
softins: Yes, I've just done some testing with "No fader sort". I added several clients, and then disconnected some to leave gaps in the used channel numbers. Re-adding them used the channel numbers that were vacated, but the faders were definitely added on the right in order of arrival.

softins: In view of that, I'll remove the above comment from the code.

src/client.cpp
softins: Because if the list size is mismatched, we don't want to emit the received, incorrect list onwards in `CClient::OnCLChannelLevelListReceived()` (line 298). The next list to arrive will probably be fine.

src/client.cpp
```diff
@@ -1386,3 +1458,179 @@ int CClient::EstimatedOverallDelay ( const int iPingTimeMs )
...
+                                                .arg ( iClientChannelID )
+                                                .arg ( iServerChannelID )
+                                                .arg ( iActiveChannels ) );
+
+                return iClientChannelID; // new client channel ID
+            }
+        }
+    }
+
+    return INVALID_INDEX;
```
pljones, line 1557: Are there any _normal_ reasons to get here?  I'd log again...

Either we expect the channel to exist, so do not request one created -- so INVALID_INDEX is very unexpected.

Or we did expected the channel might not exist, so we requested a new one -- but we got more server channels than the client knows about, so INVALID_INDEX is highly unlikely (as we generally have newer clients than servers, I'd think).

src/client.cpp
```diff
@@ -1386,3 +1458,179 @@ int CClient::EstimatedOverallDelay ( const int iPingTimeMs )
...
+
+                clientChan->oldGain = clientChan->newGain = 1.0f;
+
+                clientChan->level = 0;
+
+                clientChannelIDs[iServerChannelID] = iClientChannelID;
+
+                iActiveChannels += 1;
+
+                qInfo() << qUtf8Printable ( QString ( "> Alloc client ch %1 for server ch %2; chan count = %3" )
```
pljones, line 1547: Probably don't want this except for debugging - a busy server will cause a client to be issuing log messages rather a lot...

pljones:
OK, I'm happy with the change and wouldn't object to it as is, even without the logging amended.

src/client.cpp
```diff
@@ -1386,3 +1458,179 @@ int CClient::EstimatedOverallDelay ( const int iPingTimeMs )
...
+    qInfo() << "> Client channel list cleared";
+}
+
+void CClient::FreeClientChannel ( const int iServerChannelID )
+{
+    QMutexLocker locker ( &MutexChannels );
+
+    if ( iServerChannelID == INVALID_INDEX || iServerChannelID >= MAX_NUM_CHANNELS )
+    {
+        return;
```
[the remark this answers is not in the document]
softins: It's mainly defensive programming before using `iServerChannelID` to index an array. It should never happen.

src/client.cpp
```diff
@@ -1386,3 +1458,179 @@ int CClient::EstimatedOverallDelay ( const int iPingTimeMs )
...
+
+// find, and optionally create, a client channel for the supplied server channel ID
+// returns a client channel ID or INVALID_INDEX
+int CClient::FindClientChannel ( const int iServerChannelID, const bool bCreateIfNew )
+{
+    QMutexLocker locker ( &MutexChannels );
+
+    if ( iServerChannelID == INVALID_INDEX || iServerChannelID >= MAX_NUM_CHANNELS )
+    {
+        return INVALID_INDEX;
```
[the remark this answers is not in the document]
softins: Ditto, as above.

src/client.cpp
softins: The second parameter to the function specifies whether to create a new client channel or not. There are a few places where `FindClientChannel` is called with `false`, in which case we want the `INVALID_INDEX` return for an unallocated channel.

src/client.cpp
softins: Now that I'm satisfied with the code operation, I'm happy to comment these out, or to change them to `qDebug()`. In the latter case, they would be output from an executable that was built with `make debug`, but together with a LOT of noise from the Qt graphics code!

src/client.cpp
```diff
@@ -291,31 +281,89 @@ void CClient::OnNewConnection()
     //### TODO: END ###//
 }
 
-void CClient::OnConClientListMesReceived ( CVector<CChannelInfo> vecChanInfo )
+void CClient::OnMuteStateHasChangedReceived ( int iServerChanID, bool bIsMuted )
+{
+    // map iChanID from server channel ID to client channel ID
+    int iChanID = FindClientChannel ( iServerChanID, false );
+
+    emit MuteStateHasChangedReceived ( iChanID, bIsMuted );
```
pljones, line 289: This causes
```cpp
 vecpChanFader[static_cast<size_t> ( iChannelIdx )]->SetRemoteFaderIsMute ( bIsMute );
```
so if `iChanID` is `INVALID_INDEX`, the client will crash.

src/client.cpp
```diff
@@ -291,31 +281,89 @@ void CClient::OnNewConnection()
...
     {
-        if ( !bChanIdInUse[iId] )
+        // this relies on the received client list being in order of server channel ID
+
+        for ( i = 0, iSrvIdx = 0; i < iNumConnectedClients && iSrvIdx < MAX_NUM_CHANNELS; )
         {
-            // reset oldGain and newGain as this channel id is currently unused and will
-            // start with a server-side gain at 1 (100%) again.
-            oldGain[iId] = newGain[iId] = 1;
+            // server channel ID of this entry
```
pljones, line 321:
```suggestion
            // current "server channel ID" of this entry
```

src/client.cpp
```diff
@@ -291,31 +281,89 @@ void CClient::OnNewConnection()
...
+            const int iServerChannelID = vecChanInfo[i].iChanID;
+
+            // find matching client channel ID, creating new if necessary
+            const int iClientChannelID = FindClientChannel ( iServerChannelID, true );
```
pljones, line 325: I think I'd be happier if this were `iChanID` -- we're going to store it in `vecChanInfo[i].iChanID` later.

Talking about `vecChanInfo[i].iChanID` as `iServerChannelID` and `iClientChannelID` is a bit confusing.

pljones: In fact, between here and line 341, I don't see `i` or `iClientChannelID` or `vecChanInfo` changing -- why can't it be stored now?  (Presumably `FindClientChannel` is hiding something that should be visible here somehow.)

src/client.cpp
softins: Good catch! Highly unlikely in practice, but we shouldn't allow it. I'll protect the `emit` with a condition.

src/client.cpp
softins: Yes, in fact we don't need `iClientChannelID` at all. Line 325 could just become:

```
vecChanInfo[i].iChanID = FindClientChannel ( iServerChannelID, true );
```

With a suitably-expanded comment...

pljones:
**NOTE:** I'm happy with the scope of this PR.  The following outlines my thoughts - I'll raise a feature request shortly after this gets merged.

OK, just thinking about the wider implications...

> When talking to a server version 3.5.5 or newer, the client is told its own server channel, and will always map
this to client channel 0. This allows the user of a MIDI device or JSON-RPC client to know that channel 0 will always control
the user's own audio channel.

So we will have _inconsistent client behaviour_ based on server version, with nothing explanatory for the user to understand why.

I don't like that.  I'd rather we made the default act as if we hadn't received the "own server channel" message and only enable remapping based on capture if:
- GUI option "Own Fader First" is selected
- _NEW_ command line client-only option `--channelsort` has `own` as an argument (see below for more)

This way we can make the GUI option disabled when the server does not supply "own server channel" and document the effect of the new CLI option.

The new CLI option should additionally support one of "name", "instrument", "group", "city" and "channel".  Where none of these appears, behaviour must be as for GUI "No user sorting" (assuming no preference - e.g. headless client).  Otherwise, the channel list should be sorted accordingly.

`--ctrlmidich` would be used with "--channelsort own;channel" to get the proposed effect -- whilst making it clear this depended on the server supplying the relevant information.

softins:
> > When talking to a server version 3.5.5 or newer, the client is told its own server channel, and will always map
> > this to client channel 0. This allows the user of a MIDI device or JSON-RPC client to know that channel 0 will always control
> > the user's own audio channel.
> 
> So we will have _inconsistent client behaviour_ based on server version, with nothing explanatory for the user to understand why.

I don't think that's a problem. There are already differences in client behaviour depending whether it receives the CLIENT_ID message. When talking to a server < 3.5.5, I think the following features will work differently or not at all:

* Mute myself
* "Own fader first" setting

> I don't like that. I'd rather we made the default act as if we had _not_ received the "own server channel" message

I don't see what problem that is trying to solve, beyond the notion of consistency with archaic server versions, and I don't think it gives any advantage to the user. Version 3.5.5 was released on _26 May 2020_.

Looking in explorer at the moment, under _Any Genre 1_, 7 out of 108 servers are older than 3.5.5, and only 12 are older than 3.6.2. The oldest on _Any Genre 2_ is 3.6.1, of which there are 2 out of 51; the rest are 3.6.2 or newer. All the other directories only accept 3.6.2 or newer servers anyway.

Anyone running a private server but using a recent version of client is extremely unlikely to be using a really old version of server.

So changing the default as suggested would just be degrading the operation of the client for consistency with a minuscule set of situations. The user doesn't get to choose the channels assigned by the server, so they also don't need to choose whether their own channel gets client channel 0 or not. It's no disadvantage to the user to consistently get channel 0 for themselves by default in the overwhelming majority of cases.

I would much rather it became a non-issue by having _all_ the public directories only accept registrations from servers of 3.6.2 or newer.

Also, if we incorporate #3416, then the user would see the server versions in the Connect dialog, and could decide not to connect to an archaic version of server.

softins:
> Of course, this pulls the channel _sorting logic_ into the client itself, so that the GUI simply displays the channels in the order held by the client, which further reduces logic in the GUI code.

I think the client layer is the wrong place to do any sorting. We don't want to be changing channel assignments on the fly.

Sorting is a _presentational_ function, and so the mixer board is the right place to do it, as at present.

> (Yes, there should also be a JSON-RPC interface to set the sort order.)

The JSON-RPC interface will/should receive all the same channel info that the mixer board does. A user interface that gets its info via JSON-RPC can therefore do its own sorting for presentation to the user, based on that information and the user's settings (which don't need to be sent to the Jamulus client).

ann0see:
> Sorting is a presentational function, and so the mixer board is the right place to do it, as at present.

Yes, but it's not an UI only feature either.

> A user interface that gets its info via JSON-RPC can therefore do its own sorting for presentation to the user, based on that information and the user's settings 

I think it is confusing if JSON-RPC and GUI show different ordering.

ann0see:
> I don't think that's a problem. There are already differences in client behaviour depending whether it receives the CLIENT_ID message.

I agree - this time. It's not a huge problem.

src/client.cpp
```diff
@@ -849,8 +908,20 @@ void CClient::OnControllerInMuteMyself ( bool bMute )
...
 
-void CClient::OnClientIDReceived ( int iChanID )
+void CClient::OnClientIDReceived ( int iServerChanID )
 {
+    // if we have just connected to a running server, iActiveChannels will be 0
+    // if iActiveChannels is not 0, the server must have been restarted on the fly
+    // in that case, channels might have changed, so clear our list to get it afresh.
+    if ( iActiveChannels != 0 )
+    {
+        qInfo() << "> Server restarted?";
```
ann0see, line 918: Isn't this better qDebug()?

src/client.cpp
```diff
@@ -1386,3 +1460,183 @@ int CClient::EstimatedOverallDelay ( const int iPingTimeMs )
...
+
+    for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
+    {
+        clientChannels[i].iServerChannelID = INVALID_INDEX;
+        // all other fields will be initialised on channel allocation
+
+        clientChannelIDs[i] = INVALID_INDEX;
+    }
+
+    // qInfo() << "> Client channel list cleared";
```
ann0see, line 1481: Probably worth getting rid of this debug output.

src/client.cpp
```diff
@@ -1386,3 +1460,183 @@ int CClient::EstimatedOverallDelay ( const int iPingTimeMs )
...
+// for active channels, and storing the level value in the corresponding client channel.
+// Then the function scans the client channels in order, fetching the level values and putting
+// them back into vecLevelList in order of client channel. The mixer board will then display
+// the levels against the correct channels.
+//
+// The list size is checked against the current number of active channels to guard against
+// any unexpected temporary mismatch in size due to potential out-of-order message delivery.
+//
+// This function returns true if the list has been processed and should be passed on,
+// or false if it was the wrong size and should be discarded.
```
ann0see, line 1585: Not part of this PR, but I'd like to have a standard format like in the recorder of JSON-RPC server for documentation like this

ann0see (commented):
Otherwise seems ok (test on Debian 12)

src/client.cpp
softins: No, I don't think so. The user might want to know that the server restarted, as it may well not be his server. `qDebug()` output is normally suppressed except for debug builds, and even then needs to be specifically enabled. When doing that, it also gets swamped with a lot of output from the Qt graphics code - I tried it!

src/client.cpp
softins: I thought it would be useful to retain the `qInfo()` messages I used while debugging, just commented out, then someone wanting to re-enable them only needs to uncomment, not compose afresh or search earlier commits for them.

softins:
> > Sorting is a presentational function, and so the mixer board is the right place to do it, as at present.
> 
> Yes, but it's not an UI only feature either.
> 
> > A user interface that gets its info via JSON-RPC can therefore do its own sorting for presentation to the user, based on that information and the user's settings
> 
> I think it is confusing if JSON-RPC and GUI show different ordering.

Well let's have this conversation in any follow-up issue, as it's not relevant to this PR.

softins:
I don't think there is any reason to delay this PR until 4.0.0. It can very usefully be merged now.

src/client.cpp
pljones: Agreed, this should be `qInfo` as it's something "unexpected in normal operation" but not necessarily a problem (which the user would need warning about).

ann0see: Ok.

ann0see (approved):

PR #3429, opened 2024-11-26
Apply rate limiting to Pan as well as Gain

### LINKED ISSUE #3428: Pan change messages need to be rate-limited similarly to Gain.
**Describe the bug**

Issue #2492 reported a lag when many Gain change messages were generated, e.g. by a MIDI controller. This was fixed in #2535 by using a rate-limiting timer.

I have now observed exactly the same behaviour with Pan change messages, which is not surprising, although not mentioned in the issue above nor addressed in the PR. Adjustments to Pan are much less frequent in practice, which is probably why it has never been reported.

**To Reproduce**

Start Jamulus with a MIDI controller such as Korg NanoKontrol2, connect to a distant server, and move the Pan control for a channel by a large amount. See #2492 for more details of the testing that was done for faders, including packet tracing.

**Expected behavior**

Pan change messages need to be rate-limited in the same way as Gain-change messages.

**Screenshots**

**Operating system**

Any. Testing was done on Raspberry Pi.

**Version of Jamulus**

All versions up to 3.11.0 inclusive.

**Additional context**

softins:

This PR applies the same rate-limiting to Pan change message as is currently done for Gain change messages since #2535

CHANGELOG: Client: Apply rate-limiting to Pan change messages to eliminate lag in their effect.

**Context: Fixes an issue?**

Fixes #3428

No, it's just a performance improvement.

Tested and ready for review

Nothing

## Checklist


ann0see (approved):
Looks sensible and with a manual test (moving pan pots around quickly), it seems ok: No delay or any oddities visible.

src/client.h
```diff
@@ -255,8 +256,8 @@ class CClient : public QObject
     void SetMuteOutStream ( const bool bDoMute ) { bMuteOutStream = bDoMute; }
 
     void SetRemoteChanGain ( const int iId, const float fGain, const bool bIsMyOwnFader );
-    void OnTimerRemoteChanGain();
-    void StartDelayTimer();
+    void OnTimerRemoteChanGainOrPan();
```
pljones, line 259: It would make sense to have `SetRemoteChanPan` next to `SetRemoteChanGain` and before their shared timer methods.

src/client.cpp
```diff
@@ -500,37 +501,69 @@ void CClient::OnTimerRemoteChanGain()
...
+        {
+            // send new pan and record as old pan
+            float fPan = clientChan->oldPan = clientChan->newPan;
+            Channel.SetRemoteChanPan ( clientChan->iServerChannelID, fPan ); // translate client channel to server channel ID
+            bSent = true;
+        }
     }
 
-    // if a new gain has been sent, reset the range of channel IDs to empty and start timer
+    // if a new gain or pan has been sent, reset the range of channel IDs to empty and start timer
```
pljones, line 514: Doesn't using a single timer mean that sending a single gain and then a single pan could get twice the delay?  Does it matter?  I guess it saves complexity (at the expense of a little readability).

pljones: I guess an X-Y controller mapped to gain/pain could be very noisy, so this does probably work better.

pljones (approved):
No blockers.

src/client.cpp
softins: No, it shouldn't. When the single gain has been sent, the timer is started. If a single pan is requested while the timer is running, the new value will get stored locally and then sent as soon as the timer expires. The timer is by default 50ms if ping is less than 25ms, so pretty quick, otherwise it is increased to twice the current ping time.

I didn't feel there was any advantage in having separate gain and pan timers, as almost all users will not be twiddling both at the same time.

src/client.h
softins: Yes, good point. Done.

PR #3431, opened 2024-11-28
Add Windows native MIDI support

softins:

Add native support for MIDI devices to the ASIO (non-Jack) build for Windows. This will allow users to use
a MIDI fader bank (such as the _Korg nanoKontrol2_) without having to set up the Jack version of Jamulus.

CHANGELOG: Client: Added native MIDI support to the ASIO (non-Jack) Windows build

**Context: Fixes an issue?**

Until now, only Mac CoreAudio and Jack (for Windows or Linux) had support for MIDI fader banks.

Yes. The web page detailing how to use a fader bank needs to mention that Windows users no longer need the Jack version
to get MIDI support. It also needs to describe the `d` option for `--ctrlmidich`.

Working and ready for testing.

~Need to consider how to support systems with more than one MIDI device. Either add an option to `--ctrlmidich` to
specify the Windows MIDI Input device number (defaults to `0`), or else enumerate all connected MIDI devices
and listen to them all (I think the Mac version does the latter).~

## Checklist


softins:
Checked the original Mac version with multiple MIDI devices connected, and it does listen to them all. Updated this PR to do the same.

src/sound/asio/sound.cpp
```diff
@@ -1218,3 +1235,75 @@ int64_t CSound::Flip64Bits ( const int64_t iIn )
...
+    /* Get the number of MIDI In devices in this computer */
+    iMidiDevs = midiInGetNumDevs();
+    if ( iMidiDevs > MAX_MIDI_DEVS )
+    {
+        iMidiDevs = MAX_MIDI_DEVS;
+    }
+
+    // printf("Found %d MIDI device%s\n", iMidiDevs, iMidiDevs == 1 ? "" : "s");
+
+    // open all connected MIDI devices and set the callback function to handle incoming messages
```
pljones, line 1251: Maybe we should have some way to specify the MIDI device, rather than just opening them all?  This approach would likely cause me problems - I want Reaper to have exclusive control of my hardware (audio and MIDI) and use its virtual routing capabilities (or I'd use JACK the same way).  We don't do this for audio - i.e. greedily grab every audio device.

src/sound/asio/sound.h
```diff
@@ -134,4 +134,14 @@ class CSound : public CSoundBase
     static long      asioMessages ( long selector, long value, void* message, double* opt );
 
     char* cDriverNames[MAX_NUMBER_SOUND_CARDS];
+
+    // Windows native MIDI
+#define MAX_MIDI_DEVS 4
```
pljones, line 139: This isn't enough.  Make is 128?

pljones:
Would it be better not to put this in `asio` -- using MIDI with JACK isn't that nice, to be honest, and I'd take an alternative if it were offered...  (You don't get to configure which ports it opens, unlike it Reaper.)

src/sound/asio/sound.cpp
```diff
@@ -1218,3 +1235,75 @@ int64_t CSound::Flip64Bits ( const int64_t iIn )
...
+
+void CALLBACK CSound::MidiCallback ( HMIDIIN hMidiIn, UINT wMsg, DWORD_PTR dwInstance, DWORD_PTR dwParam1, DWORD_PTR dwParam2 )
+{
+    Q_UNUSED ( hMidiIn );
+    Q_UNUSED ( dwInstance );
+    Q_UNUSED ( dwParam2 );
+
+    if ( wMsg == MIM_DATA )
+    {
+        BYTE status = dwParam1 & 0xFF;
```
ann0see, line 1296: I'd like to see a comment here what the mask does and why it works. Maybe a link to midi specifications or similar.

ann0see (changes requested):

src/sound/asio/sound.cpp
softins: Comments added, see be04d89

src/sound/asio/sound.cpp
softins: I was going by what is already done for CoreAudio on Mac:

https://github.com/jamulussoftware/jamulus/blob/bedd7b0458d4669bb6309ed19a0f5858106fa2a4/src/sound/coreaudio-mac/sound.cpp#L72-L79

And I verified on my Mac that it did indeed hear messages from multiple devices (Korg nanoKontrol2 and Behringer UMC404HD with Midi connected to my piano). So I attempted to be consistent with that behaviour.

The problem we have is that at the Windows API level, device numbers are arbitrary, and depend on the order the devices were connected, or if left connected, the specific ports they are connected to and the order Windows finds those ports on startup. So providing some sort of selection parameter in a predictable way is tricky. It might be possible to use `midiInGetDevCaps()` to fetch the product name for each found device, such as `nanoKONTROL2` or `UMC404HD 192k`, and add command-line or even GUI settings to include or exclude devices based on that (no idea what happens if someone has two devices of the same type). But this is way more complexity than I had envisaged for this PR, and I would prefer that kind of nuance was deferred to a follow-up PR. It also requires the user to actually know what identifying string the device uses.

I have no experience with Reaper or ReaRoute, but I would be very interested to hear if my current approach causes problems with your setup in practice. If it does, we can work out how to address them, but I'd be reluctant to add a lot of complexity "just in case".

src/sound/asio/sound.h
softins: We could certainly make it 8 or 16 if necessary, but I can't imagine a use case for more than that, can you? It refers to actual MIDI _interface_ devices, such as a USB device or a card with MIDI ports, not to the number of MIDI devices connected to a MIDI interface (even that would be a modest number, surely, and at the MIDI level devices are distinguished by channel 1-16).

We need some kind of value for sizing the array of handles and limiting the loop index.

softins:
> Would it be better not to put this in `asio` -- using MIDI with JACK isn't that nice, to be honest, and I'd take a better alternative if it were offered... (You don't get to configure which ports it opens, unlike it Reaper.)

I did wonder about that, but since the coreaudio and jack implementations of MIDI input were just part of `CSound`, I followed the same principle for the non-jack Windows port in the `asio` directory.

My focus and scope here was basically to fill the omission of MIDI support in the regular non-jack Windows build.

> So I'd like something like `--winmidi` available on both Jack and ASIO Windows builds that allowed specifying behaviour. On the JACK build, it would allow `jack` as an argument, to defer to the existing Jack implementation (which would be the default on that build). On both builds, it would allow `list` to list available port names (and exit, like `--help` or `--version`). Both builds would also take one (or more?) of these names, which would then be used to filter the available ports in this implementation (rather than greedily open all ports).

These are all good ideas, but I would see them as a follow-up PR rather than holding up this one, as they do add quite a bit of complexity, which I'd like to defer or delegate!

src/sound/asio/sound.cpp
pljones: I use MIDI drums and keyboards.  Those go to Reaper.

If MME has exclusively assigned them to Jamulus, Reaper won't get the notes and I won't be able to use Jamulus.

softins: Have you experimented with the PR build to see whether MME does assign the MIDI device exclusively to Jamulus or shared?

pljones: If I open Reaper first, Jamulus gets no MIDI data.

src/sound/asio/sound.h
softins: Working on it... update tomorrow

softins:
@pljones please see the latest commits, which hopefully address most or all of your concerns.

- Moved the MIDI code into a separate class and directory, so it could be reused in the future, such as for the Jack build.
- Added output to list the found MIDI devices and their names (only happens when `--ctrlmidich` is specified).
- Added a `d` option to `--ctrlmidich` to specify which MIDI device should be used, ignoring all the others. It takes a string argument that must match the devicename of the desired device. If the `d` option is not specified, then all found MIDI devices will be opened.
- Removed the arbitrary limit on the number of MIDI devices, by using a `QVector` for the list of opened handles (only used for closing them again).

I found adding the `d` option above to be much more straightforward to implement than adding a completely new command-line option.

Here is the output from testing with different `d` options while two MIDI devices were connected:

```
C:\Users\Tony\Git\jamulus>jamulus-midi.bat

C:\Users\Tony\Git\jamulus>release\Jamulus --ctrlmidich "1;f0*8;p16*8;s32*8;m48*8"
- MIDI controller settings: 1;f0*8;p16*8;s32*8;m48*8
- allocated port number: 22134
- MIDI devices found: 2
  0: nanoKONTROL2
  1: Keystation Mini 32

C:\Users\Tony\Git\jamulus>jamulus-nano.bat

C:\Users\Tony\Git\jamulus>release\Jamulus --ctrlmidich "1;f0*8;p16*8;s32*8;m48*8;dnanoKONTROL2"
- MIDI controller settings: 1;f0*8;p16*8;s32*8;m48*8;dnanoKONTROL2
- allocated port number: 22134
- MIDI devices found: 2
  0: nanoKONTROL2
  1: Keystation Mini 32 (ignored)

C:\Users\Tony\Git\jamulus>jamulus-maudio.bat

C:\Users\Tony\Git\jamulus>release\Jamulus --ctrlmidich "1;f0*8;p16*8;s32*8;m48*8;dKeystation Mini 32"
- MIDI controller settings: 1;f0*8;p16*8;s32*8;m48*8;dKeystation Mini 32
- allocated port number: 22134
- MIDI devices found: 2
  0: nanoKONTROL2 (ignored)
  1: Keystation Mini 32

C:\Users\Tony\Git\jamulus>jamulus-other.bat

C:\Users\Tony\Git\jamulus>release\Jamulus --ctrlmidich "1;f0*8;p16*8;s32*8;m48*8;dSomething else"
- MIDI controller settings: 1;f0*8;p16*8;s32*8;m48*8;dSomething else
- allocated port number: 22134
- MIDI devices found: 2
  0: nanoKONTROL2 (ignored)
  1: Keystation Mini 32 (ignored)
```

pljones:
OK, that worked here.

Jamulus.pro
```diff
@@ -155,8 +155,10 @@ win32 {
             }
             # Important: Keep those ASIO includes local to this build target in
             # order to avoid poisoning other builds license-wise.
-            HEADERS += src/sound/asio/sound.h
+            HEADERS += src/sound/asio/sound.h \
+                src/sound/midi-win/midi.h
```
pljones, line 159: As the comment says, this is meant to the just the ASIO stuff.  Can the MIDI bits be added separately?  Maybe enable "win-midi" by default if "not jack-on-windows" and allow it to be turned on in future?

src/sound/midi-win/midi.cpp
```diff
@@ -0,0 +1,130 @@
+/******************************************************************************\
+ * Copyright (c) 2004-2024
```
pljones, line 2: Well, this probably just wants 2024 :)

pljones (approved):

src/sound/midi-win/midi.cpp
softins: Will update!

pljones:
Also checked starting Reaper first and on a populated server and it works as expected.  I'll grab the Win-Jack build now.

pljones:
Yep, that worked -- of course, it's a bit of a hunt to find the System MIDI In -- sometimes they're labelled `capture_X` -- but sometimes it uses a port name it's got from somewhere...
image
(There's only one MIDI In on my AudioFire4, it's not listed the USB Ins, and I'm pretty sure you can't squeeze MIDI _from_ the MS GS Synth...)
It's another reason I'd like not to use the `-X WinMME` on Jack ☹️...

ann0see (approved):
I don't own a midi device, but it seems good.

PR #3446, opened 2024-12-24
Add profile checkbox to enable/disable tooltips

softins:

Adds a checkbox to the My Profile dialog to enable tooltips to be enabled or disabled. The setting is stored in the `.ini` file. A Qt event filter consumes the tooltip event if tooltips are disabled, preventing the tooltip from being shown.

CHANGELOG: Client: add a checkbox to My Profile to enable/disable tooltips.

**Context: Fixes an issue?**

The PR #3252 adds additional tooltips to the Connect dialog, some of which are quite wordy. While these tooltips will be useful to new or occasional users of Jamulus, they can quickly become annoying to experienced users who have no need of the wordy reminder, particularly within the server list. Allowing tooltips to be disabled caters for both types of user.

Any Wiki page that has a screenshot and description of the the My Profile page should be updated.

Tested and working.

Review

## Checklist

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -84,11 +84,13 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     lbrInputLevelL->setAccessibleName ( strInpLevHAccText );
     lbrInputLevelL->setAccessibleDescription ( strInpLevHAccDescr );
     lbrInputLevelL->setToolTip ( strInpLevHTT );
+    lbrInputLevelL->installEventFilter ( this ); // install event filter for tooltips
     lbrInputLevelL->setEnabled ( false );
     lbrInputLevelR->setWhatsThis ( strInpLevH );
     lbrInputLevelR->setAccessibleName ( strInpLevHAccText );
     lbrInputLevelR->setAccessibleDescription ( strInpLevHAccDescr );
     lbrInputLevelR->setToolTip ( strInpLevHTT );
+    lbrInputLevelR->installEventFilter ( this ); // install event filter for tooltips
     lbrInputLevelR->setEnabled ( false );
 
     // connect/disconnect button
@@ -153,6 +155,7 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
                                 "you will not have much fun using %1." )
                                .arg ( APP_NAME ) +
                            TOOLTIP_COM_END_TEXT );
+    ledDelay->installEventFilter ( this ); // install event filter for tooltips
 
     ledDelay->setAccessibleName ( tr ( "Delay status LED indicator" ) );
 
@@ -184,6 +187,7 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     ledBuffers->setToolTip ( tr ( "If this LED indicator turns red, "
                                   "the audio stream is interrupted." ) +
                              TOOLTIP_COM_END_TEXT );
+    ledBuffers->installEventFilter ( this ); // install event filter for tooltips
 
     ledBuffers->setAccessibleName ( tr ( "Local Jitter Buffer status LED indicator" ) );
 
@@ -1515,4 +1519,16 @@ void CClientDlg::SetPingTime ( const int iPingTime, const int iOverallDelayMs, c
 
     // set current LED status
     ledDelay->SetLight ( eOverallDelayLEDColor );
+}
+
+bool CClientDlg::eventFilter ( QObject* obj, QEvent* event )
+{
+    if ( event->type() == QEvent::ToolTip )
+    {
+        // return true to suppress tooltip, false to allow it
+        return !pSettings->bShowToolTips;
+    }
+
+    // continue with normal processing for other events
+    return QObject::eventFilter ( obj, event );
 }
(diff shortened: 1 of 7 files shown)


ann0see:
Thanks. Is there a reason why it's not an advanced setting?

pljones:
Could you set the project, please, so the PR gets tracked.

pljones:
Ta.

Does this disable tooltips - such as used in the mixer view for profile information popups - everywhere?

softins:
> Does this disable tooltips - such as used in the mixer view for profile information popups - everywhere?

It certainly does so in both the settings dialog and the connect dialog. I didn't check the mixer, as I forgot there were tooltips for profile information. I can do so when I'm next at the computer. I think it probably does. If that's a problem, maybe we can add another check to the filter.

softins:
> Thanks. Is there a reason why it's not an advanced setting?
> I understand that it's UI related and most UI related settings are in my profile. However advanced users would have this disabled, so it would be an advanced setting. 

No particular reason, other than it looked a natural place to put it. I guess the same question could be raised about the Audio Alerts checkbox?

softins:
> Could you set the project, please, so the PR gets tracked.

It appears to be set already? I'm not entirely clear how the project feature works.

src/clientsettingsdlg.cpp
```diff
@@ -397,6 +397,11 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                         "A second sound device may be required to hear the alerts." ) );
     chbAudioAlerts->setAccessibleName ( tr ( "Audio Alerts check box" ) );
 
+    // show tooltips
+    chbShowToolTips->setWhatsThis ( "<b>" + tr ( "Show ToolTips" ) + ":</b> " +
```
ann0see, line 401: I believe it should be "Tooltips" by the style guide. 
@gilgongo ?

gilgongo: The Guide says "Nouns use lower case unless they have a formal definition in this style guide". So it looks like if we're to be consistent with "check box" (for example) we might need to have it as "Show tool tips". 

As an aside, we also use capitals to refer to things unique or in some way special to Jamulus, and not only terms in the Guide (hence "Audio Alerts" for example). Might add that to the Guide in fact. I like to think our slightly unusual use of capitals is inherited from the German origins of Jamulus :-)

softins:
> Does this disable tooltips - such as used in the mixer view for profile information popups - everywhere?

Yes, just tested, and with tooltips disabled, the musician profile popups are not displayed. Do we need to test for this and force just them to be enabled, or provide a separate setting, or leave them to be the same as other tooltips?

pljones:
I'd go with
> force just them to be enabled

softins:
> I'd go with
> 
> > force just them to be enabled

OK, just need to work out how ;-)

pljones:
> > I'd go with
> > > force just them to be enabled
> 
> OK, just need to work out how ;-)

Alternatively, we come up with a different way of getting the info to display.  "Hover" doesn't work well on touch screens (it lacks... touch...) and I've not found a way to emulate it that works on Android.

ann0see:
Hover might be simulated by a long press on touch devices. But that doesn't work on desktop.

softins:
Actually, I can't think of a reason it couldn't just be click/tap on the fader info box to show fuller info. It's never used for any other purpose. I'll do some experimentation.

src/clientsettingsdlg.cpp
softins: OK, I've changed this in the user-facing text

softins:
I've reimplemented the filtering to avoid using an application-level filter, which would cause _all_ events to go through it, and may have efficiency implications. Instead, each relevant dialog has its own small `eventFilter` (which needs access to `pSettings`), and this filter is installed directly onto each widget that has a tooltip set. Currently, this is only `CClientDlg` and `CClientSettingsDlg`. The same technique would need to be used for `CConnectDlg` in #3252, which was the original motivation for this PR.

By _not_ installing a filter on the fader widgets, the Musician info is always displayed, whether other tooltips are enabled or disabled.

I haven't made any changes re click vs hover.

src/clientdlg.cpp
```diff
@@ -84,11 +84,13 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     lbrInputLevelL->setAccessibleName ( strInpLevHAccText );
     lbrInputLevelL->setAccessibleDescription ( strInpLevHAccDescr );
     lbrInputLevelL->setToolTip ( strInpLevHTT );
+    lbrInputLevelL->installEventFilter ( this ); // install event filter for tooltips
```
ann0see, line 87: The comment is basically a duplicate of the method. Not sure how much it helps.

ann0see (commented):
I think on Gnome/Debian 12 there are no tooltips (unrelated). Will test on Windows.

ann0see:
Ok. On windows it works as expected.

src/clientdlg.cpp
pljones: Not really: the supplied event filter gets called by Qt for every event - the comment is to clarify that the event filter provided on `this` only handles `toolTip` events (rather than, e.g. click events).

ann0see (approved):

mcfnord:
Since this was approved but has drifted into conflict with `main`, I've rebased it as #3807 — your two commits unchanged, plus one adaptation commit folding the tooltip check into the `eventFilter` that `main` has since gained in `CClientSettingsDlg` (MIDI device-list refresh). @softins if you'd rather update this branch yourself, happy to close mine — otherwise #3807 is ready to review.

ann0see:
I don't think the connection dialog descriptions will be merged any time soon and otherwise it's not annoying -> closing.

PR #3451, opened 2024-12-28
Build: Fix windows for qt6.8.1

pljones:

See https://github.com/jamulussoftware/jamulus/pull/3407 for discussion.

This includes the ci bump commit on the branch, currently.

CHANGELOG: SKIP

**Context: Fixes an issue?**

See https://github.com/jamulussoftware/jamulus/pull/3407 for discussion.

Leaves the iOS build broken.  There doesn't seem to be any way to pin versions in the "Bump versions" workflow.

I've a [branch here](https://github.com/jamulussoftware/jamulus/compare/main...revert-ios-to-qt6_7_3-until-ffmpeg-fixed) which reverts the iOS change but the "Bump versions" will just come and bump it again.  (Build [here](https://github.com/jamulussoftware/jamulus/actions/runs/12526188602/job/34938486672).)

No

Fixes the Windows builds.

No.

## Checklist

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -227,7 +227,7 @@ jobs:
           - config_name:            MacOS (artifacts)
             target_os:              macos
             building_on_os:         macos-14
-            base_command:           QT_VERSION=6.7.3 SIGN_IF_POSSIBLE=1 TARGET_ARCHS="x86_64 arm64" ./.github/autobuild/mac.sh
+            base_command:           QT_VERSION=6.8.1 SIGN_IF_POSSIBLE=1 TARGET_ARCHS="x86_64 arm64" ./.github/autobuild/mac.sh
             # Disable CodeQL on mac as it interferes with signing the binaries (signing hangs, see #2563 and #2564)
             run_codeql:             false
             # Latest Xcode which runs on macos-14:
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -78,7 +78,7 @@ Download and install Qt e.g via the [official open source installer](https://www
 
 **Note:**
 - Use the free GPLv2 license for Open Source development, not the commercial "universal installer"
-- Select Components during installation: Expand the Qt section, find the matching version, preferrably **Qt 5.15.2**, and add the compiler components for your compiler, e.g., `MSVC 2019 32-bit/64-bit` for Visual Studio 2019
+- Select Components during installation: Expand the Qt section, find the matching version.  To match the Github builds, you will need to check the versions in `windows/deploy_windows.ps2`.  This gives both the Qt and MSVC versions (e.g. 6.8.1 and msvc2022_64 for a 64bit release).
 
 If you build with *JACK* support, install JACK via choco: `choco install --no-progress -y jack`
 
--- a/windows/deploy_windows.ps1
+++ b/windows/deploy_windows.ps1
@@ -1,9 +1,9 @@
 param (
     # Replace default path with system Qt installation folder if necessary
     [string] $QtInstallPath32 = "C:\Qt\5.15.2",
-    [string] $QtInstallPath64 = "C:\Qt\5.15.2",
+    [string] $QtInstallPath64 = "C:\Qt\6.8.1",
     [string] $QtCompile32 = "msvc2019",
-    [string] $QtCompile64 = "msvc2019_64",
+    [string] $QtCompile64 = "msvc2022_64",
     # Important:
     # - Do not update ASIO SDK without checking for license-related changes.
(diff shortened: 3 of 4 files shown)


windows/deploy_windows.ps1
```diff
@@ -1,9 +1,9 @@
 param (
     # Replace default path with system Qt installation folder if necessary
     [string] $QtInstallPath32 = "C:\Qt\5.15.2",
-    [string] $QtInstallPath64 = "C:\Qt\5.15.2",
+    [string] $QtInstallPath64 = "C:\Qt\6.8.1",
```
ann0see, line 4: This would change the local build environment. We'll need to update the COMPILING.md file probably.

ann0see:
> Leaves the iOS build broken. There doesn't seem to be any way to pin versions in the "Bump versions" workflow.

What if you just update the windows build manually and let #3407 focus only on bumping iOS only? Then we don't have build failures.

windows/deploy_windows.ps1
pljones: Not sure what you mean?

pljones:
> > Leaves the iOS build broken. There doesn't seem to be any way to pin versions in the "Bump versions" workflow.
> 
> What if you just update the windows build manually and let #3407 focus only on bumping iOS only? Then we don't have build failures.

Yup, that's the idea.

windows/deploy_windows.ps1
ann0see: As far as I know, changing this variable only affects where the script assumes Qt is if no parameters are set. The COMPILING.md file might be changed to tell the user that they should install multiple versions of Qt locally.

pljones:
```
### Supported platforms

We support the following platforms and versions:

- **Windows 10** or later
- **macOS 10.10** or later
- **Ubuntu 20.04** or later, **Debian 11** or later, most Linux flavors with recent enough Qt versions. (Currently, it may still build on Ubuntu 18.04 and Debian 10, but the binaries built by Github will not run on those versions.)

_While Android and iOS aren't officially supported, please don't break their builds._

Please try to avoid breaking any build by introducing platform-specific code. Check the Github builds all worked before raising a pull request.
Check to see if any newly introduced Qt calls, parameters, properties or constants are available in the minimum supported Qt version, which is currently **5.12.2**. Note that code _style_ in a file may be Qt 4.x. While you should normally stick to existing style, if you make large-scale modifications, updating to Qt 5.12.2 style is recommended.
Maintain C++11 compatibility throughout the code.
```
This doesn't mention _where_ Qt should be installed.  It also states 5.12.2 is the _minimum_ version: which isn't true for several platforms for several releases, anyway.  So I don't think the change here conflicts with it.

pljones: OK, I've made `COMPILING.md` reference the deploy script rather than duplicating information that needs maintaining in multiple places (along with giving an example).

COMPILING.md
```diff
@@ -78,7 +78,7 @@ Download and install Qt e.g via the [official open source installer](https://www
 
 **Note:**
 - Use the free GPLv2 license for Open Source development, not the commercial "universal installer"
-- Select Components during installation: Expand the Qt section, find the matching version, preferrably **Qt 5.15.2**, and add the compiler components for your compiler, e.g., `MSVC 2019 32-bit/64-bit` for Visual Studio 2019
+- Select Components during installation: Expand the Qt section, find the matching version.  To match the Github builds, you will need to check the versions in `window/deploy_windows.ps2`.  This gives both the Qt and MSVC versions (e.g. 6.8.1 and msvc2022_64 for a 64bit release).
```
ann0see, line 81:
```suggestion
- Select Components during installation: Expand the Qt section, find the matching version.  To match the Github builds, you will need to check the versions in `windows/deploy_windows.ps2`.  This gives both the Qt and MSVC versions (e.g. 6.8.1 and msvc2022_64 for a 64bit release).
```

ann0see (changes requested):

COMPILING.md
pljones: Ooops...

.github/workflows/autobuild.yml
```diff
@@ -251,7 +251,7 @@ jobs:
           - config_name:            iOS (artifacts)
             target_os:              ios
             building_on_os:         macos-14
-            base_command:           QT_VERSION=6.7.3 ./.github/autobuild/ios.sh
```
ann0see, line 254: I would still prefer that the Qt upgrade for iOS remains untouched in this PR.

ann0see (approved):

PR #3457, opened 2024-12-30
QML client rewrite

danryu:
**Description of changes**

NOTE: This PR is posted at the request of @ann0see, see https://github.com/orgs/jamulussoftware/discussions/3454.
It is intended as a reference or baseline to assist in the effort of converting to a QML-based version in the future.

_N.B. I had to implement a freeze on new upstream merges due to the intricate nature of these changes. So it may be necessary to cherry-pick / re-implement / port any necessary upstream changes during the last 4-6 months of 2024 to the files in question, particularly client.cpp and audiomixerboard.cpp_

While setting myself the challenge to learn QML, I have succeeded in rewriting the core Jamulus client GUI with QML, removing all widget code in the process.
To put it mildly, it was a large, non-trivial effort to disentangle all the widget display logic from the business logic.

All existing business logic is largely preserved, with small changes and additions here and there.
In general, things are much simplified when it comes to UI element handling. No need, for example, for the clever audiomixerboard template stuff, or creating and hiding 250 channelfader objects :)

Key Changes:

    clientdlg -> removed, business logic moved to client, display moved to QML
    settingsdlg -> removed, business logic moved to settings, display moved to QML
    chatdlg -> converted to chatbox
    audiomixerboard -> CChannelFader and CAudioMixerBoard classes refactored as necessary
    all *.qml files representing key components in the UI

Design
The UI design is just a demo to contain all the functioning elements. It can easily be adapted as necessary.
For this reason it has almost no styling applied.

Features
All the core features of the client have been reproduced in the QML version. There may still be some subtle bugs, but it is already quite usable (only tested on Windows 11).

Exclusions
To reduce the scope of the project (already large enough) I have not implemented the following, but they should not be challenging to re-introduce:

    server chooser dialog
    instruments/skills/flags
    reverb
    what's this stuff
    translations / language chooser stuff

I invite you to build and test it locally, and offer any suggestions or enhancements.
Especially on macOS as I don't have any Apple hardware at the moment.

CHANGELOG: 

**Context: Fixes an issue?**

## Checklist


Jamulus.pro
ann0see, line 1: You did quite a lot of changes to Jamulus.pro
It seems to be overkill.

src/audiomixerboard.cpp
```diff
@@ -25,388 +25,156 @@
 #include "audiomixerboard.h"
 
 /******************************************************************************\
-* CChanneFader                                                                 *
```
ann0see, line 28: Good catch

src/chatbox.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    // add new text in chat window
+    if (strChatText.isEmpty())
+        return;
+    // Append new message to existing history
+    m_chatHistory.append(strChatText + "<br>");
+    emit chatHistoryChanged();
+}
+
+// void CChatBox::OnAnchorClicked ( const QUrl& Url )
```
ann0see, line 98: How do message boxes work in qml?

Jamulus.pro
danryu: Can you be specific?
In this commit there are a few extra changes in Jamulus.pro included from earlier dev / simplification work. There is quite a lot of cruft IMO anyway in the .pro file, but feel free to push or suggest changes.

src/client.cpp
```diff
@@ -281,91 +377,36 @@ void CClient::OnNewConnection()
...
-        Q_ASSERT ( iActiveChannels == iNumConnectedClients );
     }
 
-    // pass the received list onwards, now containing client channel IDs
-
-    emit ConClientListMesReceived ( vecChanInfo );
+    // update mixer board with the additional client infos
+    audioMixerBoard.ApplyNewConClientList( vecChanInfo );
+    // set session status
+    setSessionStatus("CONNECTED");
```
ann0see, line 409: I don’t think we need to necessarily always have a text saying „Connected“. The state should be conveyed differently. 
if the connection stands faders are visible, else not. Potential we‘d want to say „Disconnected“ in this case.

src/client.cpp
```diff
@@ -431,139 +472,97 @@ void CClient::SetDoAutoSockBufSize ( const bool bValue )
     CreateServerJitterBufferMessage();
 }
 
-// In order not to flood the server with gain or pan change messages, particularly when using
+// In order not to flood the server with gain change messages, particularly when using
```
ann0see, line 475: Seems like you revert some recent changes like the pan timer.

src/chatbox.cpp
danryu: Closest equivalent is the Popup (https://doc.qt.io/qt-6/qml-qtquick-controls-popup.html) which I used in a few cases, see AppWindow.qml.

src/client.cpp
```diff
@@ -1494,183 +1472,391 @@ int CClient::EstimatedOverallDelay ( const int iPingTimeMs )
...
+         ( GetLevelForMeterdBLeft() > NUM_STEPS_LED_BAR - 0.5 || GetLevelForMeterdBRight() > NUM_STEPS_LED_BAR - 0.5 ) )
+    {
+        // mute locally and mute channel
+        this->setMuteOut( true );
+        audioMixerBoard.MuteMyChannel();
 
-    for ( int i = 0; i < MAX_NUM_CHANNELS; i++ )
+        qDebug() << "Feedback detected ... ";
+        // show message about feedback issue
+        setUserMsg( tr ( "Audio feedback or loud signal detected.\n\n"
```
ann0see, line 1525: Ok. That’s how errors can be shown.

src/client.cpp
```diff
@@ -1494,183 +1472,391 @@ int CClient::EstimatedOverallDelay ( const int iPingTimeMs )
...
 
-    // qInfo() << "> Client channel list cleared";
+void CClient::OnTimerDetectFeedback()
+{
+    bDetectFeedback = false;
 }
 
-void CClient::FreeClientChannel ( const int iServerChannelID )
+
+void CClient::Connect( const QString& strAddress )
```
ann0see, line 1562: This might be useful for #3372

src/client.cpp
danryu: Eesh, that's possible, I tried to avoid that but with the length of time this took me, I may have missed a few commits.

src/main.cpp
```diff
@@ -342,9 +345,6 @@ int main ( int argc, char** argv )
         // HTML status file ----------------------------------------------------
         if ( GetStringArgument ( argc, argv, i, "-m", "--htmlstatus", strArgument ) )
         {
-            qWarning() << qUtf8Printable (
```
ann0see, line 345: It seems like some of the latest changes are reverted. The question is how we can have a clean state.

ann0see (commented):
I think one of the main challenges is to keep it up to date with the changes on main. 

The question is how can we progress. 

In the best case, after 3.12.0 is out, we‘d have a codebase which is clean and we could make the qml changes.

ann0see:
Maybe out of scope, maybe not:

We'd need to adapt the CI.

ann0see:
I could imagine the following:

1. Get an approval from other main developers that we want to move to qml in Jamulus 4
2. Create a separate branch for Jamulus 4 as soon as 3.12.0 is mostly done (3.12.0 is frozen)
3. Merge further changes into the Jamulus 4 branch

ann0see:
If I find time in the following months I'd like to look at getting the CI working.

src/client.cpp
danryu: I see a few other recent changes in client.cpp, such as https://github.com/jamulussoftware/jamulus/commit/49eb6080303190b8d2c1fd73bfb38f3a085f34dc. I'm not sure what this client<->server channel mapping is about, or what server channels are, or what the requirement for this was.
For things like this, it might be best to cherry-pick the relevant commits, as the number of files affected will be minimal.

ann0see:
It would be interesting if https://github.com/jamulussoftware/jamulus/issues/3406 could be fixed with this. I'll try a qt6 build for iOS soon.

ann0see:
@coderabbitai review this please. Assess if we should put this onto a feature branch?

pljones:
@coderabbit are you able to weigh the benefits of:
* QML vs JUCE
* Qt threading vs C++ threading
* moving Qt5 platforms to Qt6 and staying on Qt6

pljones:
@coderabbitai

If this branch were rebased to current main and the identified issues in your last review resolved, what would be the main benefit of moving to QML over remaining on Qt Widgets, including the custom ones used?

pljones:
@coderabbitui so the priority is getting the architecture right _first_, then considering the presentation layer.  If we can get the separation of concerns in place as is, we're in a better place to move to QML.

Issue #3466, opened 2025-01-17
Switch arm builds to ARM runners


ann0see:
**What is the current behaviour and why should it be changed?**

Currently we cross compile the Linux builds on x86. I just got notified that GitHub now supports ARM runners: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/ I think it's worth changing the runners to ARM instead of relying on a cross compiler. 

**Has this feature been discussed and generally agreed?**

No. I think I commented it somewhere

PR #3470, opened 2025-01-24
Switch autobuild OS to ARM runner

ann0see:

Switches the build to ARM runners if an arm build is produced

CHANGELOG: Switch Linux ARM builds to be built on arm runners

**Context: Fixes an issue?**

Fixes: #3466 

No

Tested on debian arm64 on a raspberry pi. armhf and arm64 both run. I'd like to have an independent test too (best on raspbian).

External testing

## Checklist

### DIFF
--- a/.github/autobuild/linux_deb.sh
+++ b/.github/autobuild/linux_deb.sh
@@ -51,6 +51,9 @@ esac
 setup() {
     export DEBIAN_FRONTEND="noninteractive"
 
+    APT_ARCH="$(dpkg --print-architecture)"
+    export APT_ARCH
+
     setup_cross_compilation_apt_sources
 
     echo "Installing dependencies..."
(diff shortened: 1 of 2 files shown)


ann0see:
ARM runners are considered public beta. I believe that if we don't remove the possibility to build on x86, it would be ok to merge this. But I'd also like to hear a general opinion.

softins (commented):
I won't have time to test these builds for several days yet.

I've just looked at the build logs, and notice that native compilation only works for amd64 and arm64 architectures. Building armhf is still cross-compilation, but now on an arm64 runner. I assume that's expected.

ann0see:
> Building armhf is still cross-compilation, but now on an arm64 runner. I assume that's expected.

Yes. That's expected.

.github/autobuild/linux_deb.sh
```diff
@@ -61,18 +64,22 @@ setup() {
...
     fi
+    echo "Building on ${APT_ARCH} for ${TARGET_ARCH}. Thus modifying apt..."
     sudo dpkg --add-architecture "${TARGET_ARCH}"
-    # Duplicate the original Ubuntu sources and modify them to refer to the TARGET_ARCH:
-    sed -rne "s|^deb.*/ ([^ -]+(-updates)?) main.*|deb [arch=${TARGET_ARCH}] http://ports.ubuntu.com/ubuntu-ports \1 main universe multiverse restricted|p" /etc/apt/sources.list | sudo dd of=/etc/apt/sources.list.d/"${TARGET_ARCH}".list
-    # Now take the original Ubuntu sources and limit those to the build host (i.e. non-TARGET_ARCH) architectures:
-    sudo sed -re 's/^deb /deb [arch=amd64,i386] /' -i /etc/apt/sources.list
+
+    if [[ "${APT_ARCH}" == "amd64"  ]]; then
+        # Duplicate the original Ubuntu sources and modify them to refer to the TARGET_ARCH:
```
pljones, line 74: Is there a reference to where this magic comes from?  I _think_ I can see what it's doing but it would be good if there were a "standard procedure" to follow.

ann0see: https://github.com/jamulussoftware/jamulus/blob/cc78e4c3cb45394f7513f67d1e7ce4f368beda27/.github/autobuild/linux_deb.sh#L54

Gives the architecture. For the repo replacing, you'd need to ask @hoffie 
The idea is to have more than one architecture in the apt sources to allow cross compiling. I'm not fully confident though.

pljones: Yeah, it's the `sed` stuff I'm querying.  Looks hacky without an explanation.

softins (approved):
Armhf build loads and runs ok on my Pi 4.

.github/autobuild/linux_deb.sh
pljones: OK, thanks - that's clear now!

.github/autobuild/linux_deb.sh
```diff
@@ -61,18 +64,22 @@ setup() {
...
         return
     fi
+    echo "Building on ${APT_ARCH} for ${TARGET_ARCH}. Thus modifying apt..."
     sudo dpkg --add-architecture "${TARGET_ARCH}"
-    # Duplicate the original Ubuntu sources and modify them to refer to the TARGET_ARCH:
-    sed -rne "s|^deb.*/ ([^ -]+(-updates)?) main.*|deb [arch=${TARGET_ARCH}] http://ports.ubuntu.com/ubuntu-ports \1 main universe multiverse restricted|p" /etc/apt/sources.list | sudo dd of=/etc/apt/sources.list.d/"${TARGET_ARCH}".list
-    # Now take the original Ubuntu sources and limit those to the build host (i.e. non-TARGET_ARCH) architectures:
-    sudo sed -re 's/^deb /deb [arch=amd64,i386] /' -i /etc/apt/sources.list
+
+    if [[ "${APT_ARCH}" == "amd64"  ]]; then
```
pljones, line 73: What happens if `"${APT_ARCH}"` is `arm64` and `"${TARGET_ARCH}"` is `"amd64"`, for example?  Should this _explicitly error_ before this point and then take `"${APT_ARCH}" == "amd64"` as given?

softins: That's a very good question. The present logic here seems to assume that cross-compilation is only done on an `amd64` runner. That's no longer the case: we explicitly cross-compile from `arm64` to `armhf` on an ARM runner. So I wonder whether this arch duplication is still needed in that case?

It would be good to get a reference for where this technique came from and why/when it's needed. Is @hoffie still around?

pljones (approved):
Approving on the basis that we'll find out if it breaks when it breaks.

PR #3502, opened 2025-05-22
Add MIDI GUI tab and learn function

ignotus666:
Adds a MIDI tab to the Settings menu and exposes the MIDI parameters so they can be changed via the GUI. MIDI CC numbers can be automatically set using "learn" buttons that when pressed, listen for an incoming MIDI message and store it. Starting Jamulus with `--ctrlmidich` command line arguments overrides and replaces those set via the GUI.

The MIDI-in port can be enabled/disabled at runtime instead of it being determined at launch by the `--ctrlmidich` parameter or by making it permanently open. This means that user names are prepended by a number depending on whether the MIDI in port has been enabled or not (as is the case now).

I've also ditched the "Offset" term, at least for the GUI. It's a hangover from the days when that's what it was, a value hard-coded for a specific Behringer device, but in the current context it no longer makes sense.

EDIT: Added JSONRPC support.

MIDI_GUI

CHANGELOG: Client: Added MIDI tab to Settings GUI exposing MIDI parameters. MIDI Learn feature also added.

**Context: Fixes an issue?**

Needs to be documented.

Working on Linux, Windows and macOS but should be tested more thoroughly.

Further testing and code review. 

## Checklist

### DIFF
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -188,6 +188,40 @@ Results:
 | result.clients | array | The client list. See jamulusclient/clientListReceived for the format. |
 
 
+### jamulusclient/getMidiDevices
+
+Returns a list of available MIDI input devices.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params | object | No parameters (empty object). |
+
+Results:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| result | array | Array of MIDI device name strings. |
+
+
+### jamulusclient/getMidiSettings
+
+Returns all MIDI controller settings.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params | object | No parameters (empty object). |
+
+Results:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| result | object | MIDI settings object. |
+
+
 ### jamulusclient/pollServerList
 
 Request list of servers in a directory.
@@ -232,6 +266,23 @@ Parameters:
 | --- | --- | --- |
 | params.channelIndex | number | The channel index of the fader to be set. |
 | params.level | number | The fader level in range 0..100. |
+
+Results:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| result | string | Always "ok". |
+
+
+### jamulusclient/setMidiSettings
+
+Sets one or more MIDI controller settings.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params | object | Any subset of MIDI settings fields to set. |
 
 Results:
 
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -23,19 +23,21 @@
 \******************************************************************************/
 
 #include "client.h"
+#include "settings.h"
 #include "util.h"
 
 /* Implementation *************************************************************/
 CClient::CClient ( const quint16  iPortNumber,
                    const quint16  iQosNumber,
                    const QString& strConnOnStartupAddress,
-                   const QString& strMIDISetup,
                    const bool     bNoAutoJackConnect,
                    const QString& strNClientName,
                    const bool     bNEnableIPv6,
                    const bool     bNMuteMeInPersonalMix ) :
     ChannelInfo(),
     strClientName ( strNClientName ),
+    pSignalHandler ( CSignalHandler::getSingletonP() ),
+    pSettings ( nullptr ),
     Channel ( false ), /* we need a client channel -> "false" */
     CurOpusEncoder ( nullptr ),
     CurOpusDecoder ( nullptr ),
@@ -49,7 +51,7 @@ CClient::CClient ( const quint16  iPortNumber,
     bMuteOutStream ( false ),
     fMuteOutStreamGain ( 1.0f ),
     Socket ( &Channel, iPortNumber, iQosNumber, "", bNEnableIPv6 ),
-    Sound ( AudioCallback, this, strMIDISetup, bNoAutoJackConnect, strNClientName ),
+    Sound ( AudioCallback, this, bNoAutoJackConnect, strNClientName ),
     iAudioInFader ( AUD_FADER_IN_MIDDLE ),
     bReverbOnLeftChan ( false ),
     iReverbLevel ( 0 ),
@@ -68,8 +70,7 @@ CClient::CClient ( const quint16  iPortNumber,
     bJitterBufferOK ( true ),
     bEnableIPv6 ( bNEnableIPv6 ),
     bMuteMeInPersonalMix ( bNMuteMeInPersonalMix ),
-    iServerSockBufNumFrames ( DEF_NET_BUF_SIZE_NUM_BL ),
-    pSignalHandler ( CSignalHandler::getSingletonP() )
+    iServerSockBufNumFrames ( DEF_NET_BUF_SIZE_NUM_BL )
 {
     int iOpusError;
 
@@ -173,6 +174,8 @@ CClient::CClient ( const quint16  iPortNumber,
 
     QObject::connect ( pSignalHandler, &CSignalHandler::HandledSignal, this, &CClient::OnHandledSignal );
 
+    QObject::connect ( &Sound, &CSoundBase::MidiCCReceived, this, [this] ( int ccNumber ) { emit MidiCCReceived ( ccNumber ); } );
+
     // start timer so that elapsed time works
     PreciseTime.start();
 
@@ -191,6 +194,29 @@ CClient::CClient ( const quint16  iPortNumber,
         SetServerAddr ( strConnOnStartupAddress );
         Start();
     }
+}
+
+// MIDI setup will be handled after settings are assigned
+void CClient::SetSettings ( CClientSettings* settings )
+{
+    pSettings = settings;
+
+    // Apply MIDI settings
+    Sound.SetCtrlMIDIChannel ( pSettings->iMidiChannel );
+    Sound.SetMIDIControllerMapping ( pSettings->iMidiFaderOffset,
+                                     pSettings->bMidiFaderEnabled ? pSettings->iMidiFaderCount : 0,
+                                     pSettings->iMidiPanOffset,
+                                     pSettings->bMidiPanEnabled ? pSettings->iMidiPanCount : 0,
+                                     pSettings->iMidiSoloOffset,
+                                     pSettings->bMidiSoloEnabled ? pSettings->iMidiSoloCount : 0,
+                                     pSettings->iMidiMuteOffset,
+                                     pSettings->bMidiMuteEnabled ? pSettings->iMidiMuteCount : 0,
+                                     pSettings->bMidiMuteMyselfEnabled ? pSettings->iMidiMuteMyself : -1 );
+    if ( !pSettings->strMidiDevice.isEmpty() )
+    {
+        Sound.SetMIDIDevice ( pSettings->strMidiDevice );
+    }
+    Sound.EnableMIDI ( pSettings->bUseMIDIController );
 }
 
 CClient::~CClient()
@@ -1546,6 +1572,8 @@ void CClient::FreeClientChannel ( const int iServerChannelID )
                                     .arg ( iActiveChannels ) );
      */
 }
+
+void CClient::OnMidiCCReceived ( int ccNumber ) { emit MidiCCReceived ( ccNumber ); }
 
 // find, and optionally create, a client channel for the supplied server channel ID
 // returns a client channel ID or INVALID_INDEX
--- a/src/client.h
+++ b/src/client.h
@@ -119,6 +119,8 @@ public:
     // can store here other information about an active channel
 };
 
+class CClientSettings;
+
 class CClient : public QObject
 {
     Q_OBJECT
@@ -127,7 +129,6 @@ public:
     CClient ( const quint16  iPortNumber,
               const quint16  iQosNumber,
               const QString& strConnOnStartupAddress,
-              const QString& strMIDISetup,
               const bool     bNoAutoJackConnect,
               const QString& strNClientName,
               const bool     bNEnableIPv6,
@@ -293,11 +294,25 @@ public:
     CProtocol* getConnLessProtocol() { return &ConnLessProtocol; }
     //### TODO: END ###//
 
+    // MIDI control
+    void        EnableMIDI ( bool bEnable ) { Sound.EnableMIDI ( bEnable ); }
+    bool        IsMIDIEnabled() const { return Sound.IsMIDIEnabled(); }
+    QStringList GetMIDIDevNames() { return Sound.GetMIDIDevNames(); }
+    QString     GetMIDIDevice() { return Sound.GetMIDIDevice(); }
+    void        SetMIDIDevice ( const QString& strDevice ) { Sound.SetMIDIDevice ( strDevice ); }
+
     // settings
     CChannelCoreInfo ChannelInfo;
     QString          strClientName;
 
+public:
+    void SetSettings ( CClientSettings* settings );
+
 protected:
+    // Signal handler must be declared before pSettings for correct init order
+    CSignalHandler* pSignalHandler;
+    // Pointer to settings for MIDI and other config
+    CClientSettings* pSettings;
     // callback function must be static, otherwise it does not work
     static void AudioCallback ( CVector<short>& psData, void* arg );
 
@@ -407,8 +422,6 @@ protected:
     int    maxGainOrPanId;
     int    iCurPingTime;
 
-    CSignalHandler* pSignalHandler;
-
 protected slots:
     void OnHandledSignal ( int sigNum );
     void OnSendProtMessage ( CVector<uint8_t> vecMessage );
@@ -473,4 +486,8 @@ signals:
     void ControllerInFaderIsSolo ( int iChannelIdx, bool bIsSolo );
     void ControllerInFaderIsMute ( int iChannelIdx, bool bIsMute );
     void ControllerInMuteMyself ( bool bMute );
+    void MidiCCReceived ( int ccNumber );
+
+private slots:
+    void OnMidiCCReceived ( int ccNumber );
 };
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -29,7 +29,6 @@
 CClientDlg::CClientDlg ( CClient*         pNCliP,
                          CClientSettings* pNSetP,
                          const QString&   strConnOnStartupAddress,
-                         const QString&   strMIDISetup,
                          const bool       bNewShowComplRegConnList,
                          const bool       bShowAnalyzerConsole,
                          const bool       bMuteStream,
@@ -220,7 +219,7 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     MainMixerBoard->SetNumMixerPanelRows ( pSettings->iNumMixerPanelRows );
 
     // Pass through flag for MIDICtrlUsed
-    MainMixerBoard->SetMIDICtrlUsed ( !strMIDISetup.isEmpty() );
+    MainMixerBoard->SetMIDICtrlUsed ( pSettings->bUseMIDIController );
 
     // reset mixer board
     MainMixerBoard->HideAll();
@@ -401,6 +400,12 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     pSettingsMenu->addAction ( tr ( "A&dvanced Settings..." ), this, SLOT ( OnOpenAdvancedSettings() ), QKeySequence ( Qt::CTRL + Qt::Key_D ) );
 
+    pSettingsMenu->addAction (
+        tr ( "&MIDI Control Settings..." ),
+        this,
+        [this] { ShowGeneralSettings ( SETTING_TAB_MIDI ); },
+        QKeySequence ( Qt::CTRL + Qt::Key_M ) );
+
     // Main menu bar -----------------------------------------------------------
     QMenuBar* pMenu = new QMenuBar ( this );
 
@@ -535,6 +540,8 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     QObject::connect ( &ClientSettingsDlg, &CClientSettingsDlg::CustomDirectoriesChanged, &ConnectDlg, &CConnectDlg::OnCustomDirectoriesChanged );
 
     QObject::connect ( &ClientSettingsDlg, &CClientSettingsDlg::NumMixerPanelRowsChanged, this, &CClientDlg::OnNumMixerPanelRowsChanged );
+
+    QObject::connect ( &ClientSettingsDlg, &CClientSettingsDlg::MIDIControllerUsageChanged, this, &CClientDlg::OnMIDIControllerUsageChanged );
 
     QObject::connect ( this, &CClientDlg::SendTabChange, &ClientSettingsDlg, &CClientSettingsDlg::OnMakeTabChange );
 
@@ -1525,4 +1532,14 @@ void CClientDlg::SetPingTime ( const int iPingTime, const int iOverallDelayMs, c
 
     // set current LED status
     ledDelay->SetLight ( eOverallDelayLEDColor );
+}
+
+// OnOpenMidiSettings slot removed; lambda is used in menu action
+void CClientDlg::OnMIDIControllerUsageChanged ( bool bEnabled )
+{
+    // Update the mixer board's MIDI flag to trigger proper user numbering display
+    MainMixerBoard->SetMIDICtrlUsed ( bEnabled );
+
+    // Enable/disable runtime MIDI via the sound interface through the public CClient interface
+    pClient->EnableMIDI ( bEnabled );
 }
--- a/src/clientdlg.h
+++ b/src/clientdlg.h
@@ -80,7 +80,6 @@ public:
     CClientDlg ( CClient*         pNCliP,
                  CClientSettings* pNSetP,
                  const QString&   strConnOnStartupAddress,
-                 const QString&   strMIDISetup,
                  const bool       bNewShowComplRegConnList,
                  const bool       bShowAnalyzerConsole,
                  const bool       bMuteStream,
@@ -245,6 +244,8 @@ public slots:
     void OnNumClientsChanged ( int iNewNumClients );
 
     void accept() { close(); } // introduced by pljones
+
+    void OnMIDIControllerUsageChanged ( bool bEnabled );
 
 signals:
     void SendTabChange ( int iTabIdx );
--- a/src/clientrpc.cpp
+++ b/src/clientrpc.cpp
@@ -25,7 +25,9 @@
 
 #include "clientrpc.h"
 
-CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* parent ) : QObject ( parent )
+CClientRpc::CClientRpc ( CClient* pClient, CClientSettings* pSettings, CRpcServer* pRpcServer, QObject* parent ) :
+    QObject ( parent ),
+    m_pSettings ( pSettings )
 {
     /// @rpc_notification jamulusclient/chatTextReceived
     /// @brief Emitted when a chat text is received.
@@ -354,6 +356,124 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
 
         pClient->SetControllerInFaderLevel ( jsonChannelIndex.toInt(), jsonLevel.toInt() );
         response["result"] = "ok";
+    } );
+
+    /// @rpc_method jamulusclient/getMidiSettings
+    /// @brief Returns all MIDI controller settings.
+    /// @param {object} params - No parameters (empty object).
+    /// @result {object} result - MIDI settings object.
+    pRpcServer->HandleMethod ( "jamulusclient/getMidiSettings", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        QJsonObject jsonMidiParams{ { "bUseMIDIController", m_pSettings->bUseMIDIController },
+                                    { "midiDevice", m_pSettings->strMidiDevice },
+                                    { "midiChannel", m_pSettings->iMidiChannel },
+                                    { "midiMuteMyself", m_pSettings->iMidiMuteMyself },
+                                    { "midiFaderOffset", m_pSettings->iMidiFaderOffset },
+                                    { "midiFaderCount", m_pSettings->iMidiFaderCount },
+                                    { "midiPanOffset", m_pSettings->iMidiPanOffset },
+                                    { "midiPanCount", m_pSettings->iMidiPanCount },
+                                    { "midiSoloOffset", m_pSettings->iMidiSoloOffset },
+                                    { "midiSoloCount", m_pSettings->iMidiSoloCount },
+                                    { "midiMuteOffset", m_pSettings->iMidiMuteOffset },
+                                    { "midiMuteCount", m_pSettings->iMidiMuteCount },
+                                    { "bMidiFaderEnabled", m_pSettings->bMidiFaderEnabled },
+                                    { "bMidiPanEnabled", m_pSettings->bMidiPanEnabled },
+                                    { "bMidiSoloEnabled", m_pSettings->bMidiSoloEnabled },
+                                    { "bMidiMuteEnabled", m_pSettings->bMidiMuteEnabled },
+                                    { "bMidiMuteMyselfEnabled", m_pSettings->bMidiMuteMyselfEnabled },
+                                    { "bMIDIPickupMode", m_pSettings->bMIDIPickupMode } };
+        response["result"] = jsonMidiParams;
+        Q_UNUSED ( params );
+    } );
+
+    /// @rpc_method jamulusclient/setMidiSettings
+    /// @brief Sets one or more MIDI controller settings.
+    /// @param {object} params - Any subset of MIDI settings fields to set.
+    /// @result {string} result - Always "ok".
+    pRpcServer->HandleMethod ( "jamulusclient/setMidiSettings", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        bool bPreviousMIDIState = m_pSettings->bUseMIDIController;
+
+        QHash<QString, std::function<void ( const QJsonValue& )>> setters = {
+            { "bUseMIDIController", [this] ( const QJsonValue& v ) { m_pSettings->bUseMIDIController = v.toBool(); } },
+            { "midiDevice", [this] ( const QJsonValue& v ) { m_pSettings->strMidiDevice = v.toString(); } },
+            { "midiChannel", [this] ( const QJsonValue& v ) { m_pSettings->iMidiChannel = v.toInt(); } },
+            { "midiMuteMyself", [this] ( const QJsonValue& v ) { m_pSettings->iMidiMuteMyself = v.toInt(); } },
+            { "midiFaderOffset", [this] ( const QJsonValue& v ) { m_pSettings->iMidiFaderOffset = v.toInt(); } },
+            { "midiFaderCount", [this] ( const QJsonValue& v ) { m_pSettings->iMidiFaderCount = v.toInt(); } },
+            { "midiPanOffset", [this] ( const QJsonValue& v ) { m_pSettings->iMidiPanOffset = v.toInt(); } },
+            { "midiPanCount", [this] ( const QJsonValue& v ) { m_pSettings->iMidiPanCount = v.toInt(); } },
+            { "midiSoloOffset", [this] ( const QJsonValue& v ) { m_pSettings->iMidiSoloOffset = v.toInt(); } },
+            { "midiSoloCount", [this] ( const QJsonValue& v ) { m_pSettings->iMidiSoloCount = v.toInt(); } },
+            { "midiMuteOffset", [this] ( const QJsonValue& v ) { m_pSettings->iMidiMuteOffset = v.toInt(); } },
+            { "midiMuteCount", [this] ( const QJsonValue& v ) { m_pSettings->iMidiMuteCount = v.toInt(); } },
+            { "bMidiFaderEnabled", [this] ( const QJsonValue& v ) { m_pSettings->bMidiFaderEnabled = v.toBool(); } },
+            { "bMidiPanEnabled", [this] ( const QJsonValue& v ) { m_pSettings->bMidiPanEnabled = v.toBool(); } },
+            { "bMidiSoloEnabled", [this] ( const QJsonValue& v ) { m_pSettings->bMidiSoloEnabled = v.toBool(); } },
+            { "bMidiMuteEnabled", [this] ( const QJsonValue& v ) { m_pSettings->bMidiMuteEnabled = v.toBool(); } },
+            { "bMidiMuteMyselfEnabled", [this] ( const QJsonValue& v ) { m_pSettings->bMidiMuteMyselfEnabled = v.toBool(); } },
+            { "bMIDIPickupMode", [this] ( const QJsonValue& v ) { m_pSettings->bMIDIPickupMode = v.toBool(); } } };
+
+        for ( auto it = setters.constBegin(); it != setters.constEnd(); ++it )
+        {
+            if ( params.contains ( it.key() ) )
+            {
+                it.value() ( params[it.key()] );
+            }
+        }
+
+        // If midiDevice was changed and MIDI is currently enabled, restart MIDI to reconnect
+        bool bDeviceChanged = params.contains ( "midiDevice" );
+        if ( bDeviceChanged && m_pSettings->bUseMIDIController && pClient->IsMIDIEnabled() )
+        {
+            // Disable MIDI
+            pClient->EnableMIDI ( false );
+
+            // Set the new device
+            pClient->SetMIDIDevice ( m_pSettings->strMidiDevice );
+
+            // Re-enable MIDI
+            pClient->EnableMIDI ( true );
+
+            // Check if reconnection was successful
+            if ( !pClient->IsMIDIEnabled() )
+            {
+                response["error"] = CRpcServer::CreateJsonRpcError ( 1, "Failed to connect to MIDI device" );
+                return;
+            }
+        }
+        else if ( bDeviceChanged )
+        {
+            // Just update the device setting for next time MIDI is enabled
+            pClient->SetMIDIDevice ( m_pSettings->strMidiDevice );
+        }
+
+        // Apply other settings to actually enable/disable MIDI
+        pClient->SetSettings ( m_pSettings );
+
+        // Check if MIDI was requested but failed to enable
+        if ( m_pSettings->bUseMIDIController && !pClient->IsMIDIEnabled() )
+        {
+            // Restore previous state on failure
+            m_pSettings->bUseMIDIController = bPreviousMIDIState;
+            response["error"]               = CRpcServer::CreateJsonRpcError ( 1, "Failed to open MIDI port" );
+            return;
+        }
+
+        response["result"] = "ok";
+    } );
+
+    /// @rpc_method jamulusclient/getMidiDevices
+    /// @brief Returns a list of available MIDI input devices.
+    /// @param {object} params - No parameters (empty object).
+    /// @result {array} result - Array of MIDI device name strings.
+    pRpcServer->HandleMethod ( "jamulusclient/getMidiDevices", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        QStringList deviceNames = pClient->GetMIDIDevNames();
+        QJsonArray  jsonDevices;
+        for ( const QString& deviceName : deviceNames )
+        {
+            jsonDevices.append ( deviceName );
+        }
+        response["result"] = jsonDevices;
+        Q_UNUSED ( params );
     } );
 }
 
--- a/src/clientsettingsdlg.cpp
+++ b/src/clientsettingsdlg.cpp
@@ -28,7 +28,8 @@
 CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSetP, QWidget* parent ) :
     CBaseDlg ( parent, Qt::Window ), // use Qt::Window to get min/max window buttons
     pClient ( pNCliP ),
-    pSettings ( pNSetP )
+    pSettings ( pNSetP ),
+    midiLearnTarget ( None )
 {
     setupUi ( this );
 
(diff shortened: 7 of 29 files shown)


ignotus666:
@softins thanks - no worries, I guess there will be plenty of time for testing before this ever makes it in (if it does).

Hope the house move goes smoothly.

ann0see:
Nice!

I'd like to open a discussion:
1. Do you think a separate MIDI tab is useful or does it add clutter?
2. Is there a better word instead of "Learn" for the detector? How can we make the feature clear to the user? Would a dynamic text field showing currently pressed buttons be better? (I'm unsure as this could be more complicated)

src/clientsettingsdlg.cpp
```diff
@@ -1216,3 +1330,88 @@ void CClientSettingsDlg::OnAudioPanValueChanged ( int value )
...
+        button->setEnabled ( button == activeButton );
+    }
+}
+
+void CClientSettingsDlg::OnLearnButtonClicked()
+{
+    QPushButton* sender = qobject_cast<QPushButton*> ( QObject::sender() );
+
+    MidiLearnTarget target = None;
+    if ( sender == butLearnMuteMyself )
```
ann0see: Please add '{}'. That's my Personal preference.

ignotus666:
@ann0see 

1. Given the number of MIDI settings, I don't see how it would fit in somewhere else without cluttering up that section in a worse way. I think it merits its own tab.
2. "Learn" is a standard term for this feature. I see and use it in many other MIDI applications. Maybe "MIDI learn"? But I think that might make the button too large.

In any event, this is still under development as I'm finding that some of my implementations are far from perfect. E.g. the way the MIDI in port is currently kept open I think has been done wrong, so I'm working on that. I'm testing a version where the MIDI port can be enabled/disabled at runtime, as among other things it affects whether numbers are prepended to user names in the mixer board - if you're not using MIDI you won't want that.

The main purpose for opening this PR was to generate builds for other platforms for testing, but for now it's just a PoC that's far from finished, so any suggestions are welcome.

src/clientsettingsdlg.cpp
ignotus666: Done.

pljones (changes requested):
Get rid of all the unnecessary changes in spacing and resubmit the PR.  As it stands, I can't review it.

ignotus666:
> * Any UI enhancement should be mirrored by an enhancement to the JSONRPC API, if the feature is missing there

I've been working on this - should the UI immediately reflect changes from JSONRPC commands sent? It doesn't seem to do this for the existing commands but just to confirm.

src/sound/asio/sound.h
```diff
@@ -82,6 +82,10 @@ class CSound : public CSoundBase
 
     virtual float GetInOutLatencyMs() { return fInOutLatencyMs; }
 
+    // MIDI port toggle
```
pljones, line 85: Shouldn't these be declared once in `CSoundBase`?  All implementations need to define them.

src/clientdlg.cpp
```diff
@@ -990,9 +994,8 @@ void CClientDlg::ShowGeneralSettings ( int iTab )
     // open general settings dialog
     emit SendTabChange ( iTab );
     ClientSettingsDlg.show();
```
pljones, line 999: Move after setting the title?  Or does the window manager need waking up to the window's existence before the title is set?  Who knows.  Maybe safer to leave it for now.

src/clientsettingsdlg.cpp
```diff
@@ -397,6 +398,43 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
                                         "A second sound device may be required to hear the alerts." ) );
     chbAudioAlerts->setAccessibleName ( tr ( "Audio Alerts check box" ) );
 
+    // MIDI settings
+    chbUseMIDIController->setWhatsThis ( tr ( "Enable/disable MIDI-in port" ) );
+    chbUseMIDIController->setAccessibleName ( tr ( "MIDI-in port check box" ) );
+
+    QString strMidiSettings = "<b>" + tr ( "MIDI controller settings" ) + ":</b> " +
+                              tr ( "There is one global MIDI channel parameter (1-16) and two parameters you can set "
```
pljones:
```suggestion
                              tr ( "There is one global MIDI channel parameter (0-16) and two parameters you can set "
```

src/clientsettingsdlg.cpp
```diff
@@ -746,6 +784,87 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
 
     QObject::connect ( pcbxSkill, static_cast<void ( QComboBox::* ) ( int )> ( &QComboBox::activated ), this, &CClientSettingsDlg::OnSkillActivated );
 
+    // MIDI tab
```
pljones, line 807: Oooh, I want a loop... can you loop over the spinner references?

src/clientsettingsdlg.cpp
```diff
@@ -1216,3 +1355,113 @@ void CClientSettingsDlg::OnAudioPanValueChanged ( int value )
...
+        return;
+
+    // Validate MIDI CC number is within valid range (0-127)
+    if ( ccNumber < 0 || ccNumber > 127 )
+    {
+        qWarning() << "CClientSettingsDlg::OnMidiCCReceived: Invalid MIDI CC number received:" << ccNumber;
+        return;
+    }
+
+    switch ( midiLearnTarget )
```
pljones: And I'm a sucker for a static map (from typeof(midiLearnTarget) to spinner), then just a one-liner to call the setValue...

src/clientsettingsdlg.cpp
```diff
@@ -1216,3 +1355,113 @@ void CClientSettingsDlg::OnAudioPanValueChanged ( int value )
...
+    }
+}
+
+void CClientSettingsDlg::OnLearnButtonClicked()
+{
+    QPushButton* sender = qobject_cast<QPushButton*> ( QObject::sender() );
+
+    MidiLearnTarget target = None;
+
+    if ( sender == butLearnMuteMyself )
```
pljones: Oh and a map from button to target?  So `SetMidiLearnTarget ( targetMap.get( sender ), sender )`...?  (Too much java recently...)

linux/debian/rules
```diff
@@ -46,3 +46,6 @@ override_dh_auto_clean:
 	rm -rf build-gui
 	rm -rf build-nox
 	dh_clean
+
+override_dh_shlibdeps:
+	dh_shlibdeps -l/usr/lib/x86_64-linux-gnu
```
pljones: Why is this needed as part of this PR?

ignotus666: I mistakenly left it in as it won't compile on my PC without it. Fixing this and a couple of other things - will rebase and squash now.

src/sound/asio/sound.h
pljones, line 85: As these _are_ in CSoundBase, do they need to be in each inheriting class?

src/sound/soundbase.cpp
```diff
@@ -367,6 +432,7 @@ void CSoundBase::ParseMIDIMessage ( const CVector<uint8_t>& vMIDIPaketBytes )
                     {
                         const CMidiCtlEntry& cCtrl  = aMidiCtls[vMIDIPaketBytes[1]];
                         const int            iValue = vMIDIPaketBytes[2];
+                        emit                 MidiCCReceived ( vMIDIPaketBytes[1] );
```
pljones: Hm.  Formatting looks odd.

Put a blank line before and remove the aligning spaces.  (My guess is clang doesn't know what `emit` is...)

src/sound/soundbase.cpp
```diff
@@ -367,6 +432,7 @@ void CSoundBase::ParseMIDIMessage ( const CVector<uint8_t>& vMIDIPaketBytes )
...
                         ;
```
pljones: And this shouldn't be here -- clang should have caught it.

src/client.cpp
```diff
@@ -173,6 +173,8 @@ CClient::CClient ( const quint16  iPortNumber,
 
     QObject::connect ( pSignalHandler, &CSignalHandler::HandledSignal, this, &CClient::OnHandledSignal );
 
+    QObject::connect ( &Sound, &CSoundBase::MidiCCReceived, this, &CClient::OnMidiCCReceived );
```
pljones: I'd just use the modern syntax and lambda format here.  (This is one area where we're doing slow improvements rather than keeping to existing style - for "one-liner" stuff like this particularly.)

src/clientdlg.h
```diff
@@ -246,6 +246,9 @@ public slots:
 
     void accept() { close(); } // introduced by pljones
 
+    void OnOpenMidiSettings();
```
pljones: Hm.  _Public_ slots should be "service endpoints" into which other objects hook their signals, not handlers that are used internally to the class only.  (Yes, plenty of examples where the existing code doesn't follow that...)

If the method is used only once, it's eligible for turning into a lambda (subject to readability) or having a lambda call an inline private method.

src/clientrpc.cpp
```diff
@@ -355,6 +357,78 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
...
+                            { "midiMuteCount", m_pSettings->midiMuteCount } };
+        response["result"] = result;
+        Q_UNUSED ( params );
+    } );
+
+    /// @rpc_method jamulusclient/setMidiSettings
+    /// @brief Sets one or more MIDI controller settings.
+    /// @param {object} params - Any subset of MIDI settings fields to set.
+    /// @result {string} result - Always "ok".
+    pRpcServer->HandleMethod ( "jamulusclient/setMidiSettings", [=] ( const QJsonObject& params, QJsonObject& response ) {
```
pljones, line 392: Hm.  Should be able to do this with a struct containing the JSON name, the setter reference and a lambda to convert from the JSON... but I've no idea how, as it would need to be a template, I think, with dynamic casts.  And that's way above my C++ level 😀 .

pljones: Ideally, I'd like one struct with `name`, `helpText`, `settingsGetter` (method reference), `settingsSetter` (method reference), `settingsChangeSignal` (emitting `name`), `fromSettings` (conversion lambda), `toSettings` (conversion lambda).  Or something...  Then, when the UI or RPC needed to request a value, it would look up the necessary bits by name and the `connect`s could all be single lines.

(There just seems like so much repetitive template code...  Maybe ask Github CoPilot?)

ignotus666: I've had a go at addressing some of the things you've pointed out but ran into a few problems along the way with the rebasing (I've got a confusing amount of local versions that need to be organised...) and clang-format, which is really doing my head in... I'll get back to this when I have some time.

ignotus666: Ok, so I don't really know what this entails but Copilot says: " implementing this in C++ (especially with Qt types) is non-trivial and would require advanced template metaprogramming, type erasure, or a variant-based approach. (...) This is a significant refactor and may be overkill unless you have many similar settings. "

src/settings.cpp
```diff
@@ -461,6 +461,62 @@ void CClientSettings::ReadSettingsFromXML ( const QDomDocument& IniXMLDocument,
         pClient->SetAudioQuality ( static_cast<EAudioQuality> ( iValue ) );
     }
 
+    // MIDI settings
```
ignotus666: @pljones would you prefer a loop here too?

src/clientrpc.cpp
pljones: Yeah, that's what I thought... Fun if you've nothing else to do, though 😁 ...

src/main.cpp
```diff
@@ -920,25 +920,31 @@ int main ( int argc, char** argv )
...
                              strConnOnStartupAddress,
-                             strMIDISetup,
+                             "", // Always start with empty MIDI
                              bNoAutoJackConnect,
                              strClientName,
                              bEnableIPv6,
                              bMuteMeInPersonalMix );
 
-            // load settings from init-file (command line options override)
+            // Create Settings with the client pointer
```
pljones, line 926: The deleted line is correct -- but for the `Settings.Load` line.

src/main.cpp
```diff
@@ -920,25 +920,31 @@ int main ( int argc, char** argv )
...
             CClientSettings Settings ( &Client, strIniFileName );
             Settings.Load ( CommandLineOptions );
 
+            // Parse command line MIDI parameters if provided
+            CSoundBase::ParseMIDICommandLineParams ( strMIDISetup, Settings );
+            if ( Settings.bUseMIDIController )
+            {
+                Client.ApplyMIDIMapping ( Settings.GetMIDIMapString() );
```
pljones: This now seems counter-intuitive.

Maybe `CSoundBase::ParseMIDICommandLineParams ( strMIDISetup, Settings );` should be in settings (it wouldn't need to keep calling into Settings, then).

Given the command line arg has been parsed out into separate values in Settings, turning it back into a string so that it can be parsed out into separate settings again...  Client already has a Settings reference, so it can go read the values.

src/settings.h
pljones, line 166: You need to follow this style for all the new entries.

src/settings.h
```diff
@@ -201,6 +201,20 @@ class CClientSettings : public CSettings
     bool       bWindowWasShownConnect;
     bool       bOwnFaderFirst;
 
+    // MIDI settings
+    int     midiChannel        = 0; // Default MIDI channel 0
```
pljones: These really should all be `iMidi...` by convention.

src/settings.h
```diff
@@ -201,6 +201,20 @@ class CClientSettings : public CSettings
...
+    int     midiFaderOffset    = 0;
+    int     midiFaderCount     = 0;
+    int     midiPanOffset      = 0;
+    int     midiPanCount       = 0;
+    int     midiSoloOffset     = 0;
+    int     midiSoloCount      = 0;
+    int     midiMuteOffset     = 0;
+    int     midiMuteCount      = 0;
+    bool    bUseMIDIController = false;
+    QString GetMIDIMapString() const;
```
pljones: As other note, I'd like this to go away.

src/settings.cpp
pljones: Map from `"whatever"` to `[this]() { GetNumericIniSet ( IniXMLDocument, "client", "whatever", minWhatever, maxWhatever, iValue ); this->whatever = iValue; };` and then loop over an array of them?

Or do `"midichannel"` and `"usemidicontroller"` separately, with everything in the loop with the `0, 127` hardcoded...  I'd go with that, I think.

src/clientrpc.cpp
```diff
@@ -355,6 +357,78 @@ CClientRpc::CClientRpc ( CClient* pClient, CRpcServer* pRpcServer, QObject* pare
...
+        response["result"] = jsonMidiParams;
+        Q_UNUSED ( params );
+    } );
+
+    /// @rpc_method jamulusclient/setMidiSettings
+    /// @brief Sets one or more MIDI controller settings.
+    /// @param {object} params - Any subset of MIDI settings fields to set.
+    /// @result {string} result - Always "ok".
+    pRpcServer->HandleMethod ( "jamulusclient/setMidiSettings", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        if ( params.contains ( "bUseMIDIController" ) )
```
pljones: Do I trust Gemini to write C++?
```
std::unordered_map<String, std::function<void(String)>> setters = {
    { "bUseMIDIController", [this](String name) { m_pSettings->bUseMIDIController = params[name].toBool(); },
    { "midiChannel", [this](String name) { m_pSettings->midiChannel = params[name].toInt(); },
    ...
};
// Iterate through every defined setter in your map
for (const auto& [name, setter] : setters) 
{
    // Check if the incoming params collection actually has this key
    if (params.contains(name)) 
    {
        // Execute the lambda: [this](String n) { ... }
        setter(name);
    }
}
```
(And it followed Jamulus coding style better than I did...)

src/sound/soundbase.cpp
```diff
@@ -235,8 +236,72 @@ QVector<QString> CSoundBase::LoadAndInitializeFirstValidDriver ( const bool bOpe
 /******************************************************************************\
 * MIDI handling                                                                *
 \******************************************************************************/
+
+void CSoundBase::ParseMIDICommandLineParams ( const QString& strMIDISetup, CClientSettings& Settings )
```
pljones: Yeah, sorry -- move to CSettings.

src/sound/soundbase.cpp
```diff
@@ -235,8 +236,72 @@ QVector<QString> CSoundBase::LoadAndInitializeFirstValidDriver ( const bool bOpe
...
+                {
+                    QStringList slP         = sParm.mid ( 1 ).split ( "*" );
+                    Settings.midiMuteMyself = slP[0].toInt();
+                }
+            }
+        }
+        Settings.bUseMIDIController = true;
+    }
+}
 void CSoundBase::ParseCommandLineArgument ( const QString& strMIDISetup )
```
pljones: I think you need to make sure there's only one way of doing this.  This existing method is "the one true agreed way" to parse the command line.

Ideally it needs to move to CSettings and only update the values you're now storing.

Then CClient should tell CSound what to do when it initialises:
- set up the initial MIDI mapping
- check whether MIDI needs to be enabled and, if so, turn it on

The UI and RPC should tell CClient what's going on and it should update CSettings - the UI and RPC shouldn't directly update Settings -- CSettings is meant to be fairly dumb with any logic in CClient or CServer.  Currently, that's not generally how things are done though.

src/sound/soundbase.cpp
```diff
@@ -416,3 +483,13 @@ void CSoundBase::ParseMIDIMessage ( const CVector<uint8_t>& vMIDIPaketBytes )
         }
     }
 }
+
+void CSoundBase::SetMIDIMapping ( const QString& strMIDISetup )
+{
+    // Parse the MIDI mapping
+    ParseCommandLineArgument ( strMIDISetup );
```
pljones: I don't like passing this around as a string.

pljones: Actually, with the proposed change above, the whole thing can go.

src/main.cpp
```diff
@@ -920,25 +920,31 @@ int main ( int argc, char** argv )
...
         if ( bIsClient )
         {
-            // Client:
-            // actual client object
+            // Create client with empty MIDI string initially (safer initialization)
             CClient Client ( iPortNumber,
                              iQosNumber,
                              strConnOnStartupAddress,
-                             strMIDISetup,
+                             "", // Always start with empty MIDI
```
pljones: So change the method signature.

src/main.cpp
pljones: See other comment about this.

src/settings.cpp
```diff
@@ -200,6 +200,124 @@ void CSettings::PutIniSetting ( QDomDocument& xmlFile, const QString& sSection,
...
+        }
+        else if ( cType == 'o' )
+        {
+            iMidiMuteMyself = iFirst;
+        }
+    }
+
+    bUseMIDIController = true;
+}
+
```
pljones, line 352: Above should be inside the `#ifndef SERVER_ONLY` as they're client settings.

src/settings.cpp
```diff
@@ -829,16 +939,16 @@ void CClientSettings::WriteSettingsToXML ( QDomDocument& IniXMLDocument, bool is
 QString CClientSettings::GetMIDIMapString() const
```
pljones: Where will this be needed now?

src/main.cpp
```diff
@@ -535,17 +534,10 @@ int main ( int argc, char** argv )
...
-                                 "--ctrlmidich", // no short form
-                                 "--ctrlmidich",
-                                 strArgument ) )
+        // MIDI
+        if ( GetStringArgument ( argc, argv, i, "", "--ctrlmidich", strArgument ) )
         {
-            strMIDISetup = strArgument;
-            qInfo() << qUtf8Printable ( QString ( "- MIDI controller settings: %1" ).arg ( strMIDISetup ) );
-            CommandLineOptions << "--ctrlmidich";
+            CommandLineOptions << ( QString ( "--ctrlmidich=" ) + strArgument );
```
pljones: Why is this changed?

src/main.cpp
```diff
@@ -920,26 +912,13 @@ int main ( int argc, char** argv )
...
             Settings.Load ( CommandLineOptions );
-
-            // Parse command line MIDI parameters if provided
-            CSoundBase::ParseMIDICommandLineParams ( strMIDISetup, Settings );
-            if ( Settings.bUseMIDIController )
-            {
-                Client.ApplyMIDIMapping ( Settings.GetMIDIMapString() );
-            }
+            Client.SetSettings ( &Settings );
+            Client.ApplyMidiSettingsFromConfig();
```
pljones: Why isn't done by `Client.SetSettings`?

src/settings.cpp
```diff
@@ -611,6 +699,28 @@ void CClientSettings::ReadSettingsFromXML ( const QDomDocument& IniXMLDocument,
 
     // fader settings
     ReadFaderSettingsFromXML ( IniXMLDocument );
+
+    // Apply command-line MIDI parameters if present (overwrite .ini values)
+    for ( const QString& option : CommandLineOptions )
```
pljones: Move this up above the reading from the XML and only read from the XML if the command line option wasn't passed?

src/main.cpp
pljones: OK, got it -- you need the value in settings.cpp.  Mmm, use `QString ( "--ctrlmidich=%1" ).arg ( strMIDISetup )` for consistency, at least (and then you don't need the outer `( ... )` around it).

src/client.h
```diff
@@ -301,9 +300,18 @@ class CClient : public QObject
     CChannelCoreInfo ChannelInfo;
     QString          strClientName;
 
-    void ApplyMIDIMapping ( const QString& midiMap );
+public:
+    // Assign settings pointer
+    void SetSettings ( CClientSettings* settings ) { pSettings = settings; }
+
+    // Apply MIDI settings from config
+    void ApplyMidiSettingsFromConfig();
```
pljones: Move it all into SetSettings (and move SetSettings into client.cpp of course).  You won't need that `if ( *pSettings )` then, as you'll see `pSettings = settings` just having been done.

src/main.cpp
```diff
@@ -920,26 +912,13 @@ int main ( int argc, char** argv )
...
```
pljones, line 930: Unrelated rant: this is why it's stupid to load the settings file _after creating the client_!  If it had been done first - ideally letting the same "settings" code process the command line options - the resultant settings could be passed to the client.

(Why Settings is passed to ClientDlg when Client should own state, I don't know either... but ... later, later...)

src/settings.cpp
ignotus666: Done.

src/settings.cpp
ignotus666: Oops. Removed.

src/main.cpp
ignotus666: Done.

src/main.cpp
ignotus666: Done.

src/settings.cpp
ignotus666: Done.

src/client.h
ignotus666: Done.

src/settings.cpp
```diff
@@ -200,6 +200,8 @@ void CSettings::PutIniSetting ( QDomDocument& xmlFile, const QString& sSection,
     xmlKey.appendChild ( currentValue );
 }
 
+#ifndef SERVER_ONLY
+
 // Parse MIDI commmand line parameters and update MIDI variables
 void CSettings::ParseCtrlMidiCh ( const QString& strMidiMap,
```
pljones: So it should be `CClientSettings::....`

src/client.h
pljones: I still don't see why two methods are needed.  Why can't SetSettings just do it all?

src/sound/soundbase.cpp
```diff
@@ -23,6 +23,7 @@
 \******************************************************************************/
 
 #include "soundbase.h"
+#include "../settings.h"
```
pljones: If this is needed, it should be included by `soundbase.h`

src/sound/soundbase.h
```diff
@@ -62,15 +62,18 @@ class CMidiCtlEntry
 };
 
 /* Classes ********************************************************************/
+
+#ifndef SERVER_ONLY
```
pljones: Is this still needed?

src/clientsettingsdlg.cpp
```diff
@@ -397,6 +398,43 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
...
+                                   "the “Mute Myself” parameter - it only requires a single CC number as it is only "
+                                   "applied to one’s own audio stream." ) +
+                              "<br>" +
+                              tr ( "You can either type in the MIDI CC values or use the \"Learn\" button: click on "
+                                   "\"Learn\", actuate the fader/knob/button on your MIDI controller, and the MIDI CC "
+                                   "number will be detected and saved." );
+
+    lblChannel->setWhatsThis ( strMidiSettings );
+    lblMuteMyself->setWhatsThis ( strMidiSettings );
+    faderGroup->setWhatsThis ( strMidiSettings );
```
pljones: Could you have a quick check through -- or get Github CoPilot to -- to check you're following the naming conventions for all these things.  There are other widgets elsewhere you can look at.

src/main.cpp
```diff
@@ -920,25 +917,17 @@ int main ( int argc, char** argv )
...
-            // actual client object
-            CClient Client ( iPortNumber,
-                             iQosNumber,
-                             strConnOnStartupAddress,
-                             strMIDISetup,
-                             bNoAutoJackConnect,
-                             strClientName,
-                             bEnableIPv6,
-                             bMuteMeInPersonalMix );
+            CClient Client ( iPortNumber, iQosNumber, strConnOnStartupAddress, bNoAutoJackConnect, strClientName, bEnableIPv6, bMuteMeInPersonalMix );
```
pljones, line 924: I guess (again: not now, part of my earlier rant) this should be replaced with
```
CClient Client ( &Settings );
```
where Settings has already had `iPortNumber, iQosNumber, strConnOnStartupAddress, bNoAutoJackConnect, strClientName, bEnableIPv6, bMuteMeInPersonalMix` made available from `CommandLineOptions` like `--ctrlmidich` was.  I still can't see why moving the settings load before creating client or server would be bad or hard...

src/settings.cpp
pljones: Can you go through and made sure everything that's "client only" is in the "client" parts of the code (i.e. move it within `#ifndef SERVER_ONLY` or put `#ifndef SERVER_ONLY` in if there isn't any and it's a shared file).

src/client.h
ignotus666: I take it you mean that instead of calling `pClient->ApplyMidiSettingsFromConfig();`, we call `pClient->SetSettings(pSettings);` and let that take care of applying MIDI settings? (And get rid of `ApplyMidiSettingsFromConfig();` altogether.)

pljones: Exactly.

pljones:
Yes, I agree.  Once the MIDI Tab exists, it should always appear.  If it's not available because Jamulus cannot detect a MIDI port, the check box to enable the feature should be disabled with a message that makes it clear that all Jamulus knows is that it can't detect a MIDI port.

Are the existing errors _issued by Jamulus_ (i.e. as part of trying to open the port)?  If so, that's probably all that needs to be logged.  For the UI, just indicate there's a problem opening the port, uncheck and disable the checkbox.

src/client.h
ignotus666: Done.

src/settings.cpp
ignotus666: Done.

src/sound/soundbase.cpp
ignotus666: It wasn't needed any more - it was left over from a previous iteration.

src/sound/soundbase.h
ignotus666: Nope - gone.

src/clientsettingsdlg.cpp
ignotus666: Good catch - changed to follow existing style. Also caught some other things while looking at this.

src/sound/asio/sound.cpp
```diff
@@ -633,7 +633,7 @@ void CSound::EnableMIDI ( bool bEnable )
         if ( !bMidiEnabled && iCtrlMIDIChannel != INVALID_MIDI_CH )
         {
             Midi.MidiStart();
-            bMidiEnabled = true;
+            bMidiEnabled = Midi.IsActive(); // Only set to true if MIDI actually started
```
pljones: Maybe put this after the `if / else` and get rid of the other assign?

src/clientsettingsdlg.cpp
```diff
@@ -746,6 +784,65 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
...
+        pClient->SetSettings ( pSettings );
+
+        // Check if MIDI was actually enabled successfully
+        if ( checked && !pClient->IsMIDIEnabled() )
+        {
+            // MIDI failed to enable - uncheck the box and update settings
+            pSettings->bUseMIDIController = false;
+            chbUseMIDIController->setChecked ( false );
+            SetMIDIControlsEnabled ( false );
+            QMessageBox::warning ( this, tr ( "Could not open MIDI port" ), tr ( "Please check your OS configuration." ) );
```
pljones, line 896: Can the JSONRPC code return an error indication in this situation?

src/clientsettingsdlg.cpp
```diff
@@ -746,6 +784,65 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
...
+        QObject::connect ( mapping.spinBox, static_cast<void ( QSpinBox::* ) ( int )> ( &QSpinBox::valueChanged ), this, [this, mapping] ( int v ) {
+            pSettings->*( mapping.member ) = v;
+            pClient->SetSettings ( pSettings );
+        } );
+    }
+
+    QObject::connect ( chbUseMIDIController, &QCheckBox::toggled, this, [this] ( bool checked ) {
+        pSettings->bUseMIDIController = checked;
+        SetMIDIControlsEnabled ( checked );
+        pClient->SetSettings ( pSettings );
```
pljones, line 874: Only do this on success?

src/clientsettingsdlg.cpp
```diff
@@ -746,6 +784,65 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
...
+        if ( checked && !pClient->IsMIDIEnabled() )
+        {
+            // MIDI failed to enable - uncheck the box and update settings
+            pSettings->bUseMIDIController = false;
+            chbUseMIDIController->setChecked ( false );
+            SetMIDIControlsEnabled ( false );
+            QMessageBox::warning ( this, tr ( "Could not open MIDI port" ), tr ( "Please check your OS configuration." ) );
+        }
+
+        emit MIDIControllerUsageChanged ( pSettings->bUseMIDIController );
```
pljones: Only do this if it changed successfully?

src/sound/asio/sound.cpp
ignotus666: Done.

src/clientsettingsdlg.cpp
ignotus666: Done.

src/clientsettingsdlg.cpp
ignotus666: Done.

src/clientsettingsdlg.cpp
ignotus666: Done.

docs/JSON-RPC.md
```diff
@@ -271,7 +271,7 @@ Results:
 
 | Name | Type | Description |
 | --- | --- | --- |
-| result | string | Always "ok". |
+| result | string | "ok" on success, or error message if MIDI failed to enable. |
```
pljones: It looks like it's either `{"result":"ok"}` or `{"error":"...msg..."}`, rather than `"result"` in both cases.

src/clientsettingsdlg.cpp
pljones: Actually, I see why it can't be done now -- you need to _try_ to apply the settings before you can tell whether it failed.

src/settings.cpp
```diff
@@ -461,6 +580,59 @@ void CClientSettings::ReadSettingsFromXML ( const QDomDocument& IniXMLDocument,
         pClient->SetAudioQuality ( static_cast<EAudioQuality> ( iValue ) );
     }
 
+    // MIDI settings: check command line first, then fall back to XML
+    bool bMidiFromCommandLine = false;
+    for ( const QString& option : CommandLineOptions )
+    {
+        if ( option.startsWith ( "--ctrlmidich=" ) )
```
ann0see, line 663: I don't quite like manual parsing like this here (but it might be acceptable)

pljones: For 3.x it's okay.

For 4.x, it'll need to change completely.  The current command line / settings system needs an overhaul.

But I'm happy with this as a work-around.

docs/JSON-RPC.md
ignotus666: Done.

ignotus666:
Fixed the command line override - settings not included in the command line options are not altered if they were set previously in the GUI. I also fixed a bug where the MIDI device was not getting picked up at launch when set with the 'd' option.

src/settings.cpp
```diff
@@ -201,6 +201,125 @@ void CSettings::PutIniSetting ( QDomDocument& xmlFile, const QString& sSection,
...
+                                        int&           iMidiFaderCount,
+                                        int&           iMidiPanOffset,
+                                        int&           iMidiPanCount,
+                                        int&           iMidiSoloOffset,
+                                        int&           iMidiSoloCount,
+                                        int&           iMidiMuteOffset,
+                                        int&           iMidiMuteCount,
+                                        int&           iMidiMuteMyself,
+                                        bool&          bUseMIDIController,
+                                        QString*       strMIDIDevice )
```
pljones, line 224: Shouldn't this be a `QString&` so it's safely settable?  I can't remember the syntax well enough to be clear but writing to a pointer feels dangerous.  Of course, if this is just the old code moved... I guess it works...

src/settings.cpp
```diff
@@ -461,6 +580,59 @@ void CClientSettings::ReadSettingsFromXML ( const QDomDocument& IniXMLDocument,
...
+            bMidiFromCommandLine = true;
+            break;
+        }
+    }
+
+    if ( !bMidiFromCommandLine )
+    {
+        if ( GetNumericIniSet ( IniXMLDocument, "client", "midichannel", 0, 16, iValue ) )
+            iMidiChannel = iValue;
+
```
pljones: Why doesn't the MIDI device round trip to settings?

src/settings.cpp
```diff
@@ -754,6 +925,19 @@ void CClientSettings::WriteSettingsToXML ( QDomDocument& IniXMLDocument, bool is
     // Settings Tab
     SetNumericIniSet ( IniXMLDocument, "client", "settingstab", iSettingsTab );
 
+    // MIDI settings
```
pljones, line 1010: ...Why doesn't the MIDI device round trip to settings?

src/settings.cpp
ignotus666: When used it's checked before dereferencing:
```
if ( cType == 'd' )
{
    if ( strMIDIDevice != nullptr )  // Check before writing
    {
        *strMIDIDevice = sParm.mid ( 1 );
    }
    continue;
}
```
So it should be safe.

src/settings.cpp
ignotus666: Do you mean why isn't it saved? There's no way to set it other than by using it on the command line, it's a per-session option.

ignotus666:
> this GUI implementation would need to provide the equivalent of dDevicename (for Windows), and the settings from the ini file would need to be applied before the MIDI devices were opened

Settings are applied before opening MIDI devices.

I've come up with this (not finished yet but mostly working):

- A MIDI device list dropdown that's only visible in the non-Jack Windows version. I think if it's shown for any other platform (but doesn't really do anything) it will just lead to confusion.
- If no devices are connected it will show a warning message and disable the MIDI-in checkbox. If a device was previously connected and saved, but is not connected at launch, a message will pop up warning of this and that it will attempt to connect to "All devices". If none are connected, the "Failed to connect to MIDI port" message will appear and disable MIDI settings.
- It's possible to pick a different device at runtime. It will disconnect and reconnect to the chosen device. If a MIDI device is connected during runtime (where none were connected before), upon enabling the MIDI-in checkbox it will rescan for devices and list those available.
MIDI_devices

Is this something you guys would agree to, or how should I proceed?

src/settings.cpp
pljones: I'm more worried about:
```
QString str = QString("abc");

QString* _str = &str;

*_str = QString("really long string");
```
That's like trying `QString("abc") = QString("really long string")` isn't it?

Saying
```
QString str = QString("abc");

QString& _str = &str;

&_str = QString("really long string");
```
gets you `str == QString("really long string")`, I think.  But, like I say, I'd have to run it to check (or see what the compiler says).

src/settings.cpp
pljones: Ah... right, so as there's only the command line and that prevents reading or writing the settings, it's not there.

Mmm, I really would like it to be in the UI in that case, with a drop-down...  But then I'd want it cross-platform... and that's not for this change.  So OK, this'll do for now.

Please feel free to do more work here 😄 !

ignotus666:
> I can't see _why_ it fails on yours, though... Try clearing any worker caches you've got in your repo.

I've been holding back on adding the latest MIDI device selection code to this PR pending a go-ahead by you guys. If it's ok by you I'll add it and I'm fairly confident the iOS build won't fail.

pljones:
I think the default "count" was also zero, though?  Anyway, please push what's here to the PR (do rebase and squash if needed).  I don't think you can put "--" into the spinner (it would look odd) but on the heading line you could add a checkbox (it doesn't need a visible label) and have that enable the settings for that control? (Breaks your panel-based approach, though.)

If you do change the layout, you could change "MIDI-in" to "MIDI Control" and move the device selector next to it, labelled "Device", perhaps, to save a bit of space. (Or are MIDI device names long sometimes?)

pljones:
Here's what Gemini made of it:

[UX by Gemini.pdf](https://github.com/user-attachments/files/25315876/UX.by.Gemini.pdf)

Full text of prompt:
>  Attached is a screen shot of a new Settings tab being developed for Jamulus. It allows the user to configure through the UI settings previously only available from the command line (the `--ctrlmidich` values). The current design has all controls enabled by default, sharing CC0 with a count of 1. Two proposals have been made:
> - add an explicit "control active" checkbox (e.g. change the label for the panel to be a checkbox with the same label as the panel title)
> - use an "invalid value" to indicate the control isn't active (-1 for Mute Myself and a Count of 0 for the others)
>
> Which would the majority of users find easier to understand? 

It sounds like Gemini thinks there's a "Checkable Groupbox" -- but AI can make mistakes (and I wouldn't trust Gemini with coding, to be honest).

ignotus666:
Fair enough, it makes good points. Will get to work on it, and yes, checkable groupboxes are a thing. I'll make one for Mute Myself too to make it consistent.

iOS build succeeded BTW.

pljones:
> This branch is [30 commits ahead of](https://github.com/ignotus666/jamulus/compare/jamulussoftware%3Ajamulus%3Amain...refs/heads/midi-gui-learn2) and [10 commits behind](https://github.com/ignotus666/jamulus/compare/refs/heads/midi-gui-learn2...jamulussoftware%3Ajamulus%3Amain) jamulussoftware/jamulus:main.

Yes, if you update to latest on main from upstream and then rebase, that might fix your iOS build.

Mostly because of
> [Run prebuild scripts on iOS](https://github.com/ignotus666/jamulus/commit/9b1b7f6b8c8e37060f85009c7cbccc6eef25c9db)

I'd imagine.

ignotus666:
Ok, done. Each control can be enabled/disabled and omitting them from the command line disables and resets them to default values (overwriting the ini file). This was the behaviour before, but maybe just disabling them is enough?

Only tested in Linux, but seems to work ok. Here's what it looks like now:

tab

There are quite a few comments in the code to help me keep track of what's going on.

pljones:
image 
image

Windows - JACK and MME versions.  Both run with `--ctrlmidich "1;f85*3;p94*3;dnanoKONTROL"`, which was unfortunate as JACK doesn't pick the MME names up.  I think it also doesn't reliably associate `capture_N` with any particular port, so it's a bit hit and miss.  But it looks like it would work, anyway!

As you can see, though, the spinners are much too small -- they all let only a single digit show.  That's a Qt issue, though -- Qt 6 widgets aren't the same as Qt 5.

pljones:
image

Qt 6.10.2 on Linux looks fine.  It's ridiculous that Qt can't manage the UI tools so you can write the layout and _know_ it's going to render okay cross platform.  I mean, that's the point of Qt...

ignotus666:
> If the command line is specified at all, then nothing should be written back to settings, IIRC? Not that there's any examples for the client but I think the server works like that (if you run the GUI and specify a Directory to register with, then change it in the UI, I don't think it gets updated in settings). As I've mentioned, there's a lot of discussion about how that should work...
> 
> To me, if I've set everything in the UI and then as a one off run with `--ctrlmidich` from the command line, I don't want the settings I did in the UI changed.
> 
> And I probably want the saved settings for any `--ctrlmidich` value not specified to be the UI defaults.

I agree. But after having a look, the server settings actually behave more or less the way the MIDI settings do now: they skip reading the ini file if the command line option is present and later write any command line-specified values to the ini file later. What to do with non-specified controls in MIDI commands is a different matter though because AFAICT it's the only command line option that has a number of optional parameters within it, so there's not really a precedent for that scenario. I agree that they should be left alone and at most just disabled, not overwritten with defaults. Though this requires changing the current approach where the command line is checked first, to reading the xml and then checking the command line. The latest version does this:
- Any setting specified on the command line overrides what's in the ini file and is saved as such.
- Settings not specified are disabled but their values are preserved. Their "disabled" status is written to the ini file.
In short - launching and setting values with the command line is treated as though done through the GUI, and if you omit a control parameter, it's assumed you don't wish to use it.

I also fixed a bug where the "Learn" buttons for disabled controls would become active after using Learn for an active control.

> As you can see, though, the spinners are much too small -- they all let only a single digit show.

It's those stupid up/down arrows that take up so much space. Please check the latest version to see if it's fixed.

Oh, and another thing - please test launching a server. After years of being able to launch and connect my own, my bl**dy ISP has changed its policy and now I'm behind CGNAT and no IPv6 available, so I can't test that. It should be fine, but just in case.

pljones:
> Oh, and another thing - please test launching a server. After years of being able to launch and connect my own, my bl**dy ISP has changed its policy and now I'm behind CGNAT and no IPv6 available, so I can't test that. It should be fine, but just in case.

I've no IPv6, either.  IPv4 registered okay with Any Genre 1 and one of my local Directories.  Local connection worked fine.

----

I'm seeing
```
QThreadStorage: entry 3 destroyed before end of thread 0x6520f4497380
QThreadStorage: entry 2 destroyed before end of thread 0x6520f4497380
QThreadStorage: entry 1 destroyed before end of thread 0x6520f4497380
```
from the client (Audio, UI and Client threads, I think) and
```
QThreadStorage: entry 2 destroyed before end of thread 0x5b1062a0d380
QThreadStorage: entry 1 destroyed before end of thread 0x5b1062a0d380
```
from the server (UI and Server threads) when the UI shuts down normally.  @softins any idea what this is about?

pljones:
image

Stupid widget.  It's probably Windows rather than Qt, to be fair -- I think Qt tries to use the underlying platform widgets.  But you do end up not having any idea what's going to render okay.  The spinner still isn't wide enough...

I'm not sure why the Mute Myself channel doesn't align - I guess it's padding to available space.

Do you try GridLayout?  You can stick one inside a Checkable Groupbox (i.e. make the "MIDI-in" a group box around everything, replacing the one around the set of learn-ables) with a blank set of cells between each "row".  You can span just by dragging a widget and set alignment on a cell.

Otherwise vertical and horizontal alignment end up a bit hit and miss (see how "Musician Profile" and "User Interface" end up looking so clumsy on the "My Profile" tab)

softins:
> I'm seeing
> ```
> QThreadStorage: entry 3 destroyed before end of thread 0x6520f4497380
> QThreadStorage: entry 2 destroyed before end of thread 0x6520f4497380
> QThreadStorage: entry 1 destroyed before end of thread 0x6520f4497380
> ```
> from the client (Audio, UI and Client threads, I think) and
> ```
> QThreadStorage: entry 2 destroyed before end of thread 0x5b1062a0d380
> QThreadStorage: entry 1 destroyed before end of thread 0x5b1062a0d380
> ```
> from the server (UI and Server threads) when the UI shuts down normally.  @softins any idea what this is about?

I've seen it too. I think it's been around for a while, but as it only happens when exiting, it doesn't seem to be a problem. I have never investigated the cause.

ignotus666:
> Do you try GridLayout? You can stick one inside a Checkable Groupbox (i.e. make the "MIDI-in" a group box around everything, replacing the one around the set of learn-ables) with a blank set of cells between each "row". You can span just by dragging a widget and set alignment on a cell.

Followed this suggestion and I think it's better now. Please check on Windows 11 to see if we're there yet with the spinboxes (I added minimum width).

> The ideal app never discards anything you do

It's tricky to find a balance between consistency with current behaviour (command line overrides and overwrites previous settings) and not discarding previously saved (but not currently used) settings, so I think a compromise where omitted controls on the command line are disabled (not including them is taken as a wish not to use them) but values are preserved is acceptable.

ignotus666:
> I'd open a support ticket with the ISP saying something like your "connectivity is broken for IPv6 and you need it"

Rang them this morning. A customer for 20+ years and they wanted more money on top of an already extortionate price, so I talked to another ISP that will disable CGNAT and charge me 1/3 for the same package... Silly me for not checking sooner but at least a nice side-effect of this PR : )

pljones:
OK, how about getting rid of the overall "[x] MIDI-in" checkbox and using the interface widget.

Add a "Do not use MIDI In" that, when selected, it disables the rest of the UI.  This is then equivalent to leaving out `--ctrlmidich` if you never use it.

You've got "All inputs", which is the same as using `--ctrlmidich` without `;d`

And you've got the individual values, matching use of explicit `;d`.

Omitting `--ctrlmidich` no longer stops you turning it on through the UI and having state saved.

But we get to the question:
- What if I have saved state and supply `--ctrlmidich`?

ignotus666:
> OK, how about getting rid of the overall "[x] MIDI-in" checkbox and using the interface widget.
> 
> Add a "Do not use MIDI In" that, when selected, it disables the rest of the UI. This is then equivalent to leaving out --ctrlmidich if you never use it.

But isn't that essentially the same as what we've got now, only instead of disabling the MIDI-in checkbox, we enable a different one called "Do not use MIDI in"?

> 
> If the saved state was "Do not use MIDI In", you set the UI to that and don't populate the list further until the  drop down is opened. Only then do you evaluate the available devices
> For any other value, you go off and get the list, either attaching all the device the the one specified

That's the current behaviour. If MIDI-in is saved as disabled, it won't scan for any devices upon launching. It'll do that when you enable MIDI-in or click on the dropdown.

> You've got "All inputs", which is the same as using --ctrlmidich without ;d

That's only true for non-Jack Windows. The rest will either try the saved device (if there is one and it's connected) or revert to "No device connected" and let you choose from the dropdown/use a connection manager.

> So yes, if you start messing with command line parameters, it should mean your inifile is safe. Including safe from changes made through the UI to parameters passed on the command line.

But this isn't consistent with what happens to e.g. server settings. If you set a server option on the command line, it's saved, the same as if it were through the GUI. If I have delay panning off, launch it with `jamulus -s -P` and close it, the next time I open it delay panning will be on. As I commented somewhere above, `--ctrlmidich` is a bit weird in that it has a number of optional parameters within it. I still think that in order to maintain consistency, the parameters passed to it should stick, and a compromise (to not lose settings not passed) is to just disable those omitted but preserve their values.

Anyway, give the last build a whirl. I have the feeling we're overthinking this.

ann0see:
Just to check up. Do we want this in the next release or not?
If so, we'll need another translation round.

pljones:
> > Do not use MIDI in
> 
> Sounds quite unintuitive to me

You frequently see
| | |
|-|-|
| **Selected MIDI Input:** | [ None ] |

in applications.  The checkbox and list together ... well, I can't think of any example.

softins:
I'm just doing some testing here, on Windows, using a _nanoKONTROL2_. There seems to be an interaction between the first fader and the **Mute Myself** operation.

I'm starting Jamulus using `release\Jamulus --ctrlmidich "1;f0*8;p16*8;s32*8;m48*8;dnanoKONTROL2"`

All works ok if I build `main`. But if I build this PR and run it, moving the first fader causes **Mute Myself** to turn on and off. I haven't tried any "learn" functions yet, but note that the command line above doesn't include a setting for **Mute Myself**. I suspect it is defaulting to 0, conflicting with fader 0. Maybe it should default to -1 or something if not specified. I haven't yet looked at the code.

src/client.cpp
```diff
@@ -193,6 +196,29 @@ CClient::CClient ( const quint16  iPortNumber,
...
+    Sound.SetCtrlMIDIChannel ( pSettings->iMidiChannel );
+    Sound.SetMIDIControllerMapping ( pSettings->iMidiFaderOffset,
+                                     pSettings->bMidiFaderEnabled ? pSettings->iMidiFaderCount : 0,
+                                     pSettings->iMidiPanOffset,
+                                     pSettings->bMidiPanEnabled ? pSettings->iMidiPanCount : 0,
+                                     pSettings->iMidiSoloOffset,
+                                     pSettings->bMidiSoloEnabled ? pSettings->iMidiSoloCount : 0,
+                                     pSettings->iMidiMuteOffset,
+                                     pSettings->bMidiMuteEnabled ? pSettings->iMidiMuteCount : 0,
+                                     pSettings->bMidiMuteMyselfEnabled ? pSettings->iMidiMuteMyself : 0 );
```
softins:
```suggestion
                                     pSettings->bMidiMuteMyselfEnabled ? pSettings->iMidiMuteMyself : -1 );
```

Simple fix for the issue above. Tested on my machine.

ignotus666:
> No it didn't. `iMidiMuteMyself` is already qualified by `bMidiMuteMyselfEnabled`. It is only the value passed in to `SetMIDIControllerMapping()` that needs to be `-1` when mute myself is not enabled.

Oops - you're right. Undid that change.

softins:
I've also compiled and tested it successfully on macOS. Although the checkboxes appear to be smaller than normal.

softins (approved):

src/settings.cpp
ann0see: @ignotus666

ignotus666: @ann0see the hypothetical example is wrong:

`QString& _str = &str;` is not correct - &str is a pointer, not a reference.
`&_str = QString("really long string");` is invalid too - you can’t assign to an address.

It defaults to nullptr if the caller omits that argument (or passes nullptr), and then the null check makes it safe.

pljones: > `&_str = QString("really long string");` is invalid too - you can’t assign to an address.

Ah!  That's what I was thinking of.  Thanks for clearing it up.

pljones (approved):

ann0see:
@ignotus666 I assume this is fully documented on the website?

ignotus666:
@ann0see yes, I think it's all been covered.

Issue #3395, opened 2024-10-02
Windows 11 dark mode title does not show in correct color


ann0see:
**Describe the bug**

After we updated Qt in https://github.com/jamulussoftware/jamulus/pull/3262 Windows now supports a dark mode. However, the server title does not show up in the correct color:

https://github.com/jamulussoftware/jamulus/blob/08b404c40d617e2c1d73f9a4c0adcea0a3c19be1/src/clientdlg.cpp#L1471 needs to be adapted most likely.

**To Reproduce**

Download latest build from main.
Install on Windows 11 with dark mode enabled
Connect to server, see title in wrong color.
Disconnect from server

**Expected behavior**

The server title should have a visible color.

**Screenshots**

![dark mode win11 issue](https://github.com/user-attachments/assets/8fc40d6e-d074-442e-aa2c-00a90f8bd4cc)

**Operating system**

Windows 11

**Version of Jamulus**

jamulus_3.11.0dev-0e9a425 from https://github.com/ann0see/jamulus/actions/runs/11150598783?pr=160
**Additional context**

/

PR #3507, opened 2025-06-30
Add dark mode checker depending on Qt version

ann0see:
Solves darkmode color issue in mixer dialog:
Connected:
Server label in dark mode Windows connected
Disconnected:
Server label in dark mode Windows disconnected

CHANGELOG: SKIP

**Context: Fixes an issue?**

Fixes: #3395 

No

Ready for review. Working

Review

## Checklist

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -1479,9 +1479,16 @@ void CClientDlg::SetMixerBoardDeco ( const ERecorderState newRecorderState, cons
         }
         else
         {
-            if ( palette().color ( QPalette::Window ) == QColor::fromRgbF ( 0.196078, 0.196078, 0.196078, 1 ) )
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 5, 0 )
+            // for Qt 6.5.0 or later, we use the inbuilt cross platform color scheme picker.
+            if ( QGuiApplication::styleHints()->colorScheme() == Qt::ColorScheme::Dark )
+#else
+            // for earlier versions, check darkmode as proposed in https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5
+            const QPalette defaultPalette;
+            if ( defaultPalette.color ( QPalette::WindowText ).lightness() > defaultPalette.color ( QPalette::Window ).lightness() )
+#endif
             {
-                // Dark mode on macOS/Linux needs a light color
+                // Dark mode needs a light color
 
                 sTitleStyle += "color: rgb(220,220,220); }";
             }
--- a/src/clientdlg.h
+++ b/src/clientdlg.h
@@ -42,6 +42,9 @@
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 6, 0 )
 #    include <QVersionNumber>
 #endif
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 5, 0 )
+#    include <QStyleHints>
+#endif
 #include "global.h"
 #include "util.h"
 #include "client.h"


src/clientdlg.h
```diff
@@ -42,6 +42,12 @@
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 6, 0 )
 #    include <QVersionNumber>
 #endif
+#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
```
ann0see: To be checked if all of this is really needed. 
I assume that QPalette was already imported.

src/clientdlg.cpp
```diff
@@ -1479,7 +1479,18 @@ void CClientDlg::SetMixerBoardDeco ( const ERecorderState newRecorderState, cons
         }
         else
         {
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 5, 0 )
+            // for Qt 6.5.0 or later, we use the inbuilt cross platform color scheme picker.
+            if ( QGuiApplication::styleHints()->colorScheme() == Qt::ColorScheme::Dark )
+#else
+    #if defined( Q_OS_WINDOWS )
```
ann0see: The 32 bit Windows build might fail. If this is the case, we might want to guard this with a version check too.

ann0see:
macOS legacy and non legacy seem to work as expected.

ann0see:
Windows seems to work now too.

ann0see:
Light mode might have a slightly wrong color for the line surrounding the mixer board:

border.

But I suppose this is non blocking. Otherwise, I think this is good as is.

pljones:
I'd have a change log entry - helps search engines help any looking up the issue find "use the latest version" should fix it.

src/clientdlg.cpp
```diff
@@ -1479,7 +1479,16 @@ void CClientDlg::SetMixerBoardDeco ( const ERecorderState newRecorderState, cons
...
         {
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 5, 0 )
+            // for Qt 6.5.0 or later, we use the inbuilt cross platform color scheme picker.
+            if ( QGuiApplication::styleHints()->colorScheme() == Qt::ColorScheme::Dark )
+#elif defined( Q_OS_WINDOWS )
+            // for windows only, check darkmode as proposed in https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5
+            const QPalette defaultPalette;
+            if ( defaultPalette.color ( QPalette::WindowText ).lightness() > defaultPalette.color ( QPalette::Window ).lightness() )
+#else
             if ( palette().color ( QPalette::Window ) == QColor::fromRgbF ( 0.196078, 0.196078, 0.196078, 1 ) )
```
pljones: Can this not also compare against a threshold?  (Somewhere halfway between the two tones in use.)

src/clientdlg.cpp
```diff
@@ -1479,7 +1479,16 @@ void CClientDlg::SetMixerBoardDeco ( const ERecorderState newRecorderState, cons
         }
         else
         {
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 5, 0 )
+            // for Qt 6.5.0 or later, we use the inbuilt cross platform color scheme picker.
+            if ( QGuiApplication::styleHints()->colorScheme() == Qt::ColorScheme::Dark )
+#elif defined( Q_OS_WINDOWS )
```
pljones: Does this need a Qt version check in case someone builds by hand against Qt5?

src/clientdlg.cpp
ann0see: I would then rather use the code above. But this worked for macOS and probably Linux (?).

src/clientdlg.cpp
ann0see: At least the 32 bit Windows build which uses Qt 5.15 (I believe) didn't complain

src/clientdlg.cpp
pljones: Do non-`Q_OS_WINDOWS` platforms have `QPalette::Window`?  If they do, then we may as well use the same for all platforms if it works -- I'd rather have as few different ways of doing the same thing as possible.

We're looking to do "is colour scheme dark" -- if that means "is window text (**EDIT**: not darker) **ligher** than window background" and Qt supports getting the window text and window background colours in a way that supports the `lightness()` method, I can't think why we shouldn't do it the same on all.

(I really want to get rid of "does the window colour equal some random value that was present on someone's screen" -- most window managers allow a variety of colour schemes and this is as likely to be wrong for anyone customising the set up as it is right.)

ann0see: Yes, the link proposes the lightness comparison - so I assume that this is the Qt proposed way to go and it should work on all supported OS. I wanted to change as little behavior as possible, but yes - the new approach seems cleaner.

src/clientdlg.cpp
```diff
@@ -1479,7 +1479,14 @@ void CClientDlg::SetMixerBoardDeco ( const ERecorderState newRecorderState, cons
...
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 5, 0 )
+            // for Qt 6.5.0 or later, we use the inbuilt cross platform color scheme picker.
+            if ( QGuiApplication::styleHints()->colorScheme() == Qt::ColorScheme::Dark )
+#else
+            // for earlier versions, check darkmode as proposed in https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5
+            const QPalette defaultPalette;
+            if ( defaultPalette.color ( QPalette::WindowText ).lightness() > defaultPalette.color ( QPalette::Window ).lightness() )
+#endif
             {
                 // Dark mode on macOS/Linux needs a light color
```
pljones: Of course, this is now "Dark mode on QT before 6.5.0" rather than OS-specific.

pljones: (Also out of scope, I'd rather we were using named CSS colours rather than magic incantations (what's wrong with "[gainsboro](https://developer.mozilla.org/en-US/docs/Web/CSS/named-color)" or "[black](https://developer.mozilla.org/en-US/docs/Web/CSS/named-color)"?).)

ann0see: They are not magic at all. rgb(x,y,z) or #ffffff are clearly defined color codes. In webdev I usually also don't use keywords as e.g. the hex value is more precise. Also IDEs usually show the color directly or can be configured to do so:

color picker in VSCodium

pljones: The point is, someone reading the code has to work out in their head what colour they are.  I'd rather use "lightgray" than "gainsboro" for readability to be honest - I had to look up what the colour code was to see what (220, 220, 220) looked like and find the css name -- with "lightgray" I'd have a pretty good idea immediately.

PR #3533, opened 2025-09-07
3532: Add jamulusserver/setDirectory request

pljones:

CHANGELOG: Server RPC: Add jamulusserver/setDirectory request

**Context: Fixes an issue?**

Fixes: 3532

RPC documentation regenerated.

I've tested with nc, changing one of my servers from registered with a custom directory to unregistered and then back.

Reviewing.

Ideally, testing by @anprdev to confirm it meets their requirements.

## Checklist

-  [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#source-code-consistency) **`make clang_format` has gone wacky again**

### DIFF
--- a/src/serverrpc.cpp
+++ b/src/serverrpc.cpp
@@ -122,24 +122,70 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
     /// @result {string} result.city - The server city.
     /// @result {number} result.countryId - The server country ID (see QLocale::Country).
     /// @result {string} result.welcomeMessage - The server welcome message.
+    /// @result {string} result.directoryType - The directory type as a string (see EDirectoryType and SerializeDirectoryType).
+    /// @result {string} result.directoryAddress - The string used to look up the directory address (only assume valid if directoryType is "custom"
+    /// and registrationStatus is "registered").
     /// @result {string} result.directory - The directory with which this server requested registration, or blank if none.
     /// @result {string} result.registrationStatus - The server registration status as string (see ESvrRegStatus and SerializeRegistrationStatus).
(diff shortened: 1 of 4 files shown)


src/serverrpc.cpp
```diff
@@ -226,37 +272,64 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
     } );
 }
 
-QJsonValue CServerRpc::SerializeRegistrationStatus ( ESvrRegStatus eSvrRegStatus )
+const std::unordered_map<EDirectoryType, std::string> CServerRpc::sumDirectoryTypeToString = {
```
ann0see: Do we have a place where we already have those pre defined directories? Isn't it somewhere in util.h?

pljones: There's a function to return the translated name.  This is a serialisation of the enum, though, so it must not use the translations.

pljones:
> Did you check that the server UI responds as expected?

No - it would be the same as any other RPC-based change, though.  For example, `jamulusserver/setWelcomeMessage` ends up here:
```cxx
void CServer::SetWelcomeMessage ( const QString& strNWelcMess )
{
    // we need a mutex to secure access
    QMutexLocker locker ( &MutexWelcomeMessage );
    strWelcomeMessage = strNWelcMess;

    // restrict welcome message to maximum allowed length
    strWelcomeMessage = strWelcomeMessage.left ( MAX_LEN_CHAT_TEXT );
}
```
That won't update the UI, as there's no signal being raised.

pljones:
Mmm, that means `UpdateGUIDependencies` is missing something...  Possibly worth investigating?
```
void CServerDlg::UpdateGUIDependencies()
{
    const EDirectoryType directoryType = pServer->GetDirectoryType();
    cbxDirectoryType->setCurrentIndex ( static_cast<int> ( directoryType ) + 1 );
...
```

PR #3539, opened 2025-09-17
Optimize Images

ann0see:
Small improvements in image sizes (non critical and only parts should be applied in prod e.g. SVG should not be minified). The shrinkage is minimal - but imo still worth it. We'll need to shrink some of the images - but that's for a later day, I think.

Done by ImgBot.

CHANGELOG: Optimize images for slightly smaller binary size.


libs/opus/doc/opus_logo.svg
ann0see, line 1: Don't change this.

src/res/io.jamulus.jamulusserver.svg
ann0see, line 1: Don't change this.

src/res/instruments/bassoon.svg
ann0see, line 1: SVGs should not be changed.

src/res/HLEDRedSrc.png
ann0see, line 1: Likely too large anyway...

ann0see:
Since this is a small change, I'd like to get it into the upcoming release, if possible. @softins @pljones ?

src/res/HLEDRedSrc.png
ann0see: Ah. They are src files - so it should be fine as they shouldn't be in the app.

pljones:
The byte saving looks trivial.

The colours have changed in some cases, too.  Are they still okay in Fancy skin and normal?

(I've no idea why the LEDs need to be that big -- even on high definition displays, given the size of the other images.)

ann0see:
I think they are just the original files. So it's probably fine. 

I'll check the colors later.

softins:
> Since this is a small change, I'd like to get it into the upcoming release, if possible. @softins @pljones ?

I have no objection in principle, but I can't check anything as I am away from home for the next week.

ann0see:
For me the color differences are negligible. At least I don't really see much of a difference anywhere (macOS)

gilgongo (approved):
I've not looked at the changes on a compiled version, but would the brower-rendered versions be any different? Checking on a couple of monitors and a mobile phone shows very little change to the colours.

pljones:
New (left) vs Old:
image
Mmm... It shows but not enough to worry me.  I just don't see what benefit it brings.

Windows download
After: 84,229,859 bytes (assuming this is rebased to main)
Before: 84,314,507 bytes
Shrink: ~85K

Is it worth it?

ann0see:
In my opinion even small savings are worth it. You should think about that it scales with the number of users too: less downloaded bytes etc.

pljones:
Is it for 3.12.0? (No project or milestone assigned.)

ann0see:
I didn't tag it yet as I wanted a yes/no from others.

softins (approved):
Looks ok to me

PR #3550, opened 2025-10-22
Randomize and adapt ping intervals in connection dialog to reduce correlation

stefan1000:

To avoid correlation on client behaviour this change will randomize the ping intervals, adaptive to the ping time, nearer servers get pinged more often, distant ones less.

To avoid correlation of "hide connect dialog" (ping immediately stops) and "connected to a server" (user appears on a server) it will keep pinging for approx. 60 seconds after the dialog gets hidden.

In addition:
- introduced enum for UserRole data on LVC_NAME.
- cache host and port on first use, no need to resolve it every time (in UserRole data)

CHANGELOG: Randomized and adaptive ping intervals in connection dialog, keep ping running after some time after dialog gets closed.

**Context: Fixes an issue?**

See https://github.com/orgs/jamulussoftware/discussions/3545

No.

Unsure if the diagnostic code (currently only in debug builds) can/should be kept in.

## Checklist

### DIFF
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -143,9 +143,6 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     lvwServers->setColumnCount ( LVC_COLUMNS );
     lvwServers->hideColumn ( LVC_PING_MIN_HIDDEN );
     lvwServers->hideColumn ( LVC_CLIENTS_MAX_HIDDEN );
-#ifdef PING_STEALTH_MODE
-    lvwServers->hideColumn ( LVC_LAST_PING_TIMESTAMP_HIDDEN );
-#endif
 
     // per default the root shall not be decorated (to save space)
     lvwServers->setRootIsDecorated ( false );
@@ -166,8 +163,7 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     // setup timers
     TimerInitialSort.setSingleShot ( true ); // only once after list request
 #ifdef PING_STEALTH_MODE
-    TimerKeepPingAfterHide.setSingleShot ( true ); // single shot shutdown timer
-    iKeepPingAfterHideStartTime = 0;
+    TimerKeepPingAfterHide.setSingleShot ( true );
     iKeepPingAfterHideStartTimestamp = 0;
 #endif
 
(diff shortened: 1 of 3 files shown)


src/connectdlg.cpp
```diff
@@ -1032,3 +1185,164 @@ void CConnectDlg::UpdateDirectoryComboBox()
...
+        if ( count > avgPingsPerSecond * 2.0 ) // Spike = 2x average
+        {
+            iSpikesDetected++;
+        }
+    }
+
+    qDebug() << "\nPattern analysis:";
+    qDebug() << "  Average pings/sec:" << QString::number ( avgPingsPerSecond, 'f', 2 );
+    qDebug() << "  Spikes detected (>2x avg):" << iSpikesDetected << "seconds";
+    qDebug() << "  Distribution: " << ( iSpikesDetected > 10 ? "IRREGULAR (many spikes)" : "SMOOTH (good stealth)" );
```
mcfnord: I set up a time-out when a burst of pings arrive from one IP:PORT, because I thought these were non-clients. Then my own client triggered the time-out for bursting pings. Seems odd for the client to do this.

stefan1000:
no, I dont think so: if you have the connection dialog open the client will ping each server in the list to get the ping time, constantly every 2500ms. To record those messages you just need to register a jamulus server at the directory, and since you control the server you can capture all ping messages with their timestamp and IP. In addition, log the active users on each server via the explorer php or via your own tooling. When a user connects to a server the connection dialog will close and stop the pinging, and a new user name appears on a server. With some interval heuristics you can then correlate those events (user name with client IP).

AFAIK CLSendEmptyMes used for hole-punch is only sent once when the client requests the server list from the directory, its probably too coarse to do a correlation.

softins:
> But still the monitoring server will receive the client IP. This should not happen.

Of course it will, otherwise it could not reply to the ping. The capturing of the IP address by the correlator is happening at the pcap level by a separate program, not within Jamulus.

ann0see (commented):
I'd like to have a checkbox in advanced settings which enables/disables this. 

Also please rebase.

dingodoppelt:
@stefan1000 Do you have any opinions on #3882 and maybe something to add? Would this work for you as well?

dingodoppelt:
Is this still likely to get merged anytime or can we agree that this can be closed?

ann0see:
Honestly - I don't know. It stalled. I still think that we should enable the user to do *something*. If this is the correct way to do it - not sure. See my comments from above.

stefan1000:
Sorry for the late reply and not getting back earlier.

I'm not really sure if we should add this in its current form. I'd generally prefer to avoid adding more "technical" or "advanced" options that are difficult for non-technical users to understand and configure.

Additionally, to populate the connection dialog with latency times, it still needs to send ping messages to all those servers. Given the relatively low active user count/activity —along with the open-source nature of the randomization—it might still be possible to correlate users, especially if randomization isn't enabled by default for the vast majority of users.

Perhaps a combination of geolocation-based distance filtering and targeted pinging could work better: only pinging servers within a reasonable distance, and sending pings far less frequently overall.

pljones:
> only pinging servers within a reasonable distance, and sending pings far less frequently overall.

There's definitely a case to be made for a trail off on the interval, geared more heavily towards "further away" servers.  If a server is responding regularly at 200ms, you probably don't need to ping it more than once a minute or so.  If it's bouncing between 3ms (generally) and occasionally dropping off to 15ms, you want to know that, though.  So some fairly careful rules would need to be constructed.

PR #3556, opened 2025-11-06
Provides SRV support for connecting to directories from client or server

rdica:

Provides SRV DNS support for connecting to directories from a client or a server.
Supports `-e`/`--directoryaddress` for server and custom directories in both client and server GUI.
Also includes a small amount of refactoring for consistency and tidiness.

CHANGELOG: Client/Server: Add SRV support for connecting to directories.

**Context**

Currently one needs to provide both an IP/host and port number to the `-e|--directoryaddress` option if the directory server is not using the default port `22124`. This patch will enable the ability to use preconfigured SRV DNS records by a server to connect to a directory without having to provide a port number. It also supports the custom directories in both server and client GUI.

The patch expands on SRV support in client code already in `main`.

I have created SRV DNS records to test with that point to each of the seven public directory servers provided by the Jamulus team:
- anygenre1.jamulusjams.com
- anygenre2.jamulusjams.com
- anygenre3.jamulusjams.com
- rock.jamulusjams.com
- jazz.jamulusjams.com
- classical.jamulusjams.com
- choral.jamulusjams.com

You can confirm the SRV records using the following:
_Mac/Linux_
```
dig _jamulus._udp.anygenre1.jamulusjams.com srv

;; ANSWER SECTION:
_jamulus._udp.anygenre1.jamulusjams.com. 3600 IN SRV 0 0 22124 anygenre1.jamulus.io.
```
_Windows_
```
nslookup -type=srv _jamulus._udp.anygenre1.jamulusjams.com`

Server:  UnKnown
Address:  10.2.0.1

_jamulus._udp.anygenre1.jamulusjams.com SRV service location:
          priority       = 0
          weight         = 0
          port           = 22124
          svr hostname   = anygenre1.jamulus.io
```

In order to utilize this functionality for the Jamulus public space, the Jamulus team could create the seven SRV records and publish those in the same table that displays the server host/port pairs in https://jamulus.io/wiki/Running-a-Server#registered-mode

Unsure. According to `--help` output, the option `-c|--connect` doesn't mention anything about SRV support. In that same vein I submit nothing should be added for `-e|--directoryaddress` either.

## Checklist

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -588,12 +588,12 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // Send the request to two servers for redundancy if either or both of them
     // has a higher release version number, the reply will trigger the notification.
 
-    if ( NetworkUtil().ParseNetworkAddress ( UPDATECHECK1_ADDRESS, UpdateServerHostAddress, bEnableIPv6 ) )
+    if ( NetworkUtil::ParseNetworkAddress ( UPDATECHECK1_ADDRESS, UpdateServerHostAddress, bEnableIPv6 ) )
     {
         pClient->CreateCLServerListReqVerAndOSMes ( UpdateServerHostAddress );
     }
 
-    if ( NetworkUtil().ParseNetworkAddress ( UPDATECHECK2_ADDRESS, UpdateServerHostAddress, bEnableIPv6 ) )
+    if ( NetworkUtil::ParseNetworkAddress ( UPDATECHECK2_ADDRESS, UpdateServerHostAddress, bEnableIPv6 ) )
     {
         pClient->CreateCLServerListReqVerAndOSMes ( UpdateServerHostAddress );
     }
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -588,12 +588,14 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // Send the request to two servers for redundancy if either or both of them
     // has a higher release version number, the reply will trigger the notification.
 
-    if ( NetworkUtil::ParseNetworkAddress ( UPDATECHECK1_ADDRESS, UpdateServerHostAddress, bEnableIPv6 ) )
+    // Don't use SRV resolution when resolving update servers.
+
+    if ( NetworkUtil::ParseNetworkAddressBare ( UPDATECHECK1_ADDRESS, UpdateServerHostAddress, bEnableIPv6 ) )
     {
         pClient->CreateCLServerListReqVerAndOSMes ( UpdateServerHostAddress );
     }
 
-    if ( NetworkUtil::ParseNetworkAddress ( UPDATECHECK2_ADDRESS, UpdateServerHostAddress, bEnableIPv6 ) )
+    if ( NetworkUtil::ParseNetworkAddressBare ( UPDATECHECK2_ADDRESS, UpdateServerHostAddress, bEnableIPv6 ) )
     {
         pClient->CreateCLServerListReqVerAndOSMes ( UpdateServerHostAddress );
     }
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -324,17 +324,10 @@ void CConnectDlg::RequestServerList()
     // use iCustomDirectoryIndex as an index into the vector.
 
     // Allow IPv4 only for communicating with Directories
-#ifdef CLIENT_NO_SRV_CONNECT
-    if ( NetworkUtil().ParseNetworkAddress (
+    if ( NetworkUtil::ParseNetworkAddress (
              NetworkUtil::GetDirectoryAddress ( pSettings->eDirectoryType, pSettings->vstrDirectoryAddress[pSettings->iCustomDirectoryIndex] ),
              haDirectoryAddress,
              false ) )
-#else
-    if ( NetworkUtil().ParseNetworkAddressWithSrvDiscovery (
-             NetworkUtil::GetDirectoryAddress ( pSettings->eDirectoryType, pSettings->vstrDirectoryAddress[pSettings->iCustomDirectoryIndex] ),
-             haDirectoryAddress,
-             false ) )
-#endif
     {
         // send the request for the server list
         emit ReqServerListQuery ( haDirectoryAddress );
@@ -864,7 +857,7 @@ void CConnectDlg::OnTimerPing()
         // try to parse host address string which is stored as user data
         // in the server list item GUI control element
         // the data to be parsed is just IP:port, so no SRV discovery is needed
-        if ( NetworkUtil().ParseNetworkAddress ( pCurListViewItem->data ( LVC_NAME, Qt::UserRole ).toString(), haServerAddress, bEnableIPv6 ) )
+        if ( NetworkUtil::ParseNetworkAddressBare ( pCurListViewItem->data ( LVC_NAME, Qt::UserRole ).toString(), haServerAddress, bEnableIPv6 ) )
         {
             // if address is valid, send ping message using a new thread
 #if QT_VERSION >= QT_VERSION_CHECK( 6, 0, 0 )
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -318,16 +318,23 @@ void CConnectDlg::RequestServerList()
     }
     cbxDirectory->blockSignals ( false );
 
-    // Get the IP address of the directory server (using the ParseNetworAddress
+    // Get the IP address of the directory server (using the ParseNetworkAddress
     // function) when the connect dialog is opened, this seems to be the correct
     // time to do it. Note that in case of custom directories we
     // use iCustomDirectoryIndex as an index into the vector.
 
     // Allow IPv4 only for communicating with Directories
+#ifdef CLIENT_NO_SRV_CONNECT
     if ( NetworkUtil().ParseNetworkAddress (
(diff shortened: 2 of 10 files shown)


softins (dismissed):
Looks fine to me. Tested on all standard directories, using both `-e` and the Custom setting in the server GUI.

softins:
> Excellent! I look forward to testing your changes.

Updated, please could you pull and test? Thanks!

softins:
Interesting, I'll rebase the branch to `main` on my machine, and try it out.

softins:
> Interesting, I'll rebase the branch to `main` on my machine, and try it out.

Yes, it worked fine.

rdica:
Apologies, my ISP chose to perform major maintenance while I was trying to build and test.
I have a working server with a client on it on a private directory. Monitoring and will update in 24+ hrs. I'm using the default 15 min update.

softins (approved):

pljones:
`main` currently builds green for iOS.  Can we get this rebased, please, so the Auto-Build is clean.

softins:
> `main` currently builds green for iOS. Can we get this rebased, please, so the Auto-Build is clean.

Just done that and force-pushed. Curiously, the auto-build that was triggered this time only includes Android, Linux, Mac and Windows, not iOS. May need to invoke a manual auto-build of all targets. Not sure why my previous force-push built all targets.

src/serverlist.cpp
```diff
@@ -969,13 +969,27 @@ void CServerListManager::SetRegistered ( const bool bIsRegister )
         return;
     }
 
+    // It is very important to unlock the Mutex before doing address resolution,
+    // so that the event loop can run and any other timers that need the mutex
+    // can obtain it. Otherwise there is the possibility of deadlock when doing
+    // the SRV lookup, if another timer fires that needs the same mutex.
+    locker.unlock();
```
ann0see, line 978: You're fully sure that this doesn't introduce any concurrency bugs? @softins

ann0see (approved):
Seems to work. There's of course then the question if we want to have this in the client GUI too. But part of an other PR.

src/serverlist.cpp
```diff
@@ -969,13 +969,27 @@ void CServerListManager::SetRegistered ( const bool bIsRegister )
...
     // get the correct directory address
     // Note that we always have to parse the server address again since if
     // it is an URL of a dynamic IP address, the IP address might have
     // changed in the meanwhile.
     // Allow IPv4 only for communicating with Directories
-    const QString strNetworkAddress      = NetworkUtil::GetDirectoryAddress ( DirectoryType, strDirectoryAddress );
-    const bool    bDirectoryAddressValid = NetworkUtil().ParseNetworkAddress ( strNetworkAddress, DirectoryAddress, false );
+    // Use SRV DNS discovery for directory connections, fallback to A/AAAA if none.
+    const QString strNetworkAddress = NetworkUtil::GetDirectoryAddress ( DirectoryType, strDirectoryAddress );
+#ifndef CLIENT_NO_SRV_CONNECT
```
pljones: No, can't use this define name 😄 .

Please
- Update `Jamulus.pro` to make the name common between client and server name resolution
- Update the ParsNetworkAddress code to do the `#ifndef`... stuff - an "as was" wrapper passing `...., true)` to a new method that does SRV discovery unless that trailing param is false.
- Where it's _required_ to bypass SRV lookups, pass false from the calling code to the new method.

pljones: Or something like that.  Or maybe a wrapper that does the SRV look up if it's compiled in, falling back to the existing code for the final resolution.

`ParseNetworkAddress("hostname")` should _always_ be what's in the main code.

`ParseNetworkAddress` would contain the `#ifdef` and do the SRV lookup if needed.  Then drop through to the final resolution.

That should remove the number of scattered `#ifndef...` blocks.

softins: Actually `CLIENT_NO_SRV_CONNECT` does not appear in `Jamulus.pro`. It is just available for specifying explicitly. I think the only place it's documented for anyone that cares is [this comment in the original PR](https://github.com/jamulussoftware/jamulus/pull/2930#issuecomment-1296347435).

Maybe its replacement should have a commented-out mention in `Jamulus.pro`? Or maybe not?

src/serverlist.cpp
```diff
@@ -805,7 +805,7 @@ bool CServerListManager::Load()
             continue;
         }
 
-        NetworkUtil::ParseNetworkAddress ( slLine[0], haServerHostAddr, bEnableIPv6 );
+        NetworkUtil::ParseNetworkAddressBare ( slLine[0], haServerHostAddr, bEnableIPv6 );
```
pljones, line 810: Needs a comment explaining why

softins: Done

pljones: I think it stores the registered string, not the resolved string, to allow for dynamic DNS.  (Same with the server list returned from the Directory to the Client -- it's unresolved entries.)

src/global.h
pljones, line 112: These need port number -- we don't want SRV resolution on them in a remote system, we want to go directly via DNS.

src/clientdlg.cpp
```diff
@@ -588,12 +588,12 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // Send the request to two servers for redundancy if either or both of them
     // has a higher release version number, the reply will trigger the notification.
 
-    if ( NetworkUtil().ParseNetworkAddress ( UPDATECHECK1_ADDRESS, UpdateServerHostAddress, bEnableIPv6 ) )
+    if ( NetworkUtil::ParseNetworkAddress ( UPDATECHECK1_ADDRESS, UpdateServerHostAddress, bEnableIPv6 ) )
```
pljones: See comment in `global.h` -- I think these should go through `...Bare` (for safety).

src/connectdlg.cpp
```diff
@@ -856,7 +856,8 @@ void CConnectDlg::OnTimerPing()
 
         // try to parse host address string which is stored as user data
         // in the server list item GUI control element
```
pljones, line 858: As per `serverlist.cpp` comment -- I think these are _meant_ to be plain text, not resolved addresses, so that dynamic DNS can be used at all times, which would include dynamic SRV in my opinion (i.e. a temporary server moving port and IP but keeping the same "simple" registered server name).

src/serverdlg.cpp
```diff
@@ -482,12 +482,12 @@ CServerDlg::CServerDlg ( CServer* pNServP, CServerSettings* pNSetP, const bool b
     // Send the request to two servers for redundancy if either or both of them
     // has a higher release version number, the reply will trigger the notification.
 
-    if ( NetworkUtil().ParseNetworkAddress ( UPDATECHECK1_ADDRESS, UpdateServerHostAddress, pServer->IsIPv6Enabled() ) )
+    if ( NetworkUtil::ParseNetworkAddress ( UPDATECHECK1_ADDRESS, UpdateServerHostAddress, pServer->IsIPv6Enabled() ) )
```
pljones: As client.

src/clientdlg.cpp
softins: I did wonder that myself. I'll update.

src/global.h
softins: If we use `...Bare` as mentioned below, they don't need to add the port number unless it's non-default.

src/connectdlg.cpp
softins: I don't think so. A server never sends its DNS name when registering to a directory. The directory just stores the IP and port, and those are what it sends to a client requesting a list. And the alternate IP address, although still a string, is still just an IP, isn't it? During the debugging phase, I printed out what was being passed here, and it was always `IP:port`.

src/serverlist.cpp
softins: See [my comment below](https://github.com/jamulussoftware/jamulus/pull/3556#discussion_r2724122935)

src/connectdlg.cpp
pljones: I'm probably remembering some discussion with Volker about the original work done for this... Thinking back to the work I did on the server list, you're right - it's all IP/port -- returning external or and internal registered value, depending on what the incoming request matches.

OK, this is fine, then.

src/serverlist.cpp
pljones: Any compile-time option should really be available through Jamulus.pro via `CONFIG += something` to add it to the defines.

src/global.h
pljones: OK.

src/clientdlg.cpp
```diff
@@ -588,12 +588,14 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     // Send the request to two servers for redundancy if either or both of them
     // has a higher release version number, the reply will trigger the notification.
 
-    if ( NetworkUtil().ParseNetworkAddress ( UPDATECHECK1_ADDRESS, UpdateServerHostAddress, bEnableIPv6 ) )
+    // Don't use SRV resolution when resolving update servers.
```
pljones, line 591: OK, _completely_ unrelated question...

- Why do we do this in the UI?  It should be in main.cpp, really, making a call to a new `NetworkUtil::CheckForUpdate()` and either logging a message or popping up a "Do you want to download it now?" and starting the browser.  But that's definitely for 4.0, not 3.12.

(Still, it could be moved into a common method that just returns boolean back to the caller.)

pljones: Opened https://github.com/jamulussoftware/jamulus/issues/3604 to do this.

pljones:
Related issue: https://github.com/orgs/jamulussoftware/discussions/1772

Issue #3542, opened 2025-09-28
Connect Dialog with "--showallservers" does not sort "Version" normally


pljones:
**Describe the bug**

When the Client is run with `--showallservers`, the Connect Dialog shows all Servers supplied in the Directory server list response, regardless of whether they respond to a "ping" packet.  This list now contains a field displayed as "Version".

In this mode, the Client allows sorting the server list by the columns in the dialog, including the Version.

**To Reproduce**

1. Run the Client with `--showallservers` in the command line.
2. Open the Connect dialog
3. Select a Directory, e.g. Any Genre 1
4. Make sure the dialog is wide enough to expose the Version column and that the Version column is wide enough to show the versions fully
5. Click the Version column heading to sort by version (ascending).  This puts the empty values first (expected).
6. Click the Version column heading a second time to sort by version descending.

**Expected behavior**

The versions should be in "version sort" order rather than character set sort order.

**Version of Jamulus**

Jamulus 3.12.0beta2 (or anything after the Version column was introduced)

ann0see:
If possible, it should still be fixed for the RC...

softins:
I've just had an initial look at this. It looks to be quite involved to do. The connect dialog list is a `QTreeWidget`, which is a simplified subclass of `QTreeView` using a predefined model. This model appears just to have built-in column sorting of the displayed text as ASCII up or down.

To fix the version sorting looks like it would require reimplementing the widget as a `QTreeView` and providing it with a custom model subclassed from `QAbstractItemModel` and hooking in our own sorting methods via that model.

That's something I don't have time to do at the moment, nor the detailed knowledge (I sadly haven't managed to get back into Jamulus coding since moving house).

pljones:
https://codebrowser.dev/qt5/qtbase/src/widgets/itemviews/qtreewidget.h.html#_ZN15QTreeWidgetItemaSERKS_
The `QTreeWidgetItem` class overrides operator= already.  The comparison for sort is done with instances of `QTreeWidgetItem` using their `<` and `>` operators.  So for the version, could we subclass `QTreeWidgetItem`?

softins:
I've just had a long and productive conversation with ChatGPT, which has explained how to do it. I'll share the chat soon.

softins:
This is the chat: https://chatgpt.com/share/68e05067-a540-8006-9028-ec026a48f029

It shows the evolution of the solution, and it looks like the end point will be the most useful. I haven't had time to try it out yet.

[...]

PR #3557, opened 2025-11-06
Add version sorting using CMappedTreeWidgetItem

softins:

Corrects the sorting of the version column in the connect dialog when using --showallservers.
It does this by mapping the version number to a sortable text field and storing that in the `data` field of the widget item.
It also subclasses `QTreeWidgetItem` as `CMappedTreeWidgetItem` in order to provide an override for `operator<`
to sort on the `data` field instead of the `text` field.

Also increases the default column width for the version column.

CHANGELOG: Client: Fix sorting of version number in connect dialog with --showallservers

**Context: Fixes an issue?**

Fixes: #3542

No, it's just a bugfix.

Tested on RPi Linux. Ready for testing on other platforms.

Review.

## Checklist

### DIFF
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -25,6 +25,75 @@
 #include "connectdlg.h"
 
 /* Implementation *************************************************************/
+
+// mapVersionStr - converts a version number to a sortable string
+static QString mapVersionStr ( const QString& versionStr )
+{
+    QString key;
+    QString x = ">"; // default suffix is later (git, dev, nightly, etc)
+
+    // Regex for SemVer: major.minor.patch-suffix
+    QRegularExpression      semVerRegex ( R"(^(\d+)\.(\d+)\.(\d+)-?(.*)$)" );
+    QRegularExpressionMatch match = semVerRegex.match ( versionStr );
+
+    if ( !match.hasMatch() )
+    {
+        return versionStr; // fallback: plain text
+    }
+
+    int     major  = match.captured ( 1 ).toInt();
+    int     minor  = match.captured ( 2 ).toInt();
+    int     patch  = match.captured ( 3 ).toInt();
+    QString suffix = match.captured ( 4 ); // may be empty
+
+    if ( suffix.isEmpty() )
+    {
+        x = "="; // bare version number
+    }
+    else if ( suffix.startsWith ( "rc" ) || suffix.startsWith ( "beta" ) || suffix.startsWith ( "alpha" ) )
+    {
+        x = "<"; // pre-release version
+    }
+
+    // construct a sortable key mmmnnnpppksuffix, where:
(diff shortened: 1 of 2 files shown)


src/connectdlg.cpp
```diff
@@ -25,6 +25,75 @@
...
+    if ( column != CConnectDlg::LVC_VERSION )
+        return QTreeWidgetItem::operator<( other );
+
+    QVariant lhs = data ( column, Qt::UserRole );
+    QVariant rhs = other.data ( column, Qt::UserRole );
+
+    if ( !lhs.isValid() || !rhs.isValid() )
+        return QTreeWidgetItem::operator<( other );
+
+    return lhs.toString() < rhs.toString();
```
pljones, line 94: Are these meant to be wrapped in `mapVersionStr (...)`?

softins: No, they go through `mapVersionStr()` on their way to `setData()`, so that when the overloaded `operator<` fetches them using `data()`, they are ready to compare. They just need converting from a `QVariant` explicitly to ` QString`, since `QVariant` is no longer orderable in Qt6.

pljones: Ah right!

softins:
> The column could still do with being wider. It's 18 characters maximum. I make that nearly as big as the server name.

It's hard to be exact, as it varies between platforms. The main part of the version, and the alpha/beta/rc will be visible. Someone who is particularly interested in the dev commit numbers can always widen the field manually. It's no big deal to me either way.

src/connectdlg.cpp
```diff
@@ -25,6 +25,75 @@
...
+    int     major  = match.captured ( 1 ).toInt();
+    int     minor  = match.captured ( 2 ).toInt();
+    int     patch  = match.captured ( 3 ).toInt();
+    QString suffix = match.captured ( 4 ); // may be empty
+
+    if ( suffix.isEmpty() )
+    {
+        x = "="; // bare version number
+    }
+    else if ( suffix.startsWith ( "rc" ) || suffix.startsWith ( "beta" ) || suffix.startsWith ( "alpha" ) )
```
ann0see, line 53: - [ ] Document somewhere on the release docs that the application actually filters based on these suffixes

ann0see (approved):

src/connectdlg.cpp
softins: It's not so much a filter. All suffixes are shown. It's just that alpha, beta and rc sort before the base version, and all other suffixes sort after. I agree it's worth documenting somewhere suitable.

pljones: As a reminder in one of `COMPILING.md` or `CONTRIBUTING.md` -- or a comment at the top of `Jamulus.pro` perhaps.

PR #3558, opened 2025-11-06
Convert from XML::Simple to XML::LibXML

softins:

This converts the Perl script `tools/checkkeys.pl` from using the deprecated `XML::Simple` to using `XML::LibXML` instead. There are no changes in functionality or behaviour.

CHANGELOG: Tools: update checkkeys.pl to use XML::LibXML

**Context: Fixes an issue?**

`XML::Simple` is [now deprecated](https://metacpan.org/pod/XML::Simple#STATUS-OF-THIS-MODULE) as reported by @pljones on Discord. This is the only tool in the repo that used `XML::Simple`.

No. Internal operation only.

Ready to merge

Nothing

## Checklist

### DIFF
--- a/tools/checkkeys.pl
+++ b/tools/checkkeys.pl
@@ -25,44 +25,48 @@
 ##############################################################################
 
 use open qw(:std :utf8);
-use XML::Simple qw(:strict);
+use XML::LibXML;
 use Data::Dumper;
 
 my %keys;
 
 my $doneeng = 0;
+my $language;
 
 while ($ts = <*.ts>) {
-	my $xs = XMLin($ts, KeyAttr => {}, ForceArray => ['context', 'message']);
+	my $dom = XML::LibXML->load_xml(location => $ts);
 
-	# print Data::Dumper->Dump([$xs], [qw(xs)]);
+	foreach my $TS ($dom->findnodes('//TS/@language')) {
+		$language = $TS->to_literal();
+		printf "Language: %s\n", $language;
+	}
 
-	printf "Language: %s\n", $xs->{language};
-
-	foreach $context (@{$xs->{context}}) {
-		# printf "\n========================================\nContext: %s\n", $context->{name};
-		my $contextname = $context->{name};
-		$contextname =~ s/Base$//;	# merge base class with its child
+	foreach my $context ($dom->findnodes('//context')) {
+		my $contextname = $context->findvalue('./name');
+		# printf "\n========================================\nContext: %s\n", $contextname;
+		$contextname =~ s/Base$//;      # merge base class with its child
 		$contextname = 'CClientDlg+CHelpMenu' if ($contextname eq 'CClientDlg' || $contextname eq 'CHelpMenu');
 
-		foreach $message (@{$context->{message}}) {
-			# printf "  Msg: %s\n", $message->{source};
-			next if $message->{translation}{type} eq 'obsolete';
-			next if $message->{translation}{type} eq 'vanished';
-			#next if $message->{translation}{type} eq 'unfinished';	# don't skip unfinished strings, as they may still get used
+		MESSAGE: foreach my $message ($context->findnodes('./message')) {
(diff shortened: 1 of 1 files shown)


tools/checkkeys.pl
```diff
@@ -25,44 +25,48 @@
...
 
-			next unless $message->{source} =~ /\&(.)/;
+			foreach my $translation ($message->findnodes('./translation')) {
+				my $type = $translation->{type};
+				next MESSAGE if $type eq 'obsolete';
+				next MESSAGE if $type eq 'vanished';
+				#next MESSAGE if $type eq 'unfinished';	# don't skip unfinished strings, as they may still get used
 
-			push @{$keys{en}{$contextname}{uc $1}}, $message->{source} unless $doneeng;
+				# skip messages without an accelerator key
```
pljones, line 60: OK, we're still skipping even if the translation has an accelerator but the English doesn't.

Shouldn't we
1. Check that both English and translation have accelerators or both don't (for documentation consistency) -- or is this done elsewhere?
2. Then skip if both don't

If the "both the same" is done elsewhere, this is okay -- but I'd suggest mentioning that in the comment.

softins: > OK, we're still skipping even if the translation has an accelerator but the English doesn't.

I'll check, but I think that's the same with the existing tool.

> Shouldn't we
> 
> 1. Check that both English and translation have accelerators or both don't (for documentation consistency) -- or is this done elsewhere?

> 2. Then skip if both don't

That is probably a good idea, but out of scope for this PR, which was only concerned with changing XML library.

> If the "both the same" is done elsewhere, this is okay -- but I'd suggest mentioning that in the comment.

There are other improvements that could be done to this tool too, but I would do them, and the good suggestions above, separately. At the moment, the output of the tool is just a starting point to aid manual review as part of the release process.

pljones:
Ran on HEAD of softins/jamulus:perl-xml-libxml
```
$ ./tools/checkkeys.pl

Possible duplicate hotkeys:
```
I'm pretty certain that's right!

----

Not for now but I'll feature request a summary saying "Checked <X> files, found <Y> languages with <Z> hotkeys" then only emit the "Possible duplicate hotkeys:" if there are any.  (Of course, it's so long since I wrote any Perl, I'm tempted to use AI to translate it to something else -- bash with `yq` (which'll do XML) or even python -- and change that...)

softins:
> Ran on HEAD of softins/jamulus:perl-xml-libxml
> 
> ```
> $ ./tools/checkkeys.pl
> 
> Possible duplicate hotkeys:
> ```
> 
> I'm pretty certain that's right!

You need to cd to the translation directory first:

```
cd src/translations
../../tools/checkkeys.pl
```

The new version of the tool produces identical output to the old one. The output is simplistic in that it isn't able to take full account of context to know which accelerators actually conflict, and which matches are actually ok because they are in different menus, for example. But overcoming that is a bigger project than just changing XML library.

> Not for now but I'll feature request a summary saying "Checked files, found languages with hotkeys" then only emit the "Possible duplicate hotkeys:" if there are any. (Of course, it's so long since I wrote any Perl, I'm tempted to use AI to translate it to something else -- bash with `yq` (which'll do XML) or even python -- and change that...)

If rewriting it, it would need to be into Python with nested dicts. But I'm only a beginner with Python, as opposed to being very at home with Perl for 30 years!

pljones:
> You need to cd to the translation directory first:

Things like that could also be changed.  Ideally all these checks should be run from the repository root and take care of their own internal requirements.  Again, not for this PR.

tools/checkkeys.pl
pljones: Well, it _does_ change the flow, which raises the questions over what the old was doing and what the new is doing.

ann0see (approved):

pljones (approved):
Approving on the basis of the above discussion.

PR #3559, opened 2025-11-09
Add `--accessible` button to fix voice over bug on server list

ann0see:
Applies the fix by @chigkim to upstream by introducing a CLI argument.

CHANGELOG: Accessability: Add `--accessible` CLI argument to enable alternative, VoiceOver compatible server list
**Context: Fixes an issue?**

Maybe: #490 

Yes. This should have an entry somewhere in the docs.

Ready for testing on macOS. (I did not do it yet).

Testing

## Checklist

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -33,6 +33,7 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
                          const bool       bShowAnalyzerConsole,
                          const bool       bMuteStream,
                          const bool       bNEnableIPv6,
+                         const bool       bNEnableAccessiblePushButtonUi,
                          QWidget*         parent ) :
     CBaseDlg ( parent, Qt::Window ), // use Qt::Window to get min/max window buttons
     pClient ( pNCliP ),
@@ -40,11 +41,12 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
     bConnectDlgWasShown ( false ),
     bDetectFeedback ( false ),
     bEnableIPv6 ( bNEnableIPv6 ),
+    bEnableAccessiblePushButtonUi ( bNEnableAccessiblePushButtonUi ),
     eLastRecorderState ( RS_UNDEFINED ), // for SetMixerBoardDeco
     eLastDesign ( GD_ORIGINAL ),         //          "
     ClientSettingsDlg ( pNCliP, pNSetP, parent ),
     ChatDlg ( parent ),
-    ConnectDlg ( pNSetP, bNewShowComplRegConnList, parent ),
+    ConnectDlg ( pNSetP, bNewShowComplRegConnList, bNEnableIPv6, bNEnableAccessiblePushButtonUi, parent ),
     AnalyzerConsole ( pNCliP, parent )
 {
     setupUi ( this );
@@ -1286,11 +1288,11 @@ void CClientDlg::Disconnect()
     TimerDetectFeedback.stop();
     bDetectFeedback = false;
 
-    //### TODO: BEGIN ###//
-    // is this still required???
-    // immediately update status bar
+    // ### TODO: BEGIN ###//
+    //  is this still required???
+    //  immediately update status bar
     OnTimerStatus();
-    //### TODO: END ###//
+    // ### TODO: END ###//
 
     // reset LEDs
     ledBuffers->Reset();
--- a/src/clientdlg.h
+++ b/src/clientdlg.h
@@ -85,6 +85,7 @@ public:
                  const bool       bShowAnalyzerConsole,
                  const bool       bMuteStream,
                  const bool       bNEnableIPv6,
+                 const bool       bNEnableAccessiblePushButtonUi,
                  QWidget*         parent = nullptr );
 
 protected:
@@ -110,6 +111,7 @@ protected:
     bool           bConnectDlgWasShown;
     bool           bDetectFeedback;
     bool           bEnableIPv6;
+    bool           bEnableAccessiblePushButtonUi;
     ERecorderState eLastRecorderState;
     EGUIDesign     eLastDesign;
     QTimer         TimerSigMet;
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -25,7 +25,11 @@
 #include "connectdlg.h"
 
 /* Implementation *************************************************************/
-CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteRegList, const bool bNEnableIPv6, QWidget* parent ) :
+CConnectDlg::CConnectDlg ( CClientSettings* pNSetP,
+                           const bool       bNewShowCompleteRegList,
+                           const bool       bNEnableIPv6,
+                           const bool       bNEnableAccessiblePushButtonUi,
+                           QWidget*         parent ) :
     CBaseDlg ( parent, Qt::Dialog ),
     pSettings ( pNSetP ),
     strSelectedAddress ( "" ),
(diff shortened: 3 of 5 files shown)


src/clientdlg.cpp
```diff
@@ -1286,11 +1288,11 @@ void CClientDlg::Disconnect()
     TimerDetectFeedback.stop();
     bDetectFeedback = false;
 
-    //### TODO: BEGIN ###//
-    // is this still required???
-    // immediately update status bar
+    // ### TODO: BEGIN ###//
```
ann0see, line 1291: Clang format did this automatically...

src/clientdlg.cpp
```diff
@@ -33,18 +33,20 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
...
     bConnectDlgWasShown ( false ),
     bDetectFeedback ( false ),
     bEnableIPv6 ( bNEnableIPv6 ),
+    bEnableAccessiblePushButtonUi ( bNEnableAccessiblePushButtonUi ),
     eLastRecorderState ( RS_UNDEFINED ), // for SetMixerBoardDeco
     eLastDesign ( GD_ORIGINAL ),         //          "
     ClientSettingsDlg ( pNCliP, pNSetP, parent ),
     ChatDlg ( parent ),
-    ConnectDlg ( pNSetP, bNewShowComplRegConnList, parent ),
+    ConnectDlg ( pNSetP, bNewShowComplRegConnList, bNEnableIPv6, bNEnableAccessiblePushButtonUi, parent ),
```
ann0see, line 49: the ipv6 toggle wasn't passed correctly.

pljones:
Can we also have a Settings->My Profile "User Interface" checkbox, like there is for "Audio Alerts", please.  (It would need to send a signal wire to a slot in the main client, which would issue a signal wired to a slot in the Connect dialog, which would switch over the widget.)

I'd also take care with a _generic_ term like `--accessible` if it only changes _one specific_ widget in the UI.  (I mean, the Audio Alerts are explicitly an accessibility feature and it's not like you need to use `--accessible` to get _that_.)

pljones:
> > I'd also take care with a generic term like --accessible
> 
> Agree. My thought was to leave this generic to allow future accessibility features which don't "just work" out of the box.

Unless it's a compile-time option, the code is there to be used.  We _shouldn't_ introduce new code that **needs** a run time option to enable it.  It should be configurable through the UI, through RPC (if relevant -- this isn't, I suppose) or through the command line equally.  When configured through the UI or RPC, the setting should be persisted.

This is why I don't think this feature is close to ready for release.

src/clientdlg.cpp
softins: Wow, so that bug has been hiding for a long time! The `parent` argument was being converted to a `bool`, and therefore was always `true`? Although I can't think what negative effect that would have had in practice.

pljones: It's another indicator of the overall design problem.

Passing parameters is just the wrong way to do it.  This should be signals on CClient with slots wired inside CConnectDlg to get notifications of updates to the values, and getters on CClient to supply the current state.  Only CClient should provide this information.  Even things like the IPv6 toggle _should_ be controllable through the UI -- so you'd need the CConnectDlg to track changes in that state.  CClientDlg would then get passed a thread-safe reference to CClient.

(All for r4.0.0, naturally.)

softins (commented):
I have tried this out. The `QPushButton` widgets don't look good in the server list.

I read the links mentioned by @ann0see [here](https://github.com/jamulussoftware/jamulus/issues/490#issuecomment-2564841869) and following a suggestion in the second one, I changed the item widgets `QPushButton` to `QLabel`. It was necessary to create the `QLabel` with empty text, otherwise it showed up as well as the original text, and then instead of `setText()` on the widget, I used `setAccessibleName()`. It now looks fine, but how do I test it? I have a Mac (still running 10.15 Catalina), but don't know anything about screen readers.

pljones:
OK, I need to get some familiarity with the terminology...  The connect dialog contains a QTreeWidget nominally titled the "Server List".  This contains a header and then one or more server rows.  Each server row may have client rows:

```mermaid
erDiagram
    serverList ||--|| headerRow : has
    headerRow ||--|{ headerColumns : contains
    serverList ||--o{ serverRow : has
    serverRow ||--|{ serverDetailColumns : contains
    serverRow ||--o{ clientRow : has
    clientRow ||--|{ clientDetailColumns : contains
```

The QTreeWidget is the outermost thing, right -- the serverList?  And is everything else a QTreeWidgetItem?  And we'd need to implement at least the basic "read the content of the existing QTreeWidgetItem text"?

So the bit from ChatGPT "Here’s a small enhancement for the Qt 5.15 build that makes VoiceOver read both the parent and the currently selected child together" -- this would start by saying "Server List" (if that was in the accessibleText), then ... what?  Say "3ms" if you were on the ping time of a nearby server?  I don't know how the accessible navigation works for the widget and what context makes sense.  It feels like having the header column accessibleText along with the server column accessibleText would probably work better.

(ChatGPT was able to parse the mermaid diagram, by the way.)

softins:
I've not tried it yet in the Jamulus context. And I only have macOS 10.15 with Qt 5.15.2.

Apparently Qt 6.5+ should just work out of the box, even on Mac, but I don't have anything to try that on. Maybe that's why someone (you?) said it worked fine on Windows, because our Windows build uses a recent Qt6?

The original accessibility problem reports date from 2020 when Volker was still maintainer and we were on earlier Qt5.

pljones:
And the point with using the accessibleText is that it would do the same thing in Qt 6 as Qt 5, then -- if we wanted custom accessibleText, we could set _that_ and have Qt 5 pick it up, I presume, using the fix suggested for Qt 5?  Or is accessibleText not even there on the QTreeWidgetItem in Qt 5?

softins:
Wading through treacle a bit on this, with Qt 5.15.

According to ChatGPT, accessible navigation of a `QTreeWidget` should just work out of the box for Qt 6.5 onwards.

We should note that @chigkim's original report #490 and suggested fix date from 2020, when we were using Qt5.

Our standard Mac build now uses Qt 6.9, and only the legacy build uses 5.15.

Please could someone verify whether our standard Mac build (`main`, not this PR) with Qt6 actually does or doesn't provide VoiceOver accessibility for the Connect dialog already? If it does, then this PR is not actually needed, and we would be justified in just saying the legacy Mac build does not provide accessibility.

I can't verify it myself, as I only have an old Mac with macOS 10.15, so can only run the legacy version.

chigkim:
I just built the master branch with qt 6.9.3 and tried on MacOS, and the server list is still not accessible.

ann0see:
@chigkim did you start Jamulus with the --accessible flag?

ann0see:
Does the code of this PR contain all necessary changes from your build?

pljones:
OK, so the older Qt 5 patch approach works on _current_ macOS?

We have two variables and, to keep things clear for me, it helps to make sure that we're not changing both at once.

- [ ] Qt 5x plus patch -> works on current macOS -- is this correct?
- [ ] Qt 6x without patch -> expected to work on current macOS with native Qt widget but doesn't (although "expected" here isn't clear) -- is this correct?

And Tony's patch also did
- [ ] Qt 5x with a different patch -> didn't work on current macOS but doesn't -- is this correct?
- [ ] Qt 6x with a different patch -> didn't work on current macOS but doesn't -- is this correct?

(This is given that the Qt 5x build is generally considered "legacy only", so I'm trying to compare like with like.)

ann0see:
Actually, I did somehow manage to get into the server list without having enabled the accessible flag on Qt6 and VoiceOver spoke out the content. I'll try to reproduce this as this would show that the bug is actually fixed, but rather somthing weird needs to happen to start VoiceOver to speak out the contents.

Maybe a tabstop is missing in the table?

chigkim:
When I try #3573, it says unknown option --accessible. Also it doesn't work.

PR #3562, opened 2025-11-23
Where available, use the commit time rather than commit id, for sort

pljones:

As discussed elsewhere, I'd like the `--showallservers` sort to use the _commit time_ rather than _commit id_, where it's available.  This means changing the `APP_VERSION` to add the git timestamp to the end and then splitting that off for display purposes.  The define `DISPLAY` thus becomes a call to a method that does the trim.

CHANGELOG: Client: Use the commit time rather than commit id for --showallservers version sort

**Context: Fixes an issue?**

No.

When the docs for the sort on Version are written, they'd need to take this into account.

Tested locally.

Nothing _broke_ -- existing servers sort exactly the same way as before as they have no timestamp.

## Checklist

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -15,7 +15,7 @@ contains(CONFIG, "noupcasename") {
 # allow detailed version info for intermediate builds (#475)
 contains(VERSION, .*dev.*) {
     exists(".git/config") {
-        GIT_DESCRIPTION=$$system(git describe --match=xxxxxxxxxxxxxxxxxxxx --always --abbrev --dirty) # the match should never match
+        GIT_DESCRIPTION=$$system(git describe --match=xxxxxxxxxxxxxxxxxxxx --always --abbrev --dirty):$$system(git show -s "--pretty=format:%ct") # commit_id(-dirty):seconds_since_epoch
         VERSION = "$$VERSION"-$$GIT_DESCRIPTION
         message("building version \"$$VERSION\" (intermediate in git repository)")
     } else {
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -33,7 +33,7 @@ static QString mapVersionStr ( const QString& versionStr )
     QString x = ">"; // default suffix is later (git, dev, nightly, etc)
 
     // Regex for SemVer: major.minor.patch-suffix
-    QRegularExpression      semVerRegex ( R"(^(\d+)\.(\d+)\.(\d+)-?(.*)$)" );
+    QRegularExpression      semVerRegex ( R"(^(\d+)\.(\d+)\.(\d+)-?(.*):?(.*)$)" );
     QRegularExpressionMatch match = semVerRegex.match ( versionStr );
 
     if ( !match.hasMatch() )
@@ -45,6 +45,7 @@ static QString mapVersionStr ( const QString& versionStr )
     int     minor  = match.captured ( 2 ).toInt();
     int     patch  = match.captured ( 3 ).toInt();
     QString suffix = match.captured ( 4 ); // may be empty
+    QString tstamp = match.captured ( 5 ); // may be empty
 
     if ( suffix.isEmpty() )
     {
@@ -66,7 +67,7 @@ static QString mapVersionStr ( const QString& versionStr )
               .arg ( minor, 3, 10, QLatin1Char ( '0' ) )
               .arg ( patch, 3, 10, QLatin1Char ( '0' ) )
               .arg ( x )
-              .arg ( suffix );
+              .arg ( tstamp.isEmpty() ? suffix : tstamp );
 
     return key;
 }
@@ -184,13 +185,13 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     lvwServers->setColumnWidth ( LVC_NAME, 200 );
     lvwServers->setColumnWidth ( LVC_PING, 130 );
     lvwServers->setColumnWidth ( LVC_CLIENTS, 100 );
-    lvwServers->setColumnWidth ( LVC_VERSION, 110 );
+    lvwServers->setColumnWidth ( LVC_VERSION, 150 );
 #else
     lvwServers->setColumnWidth ( LVC_NAME, 180 );
     lvwServers->setColumnWidth ( LVC_PING, 75 );
     lvwServers->setColumnWidth ( LVC_CLIENTS, 70 );
     lvwServers->setColumnWidth ( LVC_LOCATION, 220 );
-    lvwServers->setColumnWidth ( LVC_VERSION, 95 );
+    lvwServers->setColumnWidth ( LVC_VERSION, 135 );
 #endif
     lvwServers->clear();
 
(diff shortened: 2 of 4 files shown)


src/protocol.cpp
pljones, line 1: `APP_VERSION` passed around in the protocol messages, to retain the full string.

src/util.h
pljones: `GetDisplayVersion` strips the seconds since epoch, where present.

src/connectdlg.cpp
```diff
@@ -33,7 +33,7 @@ static QString mapVersionStr ( const QString& versionStr )
     QString x = ">"; // default suffix is later (git, dev, nightly, etc)
 
     // Regex for SemVer: major.minor.patch-suffix
-    QRegularExpression      semVerRegex ( R"(^(\d+)\.(\d+)\.(\d+)-?(.*)$)" );
+    QRegularExpression      semVerRegex ( R"(^(\d+)\.(\d+)\.(\d+)-?(.*):?(.*)$)" );
```
pljones, line 36: Capture seconds since epoch if present.

src/connectdlg.cpp
```diff
@@ -66,7 +67,7 @@ static QString mapVersionStr ( const QString& versionStr )
               .arg ( minor, 3, 10, QLatin1Char ( '0' ) )
               .arg ( patch, 3, 10, QLatin1Char ( '0' ) )
               .arg ( x )
-              .arg ( suffix );
+              .arg ( tstamp.isEmpty() ? suffix : tstamp );
```
pljones, line 70: Use seconds since epoch, rather than commit id, if present.

src/connectdlg.cpp
```diff
@@ -1024,7 +1025,7 @@ void CConnectDlg::SetServerVersionResult ( const CHostAddress& InetAddr, const Q
 
     if ( pCurListViewItem )
     {
-        pCurListViewItem->setText ( LVC_VERSION, strVersion );
+        pCurListViewItem->setText ( LVC_VERSION, GetDisplayVersion ( strVersion ) );
```
pljones, line 1028: As this method gets passed the protocol version (`APP_VERSION`), it needs to call `GetDisplayVersion` to retain existing display behaviour.

(Also note this only gets called once each time the server is seen, not on every refresh.)

Jamulus.pro
pljones, line 1: Set the capture the commit id and commit timestamp in seconds since epoch, rather than only the commit id.

Note that it loses the `-dirty` suffix, but I've never really been fussed about that.  Most people making unofficial builds will either be _using_ git (and checking in -- or tough, they'll get their latest commit id) or using the tar ball and not getting a commit id or timestamp.

pljones: Hm, maybe it does need the `-dirty` -- if you're comparing two servers, you want to know the commit id means "this is the same source".  If someone's modified and built locally without committing, you don't want to think it's the same version as was committed.

Of course, they can modify Jamulus.pro to fake it... which I'd expect any actual "malicious actor" to do.

_And_ it only does this at all on dev versions with a git repo.

src/util.h
```diff
@@ -1373,3 +1373,5 @@ struct EnumClassHash
     }
 };
 #endif
+
+inline QString GetDisplayVersion ( QString str ) { return str.contains ( ':' ) ? str.mid ( 0, str.lastIndexOf ( ':' ) ) : str; }
```
pljones: `str.mid ( 0, str.lastIndexOf ( ':' ) )` because it has to work on Qt5.

src/global.h
```diff
@@ -374,3 +374,5 @@ bool GetNumericArgument ( int     argc,
                           double  rRangeStart,
                           double  rRangeStop,
                           double& rValue );
+
+inline QString GetDisplayVersion ( QString str ) { return str.contains ( ':' ) ? str.mid ( 0, str.lastIndexOf ( ':' ) ) : str; }
```
pljones, line 378: `GetDisplayVersion` strips the seconds since epoch, where present.

`str.mid ( 0, str.lastIndexOf ( ':' ) )` for Qt 5 compatibility.

src/global.h
```diff
@@ -73,7 +73,7 @@ LED bar:      lbr
 
 // version and application name (use version from qt prject file)
 #undef VERSION
-#define VERSION  APP_VERSION
+#define VERSION  GetDisplayVersion ( APP_VERSION )
```
pljones, line 76: `GetDisplayVersion ( APP_VERSION )` used where `APP_VERSION` is currently referenced by this define, in existing code.  `APP_VERSION` wasn't referenced anywhere directly except here before.

pljones:
Example from build that displays version but hasn't had this update, where a server is running that _has_ got this update:
image

Example from a build that displays version and has this update, for the same server (compared with the above):
image

softins:
Ack the review request, but I'm away this week, so won't be able to examine until next week.

Jamulus.pro
```diff
@@ -15,7 +15,7 @@ contains(CONFIG, "noupcasename") {
 # allow detailed version info for intermediate builds (#475)
 contains(VERSION, .*dev.*) {
     exists(".git/config") {
-        GIT_DESCRIPTION=$$system(git describe --match=xxxxxxxxxxxxxxxxxxxx --always --abbrev --dirty) # the match should never match
+        GIT_DESCRIPTION=$$system(git show -s --pretty=format:"%h:%ct") # commit:seconds-since-epoch
```
softins:
```suggestion
        # the match should never match, and then append the commit time
        GIT_DESCRIPTION=$$system(git describe --match=xxxxxxxxxxxxxxxxxxxx --always --abbrev --dirty):$$system(git show -s --pretty=format:"%ct")
```

This will retain `-dirty` when appropriate. Then I get:
```
$ qmake Jamulus.pro
Project MESSAGE: building version "3.11.0dev-6c890814-dirty:1762730446" (intermediate in git repository)
...
```

...and after committing:

```
$ qmake Jamulus.pro
Project MESSAGE: building version "3.11.0dev-824768b8:1764974439" (intermediate in git repository)
```

Jamulus.pro
```diff
@@ -15,7 +15,7 @@ contains(CONFIG, "noupcasename") {
 # allow detailed version info for intermediate builds (#475)
 contains(VERSION, .*dev.*) {
     exists(".git/config") {
-        GIT_DESCRIPTION=$$system(git describe --match=xxxxxxxxxxxxxxxxxxxx --always --abbrev --dirty) # the match should never match
+        GIT_DESCRIPTION=$$system(git describe --match=xxxxxxxxxxxxxxxxxxxx --always --abbrev --dirty):$$system(git show -s "--pretty=format:%ct") # commit_id(-dirty):seconds_since_epoch
```
pljones, line 18: Append seconds since epoch of *HEAD* commit id (so `-dirty` should _really_ mean "some time after this").

pljones:
```
~/Library/Cache/jamulus-dependencies/create-dmg/create-dmg/v1.2.2 ~/work/jamulus/jamulus
install -d /usr/local
install -d /usr/local/bin
install create-dmg /usr/local/bin/create-dmg
install -d /usr/local/share/create-dmg
cp -R support /usr/local/share/create-dmg
cp -R examples /usr/local/share/create-dmg
cp -R tests /usr/local/share/create-dmg
~/work/jamulus/jamulus
Creating disk image...
hdiutil: create failed - Resource busy
Error: Process completed with exit code 1.
```
That's a new one.  I'll rerun it...

src/connectdlg.cpp
ann0see: Why not match digits only? I believe this is a bit too permissive?

pljones: We don't care whether it's digits, just that there's a separator.

PR #3571, opened 2025-11-30
Add fader level set method to RPC remote interface

corrados:

Dear Jamulus development team :-). After a very long time, I worked on Jamulus again. I needed to add a fader level set method to the RPC remote inteface because I want to run Jamulus in headless mode together with my Edrumulus e-drum software on a Raspberry Pi. Edrumulus is remote controlled with a smart phone via a web interface and for Jamulus I want to write a small Android app which uses the RPC interface.

Note: For being able to make changes, I had to create a fork. I remember that this was problematic since I moved the jamulus project from my private space "corrados" to "jamulussoftware" and Github automatically forwarded git actions if a developer did use the old path. Since jamulus is now about 5 years in "jamulussoftware", I hope that in the meantime all developer have updated their Git repo and have cloned from the correct origin URL. If this is not the case, please let me know. Then I will delete my fork again.

CHANGELOG: Added a jamulusclient/setFaderLevel method to the RPC remote interface.

**Context: Fixes an issue?**

No.

Yes, the new method should be documented in the JSON-RPC file. I added Doxygen description to the new function similar to what was done with the other functions. It seems that the documentation file is generated from this. I would need your help how to update the documentation (is it automatically done somehow?).

I tested my implementation as follows:

Start Jamulus with: ./Jamulus --jsonrpcport 11223 --jsonrpcsecretfile secret.txt
In a Linux shell: nc localhost 11223
Then:
{"id":1,"jsonrpc":"2.0","method":"jamulus/apiAuth","params":{"secret": "my_secret_which_i_do_not_show_here"}}
{"id":1,"jsonrpc":"2.0","method":"jamulusclient/setFaderLevel","params":{"channelIndex": 0, "level": 50}}

That worked for me. The fader of the first channel moved after the last command.

As stated above: The documentation may be updated first.

## Checklist

### DIFF
--- a/src/client.h
+++ b/src/client.h
@@ -260,7 +260,7 @@ public:
     void OnTimerRemoteChanGainOrPan();
     void StartTimerGainOrPan();
 
-    void SetControllerInFaderLevel ( int iChannelIdx, int iValue );
+    void SetControllerInFaderLevel ( int iChannelIdx, int iValue )  { OnControllerInFaderLevel ( iChannelIdx, iValue ); }
 
     void SetInputBoost ( const int iNewBoost ) { iInputBoost = iNewBoost; }
 
@@ -434,7 +434,7 @@ protected slots:
     void OnCLPingWithNumClientsReceived ( CHostAddress InetAddr, int iMs, int iNumClients );
 
     void OnSndCrdReinitRequest ( int iSndCrdResetType );
-    void OnControllerInFaderLevel ( int iChannelIdx, int iValue ) { SetControllerInFaderLevel ( iChannelIdx, iValue ); }
+    void OnControllerInFaderLevel ( int iChannelIdx, int iValue );
     void OnControllerInPanValue ( int iChannelIdx, int iValue );
     void OnControllerInFaderIsSolo ( int iChannelIdx, bool bIsSolo );
     void OnControllerInFaderIsMute ( int iChannelIdx, bool bIsMute );
--- a/src/client.h
+++ b/src/client.h
@@ -260,6 +260,8 @@ public:
     void OnTimerRemoteChanGainOrPan();
     void StartTimerGainOrPan();
 
+    void SetControllerInFaderLevel ( int iChannelIdx, int iValue );
+
     void SetInputBoost ( const int iNewBoost ) { iInputBoost = iNewBoost; }
 
     void SetRemoteInfo() { Channel.SetRemoteInfo ( ChannelInfo ); }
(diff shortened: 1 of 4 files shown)


pljones:
To keep _strict separation of concerns_, there should be no direct calls to CClient methods.  CClient should have public slots to receive signals for changes that is its concern, such as tracking state.

Ideally, we'd not be adding more code to the existing method of doing it -- the "OnXxxx" method names are _more correct_, as they imply they're handling signals, which is what they should be doing.  So I'm going to object to renaming the method.

corrados:
Hi Peter, thanks for you comment.

My preferred solution would have be to do it similar as in soundbase: https://github.com/jamulussoftware/jamulus/blob/main/src/sound/soundbase.cpp#L381. But since the RPC class is not a member of the client, I cannot use the signal/slot mechanism. The way it is implementat that the RPC class communicates with the client is with public access functions provided by the client, e.g.: https://github.com/jamulussoftware/jamulus/blob/main/src/clientrpc.cpp#L260.

There is already a similar public function called SetRemoteGain (https://github.com/jamulussoftware/jamulus/blob/main/src/client.h#L258), but this does something slightly different and cannot be used by a "controller in" action.

What I now did was actually not a renaming of an existing function but to introduce a new one. The old function still exists but has a different body: https://github.com/jamulussoftware/jamulus/pull/3571/files#diff-2ee628bb1cdcbf04f484b57ea20ddf172db2dfaeb22214ebfb24ff36718a7509R437

src/client.h
```diff
@@ -260,6 +260,8 @@ class CClient : public QObject
     void OnTimerRemoteChanGainOrPan();
     void StartTimerGainOrPan();
 
+    void SetControllerInFaderLevel ( int iChannelIdx, int iValue ) { OnControllerInFaderLevel ( iChannelIdx, iValue ); }
```
pljones, line 263: Oh I see -- all the slots are currently protected...

So, "normally"...
```
src/client.cpp:161:    QObject::connect ( &Sound, &CSound::ControllerInFaderLevel, this, &CClient::OnControllerInFaderLevel );
```
is triggered by
```
src/sound/soundbase.cpp:381:                            emit ControllerInFaderLevel ( cCtrl.iChannel, iFaderLevel );
```
Ideally, it would be `CSound` wiring itself to public slots in CClient but it doesn't get a reference.

Conversely, CClient doesn't get a reference to RpcClient.  Hm.

I think I'd rather have the slots made public and connect them in RpcClient, if that's feasible -- it will follow how CSound does things more closely.

Ideally, CClient would get instantiate and passed by thread-safe reference to CSound, as well, so that it could connect to the public slots.

Again -- none of the above for here.  I'll live with this.

pljones (approved):

softins (approved):
Code looks good. I've tested it on a Pi, and the RPC successfully moves the correct fader and adjusts the volume heard.

ann0see (approved):
Can confirm that it works on macOS too.

ann0see:
I've removed needs documentation.

PR #3573, opened 2025-12-06
Try to fix QTreeWidget accessibility bug

ann0see:
Co-authored-by: GitHub Copilot

This might contain a lot of low quality code for now.

How it looks like:

image

There is a "Show/Hide accessible controls button" at the top which shows/hides the previous/next buttons + the current server info label.

Adds forward/backward buttons and label to connect dialog for accessability
Copilot's suggestion to fix: #3559 (together with some tweaking)

CHANGELOG: Accessability: Add buttons and label to go through server list

**Context: Fixes an issue?**
Yes: #490 (workaround)

Yes

Proof of concept. The code is probably extremely convoluted.

Testing, review and simplification.

## Checklist

### DIFF
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -74,22 +74,22 @@ static QString mapVersionStr ( const QString& versionStr )
 // Subclass of QTreeWidgetItem that allows LVC_VERSION to sort by the UserRole data value
 CMappedTreeWidgetItem::CMappedTreeWidgetItem ( QTreeWidget* owner ) : QTreeWidgetItem ( owner ), owner ( owner ) {}
 
-bool CMappedTreeWidgetItem::operator< ( const QTreeWidgetItem& other ) const
+bool CMappedTreeWidgetItem::operator<( const QTreeWidgetItem& other ) const
 {
     if ( !owner )
-        return QTreeWidgetItem::operator< ( other );
+        return QTreeWidgetItem::operator<( other );
 
     int column = owner->sortColumn();
 
     // we only need this override for comparing server versions
     if ( column != CConnectDlg::LVC_VERSION )
-        return QTreeWidgetItem::operator< ( other );
+        return QTreeWidgetItem::operator<( other );
 
     QVariant lhs = data ( column, Qt::UserRole );
     QVariant rhs = other.data ( column, Qt::UserRole );
 
     if ( !lhs.isValid() || !rhs.isValid() )
-        return QTreeWidgetItem::operator< ( other );
+        return QTreeWidgetItem::operator<( other );
 
     return lhs.toString() < rhs.toString();
 }


src/connectdlg.cpp
```diff
@@ -215,6 +215,90 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     // per default the root shall not be decorated (to save space)
     lvwServers->setRootIsDecorated ( false );
 
+#ifdef USE_ACCESSIBLE_SERVER_LIST
+    // Create simplified accessible navigation panel for screen readers
+    // Design: One info label + 2 navigation buttons (Previous/Next) + Toggle button
```
ann0see, line 219: Will add screenshot of macOS version

src/connectdlg.cpp
```diff
@@ -215,6 +215,90 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
...
+    lblAccessibleServerInfo->setMinimumHeight ( 50 );
+    lblAccessibleServerInfo->setFocusPolicy ( Qt::StrongFocus ); // Make it focusable for screen readers
+    lblAccessibleServerInfo->setAccessibleName ( tr ( "Current server information" ) );
+    lblAccessibleServerInfo->setAccessibleDescription ( tr ( "Shows details of the currently selected server. Use Previous/Next buttons or Alt+Up/Down to navigate." ) );
+    accessibleMainLayout->addWidget ( lblAccessibleServerInfo );
+    
+    // Create horizontal layout for navigation buttons (Previous, Next)
+    QHBoxLayout* navLayout = new QHBoxLayout();
+    
+    butAccessiblePrevious = new QPushButton ( u8"\u2190 " + tr ( "Previous Server" ), wAccessibleNavPanel );
```
ann0see: Unicode magic is not nice.

ann0see: Also, this does not go to the previous server, but rather to the previous entry - so it could also be a musician.

src/connectdlg.cpp
```diff
@@ -631,6 +725,215 @@ void CConnectDlg::OnServerAddrEditTextChanged ( const QString& )
...
+            labelText += tr ( " &nbsp; <b>Musicians:</b> %1" ).arg ( musicians );
+        }
+        if ( !location.isEmpty() )
+        {
+            labelText += tr ( "<br><b>Location:</b> %1" ).arg ( location );
+        }
+        if ( !version.isEmpty() )
+        {
+            labelText += tr ( " &nbsp; <b>Version:</b> %1" ).arg ( version );
+        }
```
ann0see: Seems like a lot of duplication...

src/connectdlg.cpp
```diff
@@ -631,6 +725,215 @@ void CConnectDlg::OnServerAddrEditTextChanged ( const QString& )
...
+        
+        // Build text for screen readers (without HTML)
+        QString accessibleText = tr ( "Server: %1" ).arg ( serverName );
```
ann0see, line 744: I believe this text is not updated. I'd be in favour of removing it and replacing it just with "Server information".

src/connectdlg.cpp
```diff
@@ -215,6 +215,90 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     // per default the root shall not be decorated (to save space)
     lvwServers->setRootIsDecorated ( false );
 
+#ifdef USE_ACCESSIBLE_SERVER_LIST
```
pljones: So this will be an additional build.  Will it be included in the macOS bundle so it all gets signed once in the Apple Store?

src/connectdlg.cpp
pljones: As a minimum, the unicode characters want putting into constants, named so someone reading the code gets some idea of what they represent when rendered.

src/connectdlg.cpp
pljones: Actually, does this need to be a compile time option?  Just have the accessibility panel hidden unless turned on in Settings->User Profile->User Interface.

src/connectdlg.cpp
```diff
@@ -215,6 +215,90 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
...
+            qWarning ( "Accessible navigation panel could not be inserted: tree widget not found in layout." );
+        }
+    }
+    else
+    {
+        qWarning ( "Accessible navigation panel could not be inserted: main layout cast failed." );
+    }
+    
+    // Initially show the accessible panel
+    wAccessibleNavPanel->setVisible ( true );
```
pljones, line 299: Using the setting, you'd set to false immediate after creating and based on the setting here.

src/connectdlg.cpp
```diff
@@ -266,6 +353,13 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
     QObject::connect ( &TimerPing, &QTimer::timeout, this, &CConnectDlg::OnTimerPing );
 
     QObject::connect ( &TimerReRequestServList, &QTimer::timeout, this, &CConnectDlg::OnTimerReRequestServList );
+
+#ifdef USE_ACCESSIBLE_SERVER_LIST
```
pljones: If the widget exists (even if not shown), these need not be compile-time.

src/connectdlg.cpp
pljones: You'd need to track a change in the setting, though.

src/connectdlg.cpp
ann0see: Not sure if we even need those characters.

src/connectdlg.cpp
ann0see: Agree. I was also not sure if it should or should not be. Best case, it's just a button/checkbox.

src/connectdlg.cpp
```diff
@@ -215,6 +215,90 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
...
+    
+    QVBoxLayout* accessibleMainLayout = new QVBoxLayout ( wAccessibleNavPanel );
+    accessibleMainLayout->setContentsMargins ( 0, 5, 0, 5 );
+    
+    // Create horizontal layout for navigation buttons (Previous, Next)
+    QHBoxLayout* navLayout = new QHBoxLayout();
+    
+    butAccessiblePrevious = new QPushButton ( u8"\u2190 " + tr ( "Previous Server" ), wAccessibleNavPanel );
+    butAccessiblePrevious->setObjectName ( "butAccessiblePrevious" );
+    butAccessiblePrevious->setAccessibleName ( tr ( "Go to previous server" ) );
```
pljones: Yeah, "Previous Button" and "Next Button" would be better for these, I think.

src/connectdlg.cpp
```diff
@@ -215,6 +215,90 @@ CConnectDlg::CConnectDlg ( CClientSettings* pNSetP, const bool bNewShowCompleteR
...
+    butAccessiblePrevious->setAccessibleDescription ( tr ( "Navigate to the previous server in the list" ) );
```
pljones: Then these can be the more vague but more accurate "Moves to the previous server list entry" and "Moves to the next server list entry"

src/connectdlg.cpp
```diff
@@ -631,6 +725,196 @@ void CConnectDlg::OnServerAddrEditTextChanged ( const QString& )
...
+    UpdateAccessibleServerInfo();
+#endif
+}
+
+#ifdef USE_ACCESSIBLE_SERVER_LIST
+void CConnectDlg::UpdateAccessibleServerInfo()
+{
+    QList<QTreeWidgetItem*> selectedItems = lvwServers->selectedItems();
+
+    if ( !selectedItems.isEmpty() && selectedItems.first()->parent() == nullptr )
```
pljones, line 731: Need to comment this to explain it has to handle Server entries and Client entries separately.

src/connectdlg.cpp
```diff
@@ -631,6 +725,196 @@ void CConnectDlg::OnServerAddrEditTextChanged ( const QString& )
...
+
+        // No server selected or musician child selected
+        lblAccessibleServerInfo->setText ( tr ( "<i>No server selected or in musician selection</i>" ) );
+        lblAccessibleServerInfo->setAccessibleName ( tr ( "No server selected or in musician selection" ) );
+        lblAccessibleServerInfo->setAccessibleDescription ( tr ( "No server selected. Use Previous/Next buttons or Alt+Up/Down to navigate servers." ) );
+        
+    }
+}
+
+void CConnectDlg::OnAccessiblePreviousClicked()
```
pljones, line 828: So these will trigger `UpdateAccessibleServerInfo` through
```
QObject::connect ( lvwServers, &QTreeWidget::itemSelectionChanged, this, &CConnectDlg::OnServerListItemSelectionChanged );
```

pljones: I don't think they should skip Clients -- it needs to step through _all_ entries in the list, otherwise the list isn't fully accessible.

src/connectdlg.cpp
```diff
@@ -631,6 +725,196 @@ void CConnectDlg::OnServerAddrEditTextChanged ( const QString& )
...
 
+void CConnectDlg::OnServerListItemSelectionChanged()
+{
+#ifdef USE_ACCESSIBLE_SERVER_LIST
+    UpdateAccessibleServerInfo();
+#endif
+}
+
+#ifdef USE_ACCESSIBLE_SERVER_LIST
+void CConnectDlg::UpdateAccessibleServerInfo()
```
pljones, line 727: Needs tests showing:
1. Empty filter, Show all musicians off
Image
2. Empty filter, Show all musicians on
Image
3. Occupied servers filter on, Show all musicians off
Image
3. Occupied servers filter on, Show all musicians on
Image

ann0see:
Maybe we could implement the following workaround:

1. extract the currently selected cell text
2. display it as text field or alert
3. have a keyboard shortcut to jump there and let it read aloud with the screen reader. Then jump back with the same shortcut.

PR #3592, opened 2026-01-08
central defense by ASN or IP4 mask

mcfnord:
feat(server): Add CentralDefense for ASN/CIDR blocking

This introduces the `CentralDefense` module, a server-side security mechanism designed to reject connections from specific Autonomous Systems (ASNs) and IP ranges (CIDRs). This allows a central service to filter out traffic from known abuse sources.

Key Implementation Details:

1. synchronous "Bouncer" Logic:
   - Hooks into `CServer::OnNewConnection` immediately after the mutex lock.
   - Uses `Qt::DirectConnection` to ensure the block check completes synchronously.
   - If a client is blocked, the connection is dropped *before* the server sends the "Welcome" message or updates the connected client list.

2. API Protection & Throttling:
   - Integrates with `ip-api.com` for ASN lookups but implements aggressive protection for the API provider.
   - **Memory-First:** Checks a local RAM cache and static CIDR list before attempting any network lookup.
   - **Queuing:** Limits concurrent requests to prevent bursts.
   - **Throttling:** Enforces a delay between outbound requests to respect API rate limits.
   - **Negative Caching:** Caches lookup results (allow or deny) to ensure subsequent connections from the same IP require zero network traffic.

3. Configuration:
   - Fetches a remote blocklist (ASN/CIDR format) on startup and refreshes periodically.
   - Parses blocklist entries robustly, ignoring comments and descriptive text.

## Checklist

### DIFF
--- a/src/centraldefense.cpp
+++ b/src/centraldefense.cpp
@@ -1,0 +1,433 @@
+#include "centraldefense.h"
+
+#include <QNetworkRequest>
+#include <QNetworkReply>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QDebug>
+#include <QHostAddress>
+#include <QRegExp>
+#include <QElapsedTimer>
+#include <QLoggingCategory>
+
+Q_LOGGING_CATEGORY(lcCentralDefense, "jamulus.centraldefense")
+
+CentralDefense::CentralDefense(const QUrl& blockListUrl,
+                               const QUrl& asnLookupBase,
+                               int refreshIntervalSeconds,
+                               QObject* parent)
+    : QObject(parent),
+      m_blockListUrl(blockListUrl),
+      m_asnLookupBase(asnLookupBase),
+      m_refreshIntervalSeconds(refreshIntervalSeconds)
+{
+    m_nam = new QNetworkAccessManager(this);
+    m_timer = new QTimer(this);
+    m_timer->setInterval(qMax(1, m_refreshIntervalSeconds) * 1000);
+    connect(m_timer, &QTimer::timeout, this, &CentralDefense::onTimerTick);
+}
+
+CentralDefense::~CentralDefense()
+{
+    stop();
+    if (m_inflightBlockList) { m_inflightBlockList->abort(); m_inflightBlockList->deleteLater(); m_inflightBlockList = nullptr; }
+    if (m_inflightReply) { m_inflightReply->abort(); m_inflightReply->deleteLater(); m_inflightReply = nullptr; }
+    if (m_inflightTimeoutTimer) { m_inflightTimeoutTimer->stop(); m_inflightTimeoutTimer->deleteLater(); m_inflightTimeoutTimer = nullptr; }
+}
+
+void CentralDefense::start()
+{
+    qCInfo(lcCentralDefense) << "starting: blocklist" << m_blockListUrl.toString()
+                             << "lookup_base" << m_asnLookupBase.toString()
+                             << "refresh_s" << m_refreshIntervalSeconds
+                             << "queue_cap" << m_maxPending;
+    fetchBlockList();
+    m_timer->start();
+}
+
+void CentralDefense::stop()
+{
+    qCInfo(lcCentralDefense) << "stopping";
+    m_timer->stop();
+}
+
+void CentralDefense::onTimerTick()
+{
+    purgeExpired();
+    fetchBlockList();
+}
+
+void CentralDefense::fetchBlockList()
+{
+    if (!m_blockListUrl.isValid()) {
+        qCWarning(lcCentralDefense) << "invalid block list URL";
+        return;
+    }
+    if (m_inflightBlockList) return; 
+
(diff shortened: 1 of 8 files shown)


src/centraldefense.cpp
```diff
@@ -0,0 +1,433 @@
...
+        QWriteLocker locker(&m_lock);
+        m_blockedCidrs = std::move(newCidrs);
+        m_blockedAsns = std::move(newAsns);
+    }
+
+    // qCInfo(lcCentralDefense) << "blocklist: fetch ok - ASNs:" << m_blockedAsns.size() << "CIDRs:" << m_blockedCidrs.size();
+    emit updated(m_blockedAsns.size(), m_blockedCidrs.size());
+}
+
+bool CentralDefense::tryParseIpv4CidrLine(const QString& line, Ipv4Cidr& outCidr)
```
ann0see, line 136: You should write the code generic such that it supports ipv6 too

ann0see (changes requested):
Very interesting! Thank you. 
I believe that we'll have a long discussion here...

src/server.cpp
```diff
@@ -301,6 +302,34 @@ CServer::CServer ( const int          iNewMaxNumChan,
 
     connectChannelSignalsToServerSlots<MAX_NUM_CHANNELS>();
 
+
+    // --- Central Defense Integration ---
+    m_centralDefense = new CentralDefense(QUrl("https://jamulus.live/asn-ip-client-blocks.txt"),
```
softins, line 307: Aha! This would explain why Jamulus Explorer is no longer able to display City, Version, OS and connected clients for the "Duet" server that exists in each genre:

Image

stefan1000:
> All great examples for implementing @Rob-NY RPC firewall. It doesn't make external calls, currently only blocks single IPs, could potentially be managed by clients directly via RPC. Any further protection should be handled at the hoster/OS firewall level.

A bad actor should not even reach the application, all of this can be done on firewall level (peek into the UDP payload if it is a jamulus client connect msg, then hand it over to some user space script that does further checking)

See e.g. https://tech-champion.com/ethical-hacking/linux-packet-payload-manipulation-netfilter-and-nfqueue-solutions/

src/server.cpp
mcfnord: Yes. There's a script version and a binary version, and the binary version is doing that (which is a bug).

stefan1000:
> > All great examples for implementing @Rob-NY RPC firewall.
> > It doesn't make external calls, currently only blocks single IPs, could potentially be managed by clients directly via RPC.
> > Any further protection should be handled at the hoster/OS firewall level.
> 
> I know of two actors who harass groups through a VPN that shifts their IP any time they want. I know of zero actors who harass groups through a singular IP address. That's why I think blocking an single IP is inadequate. And these actors target groups on multiple servers, including servers that are unmonitored. I'd like a defense that would protect against much larger attacks, and multi-IP plus unmonitored servers are facts of life any defense of the larger community needs to cover.

A jamulus server operator is free to limit access to whatever, just saying this should happen in the network middleware, not in the application.

There a valid use cases for VPNs, also for jamulus, like getting reliable UDP+IPv4 when behind a CGNAT, or avoiding sub-optimal routes because an ISP choose to take a few long latency hops more to not pay for transit. 

IP is not an identity, nor is an ASN, it is an ongoing race. If a stronger (or some kind of) user identity is needed, one could require users to authenticate via an OpenID provider (e.g., Google, Apple, Facebook, Amazon, Microsoft) before allowing access.

mcfnord:
> A jamulus server operator is free to limit access to whatever, just saying this should happen in the network middleware, not in the application.

You're right about the architectural purity of separating these concerns. Ideally, this logic _belongs_ in the middleware. My primary constraint is the 'human layer': most server operators here aren't network engineers. They need a drop-in binary that works out of the box without configuring external firewalls or middleware. 


> There a valid use cases for VPNs, also for jamulus, like getting reliable UDP+IPv4 when behind a CGNAT, or avoiding sub-optimal routes because an ISP choose to take a few long latency hops more to not pay for transit.

I know users who rely on VPNs for exactly those reasons. Is there a way we can distinguish the legitimate routing/CGNAT traffic from the mass-bot/griefer traffic, or are we forced to choose between safety and accessibility?

> IP is not an identity, nor is an ASN, it is an ongoing race. 

IP is definitely not a _strict_ identity. I view IP less as a 'fingerprint' and more as a 'contextual signal.' It’s squishy, but in a decentralized system, even a loose, probabilistic match lets me offer features—like associating a visitor with a likely persona—that I couldn't otherwise. I’m comfortable with it being a 'soft' match (70% confidence) rather than a hard, certain authentication where the stakes for errors are low.

> If a stronger (or some kind of) user identity is needed, one could require users to authenticate via an OpenID provider (e.g., Google, Apple, Facebook, Amazon, Microsoft) before allowing access.

OpenID is maybe the perfect long-term fix. It would also open up things like regional chat and social notifications. If you were architecting this, would you force a specific provider, or just support the standard generic OIDC libraries? Discord seems big in Jamulus; Google's just big.

PR #3606, opened 2026-01-25
Accessible Server List

chigkim:
Make the server list accessible by making the screen reader to announce relevant information when a server or a client is selected.

* Open connect dialog
* Press tab to focus on the table.
* Press up/down arrow to select server. If you have a screen reader on, it'll announce all the information.
* Press right/left arrow to expand/collapse.
* * If you select a client, it'll also announce instrument and country.

@pljones @ann0see

CHANGELOG: Client: Add screen-reader accessibility to server list in the Connect dialog

### DIFF
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -1126,6 +1126,7 @@ void CConnectDlg::UpdateDirectoryComboBox()
 
 void CConnectDlg::OnCurrentServerItemChanged ( QTreeWidgetItem* current, QTreeWidgetItem* )
 {
+    // Announce the currently selected server or client to screen readers.
 #if QT_VERSION >= QT_VERSION_CHECK( 6, 8, 0 )
     if ( !current )
         return;


src/connectdlg.cpp
```diff
@@ -75,22 +80,22 @@ static QString mapVersionStr ( const QString& versionStr )
 // Subclass of QTreeWidgetItem that allows LVC_VERSION to sort by the UserRole data value
 CMappedTreeWidgetItem::CMappedTreeWidgetItem ( QTreeWidget* owner ) : QTreeWidgetItem ( owner ), owner ( owner ) {}
 
-bool CMappedTreeWidgetItem::operator<( const QTreeWidgetItem& other ) const
+bool CMappedTreeWidgetItem::operator< ( const QTreeWidgetItem& other ) const
```
ann0see: These spaces break the style guide. 

```suggestion
bool CMappedTreeWidgetItem::operator<( const QTreeWidgetItem& other ) const
```

src/connectdlg.cpp
```diff
@@ -24,6 +24,11 @@
 
 #include "connectdlg.h"
 
+#if QT_VERSION >= QT_VERSION_CHECK( 6, 8, 0 )
```
ann0see, line 27: Is there a reason that we can't use this approach on Qt5 too?

chigkim: > Is there a reason that we can't use this approach on Qt5 too?

I could be wrong, but I believe QAccessibleAnnouncementEvent class became available in qt6.8.0.

src/connectdlg.cpp
chigkim: > These spaces break the style guide.

For some reason, clang-format changed that. It wasn't part of my change. Is there like particular way I should run clang-format?

ann0see: Yes. It's a known issue. You should use the clang format version which the CI uses: see the files in the .github folder.

chigkim: > Yes. It's a known issue. You should use the clang format version which the CI uses: see the files in the .github folder.

Ok, thanks!
I made it tighter by remove clang format that was not part of my change.
Just to be clean, I just pushed with --force instead of revert since it's just me.
Hopefully it passes. :)

src/connectdlg.cpp
```diff
@@ -1106,3 +1113,30 @@ void CConnectDlg::UpdateDirectoryComboBox()
...
+        QVariant countryData = current->data ( LVC_NAME, Qt::UserRole );
+        if ( countryData.isValid() )
+        {
+            announcement += ", " + countryData.toString();
+        }
+    }
+    else
+    {
+        // It's a server
+        announcement = current->text ( LVC_NAME ) + ", " + current->text ( LVC_CLIENTS ) + ", " + current->text ( LVC_LOCATION ) + ", " +
```
ann0see: In the actual dialog the order is: Server Name, Ping Time, Musicians (number), Location and then version

ann0see: Also I realized, that the number of musicians doesn't immediately update in the spoken text. If you move away and then back, it is correct though.

src/connectdlg.cpp
ann0see: Ok. Then it's fine.

src/connectdlg.cpp
```diff
@@ -1106,3 +1123,35 @@ void CConnectDlg::UpdateDirectoryComboBox()
         }
     }
 }
+
+void CConnectDlg::OnCurrentServerItemChanged ( QTreeWidgetItem* current, QTreeWidgetItem* )
```
ann0see, line 1127: Add an explanation inside the function that it's used for a11y.

chigkim: Inside OnCurrentServerItemChanged?
`// Post an announcement notifications to screen readers with information of the selected item.``
Would this enough?

src/connectdlg.cpp
```diff
@@ -1106,3 +1123,36 @@ void CConnectDlg::UpdateDirectoryComboBox()
...
+
+    if ( !current->parent() && current->text ( LVC_PING ).isEmpty() )
+    {
+        return;
+    }
+
+    QString announcement;
+    if ( current->parent() )
+    {
+        // It's a client
```
pljones, line 1142: Nice - future enhancement to have this string configurable: some people may be more interested in what instrument is being played than where someone is from.

chigkim: Right now it reports clients with name, instrument, country.

PR #3622, opened 2026-02-07
IPv6 QoS fix

rdica:

When IPv6 is enabled, this provides a fix to properly typecast `tos` so setsockopt can properly add DSCP field value.
Also adds error checking for `setsockopt()` calls and ensures correct QoS operation for dual-protocol sockets.

CHANGELOG: Client+Server: Correct QoS setting for IPv6

**Context: Fixes an issue?**

Fixes: #3623 

No, just a bugfix

Needs testing on all but Linux OSes.

## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -222,8 +222,12 @@ int main ( int argc, char** argv )
         // Quality of Service --------------------------------------------------
         if ( GetNumericArgument ( argc, argv, i, "-Q", "--qos", 0, 255, rDbleArgument ) )
         {
+#if defined( Q_OS_WIN )
+            qWarning() << "QoS is currently not available under Windows - ignoring";
+#else
             iQosNumber = static_cast<quint16> ( rDbleArgument );
             qInfo() << qUtf8Printable ( QString ( "- selected QoS value: %1" ).arg ( iQosNumber ) );
+#endif
             CommandLineOptions << "--qos";
             continue;
         }
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -129,7 +129,7 @@ void CSocket::Init ( const quint16 iNewPortNumber, const quint16 iNewQosNumber, 
         }
 #endif
 
-#if !defined( Q_OS_DARWIN )
+#if !defined( Q_OS_DARWIN ) && !defined( Q_OS_WIN )
         // set the QoS for IPv4 as well, as this is a dual-protocol socket
         if ( setsockopt ( UdpSocket, IPPROTO_IP, IP_TOS, (const char*) &tos, sizeof ( tos ) ) == -1 )
         {
@@ -162,12 +162,14 @@ void CSocket::Init ( const quint16 iNewPortNumber, const quint16 iNewQosNumber, 
             throw CGenErr ( "IPv4 requested but not available on this system.", "Network Error" );
         }
 
+#if !defined( Q_OS_WIN )
         // set the QoS
         const int tos = (int) iQosNumber; // Quality of Service
         if ( setsockopt ( UdpSocket, IPPROTO_IP, IP_TOS, (const char*) &tos, sizeof ( tos ) ) == -1 )
         {
             throw CGenErr ( "request to set ToS for IPv4 failed", "Network Error" );
         }
+#endif
 
         // preinitialize socket in address (only the port number is missing)
         UdpSocketAddr.sa4.sin_family      = AF_INET;


src/socket.cpp
```diff
@@ -118,8 +118,8 @@ void CSocket::Init ( const quint16 iNewPortNumber, const quint16 iNewQosNumber,
         setsockopt ( UdpSocket, IPPROTO_IPV6, IPV6_V6ONLY, (const char*) &no, sizeof ( no ) );
 
         // set the QoS
-        const char tos = (char) iQosNumber; // Quality of Service
-        setsockopt ( UdpSocket, IPPROTO_IPV6, IPV6_TCLASS, &tos, sizeof ( tos ) );
+        const int tos = (int) iQosNumber; // Quality of Service
```
ann0see, line 124: I believe this should be tested on windows.

https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-setsockopt

softins (commented):
I've been doing some tests of this on Linux (RPi), and it needs a little more investigation.

I verified that the `setsockopt()` previously returned an error, and that with this fix it succeeds.

I also tested it with different QoS values for server and client, running on the same machine connecting via localhost (`::1`) on IPv6, while monitoring the traffic with Wireshark. The correct QoS (DSCP+ECN) was sent from each end.

However, if the client initiated the connection to `127.0.0.1` instead, to force an IPv4 connection, both ends sent a QoS value of `0`.

If I started the client without `-6`, and connected to `127.0.0.1` or to `localhost`, then the traffic from the client to the server carried the expected QoS value, but that from the server (still running with `-6`, so a dual-protocol socket) back to the client sent QoS of `0`.

So I need to check how to get a dual-stack socket to send the correct QoS value when it has an IPv4 connection. Maybe it will also accept the `setsockopt ( UdpSocket, IPPROTO_IP, IP_TOS, ...)` call?

src/socket.cpp
softins: I'll do that soon.

pljones:
On Linux, it looks like I use `-Q 184` for Servers and `-Q 96` for Directories (with `-u 1` - not intended for real-time audio).  So I guess it does need to be flexible.

softins:
> Personally, if we're going to be making changes, I'd rather they were complete. I'm not seeing a pressing need to include it in 3.12.0 yet.

I have just pushed a new commit that adds the missing functionality for macOS sending to IPv4 over an IPv6 socket. I've tested it with Wireshark monitoring and it works.

That just leaves working out how to do it on Windows. I'm going to look at what the Mumble project does.

I think it should be in 3.12.0, with or without Windows support (preferably with, if we can), otherwise it might be another year, or at least several months, before the bug fixes we have already are in a release!

src/socket.cpp
```diff
@@ -270,6 +270,47 @@ CSocket::~CSocket()
...
+// sendto_ipv4_with_tos - helper function for macOS to set TOS when sending IPv4 over IPv6 socket
+static ssize_t sendto_ipv4_with_tos ( int fd, const void* buf, size_t len, int flags, const struct sockaddr* dest, socklen_t destlen, int tos )
+{
+    // The cmsg buffer is only set up once (tos doesn't change) so can be static
+    static union
+    {
+        unsigned char  cbuf[CMSG_SPACE ( sizeof ( int ) )];
+        struct cmsghdr h;
+    } u;
+    static socklen_t clen = 0;
```
pljones, line 288: I think `CMSG_SPACE` and `CMSG_LEN` and `CMSG_DATA` could do with a brief explanation to save looking them up.

I'm _guessing_ the first adds padding to the value of its argument to allow for `cmsg_level`, `cmsg_type` and `cmsg_len` fixed storage and that `CMSG_DATA` similarly offsets from its argument to the start of the message data space.  That would mean `CMSG_LEN` allocates the memory and returns the length.

It's all a bit opaque, though.  If the above is correct, maybe add something similar.

(It also feels incredibly old-school to be having to do these low level calls... 😄 )

softins: > I think `CMSG_SPACE` and `CMSG_LEN` and `CMSG_DATA` could do with a brief explanation to save looking them up.

They are all part of the standard API described in `man 3 cmsg` and referred to in `man 2 sendmsg` (on Linux; the man pages on macOS are less comprehensive, although the implementation is the same).

softins: Added a comment referring to the man page for `cmsg`.

pljones: Thanks.  It was just we don't mostly use low-level code but Qt wrappers, so having this virtually bare memory access needed just a little hint.

src/socket.cpp
```diff
@@ -114,12 +114,28 @@ void CSocket::Init ( const quint16 iNewPortNumber, const quint16 iNewQosNumber,
 
         // The IPV6_V6ONLY socket option must be false in order for the socket to listen on both protocols.
         // On Linux it's false by default on most (all?) distros, but on Windows it is true by default
-        const uint8_t no = 0;
-        setsockopt ( UdpSocket, IPPROTO_IPV6, IPV6_V6ONLY, (const char*) &no, sizeof ( no ) );
+        const int no = 0;
+        if ( setsockopt ( UdpSocket, IPPROTO_IPV6, IPV6_V6ONLY, (const char*) &no, sizeof ( no ) ) == -1 )
+        {
+            throw CGenErr ( "setsockopt for IPV6_V6ONLY failed", "Network Error" );
```
pljones: Can this say "to clear IPV6_V6ONLY" - otherwise people might get confused as to what went wrong.

Also, something "less technical", similar to the line 112 error -- maybe "request to clear block on IPv4 over IPv6 failed"?

src/socket.cpp
```diff
@@ -114,12 +114,28 @@ void CSocket::Init ( const quint16 iNewPortNumber, const quint16 iNewQosNumber,
...
+        }
 
         // set the QoS
-        const char tos = (char) iQosNumber; // Quality of Service
-        setsockopt ( UdpSocket, IPPROTO_IPV6, IPV6_TCLASS, &tos, sizeof ( tos ) );
+        const int tos = (int) iQosNumber; // Quality of Service
+#if !defined( Q_OS_WIN )
+        if ( setsockopt ( UdpSocket, IPPROTO_IPV6, IPV6_TCLASS, (const char*) &tos, sizeof ( tos ) ) == -1 )
+        {
+            throw CGenErr ( "setsockopt for IPV6_TCLASS failed", "Network Error" );
```
pljones:
```suggestion
            throw CGenErr ( "request to set ToS for IPv6 failed", "Network Error" );
```

src/socket.cpp
```diff
@@ -114,12 +114,28 @@ void CSocket::Init ( const quint16 iNewPortNumber, const quint16 iNewQosNumber,
...
+        }
+#endif
+
+#if !defined( Q_OS_DARWIN )
+        // set the QoS for IPv4 as well, as this is a dual-protocol socket
+        if ( setsockopt ( UdpSocket, IPPROTO_IP, IP_TOS, (const char*) &tos, sizeof ( tos ) ) == -1 )
+        {
+            throw CGenErr ( "setsockopt for IP_TOS failed", "Network Error" );
```
pljones:
```suggestion
            throw CGenErr ( "request to set ToS for IPv4 over IPv6 failed", "Network Error" );
```

src/socket.cpp
```diff
@@ -147,8 +163,11 @@ void CSocket::Init ( const quint16 iNewPortNumber, const quint16 iNewQosNumber,
         }
 
         // set the QoS
-        const char tos = (char) iQosNumber; // Quality of Service
-        setsockopt ( UdpSocket, IPPROTO_IP, IP_TOS, &tos, sizeof ( tos ) );
+        const int tos = (int) iQosNumber; // Quality of Service
+        if ( setsockopt ( UdpSocket, IPPROTO_IP, IP_TOS, (const char*) &tos, sizeof ( tos ) ) == -1 )
+        {
+            throw CGenErr ( "setsockopt for IP_TOS failed", "Network Error" );
```
pljones:
```suggestion
            throw CGenErr ( "request to set ToS for IPv4 failed", "Network Error" );
```

src/socket.cpp
softins:
```suggestion
            throw CGenErr ( "request to support IPv4 over IPv6 failed", "Network Error" );
```

src/main.cpp
```diff
@@ -222,9 +222,13 @@ int main ( int argc, char** argv )
         // Quality of Service --------------------------------------------------
         if ( GetNumericArgument ( argc, argv, i, "-Q", "--qos", 0, 255, rDbleArgument ) )
         {
+#if defined( Q_OS_WIN )
+            qWarning() << "QoS is currently not available under Windows - ignoring";
+#else
             iQosNumber = static_cast<quint16> ( rDbleArgument );
             qInfo() << qUtf8Printable ( QString ( "- selected QoS value: %1" ).arg ( iQosNumber ) );
             CommandLineOptions << "--qos";
```
pljones: It should go in `CommandLineOptions` anyway, I think.

pljones (approved):
Do we need to clarify any documentation as part of 3.12.0?

PR #2338, opened 2022-02-03
Add basic translation verification check

ann0see:
**Short description of changes**
Executes windows installer verification script and checkkeys.pl on push to src/res/translation

**Context: Fixes an issue?**
No. There was some discussion around that. 

**Does this change need documentation? What needs to be documented and how?**
No. 

**Status of this Pull Request**
Still being tested. Probably not much to do.

**What is missing until this pull request can be merged?**
Seems to work: https://github.com/ann0see/jamulus/actions/runs/1791182911 (Although we don't seem to check for syntax errors in the win installer test yet)

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md#source-code-consistency) 
- [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I've filled all the content above

PR #3625, opened 2026-02-08
Enable checkkeys in CI

ann0see:

Enables automated run of checkkeys.pl 

CHANGELOG: SKIP

**Context: Fixes an issue?**

No. It was commented out and fixed by #3590 to run again.

Ready for review. 

Probably fine to merge.

## Checklist

### DIFF
--- a/.github/workflows/translation-check.yml
+++ b/.github/workflows/translation-check.yml
@@ -1,28 +1,28 @@
-name:                Verify translations
+name: Verify translations
 
 on:
   workflow_dispatch:
   pull_request:
     paths:
-      - 'src/translation/wininstaller/**'
-      - 'tools/check-wininstaller-translations.sh'
-      - '.github/workflows/translation-check.yml'
+      - "src/translation/wininstaller/**"
+      - "tools/check-wininstaller-translations.sh"
+      - ".github/workflows/translation-check.yml"
   push:
     paths:
-      - 'src/translation/wininstaller/**'
-      - 'tools/check-wininstaller-translations.sh'
-      - '.github/workflows/translation-check.yml'
+      - "src/translation/wininstaller/**"
+      - "tools/check-wininstaller-translations.sh"
+      - ".github/workflows/translation-check.yml"
 
 jobs:
   translation-check:
-    name:            Verify translation
+    name: Verify translation
     permissions:
-      contents:      read
-    runs-on:         ubuntu-latest
+      contents: read
+    runs-on: ubuntu-latest
     steps:
-      - name:        Checkout Code
-        uses:        actions/checkout@v6
-      - name:        "Check Windows installer translations"
-        run:         ./tools/check-wininstaller-translations.sh
-      #- name:       "Check for duplicate hotkeys (will not fail)"
-      #  run:        sudo apt install libxml-simple-perl && cd src/translation/ && perl ./tools/checkkeys.pl
+      - name: Checkout Code
+        uses: actions/checkout@v6
+      - name: "Check Windows installer translations"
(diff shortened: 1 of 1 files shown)


.github/workflows/translation-check.yml
```diff
@@ -24,5 +24,5 @@ jobs:
         uses:        actions/checkout@v6
       - name:        "Check Windows installer translations"
         run:         ./tools/check-wininstaller-translations.sh
-      #- name:       "Check for duplicate hotkeys (will not fail)"
-      #  run:        sudo apt install libxml-simple-perl && cd src/translation/ && perl ./tools/checkkeys.pl
+      - name:       "Check for duplicate hotkeys (will not fail)"
+        run:        sudo apt install libxml-simple-perl && perl ./tools/checkkeys.pl
```
pljones: Does this want `&&`?  What if it fails because the base image already has it installed?

ann0see: I don't think it'll fail if it's installed already. 
In Debian at least it succeeds if it's installed.

softins: See #3558. `checkkeys.pl` no longer uses `libxml-simple-perl`. But see my more general comment (shortly) about the applicability or not of `checkkeys.pl` in CI.

softins (commented):
When I originally created `checkkeys.pl`, it was intended as a helper tool to be run to highlight potential accelerator clashes for manual review. I'm not sure whether there is any value in having it run in CI, since it doesn't provide a success/failure result, and its output will only be seen if someone actually looks at the run output. It only highlights _potential_ clashes where the same `&key` is used more than once in the same module. Many of these are false positives, as it is not able to determine whether these multiple uses are within the same scope or context.

For example, in the client dialog, it will flag up multiple uses of `&S`, but they are within the `File`, `Settings` and `Help` menus, and also the `Settings` button, and therefore will not clash with each other. See the following extract of output from [the run for this PR](https://github.com/jamulussoftware/jamulus/actions/runs/21798428789/job/62889764653?pr=3625):

```
en
	CChatDlg
		Key: E
			&Edit
			Cl&ear Chat History
	CClientDlg+CHelpMenu
		Key: A
			&Analyzer Console...
			Clear &All Stored Solo and Mute Settings
			&About Jamulus...
		Key: C
			&Connection Setup...
			Sort Users by &City
			&Chat
		Key: D
			A&dvanced Settings...
			&Disconnect
		Key: E
			&Edit
			Sort Users by Chann&el
		Key: F
			&File
			Auto-Adjust all &Faders
		Key: H
			C&hat...
			&Help
		Key: I
			Sett&ings
			Sort Users by &Instrument
		Key: L
			&Load Mixer Channels Setup...
			Set All Faders to New Client &Level
		Key: M
			&Mute Myself
			Software &Manual...
		Key: O
			C&onnect
			N&o User Sorting
			C&onnect
		Key: S
			&Save Mixer Channels Setup...
			Audio/Network &Settings...
			&Settings
			Getting &Started...
```

I have wondered from time to time if it's possible for the script to detect such situations, but since the menus are built dynamically by the code, it's extremely hard to do.

We went round this loop in 2022 with https://github.com/jamulussoftware/jamulus/pull/2338#issuecomment-1046256124 referencing 3343167c53f719032071b9533e5a13b50be5cc40

ann0see:
I still think it adds value - as it now outputs something.

.github/workflows/translation-check.yml
```diff
@@ -1,28 +1,28 @@
-name:                Verify translations
```
ann0see, line 1: My editor reformatted this again... But I think this format is also fine?

softins (approved):

PR #3657, opened 2026-04-21
Increase accepted length of version string

softins:

Increase length of accepted version number from 30 to 50 chars. This does not change any messages sent by Jamulus, only making it accept a longer string received as the version number.

CHANGELOG: Server/Client: Allow registration and display of longer version numbers.

**Context: Fixes an issue?**

Replaces #3655. If a build is made from a tree with uncommitted changes, the version string includes "-dirty", which pushes the length of the version number over 30, now that a numeric timestamp has been added. This change allows a directory and a client to correctly parse such a long version number without causing a server registration to fail or a client to fail to display the version number in the connect dialog.

No, bugfix only.

Ready

Nothing

## Checklist

### DIFF
--- a/src/global.h
+++ b/src/global.h
@@ -270,7 +270,7 @@ LED bar:      lbr
 #define MAX_LEN_SERVER_NAME         20
 #define MAX_LEN_IP_ADDRESS          15
 #define MAX_LEN_SERVER_CITY         20
-#define MAX_LEN_VERSION_TEXT        30
+#define MAX_LEN_VERSION_TEXT        50
 
 // define Settings tab indexes
 #define SETTING_TAB_USER     0


pljones (approved):

pljones:
OK, my private Directories and my Servers all updated with this.
image

mcfnord:
i know this is merged but i noticed something that does make me scratch my head:

image

Both Rising and Hot Texas are quite probably from the same source. I do think Rising might be ARM64, while Hot Texas might not be. But what's the 4 about? it looks fishy because the sequence after **dev-** should probably be a hash. Why would a hash add a 4 like that? Does it mean something? Is it related to this code change? Just curious.

softins:
> Both Rising and Hot Texas are quite probably from the same source. I do think Rising might be ARM64, while Hot Texas might not be. But what's the 4 about? it looks fishy because the sequence after **dev-** should probably be a hash. Why would a hash add a 4 like that? Does it mean something? Is it related to this code change? Just curious.

It's just a `git` thing. Commits have a 40-digit hash, but they are often quoted in abbreviated form. For `dev` versions, `Jamulus.pro` uses the command `git describe --match=xxxxxxxxxxxxxxxxxxxx --always --abbrev --dirty` to fetch the abbreviated hash for the commit at `HEAD` and append it to the version string compiled into Jamulus. Older versions of `git` abbreviate the hash to 7 digits, newer ones to 8 digits.

mcfnord:
That's surprising! In both cases I'd say I fetched "the latest" git, vintage 2026. What a twist.

softins:
> That's surprising! In both cases I'd say I fetched "the latest" git, vintage 2026. What a twist.

I meant the version of the `git` program itself that was on the host that built that particular instance of Jamulus.

pljones:
> > That's surprising! In both cases I'd say I fetched "the latest" git, vintage 2026. What a twist.
> 
> I meant the version of the `git` program itself that was on the host that built that particular instance of Jamulus.

Yes -- and unless you install `git` itself from the latest source, you'll get whatever your package manager of choice chooses to give you.  Different operating systems and Linux distros and versions will span _many_ `git` versions.

mcfnord:
I've noticed another version thing I can't explain.

I am running [a server fork with a few modifications](https://github.com/mcfnord/jamulus/tree/jamfan).

When I did this before, the Version got modified... remember the "dev-hexstring" and even the "dirty".

But I just, well, tried to grab the 3.12.1 source and make my changes... and I did! But the Version is appearing as 3.12.1. No dev, no dirty, etc.

I wager that someone could do that intentionally, but I'm not trying to do that.

What triggers the extra modifiers? Do they only appear when the build comes from github? I think my builds are compiled locally.

softins:
If you branched directly from the 3.12.1 tag, you have to change the version in `Jamulus.pro` yourself, to 3.12.1dev. That is what triggers the extra parts.

It's better to branch from `main`, as the version string will then have the dev part already.

PR #3660, opened 2026-04-23
Add new JSON-RPC API calls, with documentation

digable1:

Adds two JSON-RPC API calls:  jamulusclient/getDirectories and jamulusclient/getCurrentDirectory.

jamulusclient/getDirectories: Gets the current list of directories.
jamulusclient/getCurrentDirectory:  Gets the current directory the client is referencing.

JSON-RPC.md has been updated to include documentation for these new API calls

Nothing

Has been smoke tested to validate we're returning what we think we are.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
(100% done under Claude Pro - which means Anthropic explicitly states that no ownership/control is asserted)


ann0see:
Please add a note if this was coded with Claude (in this PR)

digable1:
line re-formatted, should now satisfy clang-format (surprised that C style doesn't like splitting lines to avoid left-right scroll, but ok - so used to Java/Javascript/Typescript styles where that is either encouraged or at least not flagged)

Also added the callout to Claude (as stated in the PR, because I used Claude Pro there is no copyright/ownership issue - Anthropic explicitly calls that out).

ann0see:
Please push the changes to this PR. It doesn't seem to be pushed.

digable1:
Take 2 (yeah I'm supposed to know what I'm doing - whoops!).

pljones:
Is there already a "set current directory" call?  Without one, I'm not sure of the purpose here.

Please clarify how this meets the "keep it simple and small" principle.

digable1:
The purpose is to obtain ping time for that server.  This means I *do* need to know what the current directory is so I can monitor serverInfoReceived and match the address to get it.

I don't need to play puppet to the Jamulus UI by being able to set the current directory (I personally don't have a use case for it, and couldn't image anybody else needing it either).  And not considering setters avoids error handling for parameter validation.

pljones:
It needs more explanation.  Set it in context.  Again, you can write in PHP and use the protocol directly to get the ping time -- the source is available and in use in several tools.  Why do you need to have an actual Jamulus Client available?

digable1:
I have a client-adjenct support tool in development.  For *this* musician's interaction in a performance/rehearsal.  Not within the context of 'general information'.

So in order to do this direct from the server, the musician would have to duplicate their selection twice - once in the Jamulus client and once again in this support tool.  And that would legitimately be a head-scratch as a musician would be correct in asking 'why', missing a critical 'this is what I already told Jamulus what I want' aspect the musician already makes with the client.

digable1:
More context on why I want to use the client with the 'this is what I already told Jamulus what I want' approach:

My client-adjacent tool might be at a stage where it won't break after 10 seconds of trying to use it (it might take 15 seconds).  As a result, I'm nervous while slightly comfortable with exposing it here to provide some insight as to why I want to easily get the ping time from Jamulus (it's _**FAR**_ from release):  https://sync-track.com/synctrack/

This work's got two purposes:
    1. An attempt to provide an aid for syncing remote live performance across continents via Jamulus (slight impediment to doing this naturally:  The laws of physics - so this is an attempt to bend them a bit).
    2. A way of helping musicians not used to playing in 'exotic' time signatures (there was a device called the Trinome from the 1960's that was a first attempt at this - the user manual references it.  And yeah I'm old enough to remember it and acquired one while I was employed)

There's a user manual button within the app, which might help.

I wrote ~5 lines of code and a paragraph on the English language version of the manual - the rest done by Claude with a lot of watching what it does (much more needs to be done - it's very brittle, especially once you get away from happy path).  Right now, Chrome-only (chromium counts) - Firefox has issues.

So I want to automate the ping time value from the jamulus client (right now, it's a manual-entry field - there *is* a Chrome extension that translates from raw TCP connections (JSON-RPC) to HTTP connections to make this possible, and it's currently very clunky without these new API calls).

pljones:
Hm.  That raises an interesting point.

My aim is that CClient "owns" client capabilities, whatever they are (even if it has them in instances of other classes it creates for itself).

Currently we're in a situation where quite a lot of client capability is scattered around in the UI classes - like knowing if a mixer channel is muted and what the "current" directory is.

Ideally, if you're using JSON-RPC, you _shouldn't_ need the GUI to know the chosen current directory - you should be able to off-load the whole of the connection dialogue to a JSON-RPC service controlling the client.

Sadly not possible right now.

If you fancy taking that on, too... 😄

src/clientrpc.cpp
```diff
@@ -183,6 +183,39 @@ CClientRpc::CClientRpc ( CClient* pClient, CClientSettings* pSettings, CRpcServe
...
+        }
+        response["result"] = arrDirectories;
+        Q_UNUSED ( params );
+    } );
+
+    /// @rpc_method jamulusclient/getCurrentDirectory
+    /// @brief Returns the currently selected directory socket address.
+    /// @param {object} params - No parameters (empty object).
+    /// @result {string} result - The socket address of the current directory, usable as params.directory in jamulusclient/pollServerList.
+    pRpcServer->HandleMethod ( "jamulusclient/getCurrentDirectory", [=] ( const QJsonObject& params, QJsonObject& response ) {
```
pljones, line 235: Go to Settings -> Advanced.  The data item you're collecting here is the "current" entry in the custom directories list, I think.

Not Connect->Directory.

src/clientrpc.cpp
```diff
@@ -183,6 +183,39 @@ CClientRpc::CClientRpc ( CClient* pClient, CClientSettings* pSettings, CRpcServe
         response["result"] = "ok";
     } );
 
+    /// @rpc_method jamulusclient/getDirectories
+    /// @brief Returns the list of directories in the same order as presented in Jamulus.
+    /// @param {object} params - No parameters (empty object).
+    /// @result {array} result - Array of directory socket address strings, usable as params.directory in jamulusclient/pollServerList.
+    pRpcServer->HandleMethod ( "jamulusclient/getDirectories", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        QJsonArray arrDirectories;
```
pljones, line 213: Ideally this would be simplified:
- CClient would own "get the user's current directories"
- CConnectDlg would put that in the Directory drop down
- CClientRpc would return that here

Otherwise we're duplicating functionality for no benefit.

src/clientrpc.cpp
pljones: So again, ideally:
- CClient would own "current directory index" which was an index into the list it returned when asked to "get the user's current directories"
- CConnectDlg would emit a "current directory index changed" signal when the selection changed
- CClient would handle that, storing the new value in CClientSettings
- CClientSettings would emit its own "current directory index changed" signal
- CConnectDlg would handle that by retrieving ensuring the dropdown was displayed correctly, then retrieving the new server list from the appropriate entry from "get the user's current directories"

I think there are JSON-RPC calls that can handle emitted signals by emitting new state, aren't there?  They not _just_ responses?
- So CClientRpc would handle the CClientSettings "current directory index changed" signal by sending the new address.

This also begs for a JSON-RPC "setCurrentDirectory", of course, to act the same as selecting a new entry in the dropdown.

digable1:
Yup - you're really going for an n-tiered architecture if I read this correctly.  Where the "source of truth" for all data is encapsulated in the 'API engine'.  And the Jamulus UI is really one of the clients, implemented in Qt with calls to the API based on user interactions (and my tool would be another client that doesn't need much - just ping times in the currently selected directory and server).  That would definitely be ideal - would need to profile to ensure we didn't add too much overhead.

And I noticed that wasn't where we are.  But I am uncomfortable with this kind of re-factor right now being I really just don't have enough context at the present.  For example, some of the choices might have been made for performance reasons to avoid the overhead of going through an API pattern, I really didn't now.  And one can really make things worse if you don't know what you're doing - and I don't.  At least in my case, they were tiny tweaks that didn't touch the core functionality of the client.

Regarding "I think there are JSON-RPC calls that can handle emitted signals by emitting new state, aren't there? They not just responses?":  Yup.  And I have a listener consuming those changes as they are emitted, as that appears to be the only place where the ping time is located on the client.

And your ideal is indeed close id nor on totally on target:

    1. It looks like it's grabbing the address from a central source and util method (it's in class NetworkUtil - that was already there) - so this method appears to be ensuring it's in the type needed (QJasonObject) for JSON-RPC.  In other projects this is a potential safety risk because you're exposing the index direct to the data source, not copying it into another buffer which won't matter if the caller later corrupts the return object (basic 101 software considerations - shallow copy vs deeper copies).  However, In this case, I can see not adding the extra performance hit (albeit small).

    2. I suggest the real issue is as you say, scattering data sources all over the client code rather than encapsulated and organized.

This is not something for somebody just jumping in on first attempt - yet (lots of dialog and bigger picture to obtain first - I'm just poking at an edge to solve an immediate problem, and indeed potentially contributing to spaghetti code).  If this really would be me, I'd be very annoying with the types and volume of questions being asked.

digable1:
So - since:

1. The ideal re-factor is going to take a while
2. Within the current landscape, this is organized (within the RPC file and document)
3. It's a tiny change, adding functionality that makes sense when interacting with a Jamulus client
4. It's really needed for my use case for a client-adjacent tool to address how to improve synchronized performances by automating ping time retrievals in the current software

Can I get approvals from 2 qualifying people as per the requirement?

ann0see:
There's no setCurrentDirectory, I suppose?

digable1:
I didn't want to provide any more setters without further discussion.  The current setters list, especially for the client side of the API, is a bit of a hodge-podge (you have personal ability to change your audio configuration such as fader levels and few other options), but I wasn't certain whether you wanted an outside app to have the capability of affecting server assignments for the client (granted, directories could be visually confusing and not as bad as an outside actor directing which server you're now a part of).

I've got two products that are on top of Jamulus coming up, remote-musician.com and sync-track.com (for 'scheduling upcoming remote events and the ability to play to metronomes or pre-recorded loops in perfect sync - yeah trying to bend a law of physics), and was thinking I doubt the Jamulus development community would want outside apps like these to interfere with a user's interaction if remote playing - even though I could *really* put such capability to good use, but so could others.

And this is the reason I haven't started a further discussion about it.

ann0see:
Ok. I'd argue that it would be fine to have a fully API controllable client. I'm asking as I currently have a bug on macOS where debugging would likely benefit from having a setter - thus I am asking. I hope that this PR will get in soon.

@dingodoppelt @pljones ?

digable1:
I would be all over a full API - I would certainly leverage it.  But I could see security implications if this was over exposed (my paranoid self would even ask why are there any setters - but there are and for many of us, very welcome.  There are so many use cases for this).

I am suggesting this be thought through - and I'm not in the in crowd so I expect I'd be considered 'half-cocked' if I went through and submitted anything in this area (there is this small item called approval of PRs).

dingodoppelt:
I'd like a setter for the connect dialog part of things meaning selecting the directory to poll. I'd argue there could be an extra setter for custom directories like adding and deleting entries.

dingodoppelt:
This works fine:
```
-> http://localhost:3000/methodSelection?method=jamulusclient/getCurrentDirectory
{"id":"0","jsonrpc":"2.0","result":"anygenre1.jamulus.app:22124"}

 -> http://localhost:3000/methodSelection?method=jamulusclient/getDirectories
{"id":"1","jsonrpc":"2.0","result":["anygenre1.jamulus.app:22124","anygenre2.jamulus.app:22224","asia.jamulus.app:22624","rock.jamulus.app:22424","jazz.jamulus.app:22324","classical.jamulus.app:22524","choral.jamulus.app:22724","127.0.0.1","127.0.0.2","127.0.0.3"]}
```
As mentioned above I can imagine two setters `jamulusclient/setCurrentDirectory` and a more tricky one `setNewDirectory` (or `addDirectory` and `deleteDirectory`)

pljones:
My view is every action the Client UI can take should be possible through JSONRPC and every signal sent to the Client UI should be sent through JSONRPC.  Total parity.  Same on the Server side.

That's why I'm against exposing JSONRPC beyond the machine the Client or Server is on _by Jamulus_.  If the machine operator wants to do that, they run their Jamulus Client or Server and control it via JSONRPC rather than the UI.  If they want to expose _their Client_ or _their Server_ over the internet, that's up to them.

digable1:
I totally agree with your second paragraph.

I have a concern with your first paragraph - which is why my very-non-official view for the client is R/O - not write.  Especially for behavior that indirectly affect other clients (which means I understand setting fader levels, etc through the API).

'Parity' with the UI client means being able to join a server via the API.  Which I would argue you *don't* want!

It means a bad actor can communicate with a cloud 'bot' and just overload a server with 5000 kazoo players with the description of 'Mass performance of Highway to Hell' (which -could- be loads of fun in a circumstance or two).

Another use case:  A client joining a server via the API and never leaving.

And another use case:  You can set "Mute myself" and that effects other clients and users on that server.

Which is why I suggested earlier that this needs to be thought through.

dingodoppelt:
> I totally agree with your second paragraph.
> 
> I have a concern with your first paragraph - which is why my very-non-official view for the client is R/O - not write. Especially for behavior that indirectly affect other clients (which means I understand setting fader levels, etc through the API).
> 
> 'Parity' with the UI client means being able to join a server via the API. Which I would argue you _don't_ want!
> 
> It means a bad actor can communicate with a cloud 'bot' and just overload a server with 5000 kazoo players with the description of 'Mass performance of Highway to Hell' (which -could- be loads of fun in a circumstance or two).
> 
> Another use case: A client joining a server via the API and never leaving.
> 
> And another use case: You can set "Mute myself" and that effects other clients and users on that server.
> 
> Which is why I suggested earlier that this needs to be thought through.

This has been possible the entire time and has been exploited as well. The possibility to join a server via RPC is not an invitation to exploit the feature.  
Nothing keeps server admins from using a firewall or scripts like [Jamban](https://github.com/dingodoppelt/jamban) to ban misbehaving clients.  
For the last 24 years this hasn't been a problem so I'd argue it still is none.

digable1:
Ok.  I don't have the historical context.  I shaddup.

pljones:
> This has been possible the entire time and has been exploited as well. The possibility to join a server via RPC is not an invitation to exploit the feature.

Yes - it's not the use of JSONRPC or not that makes it possible for third-party software to act as a Client or a Server: it's the open nature of the protocol.  There are many people running sites the community has benefited from that have needed to hand-code UDP calls.  If using JSONRCP to directly control the Client or Server enables them to avoid potential pitfalls and be able to "plug-n-play" the latest features by keeping Jamulus up to date, I see that as a good thing.

ann0see:
Is there anything blocking approval of this? I'll test it but then it should be fine for inclusion? Otherwise, @digable1 I think you can/should add yourself to util.h as contributor.

src/clientrpc.cpp
ann0see: So you think this is a bug?

mcfnord:
@digable1  Your Remote Musician site looks slick. Let me know if you see a way it might be discoverable from https://jamulus.live

I also suggest a "passwordless" authentication using an existing Discord or Google account. Those two authentication providers would cover a lot of our userbase.

I make inferences about bands that gather together routinely, but if you gather a name for a band or details like that, I could associate that name with that group. I've also begun announcing upcoming gatherings using "custom" welcome messages to users of 80 Jamulus servers on 26 hosts. Maybe we could announce your sign-up events that way. 

> It means a bad actor can communicate with a cloud 'bot' and just overload a server with 5000 kazoo players

Bad actors have done such things. I maintain a list of vectors bad actors use at https://jamulus.live/asn-ip-client-blocks.txt so at least those sources can't reach my servers. I think the platform is deeply committed to platform reliability, but we need a somewhat different faction and stack to keep the public directories reliable and free of bot attacks. I [rely on the central defense in my custom server binary](https://github.com/mcfnord/jamulus/blob/jamfan/src/centraldefense.cpp). The theory being that anyone who wanted to benefit could get a drop-in binary that repels these known attack vectors. Mostly I and my own infrastructure have been specifically targeted for harassment, but this has included griefer attacks on other server populations. At some point, a misanthrope will compel us to adopt effective counter-measures more broadly. I agree that we needn't react until they act, but it's good to have defenses ready.

digable1:
I see the "Some checks were not successful" notation in this PR, and I got an email with the same info.

What am I looking at?

1. When I click on one "Build for Linux .deb arm64", I see "Error:" (in red) - with nothing further.
2. When I look at  the logs under the settings button -> View raw logs, https://productionresultssa13.blob.core.windows.net/actions-results/155d4bc1-6c3a-46d1-9b24-ccbb833d00da/workflow-job-run-0aa5eca1-c4fa-543e-a6bd-4048a1d16536/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-07-15T19%3A15%3A24Z&sig=S9%2FjOco87Q%2BT%2F8UhTaCGvdYOuuj7Y%2FSqIy9emoTPCYo%3D&ske=2026-07-15T22%3A21%3A18Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-07-15T18%3A21%3A18Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-07-15T19%3A05%3A19Z&sv=2025-11-05, it appears there are dependency issues that I'm certainly not aware I touch.
3. I other builds passed, including Linux.deb, iOS, Android, MacOS (both of them), Windows.

Is this something at my end (I'm not sure it is)?

src/clientrpc.cpp
ann0see: @pljones ?

dingodoppelt:
> I see the "Some checks were not successful" notation in this PR, and I got an email with the same info.
> 
> What am I looking at?
> 
>     1. When I click on one "Build for Linux .deb arm64", I see "Error:" (in red) - with nothing further.
> 
>     2. When I look at  the logs under the settings button -> View raw logs, https://productionresultssa13.blob.core.windows.net/actions-results/155d4bc1-6c3a-46d1-9b24-ccbb833d00da/workflow-job-run-0aa5eca1-c4fa-543e-a6bd-4048a1d16536/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-07-15T19%3A15%3A24Z&sig=S9%2FjOco87Q%2BT%2F8UhTaCGvdYOuuj7Y%2FSqIy9emoTPCYo%3D&ske=2026-07-15T22%3A21%3A18Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-07-15T18%3A21%3A18Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-07-15T19%3A05%3A19Z&sv=2025-11-05, it appears there are dependency issues that I'm certainly not aware I touch.
> 
>     3. I other builds passed, including Linux.deb, iOS, Android, MacOS (both of them), Windows.
> 
> 
> Is this something at my end (I'm not sure it is)?

No worries, it seems ubuntu's servers are having issues again. I just restarted the failed jobs.

digable1:
@pljones - this is a stale post but in the interest of 'better late than never':  You posted [re-use comment](https://github.com/jamulussoftware/jamulus/pull/3660#discussion_r3149135538).

My impression is that what you're really saying is a good approach going forward is to step back and take some time and re-factor the product to separate the actions from how they are called.  That way, we have two clients accessing the ability to perform what Jamulus provides.

For example, connect now has common functionality as it has two clients, the UI and API.  So we extract this all action code in a centralized place so it's not scattered all over the codebase (something I didn't do with connect) - and think it through a bit.  Note:  It's actually a "workflow" pattern because the actual UI changes in these actions shouldn't be interwoven into this action (which connect currently is, including connect/disconnect).  It's an architecture change, and if this happens, it should be done in parallel so the current implementation is still working until we're ready (note:  gives us an opportunity to do things like write unit tests, etc) - all behind the scenes, nothing visible.  So the "workflow action" becomes 'current directory', 'connect', etc, with a plug-in pattern to provide hooks into the "pre-action" and "post-action" to provide the specific UI, API, space-time-continuum, implementation of those sections of this action.  Yeah I know - tiny task.

btw, question for my benefit:  Apparently marking a PR a 'draft' didn't convey what I was hoping for - a smoke signal for other people to 'stay' until I fix it and submit it right (which it looks like you did), as my https://github.com/jamulussoftware/jamulus/pull/3783/#issuecomment-4987132085 tried to explain (after the fact).

pljones:
Yes -- there's a central CClient "arbiter" of actions and information that the Jamulus Client can do and knows.  (Similarly, CServer would be for the Jamulus Server.)  Currently, lots of this is scattered without clear CClient co-ordination: lots of pointers to instances of this or that being passed to other classes.

There's a lot to change for this -- one thing is _start up_.  We can't even do that cleanly currently (if I've kept up in my mental model -- I've been to busy elsewhere lately to actually work on Jamulus code).  But things like the main Jamulus Client GUI (the mixing desk view) have too much of the logic _not_ in CClient or a non-GUI class it delegates to.  The Server isn't great, either.  Many command line options have no way to be altered at run time and don't get persisted to Server Settings.

(I'd like to rip out CLI parsing and Settings handling... and probably re-implement the GUI Settings for Client and Server... but no one would like it because it would be very unfriendly.  I know why things like VSCode, etc, have nasty settings interfaces...)

digable1:
Actually:  A different focus for Jamulus itself (probably close to a complete rewrite):

As a "plug-in-module" for the "live performance" section of a musician.  This means:

1. All capabilities are available through the API (as you've mentioned)
2. The current UI becomes one of the 'desk plug-in' options.
3. You can wire up a 'bridge' for Reaper, Audacity, etc. (start with the free DAWs first) - so those DAWs become your desk (adapted for the 'live performance' and not the 'persisted performance' that they serve so there would be desk adaptations) and Jamulus becomes a plugin.  The idea:  One-stop-shopping and technical architecture.
4. Unfortunately there would be duplications:  You don't want to depend on the DAW for EQ - heavier software footprint, more cumbersome (a lot of work if you want to wire up to that DAW, although from a user experience that would be ideal - longlong-term roadmap)

Tiny task - can be done overnight!

ann0see:
@pljones I think you should open an issue with the redesign and we should focus here on getting this in.

dingodoppelt:
> Actually: A different focus for Jamulus itself (probably close to a complete rewrite):
> 
> As a "plug-in-module" for the "live performance" section of a musician. This means:
> 
>     1. All capabilities are available through the API (as you've mentioned)
> 
>     2. The current UI becomes one of the 'desk plug-in' options.
> 
>     3. You can wire up a 'bridge' for Reaper, Audacity, etc. (start with the free DAWs first) - so those DAWs become your desk (adapted for the 'live performance' and not the 'persisted performance' that they serve so there would be desk adaptations) and Jamulus becomes a plugin.  The idea:  One-stop-shopping and technical architecture.
> 
>     4. Unfortunately there would be duplications:  You don't want to depend on the DAW for EQ - heavier software footprint, more cumbersome (a lot of work if you want to wire up to that DAW, although from a user experience that would be ideal - longlong-term roadmap)
> 
> 
> Tiny task - can be done overnight!

You basically describe the setup I use on Linux, i.e. running Jamulus as a "Plugin" launched from a single plugin host (Carla) alongside all my instruments (three windows VSTi amongst them) and effects like dynamics, EQ, MIDI filters and reverb   
It's a one-click thing including all the automagic realtime and performance tunings with no overhead from a bloated DAW. The future is now, we don't need to reinvent it, but I still see much value in stripping it down to a `libjamulus` and build whatever client on top of that. (Don't know if I'd vote for a reverb in a future client, though)

pljones:
ReaRoute, Jack or other virtual audio cabling make Jamulus "plug-able" on Windows, too.  But we've had "what about a plug-in version?" and "what about plug-in support?" on the list of future goals for a while...

dingodoppelt:
> ReaRoute, Jack or other virtual audio cabling make Jamulus "plug-able" on Windows, too. But we've had "what about a plug-in version?" and "what about plug-in support?" on the list of future goals for a while...

I'm all in on that. `libjamulus` would accomplish just that. [DPF](https://github.com/DISTRHO/DPF) is what comes to mind.

ann0see:
Tested yesterday and I'd say it's ok. However, there's still an open comment by pljones.

pljones:
> @pljones I think you should open an issue with the redesign and we should focus here on getting this in.

I'll get something written up.  I've dumped the thread into a markdown document which I'll turn into an issue for discussion.

dingodoppelt (approved):
My tests were successful and I think we should get this in, even though it is 100% generated so I can never be 100% sure if it really does what it should.

src/clientrpc.cpp
ann0see: @digable1 @dingodoppelt I'd like to have a comment here.

pljones: `m_pSettings->iCustomDirectoryIndex` is also suspicious.  That _sounds like_ it's an index into the Custom Directories list in CClientSettings, rather than the selected Connect Dialog value, which is I think what you're after.

(Of course, I don't want you going after anything stored in the UI -- _that_ UI should be storing the value in CSettings via the CClient -- so it _may_ be the right value.)

If it works, it could that it's misnamed.

I'm suspicious because there would have been a setting for the currently selected Directory prior to the concept of adding custom Directories.

pljones:
Here's the issue for the refactor: https://github.com/jamulussoftware/jamulus/issues/3801
- dumped this thread to markdown
- edited by hand down to the most pertinent points
- summarised into presentable Github Issue by Claude Sonnet 5
- hand-edited (partly for style, partly for content) prior to posting

pljones:
> [Merge branch 'jamulussoftware:main' into main](https://github.com/jamulussoftware/jamulus/pull/3660/commits/1fa6ce35b8c45512199d8f29413b74aca98a1b35)
[1fa6ce3](https://github.com/jamulussoftware/jamulus/pull/3660/commits/1fa6ce35b8c45512199d8f29413b74aca98a1b35)

I always recommend rebase rather than merge.  I find it's much safer when multiple people are working on one code base.

dingodoppelt:
Since #3814 was opened I'll close this for cleanup. We have quite a lot of influx at the moment and it can get a little confusing

PR #3664, opened 2026-04-29
FreeBSD fix for dual-stack and IPv6 QoS

### LINKED ISSUE #3478: jamulus-server: -6 makes server IPv6 only on FreeBSD
**Describe the bug**

Description is from [this downstream bug report](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=260861):
```
Adding -6 to the jamulus_server_flags does enable IPv6, but drops the socket for IPv4.

I believe this is due to different defaults between Linux and (Free)BSD. The code opening the socket should specify the correct flag to get a socket that supports both IPv4 and IPv6.
```

**To Reproduce**

See subject.

**Expected behavior**

-6 should only enable IPv6 and not disable IPv4.

**Operating system**

FreeBSD 14.2

softins:

Enhances #3622 to fix an issue with running on FreeBSD, which is subtly different from macOS.

CHANGELOG: FreeBSD: make -6 option work properly with both IPv6 and IPv4 peers.

**Context: Fixes an issue?**

Fixes #3478 and also a crash with #3622 when running with -6 on FreeBSD

No, bug fix only

Tested on both macOS and FreeBSD. No changes to Linux or Windows.

Nothing, but it is important to get this into 3.12.0

## Checklist

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -279,10 +279,16 @@ static ssize_t sendto_ipv4_with_tos ( int fd, const void* buf, size_t len, int f
     // For a description of 'struct cmsghdr' and the 'CMSG_xxx' macros, see 'man 3 cmsg' on a Linux machine.
     // The macOS man pages are less descriptive, but the API is the same, being based on the BSD socket interface.
 
+#    if defined( Q_OS_FREEBSD )
+    using tos_cmsg_type = unsigned char;
+#    else
+    using tos_cmsg_type = int;
+#    endif
+
     // The cmsg buffer is only set up once (tos doesn't change) so can be static
     static union
     {
-        unsigned char  cbuf[CMSG_SPACE ( sizeof ( int ) )];
+        unsigned char  cbuf[CMSG_SPACE ( sizeof ( tos_cmsg_type ) )];
         struct cmsghdr h;
     } u;
     static socklen_t clen = 0;
@@ -290,26 +296,30 @@ static ssize_t sendto_ipv4_with_tos ( int fd, const void* buf, size_t len, int f
     if ( clen == 0 )
     {
         // set up the cmsg buffer
-        memset ( u.cbuf, 0, sizeof ( u.cbuf ) );
+        memset ( &u, 0, sizeof ( u ) );
 
         u.h.cmsg_level = IPPROTO_IP;
         u.h.cmsg_type  = IP_TOS;
-        u.h.cmsg_len   = CMSG_LEN ( sizeof ( int ) );
-        memcpy ( CMSG_DATA ( &u.h ), &tos, sizeof ( int ) );
-        clen = (socklen_t) u.h.cmsg_len;
+        u.h.cmsg_len   = CMSG_LEN ( sizeof ( tos_cmsg_type ) );
+
+        tos_cmsg_type tosvalue = static_cast<tos_cmsg_type> ( tos & 0xFF );
+        memcpy ( CMSG_DATA ( &u.h ), &tosvalue, sizeof ( tosvalue ) );
+        clen = CMSG_SPACE ( sizeof ( tos_cmsg_type ) );
     }
 
     struct iovec iov;
+    memset ( &iov, 0, sizeof ( iov ) );
     iov.iov_base = const_cast<void*> ( buf );
     iov.iov_len  = len;
 
     struct msghdr msg;
+    memset ( &msg, 0, sizeof ( msg ) );
 
     msg.msg_name       = const_cast<sockaddr*> ( dest );
     msg.msg_namelen    = destlen;
     msg.msg_iov        = &iov;
     msg.msg_iovlen     = 1;
-    msg.msg_control    = (void*) u.cbuf;
+    msg.msg_control    = u.cbuf;
     msg.msg_controllen = clen;
 
     return sendmsg ( fd, &msg, flags );
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -129,7 +129,7 @@ void CSocket::Init ( const quint16 iNewPortNumber, const quint16 iNewQosNumber, 
         }
 #endif
 
-#if !defined( Q_OS_DARWIN ) && !defined( Q_OS_WIN )
+#if !defined( Q_OS_BSD4 ) && !defined( Q_OS_WIN )
         // set the QoS for IPv4 as well, as this is a dual-protocol socket
         if ( setsockopt ( UdpSocket, IPPROTO_IP, IP_TOS, (const char*) &tos, sizeof ( tos ) ) == -1 )
         {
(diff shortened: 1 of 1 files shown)


ann0see:
Ok. Will need to test this on macOS myself then.

src/socket.cpp
```diff
@@ -272,44 +272,55 @@ CSocket::~CSocket()
 #endif
 }
 
-#if defined( Q_OS_DARWIN )
-// sendto_ipv4_with_tos - helper function for macOS to set TOS when sending IPv4 over IPv6 socket
+#if defined( Q_OS_BSD4 )
```
ann0see, line 275: And this is guaranteed to always hold for macOS too?

softins: Yes, in [the header file](https://github.com/qt/qtbase/blob/dev/src/corelib/global/qsystemdetection.h), Q_OS_DARWIN and Q_OS_BSD4 are set on consecutive lines for Apple.

And my successful tests on Mac confirmed it.

src/socket.cpp
```diff
@@ -129,7 +129,7 @@ void CSocket::Init ( const quint16 iNewPortNumber, const quint16 iNewQosNumber,
         }
 #endif
 
-#if !defined( Q_OS_DARWIN ) && !defined( Q_OS_WIN )
+#if !defined( Q_OS_BSD4 ) && !defined( Q_OS_WIN )
```
pljones, line 132: I think we need a guidance note on where these are defined and how - specifically for Jamulus - they should be used.  (Not part of this PR, obviously.)

src/socket.cpp
```diff
@@ -272,44 +272,55 @@ CSocket::~CSocket()
...
-    static socklen_t clen = 0;
+    static size_t clen = 0;
 
     if ( clen == 0 )
     {
         // set up the cmsg buffer
-        memset ( u.cbuf, 0, sizeof ( u.cbuf ) );
+        memset ( &u, 0, sizeof ( u ) );
+
+        struct cmsghdr* cmsg = reinterpret_cast<struct cmsghdr*> ( u.cbuf );
```
pljones: OK, this is C pointer magic -- but isn't the union overlaying `struct cmsghdr align` and `unsigned char cbuf[..]` at the same address, so this could just be
```c++
struct cmsghdr* cmsg = &u.align;
```
or am I misunderstanding it completely?

softins: I think that's an excellent point. The magic was provided by ChatGPT when I asked it why the original function worked on macOS but not FreeBSD. I didn't study it in detail at the time, but I think you're right, and prefer your suggestion.

I'm happy to change it and test, but no time today.

src/socket.cpp
```diff
@@ -272,44 +272,55 @@ CSocket::~CSocket()
...
+    using tos_cmsg_type = int;
+#    endif
+
     // The cmsg buffer is only set up once (tos doesn't change) so can be static
     static union
     {
-        unsigned char  cbuf[CMSG_SPACE ( sizeof ( int ) )];
-        struct cmsghdr h;
+        unsigned char  cbuf[CMSG_SPACE ( sizeof ( tos_cmsg_type ) )];
+        struct cmsghdr align;
```
ann0see: (Why) is the naming change needed? here probably stands for header?

src/socket.cpp
```diff
@@ -272,44 +272,55 @@ CSocket::~CSocket()
...
     } u;
-    static socklen_t clen = 0;
+    static size_t clen = 0;
```
ann0see: size_t is more generic. Is this guaranteed to be sound?

src/socket.cpp
```diff
@@ -272,44 +272,55 @@ CSocket::~CSocket()
...
 
     if ( clen == 0 )
     {
         // set up the cmsg buffer
-        memset ( u.cbuf, 0, sizeof ( u.cbuf ) );
+        memset ( &u, 0, sizeof ( u ) );
```
ann0see, line 299: &u is just to make it more concise?

src/socket.cpp
```diff
@@ -272,44 +272,55 @@ CSocket::~CSocket()
...
+        cmsg->cmsg_type      = IP_TOS;
+        cmsg->cmsg_len       = CMSG_LEN ( sizeof ( tos_cmsg_type ) );
 
-        u.h.cmsg_level = IPPROTO_IP;
-        u.h.cmsg_type  = IP_TOS;
-        u.h.cmsg_len   = CMSG_LEN ( sizeof ( int ) );
-        memcpy ( CMSG_DATA ( &u.h ), &tos, sizeof ( int ) );
-        clen = (socklen_t) u.h.cmsg_len;
+        tos_cmsg_type v = static_cast<tos_cmsg_type> ( tos & 0xFF );
+        memcpy ( CMSG_DATA ( cmsg ), &v, sizeof ( v ) );
```
ann0see: v doesn't say much.  As far as I understand this, you extract a bit from tos? It should be made more clear what that is.

ann0see (approved):
Even though I'm not exactly sure about the code here.
It does work on macOS definitely. Still some questions.

src/socket.cpp
pljones: `cbuf` for "unsigned character buffer" is okay (better with a "u"?), as it implies "here's some unstructured data".

`h`, yes, a bit brief and `align` is... odd - it's "a cmsghdr struct aligned to that unsigned character buffer".  `msghdr` maybe?

src/socket.cpp
pljones: It's to make it clear it's the address of _both_ the cbuf and header.

src/socket.cpp
softins: Probably no reason. Name was chosen by ChatGPT, but h is good enough really.

src/socket.cpp
softins: ChatGPT recommended it for maximum portability, but I didn't test both with and without.

src/socket.cpp
pljones: I think it's throw-away -- it needs to be an instance variable (so `v` for "I need an instance variable of no other use"), though, otherwise you can't take the address of it to pass to memcpy.

softins: I'll revisit tomorrow, tidy up and test again.

src/socket.cpp
pljones: And lots and lots of networking code everywhere uses this approach.  (OK, it's very old networking code...)

ann0see: -> back to h

src/socket.cpp
ann0see: I'd rather not use size_t if something specific exists and it compiles.

softins: For the data type of `msg_controllen`, it looks like Linux uses `size_t`, but POSIX and BSD4 (and therefore FreeBSD and macOS) use `socklen_t`. So I've changed it back to `socklen_t`, as this code is not used on Linux.

In our case it doesn't really matter anyway, as the value is just assigned, not aliased.

src/socket.cpp
softins: It's to make sure the whole union gets cleared, whichever of the members is larger.

softins:
I've simplified the changes and re-tested on both FreeBSD and macOS. All working properly, so this is now ready.

src/socket.cpp
softins: Changed `v` to `tosvalue`

pljones (approved):

PR #3697, opened 2026-05-18
Introduce Asian directory server in code

### LINKED ISSUE #3692: Introduce new Asian directory server

The latency between Asia and Europe is high and it's wasteful to transport data over such a large geographic region. Since we have a considerably large Asian (Thai) community, there should be a Directory in Asia.
https://github.com/orgs/jamulussoftware/discussions/3629

@dtinth already proposed to set up a directory in Thailand. For Jamulus 4, we should introduce it.
Any Genre 3 is not used too much - thus, it likely makes sense to move this directory to Asia.
Any directory owner should be aware of https://github.com/orgs/jamulussoftware/discussions/3629#discussioncomment-16084970 and follow these rules.

Even though .io is [not likely to disappear, its future is unclear](https://www.eurodns.com/blog/end-of-io-domain), we should rather think of using jamulus.app for the new directory (e.g. asia.jamulus.app:22624 or asia.dir.jamulus.app:22624)

CC: @corrados @pljones 

Yes.

ann0see:

For geo-redundancy, latency and since Any Genre 3 is not heavily used, we introduce a directory in Asia. See
https://github.com/jamulussoftware/jamulus/issues/3692

CHANGELOG: Introduce Directory Server in Asia. This replaces Any Genre 3. Asian/Australian users should from now on use Any Genre Asia.

**Context: Fixes an issue?**

Fixes: #3692 

Ready for review

Review

## Checklist


ann0see:
Any Genre Asia shows up correctly in the client and server instead of Any Genre 3

ann0see:
@corrados We've now moved Any Genre 3 to Any Genre Asia hosted by @dtinth We believe it's safe to shut down Any Genre 3 directory on your end.

src/util.cpp
```diff
@@ -965,8 +965,8 @@ QString NetworkUtil::GetDirectoryAddress ( const EDirectoryType eDirectoryType,
         return strDirectoryAddress;
     case AT_ANY_GENRE2:
         return CENTSERV_ANY_GENRE2;
-    case AT_ANY_GENRE3:
-        return CENTSERV_ANY_GENRE3;
```
ann0see, line 969: We should probably rather not name it CENTSERV anymore...

pljones: Yeah, I was going to say - separate PR to move to `DIRECTORY_` and `.app` in one go?

pljones (approved):

src/util.cpp
ann0see: We could do this yes. But .app didn't resolve for me yet. @gilgongo

gilgongo: Ah, looking at the Cloudflare config-  I'd forgotten that it works using 301 redirects. For some reason I  thought it had some DNS wildcard capability. So no .app for directories right now I'm afraid.

gilgongo: > Yeah, I was going to say - separate PR to move to `DIRECTORY_` and `.app` in one go?

I suppose we could replicate the .io zone to .app and maintain both in parallel? We don't make changes that often after all. As far as I can tell, CloudFlare don't allow pointing two domains to the same zone file at least.

ann0see: Yes. .app should be the base and all .io records should cname to .app (minus MX records probably). That's the easiest config. Moreover I'd prefer not using cloudflare as proxy. GitHub is already DDOS etc protected I'd assume.

gilgongo: OK sure. I'll see if I can set up .app with the same host names as .io. Then when that's done we can CNAME all the the .io hosts to those. Might mean some slight downtime on the website cert while that gets re-generated perhaps (or some other hidden issue crops up). 

BTW we don't use the CF proxy for .io - in fact we found out we can't with GitHub Pages as it breaks their cert renewal. The proxy is only for .app at the moment so that it can do the http redirects I think.

softins: > Ah, looking at the Cloudflare config-  I'd forgotten that it works using 301 redirects. For some reason I  thought it had some DNS wildcard capability. So no .app for directories right now I'm afraid. 

301 redirects are a HTTP-only concept, and don't apply when just resolving names in DNS. We can do all we need using CNAMEs or parallel A records (and AAAA when we're ready).

gilgongo: OK so I can try this:

1. Export the .io zone file 
2. Change all .io references in the exported file to .app 
3. Remove any TXT records that have keys relating to .io (eg _github-challenge-jamulussoftware)
4. Delete all the current zone entries in .app
5. Import the new zone file into .app
6. Add TXT records with new keys for .app
7. Remove the www 301 redirect and switch Github Pages over to using .app

The worst that can happen is that the .app version of the website stops responding while it gets sorted out, but .io will be unaffected. Then once it's all working for .app, we change all the relevant CNAMEs in .io to .app.

ann0see: Hmm. Maybe we could rather wait for the main website? So just set all the directories as .app instead. 
This means duplicating only the cnames set for the .app zone similar to what we have in .io

Then wait a day. 
Then put the CNAMEs in .io to .app

Don't change anything on GitHub pages.

softins (approved):
Looks fine and runs as expected, for both Client and Server modes.

src/util.cpp
pljones: Yeah, I'd _just_ copy the Directory `.io` entries over as `.app` entries -- export _both_ zones, extract the Directories from `.io`, rename to `.app`, add to the `.app` zone file, reimport the `.app` zone file.  That should give two DNS entries for each target without any chance of anything breaking.

gilgongo: I assume we want to have the SRV records treated in the same way? So I'm going to add the following to the current .app zone (which right now only has dummy entries for the website so that the the 301 redirects can happen): https://pastebin.com/wT3s4dPm (maybe not for the updatecheck servers though?)

BTW lounge.jamulus.io was in there but it looks dead so I removed it. 

Then after that I'll re-name the directory CNAME and SRV targets in .io to point to .app.

ann0see: Get rid of anygenre3.jamulus.app

Otherwise I believe it looks good.

pljones:
```
_jamulus._udp.anygenre3.jamulus.app.	60	IN	SRV	0 0 22624 anygenre3.jamulus.app.
```
This could use the asia entry for backwards compatibility.

ann0see: I'd like to get rid of mentioning anygenre3 anywhere on jamulus.app but not on jamulus.io. 
anygenre3.jamulus.app never existed. One could also argue that asia.jamulus.io never existed (in a released version)

pljones: Ah, that's true.  So leave the SRV on jamulus.io, with the CNAME there both pointing to asia.jamulus.app and no other mention of asia.jamulus.app - and no mention of asia.jamulus.io.

gilgongo: OK I can delete SRV _jamulus._udp.asia on jamulus.io and the CNAME for asia.jamulus.io? No anygenre3 records on .app.

softins: > OK I can delete SRV _jamulus._udp.asia on jamulus.io and the CNAME for asia.jamulus.io? No anygenre3 records on .app.

Oh, that took me by surprise rather:
image

Changes like that shouldn't happen quite so fast - I don't spend _all_ my time at the computer, despite occasionally appearing to do so!

gilgongo: Sorry about that. I've not removed the SRV yet though.

pljones:
Backported to release/3_12.

PR #3699, opened 2026-05-20
Default reported OS to Unix if not Linux (e.g. FreeBSD)

softins:

Make default reported OS be Unix (`OT_UNIX` was already defined, so is not new here) if the build is not Windows, Mac, iOS, Android or Linux. For example FreeBSD.

CHANGELOG: Server: Report OS as Unix if not Windows, Mac, iOS, Android or Linux.

**Context: Fixes an issue?**

When recently testing a FreeBSD build, it was reporting its OS as Linux. It should say the more generic Unix.

Probably not. The OS is not displayed anywhere within Jamulus, so this only affects Jamulus Explorer and similar sites.

Ready and tested on both Linux and FreeBSD

Only review

## Checklist

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -1114,8 +1114,10 @@ public:
         return OT_I_OS;
 #elif defined( Q_OS_ANDROID ) || defined( ANDROID )
         return OT_ANDROID;
-#else
+#elif defined( Q_OS_LINUX )
         return OT_LINUX;
+#else
+        return OT_UNIX;
 #endif
     }
 };


ann0see (approved):
Looks sensible. Not tested but it looks trivial.

pljones (commented):
Note that this _introduces for the first time_ the possibility of `GetOperatingSystem()` returning `OT_UNIX`.

Is there anywhere this will cause problems (i.e. where the other enum values are handled and `OT_UNIX` has been overlooked)?

softins:
> Note that this _introduces for the first time_ the possibility of `GetOperatingSystem()` returning `OT_UNIX`.
> 
> Is there anywhere this will cause problems (i.e. where the other enum values are handled and `OT_UNIX` has been overlooked)?

No, the only places where the `OT_xxx` enum values are used are in `GetOperatingSystem()` and `GetOperatingSystemString()`. The value `OT_UNIX` is correctly accounted for in the latter, but that function is not called from anywhere in the codebase. Nothing else within Jamulus changes its behaviour depending on the OS value set here.

pljones:
> but that function is not called from anywhere in the codebase.

We have unused code?  Why not remove it whilst we're here?

softins:
> > but that function is not called from anywhere in the codebase.
> 
> We have unused code? Why not remove it whilst we're here?

Scope creep? This PR was meant to be a trivial change with a single purpose.

Maybe Volker envisaged a possible need to output the OS protocol value as a string sometime, in a UI or a console message?

I'm sure there are quite a few other unused sections of code scattered about the code base.

pljones (approved):

pljones:
Backported to release/3_12.

PR #3702, opened 2026-05-20
Potential fixes for 4 code quality findings

ann0see:
_This PR applies 4/4 suggestions from code quality [AI findings](https://github.com/jamulussoftware/jamulus/security/quality/ai-findings)._

Only for review. We should then decide if we actually have a memory leak here.

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -907,7 +907,14 @@ void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
 {
     if ( pSettings->bEnableAudioAlerts && iNewNumClients > iClients )
     {
-        QSoundEffect* sf = new QSoundEffect();
+        QSoundEffect* sf = new QSoundEffect ( this );
+        connect ( sf, &QSoundEffect::playingChanged, this, [sf]()
+        {
+            if ( !sf->isPlaying() )
+            {
+                sf->deleteLater();
+            }
+        } );
         sf->setSource ( QUrl::fromLocalFile ( ":sounds/res/sounds/new_user.wav" ) );
         sf->play();
     }
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -1445,8 +1445,8 @@ void CClientDlg::SetMeterStyle ( const EMeterStyle eNewMeterStyle )
         break;
 
     case MT_LED_ROUND_SMALL:
-        lbrInputLevelL->SetLevelMeterType ( CLevelMeter::MT_LED_ROUND_BIG );
-        lbrInputLevelR->SetLevelMeterType ( CLevelMeter::MT_LED_ROUND_BIG );
+        lbrInputLevelL->SetLevelMeterType ( CLevelMeter::MT_LED_ROUND_SMALL );
+        lbrInputLevelR->SetLevelMeterType ( CLevelMeter::MT_LED_ROUND_SMALL );
         break;
     }
 
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -856,9 +856,9 @@ void CClientDlg::OnChatTextReceived ( QString strChatText )
 {
     if ( pSettings->bEnableAudioAlerts )
     {
-        QSoundEffect* sf = new QSoundEffect();
-        sf->setSource ( QUrl::fromLocalFile ( ":sounds/res/sounds/new_message.wav" ) );
-        sf->play();
+        QSoundEffect sf;
+        sf.setSource ( QUrl::fromLocalFile ( ":sounds/res/sounds/new_message.wav" ) );
+        sf.play();
     }
     ChatDlg.AddChatText ( strChatText );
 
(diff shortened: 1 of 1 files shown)


ann0see:
I don't think everything here is valid.

src/clientdlg.cpp
```diff
@@ -856,9 +856,9 @@ void CClientDlg::OnChatTextReceived ( QString strChatText )
 {
     if ( pSettings->bEnableAudioAlerts )
     {
-        QSoundEffect* sf = new QSoundEffect();
-        sf->setSource ( QUrl::fromLocalFile ( ":sounds/res/sounds/new_message.wav" ) );
-        sf->play();
+        QSoundEffect sf;
+        sf.setSource ( QUrl::fromLocalFile ( ":sounds/res/sounds/new_message.wav" ) );
+        sf.play();
```
ann0see, line 861: This may be valid. Stack vs heap allocation?

src/clientdlg.cpp
```diff
@@ -1440,13 +1447,13 @@ void CClientDlg::SetMeterStyle ( const EMeterStyle eNewMeterStyle )
...
+        lbrInputLevelR->SetLevelMeterType ( CLevelMeter::MT_BAR_NARROW );
         break;
 
     case MT_LED_ROUND_SMALL:
-        lbrInputLevelL->SetLevelMeterType ( CLevelMeter::MT_LED_ROUND_BIG );
-        lbrInputLevelR->SetLevelMeterType ( CLevelMeter::MT_LED_ROUND_BIG );
+        lbrInputLevelL->SetLevelMeterType ( CLevelMeter::MT_LED_ROUND_SMALL );
+        lbrInputLevelR->SetLevelMeterType ( CLevelMeter::MT_LED_ROUND_SMALL );
         break;
     }
```
ann0see, line 1458: Unsure about those.

src/clientdlg.cpp
ann0see: -> memory leak

src/clientdlg.cpp
ann0see: @dcorson-ticino-com ?

src/clientdlg.cpp
softins: No this won't work. The original does indeed have a memory leak, as the pointer returned from `new QSoundEffect()` is only local, and disappears at the end of the block. But I think changing it so that the `QSoundEffect sf` is allocated locally on the stack may cause a crash, or at least truncate the playback, as the object will be destroyed immediately at the end of the block while the sound is still playing.

The correct solution is actually the one shown below around line 910: connect a lambda to the `playingChanged` signal and call `deleteLater()` when it is finished with.

src/clientdlg.cpp
```diff
@@ -907,7 +907,14 @@ void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
...
     {
-        QSoundEffect* sf = new QSoundEffect();
+        QSoundEffect* sf = new QSoundEffect ( this );
+        connect ( sf, &QSoundEffect::playingChanged, this, [sf]()
+        {
+            if ( !sf->isPlaying() )
+            {
+                sf->deleteLater();
+            }
+        } );
```
softins, line 917: This looks ok and is the correct way to avoid the memory leak.

src/clientdlg.cpp
softins: See 3117acde7bdbbd4cc6eebc9619214a79737b54f5 in #2356 from @henkdegroot 

I think the point is that when narrow/small is selected, it should only apply to the mixer channel meters. The above lines refer to the input meters, which need to remain wide with big LEDs, even in narrow mixer mode.

It would be worth adding explanatory comments to that effect, to forestall future confusion.

softins (changes requested):

src/clientdlg.cpp
softins: I suggest the following for the whole function:

```
void CClientDlg::SetMeterStyle ( const EMeterStyle eNewMeterStyle )
{
    // apply MeterStyle to current window
    // Note: input meter uses big LED and wide bar even in narrow mode
    switch ( eNewMeterStyle )
    {
    case MT_LED_STRIPE:
        lbrInputLevelL->SetLevelMeterType ( CLevelMeter::MT_LED_STRIPE );
        lbrInputLevelR->SetLevelMeterType ( CLevelMeter::MT_LED_STRIPE );
        break;

    case MT_LED_ROUND_BIG:
    case MT_LED_ROUND_SMALL:
        lbrInputLevelL->SetLevelMeterType ( CLevelMeter::MT_LED_ROUND_BIG );
        lbrInputLevelR->SetLevelMeterType ( CLevelMeter::MT_LED_ROUND_BIG );
        break;

    case MT_BAR_WIDE:
    case MT_BAR_NARROW:
        lbrInputLevelL->SetLevelMeterType ( CLevelMeter::MT_BAR_WIDE );
        lbrInputLevelR->SetLevelMeterType ( CLevelMeter::MT_BAR_WIDE );
        break;

    }

    // also apply MeterStyle to child GUI controls
    MainMixerBoard->SetMeterStyle ( eNewMeterStyle );
}
```

ann0see: -> https://github.com/jamulussoftware/jamulus/pull/3718

Issue #3662, opened 2026-04-26
Refactor CServer::MixEncodeTransmitData for performance


ann0see:
To me it seems as if CServer::MixEncodeTransmitData() has a lot of branches, is messy and likely worth optimizing. For example , we could likely also use a pattern like
```
                        vecfIntermProcBuf[k] += vecsData[i] * fGainL;
                        vecfIntermProcBuf[k + 1] += vecsData[i] * fGainR;
```

instead of 

```
                        if ( ( i & 1 ) == 0 )
                        {
                            // if even : left channel
                            vecfIntermProcBuf[i] += vecsData[i] * fGainL;
                        }
                        else
                        {
                            // if odd  : right channel
                            vecfIntermProcBuf[i] += vecsData[i] * fGainR;
                        }
```

https://github.com/jamulussoftware/jamulus/blob/3827902c594d54ccc5e3e8a73ff345877ca72a57/src/server.cpp#L914

As this seems to be the probably most important function in Jamulus, this warrants a close investigation

ann0see:
CC @pljones @softins

ann0see:
``` 
            // if channel gain is 1, avoid multiplication for speed optimization
            if ( fGain == 1.0f )
            {

```

Questionable if this really gives speedup...

ann0see:
```
                        vecfIntermProcBuf[i] += ( static_cast<float> ( vecsData[k] ) + vecsData[k + 1] ) / 2;
```

the /2 can definitely be made * 0.5 even if the compiler does this itself

ann0see:
```
                    const int iOffset = iB * SYSTEM_FRAME_SIZE_SAMPLES * vecNumAudioChannels[iChanCnt];
```

Probably just iOffset+=iB is enough with respective refactoring.

pljones:
Anything else needed to close this that won't make it into 4.0.0?  If not, could you change the milestone and close.  Ta.

ann0see:
I've checked the assembly and there's basically no vectorization for the code. Also I still believe that the function is insanely messy and should be refactored

dingodoppelt:
> I've checked the assembly and there's basically no vectorization for the code. Also I still believe that the function is insanely messy and should be refactored

Yes, I've seen that, too. I think we should first decide if the current audio code is valid and even does what it was intended to do and maybe rewrite parts of it. I think mixing float and short isn't helpful at all and might even introduce unwanted noise in some cases.  
[...]

PR #3706, opened 2026-05-20
Optimize bDelayPan and divisions

ann0see:

Moves delay pan checking slightly out of loop as this setting does not change often.
Moreover replaces some divisions with \/2.0f and remove some modulo computations

CHANGELOG: Server: Optimize mixing

**Context: Fixes an issue?**

Related to: https://github.com/jamulussoftware/jamulus/issues/3662

No

Ready for review

Tested and it seems to work, I still would like an external test.

## Checklist

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -982,7 +982,7 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt, const int iNumClients 
                     // stereo: apply stereo-to-mono attenuation
                     for ( i = 0, k = 0; i < iServerFrameSizeSamples; i++, k += 2 )
                     {
-                        vecfIntermProcBuf[i] += ( static_cast<float> ( vecsData[k] ) + vecsData[k + 1] ) / 2;
+                        vecfIntermProcBuf[i] += ( static_cast<float> ( vecsData[k] ) + vecsData[k + 1] ) / 2.0f;
                     }
                 }
             }
@@ -1001,7 +1001,7 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt, const int iNumClients 
                     // stereo: apply stereo-to-mono attenuation
                     for ( i = 0, k = 0; i < iServerFrameSizeSamples; i++, k += 2 )
                     {
-                        vecfIntermProcBuf[i] += fGain * ( static_cast<float> ( vecsData[k] ) + vecsData[k + 1] ) / 2;
+                        vecfIntermProcBuf[i] += fGain * ( static_cast<float> ( vecsData[k] ) + vecsData[k + 1] ) / 2.0f;
                     }
                 }
             }
@@ -1020,7 +1020,7 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt, const int iNumClients 
         const int maxPanDelay = MAX_DELAY_PANNING_SAMPLES;
 
         int iPanDelL = 0, iPanDelR = 0, iPanDel;
-        int iLpan, iRpan, iPan;
+        int iLpan, iRpan;
 
         for ( j = 0; j < iNumClients; j++ )
         {
@@ -1036,21 +1036,20 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt, const int iNumClients 
             const float fGainL = MathUtils::GetLeftPan ( fPan, false ) * fGain;
             const float fGainR = MathUtils::GetRightPan ( fPan, false ) * fGain;
 
+            const bool isMono = vecNumAudioChannels[j] == 1;
+
             if ( bDelayPan )
             {
                 iPanDel  = lround ( (float) ( 2 * maxPanDelay - 2 ) * ( vecvecfPannings[iChanCnt][j] - 0.5f ) );
                 iPanDelL = ( iPanDel > 0 ) ? iPanDel : 0;
                 iPanDelR = ( iPanDel < 0 ) ? -iPanDel : 0;
-            }
 
-            if ( vecNumAudioChannels[j] == 1 )
-            {
-                // mono: copy same mono data in both out stereo audio channels
-                for ( i = 0, k = 0; i < iServerFrameSizeSamples; i++, k += 2 )
+                if ( isMono )
                 {
-                    // left/right channel
-                    if ( bDelayPan )
+                    // mono: copy same mono data in both out stereo audio channels
+                    for ( i = 0, k = 0; i < iServerFrameSizeSamples; i++, k += 2 )
                     {
+                        // left/right channel
                         // pan address shift
 
                         // left channel
@@ -1079,54 +1078,60 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt, const int iNumClients 
                             vecfIntermProcBuf[k + 1] += vecsData[iRpan] * fGainR;
                         }
                     }
-                    else
+                }
+                else
+                {
+                    // stereo
+                    for ( i = 0; i < ( 2 * iServerFrameSizeSamples ); i += 2 )
                     {
-                        vecfIntermProcBuf[k] += vecsData[i] * fGainL;
-                        vecfIntermProcBuf[k + 1] += vecsData[i] * fGainR;
+                        // pan address shift
+
+                        iLpan = i - 2 * iPanDelL;         // left channel
+                        iRpan = ( i + 1 ) - 2 * iPanDelR; // right channel
+
+                        // interleaved channels
+                        if ( iLpan < 0 )
+                        {
+                            // get from second
+                            iLpan = iLpan + 2 * iServerFrameSizeSamples;
+                            vecfIntermProcBuf[i] += vecsData2[iLpan] * fGain;
+                        }
+                        else
+                        {
+                            vecfIntermProcBuf[i] += vecsData[iLpan] * fGain;
+                        }
+
+                        if ( iRpan < 0 )
+                        {
+                            // get from second
+                            iRpan = iRpan + 2 * iServerFrameSizeSamples;
+                            vecfIntermProcBuf[i + 1] += vecsData2[iRpan] * fGain;
+                        }
+                        else
+                        {
+                            vecfIntermProcBuf[i + 1] += vecsData[iRpan] * fGain;
+                        }
                     }
                 }
             }
             else
             {
-                // stereo
-                for ( i = 0; i < ( 2 * iServerFrameSizeSamples ); i++ )
+                if ( isMono )
                 {
-                    // left/right channel
-                    if ( bDelayPan )
+                    // mono: copy same mono data in both out stereo audio channels
+                    for ( i = 0, k = 0; i < iServerFrameSizeSamples; i++, k += 2 )
                     {
-                        // pan address shift
-                        if ( ( i & 1 ) == 0 )
-                        {
-                            iPan = i - 2 * iPanDelL; // if even : left channel
-                        }
-                        else
-                        {
-                            iPan = i - 2 * iPanDelR; // if odd  : right channel
-                        }
-                        // interleaved channels
-                        if ( iPan < 0 )
-                        {
-                            // get from second
-                            iPan = iPan + 2 * iServerFrameSizeSamples;
-                            vecfIntermProcBuf[i] += vecsData2[iPan] * fGain;
-                        }
-                        else
-                        {
-                            vecfIntermProcBuf[i] += vecsData[iPan] * fGain;
-                        }
+                        vecfIntermProcBuf[k] += vecsData[i] * fGainL;
+                        vecfIntermProcBuf[k + 1] += vecsData[i] * fGainR;
                     }
-                    else
+                }
+                else
+                {
+                    for ( i = 0; i < ( 2 * iServerFrameSizeSamples ); i += 2 )
                     {
-                        if ( ( i & 1 ) == 0 )
-                        {
-                            // if even : left channel
-                            vecfIntermProcBuf[i] += vecsData[i] * fGainL;
-                        }
-                        else
-                        {
-                            // if odd  : right channel
-                            vecfIntermProcBuf[i] += vecsData[i] * fGainR;
-                        }
+                        // left/right channel
+                        vecfIntermProcBuf[i] += vecsData[i] * fGainL;
+                        vecfIntermProcBuf[i + 1] += vecsData[i + 1] * fGainR;
                     }
                 }
             }


src/server.cpp
```diff
@@ -1079,54 +1078,61 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt, const int iNumClients
...
-                        if ( ( i & 1 ) == 0 )
-                        {
-                            // if even : left channel
-                            vecfIntermProcBuf[i] += vecsData[i] * fGainL;
-                        }
-                        else
-                        {
-                            // if odd  : right channel
-                            vecfIntermProcBuf[i] += vecsData[i] * fGainR;
-                        }
```
ann0see, line 1129: Got rid of modulo computation

ann0see (commented):
The function is still way too complex imo.

ann0see:
To avoid having the clipping to short twice, we may want someting like

```
    // convert from double to short with clipping
    for ( i = 0; i < ( ( bTargetIsMono ? 1 : 2 ) * iServerFrameSizeSamples ); i++ )
    {
        vecsSendData[i] = Float2Short ( vecfIntermProcBuf[i] );
    }
 ```
Maybe...

File: src/server.cpp
Comment (pljones) [inline src/server.cpp:1034]: This could do with expanding.

I thought the Server up-mixed all incoming channels to stereo (so it could treat them the same for all connected clients), did all the mixes in stereo - to allow a single path (either delay pan or not) - and then down-mixed to mono for clients with mono out.

The changes here seem to do something different but I don't quite follow it.

Comment (ann0see) [reply src/server.cpp:1034]: I think there's a mono only code path?

Comment (pljones) [reply src/server.cpp:1034]: There is in this new flow.  What I mean is I don't understand
a) exactly what _was_ happening before this new mono flow was introduced
b) exactly what _is_ happening now
c) whether it's therefore having the right effect

So it needs a bigger comment in the code to explain it.

Well... it needs a good answer 😄 and then a good enough comment to explain the new code.

Comment (ann0see) [reply src/server.cpp:1034]: New flow:
```
if delay pan:

// code for delay pan in mono or stereo mode
// not sure why there is even delay panning in mono mode but ok.

else:

// mono or stereo mixing
```

old flow:
```
if mono: 
  if delay pan:
     // delay pan mixing
  else:
    // non delay pan mixing
else:
  if delay pan:
    // parity based computation for delay pan
  else:
  // parity based mixing
```

Please check the code itself, not the diff as this is confusing

Comment (pljones) [reply src/server.cpp:1034]: > // not sure why there is even delay panning in mono mode but ok.

A mono client can pan any client in their mix and it will affect the output.  ("I want more of their left channel in my mix down")

A stereo client can pan any client in their mix and it will affect the output.  ("I want them moved left in my mix")

Comment (pljones) [reply src/server.cpp:1034]: > Please check the code itself, not the diff as this is confusing

I can see it moves the two `if`s inside out.  But I can't see where the panning is applied on the source channel for a mono target now.

Comment (ann0see) [reply src/server.cpp:1034]: Not sure I understand correctly. I didn't change the semantics I believe? 
I believe that a mono target may not have panning?

Comment (pljones) [reply src/server.cpp:1034]: > A mono client can pan any client in their mix and it will affect the output.

If the source is stereo, it has two channels.  The mono target can, therefore, adjust the mix - either with delaypan or not - that gets then mixed down to mono.

I think - but I'm not sure - the change here inverts the logic and prevents that.

Comment (ann0see) [reply src/server.cpp:1034]: Ok. Let's disect the logic.

vecNumAudioChannels[iChanCnt] == 1  => mono target channel. I only replaced /2 by *0.5f which is equivalent

vecNumAudioChannels[iChanCnt] != 1 => stereo target channel

Let j be arbitrary between 0 and iNumClients -1

case bDelayPan == false and vecNumAudioChannels[j] == 1 ( isMono == true):

* Old version skips setting iPanDel, iPanL, iPanR statement
* new version skips entire if (bDelayPan) section moving to the else block.

* Since cNumAudioChannels[j] == 1 , old version enters mono: "copy same mono data" ... loop.
* Since isMono == true, new version enters "copy same mono data" ... loop
* since bDelayPan == false, old version skips // pan address shift and executes `                    vecfIntermProcBuf[k] += vecsData[i] * fGainL [...]`
* since bDelayPan == false, and the entire outer conditional is already skipped before, new version checks (isMono) and executes `vecfIntermProcBuf[k] += vecsData[i] * fGainL`

=> looks valid

case bDelayPan == false and vecNumAudioChannels[j] != 1 ( isMono == false):

* old version enters `//stereo`.
* new version enters else and checks isMono -> ` // left/right channel vecfIntermProcBuf[i] += vecsData[i] * fGainL;```

* old version checks bDelayPan == false, and thus checks ` if ( ( i & 1 ) == 0 ) [...]` -> optimised in the new version

This also looks valid?


case bDelayPan == true and vecNumAudioChannels[j] == 1 ( isMono == true):

* old version enters `// mono: copy same[]...` 
* new version enters `iPanDel[...]`

* old version enters `// pan address shift [...] ilpan = I- iPanDelL...`
* new version enters same

==> also looks valid

case bDelayPan == true and vecNumAudioChannels[j] != 1 ( isMono == false)

New: 

```
// stereo
                    for ( i = 0; i < ( 2 * iServerFrameSizeSamples ); i += 2 )
                    {
                        // pan address shift

                        iLpan = i - 2 * iPanDelL;  
```

Old (unoptimised)
```
                for ( i = 0; i < ( 2 * iServerFrameSizeSamples ); i++ )
                {
                    // left/right channel
                    if ( bDelayPan )
                    {
                        // pan address shift
                        if ( ( i & 1 ) == 0 )
                        {
                            iPan = i - 2 * iPanDelL; // if even : left channel
                        }
```


Also looks valid.

Could you please specify the variable state which you think is invalid?

Comment (pljones) [reply src/server.cpp:1034]: > Since cNumAudioChannels[j] == 1 , old version enters mono: "copy same mono data" ... loop.

Uh?  How did you get that?  `vecNumAudioChannels[j]` is not equivalent to `cNumAudioChannels[j]` so I don't see where that came from.

This is what I mean -- I _think_ `cNumAudioChannels[j]` is where the sound comes from and `vecNumAudioChannels[j]` is where the sound goes to.  You look at `vecNumAudioChannels[j]` to determine `isMono` and then assume that means every connected client is mono for that output.

File: src/server.cpp [hunk 2]
```diff
@@ -982,7 +982,7 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt, const int iNumClients
                     // stereo: apply stereo-to-mono attenuation
                     for ( i = 0, k = 0; i < iServerFrameSizeSamples; i++, k += 2 )
                     {
-                        vecfIntermProcBuf[i] += ( static_cast<float> ( vecsData[k] ) + vecsData[k + 1] ) / 2;
+                        vecfIntermProcBuf[i] += ( static_cast<float> ( vecsData[k] ) + vecsData[k + 1] ) * 0.5f;
```
dingodoppelt: Couldn't we just right shift the short to halve it?

src/server.cpp
```diff
@@ -1001,7 +1001,7 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt, const int iNumClients
                     // stereo: apply stereo-to-mono attenuation
                     for ( i = 0, k = 0; i < iServerFrameSizeSamples; i++, k += 2 )
                     {
-                        vecfIntermProcBuf[i] += fGain * ( static_cast<float> ( vecsData[k] ) + vecsData[k + 1] ) / 2;
+                        vecfIntermProcBuf[i] += fGain * ( static_cast<float> ( vecsData[k] ) + vecsData[k + 1] ) * 0.5f;
```
dingodoppelt: I don't know if we do this right. Shouldn't we normalize the floats to `-1 .. 1` to get most out of the float precision? And even if we did we wouldn't mitigate clipping because adding many channels might still clip while we don't scale back or check, if clipping occurred. When converting back to short we don't care for the actual range, we just hard clip everything back into the short range. That doesn't seem right at all to me.

pljones: I thought the way the signal summation work was progressive:
* one client at 1.0f -> 1.0f / 1.0f -> 1.0f
* two clients at 1.0f -> ( 1.0f + 1.0f ) / 2.0f -> 1.0f

etc.  Clipping _should_ never happen.  However, that's ignoring panning, of course.  If you multiple L 1.0f by 1.1f and divide R 1.0f by 1.1f, you will get clipping.  I guess - as panning is on the client GUI as is level - this is seen as something the client user has under their control: drop the channel level to allow for panning.

As with all the audio code -- I could easily be wrong.  I can't follow it.

src/server.cpp
softins: Not with a `float`. Doing so with an `int` before conversion would reduce the resolution by 1 bit.

Doing `* 0.5f` or `/ 2.0f` with a float preserves the full resolution in the mantissa and just changes the exponent. I wouldn't be surprised if the compiler produces exactly the same code in both cases.

src/server.cpp
softins: > I don't know if we do this right. Shouldn't we normalize the floats to `-1 .. 1` to get most out of the float precision?

Floats are internally normalised anyway, having the max available precision in the mantissa, and then scaling with the exponent. This all happens internally within the floating-point engine.

softins:
I will study this code soon and make any comments...

src/server.cpp
pljones: `/ 2.0f` feels more intuitive, somehow.  If there's no execution speed difference, I'd rather keep it.

src/server.cpp
ann0see: I believe this was a typo and should be vecNumAudioChannels[j] instead.

https://github.com/ann0see/jamulus/blob/d6372fb358c2f8e2e038a67e28aeacb17f2291a8/src/server.cpp#L1039-L1040

should be equivalent to 

https://github.com/ann0see/jamulus/blob/2ed93d28177da93ff1b416dc39bdee71ed5bfa06/src/server.cpp#L1046

src/server.cpp
ann0see: It's unlikely to be different - however this may depend on the compiler. 0.5f is easier for the compiler to optimise than / 2, I'd argue.

src/server.cpp
ann0see: > As with all the audio code -- I could easily be wrong. I can't follow it.

Which proves that it must be refactored...

dingodoppelt: > > I don't know if we do this right. Shouldn't we normalize the floats to `-1 .. 1` to get most out of the float precision?
> 
> Floats are internally normalised anyway, having the max available precision in the mantissa, and then scaling with the exponent. This all happens internally within the floating-point engine.

I think normalization in the float representation only means to have only one non-zero digit in front of the mantissa and therefore scaling the exponent. I think we still miss out on precision because we convert big numbers to float while float has the most precision up to values of `2` so normalizing in an audio DSP sense means scaling to `-1 .. 1`.  

> I thought the way the signal summation work was progressive:

    one client at 1.0f -> 1.0f / 1.0f -> 1.0f
    two clients at 1.0f -> ( 1.0f + 1.0f ) / 2.0f -> 1.0f

> etc. Clipping should never happen.  

I couldn't find anything which takes the number of clients into account, i.e. there is no scaling happening. We just keep adding and adding...
 I think we actually do clip all the time. I added a `qDebug()` to the Float2Short function and, oh boy is this thing clipping with only two clients (one playback, one drums)

ann0see: Which may explain why people complain about quality. But why don't we hear clipping directly?

dingodoppelt: > Which may explain why people complain about quality. But why don't we hear clipping directly?

Maybe it is not "enough" to be audible. Boost your input gain in the settings and you'll hear some definite clipping artefacts.  
What we hear as the degradation in sound some people reported, are maybe just artificial overtones introduced be the square wave that is being created in the Float2Short function. -> Distortion  
  
Edit: We hear some very unpleasant distortion because we don't saturate/soft-clip but hard clip instead. For the time being a mitigation could be to apply `tanh()` instead of `Float2Short` but that would increase load on the server immensely I suppose

src/server.cpp
dingodoppelt: > Not with a `float`. Doing so with an `int` before conversion would reduce the resolution by 1 bit.
> 
> Doing `* 0.5f` or `/ 2.0f` with a float preserves the full resolution in the mantissa and just changes the exponent. I wouldn't be surprised if the compiler produces exactly the same code in both cases.

But since we hard clip to back to `short` the value will be the same in the end. If we did `(short +1) >> 1` we'd get the same rounded value back as if we did `Float2Short( static_cast<float>(short) / 2.f )` or am I wrong here?

src/server.cpp
dingodoppelt: I did some test with a few clients sending the same sine wave. At some point you clearly hear the sine wave turn into a square wave and my qDebug() output from the `Float2Short` function is permanent.

softins: > I couldn't find anything which takes the number of clients into account, i.e. there is no scaling happening. We just keep adding and adding...

I think that's correct. The output is the sum of the inputs, not their average, and statistically, channels will hit waveform peaks at different times. (unless you are sending in synchronised sine waves!)

Consider, say, four channels, one of which has sound, and three of which are currently silent. Ignoring any faders, when mixing, the output should be a 100% copy of the input channel that has sound, not just 25% of it. Otherwise adding more and more channels would make each one quieter and quieter.

softins: Further info and discussion:

https://stackoverflow.com/questions/376036/algorithm-to-mix-sound

https://dsp.stackexchange.com/questions/3581/algorithms-to-mix-audio-signals-without-clipping

dingodoppelt: I had the client clipping under normal (musical) circumstances with only two clients.  
Applying something similar to the jitter buffer logic, which measures errors over time and adjusts accordingly, could be applied to a sliding RMS which tracks clipping situations and above a certain threshold scales the input samples back.  
Users could mitigate the problem by moving down all faders or having a dedicated fader controlling the global gain alongside something indicating a clipping in the mix.  

> Consider, say, four channels, one of which has sound, and three of which are currently silent. Ignoring any faders, when mixing, the output should be a 100% copy of the input channel that has sound, not just 25% of it. Otherwise adding more and more channels would make each one quieter and quieter.  

An automatic gain correction could look like this: Only if too much clipping in a defined time period is being measured we scale every client input by a set amount before mixing. This would result in audible (smoothed) shifts in volume so these shouldn't happen too frequently.  
This `fServerGain` could be a fixed value or dynamic via RPC to start with. An external tool could be used for automated gain correction in that case or could be set by an admin if necessary.

ann0see: But this definitely seems like a hack to me - and it adds more compute on the server. It should be a solved problem. We should try to pin down how much of a problem this really is.

dingodoppelt: > We should try to pin down how much of a problem this really is.

I think we could try to get rid of the float conversion altogether since we do it but don't benefit from it. Gain scaling could be done by adding and right shifting instead of multiplication. I looked through the code and already tried some things but I don't know if that really improves performance since we still rely on a few floats (gain/pan) so multiplication or casting is inevitable at the moment.

dingodoppelt: I haven't touched the delay pan path, yet, but this is just a quick and dirty test for mixing on an integer rather than float:  
[https://github.com/jamulussoftware/jamulus/compare/main...dingodoppelt:jamulus:clipping](https://github.com/jamulussoftware/jamulus/compare/main...dingodoppelt:jamulus:clipping)  
It would make life easier if the client was sending integers for gain and pan in the first place but you get the idea. I don't know how to measure the impact this has on server performance, though.

ann0see: Intel performance counters (not sure what it's called exactly) can count cycles.

dingodoppelt: I couldn't see any improvements with my changes. I guess the compiler already knows best.

ann0see: So then probably we could skip the 0.5f optimization.

src/server.cpp
ann0see: @pljones What do you think needs to be done that we know it's correct? I believe it is but I might miss something

softins (approved):
OK, to be certain, I made separate copies of the `MixEncodeTransmitData` function, two from `main` and two from this PR. I then edited one of each to remove the `if (bDelayPan)` sections and the other one of each to remove the corresponding `else` sections instead.

I was then able to compare the "before" and "after" with no delaypan, and the "before" and "after" with delaypan, ignoring whitespace, to be sure the logic was still exactly the same in each case. I am satisfied that it is.

So this PR preserves exactly the existing behaviour of Jamulus, appears to be potentially cleaner and more efficient, and I'm approving it for merging as-is on this basis.

I think any proposal to modify and potentially improve the design of this function should be a separate PR, discussed and thoroughly tested on its own merits.

src/server.cpp
dingodoppelt: From my tests I couldn't find any speed improvements with shifts instead of multiplication

pljones (approved):
OK, you're right, comparing the full files makes it easier.

It's the mono/stereo and delayPan `if` statements that are swapped so you do all of `delayPan` (with mono and stereo versions) and then all of `!delayPan` (with mono and stereo versions), which does make more sense.

I still prefer the `/ 2` (rather, `/ 2f` - having cast to float and targeting a vector of floats) to `* 0.5f` for readability.  "Half of" is an easier concept than "50% of".

src/server.cpp
pljones: I think the problem is actually `<float_vector>[index] = <some float> / <integer>` - it's turning the RHS from float to int then back to float - because of the `/ 2` not being `/ 2f`.

ann0see: Ok. I would not expect that but probably worth changing it to /2f then?

softins (approved):

src/server.cpp
dingodoppelt, line 1032: We should have the hard coded value defined with a more readable name like `PAN_CENTER`

dingodoppelt, line 970: See below. Maybe this could be `UNITY_GAIN` ?

softins: I would like to know if this test really does gain us anything. It is just there to eliminate a `* fGain` when `fGain` is exactly `1.0f`, but I wonder what it really saves, and how often this really does match? Typically, tests for equality with floats seldom match.

I would be inclined to remove the `if ( fGain == 1.0f )` section - but that would be for a different PR after some tests.

softins: Sometimes I find replacing a value with a name that is defined elsewhere just means I have to look at that other place just to understand what the value is. Just a comment might be simpler and avoid the visual redirection.

dingodoppelt (approved):

src/server.cpp
ann0see: Agree from the theoretical perspective. 

But this may just be the default gain and still occur often.

PR #3720, opened 2026-05-30
Auto-enable IPv6 by default, unless disabled

### LINKED ISSUE #3300: Enable ipv6 client side support by default
**Current behaviour and why should it be changed?**

IPv6 client support currently has to be enabled at the command line using the -6 option. This requires the user either to edit the shortcuts/links or start the client from the command line. This varies in complexity depending on the OS, but it is a major barrier to IPv6 adoption when trying to support large groups having to make the required changes to client startup. Having IPv6 support enabled by default would remove this barrier.

Another approach could be to enable IPv6 support using a checkbox within the UI. This would allow the user to elect to enable IPv6, and perhaps there is a very marginal resource saving in not starting a feature that may not be used. 

Enable ipv6 client side support by default with no additional checkbox in the UI. That solution means no action needs to be taken at the client side at all for IPv6.

The topic has been discussed here: [https://github.com/orgs/jamulussoftware/discussions/3291](https://github.com/orgs/jamulussoftware/discussions/3291)

softins:

Auto-detects if the host supports IPv6, and enables it by default if so, unless explicitly disabled.

Removes `-6` and `--enableipv6`, replacing them with `--noipv6` to disable. If `--noipv6` is not specified, it will try first to open an IPv6 socket, and if the host does not support it, will fall back to an IPv4 socket.

`CSocket` informs `CClient` or `CServer` of IPv6 availability via a passed-in reference.

The bool `bEnableIPv6` is gone, and the actual availability of IPv6 is published by a method in `CClient` or `CServer`.

CHANGELOG: Client/Server: Enable IPv6 by default if supported by host. Provide option to disable manually.

**Context: Fixes an issue?**

Fixes #3300

The `--noipv6` option should be documented, along with an explanation that IPv6 is always enabled by default on supporting hosts.

Tested locally on Linux and ready for review.

Just review and testing on multiple platforms.

## Checklist


pljones:
If a server operator uses `--serverbindip 192.168.1.123`, does this achieve the same effect?

Does `--serverbindip` support IPv6 addresses?  Presumably there's no option of "fall back" if that works.

softins:
> If a server operator uses `--serverbindip 192.168.1.123`, does this achieve the same effect?
> 
> Does `--serverbindip` support IPv6 addresses? Presumably there's no option of "fall back" if that works.

At the moment, and ever since we introduced basic IPv6 support five years ago (!), `--serverbindip` is ignored completely if the socket is `AF_INET6` (dual-stack). It is only honoured if the socket is plain `AF_INET`. That means in existing versions that `-6` was not supplied, and in the new version of this PR, that either `--noipv6` is given, or the host does not have an IPv6 stack available.

I don't think binding a specific address to a dual-stack socket makes any sense, and probably doesn't work. If we wanted to provide that ability we would need to open two separate sockets, one for 4 and another for 6, so they could each be bound to a separate specific address. This would then involve using `select()` or `poll()` to handle them, instead of the blocking read loop we currently use. That would be a separate project for 4.1.0 or later!

Do you know what proportion of server operators actually use `--serverbindip`? I suspect it is extremely small.

For now, we could just stipulate that `--serverbindip` is only valid with `--noipv6`.

src/main.cpp
```diff
@@ -242,11 +242,11 @@ int main ( int argc, char** argv )
...
 
         // Enable IPv6 ---------------------------------------------------------
-        if ( GetFlagArgument ( argv, i, "-6", "--enableipv6" ) )
+        if ( GetFlagArgument ( argv, i, "--noipv6", "--noipv6" ) )
         {
-            bEnableIPv6 = true;
-            qInfo() << "- IPv6 enabled";
-            CommandLineOptions << "--enableipv6";
+            bDisableIPv6 = true;
+            qInfo() << "- IPv6 disabled";
```
ann0see, line 269: I'd like that this makes clear that disabling IPv6 is not recommended. Maybe add (this is not recommended)

src/main.cpp
```diff
@@ -1109,7 +1104,7 @@ QString UsageArguments ( char** argv )
            "  -Q, --qos               set the QoS value. Default is 128. Disable with 0\n"
            "                          (see the Jamulus website to enable QoS on Windows)\n"
            "  -t, --notranslation     disable translation (use English language)\n"
-           "  -6, --enableipv6        enable IPv6 addressing (IPv4 is always enabled)\n"
+           "      --noipv6            disable IPv6 addressing (IPv4 is always enabled)\n"
```
ann0see, line 1122:
```suggestion
           "      --noipv6            disable IPv6 addressing (only set this in case you have issues with IPv6. IPv4 is always enabled)\n"
```

ann0see (commented):
Will need to test this more thoroughly.

src/client.cpp
```diff
@@ -50,7 +50,8 @@ CClient::CClient ( const quint16  iPortNumber,
     bIsInitializationPhase ( true ),
     bMuteOutStream ( false ),
     fMuteOutStreamGain ( 1.0f ),
-    Socket ( &Channel, iPortNumber, iQosNumber, "", bNEnableIPv6 ),
+    bIPv6Available ( false ),
```
ann0see, line 75: Personal preference would be to have this flipped assuming IPv6 availability is there by default and we only set it to false in case we see it isn't. That would be more in line with the default enabled idea.

src/main.cpp
softins: Seems a bit verbose for a usage message, and probably better put in the documentation. Most people will probably not even be aware of IPv6, apart from the most technical.

src/client.cpp
softins: This boolean is internal only, not for user-facing presentation, and I feel this way round makes the logic cleaner. This is my reasoning:

1. We can safely assume all systems are IPv4 capable, but not necessarily IPv6 capable. So that makes the starting point ipv6avail = false
2. The first thing we do, unless the user has explicitly told us not to, it attempt to create a v6-capable socket.
3. If we succeed in doing so, we can set the ipv6avail flag to true.
4. At this point, if the flag is still false (whether by user command or failure), we know we need to create a v4 socket.

I think to invert that logic would probably result in some unnecessary `else` clauses, and make the flow less tidy. I can't see a good reason to do so.

src/main.cpp
softins: Probably better in the documentation. It could get a bit wearing telling the user off every time when it's a conscious decision that has been made.

pljones:
> Do you know what proportion of server operators actually use `--serverbindip`? I suspect it is extremely small.

From what I remember, the reason for using it was a pretty technical one (strange routing / bridging configuration or something).  I was pretty sure there'd be only the single instance of it being used and argued against it on that basis -- but the position was "there's no other way to fix it", so that made me a bit more reasonable 😄.

> For now, we could just stipulate that `--serverbindip` is only valid with `--noipv6`.

The alternative is that `--serverbindip` _implies_ `--noipv6`: so we'd get to the end of command line parsing and, if `--serverbindip` was present without `--noipv6`, issue a message to that effect.

That or error if it's mismatched.

In either case, it needs to be clear in the Release Announcement.

src/main.cpp
ann0see: I'd still like to have a clear "not recommended" thing here. It's mostly philosophical - I agree - but users should be very clear that this may degrade experience- especially if IPv6 has lower latency.

src/main.cpp
```diff
@@ -242,11 +242,11 @@ int main ( int argc, char** argv )
         }
 
         // Enable IPv6 ---------------------------------------------------------
```
ann0see: Disable IPv6?

src/main.cpp
```diff
@@ -242,11 +242,11 @@ int main ( int argc, char** argv )
...
-        if ( GetFlagArgument ( argv, i, "-6", "--enableipv6" ) )
+        if ( GetFlagArgument ( argv, i, "--noipv6", "--noipv6" ) )
```
ann0see, line 266: Also do we really want to fail if -6 was not removed before the upgrade and forgotten to be removed? At least a warning would be good if -6 was given. 

So either:
* Fail and give warning about -6 being removed. Remove warning in 4.1
* don't fail and give warning, fail in 4.1

softins: > Also do we really want to fail if -6 was not removed before the upgrade and forgotten to be removed? At least a warning would be good if -6 was given.

I don't think that's a problem. The user will see the failure once (`Jamulus: Unknown option '-6' -- use '--help' for help`), remove the `-6` from their command line, retry, and all will be good :-)

I think from 3.x to 4.x we're allowed to do things like that.

src/main.cpp
softins: Good spot - will be fixed in the next push

softins:
> > For now, we could just stipulate that `--serverbindip` is only valid with `--noipv6`.
> 
> The alternative is that `--serverbindip` _implies_ `--noipv6`: so we'd get to the end of command line parsing and, if `--serverbindip` was present without `--noipv6`, issue a message to that effect.
> 
> That or error if it's mismatched.

`--serverbindip` will work on a host with no IPv6, even if `--noipv6` is not specified.

I think what I'll do is output a warning after creating the `AF_INET6` socket, if `strServerBindIP` is not empty.

> In either case, it needs to be clear in the Release Announcement.

Agreed.

ann0see:
What happens if you have a link local/ULA but no GUA?
We should test if it works as expected with the local only addresses.

softins:
> What happens if you have a link local/ULA but no GUA? We should test if it works as expected with the local only addresses.

Well it will work in the same way as current `jamulus -6` does. I have never used a ULA, so not sure how to set one up. I did try running a server on my Pi after disabling IPv6 on my router, so the Pi only had a link-local address (starting with `FE80::`, which is different from a ULA). The server gave a warning that it couldn't determine the port's external IPv6, and fell back to `::1` for the external address it would give to a server. It still responded fine to IPv4.

ann0see:
So IPv6 connectivity doesn't work for LLAs?

I do have a network with ULAs (and no GUAs which I currently have no access to). 

Anyway. It's not related to this PR - I agree; But setups with ULAs exist - for example my WireGuard VPN needs those as I have a changing IPv6 prefix. That's common in Germany.

src/socket.cpp
```diff
@@ -102,57 +112,60 @@ void CSocket::Init ( const quint16 iNewPortNumber, const quint16 iNewQosNumber,
...
-        // FIXME: If binding a dual-protocol interface to a specific address, does it cease to be dual-protocol?
+            bIPv6Available = true; // this is a reference to CClient::bIPv6Available or CServer::bIPv6Available
 
-        // TODO - ALLOW IPV6 ADDRESS
-        // if ( !strServerBindIP.isEmpty() )
-        //{
-        //    UdpSocketInAddr.sin_addr.s_addr = htonl ( QHostAddress ( strServerBindIP ).toIPv4Address() );
-        //}
-        // END TODO - ALLOW IPV6 ADDRESS
+            qInfo() << "IPv6/IPv4 dual-protocol socket created";
```
ann0see: Dual stack socket is more in line what I would call it. Also I don't think we need to output this in prod.

softins:
> So IPv6 connectivity doesn't work for LLAs?

I think IPv6 with LLAs is point-to-point and cannot be routed via a node.

> I do have a network with ULAs (and no GUAs which I currently have no access to).

Great, are you able to do any testing and report back?

> Anyway. It's not related to this PR - I agree; But setups with ULAs exist - for example my WireGuard VPN needs those as I have a changing IPv6 prefix. That's common in Germany.

ann0see (commented):
Tested it and if ipv6 is correctly set up I think it does work (and uses it). One of my servers seems to not behave well though (the server I am running a VPN on). ~I suspect it to be a DNS problem.
Maybe not though. Probably some loopback which is not configured. ~ No. I do get pings back so there's definitely something else off.

ann0see:
> Great, are you able to do any testing and report back?

Concerning ULA - no not yet. I might be able to simulate it in future - but that may take some time. My go to network is ipv4 only and I need to do tests over a VPN to a VPS. So it's all a bit wonky.

ann0see:
> I think IPv6 with LLAs is point-to-point and cannot be routed via a node.

That's fine. If you run a local Jamulus server this should not matter?

src/socket.cpp
softins: 1. We techies might call it "stack", but I think "protocol" is more understandable to most, as that is what IPv4 and IPv6 are.
2. This message is only output once at startup, and I think it can be a useful confirmation to the user. I'd prefer to leave it in.

ann0see: But dual-stack is the "industry standard". I would rather not output something which can be misinterpreted.

ann0see:
Ok. It does NOT compile for iOS. CSocket::SendPacket there's an ifdef at the end and the Init() function is called incorrectly with the wrong number of arguments.

softins:
> > I think IPv6 with LLAs is point-to-point and cannot be routed via a node.
> 
> That's fine. If you run a local Jamulus server this should not matter?

I just tried pinging between LAN hosts using the LLA, and it required `%eth0` (interface name) at the end of the address. I'm not sure how that would translate into programming for Jamulus, or even if it's valid. ULAs worked fine (my router now gives out a ULA as well as a GUA), so I think that is the way to go with private LANs. I would say LLAs are outside the scope of this PR.

```
tony@pi:~ $ ping6 fe80::21d:aaff:feb8:fdd8
ping6: Warning: IPv6 link-local address on ICMP datagram socket may require ifname or scope-id => use: address%<ifname|scope-id>
PING fe80::21d:aaff:feb8:fdd8(fe80::21d:aaff:feb8:fdd8) 56 data bytes
^C
--- fe80::21d:aaff:feb8:fdd8 ping statistics ---
6 packets transmitted, 0 received, 100% packet loss, time 5119ms

tony@pi:~ $ ping6 fe80::21d:aaff:feb8:fdd8%eth0
PING fe80::21d:aaff:feb8:fdd8%eth0(fe80::21d:aaff:feb8:fdd8%eth0) 56 data bytes
64 bytes from fe80::21d:aaff:feb8:fdd8%eth0: icmp_seq=1 ttl=64 time=8.58 ms
64 bytes from fe80::21d:aaff:feb8:fdd8%eth0: icmp_seq=2 ttl=64 time=0.253 ms
64 bytes from fe80::21d:aaff:feb8:fdd8%eth0: icmp_seq=3 ttl=64 time=0.236 ms
64 bytes from fe80::21d:aaff:feb8:fdd8%eth0: icmp_seq=4 ttl=64 time=0.208 ms
64 bytes from fe80::21d:aaff:feb8:fdd8%eth0: icmp_seq=5 ttl=64 time=0.209 ms
^C
--- fe80::21d:aaff:feb8:fdd8%eth0 ping statistics ---
5 packets transmitted, 5 received, 0% packet loss, time 4062ms
rtt min/avg/max/mdev = 0.208/1.896/8.576/3.339 ms

tony@pi:~ $ ping6 fd09:4197:db34:1:a00:27ff:fe8e:b8ff
PING fd09:4197:db34:1:a00:27ff:fe8e:b8ff(fd09:4197:db34:1:a00:27ff:fe8e:b8ff) 56 data bytes
64 bytes from fd09:4197:db34:1:a00:27ff:fe8e:b8ff: icmp_seq=1 ttl=64 time=1.02 ms
64 bytes from fd09:4197:db34:1:a00:27ff:fe8e:b8ff: icmp_seq=2 ttl=64 time=0.898 ms
64 bytes from fd09:4197:db34:1:a00:27ff:fe8e:b8ff: icmp_seq=3 ttl=64 time=0.957 ms
64 bytes from fd09:4197:db34:1:a00:27ff:fe8e:b8ff: icmp_seq=4 ttl=64 time=0.908 ms
^C
--- fd09:4197:db34:1:a00:27ff:fe8e:b8ff ping statistics ---
4 packets transmitted, 4 received, 0% packet loss, time 3004ms
rtt min/avg/max/mdev = 0.898/0.945/1.019/0.047 ms
```

softins:
> Ok. It does NOT compile for iOS. CSocket::SendPacket there's an ifdef at the end and the Init() function is called incorrectly with the wrong number of arguments.

Ah, OK. I'll investigate...

ann0see:
Good. Then only the iOS failure needs to be fixed.

softins:
> Good. Then only the iOS failure needs to be fixed.

Done. It compiles, although I can't test it.

softins:
As well as Linux, also successfully tested on FreeBSD 15, Windows 10 and macOS 10.15

ann0see:
iOS also seems to work.

ann0see:
(Please see [TestFlight](https://testflight.apple.com/join/xZRwhf63) for the latest build)

src/socket.cpp
ann0see: I still would prefer not to output this line. Maybe for the ipv4 one but it's more of a debug output imo.

ann0see:
Even though it's not part of this PR - we may want to ask the initial poster about `--serverbindip` and should investigate if there's a need to support it on IPv6 too.

softins:
> Even though it's not part of this PR - we may want to ask the initial poster about `--serverbindip` and should investigate if there's a need to support it on IPv6 too.

This was issue #141 raised in 2020 and PR #1561 merged in 2021.

src/socket.cpp
softins: I still prefer keeping it. It provides confirmation to the user, which may be useful, and is very low cost, since it only happens once at startup. I've changed it from dual-protocol to dual-stack.

rdica:
> As well as Linux, also successfully tested on FreeBSD 15, Windows 10 and macOS 10.15

For my own edification, since FBSD defaults bindv6only to true, does this change force FBSD to have it disabled in order to successfully start jamulus server? (Its highly possible I misread the changes to `socket.cpp`)

While Linux defaults to having it disabled I have in the past specifically enabled it as I wanted separate sockets per protocol, like FBSD does.

softins:
> > As well as Linux, also successfully tested on FreeBSD 15, Windows 10 and macOS 10.15
> 
> For my own edification, since FBSD defaults bindv6only to true, does this change force FBSD to have it disabled in order to successfully start jamulus server? (Its highly possible I misread the changes to `socket.cpp`)

I don't know about `bindv6only`. My tests have been done on a vanilla install of FreeBSD 15 in a Virtualbox VM. In fact the changes I made to get FreeBSD to work properly were in #3664 which is in 3.12.0.

The current PR doesn't make any changes specific to FreeBSD - my comment was just that I had tested it still worked.

> While Linux defaults to having it disabled I have in the past specifically enabled it as I wanted separate sockets per protocol, like FBSD does.

Jamulus only has one UDP socket open, which is dual-stack when IPv6 is enabled. To have separate v4 and v6 sockets within Jamulus would need a fundamental redesign of the socket code to enable two sockets to be monitored using `poll()` or `select()`, and would likely also have an impact on latency.

ann0see (approved):
So it seems to be fine from my POV. I think we can put it into the next beta/alpha. I know @pljones may not want to?

ann0see:
The serverbindip ipv6 unavailability needs an issue raised. In theory, there should be ways to bind to ipv4, ipv6 or dual stack without needing two sockets. But this can be investigated later. IPv6 allows to have ipv4 embedded addresses for example.

pljones:
We need the capabilities work completed.

I'm happy with this - but we could hold it back until the Directories work is ready, as it doesn't really do much of use and potentially is a source of irritation we don't yet need to give people.

ann0see:
I don't think it's going to cause irritation. Directories are a separate thing and it's going to take a while.

softins:
I'd like to get this in sooner rather than later. I will need to update my TCP additions after this is in, and it will be a better foundation for the IPv6 directory stuff I've just started working on.

src/main.cpp
```diff
@@ -240,12 +241,12 @@ int main ( int argc, char** argv )
             continue;
         }
 
-        // Enable IPv6 ---------------------------------------------------------
-        if ( GetFlagArgument ( argv, i, "-6", "--enableipv6" ) )
```
pljones, line 266: OK, I think we should leave `--enableipv6` available initially -- with a warning that it's deprecated (and non-functional) now that it's the default.

This will ease transition -- i.e. server operators taking the latest version will not have to rewrite their start up files just because of this.

softins:
> The serverbindip ipv6 unavailability needs an issue raised. In theory, there should be ways to bind to ipv4, ipv6 or dual stack without needing two sockets. But this can be investigated later. IPv6 allows to have ipv4 embedded addresses for example.

Definitely to be investigated later. But I just asked Google AI:

- Is it possible to bind both ip4 and ip6 local addresses to a dual-stack socket?

And its answer:

Yes, it is possible, but how you do it depends on whether you mean binding to the "ANY" wildcard address or binding to two specific local IP addresses.

### Binding to the Wildcard Address (:: or 0.0.0.0)
  If you want a single socket to listen for both IPv4 and IPv6 traffic on all local interfaces, you only need to create a single IPv6 socket.
  - The Method: Create an IPv6 socket and bind it to the unspecified address (:: / in6addr_any).The Key Step: You must configure the `IPV6_V6ONLY` socket option to disabled (0). When disabled, the OS represents incoming IPv4 connections as IPv4-mapped IPv6 addresses (e.g., ::ffff:92.168.1.10).
  - OS Defaults: Linux typically defaults this option to disabled (allowing dual-stack by default), while Windows defaults it to enabled. Always set it explicitly using `setsockopt()` to ensure portability.

### Binding to Two Specific Local IP Addresses
  If your device has multiple distinct IP addresses (for example, a specific local IPv4 address and a specific local IPv6 address) and you want to bind to both simultaneously, you cannot do this with a single socket.
  - The Method: You must create two separate sockets. One socket is bound to the specific IPv4 address, and the second is bound to the specific IPv6 address.
  - Handling Connections: You will need to use a multiplexing mechanism (such as `select`, `poll`, `epoll`, or `kqueue`) to wait for incoming connections or data on both sockets at the same time.

...which is pretty much what I thought

src/main.cpp
softins: Ok, I'll add an update

softins:
```
 $ ./Jamulus -s -n -6
- server mode chosen
- no GUI mode chosen
IPv6 is now enabled by default: -6 and --enableipv6 are ignored and deprecated
IPv6/IPv4 dual-stack socket created
...
```

softins: Actually, they are not ignored - they make it output that message! I'll change it to say
```
IPv6 is now enabled by default: -6 and --enableipv6 have no effect and are deprecated
```

pljones (approved):

PR #3636, opened 2026-03-11
Support TCP for protocol messages

softins:
**Short description of changes**

Support fallback to TCP for protocol messages, in order to overcome potential loss of large messages due to UDP fragmentation. ~Currently an incomplete draft, for comment as development continues.~

CHANGELOG: Client/Server: Support TCP fallback for protocol messages.

**Context: Fixes an issue?**

Discussed in issue #3242.

**Does this change need documentation? What needs to be documented and how?**

It will need documentation once design and development are complete. Particularly need to explain the firewall requirements for a server or directory.

**Status of this Pull Request**

~Incomplete, still under development. Main server side complete and working. Client side development in progress.~ Complete and ready for review and testing. ~Still marked draft as~ it needs some of the debug messages to be commented out before merging.

**What is missing until this pull request can be merged?**

A lot of testing of both server and client. Intended for Jamulus 4.0.0.

## Checklist

-  [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
-  [x] I tested my code and it does what I want
-  [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#source-code-consistency) 
-  [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
-  [x] I've filled all the content above

PR #3725, opened 2026-06-05
feat: add CLM request for channel level list (message ID 1025)

mcfnord:
**🤖 AI:** Adds `PROTMESSID_CLM_REQ_CHANNEL_LEVEL_LIST` (ID 1025): a new connectionless message that causes a server to immediately reply with the existing `PROTMESSID_CLM_CHANNEL_LEVEL_LIST` (ID 1015) response.

This follows the established CLM request/response pattern used by:
- `CLReqConnClientsList` (1014) → `CLConnClientsList` (1013)
- `CLReqVersionAndOS` (1012) → `CLVersionAndOS` (1011)
- `CLReqServerList` (1007) → `CLRedServerList` (1018) and `CLServerList` (1006)

Numbered 1025: the CLM block runs 1001 to 1022 with no gaps, and 1023 and 1024 are taken by #3636.

The name is not new. The protocol documentation in `protocol.cpp` has referred to `PROTMESSID_CLM_REQ_CHANNEL_LEVEL_LIST` since a7a1549f (2020), at lines 437 and 440, for a message that was never defined. Message 28 is its connected-mode counterpart: `CClient::OnNewConnection` still sends it, and it is kept only for servers 3.4.6 to 3.5.12.

## Implementation

4 files, 28 lines:

- `protocol.h` — define ID 1025, declare evaluator, declare signal
- `protocol.cpp` — document the message in the reference block, add switch case, and one-line `EvaluateCLReqChannelLevelListMes` that emits the signal
- `server.h` — one-line `OnCLReqChannelLevelList` slot that calls the existing `CreateCLChannelLevelListMes`
- `server.cpp` — connect signal to slot

**Zero additional computation.** `vecChannelLevels` is already computed every server tick in `OnTimer()`. The handler simply forwards that existing data to the requester.

## Use case

A monitoring tool can poll the activity level on any Jamulus server by sending a single UDP datagram to the server's game port and reading the nibble-packed level list in the 1015 response. No audio connection, no additional open ports, no configuration required.

The 1015 response already encodes the number of connected clients implicitly (the list length), so the caller can detect idle servers (no clients) versus quiet ones (clients present but all levels ≤ threshold).

## Relation to prior PR

This supersedes #3724.

---

### DIFF
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -83,8 +83,8 @@
 #define PROTMESSID_CLM_CHANNEL_LEVEL_LIST     1015 // channel level list
 #define PROTMESSID_CLM_REGISTER_SERVER_RESP   1016 // status of server registration request
 #define PROTMESSID_CLM_REGISTER_SERVER_EX     1017 // register server with extended information
-#define PROTMESSID_CLM_RED_SERVER_LIST           1018 // reduced server list
-#define PROTMESSID_CLM_REQ_CHANNEL_LEVEL_LIST   1028 // request channel level list (connectionless)
+#define PROTMESSID_CLM_RED_SERVER_LIST        1018 // reduced server list
+#define PROTMESSID_CLM_REQ_CHANNEL_LEVEL_LIST 1028 // request channel level list (connectionless)
 
 // special IDs
 #define PROTMESSID_SPECIAL_SPLIT_MESSAGE 2001 // a container for split messages
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -83,7 +83,8 @@
 #define PROTMESSID_CLM_CHANNEL_LEVEL_LIST     1015 // channel level list
 #define PROTMESSID_CLM_REGISTER_SERVER_RESP   1016 // status of server registration request
 #define PROTMESSID_CLM_REGISTER_SERVER_EX     1017 // register server with extended information
-#define PROTMESSID_CLM_RED_SERVER_LIST        1018 // reduced server list
+#define PROTMESSID_CLM_RED_SERVER_LIST           1018 // reduced server list
+#define PROTMESSID_CLM_REQ_CHANNEL_LEVEL_LIST   1028 // request channel level list (connectionless)
 
 // special IDs
 #define PROTMESSID_SPECIAL_SPLIT_MESSAGE 2001 // a container for split messages
@@ -281,6 +282,7 @@ protected:
     bool EvaluateCLConnClientsListMes ( const CHostAddress& InetAddr, const CVector<uint8_t>& vecData );
     bool EvaluateCLReqConnClientsListMes ( const CHostAddress& InetAddr );
     bool EvaluateCLChannelLevelListMes ( const CHostAddress& InetAddr, const CVector<uint8_t>& vecData );
+    bool EvaluateCLReqChannelLevelListMes ( const CHostAddress& InetAddr );
     bool EvaluateCLRegisterServerResp ( const CHostAddress& InetAddr, const CVector<uint8_t>& vecData );
 
     int iOldRecID;
@@ -348,5 +350,6 @@ signals:
     void CLConnClientsListMesReceived ( CHostAddress InetAddr, CVector<CChannelInfo> vecChanInfo );
     void CLReqConnClientsList ( CHostAddress InetAddr );
     void CLChannelLevelListReceived ( CHostAddress InetAddr, CVector<uint16_t> vecLevelList );
+    void CLReqChannelLevelList ( CHostAddress InetAddr );
     void CLRegisterServerResp ( CHostAddress InetAddr, ESvrRegResult eStatus );
 };
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -106,11 +106,11 @@
 #define PROTMESSID_CLM_REGISTER_SERVER_RESP   1016 // status of server registration request
 #define PROTMESSID_CLM_REGISTER_SERVER_EX     1017 // register server with extended information
 #define PROTMESSID_CLM_RED_SERVER_LIST        1018 // reduced server list
-#define PROTMESSID_CLM_REQ_CHANNEL_LEVEL_LIST 1028 // request channel level list (connectionless)
 #define PROTMESSID_CLM_SERVER_FEATURES        1019 // server features message
 #define PROTMESSID_CLM_REQ_SERVER_FEATURES    1020 // request server features
 #define PROTMESSID_CLM_WELCOME_MESSAGE        1021 // server welcome message
 #define PROTMESSID_CLM_REQ_WELCOME_MESSAGE    1022 // request server welcome message
+#define PROTMESSID_CLM_REQ_CHANNEL_LEVEL_LIST 1025 // request channel level list (connectionless)
 
 // special IDs
 #define PROTMESSID_SPECIAL_SPLIT_MESSAGE 2001 // a container for split messages
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -310,10 +310,10 @@ protected:
     bool EvaluateCLConnClientsListMes ( const CHostAddress& InetAddr, const CVector<uint8_t>& vecData );
     bool EvaluateCLReqConnClientsListMes ( const CHostAddress& InetAddr );
     bool EvaluateCLChannelLevelListMes ( const CHostAddress& InetAddr, const CVector<uint8_t>& vecData );
-    bool EvaluateCLReqChannelLevelListMes ( const CHostAddress& InetAddr );
     bool EvaluateCLRegisterServerResp ( const CHostAddress& InetAddr, const CVector<uint8_t>& vecData );
     bool EvaluateCLReqServerFeaturesMes ( const CHostAddress& InetAddr );
     bool EvaluateCLReqWelcomeMessageMes ( const CHostAddress& InetAddr );
+    bool EvaluateCLReqChannelLevelListMes ( const CHostAddress& InetAddr );
 
     int iOldRecID;
     int iOldRecCnt;
@@ -380,8 +380,8 @@ signals:
     void CLConnClientsListMesReceived ( CHostAddress InetAddr, CVector<CChannelInfo> vecChanInfo );
     void CLReqConnClientsList ( CHostAddress InetAddr );
     void CLChannelLevelListReceived ( CHostAddress InetAddr, CVector<uint16_t> vecLevelList );
-    void CLReqChannelLevelList ( CHostAddress InetAddr );
     void CLRegisterServerResp ( CHostAddress InetAddr, ESvrRegResult eStatus );
     void CLReqServerFeatures ( CHostAddress InetAddr );
     void CLReqWelcomeMessage ( CHostAddress InetAddr );
+    void CLReqChannelLevelList ( CHostAddress InetAddr );
 };
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -280,6 +280,8 @@ CServer::CServer ( const int          iNewMaxNumChan,
 
     QObject::connect ( &ConnLessProtocol, &CProtocol::CLReqConnClientsList, this, &CServer::OnCLReqConnClientsList );
 
+    QObject::connect ( &ConnLessProtocol, &CProtocol::CLReqChannelLevelList, this, &CServer::OnCLReqChannelLevelList );
+
     QObject::connect ( &ServerListManager, &CServerListManager::SvrRegStatusChanged, this, &CServer::SvrRegStatusChanged );
 
     QObject::connect ( &JamController, &recorder::CJamController::RestartRecorder, this, &CServer::RestartRecorder );
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -292,10 +292,11 @@ CServer::CServer ( const int          iNewMaxNumChan,
 
     QObject::connect ( &ConnLessProtocol, &CProtocol::CLReqConnClientsList, this, &CServer::OnCLReqConnClientsList );
 
-    QObject::connect ( &ConnLessProtocol, &CProtocol::CLReqChannelLevelList, this, &CServer::OnCLReqChannelLevelList );
     QObject::connect ( &ConnLessProtocol, &CProtocol::CLReqServerFeatures, this, &CServer::OnCLReqServerFeatures );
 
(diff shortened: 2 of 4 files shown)


ann0see (commented):
Way better!

pljones:
Yes, much better!

mcfnord:
It says there are branch conflicts that must be resolved, and it says there's a C-like coding style failure.

I can try to fix those if it's inherently a PR you'd merge.

I've found a strong case for the feature, but I plan to try accomplishing similar results on a very lean use of message 28, [as recommended here](https://github.com/jamulussoftware/jamulus/pull/3724#issuecomment-4625016269) but [drawing criticism here](https://github.com/orgs/jamulussoftware/discussions/3729).

ann0see:
If the usecase is beneficial and a simple approach, it would not be rejected by me. However, you want to detect bots/listeners? Then please make clear why you need this. Do we want the client to mark those clients? Does it have any information leakage/privacy concerns?

mcfnord:
> you want to detect bots/listeners? Then please make clear why you need this.

Yes, I want to identify silent connections. Maybe it's a listener, maybe it's a bot that doesn't actually listen, but the thing important to me is: it's not a signal in the sense that we, an audio platform, think about signals.

I use this data in an immediate sense, where it detects someone who maybe has stayed connected while walking away for a while, and also in a long-term sense where I can determine that account xyz is always silent and might not be an active musician (so a perma-listener or a bot).


> Do we want the client to mark those clients?

Maybe! https://jamulus.live marks server cards with the "no sound" emoji (🔇) when I determine that every connection on the server is silent. This would be a helpful navigation aide in the Connection dialog. It could be per-connection, although that seems overwhelming. So maybe it just shows that the whole server is silent at the moment.

Notice how people roam around looking for activity... they end up popping in on small groups or singletons to determine if they're active connections. With the "no sound" emoji, I can look at 5 small groups on 5 servers, and see that one is actually making sound. This directs users to the active connection. If the Jamulus client indicated silence this way, it would have the same beneficial effect on navigation.

> Does it have any information leakage/privacy concerns?

I can't think of any. It's just a bit, silent or not-silent. Doesn't present an audio sample or additional metadata. It reveals "level of audio" which might have some application somehow, but I can't think of a malicious use for that detail.

I have added this feature to my own server binary and deployed it to about 20 Jamulus servers. I also get the same information through existing connected lounge lobby bots (mostly Thailand). It provides a helpful clue that the server is silent.

pljones:
> This directs users to the active connection. If the Jamulus client indicated silence this way, it would have the same beneficial effect on navigation.

Not always a beneficial impact, to be fair.  Five drummers, four bass players, three guitarists, flute, violin, tabor and harp - and still no singer...  They may not be silent but they'd probably make a better stab at fun if they split up a bit.  In my view.

pljones (approved):

ann0see:
I also thought about having a bot instrument... (I know it's only partly related)

mcfnord:
This would be cool, let's merge this.

src/protocol.h
```diff
@@ -106,6 +106,7 @@
 #define PROTMESSID_CLM_REGISTER_SERVER_RESP   1016 // status of server registration request
 #define PROTMESSID_CLM_REGISTER_SERVER_EX     1017 // register server with extended information
 #define PROTMESSID_CLM_RED_SERVER_LIST        1018 // reduced server list
+#define PROTMESSID_CLM_REQ_CHANNEL_LEVEL_LIST 1028 // request channel level list (connectionless)
```
ann0see: It's a bit weird to have a jump to 1028 here.

pljones: Agreed -- next sequential is 1023 and should be added sequentially.

pljones (changes requested):

src/protocol.h
mcfnord: **🤖 AI:** I figured if the connected-mode message is 28, then the connectionless one could be 1028. But 1025 is alright, too. Not 1023, though: #3636 already takes 1023 for `PROTMESSID_CLM_TCP_OFFERED` and 1024 for `PROTMESSID_CLM_CLIENT_ID`, so 1025 is the next one actually free.

ann0see:
Actually - after thinking a bit more, I think that we should allow the server admin disable this and expose via the server capabilities if this feature is enabled/disabled for privacy reasons. Default enabled. 

Maybe with --blacklist-messages/--blacklist-features '<feature>' which can then be extended in future?

dingodoppelt:
> Actually - after thinking a bit more, I think that we should allow the server admin disable this and expose via the server capabilities if this feature is enabled/disabled for privacy reasons. Default enabled.
> 
> Maybe with --blacklist-messages/--blacklist-features '' which can then be extended in future?

I second that. Furthermore I'd like to repeat that we might want to log requests like this so admins can monitor activity on their servers and decide whether to block or allow such traffic.

src/protocol.h
```diff
@@ -309,6 +310,7 @@ class CProtocol : public QObject
     bool EvaluateCLConnClientsListMes ( const CHostAddress& InetAddr, const CVector<uint8_t>& vecData );
     bool EvaluateCLReqConnClientsListMes ( const CHostAddress& InetAddr );
     bool EvaluateCLChannelLevelListMes ( const CHostAddress& InetAddr, const CVector<uint8_t>& vecData );
+    bool EvaluateCLReqChannelLevelListMes ( const CHostAddress& InetAddr );
```
pljones: Same order as the `#define`s, please.

src/server.cpp
```diff
@@ -292,6 +292,7 @@ CServer::CServer ( const int          iNewMaxNumChan,
 
     QObject::connect ( &ConnLessProtocol, &CProtocol::CLReqConnClientsList, this, &CServer::OnCLReqConnClientsList );
 
+    QObject::connect ( &ConnLessProtocol, &CProtocol::CLReqChannelLevelList, this, &CServer::OnCLReqChannelLevelList );
```
pljones: Same order as the `#define`s and maintain consistent line spacing.

src/server.h
```diff
@@ -383,6 +383,11 @@ public slots:
 
     void OnCLReqConnClientsList ( CHostAddress InetAddr ) { ConnLessProtocol.CreateCLConnClientsListMes ( InetAddr, CreateChannelList() ); }
 
+    void OnCLReqChannelLevelList ( CHostAddress InetAddr )
```
pljones: Same order as the `#define`s, please.

mcfnord:
**🤖 AI:** @ann0see — the control belongs with the user, not the operator.

An operator who wants to refuse this traffic already has the right layer for it — the kernel: firewall and packet filtering, outside the application, where a host's policy over its own machine sits. A `--blacklist-features` flag makes the operator the arbiter of whether a *user's* signal is shared, and that is the user's call to make.

So the gate, if there is one, is a client checkbox: the user opts out, the client carries the choice, the server honors it in the level reply — the decision staying with the person the signal describes.

It also scopes cleanly. [`CLM_CHANNEL_LEVEL_LIST`](https://github.com/jamulussoftware/jamulus/blob/127fe57177af8ddb08b475633597a61b46d4ecd4/src/protocol.h#L105) has two emitters: the [in-session broadcast](https://github.com/jamulussoftware/jamulus/blob/127fe57177af8ddb08b475633597a61b46d4ecd4/src/server.cpp#L748) that [drives the mixer meters](https://github.com/jamulussoftware/jamulus/blob/127fe57177af8ddb08b475633597a61b46d4ecd4/src/clientdlg.h#L253), and the connectionless reply this PR requests. A user opt-out applies only to the connectionless reply — the navigation view — so a hidden user still shows a live meter to the people they are playing with; only the anonymous query loses them.

This is a separate feature from the request itself, which stays as it is.

ann0see:
> This is a separate feature from the request itself, which stays as it is.

Yes - but the privacy implications should be solved before this gets merged.

mcfnord:
In your view, does unregulated disclosure of name/instrument/nation present the same concerns?

ann0see:
No since the user sets them willingly or not.

ann0see:
Server side blocking is easier. This I proposed this.

dingodoppelt:
> An operator who wants to refuse this traffic already has the right layer for it — the kernel: firewall and packet filtering, outside the application, where a host's policy over its own machine sits. A `--blacklist-features` flag makes the operator the arbiter of whether a _user's_ signal is shared, and that is the user's call to make.
> 
> So the gate, if there is one, is a client checkbox: the user opts out, the client carries the choice, the server honors it in the level reply — the decision staying with the person the signal describes.

I totally disagree here. In my perception internal data mustn't be shared. I still don't like this feature and I think we shouldn't merge this at all because it is actually not even a Jamulus feature. It's a third-party app feature.

ann0see:
> I totally disagree here. In my perception internal data mustn't be shared. 

Yes. One could imagine that it's possible to infer parts of conversations or similar from it. If that's possible, users would never know that this happened.

>  It's a third-party app feature.

Yes. But we should still be somewhat nice to external apps. A human - be it the server admin or enduser should be able to decide what kind of data he/she shares -> needs blacklisting or whitelisting.

mcfnord:
It should be a Jamulus feature. A silent gathering should show it's silent.

pljones:
> It's a third-party app feature.

I think this is an important point.

I was against opening up Jamulus to JSON-RPC on this basis.  It was, eventually, agreed that Jamulus itself would only support exposing the current state of the Client and the Server _locally_ via JSON-RPC.  (This still doesn't protect a Client user from having information exposed without their knowledge, of course, and the Server operator can then expose everything if they proxy the information out to the wider internet by whatever means.)

Of course, it's always been possible to request information about Servers and Clients using the existing (connectionless) protocol messages - but these are the same ones, exposing the same information, as expose that information to all Jamulus users.

We recently added the welcome message connectionless request, information which previously _required a connection_.  I can see the use of it.  It seems harmless enough...

...But it then does open up the floor to requests for other state information to be passed to unknown remote servers without Client or Server operator intervention, that Client users currently think of as known only to others within the session they're connected to.

Is it a line that should be held or crossed?

("Ghost probes" can, of course, already retrieve any connected information -- if there is a slot available on the Server.  That was why the welcome message request was allowed.  It appeared to be a legitimate need to retrieve information about the Server itself, rather than connection-oriented session state.)

ann0see:
@mcfnord. Please exactly describe which problem you want to solve then on your app. Is it finding servers which are just silent such that you can hide them?

pljones:
I mean, we could add "minutes since overall server sound level went above -48dB" to the connect dialog.  That's enough to tell if the server is silent without publishing every Client level (and needing to sample it _frequently_ to have it make any sense).  (A tiny thread similar to the recorder could take a push of the levels, do the maths and signal back the sum every so often.  Clients with an extended server list poll could have this included... we do need a more extensible format... let's switch to a JSON blob.)

mcfnord:
**🤖 AI:** @ann0see Yes, hiding silent servers is the first use, and there are several more. Everything below runs today on my own servers through [the fork's version of this request](https://github.com/mcfnord/jamulus/blob/27e6a66b4da4aedf86f976c79d473413f124cec8/src/server.h#L495).

- **Server visibility.** A server whose level list is all zeros [is marked quiet](https://github.com/mcfnord/JamFan22/blob/ef0cb38cadcb4a904cb23cd1ddec20f8ab913068/JamFan22/Pages/Api.cshtml.cs#L362) and gets [a 🔇 on its card](https://github.com/mcfnord/JamFan22/blob/ef0cb38cadcb4a904cb23cd1ddec20f8ab913068/JamFan22/Pages/Client.cshtml#L890) on [jamulus.live](https://jamulus.live/), and the "active only" filter hides it.
- **A per-minute record of who was audible.** Every player-server-minute is logged with [a silent/audible flag](https://github.com/mcfnord/JamFan22/blob/ef0cb38cadcb4a904cb23cd1ddec20f8ab913068/JamFan22/Services/JamulusCacheManager.cs#L584). The rest of the list reads that record.
- **Listener versus player.** A profile that was audible in under 15% of at least 20 samples is a listener; over 50% is [an active player](https://github.com/mcfnord/JamFan22/blob/ef0cb38cadcb4a904cb23cd1ddec20f8ab913068/JamFan22/CensusIndex.cs#L107). The daily essay on [jamulus.live](https://jamulus.live/) [ranks a session's players](https://github.com/mcfnord/JamFan22/blob/ef0cb38cadcb4a904cb23cd1ddec20f8ab913068/JamFan22/DailyEssayService.cs#L435) with this: confirmed active first, then unknown, then listeners, then rigs that were connected the whole time and never made a sound, and it writes "listener" or "parked" instead of an instrument it never heard.
- **Welcome messages.** The room roster handed to the welcome-message writer [tags always-silent profiles as listeners](https://github.com/mcfnord/JamFan22/blob/ef0cb38cadcb4a904cb23cd1ddec20f8ab913068/JamFan22/WelcomeContext.cs#L1638), so a newcomer is not told that a lurker is a bandmate.
- **Capacity decisions.** Offline, I count a minute as a jam only when at least two players were audible. One of my rooms sat at 37 to 54% presence in every hour of the day while its audible occupancy was 0 to 5.4%, because idle bots and parked rigs hold slots around the clock. Presence alone would have me size and place servers for them.

The nearby-musicians list does not use it, but maybe it should. Silent connections just aren't signals in the way we think of signals.

@pljones on the per-server aggregate, minutes since the overall level last exceeded -48 dB. It keeps the concern and drops the information. The concern raised above is that sound-versus-silence over time can leak something about a session and the people in it never know. A per-server clock carries exactly that signal: with one person connected it is that person's level history, with two it is the union of both. It blurs who is playing in a larger room and changes nothing in the small rooms where a leak would matter most, so it does not resolve the privacy question, it waters down the answer to a different one. What does address the concern is the client opt-out proposed above, and it works the same way for either form: the server leaves that channel out of the connectionless reply, and an opted-out singleton reports as "unknown", never "silent".

On sampling: it does not need to be frequent. [jamulus.live](https://jamulus.live/) asks each of my servers [once every 50 seconds](https://github.com/mcfnord/JamFan22/blob/ef0cb38cadcb4a904cb23cd1ddec20f8ab913068/JamFan22/harvest.cs#L370), and that is enough for the 🔇 and would be enough for the connect dialog. The time-since-sound clock is a good addition in its own right, an hour of silence means something different from a lull between songs, and it could ride in the same reply. As an addition it is welcome, but it seems to offer little privacy gain.

PR #3727, opened 2026-06-06
Feature set cl message

### LINKED ISSUE #3710: Request and receive server features cleanly by cl message rather pretending to connect a client

Currently we can only initiate message exchange with a server by pretending to be a client connecting when we only want to know certain features. There is no mechanism to have the server respond to a request from third-party tools like jamulus-php, jamilizer, jamulusjams, etc.

There should be a connection less protocol message to request and send back server features without initiating a client connection by silent audio packet following immediate disconnection.  
As I think that further integration into the client connection process could be tricky (->backwards compatibilty) we should treat this as an API feature for third-party integration first

No discussion yet, but some code to be dicussed:  
[https://github.com/jamulussoftware/jamulus/compare/main...dingodoppelt:jamulus:feature-set-cl-message](https://github.com/jamulussoftware/jamulus/compare/main...dingodoppelt:jamulus:feature-set-cl-message)

pljones:

Adds two new connection-less protocol messages, one to request and one to return server features.

CHANGELOG: Request and receive server features by connection-less message

**Context: Fixes an issue?**

Fixes 3710

Yes.  See below.  Will also need updates to Client and Server documentation.

Proof of concept.  Needs:
* Confirmation of solution
* JSONRPC support
* Connect Dialog support

See above.

## Checklist

### DIFF
--- a/src/protocol.cpp
+++ b/src/protocol.cpp
@@ -459,6 +459,19 @@ CONNECTION LESS MESSAGES
           five times for one registration request at 500ms intervals.
           Beyond this, it should "ping" every 15 minutes
           (standard re-registration timeout).
+
+
+- PROTMESSID_CLM_SERVER_FEATURES: Bitmask of enabled server features
+
+    +------------------+
+    | 4 bytes number n |
+    +------------------+
+
+
+- PROTMESSID_CLM_REQ_SERVER_FEATURES: Request bitmask of enabled server features
+
+    note: does not have any data -> n = 0
+
 */
 
 #include "protocol.h"
@@ -946,6 +959,10 @@ void CProtocol::ParseConnectionLessMessageBody ( const CVector<uint8_t>& vecbyMe
 
     case PROTMESSID_CLM_REGISTER_SERVER_RESP:
         EvaluateCLRegisterServerResp ( InetAddr, vecbyMesBodyData );
+        break;
+
+    case PROTMESSID_CLM_REQ_SERVER_FEATURES:
+        EvaluateCLReqServerFeaturesMes ( InetAddr );
         break;
     }
 }
@@ -2618,6 +2635,24 @@ bool CProtocol::EvaluateCLRegisterServerResp ( const CHostAddress& InetAddr, con
     emit CLRegisterServerResp ( InetAddr, static_cast<ESvrRegResult> ( iSvrRegResult ) );
 
     return false; // no error
+}
+
+bool CProtocol::EvaluateCLReqServerFeaturesMes ( const CHostAddress& InetAddr )
+{
+    // invoke message action
+    emit CLReqServerFeatures ( InetAddr );
+
+    return false; // no error
+}
+
+void CProtocol::CreateCLServerFeaturesMes ( const CHostAddress& InetAddr, const uint32_t iFeatures )
+{
+    int              iPos = 0; // init position pointer
+    CVector<uint8_t> vecData ( 4 );
+
+    PutValOnStream ( vecData, iPos, iFeatures, 4 );
+
+    CreateAndImmSendConLessMessage ( PROTMESSID_CLM_SERVER_FEATURES, vecData, InetAddr );
 }
 
 /******************************************************************************\
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -45,6 +45,7 @@
 \******************************************************************************/
 
 #include "server.h"
+#include "util.h"
 
 // CServer implementation ******************************************************
 CServer::CServer ( const int          iNewMaxNumChan,
@@ -291,6 +292,8 @@ CServer::CServer ( const int          iNewMaxNumChan,
 
     QObject::connect ( &ConnLessProtocol, &CProtocol::CLReqConnClientsList, this, &CServer::OnCLReqConnClientsList );
 
+    QObject::connect ( &ConnLessProtocol, &CProtocol::CLReqServerFeatures, this, &CServer::OnCLReqServerFeatures );
+
     QObject::connect ( &ServerListManager, &CServerListManager::SvrRegStatusChanged, this, &CServer::SvrRegStatusChanged );
 
     QObject::connect ( &JamController, &recorder::CJamController::RestartRecorder, this, &CServer::RestartRecorder );
@@ -470,6 +473,60 @@ void CServer::OnNewConnection ( int iChID, int iTotChans, CHostAddress RecHostAd
 
     // logging of new connected channel
     Logging.AddNewConnection ( RecHostAddr.InetAddr, iTotChans );
+}
+
+void CServer::OnCLReqServerFeatures ( CHostAddress RecHostAddr )
+{
+    // This is a bitmask of features enabled at the server.
+    // EFeatureSet from util.h is used to shift each bool into position
+    uint32_t iFeatures = 0;
+
+    // Use 64 samples frame size mode? (argument -F)
+    iFeatures |= ( !bUseDoubleSystemFrameSize << FS_FAST_UPDATE );
+
+    // Multithreading enabled? (argument -T)
+    iFeatures |= ( bUseMultithreading << FS_MULTITHREADING );
+
+    // Recording directory set? (argument -R)
+    // If a recording directory is set a server could potentially record all client audio
+    iFeatures |= ( GetRecorderInitialised() << FS_RECORDER_ENABLED );
+
+    // Will an idle server start recording when a client joins or is it already recording an active session?
+    // (argument --norecord disables recording by default)
+    iFeatures |= ( ( JamController.GetRecorderState() == RS_RECORDING ) << FS_IS_RECORDING );
+
+    // Delay pan enabled? (argument -P)
+    iFeatures |= ( bDelayPan << FS_DELAY_PAN );
+
+    // IPv6 available? (argument --noipv6 disables this feature)
+    iFeatures |= ( bIPv6Available << FS_IPV6_AVAILABLE );
+
+    // "Max" audio quality setting enabled? (argument --noraw disables this feature)
+    iFeatures |= ( !bDisableRaw << FS_RAW_AUDIO );
+
+    // Disconnect all clients on quit? (argument -d)
+    iFeatures |= ( bDisconnectAllClientsOnQuit << FS_DISCONONQUIT );
+
+    // Has welcome message? (argument -w)
+    iFeatures |= ( !strWelcomeMessage.isEmpty() << FS_HAS_WELCOME_MESSAGE );
+
+    // Logging enabled? (argument -l)
+    iFeatures |= ( Logging.IsLogging() << FS_IS_LOGGING );
+
+    // Licence agreement required? (argument -L)
+    iFeatures |= ( ( eLicenceType != LT_NO_LICENCE ) << FS_HAS_LICENCE );
+
+    // TODO:
+    // Running a GUI? (argument -n disables the GUI)
+    // iFeatures |= (  << FS_HAS_GUI );
+    //
+    // // RPC interface enabled? (argument --jsonrpcport)
+    // iFeatures |= (  << FS_RPC_ENABLED );
+
+    // qDebug() << QString::number(iFeatures, 2).rightJustified(32, '0');
+
+    // Create and send the message
+    ConnLessProtocol.CreateCLServerFeaturesMes ( RecHostAddr, iFeatures );
 }
 
 void CServer::OnServerFull ( CHostAddress RecHostAddr )
--- a/src/serverlogging.h
+++ b/src/serverlogging.h
@@ -66,6 +66,7 @@ public:
     void AddServerStopped();
 
     void AddNewConnection ( const QHostAddress& ClientInetAddr, const int iNumberOfConnectedClients );
+    bool IsLogging() { return bDoLogging; }
 
 protected:
     void    operator<< ( const QString& sNewStr );
--- a/src/util.h
+++ b/src/util.h
@@ -617,6 +617,25 @@ enum EDirectoryType
     AT_CUSTOM               = 7 // Must be the last entry!
 };
 
+// Server feature set ----------------------------------------------------------
+enum EFeatureSet
+{
+    // used for protocol -> enum values must be fixed
+    FS_FAST_UPDATE         = 0,
+    FS_MULTITHREADING      = 1,
+    FS_RECORDER_ENABLED    = 2,
+    FS_IS_RECORDING        = 3,
+    FS_DELAY_PAN           = 4,
+    FS_IPV6_AVAILABLE      = 5,
+    FS_RAW_AUDIO           = 6,
+    FS_DISCONONQUIT        = 7,
+    FS_HAS_WELCOME_MESSAGE = 8,
+    FS_IS_LOGGING          = 9,
+    FS_HAS_LICENCE         = 10,
+    FS_HAS_GUI             = 11,
+    FS_RPC_ENABLED         = 12
+};
+
 inline QString DirectoryTypeToString ( EDirectoryType eAddrType )
 {
     switch ( eAddrType )
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -106,6 +106,8 @@
 #define PROTMESSID_CLM_REGISTER_SERVER_RESP   1016 // status of server registration request
 #define PROTMESSID_CLM_REGISTER_SERVER_EX     1017 // register server with extended information
 #define PROTMESSID_CLM_RED_SERVER_LIST        1018 // reduced server list
+#define PROTMESSID_CLM_SERVER_FEATURES        1019 // server features message
+#define PROTMESSID_CLM_REQ_SERVER_FEATURES    1020 // request server features
 
 // special IDs
 #define PROTMESSID_SPECIAL_SPLIT_MESSAGE 2001 // a container for split messages
@@ -177,6 +179,7 @@ public:
     void CreateCLReqConnClientsListMes ( const CHostAddress& InetAddr );
     void CreateCLChannelLevelListMes ( const CHostAddress& InetAddr, const CVector<uint16_t>& vecLevelList, const int iNumClients );
     void CreateCLRegisterServerResp ( const CHostAddress& InetAddr, const ESvrRegResult eResult );
+    void CreateCLServerFeaturesMes ( const CHostAddress& InetAddr, const uint32_t iResult );
 
     static bool ParseMessageFrame ( const CVector<uint8_t>& vecbyData,
                                     const int               iNumBytesIn,
@@ -304,6 +307,7 @@ protected:
     bool EvaluateCLReqConnClientsListMes ( const CHostAddress& InetAddr );
     bool EvaluateCLChannelLevelListMes ( const CHostAddress& InetAddr, const CVector<uint8_t>& vecData );
     bool EvaluateCLRegisterServerResp ( const CHostAddress& InetAddr, const CVector<uint8_t>& vecData );
+    bool EvaluateCLReqServerFeaturesMes ( const CHostAddress& InetAddr );
 
     int iOldRecID;
     int iOldRecCnt;
@@ -371,4 +375,5 @@ signals:
     void CLReqConnClientsList ( CHostAddress InetAddr );
     void CLChannelLevelListReceived ( CHostAddress InetAddr, CVector<uint16_t> vecLevelList );
     void CLRegisterServerResp ( CHostAddress InetAddr, ESvrRegResult eStatus );
+    void CLReqServerFeatures ( CHostAddress InetAddr );
 };
--- a/src/server.h
+++ b/src/server.h
@@ -396,6 +396,8 @@ public slots:
 
     void OnCLUnregisterServerReceived ( CHostAddress InetAddr ) { ServerListManager.Remove ( InetAddr ); }
 
+    void OnCLReqServerFeatures ( CHostAddress InetAddr );
+
     void OnCLDisconnection ( CHostAddress InetAddr );
 
     void OnAboutToQuit();


src/server.cpp
```diff
@@ -472,6 +453,22 @@ void CServer::OnNewConnection ( int iChID, int iTotChans, CHostAddress RecHostAd
     Logging.AddNewConnection ( RecHostAddr.InetAddr, iTotChans );
 }
 
+void CServer::OnCLReqServerInfo ( CHostAddress RecHostAddr )
+{
+    uint32_t iFeatures = 0;
+
+    iFeatures |= ( !bUseDoubleSystemFrameSize                            << FS_SMALL_NETW_BUF );
```
pljones: Don't conflate "small network buffers" in the client with "double frame size" in the server -- you've looked at this now 😄.

src/util.h
```diff
@@ -613,6 +591,20 @@ enum EDirectoryType
     AT_CUSTOM               = 7 // Must be the last entry!
 };
 
+// Server feature set ----------------------------------------------------------
+enum EFeatureSet
+{
+    // used for protocol -> enum values must be fixed
+    FS_SMALL_NETW_BUF       = 0,
```
pljones: This should be `FS_FAST_UPDATE` and be true if `bUseDoubleSystemFrameSize` is false.

src/util.h
```diff
@@ -613,6 +591,20 @@ enum EDirectoryType
...
+    FS_MULTITHREADING       = 1,
+    FS_RECORDER_INIT        = 2,
+    FS_DISABLE_RECORDING    = 3,
+    FS_IS_RECORDING         = 4,
```
pljones: I don't think we need three flags:
* `FS_RECORDER_ENABLED` - true if `GetRecorderInitialised() && !GetDisableRecording()`
* `FS_IS_RECORDING` - true if `JamController.GetRecorderState() == RS_RECORDING` (you had `!=`?)

src/util.h
```diff
@@ -613,6 +591,20 @@ enum EDirectoryType
...
+    FS_DELAY_PAN            = 5,
+    FS_ENABLE_IPV6          = 6,
+    FS_RAW_AUDIO            = 7
+};
```
pljones, line 637: Might be worth adding:
* `FS_DISCONONQUIT` -- true if the connection will drop if the server exits

ann0see:
Could you please also check which protocol messages could be replaced by this?

src/protocol.cpp
```diff
@@ -459,6 +437,19 @@ CONNECTION LESS MESSAGES
           five times for one registration request at 500ms intervals.
           Beyond this, it should "ping" every 15 minutes
           (standard re-registration timeout).
+
+
+- PROTMESSID_CLM_SERVER_INFO: Bitmask of enabled server features
```
ann0see: I initially phrased it as "capabilities". FYI. 

There were old versions which didn't support panning I think. Assuming this message was already implemented at this time, We would have used this message to advertise pan support. 

In think we should make it very clear what this message can and cannot encode

src/server.cpp
```diff
@@ -472,6 +453,22 @@ void CServer::OnNewConnection ( int iChID, int iTotChans, CHostAddress RecHostAd
...
+    iFeatures |= ( bUseMultithreading                                    << FS_MULTITHREADING );
+    iFeatures |= ( GetRecorderInitialised()                              << FS_RECORDER_INIT );
+    iFeatures |= ( GetDisableRecording()                                 << FS_DISABLE_RECORDING );
+    iFeatures |= ( (JamController.GetRecorderState() != RS_RECORDING)    << FS_IS_RECORDING );
+    iFeatures |= ( bDelayPan                                             << FS_DELAY_PAN );
+    iFeatures |= ( bEnableIPv6                                           << FS_ENABLE_IPV6 );
+    // iFeatures |= ( bDisableRaw              << FS_RAW_AUDIO );
```
ann0see: Shouldn't be un commented?

src/protocol.cpp
pljones: It's a new message being added for this new feature.

ann0see: Yes. I know. But we should specify what this feature can and what it can't expose. I'm thinking of a specification/architectural viewpoint

ann0see (commented):
Maybe add "has welcome message" also

pljones:
My view is that it's exposing the operating state of the Server.

If you're running with the UI or using JSONRPC, some state can change (one day, it'll all be controllable, hopefully).

However, it's not for large blocks of data: client list, welcome message, etc.

I see it as providing answers to questions a Client user might have about Server capabilities (I also favour using that term) - so on/off states.

That's why I think it should be stuff that gets shown on the Connect Dialog (although I'm thinking of a "toggle extra connect info" option in Advanced Settings).

ann0see:
Ok. Can it be used by the client to to enable/disable certain features such as raw audio? 
I would say yes.

src/server.cpp
```diff
@@ -472,6 +453,22 @@ void CServer::OnNewConnection ( int iChID, int iTotChans, CHostAddress RecHostAd
...
+{
+    uint32_t iFeatures = 0;
+
+    iFeatures |= ( !bUseDoubleSystemFrameSize                            << FS_SMALL_NETW_BUF );
+    iFeatures |= ( bUseMultithreading                                    << FS_MULTITHREADING );
+    iFeatures |= ( GetRecorderInitialised()                              << FS_RECORDER_INIT );
+    iFeatures |= ( GetDisableRecording()                                 << FS_DISABLE_RECORDING );
+    iFeatures |= ( (JamController.GetRecorderState() != RS_RECORDING)    << FS_IS_RECORDING );
+    iFeatures |= ( bDelayPan                                             << FS_DELAY_PAN );
+    iFeatures |= ( bEnableIPv6                                           << FS_ENABLE_IPV6 );
```
softins: `bEnableIPv6` no longer exists. Should now be `pServer->IsIPv6Available()` or `pClient->IsIPv6Available()` as appropriate. And `FS_ENABLE_IPV6` should be `FS_IPV6_AVAILABLE`.

src/util.h
```diff
@@ -613,6 +591,20 @@ enum EDirectoryType
...
+enum EFeatureSet
+{
+    // used for protocol -> enum values must be fixed
+    FS_SMALL_NETW_BUF       = 0,
+    FS_MULTITHREADING       = 1,
+    FS_RECORDER_INIT        = 2,
+    FS_DISABLE_RECORDING    = 3,
+    FS_IS_RECORDING         = 4,
+    FS_DELAY_PAN            = 5,
+    FS_ENABLE_IPV6          = 6,
```
softins: Now needs to be `FS_IPV6_AVAILABLE`, as mentioned above.

softins:
> Maybe add "has welcome message" also

Yes, I agree. I plan to add a separate protocol request to fetch the welcome message. Would be useful for Explorer, and we could add a means within the Connect Dialog to right-click on a host and view the welcome message.

softins:
> Ok. Can it be used by the client to to enable/disable certain features such as raw audio? I would say yes.

I'm not sure that's very useful, and may add significant complication for some features, which set up things at initialisation time. I would treat this PR as just a way of exposing information, and any means of control could be investigated and proposed as a subsequent PR.

dingodoppelt:
```
--- ./src/server.cpp	(original)
+++ ./src/server.cpp	(reformatted)
@@ -479,15 +479,15 @@
 {
     uint32_t iFeatures = 0;
 
-    iFeatures |= ( !bUseDoubleSystemFrameSize                            << FS_FAST_UPDATE );
-    iFeatures |= ( bUseMultithreading                                    << FS_MULTITHREADING );
-    iFeatures |= ( GetRecorderInitialised() && !GetDisableRecording()    << FS_RECORDER_ENABLED );
-    iFeatures |= ( (JamController.GetRecorderState() == RS_RECORDING)    << FS_IS_RECORDING );
-    iFeatures |= ( bDelayPan                                             << FS_DELAY_PAN );
-    iFeatures |= ( IsIPv6Available()                                     << FS_IPV6_AVAILABLE );
-    iFeatures |= ( bDisableRaw                                           << FS_RAW_AUDIO );
-    iFeatures |= ( bDisconnectAllClientsOnQuit                           << FS_DISCONONQUIT );
-    iFeatures |= ( !strWelcomeMessage.isEmpty()                          << FS_HAS_WELCOME_MESSAGE );
+    iFeatures |= ( !bUseDoubleSystemFrameSize << FS_FAST_UPDATE );
+    iFeatures |= ( bUseMultithreading << FS_MULTITHREADING );
+    iFeatures |= ( GetRecorderInitialised() && !GetDisableRecording() << FS_RECORDER_ENABLED );
+    iFeatures |= ( ( JamController.GetRecorderState() == RS_RECORDING ) << FS_IS_RECORDING );
+    iFeatures |= ( bDelayPan << FS_DELAY_PAN );
+    iFeatures |= ( IsIPv6Available() << FS_IPV6_AVAILABLE );
+    iFeatures |= ( bDisableRaw << FS_RAW_AUDIO );
+    iFeatures |= ( bDisconnectAllClientsOnQuit << FS_DISCONONQUIT );
+    iFeatures |= ( !strWelcomeMessage.isEmpty() << FS_HAS_WELCOME_MESSAGE );
 
     ConnLessProtocol.CreateCLServerInfoMes ( RecHostAddr, iFeatures );
 }
```

I don't know if the style check is right this time...

dingodoppelt:
I'd suggest adding two flags to see if the server 
1. is using a log file `!strLoggingFileName.isEmpty()`
2. will show a licence agreement dialog after connection `eLicenceType != LT_NO_LICENCE`

any objections?

softins:
> I don't know if the style check is right this time...

What OS do you do your dev on? If you have `clang-format` version 14 installed, you can just do `make clang_format` before committing and it will format all the style to the same standard as the CI expects.

dingodoppelt:
> > I don't know if the style check is right this time...
> 
> What OS do you do your dev on? If you have `clang-format` version 14 installed, you can just do `make clang_format` before committing and it will format all the style to the same standard as the CI expects.

I use kate on arch. I just used the "format" feature of the built-in LSP-client for the first time 👍

softins:
Would be good if you can install `clang-format`. The Jamulus tree has a file with all the correct format rules. Unless that LSP format feature uses the same file of rules?

dingodoppelt:
> Would be good if you can install `clang-format`. The Jamulus tree has a file with all the correct format rules. Unless that LSP format feature uses the same file of rules?

Yes, that file is used by kate. I just ran `clang-format --dry-run --style=file:.clang-format src/*` manually for a test and it wants to change lots of comments and some code. Is that expected and why doesn't the auto style check complain?

src/protocol.cpp
dingodoppelt: We are not exclusively checking for features available in newer versions. There are status flags like recording enabled and if there is a welcome message. Those aren't strictly capabilities so I think calling it a "feature set" makes more sense here. I'd therefore propose to call it `PROTMESSID_CLM_SERVER_FEATURE_SET`

softins:
> Yes, that file is used by kate. I just ran `clang-format --dry-run --style=file:.clang-format src/*` manually for a test and it wants to change lots of comments and some code. Is that expected and why doesn't the auto style check complain?

It may depend on which version of `clang-format` you have. When I had 13, it used to make some unexpected changes. When I moved to 14 it was fine. I don't know about versions newer than that.

dingodoppelt:
> It may depend on which version of `clang-format` you have. When I had 13, it used to make some unexpected changes. When I moved to 14 it was fine. I don't know about versions newer than that.

Ah, I'm on `clang-format version 22.1.6`

dingodoppelt:
The `.clang-format` file says:  
```
# Style for Clang-Format version 10
# https://releases.llvm.org/10.0.0/tools/clang/docs/ClangFormatStyleOptions.html
```
Could that become an issue when not updated or is only the comment outdated?

pljones:
> I'd suggest adding two flags to see if the server
> 1. is using a log file `!strLoggingFileName.isEmpty()`
> 2. will show a licence agreement dialog after connection `eLicenceType != LT_NO_LICENCE`
>
> any objections?

Both make sense to me.  I'd actually thought of the licence one myself and forgot to add it!

src/server.cpp
ann0see, line 485: I'd like to have comments which describe exactly which feature is enabled - especially if it's non obvious like the recorder. Yes, we do have FS_* but that may get ambiguous for non trivial features in future.

dingodoppelt:
How about:
1. Is the server a directory?
2. Is the server registered with a directory?

src/util.h
dingodoppelt: > I don't think we need three flags:
> 
>     * `FS_RECORDER_ENABLED` - true if `GetRecorderInitialised() && !GetDisableRecording()`
> 
>     * `FS_IS_RECORDING` - true if `JamController.GetRecorderState() == RS_RECORDING` (you had `!=`?)

From my understanding `JamController.GetRecorderState() == RS_RECORDING` is only true when `GetRecorderInitialised() && !GetDisableRecording()` is true (recorder may be actively recording or idling) while both are not transparently reflecting if the server is *currently* recording.  
I think we need to know...
- if the recorder is *initialised* -> a recording directory is set with `-R`
- if recording will start automatically when a client joins the server -> recording is *enabled* by NOT using `--norecord`
- if the server is *currently* recording which is currently not known in the jamController. @pljones Is there a way for CServer to know for sure if a recording is currently taking place on a server?

src/server.cpp
dingodoppelt: While writing these I came to the conclusion that the recorder flags need a makeover and I'll push the comments once I figured out how to handle recording state properly.

src/util.h
pljones: I think only two states need to be shown -- either the Server _is_ recording or the Server _could_ be recording.  The reason _why_ is not really relevant.  So that's what I was trying to reflect.

I think you're right that I got the logic wrong, though.

(I mean, that `--norecord` is in use doesn't stop someone monitoring their server with JSONRPC and turning on recording automatically when a Client joins.  It's just a different mechanism.)

pljones:
> How about:
> 
> 1. Is the server a directory?
> 2. Is the server registered with a directory?

I don't think these are relevant to the use case.

* "Is a Server a Directory?" -- a Client is not affected by this directly.  They can go to explorer.jamulus.app and check, if they want to find out.  There's no need to add it.
* "Is the server registered with a directory?" -- This doesn't help the Client at all, on its own.

src/util.h
dingodoppelt: > I think only two states need to be shown -- either the Server _is_ recording or the Server _could_ be recording. The reason _why_ is not really relevant. So that's what I was trying to reflect.

So we will stick with `FS_RECORDER_ENABLED` which tells us if the server *could* be recording and `FS_IS_RECORDING` to see the server's current recording state.  
@pljones Do you have a suggestion on how to query the jamController for its current recording state? I think for knowing that we need to request information from the jamRecorder directly which isn't visible to CServer.

dingodoppelt:
>     * "Is a Server a Directory?" -- a Client is not affected by this directly.  They can go to explorer.jamulus.app and check, if they want to find out.  There's no need to add it.
> 
>     * "Is the server registered with a directory?" -- This doesn't help the Client at all, on its own.

I'm not thinking this from a client perspective. A script might need this info for setting up things and I would argue that we might want to reflect the current state of the server with the highest possible coverage.

dingodoppelt:
We might also want to know if the RPC interface is enabled with a potential `FS_RPC_ENABLED`

ann0see:
> We might also want to know if the RPC interface is enabled

Technically it should not be exposed to the internet so I'd argue that we don't want to expose it in the message.

src/util.h
pljones: > `FS_IS_RECORDING` - true if `jamController.GetRecorderState() == RS_RECORDING`

Doesn't that work?  It _should_ only be true when a client is connected and recording is enabled.  (Presuming there's a handy reference to the JamController instance in the Server instance...  Needs a thread-safe call, as it's running in a separate thread.  I never remember if Qt makes that magic "just work" but it always makes me nervous...)

dingodoppelt: > > `FS_IS_RECORDING` - true if `jamController.GetRecorderState() == RS_RECORDING`
> 
> Doesn't that work? It _should_ only be true when a client is connected and recording is enabled.  

The problem is `FS_IS_RECORDING` gives the same value as  `FS_RECORDER_ENABLED`. I just needed to get rid of `&& !GetDisableRecording()`

pljones:
> I'm not thinking this from a client perspective. A script might need this info for setting up things and I would argue that we might want to reflect the current state of the server with the highest possible coverage.

A script running where?  And, as always, it needs a use case before we just add it.  I'd argue we stick to KISS.

I'd say a separate "What directory are you registered with?" request would be better:
- "Myself, I am a directory"
- "\<X\>"
- None

tells the client (small c) what it needs to know in one request / reply pair.  (And still it needs a use case before we'd add it.)

pljones:
> We might also want to know if the RPC interface is enabled with a potential `FS_RPC_ENABLED`

This one I can see is of more interest -- it lets the Client know that the server state may be controlled and changed whilst they are connected, and that information about the connection may be monitored externally to the server.

Similarly, we should expose whether the server is running headless or with the GUI up.

dingodoppelt:
> > We might also want to know if the RPC interface is enabled with a potential `FS_RPC_ENABLED`
> 
> This one I can see is of more interest -- it lets the Client know that the server state may be controlled and changed whilst they are connected, and that information about the connection may be monitored externally to the server.

This is what I thought. I can see @ann0see's point, too. Would it be a safety concern if that information was available? I think it is of interest if the server is "managed" and therefore subject to unexpected behaviour. I, for example, use RPC for a chat where users can chat with the musicians on the server without appearing in the mixerboard. This flag could be used to have third-party apps display that info or filter for it and I think that would be really useful.

> Similarly, we should expose whether the server is running headless or with the GUI up.

Good idea, but not currently not known to CServer (same goes for `FS_RPC_ENABLED`...)  
These two are not as easy as the others were ;)

pljones:
I think we need to draw a line _now_ for getting this into 4.0.0 as soon as possible.

pljones:
Anyone got anything pending on this?

ann0see:
Review wise it seems good (we can always add new messages). But we may want to think about where in the client to use this.

src/server.cpp
```diff
@@ -472,6 +475,60 @@ void CServer::OnNewConnection ( int iChID, int iTotChans, CHostAddress RecHostAd
...
+    iFeatures |= ( ( eLicenceType != LT_NO_LICENCE ) << FS_HAS_LICENCE );
+
+    // TODO:
+    // Running a GUI? (argument -n disables the GUI)
+    // iFeatures |= (  << FS_HAS_GUI );
+    //
+    // // RPC interface enabled? (argument --jsonrpcport)
+    // iFeatures |= (  << FS_RPC_ENABLED );
+
+    // qDebug() << QString::number(iFeatures, 2).rightJustified(32, '0');
```
dingodoppelt, line 526: I'd like to leave this in for the moment for documentation purposes. This is supposed to be used for possible future enhancements and I think the comments reflect that well enough.

ann0see (changes requested):
Testing with wireshark seems to show that e.g. -L does not get correctly encoded. We might just send one byte over the wire.

dingodoppelt:
> Testing with wireshark seems to show that e.g. -L does not get correctly encoded. We might just send one byte over the wire.

Ouch, I got that terribly wrong. Fixed

softins:
> > Testing with wireshark seems to show that e.g. -L does not get correctly encoded. We might just send one byte over the wire.
> 
> Ouch, I got that terribly wrong. Fixed

Yes, I found that too, but am still testing. I have a couple of other minor comments. Will post my review soon.

src/protocol.cpp
```diff
@@ -2620,6 +2637,24 @@ bool CProtocol::EvaluateCLRegisterServerResp ( const CHostAddress& InetAddr, con
...
+
+    return false; // no error
+}
+
+void CProtocol::CreateCLServerInfoMes ( const CHostAddress& InetAddr, const uint32_t iFeatures )
+{
+    int              iPos = 0; // init position pointer
+    CVector<uint8_t> vecData ( sizeof ( uint32_t ) );
+
+    PutValOnStream ( vecData, iPos, iFeatures, 4 );
```
softins:
```suggestion
    CVector<uint8_t> vecData ( 4 );

    PutValOnStream ( vecData, iPos, iFeatures, 4 );
```

It looks odd expressing the same related values two ways, and the protocol spec is specific about the number. All the other methods with fixed length messages just use a number.

src/server.cpp
```diff
@@ -472,6 +475,60 @@ void CServer::OnNewConnection ( int iChID, int iTotChans, CHostAddress RecHostAd
...
+
+    // Will an idle server start recording when a client joins or is it already recording an active session?
+    // (argument --norecord disables recording by default)
+    iFeatures |= ( ( JamController.GetRecorderState() == RS_RECORDING ) << FS_IS_RECORDING );
+
+    // Delay pan enabled? (argument -P)
+    iFeatures |= ( bDelayPan << FS_DELAY_PAN );
+
+    // IPv6 available? (argument --noipv6 disables this feature)
+    iFeatures |= ( IsIPv6Available() << FS_IPV6_AVAILABLE );
```
softins:
```suggestion
    iFeatures |= ( bIPv6Available << FS_IPV6_AVAILABLE );
```

Within `CServer`, `bIPv6Available` is accessible directly. The getter method is just for making it available outside the class methods.

src/server.cpp
```diff
@@ -472,6 +475,60 @@ void CServer::OnNewConnection ( int iChID, int iTotChans, CHostAddress RecHostAd
...
+    iFeatures |= ( !bDisableRaw << FS_RAW_AUDIO );
+
+    // Disconnect all clients on quit? (argument -d)
+    iFeatures |= ( bDisconnectAllClientsOnQuit << FS_DISCONONQUIT );
+
+    // Has welcome message? (argument -w)
+    iFeatures |= ( !strWelcomeMessage.isEmpty() << FS_HAS_WELCOME_MESSAGE );
+
+    // Logging enabled? (argument -l)
+    iFeatures |= ( Logging.bIsLogging() << FS_IS_LOGGING );
```
softins:
```suggestion
    iFeatures |= ( Logging.IsLogging() << FS_IS_LOGGING );
```

We just use the `b` prefix for boolean variables, not for their accessor methods.

src/serverlogging.h
```diff
@@ -66,6 +66,7 @@ class CServerLogging
     void AddServerStopped();
 
     void AddNewConnection ( const QHostAddress& ClientInetAddr, const int iNumberOfConnectedClients );
+    bool bIsLogging() { return bDoLogging; }
```
softins:
```suggestion
    bool IsLogging() { return bDoLogging; }
```

As comment above.

softins (changes requested):
A few minor points of style, but other than that, this looks good, and I've tested it.

dingodoppelt:
> A few minor points of style, but other than that, this looks good, and I've tested it.

Thanks for your suggestions! Applied :)

softins (approved):
Looking good - thank you!

src/protocol.cpp
dingodoppelt: We can still rename this if necessary, any thoughts?

dingodoppelt: We actually should rename this, as I just found `strServerInfo` being used for the argument `--serverinfo` which is something completely different.  
I'd go with `PROTMESSID_CLM_SERVER_FEATURE_SET` if there are no objections or other suggestions.

softins: OK. Another possibility might be `PROTMESSID_CLM_SERVER_CONFIG`? I don't have a strong view, but do agree with the need to avoid the confusion of two similar names meaning different things.

ann0see: I prefer SERVER_FEATURE_SET

softins: Or "features" instead of "feature set"?

softins: > I prefer SERVER_FEATURE_SET

I think I prefer the simpler SERVER_FEATURES without the set

dingodoppelt: "FEATURES" it is. The shorter the better :+1:

softins:
... and the comments for the #defines in protocol.h :)

dingodoppelt:
> ... and the comments for the #defines in protocol.h :)

Fixed :rofl:

softins (approved):

dingodoppelt:
Should I rebase or squash or both?

softins:
> Should I rebase or squash or both?

Probably worth combining the many small commits, so I guess it's a good idea rebasing to current `main` at the same time.

softins (approved):

pljones:
Merging (I can't approve the PR as I raised it, even though it's @dingodoppelt's code).

PR #3728, opened 2026-06-06
Fix memory leak found by copilot

ann0see:

Tries to fix memory leak found by copilot: See: https://github.com/jamulussoftware/jamulus/pull/3702 This should be valiated. I checked that the audio alerts still play - and it seems to work, but further investigation is still needed. I experienced hearing an audio-alert twice - but that may just be due to my testing mythology.

CHANGELOG: Bugfix: Fix memory leak for audio alerts

**Context: Fixes an issue?**

See: https://github.com/jamulussoftware/jamulus/pull/3702

No

Needs review and testing.

Testing

## Checklist

### DIFF
--- a/src/clientdlg.cpp
+++ b/src/clientdlg.cpp
@@ -876,9 +876,7 @@ void CClientDlg::OnChatTextReceived ( QString strChatText )
 {
     if ( pSettings->bEnableAudioAlerts )
     {
-        QSoundEffect* sf = new QSoundEffect();
-        sf->setSource ( QUrl::fromLocalFile ( ":sounds/res/sounds/new_message.wav" ) );
-        sf->play();
+        PlayAudioAlert ( QUrl::fromLocalFile ( ":sounds/res/sounds/new_message.wav" ) );
     }
     ChatDlg.AddChatText ( strChatText );
 
@@ -927,9 +925,7 @@ void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
 {
     if ( pSettings->bEnableAudioAlerts && iNewNumClients > iClients )
     {
-        QSoundEffect* sf = new QSoundEffect();
-        sf->setSource ( QUrl::fromLocalFile ( ":sounds/res/sounds/new_user.wav" ) );
-        sf->play();
+        PlayAudioAlert ( QUrl::fromLocalFile ( ":sounds/res/sounds/new_user.wav" ) );
     }
 
     // iNewNumClients will be zero on the first trigger of this signal handler when connecting to a new server.
--- a/src/clientdlg.h
+++ b/src/clientdlg.h
@@ -60,7 +60,6 @@
 #include <QMessageBox>
 #include <QFileDialog>
 #include <QActionGroup>
-#include <QSoundEffect>
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 6, 0 )
 #    include <QVersionNumber>
 #endif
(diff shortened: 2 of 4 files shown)


src/clientdlg.cpp
```diff
@@ -927,7 +933,13 @@ void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
...
-        QSoundEffect* sf = new QSoundEffect();
+        QSoundEffect* sf = new QSoundEffect ( this );
+        connect ( sf, &QSoundEffect::playingChanged, this, [sf]() {
+            if ( !sf->isPlaying() )
+            {
+                sf->deleteLater();
+            }
+        } );
         sf->setSource ( QUrl::fromLocalFile ( ":sounds/res/sounds/new_user.wav" ) );
         sf->play();
```
ann0see: Can't we free sf here? play() is not blocking?

softins (approved):
This looks good. I've built and tried it, and verified by adding debug messages that the `deleteLater()` is being called when it should be in each case.

src/clientdlg.cpp
```diff
@@ -927,7 +933,13 @@ void CClientDlg::OnNumClientsChanged ( int iNewNumClients )
 {
     if ( pSettings->bEnableAudioAlerts && iNewNumClients > iClients )
     {
-        QSoundEffect* sf = new QSoundEffect();
+        QSoundEffect* sf = new QSoundEffect ( this );
```
pljones: I think I'd feel happier with this duplicated code factored out into a method. That way any future sound effects can be played by the same code, without having to write another copy.

ann0see:
Weird. It compiled for me locally.

softins:
> Weird. It compiled for me locally.

I think it might be better to put `PlayAudioAlert()` in `clientdlg` instead of moving it to `util`.

softins:
> > Weird. It compiled for me locally.
> 
> I think it might be better to put `PlayAudioAlert()` in `clientdlg` instead of moving it to `util`.

~Yes, you definitely should, as it is specific to the client GUI.~

The Linux build in CI builds both a GUI and a headless version, and it is the headless one that is failing, because the include paths for QtWidgets, QtMultimedia and QtGUI are not in the headless build.

If you want to keep `PlayAudioAlert()` in `util`, you need to put the `#include <QSoundEffect>` within the `#ifndef HEADLESS` section with the other GUI includes, and also protect the declaration and definition of `PlayAudioAlert()` with `#ifndef HEADLESS` too. ~I think it's easier and more logical just to move it back to `clientdlg`.~

softins:
I suppose if you might want to use it in `serverdlg` too, you could keep it in `util`, and just do the `#ifndef` thing.

softins:
I just pushed a commit to try it.

ann0see:
Thanks. The idea was that one could use audio alerts everywhere. So util.\* seemed the best fitting location.

softins (approved):

pljones:
> > Weird. It compiled for me locally.
> 
> I think it might be better to put `PlayAudioAlert()` in `clientdlg` instead of moving it to `util`.

Not in the _graphical_ UI, if it's avoidable.  I'd rather we had the assistive UI support separately, so it could be used even if the graphical UI was not in used.  There's too much functional code there.

softins:
> Not in the _graphical_ UI, if it's avoidable. I'd rather we had the assistive UI support separately, so it could be used even if the graphical UI was not in used. There's too much functional code there.

It's ok, we've left it in `util.*`, but excluded for headless compilation, as `QSoundEffect` is a GUI-only module in Qt anyway.

I think this is ready to go in now.

pljones (approved):

pljones:
Backported to release/3_12.

Issue #3029, opened 2023-03-22
ban / blacklist misbahaved users


yeahletters:
Is there a function to ban or blacklist certain users with unique IP addresses on my own jamulus server? I had an unwanted guest (troll) posting profanties in the chat using capitals.

ann0see:
Hi @yeahletters That's bad to hear.

 Yes, there are unofficial ones like https://github.com/dingodoppelt/jamban and @Rob-NY also has one. 

We have multiple requests like https://github.com/jamulussoftware/jamulus/discussions/1057 or https://github.com/jamulussoftware/jamulus/discussions/1371 but I think there's a general disagreement about this topic I'm afraid.

PR #3730, opened 2026-06-08
Add new RPC methods and notifications (mostly for scripting interaction) to server and client

dingodoppelt:

Add the following RPC methods and notifications to server and client. These are useful mostly for scripting interaction.

- ~~jamulus/getAvailableMethods~~
- jamulusclient/setMuted
- jamulusserver/broadcastChatMessage
- ~~jamulusserver/getClientDetails~~
- jamulusserver/chatMessageReceived
- jamulusserver/clientConnected
- jamulusserver/clientDisconnected
- ~jamulusserver/privateChatMessage~

CHANGELOG: Add new RPC methods and notifications (mostly for scripting interaction) to server and client

**Context: Fixes an issue?**

Relates to #3029 #3447 #2828 #1057 #1371 
This is needed for [Jamban](https://github.com/dingodoppelt/jamban) and others.

I guess the automated RPC documentation is sufficient?

up for discussion

Approval

## Checklist

### DIFF
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -291,6 +291,23 @@ Results:
 | result | string | Always "ok". |
 
 
+### jamulusclient/setMuted
+
+Mutes or unmutes the client.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.muted | boolean | muted (true or false). |
+
+Results:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| result | string | Always "ok". |
+
+
 ### jamulusclient/setName
 
 Sets your name.
@@ -317,6 +334,23 @@ Parameters:
 | Name | Type | Description |
 | --- | --- | --- |
 | params.skillLevel | string | The new skill level (beginner, intermediate, expert, or null). |
+
+Results:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| result | string | Always "ok". |
+
+
+### jamulusserver/broadcastChatMessage
+
+Sends a message (as the server) to all connected clients. This can be used to broadcast messages from external sources (e.g. scripts or  monitoring tools).
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.chatMessage | string | The chat message text. |
 
 Results:
 
@@ -619,5 +653,40 @@ Parameters:
 | params.servers[*].countryId | number | Server country ID (see QLocale::Country). |
 | params.servers[*].country | string | Server country. |
 | params.servers[*].city | string | Server city. |
+
+
+### jamulusserver/chatMessageReceived
+
+Emitted when a chat message is received from either a Jamulus or RPC client and to be broadcast to all connected clients.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.chatMessage | string | Chat message text. |
+
+
+### jamulusserver/clientConnected
+
+Emitted when a client has connected to the server.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.id | number | The channel ID assigned to the client. |
+| params.address | string | The client's address. |
+| params.totalChannels | number | Number of total channels connected to the server. |
+
+
+### jamulusserver/clientDisconnected
+
+Emitted when a client has disconnected from the server.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.id | number | The channel ID assigned to the client. |
 
 
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -663,6 +663,7 @@ Parameters:
 
 | Name | Type | Description |
 | --- | --- | --- |
+| params.id | number | Channel ID of sending client or -1 for RPC sent messages. |
 | params.chatMessage | string | Chat message text. |
 
 
--- a/src/client.h
+++ b/src/client.h
@@ -328,6 +328,7 @@ public:
     // settings
     CChannelCoreInfo ChannelInfo;
     QString          strClientName;
+    void             OnRPCInMuteMyself ( bool bMute ) { OnControllerInMuteMyself ( bMute ); }
 
 public:
     void SetSettings ( CClientSettings* settings );
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -473,6 +473,8 @@ void CServer::OnNewConnection ( int iChID, int iTotChans, CHostAddress RecHostAd
 
     // logging of new connected channel
     Logging.AddNewConnection ( RecHostAddr.InetAddr, iTotChans );
+
+    emit ClientConnected ( iChID, RecHostAddr.InetAddr, iTotChans );
 }
 
 void CServer::OnCLReqServerFeatures ( CHostAddress RecHostAddr )
@@ -929,10 +931,7 @@ void CServer::DecodeReceiveData ( const int iChanCnt, const int iNumClients )
             // and emit the client disconnected signal
             if ( eGetStat == GS_CHAN_NOW_DISCONNECTED )
             {
-                if ( JamController.GetRecordingEnabled() )
-                {
-                    emit ClientDisconnected ( iCurChanID ); // TODO do this outside the mutex lock?
-                }
+                emit ClientDisconnected ( iCurChanID ); // TODO do this outside the mutex lock?
 
                 FreeChannel ( iCurChanID ); // note that the channel is now not in use
 
@@ -1368,6 +1367,7 @@ void CServer::SendChatTextToAllConChannels ( const QString& strChatText )
             vecChannels[i].CreateChatTextMes ( strChatText );
         }
     }
+    emit sentChatMessage ( strChatText );
 }
 
 void CServer::SendChatTextToConChannel ( const int iCurChanID, const QString& strChatText )
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1354,10 +1354,10 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int iCurChanID, con
                                          ChanName.toHtmlEscaped() + "</b></font> " + strChatText.toHtmlEscaped();
 
     // Send chat text to all connected clients ---------------------------------
-    SendChatTextToAllConChannels ( strActualMessageText );
+    SendChatTextToAllConChannels ( iCurChanID, strActualMessageText );
 }
 
-void CServer::SendChatTextToAllConChannels ( const QString& strChatText )
+void CServer::SendChatTextToAllConChannels ( const int iSendingChanID, const QString& strChatText )
 {
     // Send chat text to all connected clients ---------------------------------
     for ( int i = 0; i < iMaxNumChannels; i++ )
@@ -1367,7 +1367,8 @@ void CServer::SendChatTextToAllConChannels ( const QString& strChatText )
             vecChannels[i].CreateChatTextMes ( strChatText );
         }
     }
-    emit sentChatMessage ( strChatText );
+    // forward the message to the RPC server
+    emit sentChatMessage ( iSendingChanID, strChatText );
 }
 
 void CServer::SendChatTextToConChannel ( const int iCurChanID, const QString& strChatText )
--- a/src/server.h
+++ b/src/server.h
@@ -331,6 +331,8 @@ signals:
     void Started();
     void Stopped();
     void ClientDisconnected ( const int iChID );
+    void ClientConnected ( const int iChID, const QHostAddress RecHostAddr, const int iTotChans );
+    void sentChatMessage ( const QString& strChatText );
     void SvrRegStatusChanged();
     void AudioFrame ( const int              iChID,
                       const QString          stChName,
--- a/src/server.h
+++ b/src/server.h
@@ -191,7 +191,7 @@ public:
     void SetEnableDelayPanning ( bool bDelayPanningOn ) { bDelayPan = bDelayPanningOn; }
     bool IsDelayPanningEnabled() { return bDelayPan; }
 
-    void SendChatTextToAllConChannels ( const QString& strChatText );
+    void SendChatTextToAllConChannels ( const int iSendingChanID, const QString& strChatText );
     void SendChatTextToConChannel ( const int iCurChanID, const QString& strChatText );
 
 protected:
@@ -332,7 +332,7 @@ signals:
     void Stopped();
     void ClientDisconnected ( const int iChID );
     void ClientConnected ( const int iChID, const QHostAddress RecHostAddr, const int iTotChans );
-    void sentChatMessage ( const QString& strChatText );
+    void sentChatMessage ( const int iSendingChanID, const QString& strChatText );
     void SvrRegStatusChanged();
     void AudioFrame ( const int              iChID,
                       const QString          stChName,
--- a/src/serverrpc.cpp
+++ b/src/serverrpc.cpp
@@ -56,6 +56,57 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
         Q_UNUSED ( params );
     } );
 
+    /// @rpc_notification jamulusserver/clientConnected
+    /// @brief Emitted when a client has connected to the server.
+    /// @param {number} params.id - The channel ID assigned to the client.
+    /// @param {string} params.address - The client's address.
+    /// @param {number} params.totalChannels - Number of total channels connected to the server.
+    connect ( pServer, &CServer::ClientConnected, [=] ( const int iChanID, const QHostAddress RecHostAddr, const int iTotChans ) {
+        pRpcServer->BroadcastNotification ( "jamulusserver/clientConnected",
+                                            QJsonObject{
+                                                { "id", iChanID },
+                                                { "address", RecHostAddr.toString() },
+                                                { "totalChannels", iTotChans },
+                                            } );
+    } );
+
+    /// @rpc_notification jamulusserver/clientDisconnected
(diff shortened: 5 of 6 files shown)


ann0see:
@mcfnord you might be interested in this.

src/serverrpc.cpp
```diff
@@ -75,6 +130,74 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
...
+    /// @result {string} result.clients[*].address - The client’s address (ip:port).
+    /// @result {string} result.clients[*].name - The client’s name.
+    /// @result {string} result.clients[*].city - The client’s city.
+    /// @result {string} result.clients[*].country - The client’s country.
+    /// @result {string} result.clients[*].instr - The client’s instrument.
+    /// @result {string} result.clients[*].instrpic - The client’s instrument picture.
+    /// @result {string} result.clients[*].skill - The client’s skill.
+    /// @result {number} result.clients[*].jitterBufferSize - The client’s jitter buffer size.
+    /// @result {number} result.clients[*].channels - The number of audio channels of the client.
+    pRpcServer->HandleMethod ( "jamulusserver/getClientDetails", [=] ( const QJsonObject& params, QJsonObject& response ) {
```
dingodoppelt: This will need to go. Before there was a change to `jamulusserver/getClients` in #2918 I needed some more info. I will have to check and change my scripts because the output is slightly rearranged but the data is the same

docs/JSON-RPC.md
```diff
@@ -95,6 +95,23 @@ Results:
 | result | string | "ok" on success |
 
 
+### jamulus/getAvailableMethods
+
+Returns all available rpc methods.
```
ann0see: Do we really need this?

dingodoppelt: I do, but I might be the only one. Having it there doesn't hurt and as stated this is mostly for interactions with scripts so I'd be alright with some obscure functions while they're not crashing things. I've been running this for a few years now and it turned out to be harmless ;)

ann0see: Hmm. I'd prefer to not have this and rather rely on machine readable documentation. 
If it was a REST API you'd have some Open API spec or so.

dingodoppelt: I only use it to populate a menu which could be hard coded as well, so this can go eventually if nobody else needs it besides me.

src/client.h
```diff
@@ -329,6 +329,7 @@ class CClient : public QObject
     // settings
     CChannelCoreInfo ChannelInfo;
     QString          strClientName;
+    void             OnRPCInMuteMyself ( bool bMute ) { OnControllerInMuteMyself ( bMute ); }
```
pljones, line 331: Does this trigger a change in state in the UI?

docs/JSON-RPC.md
```diff
@@ -325,6 +342,23 @@ Results:
 | result | string | Always "ok". |
 
 
+### jamulusserver/broadcastChatMessage
+
+Sends a chat message to all connected clients.
```
pljones:
```suggestion
Sends a chat message from this client to the server.  The server will send this to all connected clients.
```

docs/JSON-RPC.md
```diff
@@ -396,6 +430,24 @@ Results:
 | result.registrationStatus | string | The server registration status as string (see ESvrRegStatus and SerializeRegistrationStatus). |
 
 
+### jamulusserver/privateChatMessage
+
+Sends a chat message to a single connected client.
```
pljones: Please split this out into a separate PR.

pljones (changes requested):
Needs splitting.  The private message feature should be separate.

src/client.h
dingodoppelt: Yes it does

docs/JSON-RPC.md
dingodoppelt: What exactly is "this client" in this case? Should'nt it rather be "this _RPC_ client" (like `nc`, `curl` or some script)? Since this is a server method I think my explanation is sufficient in making clear what this method does.

docs/JSON-RPC.md
dingodoppelt: > **Status of this Pull Request**
up for discussion and happy to receive proposals for more messages or notifications

If that's really necessary I should also get rid of the text above. Since I know that there are some RPC methods floating around in different forks I had hoped to even see some more proposals for methods and notifications in this PR.

dingodoppelt:
> Needs splitting. The private message feature should be separate.

In addition to `jamulusserver/broadcastChatMessage` and `jamulusserver/chatMessageReceived` I think this is the right place for it, tbh. It would make more sense to split the whole PR into three smaller ones
1. Chat
2. Client connection/disconnection
3. set client muted status

but I don't see why only the private chat message should be handled independently.

docs/JSON-RPC.md
pljones: The Client in Jamulus terms is the Client exchanging UDP messages with the Server.  (I should have capitalised both for compliance with the style guide, my bad.)

Maybe
```suggestion
Sends the chat message in the request from the Client to the Server.  The Service will send the message to all connected Clients.
```

pljones:
> but I don't see why only the private chat message should be handled independently.

Because the feature is entirely new.

docs/JSON-RPC.md
dingodoppelt: > The Client in Jamulus terms is the Client exchanging UDP messages with the Server. 

But we are not a client in this case, we are the RPC server and we are processing a request from a client like curl or nc. I'm confused

Edit: To clarify: this is a server method, not client. This is not the client sending a normal chat message just triggered via RPC, it is actually the server itself broadcasting a message to all connected clients.

pljones: Ah right.  I missed that, sorry.  Still, make the doc clearer 😄 .

src/server.h
```diff
@@ -328,6 +330,9 @@ class CServer : public QObject, public CServerSlots<MAX_NUM_CHANNELS>
     void Started();
     void Stopped();
     void ClientDisconnected ( const int iChID );
+    void ClientDisconnectedRPC ( const int iChID );
```
pljones: Why does this need to be a separate signal?  Two slots can connect to the same signal.

src/server.cpp
```diff
@@ -1306,6 +1310,22 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int iCurChanID, con
...
+    // Send chat text to all connected clients ---------------------------------
+    for ( int i = 0; i < iMaxNumChannels; i++ )
+    {
+        if ( vecChannels[i].IsConnected() )
+        {
+            // send message
+            vecChannels[i].CreateChatTextMes ( strChatText );
+        }
+    }
+    emit receivedChatMessage ( strChatText );
```
pljones: This doesn't seem the right place (now I've found it).  The name is "receivedChatMessage" but the method is sending chat messages.

Shouldn't this be in the existing server message that handles inbound chat?

src/server.cpp
```diff
@@ -1306,6 +1310,22 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int iCurChanID, con
             vecChannels[i].CreateChatTextMes ( strActualMessageText );
         }
     }
+    emit receivedChatMessage ( strActualMessageText );
```
pljones: Same as the other one -- this name says "received" and the method is Send.

You probably want this emit in the caller of this method.

src/server.cpp
```diff
@@ -1306,6 +1310,22 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int iCurChanID, con
...
+}
+
+// external chat
+void CServer::CreateAndSendExtChatTextForAllConChannels ( const QString& strChatText )
```
pljones: Please refactor `CServer::CreateAndSendChatTextForAllConChannels` to create the message text and then call this method to send it out to remove the duplicated loop.

src/server.cpp
pljones: Actually, with my suggested refactor, this looks more like the right place.

src/server.cpp
pljones: Even with the refactor, you don't want to be saying a chat message was received at this point -- it's false: either a chat message was received from the Jamulus Client and handled by `CServer::CreateAndSendChatTextForAllConChannels` above or from the JSONRPC client (`pRpcServer->HandleMethod ( "jamulusserver/broadcastChatMessage", [=] ( const QJsonObject& params, QJsonObject& response )`) - which is a different thing altogether.

pljones (changes requested):

src/server.h
dingodoppelt: Because it broke the recorder. I needed an independent signal to not crash the server

src/server.cpp
dingodoppelt: ~`CreateAndSendChatTextForAllConChannels`~
~~1. is protected and can't be called from the RPC server and 
2. it alters the chat text so I'd need to pass an additional parameter to tell it not to do that for RPC sent messages.
3. it expects the message to be from a connected client but there is none because the RPC server is not a connected client.~~

Edit: Forget all the above ;)

The signal `receivedChatMessage` was first used when I wanted a notification (`jamulusserver/chatMessageReceived`) at the server after a client has sent a chat message which the server broadcast to the other clients.  
I can rename it to just `chatMessage` and keep the notification called as it is because it really is just a notification on a received chat message.

src/server.cpp
dingodoppelt: Would it be clearer if I just sent the signal first line in the function? At that point the server hasn't sent anything and only really received a message. (either by an RPC or Jamulus client)

src/server.cpp
dingodoppelt: I renamed it to `sentChatMessage` but left the RPC notification unchanged because its name still fits. This will only notify RPC on a public chat message that is supposed to be seen by every client. Doesn't apply to for example the welcome message or an imaginary private chat message.

src/server.h
pljones: Mmm...  Any idea why?  Signals should be thread-safe if the connect is done right.

dingodoppelt: `ClientDisconnected` is only emit if `JamController.GetRecordingEnabled() == true`. If you emit it somewhere else with recording not enabled the server crashes.

dingodoppelt: This signal was the last one I got working because of threads.  
```
    /// @rpc_notification jamulusserver/clientDisconnected
    /// @brief Emitted when a client has disconnected from the server.
    /// @param {number} params.id - The channel ID assigned to the client.
    connect (
        pServer,
        &CServer::ClientDisconnectedRPC,
        this,
        [=] ( const int iChanID ) {
            pRpcServer->BroadcastNotification ( "jamulusserver/clientDisconnected",
                                                QJsonObject{
                                                    { "id", iChanID },
                                                } );
        },
        Qt::QueuedConnection );
```
Connecting it to `this` and `Qt::QueuedConnection` made it thread safe to the actual client ID arrives at the RPC server. Before it was reporting `151` as the client ID but wasn't crashing. I think the issue must be somewhere in the recorder but I haven't had a look, yet.

pljones: Yeah, the recorder uses it to ask if anyone's still connected, IIRC, which is why I was pretty sure it was threadsafe (as the recorder is on another thread).

Weird it's not happy in your case.

---

Is the recorder using a slot or a lambda?  Try a slot and see if that makes a difference.

dingodoppelt: Weird indeed, because now it seems to work the way it was supposed to in the beginning. I don't know why. Maybe some bug in Qt? I will check if this now works on my Ubuntu servers as well as those were crashing before.

dingodoppelt: When no argument `--recording` is given the server is now a little more chatty than before:  
```
2026-06-17 15:30:52,, server idling -------------------------------------
2026-06-17 15:30:55, ::1, connected (1)
2026-06-17 15:30:58, ::1, connected (2)
CJamRecorder::OnDisconnected: channel 1 disconnected but not recording
CJamRecorder::OnDisconnected: channel 1 disconnected but no currentSession
CJamRecorder::OnDisconnected: channel 0 disconnected but not recording
CJamRecorder::OnDisconnected: channel 0 disconnected but no currentSession
2026-06-17 15:31:01,, server idling -------------------------------------
```

src/server.cpp
```diff
@@ -872,10 +874,7 @@ void CServer::DecodeReceiveData ( const int iChanCnt, const int iNumClients )
             // and emit the client disconnected signal
             if ( eGetStat == GS_CHAN_NOW_DISCONNECTED )
             {
-                if ( JamController.GetRecordingEnabled() )
```
dingodoppelt, line 932: This signal was kind of exclusive to the recorder. Could this be handled in a more general way?

src/server.h
pljones: Well, it makes me happy! 😀

src/server.cpp
pljones: Yeah, no problem with this.

pljones (approved):

dingodoppelt:
Would it make sense to pull 80a5147489a329e5da1a0cbbaeaa41b9835d2c8f into this to have the channel IDs sanity checked as they might be invalid when called over RPC? The boundary check isn't reviewed yet.

Edit: I might as well rebase this on main once #3741 is pulled

src/serverrpc.cpp
```diff
@@ -56,6 +56,57 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
...
+        pRpcServer->BroadcastNotification ( "jamulusserver/clientDisconnected",
+                                            QJsonObject{
+                                                { "id", iChanID },
+                                            } );
+    } );
+
+    /// @rpc_notification jamulusserver/chatMessageReceived
+    /// @brief Emitted when a chat message is received from either a Jamulus or RPC client and to be broadcast to all connected clients.
+    /// @param {string} params.chatMessage - Chat message text.
+    connect ( pServer, &CServer::sentChatMessage, [=] ( const QString& strChatText ) {
```
dingodoppelt: It'd be nice if we saw the channel ID from the client sending the message as well.  
It *must* be impossible to fake a client ID over RPC of course so we'd need to set a channel ID `<0` and check for it.  
With this we'd be able to respond to requests privately over RPC.

dingodoppelt: 3732f3917311b79dc67e7c812cfb163560daabc4 adds the sender's channel ID to the notification, sets the ID to `-1` for RPC sent messages

src/serverrpc.cpp
```diff
@@ -56,6 +56,60 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
...
+    /// @param {string} params.chatMessage - The chat message text.
+    /// @result {string} result - Always "ok".
+    pRpcServer->HandleMethod ( "jamulusserver/broadcastChatMessage", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        auto jsonChatMessage = params["chatMessage"];
+        if ( !jsonChatMessage.isString() )
+        {
+            response["error"] = CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: chatMessage is not a string" );
+            return;
+        }
+
```
pljones, line 110: Meh....  Now I want a named constant for the -1.

dingodoppelt: Good idea, there you go!

pljones (approved):

src/server.cpp
```diff
@@ -929,10 +931,7 @@ void CServer::DecodeReceiveData ( const int iChanCnt, const int iNumClients )
...
-                {
-                    emit ClientDisconnected ( iCurChanID ); // TODO do this outside the mutex lock?
-                }
+                emit ClientDisconnected ( iCurChanID ); // TODO do this outside the mutex lock?
```
ann0see, line 934: Do you want to say with the comment that you are unsure if there might be concurrency buts?

src/serverrpc.cpp
```diff
@@ -56,6 +59,60 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
...
+                                                { "id", iChanID },
+                                            } );
+    } );
+
+    /// @rpc_notification jamulusserver/chatMessageReceived
+    /// @brief Emitted when a chat message is received from either a Jamulus or RPC client and to be broadcast to all connected clients.
+    /// @param {number} params.id - Channel ID of sending client or -1 for RPC sent messages.
+    /// @param {string} params.chatMessage - Chat message text.
+    connect ( pServer, &CServer::sentChatMessage, [=] ( const int iSendingChanID, const QString& strChatText ) {
+        pRpcServer->BroadcastNotification ( "jamulusserver/chatMessageReceived",
```
ann0see, line 91: We might want to be clear what this enables - privacy wise. Server operators could log chat messages now - and this should be made very clear in the privacy policy.

ann0see (changes requested):
Tested and works ok. Please see my comments.

src/server.cpp
dingodoppelt: I think that comment was added by @pljones?

src/serverrpc.cpp
dingodoppelt: That's true. Nonetheless it has never been an issue logging the chat with little effort like connecting a client like many servers do or adding a few lines of code for that purpose.

ann0see (approved):

src/server.cpp
pljones: Yeah - I wanted the ClientDisconnected to stop the recorder (which is on a separate Qt thread, so thread-safe - and hopefully non-interrupting therefore).  My concern over the mutex was how it would affect CServer itself - it shouldn't get interrupted within the mutex by something handling the signal.

PR #3731, opened 2026-06-09
Add jamulusserver/privateChatMessage RPC method for sending messages to single channels

dingodoppelt:

This makes it possible to query the server over RPC to send a chat text to a connected channel. The method takes the chat text and the channel id as arguments.

CHANGELOG: Add jamulusserver/privateChatMessage RPC method for sending messages to single channels

**Context: Fixes an issue?**
No

Part of auto generated RPC docs

Requesting review

## Checklist

-  [X] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#source-code-consistency) 

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1381,16 +1381,13 @@ void CServer::SendChatTextToAllConChannels ( const int iSendingChanID, const QSt
 
 bool CServer::SendChatTextToConChannel ( const int iCurChanID, const QString& strChatText )
 {
-    if ( MathUtils::InRange<int> ( iCurChanID, 0, iMaxNumChannels - 1 ) && vecChannels[iCurChanID].IsConnected() )
+    if ( !MathUtils::InRange<int> ( iCurChanID, 0, iMaxNumChannels - 1 ) || !vecChannels[iCurChanID].IsConnected() )
     {
-        // send message
-        vecChannels[iCurChanID].CreateChatTextMes ( strChatText );
+        return false;
     }
-    else
-    {
-        return true;
-    }
-    return false;
+    // send message
+    vecChannels[iCurChanID].CreateChatTextMes ( strChatText );
+    return true;
 }
 
 void CServer::CreateAndSendRecorderStateForAllConChannels()
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1379,13 +1379,18 @@ void CServer::SendChatTextToAllConChannels ( const int iSendingChanID, const QSt
     emit sentChatMessage ( iSendingChanID, strChatText );
 }
 
-void CServer::SendChatTextToConChannel ( const int iCurChanID, const QString& strChatText )
+bool CServer::SendChatTextToConChannel ( const int iCurChanID, const QString& strChatText )
 {
     if ( MathUtils::InRange<int> ( iCurChanID, 0, iMaxNumChannels - 1 ) && vecChannels[iCurChanID].IsConnected() )
     {
         // send message
         vecChannels[iCurChanID].CreateChatTextMes ( strChatText );
     }
+    else
+    {
+        return true;
+    }
+    return false;
 }
 
 void CServer::CreateAndSendRecorderStateForAllConChannels()
--- a/src/serverrpc.cpp
+++ b/src/serverrpc.cpp
@@ -113,6 +113,28 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
         response["result"] = "ok";
     } );
 
+    /// @rpc_method jamulusserver/privateChatMessage
+    /// @brief Sends a chat message to a single connected client.
+    /// @param {string} params.chatMessage - The chat message text.
+    /// @param {number} params.id - The client's channel id.
+    /// @result {string} result - Always "ok".
+    pRpcServer->HandleMethod ( "jamulusserver/privateChatMessage", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        auto      jsonChatMessage = params["chatMessage"];
+        const int id              = params["id"].toInt();
+        if ( !jsonChatMessage.isString() )
+        {
+            response["error"] = CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: chatMessage is not a string" );
+            return;
+        }
+
+        if ( pServer->SendChatTextToConChannel ( id, jsonChatMessage.toString() ) )
+        {
+            response["error"] = "invalid channel ID";
+            return;
+        }
+        response["result"] = "ok";
+    } );
+
     /// @rpc_method jamulusserver/getRecorderStatus
     /// @brief Returns the recorder state.
     /// @param {object} params - No parameters (empty object).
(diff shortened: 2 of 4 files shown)


pljones:
Thanks.

src/serverrpc.cpp
```diff
@@ -292,6 +292,24 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
...
     } );
+
+    /// @rpc_method jamulusserver/privateChatMessage
+    /// @brief Sends a chat message to a single connected client.
+    /// @param {string} params.chatMessage - The chat message text.
+    /// @param {number} params.id - The client's channel id.
+    /// @result {string} result - Always "ok".
+    pRpcServer->HandleMethod ( "jamulusserver/privateChatMessage", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        auto      jsonChatMessage = params["chatMessage"];
+        const int id              = params["id"].toInt();
```
pljones: This needs validating, probably in `CServer::CreateAndSendChatTextToChannel`

pljones (changes requested):

src/serverrpc.cpp
dingodoppelt: [CServer::CreateAndSendChatTextToChannel](https://github.com/jamulussoftware/jamulus/pull/3731/changes/BASE..b1f38f38c96534aed3ce94d1904ec541b4939e2c#diff-d0ecb51621381529aa2b31bb04fecee4a1b1b7dc0be53978c4878f2c4d4bfe1fR1315) does actually validate the channel ID and will only send the message if the channel is present. It was copied from the original function for sending chat messages.

src/server.cpp
```diff
@@ -1308,6 +1308,17 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int iCurChanID, con
     }
 }
 
+// private chat
+void CServer::CreateAndSendChatTextToChannel ( const QString& strChatText, const int id )
+{
+    // Check if channel is connected ---------------------------------
+    if ( vecChannels[id].IsConnected() )
```
pljones: This will potentially throw an exception if `id` is out of bounds.  The vector isn't MAX_INT in size, nor are negative indices valid.

src/serverrpc.cpp
pljones: Commented there.

src/server.cpp
dingodoppelt: Yes, but this applies to the original code as well, although it is never called with invalid values at the moment. An idea would be to refactor this on main first so I can rebase this PR and #3730 since this refactor is needed for both PRs.

pljones: Not a bad idea, to be honest.  We're hoping to get the JSONRPC out of "experimental" but index out of bounds not being checked for is a red flag...

src/server.cpp
dingodoppelt, line 1359: This is a big issue for dark theme users as the colours assume a light theme is used to be legible. This should be handled in the client entirely based on system theme settings. I've seen we use `styleHints` already. Maybe this can be addressed as well when refactoring the chat message delivery?

ann0see: Uhm yes. I know that iOS shows this bug.

dingodoppelt: This is a tricky one. We can't easily refactor this into the client as only the timestamp and name are coloured, not the message itself :-/

src/server.cpp
dingodoppelt: I did the refactor alongside validation in 80a5147489a329e5da1a0cbbaeaa41b9835d2c8f  
Is that good enough for a PR or something to be discussed first?

pljones: That looks good.

dingodoppelt: Opened #3741

src/server.cpp
dingodoppelt: Opened #3740

pljones (approved):

src/serverrpc.cpp
ann0see: I still think that it should then *not* return OK.
Furthermore, if the id is not given it "defaults" to 0.

At least if id is unset, it should fail.

src/serverrpc.cpp
```diff
@@ -292,6 +292,24 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
         response["result"] = "acknowledged";
         Q_UNUSED ( params );
     } );
+
+    /// @rpc_method jamulusserver/privateChatMessage
+    /// @brief Sends a chat message to a single connected client.
+    /// @param {string} params.chatMessage - The chat message text.
+    /// @param {number} params.id - The client's channel id.
+    /// @result {string} result - Always "ok".
```
ann0see: How do we document errors? It should not rather be "ok" on success, no?

src/serverrpc.cpp
pljones: I'm guessing we have a max message length for UDP CM chat messages?  The JSONRPC request message length should be validated against that and an error returned.

And yes, it needs to return an error for an incorrect channel id.

ann0see: Isn't JSON-RPC using a TCP socket, if so, it should be of basically "infinite" length?

pljones: It needs to not overrun the existing request expectations.

src/serverrpc.cpp
dingodoppelt: If an invalid channel ID is supplied we fail with an error response now.

src/serverrpc.cpp
dingodoppelt: I limited the chat message size to `MAX_LEN_CHAT_TEXT` as defined in `global.h` and in line with current chat implementation

src/serverrpc.cpp
```diff
@@ -292,6 +295,30 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
...
```
ann0see: Thanks. you should make clear that it is ok on success only (in the docs).

ann0see (changes requested):
Documentation

src/serverrpc.cpp
dingodoppelt: I was thinking about that but other methods handle this the same way. I think `result = "ok"` is the equivalent to an `ACK` message

ann0see: I find it confusing still. 

https://github.com/jamulussoftware/jamulus/blob/main/docs/JSON-RPC.md#jamulusclientpollserverlist

Would be more in line with what I'd expect.

dingodoppelt: Other docs should be fixed in a dedicated PR. It seems `"ok"` was used as an `ACK` in other methods and only the PR you referenced changed that.

ann0see (approved):

src/server.cpp
```diff
@@ -1379,13 +1379,18 @@ void CServer::SendChatTextToAllConChannels ( const int iSendingChanID, const QSt
...
     if ( MathUtils::InRange<int> ( iCurChanID, 0, iMaxNumChannels - 1 ) && vecChannels[iCurChanID].IsConnected() )
     {
         // send message
         vecChannels[iCurChanID].CreateChatTextMes ( strChatText );
     }
+    else
+    {
+        return true;
+    }
+    return false;
```
ann0see: Still a bit weird to return false on success - but probably that's fine.

pljones: Might be better as
```suggestion
bool CServer::SendChatTextToConChannel ( const int iCurChanID, const QString& strChatText )
{
    if ( !MathUtils::InRange<int> ( iCurChanID, 0, iMaxNumChannels - 1 ) || !vecChannels[iCurChanID].IsConnected() )
    {
        return false;
    }
    // send message
    vecChannels[iCurChanID].CreateChatTextMes ( strChatText );
    return true;
}
```

src/serverrpc.cpp
```diff
@@ -113,6 +113,30 @@ CServerRpc::CServerRpc ( CServer* pServer, CRpcServer* pRpcServer, QObject* pare
...
+        const int     id              = params["id"].toInt ( INVALID_CLIENT_ID );
+        const QString chatMessage     = jsonChatMessage.toString();
+        if ( chatMessage.isEmpty() || chatMessage.size() > MAX_LEN_CHAT_TEXT )
+        {
+            response["error"] =
+                CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: chatMessage is not a string or malformed" );
+            return;
+        }
+
+        if ( pServer->SendChatTextToConChannel ( id, chatMessage ) )
```
pljones: And similarly
```suggestion
        if ( !pServer->SendChatTextToConChannel ( id, chatMessage ) )
```

PR #3739, opened 2026-06-16
Enable autovectorization on GCC compilers

ann0see:

Enables autovectorization throughout the whole codebase. We found out that there is no vectorization especially of the server side mixing code (https://github.com/jamulussoftware/jamulus/issues/3662#issuecomment-4712581491)
 I consider this safe to default enable; in case of bugs we'd have a compiler bug.

CHANGELOG: Performance: Enable auto vectorization by default

**Context: Fixes an issue?**

Related to: https://github.com/jamulussoftware/jamulus/issues/3662#issuecomment-4712581491

Ready for testing on all supported OS.

Build with multiple compilers

## Checklist

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -1176,6 +1176,9 @@ contains(CONFIG, "opus_shared_lib") {
     }
 }
 
+# Always enable auto vectorization
+QMAKE_CXXFLAGS+=-ftree-vectorize
+
 # disable version check if requested (#370)
 contains(CONFIG, "disable_version_check") {
     message(The version check is disabled.)


dingodoppelt:
As mentioned in #3662 my compiler seems to autovectorize already. I see `-O2` is set when I compile. `-O3` includes `-ftree-vectorize`. Have you looked into setting `-O3` as an alternative as well?  
How can we autobuild with debug symbols? It would be interesting to know if some runners already automagically handle things like my compiler does.

mcfnord:
I benchmarked this change on Linux to help move it forward. Summary: **the claim in the PR description is correct** — with the current `-O2` build flags, g++ vectorizes none of the mixing loops in `CServer::MixEncodeTransmitData`, and this one-line flag fixes that, making the mix loops **2.2–3.8× faster**. It also answers @dingodoppelt's `-O3` question: `-O3` produces essentially identical mixing performance, so `-ftree-vectorize` alone gets all of the benefit with a much smaller change.

**Setup**: g++ 13.3.0 (Ubuntu 24.04, x86-64), current `main` (6b7a0ad), compiled with the exact flags qmake generates (`-pipe -O2 -std=gnu++11 -fPIC` + project defines/includes), comparing three variants: stock `-O2`, `-O2 -ftree-vectorize` (this PR), and `-O3`.

### 1. Compiler vectorization reports on the real `server.cpp`

`-fopt-info-vec-optimized`, filtered to `MixEncodeTransmitData`:

| Variant | Vectorized loops in the mix function |
|---|---|
| `-O2` (current) | **0** (only one trivial SLP basic-block) |
| `-O2 -ftree-vectorize` (this PR) | **8** — all four mono-target mix loops, both stereo-target no-delay-pan mix loops, both `Float2Short` conversion loops |
| `-O3` | same 8 loops |

So on GCC 13 nothing in the mixing path vectorizes today, and the PR's flag reaches exactly the same set of loops as `-O3`. (The delay-pan loops don't vectorize under any variant due to the wrap-around conditionals — pre-existing, unchanged.)

### 2. Microbenchmark of the mixing loops

Loop bodies lifted verbatim from `MixEncodeTransmitData` (`CVector` is `std::vector`, runtime trip count, same access pattern). Each measurement is one full target-channel frame: buffer reset + mix of N clients + `Float2Short` conversion; best of 30 runs × 20 000 reps. Times in ns per target-channel frame:

| Frame | Clients | Mode | `-O2` | `-O2 -ftree-vectorize` | `-O3` | Speedup (PR vs -O2) |
|---:|---:|---|---:|---:|---:|---:|
| 64 | 10 | mono | 1091 | 285 | 260 | **3.8×** |
| 64 | 10 | stereo | 1024 | 407 | 418 | **2.5×** |
| 64 | 50 | mono | 3235 | 1104 | 978 | **2.9×** |
| 64 | 50 | stereo | 4931 | 1986 | 1822 | **2.5×** |
| 128 | 10 | mono | 1339 | 419 | 551 | **3.2×** |
| 128 | 10 | stereo | 2642 | 1177 | 1183 | **2.2×** |
| 128 | 50 | mono | 7873 | 2463 | 2659 | **3.2×** |
| 128 | 50 | stereo | 14248 | 5232 | 5130 | **2.7×** |

Whole-server impact scales with N² (each of N clients gets its own N-way mix). Example: 50 mono clients at the 128-sample frame size = 50 × 7.9 µs = ~394 µs of mixing per 2.67 ms frame period (≈15% of a core) at `-O2`, dropping to ~123 µs (≈4.6%) with this PR. Stereo: ≈27% → ≈10%.

### 3. Safety notes

- These loops vectorize over the *sample* index; each output sample keeps its exact scalar operation order (no FP reassociation — `-ftree-vectorize` does not imply `-ffast-math`), so output is bit-identical, consistent with the "safe to default-enable" assessment.
- The full autobuild matrix on this PR is green, including both Windows builds — MSVC ignores the unknown `-ftree-vectorize` with a warning, so the flag is inert (not harmful) there. If the warning noise bothers anyone it could be scoped to gcc/clang mkspecs, but it's not a correctness issue.
- @dingodoppelt: on g++ 13, plain `-O2` definitely does not enable loop vectorization (`-ftree-vectorize` is only implied from `-O3` up) — if your local builds vectorize at `-O2`, something else in your flags is adding it. And per the table above, `-O3` buys nothing further for this code, so the PR's minimal flag seems like the right call.

Benchmark source available on request.

dingodoppelt (approved):

pljones:
> MSVC ignores the unknown -ftree-vectorize with a warning

I think it would be worth putting a guard in the `Jamulus.pro`, in that case, with a comment saying it's only in GCC.

Additionally, it might be worth investigating what the equivalent in the MSVC builds looks like - in terms of performance now and potential for optimisation.  This step isn't a blocker, though.

pljones:
I've added this to 4.0.0 as it's a small change with potentially wide-scale benefit right now.

mcfnord:
@pljones — the GCC-only guard you asked for. I checked all three compilers to get the scope right:

- **GCC 13.3** (Linux autobuilder): does *not* auto-vectorize at `-O2` (`-ftree-vectorize` is off until `-O3`), so the flag is genuinely needed — it takes `MixEncodeTransmitData` from 4 to 100 packed FP ops in the disassembly.
- **Clang 18** (macOS): already vectorizes these loops at `-O2` without the flag, and doesn't warn if it's passed — so it doesn't need it.
- **MSVC**: warns on the unknown flag.

So the flag is only needed for GCC:

```qmake
# GCC does not auto-vectorize at -O2 (only from -O3 up); Clang already does, MSVC
# would warn. Scope the flag to GCC mkspecs (linux-g++, win32-g++, ...). (#3662)
*-g++* {
    QMAKE_CXXFLAGS += -ftree-vectorize
}
```

On "safe to default-enable" — I verified it rather than just reasoning it. Same inputs through a pragma-forced-scalar loop and the vectorized loop, built with `-O2 -ftree-vectorize` (no `-march`, so SSE2 / no FMA, matching the shipped target), `memcmp` of the float accumulators: byte-for-byte identical, mono and stereo. No `-ffast-math`, and the loop vectorizes over the sample index, so every output sample keeps its exact scalar op order.

mcfnord:
Follow-up: the isolated mix-loop speedup above, now as **end-to-end whole-server CPU under load**.

Same 1-core harness (server pinned to one core with `taskset`, N tone-generating clients, `perf stat` task-clock, stock `-O2` vs this PR interleaved, 4 reps each):

| clients | `-O2` | `+ -ftree-vectorize` | saving |
|---|---|---|---|
| 20 | 0.330 | 0.313 | **5.2%** |
| 40 | 0.729 | 0.625 | **14.3%** |

(CPU = fraction of one core; ±1 sd ≤ 0.007, so the 40-client gap is well outside run-to-run spread.) The saving **grows with client count** — as expected, since the vectorized loops are the O(N²) mixing plus the O(N) `Float2Short` conversions, so their share of server CPU rises with load. A single core saturates past ~50 clients, so measuring higher counts needs multi-core / `-T`; but the direction is what matters — this helps most exactly where it's needed, on busy servers. Output stays bit-identical (the scalar-vs-vectorized `memcmp` I noted earlier).

mcfnord:
MY LLM WROTE:

The stated blocker on this PR is "Build with multiple compilers." Here is that data, plus a correction we owe @dingodoppelt.

Method: two worktrees differing by exactly one commit — A = `7f559949` (parent), B = `d8567ab9` (this PR). Full qmake builds, flags read out of the generated Makefile rather than hand-constructed, so this measures the build the PR actually produces.

**Builds clean on every compiler tested, with no new warnings:**

| Compiler | Target | Builds | New warnings in B |
|---|---|---|---|
| gcc 13.3.0 | x86-64 Linux | ✓ | none (1 pre-existing `-Wmaybe-uninitialized` in both A and B) |
| gcc 14.2.0 | aarch64 Linux | ✓ | none |
| Apple clang 14.0.0 | x86-64 macOS | ✓ | none |
| MSVC | Windows (CI) | ✓ | `D9002: ignoring unknown option '-ftree-vectorize'`, once per TU (97 and 104 in the two Windows jobs) |
| clang | iOS, Android NDK (CI) | ✓ | none |
| gcc | Linux arm64, armhf (CI) | ✓ | none |

**The flag reaches the real compile line.** In the generated Makefile, `-ftree-vectorize` appears on the `src/server.cpp` command in B and not in A, on all three locally built compilers.

**What it actually changes, per compiler.** Counting vectorized loops in `MixEncodeTransmitData` (`src/server.cpp` 953–1243) via `-fopt-info-vec-optimized`, with `--param vect-epilogues-nomask=0` so each source loop is counted once — GCC 12+ vectorizes the epilogue too and reports it as a second line at half vector width:

- gcc 13.3 x86-64: A = 0 loops, **B = 8**
- gcc 11.5 x86-64: A = 0 loops, **B = 8** — the same eight source loops, so Ubuntu 22's default GCC gets the full main-loop benefit
- gcc 14.2 aarch64: A = 0 loops, **B = 8** — the same eight source lines again
- `-O3` (`dynamic` cost model) on x86-64: the same 8. Going to `-O3` buys nothing further here.

ARM gets the same eight loops. One differs in width: at `src/server.cpp:1136`, the stereo/interleaved branch of the gain mix, gcc 14.2 picks 8-byte vectors where x86-64 uses 16 — `Preferring vector mode V4HI to vector mode V8HI` in its own vectorizer dump. The mono branch eight lines above gets the full width. Confirmed in the binary, not just in the compiler's log: packed-NEON FP instructions rise from 28 to 68.

**On clang the flag is a genuine no-op, and we checked that rather than assuming it.** On Apple clang 14, `server.o` is byte-identical between A and B (`0ffaa22dd167216d287a9734f3bd150d` both). Of 203 objects, 201 are byte-identical; the 2 that differ are Qt resource blobs whose `otool -tV` disassembly is identical — the differing bytes are embedded resource data, not code. Clang accepts the flag silently (no "unknown argument" diagnostic) because it already auto-vectorizes at `-O2`: that same `server.o` contains 24 `addps` and 16 `mulps` in both builds. So this PR is a GCC-only change in effect, harmless elsewhere.

**Correction: our earlier `-O2` claim was wrong.** On 2026-07-19 we told @dingodoppelt that "on g++ 13, plain `-O2` definitely does not enable loop vectorization — if your local builds vectorize at `-O2`, something else in your flags is adding it." That is incorrect. GCC 12+ enables `-ftree-loop-vectorize` by default at `-O2` with the `very-cheap` cost model, which refuses loops needing a scalar epilogue. What this PR's flag does is move the cost model to `cheap`, permitting those loops. His build was behaving exactly as documented; our explanation was the thing at fault. Apologies for the noise.

**One scope note:** `QMAKE_CXXFLAGS` is C++-only, so the flag does not reach the bundled Opus C sources (0 occurrences on any `.c` build line). Whether `QMAKE_CFLAGS` would be worth adding for Opus is a separate question we have not measured.

Reproduction, on any GCC ≥ 12:

```
git worktree add A 7f559949 && git worktree add B d8567ab9
for V in A B; do (cd $V && qmake Jamulus.pro && make -j"$(nproc)" \
  2>&1 | tee build-$V.log); done
# vectorized-loop counts for the mixing function:
for V in A B; do (cd $V && g++ -c -O2 $( [ $V = B ] && echo -ftree-vectorize ) \
  -fopt-info-vec-optimized --param vect-epilogues-nomask=0 ... src/server.cpp 2>&1 \
  | awk -F: '$2>=953 && $2<=1243' | grep -c "loop vectorized"); done
```

(substitute the include/define flags from the `server.o` line in the generated Makefile — that is how these numbers were taken)

pljones:
Android clang presumably comes under the "no effect" applied to clang?  And is _anyone_ running Jamulus on a device with so little storage unrolling would be an issue (i.e. "packed-NEON FP instructions rise from 28 to 68")?

mcfnord:
MY LLM WROTE:

**Q1 — yes, Android clang is a no-op.** NDK r26d (`clang-17`, `aarch64-linux-android34`) on a standalone extract of the four `MixEncodeTransmitData` loops: `-O2` vs `-O2 -ftree-vectorize` = **byte-identical object** (md5 `16506ee2…`). The flag draws no diagnostic, and `-O2` alone already emits 38 packed-NEON FP instructions — it auto-vectorizes without the flag. Same as the Apple-clang `server.o` I posted earlier.

**Q2 — negligible, near-nil where storage is tightest.** The `28→68` is instruction *count* in one function, not size (vectorization mostly *replaces* scalar ops). Whole-binary delta, stripped:

| Compiler / arch | delta |
|---|---|
| gcc 14.2 aarch64 | **+72 B** |
| Apple clang 14 x86-64 | **+0 B** |
| gcc 13.3 x86-64 | +12 KB (+0.31%) |

On ARM and any clang target it's 72 bytes or nothing. No device this tips over.

pljones (approved):

ann0see:
So we should narrow this to gcc only?

softins:
I just built the latest `main` on Windows with MSVC 2019, and am getting the following warning:

```
Command line warning D9002 : ignoring unknown option '-ftree-vectorize'
```

pljones:
> I just built the latest `main` on Windows with MSVC 2019, and am getting the following warning:
> 
> ```
> Command line warning D9002 : ignoring unknown option '-ftree-vectorize'
> ```

Yes: https://github.com/jamulussoftware/jamulus/pull/3739#issuecomment-5038945793

PR #3741, opened 2026-06-17
Refactor server chat handling by extracting methods

dingodoppelt:

This refactors `CServer::CreateAndSendChatTextForAllConChannels` into three functions for sending messages.
The extracted methods
1. create the full chat message including timestamp and name of the sending client,
2. iterate over all possible channel numbers to
3. send the message text to a channel

As it is now possible to call the function to send a message to a channel directly, the channel number needs to be boundary checked.  
A function `InRange()` was introduced into `util.h` for that purpose.

CHANGELOG: Refactor server chat handling by extracting methods

**Context: Fixes an issue?**

Not yet ;)

No

Ready for review

Review, Approval

## Checklist

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1306,7 +1306,7 @@ void CServer::SendChatTextToAllConChannels ( const QString& strChatText )
     // Send chat text to all connected clients ---------------------------------
     for ( int i = 0; i < iMaxNumChannels; i++ )
     {
-        SendChatTextToConChannel ( i, strChatText );
+        if ( vecChannels[i].IsConnected() ) vecChannels[i].CreateChatTextMes ( strChatText );
     }
 }
 
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1298,13 +1298,24 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int iCurChanID, con
                                          ChanName.toHtmlEscaped() + "</b></font> " + strChatText.toHtmlEscaped();
 
     // Send chat text to all connected clients ---------------------------------
+    SendChatTextToAllConChannels ( strActualMessageText );
+}
+
+void CServer::SendChatTextToAllConChannels ( const QString& strChatText )
+{
+    // Send chat text to all connected clients ---------------------------------
     for ( int i = 0; i < iMaxNumChannels; i++ )
     {
-        if ( vecChannels[i].IsConnected() )
-        {
-            // send message
-            vecChannels[i].CreateChatTextMes ( strActualMessageText );
-        }
+        SendChatTextToConChannel ( i, strChatText );
+    }
+}
+
+void CServer::SendChatTextToConChannel ( const int iCurChanID, const QString& strChatText )
+{
+    if ( MathUtils::InRange<int> ( iCurChanID, 0, iMaxNumChannels - 1 ) && vecChannels[iCurChanID].IsConnected() )
+    {
+        // send message
+        vecChannels[iCurChanID].CreateChatTextMes ( strChatText );
     }
 }
 
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1306,7 +1306,10 @@ void CServer::SendChatTextToAllConChannels ( const QString& strChatText )
     // Send chat text to all connected clients ---------------------------------
     for ( int i = 0; i < iMaxNumChannels; i++ )
     {
-        if ( vecChannels[i].IsConnected() ) vecChannels[i].CreateChatTextMes ( strChatText );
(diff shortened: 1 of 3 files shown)


src/util.h
```diff
@@ -1223,6 +1223,13 @@ class MathUtils
             return powf ( 10.0f, ( fInValueRange0_1 - 1.0f ) * AUD_MIX_FADER_RANGE_DB / 20.0f );
         }
     }
+
+    // return true if a value is inside a given range, otherwise false
+    template<typename T>
```
ann0see, line 1228: Seems a bit overkill for me to use a template IMO.

dingodoppelt: Since there is no `==` involved I thought it made sense to use a template here to make it work with floats as well. Is there any performance or stability penalty for using a template? If so I'm happy to change this into a fixed type. Otherwise I'd like to keep it that way because it is a utility function after all.  
If we really wanted to make this exclusive to the chat message refactor we could just get rid of this whole function and check in the one place in the code it gets actually used. (I thought I'd be smart here and make it available in case somebody else needs it in the future but my clairvoyant powers might have failed me here ;)

pljones: Templates in C++ are compile-time, as far as I know, so the compiler should be as able to optimise its own output as it would hand-coded classes.

src/util.h
```diff
@@ -1223,6 +1223,13 @@ class MathUtils
...
+    static inline bool InRange ( T value, T lower, T upper )
```
pljones, line 1229: The Java/C# programmer in me wants this as an extension to `Comparable<T>`, so that
```
(<Comparable<Int>>32).inRange(0, 100)
```
returns true.  Don't know if C++ can do better than what you've got here, though.  (What you _should_ do is constraint `T` somehow, really.)

src/util.h
```diff
@@ -1223,6 +1223,13 @@ class MathUtils
             return powf ( 10.0f, ( fInValueRange0_1 - 1.0f ) * AUD_MIX_FADER_RANGE_DB / 20.0f );
         }
     }
+
+    // return true if a value is inside a given range, otherwise false
```
pljones: a) Use a doccomment
b) Make it clear it's inclusive of the lower and upper bounds
```suggestion
    // return true if a value is within the lower and upper bounds (inclusively), otherwise false
```

src/server.cpp
```diff
@@ -1297,14 +1297,25 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int iCurChanID, con
...
+{
     // Send chat text to all connected clients ---------------------------------
     for ( int i = 0; i < iMaxNumChannels; i++ )
     {
-        if ( vecChannels[i].IsConnected() )
-        {
-            // send message
-            vecChannels[i].CreateChatTextMes ( strActualMessageText );
-        }
+        SendChatTextToConChannel ( i, strChatText );
```
pljones: At this point, it's guaranteed that `MathUtils::InRange` will return true, so I think it's worth removing the overhead.
```suggestion
        if ( vecChannels[i].IsConnected() ) vecChannels[i].CreateChatTextMes ( strChatText );
```

dingodoppelt: I was wondering, too and decided to not duplicate code but I think you're right here.

src/util.h
dingodoppelt: > Don't know if C++ can do better than what you've got here, though. (What you _should_ do is constraint `T` somehow, really.)

This function cannot be called from the outside so I think we should be safe here.  
The first placed I searched for range validating functions was the Jamulus source code, then Qt docs because I thought there *must* be something there. Then I wrote this :) If we don't need it and it might cause issues I'm happy to remove it and do the sanity checks locally.

pljones (approved):

src/server.cpp
```diff
@@ -1297,17 +1297,31 @@ void CServer::CreateAndSendChatTextForAllConChannels ( const int iCurChanID, con
     const QString strActualMessageText = "<font color=\"" + sCurColor + "\">(" + QTime::currentTime().toString ( "hh:mm:ss AP" ) + ") <b>" +
                                          ChanName.toHtmlEscaped() + "</b></font> " + strChatText.toHtmlEscaped();
 
+    // Send chat text to all connected clients ---------------------------------
```
ann0see, line 1300: Well, I don't like the `---` but i guess that's style.

ann0see (approved):

src/server.cpp
dingodoppelt: Copied and pasted this reluctantly ;)

PR #3742, opened 2026-06-18
Add CLM protocol to get Welcome Message

softins:

Adds two CLM protocol messages: one to request the server's Welcome Message, and the other to respond with it.

CHANGELOG: Server: Add protocol for requesting a server's Welcome Message without connecting

**Context: Fixes an issue?**

Provides a way for a client to fetch the Welcome Message without needing to start an audio connection.
It will be very useful for Jamulus Explorer, and could be used in the future by a Jamulus client.

Not at present. It is documented in protocol.cpp. If in the future a Jamulus client uses the feature to fetch a Welcome Message, that usage would need to be documented.

Ready and tested.

Just review

## Checklist

### DIFF
--- a/src/protocol.cpp
+++ b/src/protocol.cpp
@@ -472,6 +472,18 @@ CONNECTION LESS MESSAGES
 
     note: does not have any data -> n = 0
 
+
+- PROTMESSID_CLM_WELCOME_MESSAGE: Server welcome message
+
+    +------------------+--------------------------------------+
+    | 2 bytes number n | n bytes UTF-8 string welcome message |
+    +------------------+--------------------------------------+
+
+
+- PROTMESSID_CLM_REQ_WELCOME_MESSAGE: Request server welcome message
+
+    note: does not have any data -> n = 0
+
 */
 
 #include "protocol.h"
@@ -963,6 +975,10 @@ void CProtocol::ParseConnectionLessMessageBody ( const CVector<uint8_t>& vecbyMe
 
     case PROTMESSID_CLM_REQ_SERVER_FEATURES:
         EvaluateCLReqServerFeaturesMes ( InetAddr );
+        break;
+
+    case PROTMESSID_CLM_REQ_WELCOME_MESSAGE:
+        EvaluateCLReqWelcomeMessageMes ( InetAddr );
         break;
     }
 }
@@ -2653,6 +2669,35 @@ void CProtocol::CreateCLServerFeaturesMes ( const CHostAddress& InetAddr, const 
     PutValOnStream ( vecData, iPos, iFeatures, 4 );
 
     CreateAndImmSendConLessMessage ( PROTMESSID_CLM_SERVER_FEATURES, vecData, InetAddr );
+}
+
+bool CProtocol::EvaluateCLReqWelcomeMessageMes ( const CHostAddress& InetAddr )
+{
+    // invoke message action
+    emit CLReqWelcomeMessage ( InetAddr );
+
+    return false; // no error
+}
+
+void CProtocol::CreateCLWelcomeMessageMes ( const CHostAddress& InetAddr, const QString strWelcomeMessage )
+{
+    int iPos = 0; // init position pointer
+
+    // convert chat text string to utf-8
+    const QByteArray strUTF8WelcomeMessage = strWelcomeMessage.toUtf8();
+
+    const int iStrUTF8Len = strUTF8WelcomeMessage.size(); // get utf-8 str. size / string
+
+    // size of message body
+    const int iEntrLen = 2 + iStrUTF8Len; // utf-8 str. size / string
+
+    // build data vector
+    CVector<uint8_t> vecData ( iEntrLen );
+
+    // chat text
+    PutStringUTF8OnStream ( vecData, iPos, strUTF8WelcomeMessage );
+
+    CreateAndImmSendConLessMessage ( PROTMESSID_CLM_WELCOME_MESSAGE, vecData, InetAddr );
 }
 
 /******************************************************************************\
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -108,6 +108,8 @@
 #define PROTMESSID_CLM_RED_SERVER_LIST        1018 // reduced server list
 #define PROTMESSID_CLM_SERVER_FEATURES        1019 // server features message
 #define PROTMESSID_CLM_REQ_SERVER_FEATURES    1020 // request server features
+#define PROTMESSID_CLM_WELCOME_MESSAGE        1021 // server welcome message
+#define PROTMESSID_CLM_REQ_WELCOME_MESSAGE    1022 // request server welcome message
 
 // special IDs
 #define PROTMESSID_SPECIAL_SPLIT_MESSAGE 2001 // a container for split messages
@@ -180,6 +182,7 @@ public:
     void CreateCLChannelLevelListMes ( const CHostAddress& InetAddr, const CVector<uint16_t>& vecLevelList, const int iNumClients );
     void CreateCLRegisterServerResp ( const CHostAddress& InetAddr, const ESvrRegResult eResult );
     void CreateCLServerFeaturesMes ( const CHostAddress& InetAddr, const uint32_t iResult );
+    void CreateCLWelcomeMessageMes ( const CHostAddress& InetAddr, const QString strWelcomeMessage );
 
     static bool ParseMessageFrame ( const CVector<uint8_t>& vecbyData,
                                     const int               iNumBytesIn,
@@ -308,6 +311,7 @@ protected:
     bool EvaluateCLChannelLevelListMes ( const CHostAddress& InetAddr, const CVector<uint8_t>& vecData );
     bool EvaluateCLRegisterServerResp ( const CHostAddress& InetAddr, const CVector<uint8_t>& vecData );
     bool EvaluateCLReqServerFeaturesMes ( const CHostAddress& InetAddr );
+    bool EvaluateCLReqWelcomeMessageMes ( const CHostAddress& InetAddr );
 
     int iOldRecID;
     int iOldRecCnt;
@@ -376,4 +380,5 @@ signals:
     void CLChannelLevelListReceived ( CHostAddress InetAddr, CVector<uint16_t> vecLevelList );
     void CLRegisterServerResp ( CHostAddress InetAddr, ESvrRegResult eStatus );
     void CLReqServerFeatures ( CHostAddress InetAddr );
+    void CLReqWelcomeMessage ( CHostAddress InetAddr );
 };
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -294,6 +294,8 @@ CServer::CServer ( const int          iNewMaxNumChan,
 
     QObject::connect ( &ConnLessProtocol, &CProtocol::CLReqServerFeatures, this, &CServer::OnCLReqServerFeatures );
 
+    QObject::connect ( &ConnLessProtocol, &CProtocol::CLReqWelcomeMessage, this, &CServer::OnCLReqWelcomeMessage );
+
     QObject::connect ( &ServerListManager, &CServerListManager::SvrRegStatusChanged, this, &CServer::SvrRegStatusChanged );
 
     QObject::connect ( &JamController, &recorder::CJamController::RestartRecorder, this, &CServer::RestartRecorder );
@@ -527,6 +529,12 @@ void CServer::OnCLReqServerFeatures ( CHostAddress RecHostAddr )
 
     // Create and send the message
     ConnLessProtocol.CreateCLServerFeaturesMes ( RecHostAddr, iFeatures );
+}
+
+void CServer::OnCLReqWelcomeMessage ( CHostAddress RecHostAddr )
+{
+    // Create and send the message
+    ConnLessProtocol.CreateCLWelcomeMessageMes ( RecHostAddr, strWelcomeMessage );
 }
 
 void CServer::OnServerFull ( CHostAddress RecHostAddr )
--- a/src/server.h
+++ b/src/server.h
@@ -401,6 +401,8 @@ public slots:
 
     void OnCLReqServerFeatures ( CHostAddress InetAddr );
 
+    void OnCLReqWelcomeMessage ( CHostAddress InetAddr );
+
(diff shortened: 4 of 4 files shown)


dingodoppelt:
Looks perfectly fine (by the code, that is. haven't tested, yet)  
Could it be useful if we somehow logged requests like this one (or server features cl message)?  
I just recently noticed some client was connecting to my public server every 30 seconds to do whatever. I only noticed because the connection was logged.

ann0see:
Meaning we should probably change the client functionality in future to request the server welcome message instead of the server pushing it.

softins:
> Meaning we should probably change the client functionality in future to request the server welcome message instead of the server pushing it.

I don't think there is any point in doing that. The server still needs to be able to push chat messages to the client. The welcome message is just the first chat message to be pushed after connecting. I don't see that as a problem to be solved.

What may possibly be useful would be providing some way within the connect dialog to check the welcome message of a server before deciding whether to connect to it.

mcfnord:
Showing the welcome message from the connect dialog would be kinda neat. 

I've made a welcome message replacement that gives an LLM-derived welcome to joiners. If you made a welcome message visible before joining, I'd make that an LLM-derived general welcome message, too. In the coolest scenario, the welcome message could reveal what the people on the jam are playing now, or whether they're a regular crowd, stuff like that.

@dingodoppelt  I think I have a sample cycle on a 29-second cycle, and it's pretty new code, but I think it's connectionless so it wouldn't be a true connection. I think this code is a connection metadata request (names, instruments, countries, jitter). This is helpful to maintain GUI responsiveness. I could probably change the rate to 55-seconds. My systems run on 1-minute chunks so it's useful to have a sample per each minute. Immediately after the metadata request I ask for an audio nibble request for those servers that support this. I have to re-confirm the metadata before gathering the channel nibbles to make sure the channels haven't changed.

I've also re-implemented servers.php to sample metadata more rapidly during prime time and more slowly during slow hours, but I don't think those intervals are ~30s.

pljones:
> Meaning we should probably change the client functionality in future to request the server welcome message instead of the server pushing it.

No - it's the server administrator's decision -- not the client user's. Remember, we still have the client being required to accept the server terms as a server option: the server can refuse the connection at that point.  The welcome message forms part of that exchange.

src/protocol.cpp
```diff
@@ -472,6 +472,18 @@ CONNECTION LESS MESSAGES
 
     note: does not have any data -> n = 0
 
+
+- PROTMESSID_CLM_WELCOME_MESSAGE: Server welcome message
```
pljones, line 476: Can I suggest this is loosened a bit.

Currently, the welcome message is sent to the chat window because it looks like a chat message sent from the server to that client.

I'd like the same protocol here with the exception that it's connectionless.  The request should come in and, should the server choose to respond (or support the request), the response should go out "looking like a chat message".

I'm hoping that way there can be more reuse of code than the approach here.

(Yes, it still needs two new protocol messages.)

softins: Not sure I understand what you're getting at here. The bit `FS_HAS_WELCOME_MESSAGE` in server features is true if the server operator gave something to `-w` on invocation, i.e. the welcome message is non-empty. The intention of this message is just to tell the caller what those contents are. I'm not sure what else it could be doing.

And a chat message just looks like `| 2 bytes number n | n bytes UTF-8 string |`, which is exactly what the contents of this message is too.

pljones: I'm not talking about server features.

The server welcome message sent when a client connects to the server is sent as a chat message.

The code to do that exists.

You're writing completely new code to do exactly the same thing here.

Then saying, no, it's not a chat message, it's a welcome message.

That unnecessarily constrains the future use of the protocol message.

softins: > The server welcome message sent when a client connects to the server is sent as a chat message.

Yes, because that's how it is displayed to the client user

> The code to do that exists.

Yes, in the connected-mode `PROTMESSID_CHAT_TEXT`

> You're writing completely new code to do exactly the same thing here.

That's because connected mode is not appropriate here. There is no existing CLM message that encodes just a variable-length string. And the amount of extra code is trivially small, compared to the size of the code base, so I'm not sure why you think it's a problem.

> Then saying, no, it's not a chat message, it's a welcome message.

Well that's true. The request was specifically "send me the contents of your welcome message".

> That unnecessarily constrains the future use of the protocol message.

It's not as if protocol messages are in short supply! I do feel it's picking nits for the sake of it, which can be rather exasperating.

But maybe you are just suggesting that we call it `CLM_CHAT_TEXT` instead of `CLM_WELCOME_MESSAGE`? I suppose in that case we could use it for sending potentially (but rarely) large welcome message and chat texts over any long-lived TCP connection, in the same way as I propose to use `CLM_CONN_CLIENTS_LIST` instead of the connected-mode `CONN_CLIENTS_LIST`. That would address the potential point alluded to in your [other comment](https://github.com/jamulussoftware/jamulus/pull/3636#discussion_r3446313100)

pljones: Yes, that's what I'm getting at.

softins: > Yes, that's what I'm getting at.

Well I disagree fundamentally with the objective. And I will explain why. Although it is becoming _very_ tedious always having to explain and defend my position at great length, after putting a _lot_ of thought into my submissions. This PR could have been in 4.0.0beta1 if it weren't for this objection. And it wouldn't be the first PR I have ever closed and abandoned due to unhelpful and intransigent feedback from @pljones. Although in this case, I feel it is too important a feature to abandon.

- I don't believe there is any problem with having two different protocol messages having a similar or same message format, _if they mean different things_. In this case there would be no problem in having two messages, `CLM_WELCOME_MESSAGE` and `CLM_CHAT_TEXT`, because they are intended for different purposes:

  - `CLM_WELCOME_MESSAGE` would be used precisely to respond to `CLM_REQ_WELCOME_MESSAGE`, and is unambiguously "this is the welcome message that the server operator has configured". It is intended just to answer that query, and would never contain anything else. It isn't something that needs to be defended against "unnecessary contstraint on future use".
  - `CLM_CHAT_TEXT` (to be defined in the future, potentially as part of the TCP PR), is more general, meaning "this is to be displayed to the user in their chat dialog", in the same way as connected-mode `CHAT_TEXT` is. This is a different semantic, and is _not_ negated by the fact that the welcome message happens (by design) to be the first chat text sent by the server to a new client. The difference is context.
  - Using a `CLM_CHAT_TEXT` to respond to a `CLM_REQ_WELCOME_MESSAGE` is an unnecessary conflation of two concepts, and requires the message receiver potentially to interpret it in context.

- So I maintain that `CLM_WELCOME_MESSAGE` as proposed in this PR is the correct way to do it, and I don't propose to change that.

I would be interested in more constructive feedback and review from @ann0see and @dingodoppelt.

ann0see: Please be constructive...

I think it's mainly a style question again?
We need to be clear about *why* we add all those new messages.

* Until now, protocol messages were almost always for exchanging information between Jamulus Client and Jamulus Server only. This changes with 4.0.0. The PR by @dingodoppelt introducing SERVER_FEATURES might be used by external tools too - better let's say the reason it was implemented was not a client need. 

Peter (and me) would probably prefer to use this message in the client some day instead of introducing too many new protocol messages since it feels redundant.

Here, we have a very clear feature specification:
Get the welcome message from a server without connecting. Again from the viewpoint of an external tool (Explorer). For this, it is enough to have a specific response just containing the welcome message.
If we decide that this is *just* the welcome message, I agree `CLM_CHAT_TEXT` would be overkill. We don't know *yet* if there is any feature which will require the server to send a connection-less chat message which for now would make `CLM_WELCOME_MESSAGE` enough. 

Curently, the server has no notion of what a chat message vs a welcome message is (it just does some html based hackery which is unclean engineering).

For now, i would decide that we can go with either definition - however, from a clean semantic meaning, I'd likely go with `CLM_WELCOME_MESSAGE` and define the scope strictly.

I hope I got this right then and you both can agree?

src/protocol.cpp
```diff
@@ -2655,6 +2671,35 @@ void CProtocol::CreateCLServerFeaturesMes ( const CHostAddress& InetAddr, const
...
 
+bool CProtocol::EvaluateCLReqWelcomeMessageMes ( const CHostAddress& InetAddr )
+{
+    // invoke message action
+    emit CLReqWelcomeMessage ( InetAddr );
+
+    return false; // no error
+}
+
+void CProtocol::CreateCLWelcomeMessageMes ( const CHostAddress& InetAddr, const QString strWelcomeMessage )
```
ann0see, line 2682: What you could do would be to point to the respective place on the server where the connected welcome message would be sent.

src/protocol.cpp
dingodoppelt: +1 for `CLM_WELCOME_MESSAGE`.  
I guess it makes sense to think about turning this new kind of message into something more in line with existing code. We currently only have messages for exchange between a client and a server and from a stylistic point of view I can see why PRs that don't exactly fit that framework are more difficult to review because more consideration goes into every personal review.  
I think we might need to acknowledge that the Jamulus explorer has been "exploiting" that design at least partly for years now. Since there is no category for protocol messages that are to be exchanged with anything else than a client we should introduce one with a clear scope: External communication to non-jamulus clilents only. Therefore it'd be possible and probably necessary to add logging to these messages (`CLM_SERVER_FEATURES`) which would enable admins to use stuff like `fail2ban`.

softins: > Please be constructive...

I _always_ try to be, as I hope you know. I just reached a frustration threshold. If I were just going to rant, I wouldn't still include reasoned technical detail in support of my position.

> I think it's mainly a style question again? We need to be clear about _why_ we add all those new messages.
> 
> * Until now, protocol messages were almost always for exchanging information between Jamulus Client and Jamulus Server only. This changes with 4.0.0. The PR by @dingodoppelt introducing SERVER_FEATURES might be used by external tools too - better let's say the reason it was implemented was not a client need.
> 
> Peter (and me) would probably prefer to use this message in the client some day instead of introducing too many new protocol messages since it feels redundant.
> 
> Here, we have a very clear feature specification: Get the welcome message from a server without connecting. Again from the viewpoint of an external tool (Explorer). For this, it is enough to have a specific response just containing the welcome message. If we decide that this is _just_ the welcome message, I agree `CLM_CHAT_TEXT` would be overkill. We don't know _yet_ if there is any feature which will require the server to send a connection-less chat message which for now would make `CLM_WELCOME_MESSAGE` enough.

I can see some functional user benefit in a future version of the Jamulus client, such as in the Connect Dialog, allowing the user to request the server's welcome message in a connectionless way _before_ the user decides whether to connect. Such as providing a button, or a right-click menu on a server entry.

> Curently, the server has no notion of what a chat message vs a welcome message is (it just does some html based hackery which is unclean engineering).

That's right, the fundamental question is what the response _means_ and what client should _do_ with it:
- If it receives a `CLM_WELCOME_MESSAGE`, it knows that is it actually a welcome message, and to display it in some way to a user who requested it, in a connectionless way (e.g. a popup).
- If it receives a `CHAT_TEXT` or a (possibly, in the future) `CLM_CHAT_TEXT`, it knows to display it in the user's chat window for the active connection, and that it could contain anything the server wants to send (either initial welcome message to a new client, chat from another user, or a message from an RPC client run by the server operator).

> For now, i would decide that we can go with either definition - however, from a clean semantic meaning, I'd likely go with `CLM_WELCOME_MESSAGE` and define the scope strictly.
> 
> I hope I got this right then and you both can agree?

I hope so. I still maintain that this PR is fine in its current form. Logging for both it and the server features request could be added later, if needed.

src/protocol.cpp
softins: No, because connected mode is different from CLM mode. The welcome message is stored as a string within the `CServer` class, and it is trivial to encode and send it as a CLM.

src/protocol.cpp
softins: > +1 for `CLM_WELCOME_MESSAGE`. I guess it makes sense to think about turning this new kind of message into something more in line with existing code. We currently only have messages for exchange between a client and a server and from a stylistic point of view I can see why PRs that don't exactly fit that framework are more difficult to review because more consideration goes into every personal review. I think we might need to acknowledge that the Jamulus explorer has been "exploiting" that design at least partly for years now. Since there is no category for protocol messages that are to be exchanged with anything else than a client we should introduce one with a clear scope: External communication to non-jamulus clilents only. Therefore it'd be possible and probably necessary to add logging to these messages (`CLM_SERVER_FEATURES`) which would enable admins to use stuff like `fail2ban`.

The client-server interface with Jamulus is defined to be the protocol. The Explorer back-end `jamulus-php` is _designed_ just to look to the server exactly like a Jamulus client. I don't see any need to change that. I wouldn't want to have to add tagging to protocol messages to say "this is from an actual Jamulus client" vs "this is from an external tool".

I think any logging needs to be thought about further, and shouldn't hold up this PR in the meantime.

dingodoppelt: > The client-server interface with Jamulus is defined to be the protocol. The Explorer back-end `jamulus-php` is _designed_ just to look to the server exactly like a Jamulus client. I don't see any need to change that.

That's what I meant by "exploit" (probably not the right word). It uses client mechanics while it really is not a client. It is not only `jamulus-php` using this approach.

>  I wouldn't want to have to add tagging to protocol messages to say "this is from an actual Jamulus client" vs "this is from an external tool".

Why would we need tagging here? I'm suggesting not to use the existing protocol messages and therefore introducing a new category of protocol messages: External _non_ Jamulus clients.

> 
> I think any logging needs to be thought about further, and shouldn't hold up this PR in the meantime.

True as it would apply to the newly introduced `CLM_SERVER_FEATURES` as well.  
Still I feel it is important at least for admins to be able to monitor their servers. The silent connections go unnoticed by the users and most admins who don't actively look through their logs. I'm not suggesting logging existing protocol messages, just external communication.

softins: > Why would we need tagging here? I'm suggesting not to use the existing protocol messages and therefore introducing a new category of protocol messages: External _non_ Jamulus clients.

I'm not convinced of the benefit of that. But either way, that is a separate, and much bigger, conversation.

pljones (approved):

softins:
Thank you!

ann0see (approved):
Looks ok - however I didn't test it.

Issue #3613, opened 2026-01-29
Accessibility: Chat History is not properly accessible with Screen Reader


chigkim:
Not sure whether this is another Qt bug with QTextBrowser, but VoiceOver cannot navigate the chat history (txvChatWindow) reliably, especially when there is a long message.
When you press the up or down arrow in the chat history and VoiceOver encounters a long entry (for example, a server intro message), it seems to treat the text as multiple wrapped lines, but it only reads the first few words in the beginning. After that, navigation through the rest of the history becomes broken.
Is it possible to present each chat entry as a separate accessibility element instead of one large block of text within the QTextBrowser? That might resolve the issue.

chigkim:
Cool, when you guys redesign, please let me know. Happy to test for accessibility with screen readers.

mcfnord:
@chigkim Claude and I have implemented possible fix for this problem in a client test. I've run the client on Windows and I think the new user experience is adequate, and arguably easier for the typical use case. 

You can test a [Mac build download here](https://github.com/mcfnord/jamulus/actions/runs/27832372114/artifacts/7752609847).  Does it work for VoiceOver?

PR #3748, opened 2026-06-20
chatdlg: replace QTextBrowser with QListView for screen reader accessibility

mcfnord:
Replaces \`QTextBrowser\` with \`QListView\` + \`QStyledItemDelegate\` in the chat dialog so that each message is a discrete item in the OS accessibility tree.

With \`QTextBrowser\`, VoiceOver (and other screen readers that consume \`QAccessible\`) sees the entire chat history as a single text block. With \`QListView\`, each appended message becomes a \`QAccessible::ListItem\`, letting the user arrow-key through individual messages.

\`ChatDelegate\` renders each item as HTML via \`QTextDocument\`, preserving the existing rich-text formatting. Link clicks are detected in \`eventFilter\` by installing a filter on the viewport (mouse events in \`QListView\` go to the viewport child widget, not the view itself) and using \`QTextDocument::documentLayout()->anchorAt()\` to hit-test the click position. Hovering over a link shows a pointing-hand cursor. Clicking routes through the existing \`OnAnchorClicked\` confirmation dialog. Right-click or Ctrl+C copies the selected message's plain text to the clipboard.

Link clicks have been confirmed working on Windows, perhaps on Linux. Does VoiceOver read individual messages as expected on macOS?

**Fixes an issue?**

Addresses #3613.

No documentation changes needed.

Builds and runs. Not tested on macOS. does VoiceOver navigate individual messages? 

I don't see pointer cursor on hover; probably worth investigating.

## Checklist

- [ ] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I tested my code and it does what I want... but does it do things I don't want?
- [ ] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#source-code-consistency)


pljones:
Thanks - we've been looking to get more of the assistive side of Jamulus fixed.

Could you also provide screenshots (ideally with some rich HTML text - and welcome messages) and a comparative walk through of use with the existing interface - not just for screen-reader use but generally.

-- Peter

mcfnord:
I found a problem with this approach (links not clickable) but Claude says it's fixable. I agree that we'd need to assure any visual changes are unimportant. My hope is that the only useability difference is arguably an improvement: you can copy the whole line rather than pieces of it, but no copy of multiple lines anymore. This is a more common pattern in tools like Slack.

src/chatdlg.cpp
```diff
@@ -65,16 +43,23 @@ CChatDlg::CChatDlg ( QWidget* parent ) : CBaseDlg ( parent, Qt::Window ) // use
...
+    // clear edit line
     edtLocalInputText->clear();
 
-    // we do not want to show a cursor in the chat history
-    txvChatWindow->setCursorWidth ( 0 );
-
     // set a placeholder text to make sure where to type the message in (#384)
     edtLocalInputText->setPlaceholderText ( tr ( "Type a message here" ) );
 
+    // Set up the list model and delegate for accessible per-message rows ------
```
ann0see, line 52:
```suggestion
    // Set up the list model and delegate for accessible per-message rows
```

Check for AI artifacts like this.

ann0see (commented):
Tested locally on linux and there is a slight difference. I'd be interested what happens if you send broken html form the server side (e.g. a broken welcome message).

mcfnord:
> Tested locally on linux and there is a slight difference. I'd be interested what happens if you send broken html form the server side (e.g. a broken welcome message).

I want a Mac user to test with VoiceOver!

dingodoppelt:
Do @chigkim or @OnjLouis have anything to add to this maybe?

ann0see:
I tried it on macOS but could not succeed to tab inside the chat message. It also did not speak once i clicked onto the message. so it didn't fix it unfortunately. 

I think that we probably need to resructure the UI anyway. It should probably be restructured  such that it works like modern messengers.

mcfnord:
Updated with two fixes targeting the macOS VoiceOver silence:

**Fix 1 — `Qt::AccessibleTextRole` (root cause of "didn't speak")**

`Qt::DisplayRole` stores the raw HTML string. VoiceOver reads this role directly — it sees angle-bracket markup and either reads garbled text or says nothing. The fix: after HTML processing, extract plain text with `QTextDocument::toPlainText()` and store it as `Qt::AccessibleTextRole` on each item. VoiceOver reads `AccessibleTextRole` first when naming a list item, so it now gets clean text when the user navigates to a row.

**Fix 2 — `QAccessibleTableModelChangeEvent::RowsInserted` + corrected `QAccessibleValueChangeEvent` (live-region-style announcement)**

The previous `QAccessibleValueChangeEvent` fired before HTML processing and before the row was even appended — it carried unprocessed text and referenced the whole widget at the wrong moment. Replaced with:
1. `QAccessibleTableModelChangeEvent::RowsInserted` targeting the exact row — tells the AT stack a new list item exists.
2. `QAccessibleValueChangeEvent` with the plain text, fired after `appendRow` — this is what drives VoiceOver to announce an incoming message without requiring the user to navigate to it.

Whether (2) produces true live-region auto-announcement on macOS depends on VoiceOver's policy for list widgets, but it's the correct signal sequence; the previous approach was definitely wrong.

mcfnord:
this conflict thing seems so extra... i think i'll make a new PR

PR #3750, opened 2026-06-22
chatdlg: replace QTextBrowser with QListView for screen reader accessibility

mcfnord:
Replaces `QTextBrowser` with `QListView` + `QStyledItemDelegate` in the chat dialog so that each message is a discrete item in the OS accessibility tree.

With `QTextBrowser`, VoiceOver (and other screen readers that consume `QAccessible`) sees the entire chat history as a single text block. With `QListView`, each appended message becomes a `QAccessible::ListItem`, letting the user arrow-key through individual messages.

`ChatDelegate` renders each item as HTML via `QTextDocument`, preserving the existing rich-text formatting. Link clicks are detected in `eventFilter` by installing a filter on the viewport and using `QTextDocument::documentLayout()->anchorAt()` to hit-test the click position. Hovering over a link shows a pointing-hand cursor. Clicking routes through the existing `OnAnchorClicked` confirmation dialog. Right-click or Ctrl+C copies the selected message's plain text to the clipboard.

## macOS VoiceOver fixes (addresses the silence reported in testing)

Two issues prevented VoiceOver from speaking on macOS:

1. **`Qt::AccessibleTextRole`** — `Qt::DisplayRole` stores raw HTML. VoiceOver reads this role verbatim and sees angle-bracket markup, producing silence or garbled output. Fix: `QTextDocument::toPlainText()` strips the tags, and the result is stored as `Qt::AccessibleTextRole` on each item — the role VoiceOver actually queries for the item's accessible name.

2. **Correct accessibility events** — Replaced the previous `QAccessibleValueChangeEvent` (which fired before HTML processing and before the row existed) with `QAccessibleTableModelChangeEvent::RowsInserted` targeting the exact new row, followed by `QAccessibleValueChangeEvent` carrying the plain text. This is the correct signal sequence for live-region-style auto-announcement of incoming messages on macOS.

**Fixes an issue?**

Addresses #3613.

No documentation changes needed.

**Status**

Builds and runs. VoiceOver on macOS should now speak individual messages when navigating the list and announce incoming messages automatically. Please test with VoiceOver — specifically: does arrowing to a message read it aloud, and does an incoming message get announced without user action?

## Checklist

- [ ] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [ ] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#source-code-consistency)

### DIFF
--- a/src/chatdlg.cpp
+++ b/src/chatdlg.cpp
@@ -65,15 +65,23 @@ CChatDlg::CChatDlg ( QWidget* parent ) : CBaseDlg ( parent, Qt::Window ) // use 
 
     edtLocalInputText->setAccessibleName ( tr ( "New chat text edit box" ) );
 
-    // clear chat window and edit line
-    txvChatWindow->clear();
+    // clear edit line
     edtLocalInputText->clear();
-
-    // we do not want to show a cursor in the chat history
-    txvChatWindow->setCursorWidth ( 0 );
 
     // set a placeholder text to make sure where to type the message in (#384)
     edtLocalInputText->setPlaceholderText ( tr ( "Type a message here" ) );
+
+    // Set up the list model and delegate for accessible per-message rows ------
+    m_pChatModel = new QStandardItemModel ( this );
+    txvChatWindow->setModel ( m_pChatModel );
+    txvChatWindow->setItemDelegate ( new ChatDelegate ( txvChatWindow ) );
+    txvChatWindow->setSelectionMode ( QAbstractItemView::SingleSelection );
+    txvChatWindow->setEditTriggers ( QAbstractItemView::NoEditTriggers );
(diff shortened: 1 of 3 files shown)


mcfnord:
Someone please try this on a  Mac using VoiceOver.

ann0see:
No. Unfortunately this also does not fix it. 

1. There's no way to tab into the field
2. Even if i click into it, voice over does not add focus

mcfnord:
I guess someone who has a Mac needs to make it work.

pljones:
I get the feeling no one at Qt itself has a mac...

Issue #2344, opened 2022-02-05
Allow mono input device and channel selection for stereo input devices.


pgScorpio:
**Describe the bug**

Momentarily it is impossible to use a mono input device and also impossible to select input channels for stereo devices, though there seems to be no reason why this should not be possible, since it is possible to select twice the same input channel with devices with more than two input channels.
Also in some cases it would be handy to swap L/R output channels which is currently also impossible for two channel devices.

**To Reproduce**

Just try different devices:
mono input device: Error not 2 ch.
stereo input device: no input/output selection possible
more than 2 channel device: all is possible.

**Expected behavior**

Always being able to select L/R channels for input and output. (and for mono input devices always select twice the same mono input)

**Screenshots**

n.a.

**Operating system**

All OS'es

**Version of Jamulus**

All versions

**Additional context**

Also discussed on [discussion #2268](https://github.com/jamulussoftware/jamulus/discussions/2268#discussioncomment-2114430)

changes for v3.8.1 windows release:

Add in sound.h: 
(In my opinion these and some other definitions in sound.h should actually be defined in global.h since they are client, and not OS, related, but are of the same category as MAX_NUM_IN_OUT_CHANNELS which is defined in global.h)
```
42 #define MIN_IN_CHANNELS 1
43 #define MIN_OUT_CHANNELS 2

```
Chances in sound.cpp:
```
168 if ( ( lNumInChan < MIN_IN_CHANNELS ) || ( lNumOutChan < MIN_OUT_CHANNELS ) )

586 vSelectedInputChannels[1]  = GetNumInputChannels() > 1 ? 1 : 0;

588 vSelectedOutputChannels[1] = GetNumOutputChannels() > 1 ? 1 : 0;
```
Same (alike) changes also have to be done for Mac/Linux/Android  sound.cpp, sound.h

Chance in clientsettingsdlg.cpp
```
823 if ( ( iNumInChannels < MIN_IN_CHANNELS  ) || ( iNumOutChannels < MIN_OUT_CHANNELS  ) )

```
I don't think there is any other impact on the GUI, but that has to be further checked/tested.

ann0see:
Great thanks. If no discussion comes up the next 2 day, feel free to open a PR with these changes. I hope it doesn’t have an impact on other functionality (you never know…)

henkdegroot:
> should actually be defined in global.h since they are client, and not OS,

[...]

Issue #96, opened 2020-04-18
Multi Channel Fader on one Jamulus


nicolopozzato:
I play with some friends... All of them have a simple request...

A way to send to the Jamulus Server multiple audio over different fader in the mixer, all of this can be Mono or Stereo.

Example:
I have a microphone, it's mono, so in the first channel I set it in mono, then i can select what device, and the input channel of that device.
After that I have a Keyboard, so in the second channel i can set it in stereo (stereo->stereo, for send two channel, or a stereo->mono to mix down a mono track), then i can select what device, and then i can select the two channel.

After the input I can select if i want a mono or stereo return from the server to hear in the headphone.

We ask this because we all have to open more Jamulus each time, and this leads us to throw away server bandwidth.
We need everyone to do their own mix on headphones, some musicians have two or more keyboards and a microphone, so they start jamulus many times.

If you put together all the open Jamulus as a second or more channel (99% in stereo) we find ourselves having a lot of bandwidth thrown away by the server because the server sends the mix to all the jamulus.

It would be nice to have a setting for unlimited channels. every channel could be mono or true stereo or a strereo mix down mono, every channel could select a different audio device and different channels, also a different name.

In this way you can eliminate the complexity of running multiple Jamulus client and the extra bandwith trown away.

pljones:
You can hook up a DAW to Jamulus - but it depends on having routing that both ends understand.  JACK on Linux is fine.  I can't comment on MacOS.  On Windows, I run it using Reaper with about 60 or 70 different local channels running through the mixing desk before being routed out to Jamulus and the return audio fed back again.

nicolopozzato:
Ok but all the channel are local, only for you... The other member see only one channel. I was asking for a way to send multiple channel like I was using multiple Jamulus, so every member can adjust their own audio.

pljones:
You can run multiple instances of Jamulus, too, if you do it right - that gives that level of control, too.

nicolopozzato:
But running multiple Jamulus add up also all the output from the server, but they are not needed... So I was thinking at a solution to send multiple separate channel in a Jamulus Server, but to have only one feed to ear from the server.

pljones:
Two input channels on one client of software adds to the server just as much as two input channels on two clients.  It makes no difference - it's just as much network load and CPU load.

nicolopozzato:
[...]

PR #3746, opened 2026-06-19
Claude: add support for multichannel interfaces on iOS/iPadOS

rdica:
**Short description of changes**

Adds support for external multichannel interfaces and provides ability to choose inputs on iOS/iPadOS devices.

CHANGELOG: Added support for external multichannel interfaces and ability to choose inputs on iOS/iPadOS devices.

**Context: Fixes an issue?**

Fixes: 3735

**Does this change need documentation? What needs to be documented and how?**

I don't know.

**Status of this Pull Request**

**What is missing until this pull request can be merged?**

This absolutely needs testing as I have no means to do so.

I don't really expect this to work but hopefully it provides a starting point.

Claude summary:
What was missing: `src/sound/coreaudio-ios/sound.mm` hardcoded a 2-channel (stereo) input format on the RemoteIO audio unit and never queried `AVAudioSession` for a device's actual channel count, so any external multichannel interface (USB/Lightning audio interface) was silently truncated to its first two channels.

Changes (`src/sound/coreaudio-ios/sound.{h,mm}`):

- Negotiate channel count via `AVAudioSession.maximumInputNumberOfChannels` / `setPreferredInputNumberOfChannels:` / `inputNumberOfChannels` whenever a device is selected (`SwitchDevice`).
- Split the single audio format into separate input (now `iNumInChan`-wide) and output (still stereo, since iOS playback is always 2-channel) `AudioStreamBasicDescription`s applied to the RemoteIO unit.
- `GetAvailableInOutDevices()` now lists every `AVAudioSessionPortDescription` in availableInputs (built-in mic, headset, external interfaces) instead of just "default" vs "built-in mic".
- Added `UpdateInputChannelInfo()` to populate per-channel names (from `AVAudioSessionChannelDescription`) and clamp the current selection when the device changes.
- `processBufferList()` now indexes the interleaved N-channel buffer using the user-selected left/right channel instead of assuming channel 0/1.
- Implemented `GetNumInputChannels`, `GetInputChannelName`, `SetLeft/RightInputChannel`, `GetLeft/RightInputChannel` overrides — these plug straight into the existing generic channel-selection UI in `clientsettingsdlg.cpp`, which is already gated on `__APPLE__` (true for iOS) and just was never exercised because the backend always reported 2 channels.

No GUI changes were needed — the settings dialog's channel-selection frame already supports any platform reporting more than 2 channels.

[...]

PR #3757, opened 2026-06-28
Add two-input stereo-output microphone mode

seanogdelaney:
Summary:
- add a Two-in/Stereo-out client audio mode
- send the second input as an auxiliary mono Mic channel
- use the Vocal instrument icon for the Mic channel
- persist and expose the new audio-channel setting

- Implements #3756.
- Implements a narrow subset of #96.
- Related to #2344.

Validation:
- Compiles on Linux
- Runs on Linux
- Two channels appear in mixer as expected, with second suffixed 'Mic', using vocal icon.

Notes:
Running two jamulus clients can achieve this, but is a bit awkward.
This way makes it more accessible (less flexible, but easier).

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -94,28 +94,28 @@ public:
         }
     }
 
-    void Configure ( const CHostAddress&      ServerAddress,
-                     const CChannelCoreInfo&  AuxiliaryChannelInfo,
-                     OpusCustomMode*          pNewOpusMode,
-                     const EAudComprType      eNewAudioCompressionType,
-                     const int                iNewCodedBytes,
-                     const int                iNewFrameSizeSamples,
-                     const int                iNewFrameSizeFactor,
-                     const int                iNewSockBufNumFrames,
-                     const int                iNewServerSockBufNumFrames,
-                     const bool               bNewDoAutoSockBufSize )
+    void Configure ( const CHostAddress&     ServerAddress,
+                     const CChannelCoreInfo& AuxiliaryChannelInfo,
+                     OpusCustomMode*         pNewOpusMode,
+                     const EAudComprType     eNewAudioCompressionType,
+                     const int               iNewCodedBytes,
+                     const int               iNewFrameSizeSamples,
+                     const int               iNewFrameSizeFactor,
+                     const int               iNewSockBufNumFrames,
+                     const int               iNewServerSockBufNumFrames,
+                     const bool              bNewDoAutoSockBufSize )
     {
         const bool bEncoderChanged = ( pOpusMode != pNewOpusMode );
 
-        ChannelInfo               = AuxiliaryChannelInfo;
-        pOpusMode                 = pNewOpusMode;
-        eAudioCompressionType     = eNewAudioCompressionType;
-        iCodedBytes               = iNewCodedBytes;
-        iFrameSizeSamples         = iNewFrameSizeSamples;
-        iFrameSizeFactor          = iNewFrameSizeFactor;
-        iSockBufNumFrames         = iNewSockBufNumFrames;
-        iServerSockBufNumFrames   = iNewServerSockBufNumFrames;
-        bDoAutoSockBufSize        = bNewDoAutoSockBufSize;
+        ChannelInfo             = AuxiliaryChannelInfo;
+        pOpusMode               = pNewOpusMode;
+        eAudioCompressionType   = eNewAudioCompressionType;
+        iCodedBytes             = iNewCodedBytes;
+        iFrameSizeSamples       = iNewFrameSizeSamples;
+        iFrameSizeFactor        = iNewFrameSizeFactor;
+        iSockBufNumFrames       = iNewSockBufNumFrames;
+        iServerSockBufNumFrames = iNewServerSockBufNumFrames;
+        bDoAutoSockBufSize      = bNewDoAutoSockBufSize;
 
         Channel.SetAddress ( ServerAddress );
         Channel.SetSockBufNumFrames ( iSockBufNumFrames );
@@ -145,8 +145,7 @@ public:
             }
         }
 
-        opus_custom_encoder_ctl ( pOpusEncoder,
-                                  OPUS_SET_BITRATE ( CalcBitRateBitsPerSecFromCodedBytes ( iCodedBytes, iFrameSizeSamples ) ) );
+        opus_custom_encoder_ctl ( pOpusEncoder, OPUS_SET_BITRATE ( CalcBitRateBitsPerSecFromCodedBytes ( iCodedBytes, iFrameSizeSamples ) ) );
 
         vecCodedData.Init ( iCodedBytes );
         vecReceiveData.Init ( iCodedBytes );
@@ -194,9 +193,7 @@ public:
         }
     }
 
-    void SetSocketBufferSettings ( const int  iNewSockBufNumFrames,
-                                   const int  iNewServerSockBufNumFrames,
-                                   const bool bNewDoAutoSockBufSize )
+    void SetSocketBufferSettings ( const int iNewSockBufNumFrames, const int iNewServerSockBufNumFrames, const bool bNewDoAutoSockBufSize )
     {
         iSockBufNumFrames       = iNewSockBufNumFrames;
         iServerSockBufNumFrames = iNewServerSockBufNumFrames;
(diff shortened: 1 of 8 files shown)


dingodoppelt:
From a first skim through the code I could gather that `CAuxiliaryMonoSender` basically duplicates `CClient` without raw audio transmission. Is there a reason for not trying to reuse `CClient`?  
I guess since it is client only code we now occupy two channels with one client?  
Have you thought about less invasive alternatives like UI only changes, i.e. splitting stereo clients (with vocal and instrument combination) into separate channels?

seanogdelaney:
Thanks. Yes, the auxiliary sender is intentionally a cut-down client-side
network peer, so it does occupy a second server channel. That is the point:
the extra vocal input appears as its own fader/channel in the mixer for all
participants, without requiring server or protocol changes.

I did consider reusing "CClient", but "CClient" owns much more than the
transmit path: sound device lifecycle, receive/mix/playback, UI-facing
state, raw audio handling, local monitor logic, etc. A second "CClient"
would either try to own another sound device or require a larger refactor
to separate “Jamulus network/audio stream endpoint” from “full client”. The
auxiliary sender is intended to avoid that larger architectural change.

I agree there is overlap. It would be possible to refactor CClient into a
reusable base plus the existing full client, with the auxiliary sender
deriving from (or using) the reusable part. That looked like a
significantly larger architectural change. For this PR, a small private
helper seemed simpler and less invasive.

I think you may misunderstand. A UI-only change won't work here. Each
client channel exists on the server, where it is mixed and sent to each
clients. To expose separate vocal and instrument faders, the server must
receive them as separate channels. Otherwise we'd need to modify the server
to split the incoming stereo stream into separate mono channels when this
client mode is enabled, which would be a considerably more invasive change.

dingodoppelt:
> I think you may misunderstand. A UI-only change won't work here. Each client channel exists on the server, where it is mixed and sent to each clients. To expose separate vocal and instrument faders, the server must receive them as separate channels. Otherwise we'd need to modify the server to split the incoming stereo stream into separate mono channels when this client mode is enabled, which would be a considerably more invasive change.

I find this change very invasive by it's sheer size while it is not very clear what is to be achieved.  
If we are using two server slots, why is this mono-only/non-raw? A client already sends two mono signals in all available audio qualities. This should be exploited first before duplication of code.   
Why do you think changes to the server would be more invasive? A server would receive a stereo packet and treat it like two mono paths. I'd start from there to keep changes simple and small.

src/client.cpp
```diff
@@ -259,6 +444,86 @@ CClient::~CClient()
     opus_custom_mode_destroy ( Opus64Mode );
 }
 
+EAudChanConf CClient::GetEffectiveAudioChannels() const { return IsAuxiliaryMonoSenderEnabled() ? CC_MONO : eAudioChannelConf; }
+
+bool CClient::CanUseAuxiliaryMonoSender()
+{
+    return ( eAudioQuality != AQ_RAW ) && ( Sound.GetNumInputChannels() >= 2 ) && ( Sound.GetLeftInputChannel() >= 0 ) &&
```
dingodoppelt, line 451: Why is raw audio transmission left out?

src/util.h
```diff
@@ -504,7 +504,8 @@ enum EAudChanConf
     // used for settings -> enum values should be fixed
     CC_MONO               = 0,
     CC_MONO_IN_STEREO_OUT = 1,
-    CC_STEREO             = 2
+    CC_STEREO             = 2,
+    CC_TWO_IN_STEREO_OUT  = 3
```
dingodoppelt, line 508: I'd say stereo already covers this. This should rather tell the server to treat a stereo signal as two mono signals and have the client display two channels instead of one.

seanogdelaney:
Firstly, respect (to you and the community) for such fast responses on issues / PRs. It's much appreciated!

You express three very reasonable concerns: intent clarity, code size, and the server-side alternative. My perspective is:

**Intent Clarity**: The aim is "*mixer shows 1 fader per source*" instead of "*1 fader per person*". The most common situation is a person who sings and plays an instrument (two sources). This PR addresses that case, but could be extended later (e.g. stereo instrument + mic), see for example issue #96.

**Code size**: This change is about 300 real lines of code, mostly isolated in the auxiliary sender. If it can be achieved in less code (without sacrificing quality / maintainability), I'd like that very much.

**Server-side alternative**: I agree it's possible (i.e. a client check-box triggers the server to split a stereo channel into two mono channels before the mixer stage). It's not clear to me that this would be less code, or less invasive. It also doesn't provide a natural path to cases "stereo DJ decks + mic", or the more general multichannel support discussed in #96.

So my thinking was to implement the more general model from the outset: independently mixable sources are sent as independent channels. This PR just implements the simplest two-source case.

dingodoppelt:
> This way makes it more accessible (less flexible, but easier).

Is there really a usecase for this? Why is this better than using two clients or simply mixing a stereo mix and feeding that into Jamulus for the sake of stereo and raw audio? I don't know if people will use this when there is not even a stereo signal available.  
By the approach taken a lot of  code is duplicated and would have to be maintained in parallel. I really think the server is the place to start here and have clients do the visuals.  
I could see some benefit in this, if...
- There was only one channel occupied at the server
- There was stereo
- There was raw audio

In general I think we shouldn't give two channels to one client. I know this is being done but I have never liked it and I still think people shouldn't do it.  
Missing stereo and raw audio support will be a deal breaker for many.

seanogdelaney:
Thanks, that helps clarify the concern.

**Use case:** I believe there is a real use case here. A mono instrument plus mono vocal is common for people playing in Jamulus “rock band” style sessions: bass + vocal, guitar + vocal, etc. In that situation, independent mixer faders are useful even though neither source needs stereo.

**Why not two Jamulus clients:** that is what I currently do, but it is hard to recommend to less technical users. It requires either scripts with separate ini files, or repeated manual configuration. Many users will find that awkward or conceptually confusing.

**Code duplication:** I agree this is the main weakness of the current implementation. If you are open to it, I can reduce this by factoring the relevant sender-side parts out of `CClient` rather than maintaining parallel code.

**Stereo / raw audio:** agreed, this PR does not cover those cases. They would require more code, so I would prefer to handle them in a follow-up PR if this general direction is acceptable.

**Multiple mixer channels per user:** I understand this is the central design disagreement. I do not want to argue against the maintainers’ preferred model. My motivation is simply that I believe many users would find this useful, myself included, and that independently mixable sources mapping to independent channels is a natural model for this use case.

dingodoppelt:
> Thanks, that helps clarify the concern.
> 
> **Use case:** I believe there is a real use case here. A mono instrument plus mono vocal is common for people playing in Jamulus “rock band” style sessions: bass + vocal, guitar + vocal, etc. In that situation, independent mixer faders are useful even though neither source needs stereo.
> 

I use a similar setup (multiple instruments, analog and digital through some effect chains) and I surely need stereo for everything. I'm mixing all my channels locally and just send a stereo mix.

> **Why not two Jamulus clients:** that is what I currently do, but it is hard to recommend to less technical users. It requires either scripts with separate ini files, or repeated manual configuration. Many users will find that awkward or conceptually confusing.

To launch the client with a separate INI file you don't need a script, just the cli option `-i`. This is well documented and not at all awkward or confusing in my opinion. At least on Linux this is a one-click matter.
 
> **Code duplication:** I agree this is the main weakness of the current implementation. If you are open to it, I can reduce this by factoring the relevant sender-side parts out of `CClient` rather than maintaining parallel code.

I'd say `CServer` is a better place to make this happen as we already have a path to send two mono channels (stereo configuration)

> **Multiple mixer channels per user:** I understand this is the central design disagreement. I do not want to argue against the maintainers’ preferred model. My motivation is simply that I believe many users would find this useful, myself included, and that independently mixable sources mapping to independent channels is a natural model for this use case.

"Multiple" is not what you actually propose here. It is "two mono signals" which is a stereo signal in current Jamulus terms. This only applies to users using mono for all their instruments exclusively. This may be ok for a bass player but pianists, guitarists and vocalists generally use stereo signals for the sake of sound fidelity (this is what we tell users when they join a server with mono signals) and those users wouldn't make use of this feature.

dingodoppelt:
We have #3746 that introduces multichannel support but only on iOS.  
If we had this cross platform we could think about acquiring more than two channels and send those simultaneously from one client. Either in multiple messages or over dedicated sockets.  
Since the limitation mono/compressed is not a technical boundary but a deliberate design decision and tailored to a very limited use case I don't think this is the right approach. Not many users will make use of this I fear.

dingodoppelt:
If running two clients seems awkward you may want to try running the secondary client without a GUI (`-n`) so you don't receive chat messages twice and stuff like that. An RPC method has just been introduced to mute yourself or you could use MIDI to control the headless client.  
In theory you could have even more clients with stereo/raw sound that way.

src/client.cpp
```diff
@@ -49,6 +49,186 @@
...
+    void Stop()
+    {
+        if ( Channel.IsEnabled() )
+        {
+            Channel.SetEnable ( false );
+            ConnLessProtocol.CreateCLDisconnection ( Channel.GetAddress() );
+        }
+    }
+
+    void Process ( const CVector<int16_t>& vecMonoSamples, const bool bMute )
```
dingodoppelt, line 170: This would have to be maintained separately from the main code for audio transmission and is already out of date as it doesn't support raw audio or even stereo while there is no clear reason other than covering a singular use case.  
While we have a socket, why not use it to its full potential?

src/client.cpp
```diff
@@ -49,6 +49,186 @@
...
+        int iUnused;
+        for ( int i = 0, iOffset = 0; i < iFrameSizeFactor; i++, iOffset += iFrameSizeSamples )
+        {
+            iUnused = opus_custom_encode ( pOpusEncoder,
+                                           bMute ? &vecZeros[iOffset] : &vecMonoSamples[iOffset],
+                                           iFrameSizeSamples,
+                                           &vecCodedData[0],
+                                           iCodedBytes );
+            Channel.PrepAndSendPacket ( &Socket, vecCodedData, iCodedBytes );
+            Channel.GetData ( vecReceiveData, iCodedBytes );
```
dingodoppelt, line 181: The server sends audio that will never be played back from the client. Is this really necessary? Shouldn't the server rather not return anything, just consume additional samples from the client? `CAuxiliaryMonoSender` doesn't imply receiving as well.

dingodoppelt:
#1063 and #1840 are still open and it hasn't been agreed upon a strategy to pursue with this but the ideas seem to point in a different direction of a more general way of sending multichannel audio. Use cases differ wildly and we need the widest possible coverage.

pljones:
A another user on a server, I have no idea other than each fader is a real person.  I want to see what each fader represents (i.e. if someone is a singer and sitting next to them is a guitar player, I want both their names and instruments).  So, effectively, I see no model - from other Jamulus users' perspectives - where a fader should not reflect the profile information as well as the audio channel.

seanogdelaney:
Thanks all for the comments, and reviewing effort.

I agree. This patch works but is architecturally flawed. Issue 1840 is the best hope of a general solution, if an architecture can be agreed.

I might try implementing something for that, if only to illuminate and encourage discussion.

PR #3772, opened 2026-07-06
Add jamulusclient/setInstrumentCode JSONRPC method

### LINKED ISSUE #2466: JSON-RPC: Add client method to set instrument
*Raised by @softins in https://github.com/jamulussoftware/jamulus/pull/1975#pullrequestreview-901000643*
>  I must admit I struggle to understand a reason for making `skillLevel` changeable via the RPC, but not an instrument. I would only ever need to change my skill level if I was also changing my instrument, as my skill level on a given instrument is pretty static!
> [...] we need to be able to set the user's instrument. In fact, the complement to `jamulusclient/getChannelInfo` maybe should be `jamulusclient/setChannelInfo` for consistency?

I assume agreement that a way for setting the instrument should be added.
There's no agreement on the particular implementation yet.

1. Add `jamulusclient/setInstrument` and keep `setName` and `setSkillLevel`
  - :+1: Most fine-grained solution
  - :+1: Saves the API user from having to request and (re)set all existing, unchanged values
  - :-1:  Amount of code
2. Add `jamulusclient/setChannelInfo` and drop `setName` and `setSkillLevel`
  - :+1: Consistent with `getChannelInfo`
  - :+1: Uses a single protocol message instead of one per change
  - :-1: Requires API users to request the current data first and pre-fill the fields which should not be changed. Maybe we could make the fields optional, though?

I'd go with (2). I read @pljones' https://github.com/jamulussoftware/jamulus/pull/1975#issuecomment-1059839552 as also supporting that approach.

dingodoppelt:
Change your instrument by code number via JSONRPC (client only method)

CHANGELOG: Add jamulusclient/setInstrumentCode JSONRPC method

**Context: Fixes an issue?**

I accidently recommended in #3769 to set the instrument via JSONRPC while that wasn't actually possible.
This fixes it.  
Fixes #2466

Already documented in the generated JSONRPC docs.

Approval

## Checklist

### DIFF
--- a/src/clientrpc.cpp
+++ b/src/clientrpc.cpp
@@ -328,6 +328,32 @@ CClientRpc::CClientRpc ( CClient* pClient, CClientSettings* pSettings, CRpcServe
         response["result"] = "ok";
     } );
 
+    /// @rpc_method jamulusclient/setInstrumentCode
+    /// @brief Sets your instrument code.
+    /// @param {number} params.instrCode - The new instrument code.
+    /// @result {string} result - Always "ok".
+    pRpcServer->HandleMethod ( "jamulusclient/setInstrumentCode", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        auto jsonInstrCode = params["instrCode"];
+
+        if ( !jsonInstrCode.isDouble() )
+        {
+            response["error"] = CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: instrCode is not a number" );
+            return;
+        }
+
+        const int iNInstrument = jsonInstrCode.toInt();
+
+        if ( CInstPictures::GetName ( iNInstrument ).isEmpty() )
+        {
+            response["error"] = CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: unknown instrCode" );
+            return;
+        }
+
+        pClient->ChannelInfo.iInstrument = iNInstrument;
+        pClient->SetRemoteInfo();
+        response["result"] = "ok";
+    } );
+
     /// @rpc_method jamulusclient/sendChatText
     /// @brief Sends a chat text message.
     /// @param {string} params.chatText - The chat text message.
--- a/src/clientrpc.cpp
+++ b/src/clientrpc.cpp
@@ -341,7 +341,7 @@ CClientRpc::CClientRpc ( CClient* pClient, CClientSettings* pSettings, CRpcServe
             return;
         }
 
-        const int iNInstrument = jsonInstrCode.toInt();
+        const int iNInstrument = static_cast<int> ( jsonInstrCode.toDouble() );
 
         if ( CInstPictures::GetName ( iNInstrument ).isEmpty() )
         {
--- a/docs/JSON-RPC.md
+++ b/docs/JSON-RPC.md
@@ -274,6 +274,23 @@ Results:
 | result | string | Always "ok". |
 
 
+### jamulusclient/setInstrumentCode
+
+Sets your instrument code.
+
+Parameters:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| params.instrCode | number | The new instrument code. |
+
+Results:
+
+| Name | Type | Description |
+| --- | --- | --- |
+| result | string | Always "ok". |
+
+
 ### jamulusclient/setMidiSettings
 
 Sets one or more MIDI controller settings.


pljones (approved):

src/clientrpc.cpp
```diff
@@ -331,6 +331,32 @@ CClientRpc::CClientRpc ( CClient* pClient, CClientSettings* pSettings, CRpcServe
         response["result"] = "ok";
     } );
 
+    /// @rpc_method jamulusclient/setInstrumentCode
+    /// @brief Sets your instrument code.
+    /// @param {number} params.instrCode - The new instrument code.
+    /// @result {string} result - Always "ok".
+    pRpcServer->HandleMethod ( "jamulusclient/setInstrumentCode", [=] ( const QJsonObject& params, QJsonObject& response ) {
```
ann0see, line 335: We should have one call for instrument, name, skill level, imo.

If it was REST, we'd have PUT client/profile {name: ..., instrument: ..., skill level: ..., city:... } etc.

dingodoppelt: Do you mean to add `setCity` and `setCountryCode` as those are still missing? Skill and name are already settable, instrument will be with this merged.

ann0see: Yes. However I think an alternative design would be one method to set all metadata for the client

pljones: For the JSONRPC, it's setter/getter mapping to the CClient entrypoints - or, here, CClientSettings (although I'd rather CClient orchestrated that...).

dingodoppelt: > Yes. However I think an alternative design would be one method to set all metadata for the client

True. We submit our changes with `pClient->SetRemoteInfo()` so this is already happening (kind of).  
Should the name be reused as well or should it rather be `setChannelInfo` analogous to `getChannelInfo`?
The method should still let users set single fields, i.e. only those supplied in the request, and leave the others untouched. There might be valid and invalid fields in the request which need to be checked for.

ann0see: I think setChannelInfo would be good.

dingodoppelt: > For the JSONRPC, it's setter/getter mapping to the CClient entrypoints - or, here, CClientSettings (although I'd rather CClient orchestrated that...).

At the moment we have one getter `getChannelInfo` but multiple setters. This looks not too elegant but on the other hand I don't see see the need for more than one getter.  
Implementing a single setter would be trickier than having multiple setters for we have to check more input fields for validity at the same time and need to define behaviour for every possible combination of invalid and valid fields.  
Any ideas on how to proceed here? @ann0see @pljones

ann0see: Can't we parse the json and validate it against a json schema?

This should make optional fields trivial

dingodoppelt: Yes, invalid json should fail right away.  
After successful parsing we check which fields are valid.  (name, instrument code, shoesize, ...)
We then check each field's value for validity (type, length) and set the valid ones.  
We could return the number of set fields or maybe a more verbose message?

pljones: If it were REST, of course, you'd return the mutated state and let the caller worry about what worked and what didn't... (There's not even an HTTP status code that indicates a partially processed request.)

If the client were to send over JSONRPC
```json
{
  "name": "Bob",
  "instrumentCode": <violin>,
  "mood": "Bachian mode",
}
```
I'd expect a
```
HTTP 200 OK

{ "status": "partially successful", "warnings": [ { "property": "mood", "result": "unrecognised" } ] }
```
or something, if you were going to get really fancy about it.  Otherwise, I think "ok" and ignore what's not implemented.

Ideally, if it's not going to have a complex result returned, it would be atomic: capture the current state; attempt each update; if any failed, revert to the previous state and return "error".

pljones (approved):

src/clientrpc.cpp
```diff
@@ -331,6 +331,32 @@ CClientRpc::CClientRpc ( CClient* pClient, CClientSettings* pSettings, CRpcServe
...
+
+        const int _iInstrument = jsonInstrCode.toInt();
+
+        if ( CInstPictures::GetName ( _iInstrument ).isEmpty() )
+        {
+            response["error"] = CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: unknown instrCode" );
+            return;
+        }
+
+        pClient->ChannelInfo.iInstrument = _iInstrument;
```
mcfnord: Minor style: the leading underscore on `_iInstrument` is non-idiomatic here — no other local in this handler file uses one (siblings use `jsonName`, `strSkillLevel`, etc.). Renaming to `iInstrCode` also lines it up with the `instrCode` param. Single apply-all suggestion so the variable stays consistent:

```suggestion
        const int iInstrCode = jsonInstrCode.toInt();

        if ( CInstPictures::GetName ( iInstrCode ).isEmpty() )
        {
            response["error"] = CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: unknown instrCode" );
            return;
        }

        pClient->ChannelInfo.iInstrument = iInstrCode;
```

mcfnord (commented):
One small polish suggestion on the new handler — everything else looks good.

src/clientrpc.cpp
dingodoppelt: I changed it to what it is elsewhere in the code

pljones (approved):

ann0see:
Small observation: could we reject non-integer codes? Qt's `QJsonValue::toInt()` returns 0 for values like `3.7`, so `setInstrumentCode(3.7)` returns "ok" but silently sets instrument to 0 ("None") instead of being rejected. A quick integrality check (e.g. `jsonInstrCode.toInt() != jsonInstrCode.toDouble()`) would match the existing invalid-params handling. Not blocking though.

Disclaimer: This comment was written with AI assistance to test for bugs.

dingodoppelt:
> Small observation: could we reject non-integer codes? Qt's `QJsonValue::toInt()` returns 0 for values like `3.7`, so `setInstrumentCode(3.7)` returns "ok" but silently sets instrument to 0 ("None") instead of being rejected. A quick integrality check (e.g. `jsonInstrCode.toInt() != jsonInstrCode.toDouble()`) would match the existing invalid-params handling. Not blocking though.
> 
> Disclaimer: This comment was written with AI assistance to test for bugs.

I assumed `toInt()` to behave like `static_cast<int>`. I don't know why somebody would send a float value but I think a simple cast just truncates the decimal place and still gets sanity checked afterwards.

src/clientrpc.cpp
```diff
@@ -328,6 +328,32 @@ CClientRpc::CClientRpc ( CClient* pClient, CClientSettings* pSettings, CRpcServe
...
+    pRpcServer->HandleMethod ( "jamulusclient/setInstrumentCode", [=] ( const QJsonObject& params, QJsonObject& response ) {
+        auto jsonInstrCode = params["instrCode"];
+
+        if ( !jsonInstrCode.isDouble() )
+        {
+            response["error"] = CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: instrCode is not a number" );
+            return;
+        }
+
+        const int iNInstrument = jsonInstrCode.toInt();
```
dingodoppelt:
```suggestion
        const int iNInstrument = static_cast<int> ( jsonInstrCode.toDouble() );
```

This simply truncates the decimal place (as initially intended)

ann0see: Yes, Probably that's then better. The thing is: Users may send any kind of invalid data and this should always be guarded.

dingodoppelt: This applies to `jamulusclient/setFaderLevel` as well. Basically all `toInt()` calls should be checked

pljones:
`toInt()` has a variant that sets a success indicator.  That's should always be used when validating external input (even stored in a settings ini file, really).

ann0see (approved):

Issue #3379, opened 2024-09-21
Move reverb silder to advanced settings


ann0see:
**What is the current behaviour and why should it be changed?**

Currently, the reverb slider takes up a lot of space in the main mixer board. It is assumed that this could be mistaken as general volume slider. In order to simplify the UI, the reverb slider should go to advanced settings instead.

In future - if we support plugins, the reverb feature could be removed and replaced by custom reverb plugins.

**Describe possible approaches**

Move the slider to advanced settings.

**Has this feature been discussed and generally agreed?**

See: https://github.com/orgs/jamulussoftware/discussions/3341#discussioncomment-10576183

ann0see:
This is a UI change and will probably stir confusion for people who are using the reverb slider. They might think that the feature was removed.
Therefore, we must document this and make clear that this makes the UI simpler.

PR #3787, opened 2026-07-17
Use MVerb by Martin Eastwood for reverberation

dingodoppelt:

This replaces the inbuilt reverb with MVerb by Martin Eastwood.

CHANGELOG: Use MVerb by Martin Eastwood for reverberation

**Context: Fixes an issue?**

This replaces the inbuilt reverb with a professional grade reverb algorithm:
> MVerb is a free, studio quality, open-source reverb. Its release is intended to provide a practical demonstration of Dattorro's figure-of-eight reverb structure and provide the open source community with a high quality reverb.

- [x] Since we pull MVerb as a submodule build instructions are needed in `COMPILING.md`~, similar to the Android build instructions.~  
- [x] Since the introduction of presets we might need documentation for that.  
Corresponding PR: https://github.com/jamulussoftware/jamuluswebsite/pull/1153

Up for discussion

As this has not been tested thoroughly I'd like to keep it a draft for now 

## Checklist


dingodoppelt:
~The builds fail, because MVerb was pulled in as a submodule and we seem to only clone them for android builds.~

__Edit:__ fixed

pljones:
> The builds fail, because MVerb was pulled in as a submodule and we seem to only clone them for android builds.

I wonder if the sub-module clone can be repo-specific, so we can have one that only does liboboe for Android and another that does this for all platforms, included in the main clone step.

That suggests we also need build documentation changes included here.

dingodoppelt:
> I wonder if the sub-module clone can be repo-specific, so we can have one that only does liboboe for Android and another that does this for all platforms, included in the main clone step.

`autobuild.yml` currently handles submodules only for android, because oboe is our only submodule.  
I have never had reasonable success with autobuilds on github, especially since I don't know how to properly test these scripts locally before releasing my embarrassing attempts to the world. Any advice on how to do that the right way?

> That suggests we also need build documentation changes included here.

Haven't though about that but, yes, certainly.


**Edit:**
We could have Jamulus.pro check for existence of MVerb.h. If not found we can fail with an error message telling the user to run a target `make submodules` which would clone MVerb for all platforms and oboe only for android or just do it automatically. Just an idea because I just remembered typing `make submodules` at least once somewhere else.

ann0see:
> how to properly test these scripts locally before releasing my embarrassing attempts to the world.

If it is **not** in .github they should be designed to build .deb/.dmg/.exe/.ipa/.apk I believe that this is not achieved at the moment. However, you can try and see where they fail.

ann0see:
For my taste, it is a bit too "far" .

dingodoppelt:
> For my taste, it is a bit too "far" .

It is nothing that is "needed" but the current reverb sounds horrible, tbh.
If we ever really needed a reverb (and I heavily doubt that) we should have used this in the first place or rip it out entirely, which would be my second choice after at least trying to improve things. Have you listened to MVerb in comparison to the current implementation?

ann0see:
Yes. I think the reverb implemented here (this PR) sounds too far. Yes, probably better than what we have at the moment.

ann0see:
Actually, we could also only include the reverb if the submodule was loaded.

dingodoppelt:
> Yes. I think the reverb implemented here sounds too far. Yes, probably better than what we have at the moment.

I used one of the presets just to have at least something to start from. I'm not with my usual gear but will be tomorrow. I'll tweak some settings and try with different instruments and hopefully find a useful preset.

dingodoppelt:
Just a thought: We could easily expose the reverb parameters via JSON-RPC and store them in the .ini file

pljones:
You should be able to test the autobuild scripts on your own repo without raising a PR by pushing to a branch starting with "autobuild" -- I _think_.

dingodoppelt:
> You should be able to test the autobuild scripts on your own repo without raising a PR by pushing to a branch starting with "autobuild" -- I _think_.

Yes, that works. But I meant _locally_ as in _my machine_. I found something which involved containers and didn't work for windows so I guess I'll stick to my github repo.

ann0see:
It should definitely have the presets as configuration option in the Advanced Setup screen too.

dingodoppelt:
> It should definitely have the presets as configuration option in the Advanced Setup screen too.

I wasn't too sure about how to handle this. Either make it JSON-RPC exclusive or have power users and nerds edit their INI files, which is dead easy. I didn't want to bloat the UI with a setting nobody actually requested.  
I've never worked on UI things so that might take a little while to get into. I just got the autobuilds fixed so I'll push that first.

ann0see:
I also only have limited UI understanding. I think @ignotus666 knows more.

dingodoppelt:
> For my taste, it is a bit too "far" .

I guess this is due to the "predelay" parameter setting. I chose a preset with no pre-delay.  
We can always add new presets but I just copied those from the VST plugin for now.

dingodoppelt:
> It should definitely have the presets as configuration option in the Advanced Setup screen too.

Done

dingodoppelt:
- Linux amd64 works (tested debian autobuild binary and local arch linux build with gcc16 and qt6)
- Android works (tested on Android 17/GrapheneOS)

ignotus666:
@dingodoppelt It shouldn't be too hard to add a preset dropdown to the UI.
On the topic of effects, I've been working on an experimental branch for personal use that has an effects dialog with reverb, compression and EQ. It saves presets, so maybe you can have a look at that part. The UI is modified but the principle should be the same. I'm going to be super busy for the forseeable future so unfortunately can't be of much help. I try to keep the branch up to date and generate Win, Mac and Linux builds.
[Branch](https://github.com/ignotus666/jamulus/tree/my-build-effects)
[Builds](https://github.com/ignotus666/jamulus/actions)

Edit: Never mind, I see it's been added.

ann0see:
Thank you for adding the UI. For me it was quite confusing (while singing) still as I had the feeling that intonation was way harder. But it seems to depend on the selected preset. 

There was also https://github.com/jamulussoftware/jamulus/issues/3379 (moving the reverb slider to advanced settings) which I think we should do some day. The reverb is not used often - I suppose. And I think it clutters the UI.

dingodoppelt:
> There was also #3379 (moving the reverb slider to advanced settings) which I think we should do some day. The reverb is not used often - I suppose. And I think it clutters the UI.

Good idea. Alternatively get rid of the reverb altogether. With this change I hope the reverb sees a little more use. We'll hear ;)

ann0see:
If this was a new feature, I'd argue that it probably violates the do one thing and do it well approach - unless we add full plugin support.

dingodoppelt:
> If this was a new feature, I'd argue that it probably violates the do one thing and do it well approach - unless we add full plugin support.

I don't think we should go full plugin support. We'd be writing a plugin host and there must be a reason why there aren't many. If we made Jamulus a plugin (Linux already does that) users could load it into their effects chain and have all plugins working and not only the chosen few JamulusVST would be able to provide. I think I've mentioned [DPF](https://github.com/distrho/dpf) already in that respect.

pljones:
>  If we made Jamulus a plugin (Linux already does that)

Eh?  You mean Jack?  Which is on Windows as well.  As is ReaRoute.  And similar solutions on macOS.

To make it clear: I see the problem as two-fold.  What I want others to hear as an effect to my input and what I don't want others to hear but I want to hear as an effect to my input.  (To be even clearer: neither of those applies to the server mix returned.)

```
<audio in> -> <inline effect> -> Jamulus -> (everyone including) me
```

```
<audio in> -- --------------- -> Jamulus -> (everyone including) me
   \_ -> <send effect> -----------------------------> <return> _/
```

I don't think either of these belongs in Jamulus.  Whilst having "soft reverb on my vocals makes me feel more comfortable" might work for "what I hear", it doesn't necessarily work in the mix (causing it to get muddy at times).

I think there's still a perception that the mixing desk actually controls something other than a single audio signal that's received from the server.  That leads to an idea that having "effects in Jamulus" would allow per-channel routing to those effects.  Of course, that's not structurally possible.

ann0see:
After some more thinking: I think we should indeed detect if the submodule was cloned. If not the reverb feature should be hidden and while running qmake a warning should be issued. 

I suspect that you could check if the folder contains the reverb code and if not set pre processor flags -> use ifdefs in the code to remove the reverb functionality. For the UI hide or remove it it if reverb is not compiled. I would personally not use reverb even in this upgraded state.

dingodoppelt:
> After some more thinking: I think we should indeed detect if the submodule was cloned. If not the reverb feature should be hidden and while running qmake a warning should be issued.
> 
> I suspect that you could check if the folder contains the reverb code and if not set pre processor flags -> use ifdefs in the code to remove the reverb functionality. For the UI hide or remove it it if reverb is not compiled. I would personally not use reverb even in this upgraded state.

I'd then argue it should be on by default and a qmake option `noreverb` should be introduced. We could clone the submodule if it isn't there and `noreverb` isn't set.

ann0see:
Maybe yes. 

Better ask the user: Reverb not found. Want to clone?

pljones:
> Better ask the user: Reverb not found. Want to clone?

No, `qmake` and `make` aren't interactive processes.  You pass flags.

ann0see:
Ok. Fair point.

dingodoppelt:
> Maybe yes.
> 
> Better ask the user: Reverb not found. Want to clone?

QMake will try to clone the submodule if it can't find it and `noreverb` is *not* set. If the clone fails it will fall back to `noreverb`and build nonetheless. It could be argued we want to fail instead because the user intended to build with reverb because `noreverb` wasn't set.

Edit, a small excerpt from the autobuilds:
```shell
Submodule 'libs/mverb' (https://github.com/martineastwood/mverb.git) registered for path 'libs/mverb'
Cloning into '/home/runner/work/jamulus/jamulus/libs/mverb'...
Submodule path 'libs/mverb': checked out '9f9512e9d990fcf265e00fc7c2ee47ad20c8019f'
Project MESSAGE: building with MVerb
```

Jamulus.pro
```diff
@@ -42,6 +42,21 @@ contains(CONFIG, "nosound") {
     warning("\"nosound\" is deprecated: please use \"serveronly\" for a server-only build.")
 }
 
+!contains(CONFIG, "serveronly") {
+    !contains(CONFIG, "noreverb"):!exists($$PWD/libs/mverb/MVerb.h) {
+        !system("git -C $$PWD submodule update --init --force libs/mverb") {
+            message("MVerb not found and could not be cloned.")
+            CONFIG += noreverb
```
dingodoppelt: Should we rather be failing here? The user requested reverb but won't find it in the actual build. That might be confusing

ann0see: Yes. Fair point. Probably give the exact clone command as error message again and say it can be skipped by noreverb.

dingodoppelt: QMake will still try to clone the submodule. If that doesn't succeed we throw an error and tell the user to clone manually or build without reverb by adding `noreverb` to `CONFIG`. This way the autobuilds still work without changes to `autobuild.yml`.

dingodoppelt: I got the logic wrong and it wouldn't always clone the submodule. This is hopefully fixed now (says Google Gemini)

ann0see (changes requested):
Weirdly no reverb seems to hide the pan pots on macOS.

src/clientdlg.cpp
```diff
@@ -720,7 +726,7 @@ void CClientDlg::UpdateRevSelection()
     // update visibility of the pan controls in the audio mixer board (pan is not supported for mono)
     MainMixerBoard->SetDisplayPans ( pClient->GetAudioChannels() != CC_MONO );
```
ann0see, line 720: You may want to check if there's a bug here concerning the pan control visibility if we do not compile this function.

dingodoppelt: > You may want to check if there's a bug here concerning the pan control visibility if we do not compile this function.

I don't need to check, this *is* clearly a bug. Sorry, should be fixed with the latest commit.

Jamulus.pro
dingodoppelt: Excerpt from the autobuilds
```shell
Project MESSAGE: The target name is jamulus instead of Jamulus.
Project MESSAGE: building version "3.12.4dev-f6b3220:1786620718" (intermediate in git repository)
Submodule 'libs/mverb' (https://github.com/martineastwood/mverb.git) registered for path 'libs/mverb'
Cloning into '/home/runner/work/jamulus/jamulus/libs/mverb'...
Submodule path 'libs/mverb': checked out '9f9512e9d990fcf265e00fc7c2ee47ad20c8019f'
Project MESSAGE: building with MVerb
Project MESSAGE: JACK Audio Interface Enabled.
Project MESSAGE: The target name is jamulus instead of Jamulus.
Project MESSAGE: building version "3.12.4dev-f6b3220:1786620718" (intermediate in git repository)
Project MESSAGE: building with MVerb
Project MESSAGE: JACK Audio Interface Enabled.
Project MESSAGE: The target name is jamulus instead of Jamulus.
Project MESSAGE: building version "3.12.4dev-f6b3220:1786620718" (intermediate in git repository)
Project MESSAGE: building with MVerb
Project MESSAGE: JACK Audio Interface Enabled.
mkdir -p build-nox && cd build-nox && qmake "CONFIG+=headless serveronly" TARGET=jamulus-headless PREFIX=/usr ../Jamulus.pro
Info: creating stash file /home/runner/work/jamulus/jamulus/build-nox/.qmake.stash
Project MESSAGE: building version "3.12.4dev-f6b3220:1786620718" (intermediate in git repository)
Project MESSAGE: building without reverb plugin
Project MESSAGE: Headless mode activated.
Project MESSAGE: Restricting build to server-only due to CONFIG+=serveronly.
Project MESSAGE: server only, skipping client rpc
Project MESSAGE: building version "3.12.4dev-f6b3220:1786620718" (intermediate in git repository)
Project MESSAGE: building without reverb plugin
Project MESSAGE: Headless mode activated.
Project MESSAGE: Restricting build to server-only due to CONFIG+=serveronly.
Project MESSAGE: server only, skipping client rpc
Project MESSAGE: building version "3.12.4dev-f6b3220:1786620718" (intermediate in git repository)
Project MESSAGE: building without reverb plugin
Project MESSAGE: Headless mode activated.
Project MESSAGE: Restricting build to server-only due to CONFIG+=serveronly.
Project MESSAGE: server only, skipping client rpc
```

src/plugins/audioreverb.cpp
```diff
@@ -2,206 +2,69 @@
...
-    http://ccrma.stanford.edu/software/stk
-
-    Original description:
-    This class is derived from the CLM JCRev function, which is based on the use
-    of networks of simple allpass and comb delay filters. This class implements
-    three series allpass units, followed by four parallel comb filters, and two
-    decorrelation delay lines in parallel at the output.
+    The following code calls MVerb for reverberation.
+    MVerb was written by Martin Eastwood.
+    https://github.com/martineastwood/mverb
```
pljones: Gemini suggests:
> The following code uses the MVerb library by Martin Eastwood.
> Source: https://github.com/martineastwood/mverb
> License: GNU General Public License v3.0 (GPL-3.0)

src/util.cpp
```diff
@@ -524,10 +524,9 @@ CAboutDlg::CAboutDlg ( QWidget* parent ) : CBaseDlg ( parent )
 #        endif
 #        ifndef HEADLESS
                             "<p>" +
-                            tr ( "Audio reverberation code by Perry R. Cook and Gary P. Scavone" ) +
-                            ", 1995 - 2021"
-                            ", The Synthesis ToolKit in C++ (STK)"
-                            ", <i><a href=\"https://ccrma.stanford.edu/software/stk\">https://ccrma.stanford.edu/software/stk</a></i>"
+                            tr ( "Audio reverberation powered by MVerb" ) +
+                            ", Copyright (c) 2010 Martin Eastwood"
```
pljones: Needs to say GPL3.0 here, as well.

pljones: And "GPL3.0" here.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "libs/oboe"]
 	path = libs/oboe
 	url = ../../google/oboe.git
+[submodule "libs/mverb"]
+	path = libs/mverb
+	url = https://github.com/martineastwood/mverb.git
```
pljones, line 6: https://github.com/jamulussoftware/jamulus/blob/main/COPYING#L1694-L1721
All of this needs doing -- add
```
Jamulus uses MVerb - Copyright (c) 2010 Martin Eastwood

MVerb is free software: you can redistribute it and/or modify
it under the terms of the GNU General Public License as published by
the Free Software Foundation, either version 3 of the License, or
at your option) any later version.
```
That should do as we already include GPL3.0.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "libs/oboe"]
 	path = libs/oboe
 	url = ../../google/oboe.git
```
pljones, line 3: Relative URL?!

ann0see:
We should probably bump to c++17.

```

 src/plugins/audioreverb.h:70:22: warning: inline variables are only available with ‘-std=c++17’ or ‘-std=gnu++17’ [-Wc++17-extensions]
   70 |     constexpr static inline float const presets[NUM_REV_PRESETS][MVerb<float>::NUM_PARAMS] = { { 0., .5, 1., .5, 0., .5, 1., 1., .75 },
      |                      ^~~~~~
In file included from src/client.h:65,
                 from src/analyzerconsole.h:56,
                 from src/analyzerconsole.cpp:47:
src/plugins/audioreverb.h:70:22: warning: inline variables are only available with ‘-std=c++17’ or ‘-std=gnu++17’ [-Wc++17-extensions]
```

Jamulus.pro
```diff
@@ -90,6 +104,8 @@ DEFINES += APP_VERSION=\\\"$$VERSION\\\" \
 DEFINES += QT_NO_DEPRECATED_WARNINGS
 
 win32 {
+    # fixes error C7525: inline variables require at least '/std:c++17'
+    CONFIG += c++17
```
ann0see, line 108: Probably need to set this globally...

src/clientsettingsdlg.cpp
```diff
@@ -379,7 +379,21 @@ CClientSettingsDlg::CClientSettingsDlg ( CClient* pNCliP, CClientSettings* pNSet
...
+#ifndef NO_REVERB
+    // reverb preset
+    QString strReverbPreset = "<b>" + tr ( "Reverb Preset" ) + ":</b> " +
+                              tr ( "Jamulus uses MVerb by Martin Eastwood for reverberation. "
+                                   "MVerb comes with a set of presets you can select here. "
+                                   "Available Presets: Subtle, Stadium, Cupboard, Dark, Halves, Drum Room, Club " );
+    lblReverbPreset->setWhatsThis ( strReverbPreset );
+    cbxReverbPreset->setWhatsThis ( strReverbPreset );
+    cbxReverbPreset->setAccessibleName ( tr ( "Reverb Preset combo box" ) );
+#endif
```
ann0see: Purely a style question: but why don't you use #if, #else, etc?

src/clientsettingsdlgbase.ui
```diff
@@ -2228,7 +2221,7 @@
           <item row="1" column="0">
            <spacer name="verticalSpacer_19">
             <property name="orientation">
-             <enum>Qt::Vertical</enum>
+             <enum>Qt::Orientation::Vertical</enum>
```
ann0see, line 2224: In this file, you changed some other properties. I assume that QT Designer did this?

ann0see (commented):
See comments. Now I believe functionality wise I'd be ok with it.

src/clientsettingsdlg.cpp
dingodoppelt: "Stilblüte" fixed ;)

src/clientsettingsdlgbase.ui
dingodoppelt: Yes, I didn't edit that file at all

Jamulus.pro
dingodoppelt: Maybe this should be a separate PR? Found for example this in `threadpool.h`:  
```cpp
// compatibility with C++17 deprecation of result_of
#if __cplusplus >= 201703L
// std::invoke_result_t should be used from C++17 onwards, but is not available in C++14 and earlier
template<class F, class... Args>
using threadpool_result_t = std::invoke_result_t<F, Args...>;
#else
// std::result_of should be used on C++14 and earlier, but is deprecated in C++17
template<class F, class... Args>
using threadpool_result_t = typename std::result_of<F ( Args... )>::type;
#endif
```

ann0see: Yes. I agree. We'd need to bump CONTRIBUTING.md then also and could probably port a lot of stuff over (std::clamp for example)

dingodoppelt: My autobuilds checked out fine with c++17 set globally already

ann0see:
I think we could probably also add this for 4.0.0?

ann0see (approved):

src/plugins/audioreverb.cpp
```diff
@@ -2,206 +2,69 @@
...
-    const float scaler     = static_cast<float> ( iSampleRate ) / 44100.0f;
-
-    if ( scaler != 1.0f )
-    {
-        for ( int i = 0; i < 9; i++ )
-        {
-            int delay = static_cast<int> ( floorf ( scaler * lengths[i] ) );
+    // Jamulus uses interleaved stereo, mverb operates on a 2-dimensional array instead
+    // Calculate the number of frames for each channel ( iStereoBlockSizeSam / 2 )
+    numFrames = iStereoBlockSizeSam >> 1;
```
pljones: OK, so the comment is there.  I'd rather leave optimisation to the compiler and have the meaning in the code.

src/plugins/audioreverb.cpp
```diff
@@ -2,206 +2,69 @@
...
 
-            if ( ( delay & 1 ) == 0 )
-            {
-                delay++;
-            }
+    // These buffers get filled with dry signal and are then passed to mverb
+    // They need to be vectors as the windows builds fail when arrays are used
```
pljones:
```suggestion
    // They need to be vectors as the Windows builds fail when arrays are used
```

src/plugins/audioreverb.cpp
```diff
@@ -2,206 +2,69 @@
...
+}
 
-    for ( int i = 0; i < 4; i++ )
+void CAudioReverb::loadPreset()
+{
+    for ( int i = 0; i < MVerb<float>::NUM_PARAMS; i++ )
     {
-        combDelays[i].Init ( lengths[i] );
-        combFilters[i].setPole ( 0.2f );
+        mverb->setParameter ( i, presets[iPreset][i] );
```
pljones, line 45: How do we know here that `iPreset` is valid?

src/plugins/audioreverb.h
```diff
@@ -2,56 +2,76 @@
...
 
 #pragma once
 #include "util.h"
+#include "libs/mverb/MVerb.h"
 
 class CAudioReverb
 {
 public:
-    CAudioReverb() {}
+    CAudioReverb()
```
pljones: Borderline case for moving this constructor into the cpp file.  I'm tempted to it should be.  I prefer only one-liner declarations with justified exceptions.

src/plugins/audioreverb.h
```diff
@@ -2,56 +2,76 @@
...
+        loadPreset();
+    }
 
-    void Init ( const EAudChanConf eNAudioChannelConf, const int iNStereoBlockSizeSam, const int iSampleRate, const float fT60 = 1.1f );
+    void Init ( const EAudChanConf eNAudioChannelConf, const int iNStereoBlockSizeSam );
 
     void Clear();
-    void Process ( CVector<int16_t>& vecsStereoInOut, const bool bReverbOnLeftChan, const float fAttenuation );
+    void Process ( CVector<int16_t>& vecsStereoInOut, const bool bReverbOnLeftChan, const float fReverbGain );
+    void setPreset ( const int iNPreset )
```
pljones: This one is definitely too complex for the .h.

src/plugins/audioreverb.h
```diff
@@ -2,56 +2,76 @@
...
+    int          iStereoBlockSizeSam;
+    float        fMaxShort;
+    int          iPreset;
+
+    int                numFrames;
+    std::vector<float> bufL;
+    std::vector<float> bufR;
 
-    class COnePole
+    enum
```
pljones: **Every** other enum is `enum ENameOfEnum` format.  That should be adhered to here.

src/plugins/audioreverb.h
```diff
@@ -2,56 +2,76 @@
...
-        COnePole() : fA ( 0 ), fB ( 0 ) { Reset(); }
-        void  setPole ( const float fPole );
-        float Calc ( const float fIn );
-        void  Reset() { fLastSample = 0; }
-
-    protected:
-        float fA;
-        float fB;
-        float fLastSample;
+        SUBTLE = 0,
```
pljones: I'm also a bit puzzled why there's no `OFF` preset that causes the code to bypass the reverb processing (and maybe have that disable the main dialog controls).

Maybe the settings page isn't quite right for this after all.  We've got the UI on the main panel.  Perhaps a pop-up chooser for the preset would be better than having to open Settings to change preset.

src/plugins/audioreverb.cpp
dingodoppelt: The constructor initialises a valid value and the setter checks for validity with `MathsUtils::InRange()`. Did I miss something?

src/plugins/audioreverb.h
dingodoppelt: > I'm also a bit puzzled why there's no `OFF` preset that causes the code to bypass the reverb processing (and maybe have that disable the main dialog controls).
> 
> Maybe the settings page isn't quite right for this after all. We've got the UI on the main panel. Perhaps a pop-up chooser for the preset would be better than having to open Settings to change preset.

Moving the preset chooser to the main UI and adding an `OFF` preset would make sense. Right now I think it can confuse people I the reverb fader just did nothing when `OFF` was selected (in the settings dialogue).

src/plugins/audioreverb.cpp
dingodoppelt: Considering the AI review it seems I did miss something

Jamulus.pro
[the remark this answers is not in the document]
dingodoppelt: See https://github.com/jamulussoftware/jamulus/pull/3787#discussion_r3790037071

This should be a separate PR. Builds for mverb work. C++17 is only set where needed at the moment.

src/plugins/audioreverb.h
```diff
@@ -2,56 +2,47 @@
...
+    float        fMaxShort;
+    int          iPreset;
+
+    int                numFrames;
+    std::vector<float> bufL;
+    std::vector<float> bufR;
+
+    // Parameters are set iteratively by enum. See MVerb.h for reference.
+    // NOTE: parameters "GAIN" and "MIX" must be "1."
+    constexpr static inline float const presets[RP_NUM_REV_PRESETS][MVerb<float>::NUM_PARAMS] = { { 0., .5, 1., .5, 0., .5, 1., 1., .75 },
```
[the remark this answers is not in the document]
dingodoppelt: See https://github.com/jamulussoftware/jamulus/pull/3787#discussion_r3790037071

src/plugins/audioreverb.h
ann0see: Only once we remove the reverb to settings this makes sense, I think.

src/plugins/audioreverb.h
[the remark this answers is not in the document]
ann0see: @coderabbitai please open an issue to investigate a global move to a more up to date C++ update. Say exactly what needs to be changed throughout the codebase. Thanks

src/util.h
```diff
@@ -639,6 +639,19 @@ enum EFeatureSet
     FS_RPC_ENABLED         = 12
 };
 
+// Reverb presets
```
pljones, line 642: Does this block want wrapping in
```
#ifndef NO_REVERB
...
#endif
```
?

src/plugins/audioreverb.h
pljones: > Right now I think it can confuse people I the reverb fader just did nothing when OFF was selected (in the settings dialogue).

I was thinking "OFF" would disable the slider (the UI does show a slider being disabled, just about).  If they'd had the slider not at zero and another preset selected, I really don't think switching it to "OFF" would be confusing in itself -- it's just that the slider should stop being available (and have no effect).  So "OFF" needs to stop the slider moving.

pljones:
Of course, one thing I've totally missed:
* JSON-RPC and GUI alignment.

As this is a substantial replacement of an existing UI and Client feature, it needs to have appropriate JSON-RPC support.

Now that really does open a can of worms.  If we're going to have pluggable effects, the JSON-RPC API needs to:
* probe the Client to determine what effects it has (initially just the reverb)
* probe what controls the effect exposes (preset and wet level for the reverb)
* send correctly-typed updates to the Client, targeting a specific effect, to adjust the controls (and the Client needs to make sure the UI reflects these)
* receive notifications from the Client for updates made in the UI to controls on a specific effect

dingodoppelt:
> Now that really does open a can of worms. If we're going to have pluggable effects, the JSON-RPC API needs to:
> 

That's why I stated we could alternatively remove the reverb, because it doesn't really belong here in the first place. See https://github.com/jamulussoftware/jamulus/pull/3787#issuecomment-5038723029  
I just worked on the reverb because it was already there. I'd vote against any DSP in Jamulus generally.

ann0see:
We should remain sensible with what we request as quality bar. I'm ok with not providing 100% json rpc support if it's a huge undertaking. Opening an issue documenting what's missing is appreciated though.

pljones:
Agreed -- it'll need doing before 4.0.0.

PR #3774, opened 2026-07-10
Make IPv4 and IPv6 use separate sockets

softins:
**Short description of changes**

Instead of using a single IP socket as dual-stack for IPv4 and IPv6, change to a separate socket for each protocol, but both contained within a single `CSocket`. This has the following advantages:
1. Ability to specify bind addresses for both IPv4 and IPv6.
2. Avoids the potential lock/hang issue with macOS dual-stack sending, possibly fixing #3758
3. No need to have a separate function to send IPv4 with TOS over an IPv6 socket on BSD variants

The sockets are set to non-blocking. In `OnDataReceived()`, use `poll()` to wait for either or both sockets to become ready, then read and process all available data before returning.

When sending a packet, the correct socket is selected depending on the protocol of the destination address.

CHANGELOG: Client/Server: Improve handling of IPv6 by using separate sockets

**Context: Fixes an issue?**

May fix #3758

**Does this change need documentation? What needs to be documented and how?**

New option `--serverbindip6` will need including in relevant documentation.

**Status of this Pull Request**

Work in progress

**What is missing until this pull request can be merged?**

Cross-platform compilation needs to be tidied up and tested, particularly in `socket.cpp`. Testing on all supported platforms. So far, only tested on a Mac with a variable-quality internet connection!

## Checklist

-  [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
-  [x] I tested my code and it does what I want
-  [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#source-code-consistency) 
-  [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
-  [x] I've filled all the content above

PR #3788, opened 2026-07-17
Atomic bRun in socket

softins:

Make the bRun boolean in the socket class atomic, for thread safety. Also remove the CSocket::Close method, as it is no longer needed with non-blocking socket reads.

CHANGELOG: Internal: Improve thread safety in socket handling

**Context: Fixes an issue?**

Replaces #3786 and expands on it, following on from #3774

No, bug fix and refactor only

Ready for review

Checking on all main platforms

## Checklist

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -531,7 +531,7 @@ bool CSocket::GetAndResetbJitterBufferOKFlag()
     return true;
 }
 
-void CSocket::OnDataReceived()
+void CSocket::OnDataReceived ( std::atomic<bool>& bRun )
 {
     /*
        The strategy of this function is that only the "put audio" function is
@@ -585,7 +585,7 @@ void CSocket::OnDataReceived()
             sockaddr_in sa4;
             socklen_t   sa4len = sizeof ( sa4 );
 
-            while ( true )
+            while ( bRun )
             {
                 const long iNumBytesRead =
                     recvfrom ( UdpSocket4, (char*) &vecbyRecBuf[0], MAX_SIZE_BYTES_NETW_BUF, 0, (struct sockaddr*) &sa4, &sa4len );
@@ -634,7 +634,7 @@ void CSocket::OnDataReceived()
             sockaddr_in6 sa6;
             socklen_t    sa6len = sizeof ( sa6 );
 
-            while ( true )
+            while ( bRun )
             {
                 const long iNumBytesRead =
                     recvfrom ( UdpSocket6, (char*) &vecbyRecBuf[0], MAX_SIZE_BYTES_NETW_BUF, 0, (struct sockaddr*) &sa6, &sa6len );
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -386,41 +386,6 @@ void CSocket::Init ( const quint16  iNewPortNumber,
     }
 }
 
-void CSocket::Close()
-{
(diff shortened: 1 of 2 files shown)


ann0see (commented):
Looks good but I'll need to test it on ARM and x86. 

We'll need to hunt for similar bugs then. Especially in the windows audio code.

src/socket.cpp
pljones, line 397: Which socket(s) are affected here?  For a Server, is it calling `WSACleanup()` on each Client disconnect, for example?

> WSACleanup terminates Windows Sockets operations for all threads

I'd be wary of doing that if so.  The CChannel CCSocket instances would get invalidated.

softins: No, even on a server all the traffic goes through a single `CSocket` instance, which is created at startup and persists until program termination. The WSA stuff only happens in the constructor and destructor for `CSocket`.

pljones:
I'll give the Windows Client and Linux Server a test today.

ann0see:
Actually, testing on iOS would also be beneficial

softins:
Worked fine on arm64 Pi. I'll test on x86 Mac later.

softins:
Seems to behave fine as both server and client on Mac x86 10.15

ann0see (approved):
I am now back on the initial network where I saw the hang, and it seems good now. 
Tested with macOS Apple Silicon

pljones:
Yeah, no problems between WIndows and Linux yesterday (both x64 machines, so unlikely to suffer).

pljones (approved):

ann0see:
@pljones 3.12.3?

pljones:
If it cherry-pick's cleanly, yes, I think this is worth having and small enough.  If it doesn't "just work" on the cherry-pick, we'll either have to pick it up separately or leave it.  I'd rather not muddy things too much.

(Basically that'd be my rule for things: does it cherry-pick cleanly.  Maybe needing a bit of help if the context has drifted but not a wholesale rewrite/refactor.)

I'll label it for now in the hope it works.

softins:
> If it cherry-pick's cleanly, yes, I think this is worth having and small enough. If it doesn't "just work" on the cherry-pick, we'll either have to pick it up separately or leave it. I'd rather not muddy things too much.

No, both the commits in this PR are dependent on the change to non-blocking sockets done as part of the IPv4/IPv6 split. I think I agree with @pljones that the latter doesn't want to be backported.

softins:
> > If it cherry-pick's cleanly, yes, I think this is worth having and small enough. If it doesn't "just work" on the cherry-pick, we'll either have to pick it up separately or leave it. I'd rather not muddy things too much.
> 
> No, both the commits in this PR are dependent on the change to non-blocking sockets done as part of the IPv4/IPv6 split. I think I agree with @pljones that the latter doesn't want to be backported.

We could just change the declaration of `bRun` to be `std::atomic<bool>` and leave everything else unchanged. That would at least fix the potential race condition.

pljones:
Following from reverting
```
 * 20697a7e 2026-08-02 Drop CSocketThread::bRun conversion, superseded by #3788 [jrd]
```
on `release/3_12`, we need to make progress getting this PR backported.  @softins are you okay to open a PR for it?

softins:
This PR doesn't need backporting. Apart from the atomic `bRun`, everything else in it is only applicable to the dual-socket architecture that is for 4.0.0 but not being backported.

PR #3789, opened 2026-07-17
Add AGENTS.md: a 22-line agent entry point

mcfnord:
Now a single new file, per review. AGENTS.md is 22 lines — short enough for any context window — and leads with hard PR-granularity rules: smallest independently mergeable unit, one PR per new file / independent fix / typo, every PR a standalone option reviewable in minutes. The rest is links to the existing docs plus the few facts an agent can't cheaply discover (no test suite → how to verify; real-time path; untrusted input; generated files; CHANGELOG/AUTOBUILD conventions).

It starts from @ann0see's #3785 draft — structure and much content survive. If you'd rather land AGENTS.md via #3785, I'm happy to close this one; maintainers' choice.

Everything else from earlier revisions of this PR is now its own single-file PR:
- #3790 — docs/DEPLOY.md
- #3791 — docs/ARCHITECTURE.md
- #3792 — CONTRIBUTING.md typo fix
- #3793 — CONTRIBUTING.md pointer paragraph for agent-assisted contributions

The GCC 13 moc issue formerly documented in COMPILING.md here will return as a build fix instead, per review.

CHANGELOG: SKIP

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3790, opened 2026-07-17
Add docs/DEPLOY.md: deploying a self-built server binary

mcfnord:
Split out of #3789 per review — one new file per PR.

DEPLOY.md covers what happens *after* COMPILING.md when you put a self-built headless server binary on production hosts: CPU architecture and glibc/libstdc++ matching between build and target machines, `ldd` verification, running under the shipped `linux/debian/jamulus-headless.service` unit, UDP receive-buffer tuning, firewall layering (cloud filters in front of host firewalls, JSON-RPC never internet-exposed), and a post-deploy verification checklist.

It is the distillation of operating self-built server binaries across Ubuntu 22.04/24.04/26.04, Oracle Linux, x86-64 and ARM, and sub-1GB-RAM machines. Every rule corresponds to a real outage I have caused or debugged.

Docs only; no code changes.

CHANGELOG: SKIP

### DIFF
--- a/docs/DEPLOY.md
+++ b/docs/DEPLOY.md
@@ -3,7 +3,18 @@
 Author(s):
 * The Jamulus Development Team
 
-Licensed under AGPL 3.0 or any later version. See [COPYING](../COPYING) for details.
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see [<https://www.gnu.org/licenses/>](https://www.gnu.org/licenses/).
 
 ---
 
--- a/docs/DEPLOY.md
+++ b/docs/DEPLOY.md
@@ -1,0 +1,58 @@
+### Copyright (c) 2026
+
+Author(s):
+* The Jamulus Development Team
+
+Licensed under AGPL 3.0 or any later version. See [COPYING](../COPYING) for details.
+
+---
+
+# Deploying a Jamulus Server
+
+[COMPILING.md](../COMPILING.md) ends when the binary exists. This document covers the step after that: putting a self-built headless server binary on a production host and verifying it actually runs. Most self-inflicted server outages happen in this step, and every rule below corresponds to a real-world failure.
+
+For configuring and operating a server (registration, recording, welcome message, etc.), see the [Server manual](https://jamulus.io/wiki/Running-a-Server).
+
+## Build for the target, not for the build machine
+
+- **CPU architecture must match.** A binary built on x86-64 will not run on an ARM host (and vice versa) — the service crash-loops with `Exec format error`. Before copying, compare `file ./Jamulus` on the build machine with `uname -m` on the target.
+- **Build on the oldest OS release you deploy to.** Binaries depend on the glibc/libstdc++ of the build machine. A binary built on Ubuntu 24.04 fails on 22.04 with `GLIBCXX_3.4.32 not found`, while a 22.04 build runs fine on 24.04 and later. Newer hosts run older binaries; the reverse never holds.
+- **Check shared libraries after every copy.** `ldd /path/to/jamulus | grep "not found"` must print nothing. This catches a missing Qt runtime package before systemd shows you a crash loop. The minimal headless runtime needs the Qt core, network, concurrent and xml libraries (see [COMPILING.md](../COMPILING.md)).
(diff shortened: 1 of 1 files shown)


docs/DEPLOY.md
```diff
@@ -0,0 +1,58 @@
+### Copyright (c) 2026
+
+Author(s):
+* The Jamulus Development Team
+
+Licensed under AGPL 3.0 or any later version. See [COPYING](../COPYING) for details.
+
+---
```
ann0see, line 19: I suppose this is not right - we need the full header.

docs/DEPLOY.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Deploying a Jamulus Server
+
+[COMPILING.md](../COMPILING.md) ends when the binary exists. This document covers the step after that: putting a self-built headless server binary on a production host and verifying it actually runs. Most self-inflicted server outages happen in this step, and every rule below corresponds to a real-world failure.
```
ann0see, line 23: Hmm. I think we should link the website directly. But I'll need to have a thought on this.

docs/DEPLOY.md
```diff
@@ -0,0 +1,58 @@
...
+
+For configuring and operating a server (registration, recording, welcome message, etc.), see the [Server manual](https://jamulus.io/wiki/Running-a-Server).
+
+## Build for the target, not for the build machine
+
+- **CPU architecture must match.** A binary built on x86-64 will not run on an ARM host (and vice versa) — the service crash-loops with `Exec format error`. Before copying, compare `file ./Jamulus` on the build machine with `uname -m` on the target.
+- **Build on the oldest OS release you deploy to.** Binaries depend on the glibc/libstdc++ of the build machine. A binary built on Ubuntu 24.04 fails on 22.04 with `GLIBCXX_3.4.32 not found`, while a 22.04 build runs fine on 24.04 and later. Newer hosts run older binaries; the reverse never holds.
```
ann0see, line 30: This is clearly just targeted at AI.

docs/DEPLOY.md
```diff
@@ -0,0 +1,58 @@
...
+- **Check shared libraries after every copy.** `ldd /path/to/jamulus | grep "not found"` must print nothing. This catches a missing Qt runtime package before systemd shows you a crash loop. The minimal headless runtime needs the Qt core, network, concurrent and xml libraries (see [COMPILING.md](../COMPILING.md)).
+- **Low-memory hosts:** on machines with ≤ 1 GB RAM, build with `make -j1`, or build on a bigger machine of the same OS/architecture and copy the binary. If you must add temporary swap to survive a build, remove it afterwards — see below.
+
+## Run under systemd
+
+Use the unit shipped in [`linux/debian/jamulus-headless.service`](../linux/debian/jamulus-headless.service) as your starting point — it already encodes hard-won defaults (dedicated `jamulus` user, `Nice=-20`, real-time I/O scheduling, `MemorySwapMax=0`, `Restart=on-failure`). Manage the server only through `systemctl`; never kill the process by PID.
```
ann0see, line 36: Add a README.md to linux/debian explaining this then (if it doesn't exist yet)

docs/DEPLOY.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Host tuning
+
+- **Enlarge the UDP receive buffer.** Under load, default kernel buffers drop packets, which musicians hear as dropouts. Set in `/etc/sysctl.d/99-jamulus.conf`:
```
ann0see, line 40: Sou you'd use an agent to deploy Jamulus to a machine automatically???

ann0see:
Not sure about those tbh. I'd rather add README.md files in the respective folders, link to the webpage or add comments to the files.

docs/DEPLOY.md
mcfnord: It's targeted at running a fleet, which AI makes much more realistic. some of my Linux instances are really tiny, and compiling on them is fraught with peril. So I've sorta taught it to build certain binaries in certain places and copy them along when done. All that version-specific gotcha stuff has been very useful in this. still seems like something a human would also want to be aware of.

docs/DEPLOY.md
mcfnord: Yes. I run 81 servers on 26 instances. I paste in a single authentication secret and then never visit the Linux instance again.

ann0see: TBH I'd rather use something deterministic like ansible for this.

ann0see:
Even though unrelated: could you run Faible on the whole code base and let it find concurrency bugs especially concerning missing std::atomic booleans. Open an issue with the findings and files only. That would be very much appreciated

docs/DEPLOY.md
mcfnord: Done in #3799 — one new file, linux/debian/README.md.

docs/DEPLOY.md
mcfnord: Fixed — full AGPL header now (also applied the same fix to docs/ARCHITECTURE.md in #3791).

ann0see:
I hope it's ok that I close this. I think that it's better to put the stuff into other files/READMEs. Then on your fork you can use this markdown as you need it

docs/DEPLOY.md
pljones: > TBH I'd rather use something deterministic like ansible for this.

Yes - or MicroK8s against a revision controlled deployment state.  I would trust an AI Agent any more than I'd trust a human agent.  You end believing both when they tell you something is working.  If you have to fix it in Git to get it working, it tends to stay working or you know why due to the revision history.

PR #3791, opened 2026-07-17
Add docs/ARCHITECTURE.md: system overview, source map, threading model

mcfnord:
Split out of #3789 per review — one new file per PR.

A single page answering the questions every new contributor (human or agent) otherwise reconstructs from scratch: how audio flows client → server → client, which class lives in which file, which threads exist and which of them are real-time, how a UDP "connection" works without a handshake, how directory servers and hole punching fit in, and where common kinds of changes go. It states two invariants: never block the real-time path, and treat all network input as untrusted.

Everything in it is verified against current `main` (`socket.cpp` `ProcessPacket()`, `CHighPrecisionTimer`/`OnTimer`, `CChannel`, `CServerListManager`).

Two things worth discussing here, per the comments in #3789:
- AGENTS.md stays extremely short and just links here; this file is the long-form home. If any part of this belongs in AGENTS.md instead (or vice versa), happy to move it.
- Agents can enumerate `src/` themselves, so the source map is aimed at humans — it carries the class→file responsibilities that enumeration doesn't reveal. Can trim if you prefer.

Docs only; no code changes.

CHANGELOG: SKIP

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -3,7 +3,18 @@
 Author(s):
 * The Jamulus Development Team
 
-Licensed under AGPL 3.0 or any later version. See [COPYING](../COPYING) for details.
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see [<https://www.gnu.org/licenses/>](https://www.gnu.org/licenses/).
 
 ---
 
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,0 +1,70 @@
+### Copyright (c) 2026
+
+Author(s):
+* The Jamulus Development Team
+
+Licensed under AGPL 3.0 or any later version. See [COPYING](../COPYING) for details.
+
+---
+
+# Jamulus Architecture
+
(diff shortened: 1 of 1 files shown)


ann0see (commented):
I suppose we'll need to make this more detailed. Also explain classes, ...

mcfnord:
Added a "Key classes and how they connect" section — the ownership tree (`CServer` → `vecChannels[]` → `SockBuf`/`Protocol`, plus `ConnLessProtocol`, and the client side) with a short responsibility note per class, member names as declared in the headers. Also upgraded to the full AGPL header per your comment on #3790.

Is that the level of detail you had in mind, or do you want it to go deeper (e.g. per-class method walkthroughs)? My instinct is to keep it at "what state lives where and who calls whom" and let the code carry the rest, so the doc doesn't rot.

pljones:
Whilst keeping this as the "this is the current architecture" guide, I'd like _somewhere_ that directs people on what the "target architecture" is: the clear separation of concerns as raised in https://github.com/jamulussoftware/jamulus/issues/3801.  I guess that Issue can have that document as a deliverable.

ann0see:
While definitely interesting, I think we need to make this lower priority. Focus on a lean AGENTS.md first, then add the ARCHITECTURE.md later.

mcfnord:
AGENTS.md should come first, sure. And it can stay lean, sure. But I support a more radical plan.

In 2003, I'd ship content on disks that couldn't be updated until the next release, next year.

So much has changed since then.

Facebook grew from a "move fast and break things" mindset.

Amazon says most decisions are two-way doors. You can go through. You can go right back. So why not just go through?

Taking a nod from those hyperscalers, we could throw everything and the kitchen sink into 3 or 4 Markdown files.

Immediately our surging agentic minions would start producing cleaner PRs.

Then we could describe our frustrations with those PRs, and rapidly update our guidance for agents.

Ready-fire-aim!

It would be glorious!

ann0see:
I suggest you iterate on it locally and then open an issue linking a proposal on your fork. I think we need to clean up and this is not yet in a state I'd be comfortable with. Thus closing

PR #3794, opened 2026-07-17
docs/JAMULUS_PROTOCOL.md: message reference, directory flows, small fixes

mcfnord:
**🤖 AI:** One file. Fills the main gaps in the protocol doc, verified against `src/protocol.cpp` / `src/socket.cpp` / `src/serverlist.cpp` on current main:

- How a receiver tells protocol packets from audio packets (they share one port; parse-or-audio in `CSocket::ProcessPacket()`), the CRC parameters, and how acknowledge/retransmit actually works — previously the doc said only "must be acknowledged".
- A complete message-ID reference table for both message classes, with the legacy IDs called out. This is the piece a Wireshark user or protocol reader currently has to reconstruct from `protocol.h`.
- A new section on directories: registration and refresh/timeout intervals, server-list responses (why both reduced and full lists are sent), and the `CLM_SEND_EMPTY_MESSAGE`/`CLM_EMPTY_MESSAGE` NAT hole-punch flow — previously undocumented here despite being a third of the connectionless messages.
- Fixes message names that drifted from `protocol.h`: `REQ_CHANNEL_INFOS`, `CHANNEL_INFOS`, `REQ_CONN_CLIENTS_LIST`, `REQ_SPLIT_MESS_SUPPORT`.

### Status

Documentation only; no linked issue and nothing on the website changes. Review threads from @pljones, @ann0see and @softins addressed in place; two open questions from @pljones remain on the thread. CodeRabbit's two findings (the `ACKN` exception to the acknowledge rule, exact message names in the flow diagram) fixed in [6b20e743](https://github.com/jamulussoftware/jamulus/pull/3794/commits/6b20e74398d610a07d4f465de5c185c5d8b153be).

CHANGELOG: SKIP

---

### DIFF
--- a/docs/JAMULUS_PROTOCOL.md
+++ b/docs/JAMULUS_PROTOCOL.md
@@ -120,7 +120,7 @@ Connection-based messages (acknowledged; `PROTMESSID_` prefix omitted). Full pay
 | 35 | `SPLIT_MESS_SUPPORTED` | Split messages are supported |
 | 36 | `RAWAUDIO_SUPPORTED` | Raw (uncompressed) audio is supported |
 
-IDs 12, 14, 15, 17, 19, 22 and 28 are legacy messages no longer sent (28, `REQ_CHANNEL_LEVEL_LIST`, is still understood for compatibility with Servers 3.4.6–3.5.12).
+IDs 12, 14, 15, 17, 19 and 22 are legacy messages no longer sent. `REQ_CHANNEL_LEVEL_LIST (28)` is still sent by the Client once per new connection, for compatibility with Servers 3.4.6–3.5.12; current Servers send `CLM_CHANNEL_LEVEL_LIST (1015)` without being asked.
 
 `SPECIAL_SPLIT_MESSAGE (2001)` sits outside both ID ranges because it is not a message in its own right: it is the transport container for the fragments of an oversized connection-based message (see the split message container above), with the original message ID carried inside the container. Each fragment frame has its own sequence counter and is acknowledged and retransmitted like any connection-based message.
 
@@ -168,15 +168,16 @@ The Server on a new Client connection will:
 
 This is defined in `CServer::OnNewConnection()`
 
-The Client on a new connection will:
+For the Client, a new connection begins when the first audio packet from the Server arrives (`CChannel::PutAudioData()` returns `PS_NEW_CONNECTION`). The Server's `REQ_*` messages above have been answered by then; independently of them, the Client will:
 
 - Send its Channel info with a `CHANNEL_INFOS (25, 0x1900)` message
 - Request the list of connected Clients with a `REQ_CONN_CLIENTS_LIST (16, 0x1000)` message
 - Set the Server-side Jitter Buffer value with a `JITT_BUF_SIZE (10, 0x0a00)` message
+- Request the channel level list with a `REQ_CHANNEL_LEVEL_LIST (28, 0x1c00)` message, for compatibility with Servers 3.4.6 to 3.5.12
 
-This is defined in `CClient::OnNewConnection()`
+This is defined in `CClient::OnNewConnection()`. The Client sends these unrequested because the Server may still hold the Client's channel from a previous session, in which case it would not treat this as a new connection and would not send its `REQ_*` messages.
 
-At the end of the session, the Client repeatedly sends a `CLM_DISCONNECTION (1010, 0xf203)` message, until the Server stops streaming audio to it.
+To disconnect, the Client stops sending audio and waits about 100 ms, answering every audio packet that still arrives from the Server in that window with a `CLM_DISCONNECTION (1010, 0xf203)` message; it then sends one final `CLM_DISCONNECTION` and is done (`CClient::Stop()`). The Server disconnects the channel as soon as it receives the message (`CServer::OnCLDisconnection()`). If the message is lost, the Server's channel times out after 30 seconds without audio from the Client (`CON_TIME_OUT_SEC_MAX`).
 
 A typical flow would be:
 
@@ -279,8 +280,8 @@ A Directory is a Jamulus Server acting as a registry (implemented in `src/server
 
 - A Server registers with `CLM_REGISTER_SERVER_EX (1017)` (older versions: `CLM_REGISTER_SERVER (1004)`) and receives `CLM_REGISTER_SERVER_RESP (1016)` carrying the result (registered, list full, version too old, requirements not fulfilled). If no response arrives, registration is retried every 500 ms, up to 5 times.
 - Registration is refreshed every 15 minutes; the Directory drops a Server it has not heard from for 33 minutes. `CLM_UNREGISTER_SERVER (1005)` removes the entry immediately at shutdown or when changing Directory through the Server UI.
-- A Client requests the list with `CLM_REQ_SERVER_LIST (1007)`. The Directory answers with both `CLM_RED_SERVER_LIST (1018)` (a shorter form that reduces UDP fragmentation) and `CLM_SERVER_LIST (1006)` (the full information). The Client then pings each listed Server with `CLM_PING_MS_WITHNUMCLIENTS (1002)` to display latency and occupancy.
-- NAT hole punching: when it answers a list request, the Directory also sends every registered Server a `CLM_SEND_EMPTY_MESSAGE (1008)` carrying the Client's public address; each Server responds by sending `CLM_EMPTY_MESSAGE (1009)` to that address, in order to open its own NAT/firewall for the Client's subsequent packets. It doesn't matter whether the Client receives this message, as it will ignore it. The Directory and its registered Servers also ping each other about once a minute to keep their NAT mappings alive.
+- A Client requests the server list with `CLM_REQ_SERVER_LIST (1007)`. The Directory answers with both `CLM_RED_SERVER_LIST (1018)` (a shorter form that reduces UDP fragmentation) and `CLM_SERVER_LIST (1006)` (the full information). The Client then pings each listed Server with `CLM_PING_MS_WITHNUMCLIENTS (1002)` to display latency and occupancy.
+- NAT hole punching: when it answers a server list request, the Directory also sends every registered Server a `CLM_SEND_EMPTY_MESSAGE (1008)` carrying the Client's public address; each Server responds by sending `CLM_EMPTY_MESSAGE (1009)` to that address, in order to open its own NAT/firewall for the Client's subsequent packets. It doesn't matter whether the Client receives this message, as it will ignore it. The Directory and its registered Servers also ping each other about once a minute to keep their NAT mappings alive.
 
 ---
 
--- a/docs/JAMULUS_PROTOCOL.md
+++ b/docs/JAMULUS_PROTOCOL.md
@@ -42,12 +42,17 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see [<https://www.gnu.org/licenses/>](https://www.gnu.org/licenses/).
 
-# The Jamulus Audio Protocol
+# The Jamulus audio protocol
 
-Jamulus uses connectionless UDP packets to communicate between the client and server, and additionally for directory server registration. The `src/protocol.cpp` file contains much of the details of the packets themselves, whereas this document is intended to form a higher-level view of the protocol interactions.
-Some of the messages need to be acknowledged, some do not. If a message ID is less than 1000, the message must be acknowledged in under `SEND_MESS_TIMEOUT_MS` ms.
+Jamulus uses UDP to communicate between the Client and Server, and additionally for registration with a Directory. The `src/protocol.cpp` file contains much of the details of the packets themselves, whereas this document is intended to form a higher-level view of the protocol interactions.
 
-All of this information can be discovered from reading the code, but hopefully is quicker to digest when available in one location. There is a wireshark dissector available too, [here](https://github.com/softins/jamulus-wireshark), if you would like to inspect the packet flow.
+UDP offers no delivery guarantee and no notion of a connection, so Jamulus layers its own session and reliability semantics on top of it. A Client and Server count as *connected* once the Client is sending valid audio packets and the Server has assigned it a channel. The two message classes below are named relative to that session, not to anything at the transport level.
+
+Messages with an ID below 1000 are connection-based: they apply to an established session, and each is acknowledged by an `ACKN (1)` message carrying the same sequence counter. Until that acknowledgement arrives, the sender retransmits the message every `SEND_MESS_TIMEOUT_MS` (400) ms. The protocol layer sets no retry limit: retransmission ends when the message is acknowledged, or when the channel clears the send queue via `CProtocol::Reset()` — on disconnect, on channel time-out, or when the protocol is disabled.
+
+Messages with an ID from 1000 to 1999 (`CLM_*`) are connectionless: they need no established session and are never acknowledged.
+
+All of this information can be discovered from reading the code, but hopefully is quicker to digest when available in one location. There is a Wireshark dissector available too, [here](https://github.com/softins/jamulus-wireshark), if you would like to inspect the packet flow.
 
 ---
 
@@ -68,6 +73,10 @@ LENgth of the data precedes the data and is followed by a CRC for the packet.
 
 Data is sent little-endian, i.e. not network byte-order.
 
+The CRC is 16 bits, generator polynomial x¹⁶ + x¹² + x⁵ + 1 (CCITT), initial state all ones, calculated over the entire message and transmitted inverted.
+
+Audio and protocol messages share one UDP port. A receiver classifies every incoming datagram by attempting to parse it as a protocol frame — zero TAG bytes, consistent length, valid CRC. Anything that fails this parse is treated as an audio packet (see `CSocket::ProcessPacket()` in `src/socket.cpp`).
+
 Where a message will not fit into the maximum packet size before fragmentation, a split message container is used.
 
 ```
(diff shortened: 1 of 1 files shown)


docs/JAMULUS_PROTOCOL.md
```diff
@@ -83,6 +87,64 @@ DATA is the fragment data to be re-assembled
...
+| 27 | `LICENCE_REQUIRED` | Server requires licence agreement |
+| 29 | `VERSION_AND_OS` | Version and operating system |
+| 30 | `CHANNEL_PAN` | Set a channel's pan in your mix |
+| 31 | `MUTE_STATE_CHANGED` | Your signal was (un)muted at another client |
+| 32 | `CLIENT_ID` | Your channel ID on the server |
+| 33 | `RECORDER_STATE` | Jam recorder state |
+| 34 | `REQ_SPLIT_MESS_SUPPORT` | Request split-message support |
+| 35 | `SPLIT_MESS_SUPPORTED` | Split messages are supported |
+| 36 | `RAWAUDIO_SUPPORTED` | Raw (uncompressed) audio is supported |
+| 2001 | `SPECIAL_SPLIT_MESSAGE` | Container for split messages |
```
pljones: Without commentary, this looks peculiar.

`2001` is a different type of message from the (other) connection-oriented messages.  I'd not put it in the same table.

docs/JAMULUS_PROTOCOL.md
```diff
@@ -205,6 +267,17 @@ Some typical messages could be:
 
 ---
 
+## Directory Registration and Server Lists
+
+A directory is a Jamulus server acting as a registry (implemented in `src/serverlist.cpp`, both roles). All directory traffic uses connectionless messages:
```
pljones: Can you check through and make sure the following adhere to the style guide:
- Client
- Server
- Directory

all always capitalised when referring to Jamulus "running as a...".  The style guide covers a few other things.

Probably applies to the other docs as well.

docs/JAMULUS_PROTOCOL.md
```diff
@@ -205,6 +267,17 @@ Some typical messages could be:
...
+
+- A server registers with `CLM_REGISTER_SERVER_EX (1017)` (older versions: `CLM_REGISTER_SERVER (1004)`) and receives `CLM_REGISTER_SERVER_RESP (1016)` carrying the result (registered, list full, version too old, requirements not fulfilled). If no response arrives, registration is retried every 500 ms, up to 5 times.
+- Registration is refreshed every 15 minutes; the directory drops a server it has not heard from for 33 minutes. `CLM_UNREGISTER_SERVER (1005)` removes the entry immediately at shutdown.
```
pljones:
```suggestion
- Registration is refreshed every 15 minutes; the directory drops a server it has not heard from for 33 minutes. `CLM_UNREGISTER_SERVER (1005)` removes the entry immediately at shutdown or when changing Directory through the Server UI.
```

docs/JAMULUS_PROTOCOL.md
```diff
@@ -68,6 +68,10 @@ LENgth of the data precedes the data and is followed by a CRC for the packet.
 
 Data is sent little-endian, i.e. not network byte-order.
 
+The CRC is 16 bits, generator polynomial x¹⁶ + x¹² + x⁵ + 1 (CCITT), initial state all ones, calculated over the entire message and transmitted inverted.
+
+Audio and protocol messages share one UDP port. A receiver classifies every incoming datagram by attempting to parse it as a protocol frame — zero TAG bytes, consistent length, valid CRC. Anything that fails this parse is treated as an audio packet (see `CSocket::ProcessPacket()` in `src/socket.cpp`).
```
ann0see, line 78: This will probably change soon since we have TCP support then. Probably worth thinking about rephrasing this.

docs/JAMULUS_PROTOCOL.md
```diff
@@ -83,32 +87,91 @@ DATA is the fragment data to be re-assembled
 
 This forms the data component of the packet above.
 
-## Client Session with a Server
+## Message reference
+
+Connection-based messages (acknowledged; `PROTMESSID_` prefix omitted). Full payload layouts are in the header comment of `src/protocol.cpp`.
+
+| ID | Name | Purpose |
```
ann0see, line 99: Thing is that this may get out of date. Maybe it's better to just link the source file. It should already have descriptions

ann0see (changes requested):

docs/JAMULUS_PROTOCOL.md
softins: I think it's good to have as part of this document. Once a message has been defined and merged to `main`, it is set in stone and will not change. The only out-of-date-ness would be if a new message was defined and didn't get added here. We can always add a comment in protocol.h reminding to update this file too.

docs/JAMULUS_PROTOCOL.md
softins: In the TCP PR, we can add a sentence or paragraph here explaining that certain messages can arrive over TCP, but any that don't parse correctly are dropped and not treated as audio.

docs/JAMULUS_PROTOCOL.md
```diff
@@ -83,32 +87,91 @@ DATA is the fragment data to be re-assembled
...
 
-- Tell the client connection its ID, with a `CLIENT_ID (32, 0x2000)` message.
-- Reset the connected client list with a `CONN_CLIENTS_LIST (24, 0x1800)` message.
-- Determine if the client supports split messages, with a `REQ_SPLIT_MESSAGE_SUPPORT (34, 0x2200)` message.
-- Request the details of the audio packets from the client with a `REQ_NETW_TRANSPORT_PROPS (21, 0x1500)` message,
-- Request the number of jitter buffer value to use, with a `REQ_JITT_BUF_SIZE (11, 0x0B00)` message.
-- Request the details of the channel info, with a `REQ_CHANNELS_INFOS (23, 0x1700)` message.
-- Send the version and OS of the server, with a `VERSION_AND_OS (29, 0x1d00)` message.
+- Tell the Client connection its ID, with a `CLIENT_ID (32, 0x2000)` message.
+- Reset the connected Client list with a `CONN_CLIENTS_LIST (24, 0x1800)` message.
```
softins: This could be interpreted to mean the server clears its own list. Suggest this clarification:

```suggestion
- Send the Client an empty connected Client list with a `CONN_CLIENTS_LIST (24, 0x1800)` message.
```

docs/JAMULUS_PROTOCOL.md
```diff
@@ -205,18 +268,29 @@ Some typical messages could be:
...
 ---
 
-## Audio Packet Structure
+## Directory registration and Server lists
+
+A Directory is a Jamulus Server acting as a registry (implemented in `src/serverlist.cpp`, both roles). All Directory traffic uses connectionless messages:
+
+- A Server registers with `CLM_REGISTER_SERVER_EX (1017)` (older versions: `CLM_REGISTER_SERVER (1004)`) and receives `CLM_REGISTER_SERVER_RESP (1016)` carrying the result (registered, list full, version too old, requirements not fulfilled). If no response arrives, registration is retried every 500 ms, up to 5 times.
+- Registration is refreshed every 15 minutes; the Directory drops a Server it has not heard from for 33 minutes. `CLM_UNREGISTER_SERVER (1005)` removes the entry immediately at shutdown or when changing Directory through the Server UI.
+- A Client requests the list with `CLM_REQ_SERVER_LIST (1007)`. The Directory answers with both `CLM_RED_SERVER_LIST (1018)` (a reduced form that avoids UDP fragmentation) and `CLM_SERVER_LIST (1006)`. The Client then pings each listed Server with `CLM_PING_MS_WITHNUMCLIENTS (1002)` to display latency and occupancy.
```
softins:
```suggestion
- A Client requests the list with `CLM_REQ_SERVER_LIST (1007)`. The Directory answers with both `CLM_RED_SERVER_LIST (1018)` (a shorter form that reduces UDP fragmentation) and `CLM_SERVER_LIST (1006)` (the full information). The Client then pings each listed Server with `CLM_PING_MS_WITHNUMCLIENTS (1002)` to display latency and occupancy.
```

docs/JAMULUS_PROTOCOL.md
```diff
@@ -205,18 +268,29 @@ Some typical messages could be:
...
+- NAT hole punching: when it answers a list request, the Directory also sends every registered Server a `CLM_SEND_EMPTY_MESSAGE (1008)` carrying the Client's public address; each Server responds by sending `CLM_EMPTY_MESSAGE (1009)` to that address, opening its own NAT/firewall for the Client's subsequent packets. The Directory and its registered Servers also ping each other about once a minute to keep their NAT mappings alive.
```
softins:
```suggestion
- NAT hole punching: when it answers a list request, the Directory also sends every registered Server a `CLM_SEND_EMPTY_MESSAGE (1008)` carrying the Client's public address; each Server responds by sending `CLM_EMPTY_MESSAGE (1009)` to that address, in order to open its own NAT/firewall for the Client's subsequent packets. It doesn't matter whether the Client receives this message, as it will ignore it. The Directory and its registered Servers also ping each other about once a minute to keep their NAT mappings alive.
```

docs/JAMULUS_PROTOCOL.md
```diff
@@ -205,18 +268,29 @@ Some typical messages could be:
...
+
+---
+
+## Audio packet structure
 
 The OPUS codec is used to compress the audio over the network and the packets are documented [here](https://datatracker.ietf.org/doc/html/rfc6716).
 
-Jamulus uses a custom OPUS encoder / decoder, giving some different frame sizes, but always uses a 48kHz sample rate. OPUS and OPUS64 codecs are the only supported options currently.
+Jamulus uses a custom OPUS encoder / decoder, giving some different frame sizes, but always uses a 48 kHz sample rate. OPUS and OPUS64 codecs are the only supported options currently.
```
softins: I guess we should mention the new raw audio option here now.

softins (commented):
This looks good. Just a few comments and suggested changes.

ann0see:
Please squash the commits here.

docs/JAMULUS_PROTOCOL.md
mcfnord: MY LLM WROTE:

Good catch — added: a sentence on the raw (uncompressed PCM) audio option, how it's advertised (`FS_RAW_AUDIO` in `CLM_SERVER_FEATURES`), how it's disabled (`--noraw`), and how the Server tells it apart from OPUS (packet-size comparison, no distinct message type).

docs/JAMULUS_PROTOCOL.md
```diff
@@ -42,12 +42,12 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see [<https://www.gnu.org/licenses/>](https://www.gnu.org/licenses/).
 
-# The Jamulus Audio Protocol
+# The Jamulus audio protocol
 
-Jamulus uses connectionless UDP packets to communicate between the client and server, and additionally for directory server registration. The `src/protocol.cpp` file contains much of the details of the packets themselves, whereas this document is intended to form a higher-level view of the protocol interactions.
-Some of the messages need to be acknowledged, some do not. If a message ID is less than 1000, the message must be acknowledged in under `SEND_MESS_TIMEOUT_MS` ms.
+Jamulus uses connectionless UDP packets to communicate between the Client and Server, and additionally for registration with a Directory. The `src/protocol.cpp` file contains much of the details of the packets themselves, whereas this document is intended to form a higher-level view of the protocol interactions.
```
pljones: (Nit: "connectionless UDP packets" is a horrible phrase...  UDP is a connectionless IP protocol, so "connectionless" is redundant.  It's also a packet-based protocol (inherently), so "packets" is redundant.  But there you are...)

docs/JAMULUS_PROTOCOL.md
```diff
@@ -42,12 +42,12 @@ GNU General Public License for more details.
...
+Messages with an ID below 1000 are connection-based: each one is acknowledged by an `ACKN (1)` message carrying the same sequence counter, and the sender retransmits it every `SEND_MESS_TIMEOUT_MS` ms until the acknowledgement arrives. Messages with an ID from 1000 to 1999 (`CLM_*`) are connectionless: they work without an established audio connection and are never acknowledged.
```
pljones: And then it has the nerve to go from saying "Jamulus uses connectionless..." to "connection-based"... and "... an established audio connection..." without any explanation.  I think a _little_ scene-setting is needed.

pljones: "the sender retransmits it every `SEND_MESS_TIMEOUT_MS` ms until the acknowledgement arrives"

Forever?  I know it's not a full spec but given it's got this partly stated, it might as well put the upper bound in.

docs/JAMULUS_PROTOCOL.md
mcfnord: MY LLM WROTE:

Agreed on both counts — dropped to just "Jamulus uses UDP to communicate...". Fixed in be4b4e71.

docs/JAMULUS_PROTOCOL.md
mcfnord: MY LLM WROTE:

Fair — it was using "connection-based" as a classification before establishing what a connection *is* here. Added a short paragraph between the two: UDP gives no delivery guarantee and no connection, so Jamulus layers its own session semantics on top; a Client and Server count as connected once the Client is sending valid audio packets and the Server has assigned it a channel. The two message-class names now have something to refer to, and I say explicitly they are named relative to that session rather than to anything at the transport level.

mcfnord: MY LLM WROTE:

I went to look up the bound and there is not one — so the honest answer to "Forever?" is very nearly yes, and the sentence was hiding that rather than just being incomplete.

`OnTimerSendMess()` (`protocol.h:332`) is just `SendMessage()`, and `SendMessage()` restarts `TimerSendMess` unconditionally whenever the queue is non-empty (`protocol.cpp:552-566`). No retry counter anywhere. Retransmission stops on exactly two events: a matching `ACKN` pops the queue (`protocol.cpp:711-716`), or `CProtocol::Reset()` clears it (`protocol.cpp:516`) — called from `channel.cpp` at 167, 394 and 678, i.e. channel disabled, protocol disabled while sending, and channel time-out/disconnect.

So the real bound is the channel lifetime, not a retry cap. Documented it that way rather than inventing a limit, and added the concrete 400 ms. If you would rather this were a bounded retry, that is a code change and I am happy to raise it separately.

docs/JAMULUS_PROTOCOL.md
```diff
@@ -44,8 +44,13 @@ along with this program.  If not, see [<https://www.gnu.org/licenses/>](https://
...
+Jamulus uses UDP to communicate between the Client and Server, and additionally for registration with a Directory. The `src/protocol.cpp` file contains much of the details of the packets themselves, whereas this document is intended to form a higher-level view of the protocol interactions.
+
+UDP offers no delivery guarantee and no notion of a connection, so Jamulus layers its own session and reliability semantics on top of it. A Client and Server count as *connected* once the Client is sending valid audio packets and the Server has assigned it a channel. The two message classes below are named relative to that session, not to anything at the transport level.
+
+Messages with an ID below 1000 are connection-based: they apply to an established session, and each is acknowledged by an `ACKN (1)` message carrying the same sequence counter. Until that acknowledgement arrives, the sender retransmits the message every `SEND_MESS_TIMEOUT_MS` (400) ms. The protocol layer sets no retry limit: retransmission ends when the message is acknowledged, or when the channel clears the send queue via `CProtocol::Reset()` — on disconnect, on channel time-out, or when the protocol is disabled.
```
pljones, line 51: Mm, leaves me asking how big an exploit surface that is -- I don't think it's that large as there aren't very many protocol messages that a malicious client could send to a server after initiating a connection that require a response that could have the ACKN unsent.  But a spoofable sender address would open this.

docs/JAMULUS_PROTOCOL.md
pljones: Thanks.

docs/JAMULUS_PROTOCOL.md
mcfnord: **🤖 AI:** Traced it. `CreateAndSendMessage()` (`protocol.cpp:577`) is the only entry point onto the retransmit-until-ACK queue (`SendMessQueue`), used by 21 distinct message types — jitter-buffer size, client ID, channel gain/pan, mute state, connected-client list, channel info, chat text, network-transport props, split-message support, raw-audio support, licence-required, opus-supported, channel-level-list request, version/OS, recorder state. Audio doesn't go through this path — it's UDP fire-and-forget, separate from the ACK'd control-message queue.

`SendMessQueue` (`protocol.h:321`, a `std::list`) has no size cap — nothing like `MAX_SIZE_MESS_QUEUE` exists. There's also no dedup: `EnqueueMessage()` (`protocol.cpp:519`) unconditionally `push_back()`s on every call, with no check for an existing entry of the same `iID`. ACK handling only pops `SendMessQueue.front()` when its `iCnt` and `iID` match the incoming `ACKN` (`protocol.cpp:709-719`) — strict FIFO, one message retransmitted at a time regardless of queue depth.

So repeated calls to the same setter do pile up, and there's a legitimate path that does exactly that: `CAudioMixerBoard::UpdateGainValue()` (`audiomixerboard.cpp:1858`) emits `ChangeChanGain` on every unsuppressed fader update, unthrottled, which reaches `CreateChanGainMes()` via `CChannel::SetRemoteChanGain()` (`channel.h:134`). Measured directly with a minimal headless harness linking the unmodified `protocol.cpp`/`.h` and leaving `MessReadyForSending` unconnected, so no `ACKN` is ever processed: 200 calls to `CreateChanGainMes()` on one channel grow `SendMessQueue` by exactly one entry per call, zero pops, ending at depth 200; repeated at 50000 calls, still perfectly linear with no cap encountered. Combined with the unbounded retransmit window from the earlier reply, a connection whose ACKs are being dropped or spoofed accumulates one queue entry per UI action indefinitely, not just per handshake step.

pljones: Ah, so _every_ connection-oriented message, regardless of whether the recipient _expects_ it (e.g. Client _should_ not send something to Server...) get an ACKN back anyway?

pljones: Anyway, it needs fixing.

mcfnord: **🤖 AI:** Yes, and it is wider than the message type. For a connected channel the acknowledgement is unconditional for anything that passes the frame and CRC check: `CreateAndImmSendAcknMess` sits outside the `switch ( iRecIDModified )` and outside the `if ( bEvaluateMessage )` guard, so no branch inside either can suppress it.

Now it is measured, against a server built from an unmodified `main`, over loopback. Each of these was sent to a connected channel 10 times and acknowledged all 10:

- `PROTMESSID_REQ_JITT_BUF_SIZE`, handled — the control
- `PROTMESSID_DISCONNECTION`, defined but with no `case`
- ID 99, not defined at all
- `PROTMESSID_CLIENT_ID`, which only ever travels server to client
- `PROTMESSID_JITT_BUF_SIZE` with a wrong-sized body
- one part of a 3-part split message, so never reassembled or evaluated

The last two are the ones worth noting: a body the evaluator rejects on size still gets acknowledged, and so does a split fragment that is never evaluated at all — the sender is told the message arrived and was accepted when nothing has read it.

Negative controls, same rig, 0/10 each: an `PROTMESSID_ACKN` itself, a frame with a corrupted CRC, and a connectionless ID.

One bound on it — this needs an established channel. An unconnected peer sending the same undefined ID gets 0/10, because the protocol path calls `FindChannel` without allowing a new channel, so there is nothing to parse into.

pljones: > the sender is told the message arrived and was accepted when nothing has read it.

That part is explicitly intentional in the design, I believe.

The only issue here is the "forever" part, really.

mcfnord:
**🤖 AI:** Squashed to a single commit. Ready for re-review.

docs/JAMULUS_PROTOCOL.md
pljones: .... which is under separate work ...

docs/JAMULUS_PROTOCOL.md
```diff
@@ -83,32 +92,91 @@ DATA is the fragment data to be re-assembled
...
+- Determine if the Client supports split messages, with a `REQ_SPLIT_MESSAGE_SUPPORT (34, 0x2200)` message.
+- Request the details of the audio packets from the Client with a `REQ_NETW_TRANSPORT_PROPS (21, 0x1500)` message,
+- Request the Jitter Buffer value to use, with a `REQ_JITT_BUF_SIZE (11, 0x0B00)` message.
+- Request the details of the Channel info, with a `REQ_CHANNEL_INFOS (23, 0x1700)` message.
+- Send the version and OS of the Server, with a `VERSION_AND_OS (29, 0x1d00)` message.
 
 This is defined in `CServer::OnNewConnection()`
 
-The client on a new connection will:
+The Client on a new connection will:
```
pljones, line 171: "On new connection" - i.e. regardless of the `REQ_xxx` messages coming from the Server?  Needs to be made clear either way.

Fundamentally what - for the Client - is a "new connection"?  It's clearly stated for the Server.  (Is it "when I start sending audio" or "when I get a Channel ID assigned" or something else...?)

docs/JAMULUS_PROTOCOL.md
```diff
@@ -83,32 +92,91 @@ DATA is the fragment data to be re-assembled
...
-- Request the list of connected clients with a `REQ_CONN_CLIENT_LIST (16, 0x1000)` message
-- Set the server-side jitter buffer value with a `JITT_BUF_SIZE (10, 0x0a00)` message
+- Send its Channel info with a `CHANNEL_INFOS (25, 0x1900)` message
+- Request the list of connected Clients with a `REQ_CONN_CLIENTS_LIST (16, 0x1000)` message
+- Set the Server-side Jitter Buffer value with a `JITT_BUF_SIZE (10, 0x0a00)` message
 
 This is defined in `CClient::OnNewConnection()`
 
-At the end of the session, the client calls the `CLM_DISCONNECTION (1010, 0xf203)` message, until the server stops streaming audio to it.
+At the end of the session, the Client repeatedly sends a `CLM_DISCONNECTION (1010, 0xf203)` message, until the Server stops streaming audio to it.
```
pljones, line 179: So
a) this is CLM so no ACKN but
b) it's up to the Server to stop sending or the Client is stuck?

docs/JAMULUS_PROTOCOL.md
```diff
@@ -205,18 +273,29 @@ Some typical messages could be:
...
 ---
 
-## Audio Packet Structure
+## Directory registration and Server lists
+
+A Directory is a Jamulus Server acting as a registry (implemented in `src/serverlist.cpp`, both roles). All Directory traffic uses connectionless messages:
+
+- A Server registers with `CLM_REGISTER_SERVER_EX (1017)` (older versions: `CLM_REGISTER_SERVER (1004)`) and receives `CLM_REGISTER_SERVER_RESP (1016)` carrying the result (registered, list full, version too old, requirements not fulfilled). If no response arrives, registration is retried every 500 ms, up to 5 times.
+- Registration is refreshed every 15 minutes; the Directory drops a Server it has not heard from for 33 minutes. `CLM_UNREGISTER_SERVER (1005)` removes the entry immediately at shutdown or when changing Directory through the Server UI.
+- A Client requests the list with `CLM_REQ_SERVER_LIST (1007)`. The Directory answers with both `CLM_RED_SERVER_LIST (1018)` (a shorter form that reduces UDP fragmentation) and `CLM_SERVER_LIST (1006)` (the full information). The Client then pings each listed Server with `CLM_PING_MS_WITHNUMCLIENTS (1002)` to display latency and occupancy.
```
pljones, line 282:
```suggestion
- A Client requests the server list with `CLM_REQ_SERVER_LIST (1007)`. The Directory answers with both `CLM_RED_SERVER_LIST (1018)` (a shorter form that reduces UDP fragmentation) and `CLM_SERVER_LIST (1006)` (the full information). The Client then pings each listed Server with `CLM_PING_MS_WITHNUMCLIENTS (1002)` to display latency and occupancy.
```
(my pedantry...)

docs/JAMULUS_PROTOCOL.md
```diff
@@ -205,18 +273,29 @@ Some typical messages could be:
...
+- NAT hole punching: when it answers a list request, the Directory also sends every registered Server a `CLM_SEND_EMPTY_MESSAGE (1008)` carrying the Client's public address; each Server responds by sending `CLM_EMPTY_MESSAGE (1009)` to that address, in order to open its own NAT/firewall for the Client's subsequent packets. It doesn't matter whether the Client receives this message, as it will ignore it. The Directory and its registered Servers also ping each other about once a minute to keep their NAT mappings alive.
```
pljones, line 283:
```suggestion
- NAT hole punching: when it answers a server list request, the Directory also sends every registered Server a `CLM_SEND_EMPTY_MESSAGE (1008)` carrying the Client's public address; each Server responds by sending `CLM_EMPTY_MESSAGE (1009)` to that address, in order to open its own NAT/firewall for the Client's subsequent packets. It doesn't matter whether the Client receives this message, as it will ignore it. The Directory and its registered Servers also ping each other about once a minute to keep their NAT mappings alive.
```

docs/JAMULUS_PROTOCOL.md
mcfnord: **🤖 AI:** Both questions are answered in the text now, from packet captures of a main build (72e856af) of client and server on loopback, not from the source alone. For the Client, a new connection is the first audio packet from the Server: in the capture `CHANNEL_INFOS` left within 1 ms of that packet, after the Server's `REQ_*` had already been answered, and `REQ_CHANNEL_LEVEL_LIST` goes out too, which the list omitted. On disconnect the Client answers packets still arriving during its 100 ms wait, then sends one final `CLM_DISCONNECTION` and stops: 277 further packets from a fake server drew 0 replies. The real Server stopped streaming within 1 ms of the first `CLM_DISCONNECTION`, and a SIGKILLed client was dropped 30.0 s after its last packet. Both suggestions applied.

docs/JAMULUS_PROTOCOL.md
mcfnord: **🤖 AI:** 

a) Yes, connectionless, so no `ACKN`. 

b) No, the Client is never stuck. [`CClient::Stop()`](https://github.com/jamulussoftware/jamulus/blob/c002c51f38cd75ea9c28834756ee453dad5adc25/src/client.cpp#L1096-L1130) disables the channel, waits about 100 ms, sends one `CLM_DISCONNECTION` and returns, whatever the Server does; its own comment says it never evaluates a reply. Measured against a fake server that ignores `CLM_DISCONNECTION` and keeps streaming: the Client answered the packets inside its 100 ms window, sent the final one, and 277 further packets over 3 s drew 0 replies. The real Server drops the channel within 1 ms of the first `CLM_DISCONNECTION` ([`OnCLDisconnection`](https://github.com/jamulussoftware/jamulus/blob/c002c51f38cd75ea9c28834756ee453dad5adc25/src/server.cpp#L558-L568)); if that message is lost, its 30 s timeout ([`CON_TIME_OUT_SEC_MAX`](https://github.com/jamulussoftware/jamulus/blob/c002c51f38cd75ea9c28834756ee453dad5adc25/src/channel.h#L66)) does it. The sentence quoted here was wrong on both halves and is rewritten at [line 180](https://github.com/mcfnord/jamulus/blob/6b20e74398d610a07d4f465de5c185c5d8b153be/docs/JAMULUS_PROTOCOL.md?plain=1#L180).

docs/JAMULUS_PROTOCOL.md
```diff
@@ -205,18 +274,29 @@ Some typical messages could be:
 
 ---
 
-## Audio Packet Structure
+## Directory registration and Server lists
+
+A Directory is a Jamulus Server acting as a registry (implemented in `src/serverlist.cpp`, both roles). All Directory traffic uses connectionless messages:
+
+- A Server registers with `CLM_REGISTER_SERVER_EX (1017)` (older versions: `CLM_REGISTER_SERVER (1004)`) and receives `CLM_REGISTER_SERVER_RESP (1016)` carrying the result (registered, list full, version too old, requirements not fulfilled). If no response arrives, registration is retried every 500 ms, up to 5 times.
+- Registration is refreshed every 15 minutes; the Directory drops a Server it has not heard from for 33 minutes. `CLM_UNREGISTER_SERVER (1005)` removes the entry immediately at shutdown or when changing Directory through the Server UI.
```
pljones:
```suggestion
- Registration is refreshed every 15 minutes; the Directory drops a Server it has not heard from for 33 minutes. `CLM_UNREGISTER_SERVER (1005)` removes the entry immediately at Server shutdown or when changing Directory through the Server UI.
```
Minor, just I had to read the sentence three times...

pljones (approved):
I'm happy as it is but for a tiny niggle.

docs/JAMULUS_PROTOCOL.md
```diff
@@ -42,12 +42,17 @@ GNU General Public License for more details.
...
+UDP offers no delivery guarantee and no notion of a connection, so Jamulus layers its own session and reliability semantics on top of it. A Client and Server count as *connected* once the Client is sending valid audio packets and the Server has assigned it a channel. The two message classes below are named relative to that session, not to anything at the transport level.
+
+Messages with an ID below 1000 are connection-based: they apply to an established session, and each is acknowledged by an `ACKN (1)` message carrying the same sequence counter. Until that acknowledgement arrives, the sender retransmits the message every `SEND_MESS_TIMEOUT_MS` (400) ms. The protocol layer sets no retry limit: retransmission ends when the message is acknowledged, or when the channel clears the send queue via `CProtocol::Reset()` — on disconnect, on channel time-out, or when the protocol is disabled. `ACKN` itself is the one exception in this range: it is sent immediately, never queued, and is neither acknowledged nor retransmitted.
```
ann0see: ACKs are Never ACK'd as this would lead to indefinitely ACK'ing that should be a general knowledge thing for a LLM.

docs/JAMULUS_PROTOCOL.md
```diff
@@ -83,32 +92,92 @@ DATA is the fragment data to be re-assembled
...
+| 29 | `VERSION_AND_OS` | Version and operating system |
+| 30 | `CHANNEL_PAN` | Set a Channel's pan in your mix |
+| 31 | `MUTE_STATE_CHANGED` | Your signal was (un)muted at another Client |
+| 32 | `CLIENT_ID` | Your Channel ID on the Server |
+| 33 | `RECORDER_STATE` | Jam recorder state |
+| 34 | `REQ_SPLIT_MESS_SUPPORT` | Request split-message support |
+| 35 | `SPLIT_MESS_SUPPORTED` | Split messages are supported |
+| 36 | `RAWAUDIO_SUPPORTED` | Raw (uncompressed) audio is supported |
+
+IDs 12, 14, 15, 17, 19 and 22 are legacy messages no longer sent. `REQ_CHANNEL_LEVEL_LIST (28)` is still sent by the Client once per new connection, for compatibility with Servers 3.4.6–3.5.12; current Servers send `CLM_CHANNEL_LEVEL_LIST (1015)` without being asked.
```
ann0see, line 123: We may want to drop this backwards compatibility one day.

docs/JAMULUS_PROTOCOL.md
```diff
@@ -205,18 +274,29 @@ Some typical messages could be:
 
 ---
 
-## Audio Packet Structure
+## Directory registration and Server lists
+
+A Directory is a Jamulus Server acting as a registry (implemented in `src/serverlist.cpp`, both roles). All Directory traffic uses connectionless messages:
```
ann0see: both roles should be made more explicit.

ann0see (approved):
I'd still like an explicit approval by @softins as he knows the protocol way better than I do.

mcfnord:
**🤖 AI:** @softins, the doc moved after the approvals above — two corrections, both in sentences you shaped.

The [raw audio option](https://github.com/jamulussoftware/jamulus/pull/3794#discussion_r3638680793) is described in the audio-packet section, but the new-connection list ran `CLIENT_ID (32)` straight into `CONN_CLIENTS_LIST (24)`. [`OnNewConnection()`](https://github.com/jamulussoftware/jamulus/blob/267bf6b762202954200839135778cdc194a2ab4e/src/server.cpp#L403-L414) sends `RAWAUDIO_SUPPORTED (36)` between them, unless [`--noraw`](https://github.com/jamulussoftware/jamulus/blob/267bf6b762202954200839135778cdc194a2ab4e/src/main.cpp#L519-L522) set `bDisableRaw`. Added there.

The [hole-punching sentence](https://github.com/jamulussoftware/jamulus/pull/3794#discussion_r3638670619) said the Directory sends `CLM_SEND_EMPTY_MESSAGE (1008)` to every registered Server. [`RetrieveAll()`](https://github.com/jamulussoftware/jamulus/blob/267bf6b762202954200839135778cdc194a2ab4e/src/serverlist.cpp#L733-L740) sends it only where [`IsPrivateNetworkIP()`](https://github.com/jamulussoftware/jamulus/blob/267bf6b762202954200839135778cdc194a2ab4e/src/serverlist.cpp#L722) rejects the registered address, and its loop starts at entry 1, so the Directory's own entry gets none either. Narrowed to match.

Both spotted by CodeRabbit.

PR #3795, opened 2026-07-17
docs/TRANSLATING.md: fix translators-list file reference

mcfnord:
One-line fix: TRANSLATING.md points translators at `src/utils.h`, which doesn't exist. The contributors/translators list lives in `src/util.cpp` in `CAboutDlg::CAboutDlg()` (matching what CONTRIBUTING.md says).

CHANGELOG: SKIP

### DIFF
--- a/docs/TRANSLATING.md
+++ b/docs/TRANSLATING.md
@@ -334,7 +334,7 @@ To add a new translation to the Installer (while undertaking the above steps wit
 
 Start translating the new language using Hosted Weblate or Git (as described in the first section of this guide). If you plan to work on the translation for a longer period, please contact the administrators to be invited to the Jamulus organisation as well as translators channel on Discord or Matrix.
 
-Since every contributor can and should be added to the in-app contributor list, please remind us to add you to the `src/utils.h` file as translator or open a pull request yourself.
+Since every contributor can and should be added to the in-app contributor list, please remind us to add you to the translators list in `src/util.cpp` (in `CAboutDlg::CAboutDlg()`) or open a pull request yourself.
 
 ---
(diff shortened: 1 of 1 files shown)


docs/TRANSLATING.md
```diff
@@ -334,7 +334,7 @@ To add a new translation to the Installer (while undertaking the above steps wit
 
 Start translating the new language using Hosted Weblate or Git (as described in the first section of this guide). If you plan to work on the translation for a longer period, please contact the administrators to be invited to the Jamulus organisation as well as translators channel on Discord or Matrix.
 
-Since every contributor can and should be added to the in-app contributor list, please remind us to add you to the `src/utils.h` file as translator or open a pull request yourself.
+Since every contributor can and should be added to the in-app contributor list, please remind us to add you to the translators list in `src/util.cpp` (in `CAboutDlg::CAboutDlg()`) or open a pull request yourself.
```
ann0see, line 337:
```suggestion
Since every contributor can and should be added to the in-app contributor list, please remind us to add you to the translators list in `src/util.cpp` or open a pull request yourself.
```

As the method name may get changed some time. As before, I expect humans and agents to find it e.g with grep. 
We might want to update the comment to say translators list

pljones: Ideally it wouldn't need a C++ file editing, it would be a plain text file preprocessed into a static value included at compile time...  That's for another time.   But my point is the file name could change just as easily.

pljones (approved):

ann0see (approved):
Approving for now. This might get stale if we change the location of the list.

PR #3799, opened 2026-07-17
Add linux/debian/README.md: packaging overview and headless service

mcfnord:
Requested by @ann0see in https://github.com/jamulussoftware/jamulus/pull/3790#discussion_r3606115668 ("Add a README.md to linux/debian explaining this then").

One new file. Explains what `deploy_deb.sh` builds from this directory (the two packages and their qmake configs), notes that `changelog` is generated at build time, and documents `jamulus-headless.service`: what it runs, as which user, and where to configure it — linking to the server options page on jamulus.io rather than duplicating it.

CHANGELOG: SKIP

### DIFF
--- a/linux/debian/README.md
+++ b/linux/debian/README.md
@@ -1,0 +1,34 @@
+# Debian packaging
+
+Author(s):
+* The Jamulus Development Team
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see [<https://www.gnu.org/licenses/>](https://www.gnu.org/licenses/).
+
+---
+
+## Files in this folder
+
+`linux/deploy_deb.sh` (run from the repository root) uses this directory to build two packages with `debuild`:
+
+- **jamulus** — the desktop client/server (built with `CONFIG+=noupcasename`, which renames the target from `Jamulus` to `jamulus`).
+- **jamulus-headless** — a server-only binary (`CONFIG+=headless serveronly` with `TARGET=jamulus-headless`) with no GUI library dependencies, plus a systemd service.
+
+Most files here are standard Debian packaging (`control`, `rules`, `copyright`, `*.install`, …). `changelog` is regenerated at build time from the top-level `ChangeLog` — don't edit it by hand.
+
+## jamulus-headless.service
+
+Installed to `lib/systemd/system` by the headless package. It runs `/usr/bin/jamulus-headless -s -n` as the unprivileged `jamulus` system user (created in `jamulus-headless.postinst`), restarts on failure, and raises CPU and I/O scheduling priority for real-time audio.
+
+Start it with `sudo systemctl enable --now jamulus-headless`. To name, publish or otherwise configure the server, adjust the `ExecStart=` options — see https://jamulus.io/wiki/Running-a-Server#server-mode-related-options.


linux/debian/README.md
```diff
@@ -0,0 +1,34 @@
...
+along with this program.  If not, see [<https://www.gnu.org/licenses/>](https://www.gnu.org/licenses/).
+
+---
+
+# Debian Packaging
+
+`linux/deploy_deb.sh` (run from the repository root) uses this directory to build two packages with `debuild`:
+
+- **jamulus** — the desktop client/server (built with `CONFIG+=noupcasename`).
+- **jamulus-headless** — a server-only binary (`CONFIG+=headless serveronly`) with no GUI library dependencies, plus a systemd service.
```
pljones: `noupcasename` missing here?

mcfnord: Not missing — `rules` sets `TARGET=jamulus-headless` explicitly for that build, which overrides the target name directly (`noupcasename` only renames the default `Jamulus` → `jamulus`). Updated the README to spell out both mechanisms so the asymmetry is clear.

pljones: Thanks.

I wonder why, though.  Unless there's a good reason, we should be consistent.  Is `TARGET` used elsewhere?  (I mean, we added `noupcasename` specifically for Debian, if I remember correctly.)

linux/debian/README.md
```diff
@@ -0,0 +1,34 @@
+### Copyright (c) 2026
```
ann0see:
```suggestion
### Debian packaging
```

linux/debian/README.md
```diff
@@ -0,0 +1,34 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see [<https://www.gnu.org/licenses/>](https://www.gnu.org/licenses/).
+
+---
+
+# Debian Packaging
```
ann0see:
```suggestion
## Files in this folder
```

linux/debian/README.md
```diff
@@ -0,0 +1,34 @@
+### Debian packaging
```
ann0see: Actually better

```suggestion
# Debian packaging
```

linux/debian/README.md
```diff
@@ -0,0 +1,34 @@
...
+along with this program.  If not, see [<https://www.gnu.org/licenses/>](https://www.gnu.org/licenses/).
+
+---
+
+## Files in this folder
+
+`linux/deploy_deb.sh` (run from the repository root) uses this directory to build two packages with `debuild`:
+
+- **jamulus** — the desktop client/server (built with `CONFIG+=noupcasename`, which renames the target from `Jamulus` to `jamulus`).
+- **jamulus-headless** — a server-only binary (`CONFIG+=headless serveronly` with `TARGET=jamulus-headless`, so `noupcasename` is not needed) with no GUI library dependencies, plus a systemd service.
```
ann0see:
```suggestion
- **jamulus-headless** — a server-only binary (`CONFIG+=headless serveronly` with `TARGET=jamulus-headless`) with no GUI library dependencies, plus a systemd service.
```

ann0see (changes requested):

linux/debian/README.md
mcfnord: > [!NOTE]

`TARGET` isn't used anywhere else in `Jamulus.pro` for naming — it's set in exactly two places: the `noupcasename` block (`TARGET = jamulus`) and the qmake command line in `rules` (`TARGET=jamulus-headless`). A qmake command-line variable always overrides an assignment in the `.pro` file, so passing `TARGET=jamulus-headless` on the command line makes `noupcasename` redundant for that build regardless — it would just get overridden right back. So there's no real inconsistency: `noupcasename` is the mechanism for the desktop build (there's no better name-changing hook at that call site), and an explicit `TARGET=` override is the mechanism for headless, because that build wants a completely different binary name, not just a case change. I've trimmed the README line per the suggestion below so it states the facts without asserting a design opinion — happy to spell the reasoning out inline instead if you'd rather have it in the doc than in this thread.

ann0see (commented):
Once everything is squashed into one commit I think this is ready.

linux/debian/README.md
pljones: It might be worth documenting in the `Jamulus.pro` rather than surfacing it to the debian-specific README.  Anyone maintaining `Jamulus.pro` would need to be aware of the intent.

Probably a separate PR, though, at this point.

ann0see (approved):

pljones (approved):

Issue #3798, opened 2026-07-17
Audit: non-atomic booleans and ints shared across threads (follow-up to #3786)


mcfnord:
**Background.** In #3786 the discussion converged on two points: `CSocketThread::bRun` is written by one thread and read by another, so it should be `std::atomic<bool>`; and the same pattern probably exists elsewhere in the codebase. I audited `main` (55e0d62d) file by file, tracing each shared flag against the threads that actually touch it. Below is everything I found, with file/line references only (no patch — that can follow separately if there's interest).

One data point up front: there are currently **zero** uses of `std::atomic` or `QAtomic*` anywhere in `src/`.

**Why these matter, briefly.** Each item is a data race in the C++ memory-model sense (concurrent access, at least one write, no synchronization), which is formally undefined behavior. In practice the realistic failure modes are: (a) the compiler legally caching or hoisting a repeated read (more likely under LTO/aggressive optimization), and (b) delayed or reordered visibility on weakly-ordered CPUs — which today includes Apple Silicon Macs, Android (Oboe), and ARM Linux servers such as Raspberry Pi. For the non-`bool` items, torn reads of multi-byte values are also possible. On x86 most of these "work" today, which is why they've been harmless-looking for years.

**Thread model as audited.**
- Client: main/Qt event thread (GUI, protocol — protocol messages arrive via queued signal, so parsing runs on main), socket thread (`CSocketThread`, time-critical), audio-driver callback thread (JACK/CoreAudio/ASIO/Oboe).
- Server: main thread (including `OnTimer` mixing, reached via queued signal from the timer thread, and the protocol slots which lock `Mutex`), `CHighPrecisionTimer` thread (non-Windows), socket thread, `CThreadPool` workers, recorder thread.

---

### A. Thread-lifecycle `bRun` flags (the exact #3786 class)

**A1. `CSocketThread::bRun` — `src/socket.h:241`.**
Written `false` by the main thread in `Stop()` (socket.h:213); read in the thread loop (socket.h:231) on the socket thread. This is the #3786 case, unchanged on `main`.

**A2. `CHighPrecisionTimer::bRun` (non-Windows) — `src/util.h:1355`.**
[...]

PR #3793, opened 2026-07-17
CONTRIBUTING.md: point agent-assisted contributors at AGENTS.md

mcfnord:
Split out of #3789 per review — one file per PR.

Adds one paragraph to CONTRIBUTING.md: AI coding agents start at AGENTS.md, everything in CONTRIBUTING.md applies to agent-assisted work without exception, and the contributor remains the author, expected to stand behind every line.

Depends on an AGENTS.md landing (#3789 or #3785) for the link to resolve.

CHANGELOG: SKIP

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3779, opened 2026-07-11
Investigate Windows Hang if using ASIO4ALL especially


ann0see:
**Describe the bug**
The client UI hangs (freezes indefinitely, not just briefly; not easily reproducible) on Windows when using the ASIO4ALL driver. I believe this happens when changing the input/output device or the input/output channel selection in the Settings dialog while connected/audio is running.

-- AI --
Static source analysis by Claude (see below) points to a cross-thread lock-order inversion on the ASIO backend's ASIOMutex. The device-change path re-initializes the ASIO buffers while holding ASIOMutex, and the driver's audio callback also needs ASIOMutex — if ASIO4ALL fires a buffer switch during buffer teardown/recreation, the two block each other permanently.
Note: This report is based on AI-assisted static code analysis. It was not reproduced or confirmed on Windows/ASIO4ALL by the author of this report but occurred last time I used the windows build. The stack-trace request under "Additional context" is intended to confirm it.

Key code facts:
- CSound::Init() holds ASIOMutex across driver calls ASIODisposeBuffers(), ASIOCreateBuffers(), ASIOGetLatencies() — src/sound/asio/sound.cpp:448-510.
- CSound::bufferSwitch() takes ASIOMutex for its entire body and does not check bRun — src/sound/asio/sound.cpp:701-1216.
- ASIO is the only backend that does this. CoreAudio checks pSound->bRun and uses MutexAudioProcessCallback (src/sound/coreaudio-mac/sound.cpp:1026,1086); JACK uses MutexAudioProcessCallback (src/sound/jack/sound.cpp:366). For ASIO, CSoundBase::Stop() locks MutexAudioProcessCallback (src/sound/soundbase.cpp:74) which the ASIO callback never takes, so it provides no callback synchronization — ASIO relies solely on ASIOMutex.tryLock(5000) (src/sound/asio/sound.cpp:534).
- Aggravator: CClient::Init() calls Sound.Init() 4× per change (src/client.cpp:1105-1107,1122), so each device/channel change triggers 4 dispose/create cycles, each of which can prompt ASIO4ALL to post kAsioResetRequest → queued reinit → repeat (reset storm). Device-change entry points: SetSndCrdDev (src/client.cpp:739), SetSndCrdLeft/RightInput/OutputChannel (:770,790,810,830), all of which do Stop → Init() → Start.
**To Reproduce**
1. Windows, Jamulus client, sound driver set to ASIO4ALL (with a real device routed through it).
[...]

PR #3800, opened 2026-07-17
Make cross-thread run/status flags std::atomic (#3798 batches A/B)

mcfnord:
This follows up on the discussion in #3786 and covers batches **A** and **B** of the audit in #3798 — the thread-lifecycle `bRun` flags and the jitter-buffer status flags. Offering it as an **experiment**: it is the least invasive slice of #3798, intended to establish the `std::atomic` pattern in the codebase and let us observe it through CI and normal use before deciding whether the more involved findings (C–F) are worth touching.

### What changes

Five members become `std::atomic<bool>`, keeping plain operator syntax at every call site (per the conclusion in #3786 that the overloaded operators are simplest, and the default sequentially-consistent operations cost nothing at these rates):

| Member | Written by | Read by |
|---|---|---|
| `CHighPrecisionTimer::bRun` (Mac/Linux variant) | main thread (`Start()`/`Stop()`) | timer thread loop; socket thread via `IsRunning()` |
| `CSoundBase::bRun` | main thread (`Start()`/`Stop()`) | audio callback thread (JACK/CoreAudio/Oboe/ASIO) |
| `CSoundBase::bCallbackEntered` | audio callback thread | GUI sound-device check |
| `CSocket::bJitterBufferOK` | socket thread | GUI read-and-reset |
| `CClient::bJitterBufferOK` | sound thread | GUI read-and-reset |

A sixth member, `CSocketThread::bRun` (the exact #3786 case), was originally part of this PR but has been dropped: #3788 already converts it — and removes `CSocket::Close()` in favour of non-blocking reads, which also disposes of finding F1 in #3798. The `<atomic>` include added to `socket.h` here matches #3788's placement so the two branches merge cleanly in either order.

The only non-declaration change: the two `GetAndResetbJitterBufferOKFlag()` functions now use `exchange ( true )` instead of a separate read and reset. Behavior is the same, and it closes a small lost-update window — a "not OK" written by the socket/sound thread between the read and the reset used to be dropped silently.

### Behavior

No intended behavior change. On x86 these compile to essentially the same instructions; the difference is that the compiler may no longer cache the flag across loop iterations, and on weakly-ordered CPUs (Apple Silicon, Android, ARM Linux servers) the cross-thread store is guaranteed to become visible and correctly ordered.

### Tested

- Full Linux client+server build (Qt 5.15, JACK enabled): compiles with no new warnings.
- Headless server smoke test: starts, runs, exits cleanly on SIGTERM — this exercises the modified `Stop()` paths of the socket and timer threads.
- macOS / Windows / Android builds not tested locally (relying on CI); the touched accesses in those backends are plain reads/writes covered by the operator overloads.

### DIFF
--- a/src/client.h
+++ b/src/client.h
@@ -51,6 +51,7 @@
 #include <QString>
 #include <QDateTime>
 #include <QMutex>
+#include <atomic>
 #ifdef USE_OPUS_SHARED_LIB
 #    include "opus/opus_custom.h"
 #else
@@ -431,9 +432,9 @@ protected:
     bool        bEnableAudioAlerts;
     bool        bEnableOPUS64;
 
-    bool   bJitterBufferOK;
-    bool   bMuteMeInPersonalMix;
-    QMutex MutexDriverReinit;
+    std::atomic<bool> bJitterBufferOK;
+    bool              bMuteMeInPersonalMix;
+    QMutex            MutexDriverReinit;
 
     // server settings
     int  iServerSockBufNumFrames;
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -518,17 +518,10 @@ void CSocket::SendPacket ( const CVector<uint8_t>& vecbySendBuf, const CHostAddr
 
 bool CSocket::GetAndResetbJitterBufferOKFlag()
 {
-    // check jitter buffer status
-    if ( !bJitterBufferOK )
-    {
-        // reset flag and return "not OK" status
-        bJitterBufferOK = true;
-        return false;
-    }
-
-    // the buffer was OK, we do not have to reset anything and just return the
-    // OK status
-    return true;
+    // atomically read the jitter buffer status and reset it to OK so that a
+    // "not OK" set by the socket thread between a separate read and reset
+    // cannot be lost
+    return bJitterBufferOK.exchange ( true );
 }
 
 void CSocket::OnDataReceived()
--- a/src/socket.h
+++ b/src/socket.h
@@ -49,8 +49,8 @@
 #include <QObject>
 #include <QThread>
 #include <QMutex>
-#include <atomic>
 #include <vector>
+#include <atomic>
 #include "global.h"
 #include "protocol.h"
 #include "util.h"
@@ -238,8 +238,8 @@ protected:
             }
         }
 
-        CSocket*          pSocket;
-        std::atomic<bool> bRun;
+        CSocket* pSocket;
+        bool     bRun;
     };
 
     void Init()
--- a/src/socket.h
+++ b/src/socket.h
@@ -49,6 +49,7 @@
 #include <QObject>
 #include <QThread>
 #include <QMutex>
+#include <atomic>
 #include <vector>
 #include "global.h"
 #include "protocol.h"
@@ -124,7 +125,7 @@ protected:
 
     bool bIsClient;
 
-    bool bJitterBufferOK;
+    std::atomic<bool> bJitterBufferOK;
 
     // This is a reference to CClient::bIPv6Available or CServer::bIPv6Available,
     // to inform the Client or Server which type of socket was created at startup.
@@ -237,8 +238,8 @@ protected:
             }
         }
 
-        CSocket* pSocket;
-        bool     bRun;
+        CSocket*          pSocket;
+        std::atomic<bool> bRun;
     };
 
     void Init()
--- a/src/util.h
+++ b/src/util.h
@@ -47,6 +47,7 @@
 #pragma once
 #include <vector>
 #include <algorithm>
+#include <atomic>
 #ifdef _WIN32
 #    include <winsock2.h>
 #    include <ws2tcpip.h>
@@ -1352,11 +1353,11 @@ public:
 protected:
     virtual void run();
 
-    bool     bRun;
+    std::atomic<bool> bRun;
 
 #    if defined( __APPLE__ ) || defined( __MACOSX )
-    uint64_t Delay;
-    uint64_t NextEnd;
+    uint64_t          Delay;
+    uint64_t          NextEnd;
 #    else
     long     Delay;
     timespec NextEnd;
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -631,14 +631,12 @@ bool CClient::GetAndResetbJitterBufferOKFlag()
     // get the socket buffer put status flag and reset it
     const bool bSocketJitBufOKFlag = Socket.GetAndResetbJitterBufferOKFlag();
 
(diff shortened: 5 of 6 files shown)


src/socket.h
```diff
@@ -237,8 +238,8 @@ class CHighPrioSocket : public QObject
             }
         }
 
-        CSocket* pSocket;
-        bool     bRun;
+        CSocket*          pSocket;
+        std::atomic<bool> bRun;
```
softins: Note that this one is already addressed as part of #3788, which is just awaiting a second approval.

ann0see:
Also you need to fix the coding style violations

mcfnord:
> Also you need to fix the coding style violations

This error class should be prevented when help about the formatting rules (or a link to the rules) appears in AGENTS.md. I don't see your draft of AGENTS.md as a PR, and I think you said you'd be providing one, so I remember yesterday my agent linked to my draft in a comment. For now, my CLAUDE.md has added the following guidance, including the note that eventually this will reach the AGENTS.md file when it exists:

## Upstream PRs (jamulussoftware/jamulus)

Upstream `CONTRIBUTING.md` is not auto-loaded context — read it before starting any upstream-bound change. (This section migrates to upstream `AGENTS.md` when that lands; the draft on the `agent-deploy-docs` branch already carries these rules, and PR #3793 adds the CONTRIBUTING.md pointer to it.)

**Style, before every commit**: `clang-format-14 -i` on all touched `.cpp/.h/.mm` (installed locally 2026-07-18; CI pins clang-format **14** in `.github/workflows/coding-style-check.yml`). `make clang_format` is the repo target but invokes unversioned `clang-format`, which this box doesn't have. Beware `AlignConsecutiveDeclarations`: widening one declaration's type re-pads its *neighbors* — this cost PR #3800 a red X.
- Qt ≥ 5.12.2 APIs only (`QT_VERSION_CHECK` guards); C++11; smallest-possible single-topic PRs; features agreed in an issue/Discussion before coding.
- Working pattern (issue #3798 → PR #3800): worktree off freshly fetched `main` in the scratchpad (never build upstream code in `/root/jamulus` — it clobbers the deploy-cached `./Jamulus`), full build + headless SIGTERM smoke test, push branch to fork, `gh pr create --repo jamulussoftware/jamulus --head mcfnord:<branch> --draft`.

src/socket.h
mcfnord: Thanks — dropped from this PR in favour of #3788 (kept the `bJitterBufferOK` conversion in the same file, and matched your `<atomic>` include placement so the branches merge cleanly in either order). Nice bonus: removing `CSocket::Close()` also disposes of finding F1 in #3798.

src/socket.cpp
```diff
@@ -518,17 +518,10 @@ void CSocket::SendPacket ( const CVector<uint8_t>& vecbySendBuf, const CHostAddr
...
-        return false;
-    }
-
-    // the buffer was OK, we do not have to reset anything and just return the
-    // OK status
-    return true;
+    // atomically read the jitter buffer status and reset it to OK so that a
+    // "not OK" set by the socket thread between a separate read and reset
+    // cannot be lost
+    return bJitterBufferOK.exchange ( true );
```
softins, line 524: This is nice, and more concise than the original!

mcfnord:
> [!NOTE]

Style violations fixed; all checks pass on the current head, including **Verify C-like coding style**.

Worth noting since it postdates this PR: `CSoundBase::bRun` is the declaration agreed in #3779 on 2026-07-21, and it is already in here. No separate PR needed for it. (This is the flag-visibility half only — the `ASIOMutex` lock-order problem in that thread is untouched.)

Unless you would rather I narrow this to that one declaration, I will drop "Experiment" from the title and treat it as ready for review.

mcfnord:
MY LLM WROTE:

Follow-up to the TSan evidence posted on #3802 (https://github.com/jamulussoftware/jamulus/pull/3802#issuecomment-5081840329): that run only tested #3800+#3802 merged together against unfixed `main`, so it didn't show whether this PR needs #3802 to fix its own claimed race. Built two more isolated worktrees on the same Pi 4 (same TSan flags, same load): `main` + only this PR (#3800) merged, and separately `main` + only #3802 merged.

**Result:** #3800 alone kills the `CHighPrecisionTimer::bRun` race completely — `Start()`'s write racing `isActive()`'s read, zero occurrences either side after the fix — and #3802 alone leaves that same race fully present, unaffected. This PR's fix stands on its own; it doesn't depend on #3802. (For completeness: #3802 alone independently kills its own `iConTimeOut` race and leaves `bRun` untouched — posted the mirror of this on #3802.)

Same caveat as before: raw TSan warning counts aren't comparable run-to-run (first-seen-per-race-pair reporting is noisy); the signal is presence/absence of this specific race's `SUMMARY` signature, and that's clean in both directions.

ann0see:
Once this is merged, you should rerun the tsan check on the fixed main branch.

ann0see (approved):

pljones:
Milestone on _this_ PR set to 4.0.0.  If and only if this can be cleanly cherry-picked after merge can it be tagged for backport to 3.12.4.

Otherwise a new PR will be needed that's clean against branch release/3_12.

src/util.h
```diff
@@ -1352,11 +1353,11 @@ class CHighPrecisionTimer : public QThread
 protected:
     virtual void run();
 
-    bool     bRun;
```
pljones, line 1355: Not part of all this -- we need better names than reusing `bRun` several times.  Either it should have a clear, distinctive name or it's the same data item and it's currently mismanaged.  I'm hoping it's the former...

(I'd only apply this to Class-scope variables: method variables can do whatever they want.)

src/client.h
```diff
@@ -431,9 +432,9 @@ class CClient : public QObject
     bool        bEnableAudioAlerts;
     bool        bEnableOPUS64;
 
-    bool   bJitterBufferOK;
```
pljones, line 434: Same here -- we shouldn't have two distinct concepts with the same identity (i.e. variable name).  Not part of this PR.

pljones (approved):

pljones:
@mcfnord, see https://github.com/jamulussoftware/jamulus/pull/3800#issuecomment-5133980771

I'll see if it cherry-picks cleanly for me and confirm but it won't get backported until there's a reply to @ann0see's comment.

pljones:
Oh, and could you rebase all your open PRs to HEAD of main of jamulussoftware/jamulus, please -- makes spotting the history of each a bit easier.  (Probably won't help resolving conflicts on the cherry-pick: it wasn't clean but I think I can fix it.)

pljones:
Backport merged based on @softins validation.

PR #3802, opened 2026-07-18
Make cross-thread audio params and channel/server flags std::atomic

mcfnord:
Second slice of the audit in #3798, following the same experimental approach as #3800: **only the mechanical conversions** from batches **C** (GUI-thread parameters read by the audio callback), **D** (client-side channel state), and **E** (the server disconnect flag). Everything that needs an actual design decision is deliberately left out — see "Not included" below.

### What changes

Nine members become `std::atomic`, keeping plain operator syntax at every call site (same rationale as #3800: per #3786, the overloaded operators are the simplest form, and default sequentially-consistent operations are negligible at these rates). No control flow changes at all this time — every access stays a plain load or store through the operators.

| Member | Written by | Read by |
|---|---|---|
| `CClient::bMuteOutStream` | main thread (setter) | sound thread, every frame |
| `CClient::fMuteOutStreamGain` | main thread (`SetRemoteChanGain`, under `MutexGainOrPan`) | sound thread, no lock |
| `CClient::iAudioInFader` | main thread (setter) | sound thread, every frame |
| `CClient::bReverbOnLeftChan` | main thread (setter) | sound thread, every frame |
| `CClient::iReverbLevel` | main thread (setter) | sound thread, every frame |
| `CClient::iInputBoost` | main thread (setter) | sound thread, every frame |
| `CChannel::bIsEnabled` | main thread (under `Mutex`) | socket thread in `PutAudioData()`, no lock |
| `CChannel::iConTimeOut` | main thread (`Disconnect()`, no lock) | socket/audio threads (under `MutexSocketBuf`), `IsConnected()` readers |
| `CServer::bChannelIsNowDisconnected` | thread-pool workers | main thread after `future.wait()` |

`fMuteOutStreamGain` is an addendum to the batch-C table in #3798, found while preparing this patch: it is written by the GUI thread under `MutexGainOrPan` but read lock-free by the sound thread in `ProcessAudioDataIntern()` — a writer-side-only mutex synchronizes nothing, and the `int` members in the same table can additionally tear. It is the same mechanical pattern, so it is included here.

The one comment change: the note above the `bChannelIsNowDisconnected` store (`server.cpp`) claimed "no mutex is needed … an atomic write". As discussed in #3786, a plain `bool` store is not an atomic write in the C++ memory model. With the member now `std::atomic<bool>`, the comment is reworded to be true.

### Not included (needs discussion, not mechanics)

- **C**: `SetReverbOnLeftChan()` also calls `AudioReverb.Clear()` on the GUI thread while the sound thread may be inside `AudioReverb.Process()` — concurrent mutation of filter state. The clean fix (deferring the clear onto the audio thread) changes behavior slightly and deserves its own review.
- **D3**: `iNetwFrameSize`/`iNetwFrameSizeFact` are written under `Mutex` but read under `MutexSocketBuf`; fixing that is a locking-design choice.
- **F**: the `Close()` fd overwrite race is the intended unblock mechanism and needs a considered design if it is to change at all.

### Behavior

No intended behavior change. The `iConTimeOut -= n` in `PutAudioData()` becomes an atomic `fetch_sub`, but that path already runs under `MutexSocketBuf`, so ordering there is unchanged; the fix is for the lock-free writers/readers elsewhere.

### Tested

- Full Linux client+server build (Qt 5.15, JACK enabled): compiles with no new warnings; `clang-format-14` clean.
- Headless server smoke test: starts, runs, exits cleanly on SIGTERM (`OnHandledSignal: 15`).

Note for whoever merges: #3800 also adds `#include <atomic>` to `src/client.h`, so whichever of the two lands second will need a trivial rebase of that one line.

### DIFF
--- a/src/channel.h
+++ b/src/channel.h
@@ -49,6 +49,7 @@
 #include <QThread>
 #include <QDateTime>
 #include <QFile>
+#include <atomic>
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 6, 0 )
 #    include <QVersionNumber>
 #endif
@@ -231,14 +232,14 @@ protected:
     // network protocol
     CProtocol Protocol;
 
-    int iConTimeOut;
-    int iConTimeOutStartVal;
-    int iFadeInCnt;
-    int iFadeInCntMax;
+    std::atomic<int> iConTimeOut;
+    int              iConTimeOutStartVal;
+    int              iFadeInCnt;
+    int              iFadeInCntMax;
 
-    bool bIsEnabled;
-    bool bIsServer;
-    bool bIsIdentified;
+    std::atomic<bool> bIsEnabled;
+    bool              bIsServer;
+    bool              bIsIdentified;
 
     int iNetwFrameSizeFact;
     int iNetwFrameSize;
--- a/src/client.h
+++ b/src/client.h
@@ -51,6 +51,7 @@
 #include <QString>
 #include <QDateTime>
 #include <QMutex>
+#include <atomic>
 #ifdef USE_OPUS_SHARED_LIB
 #    include "opus/opus_custom.h"
 #else
@@ -390,8 +391,8 @@ protected:
     EAudChanConf           eAudioChannelConf;
     int                    iNumAudioChannels;
     bool                   bIsInitializationPhase;
-    bool                   bMuteOutStream;
-    float                  fMuteOutStreamGain;
+    std::atomic<bool>      bMuteOutStream;
+    std::atomic<float>     fMuteOutStreamGain;
     CVector<unsigned char> vecCeltData;
 
     bool            bIPv6Available; // must be before Socket - passed by reference to Socket
@@ -402,11 +403,11 @@ protected:
 
     CVector<uint8_t> vecbyNetwData;
 
-    int          iAudioInFader;
-    bool         bReverbOnLeftChan;
-    int          iReverbLevel;
-    CAudioReverb AudioReverb;
-    int          iInputBoost;
+    std::atomic<int>  iAudioInFader;
+    std::atomic<bool> bReverbOnLeftChan;
+    std::atomic<int>  iReverbLevel;
+    CAudioReverb      AudioReverb;
+    std::atomic<int>  iInputBoost;
 
     int iSndCrdPrefFrameSizeFactor;
     int iSndCrdFrameSizeFactor;
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -944,9 +944,9 @@ void CServer::DecodeReceiveData ( const int iChanCnt, const int iNumClients )
 
                 FreeChannel ( iCurChanID ); // note that the channel is now not in use
 
-                // note that no mutex is needed for this shared resource since it is not a
-                // read-modify-write operation but an atomic write and also each thread can
-                // only set it to true and never to false
+                // note that no mutex is needed for this shared resource since it is a
+                // std::atomic write (not a read-modify-write operation) and also each
+                // thread can only set it to true and never to false
                 bChannelIsNowDisconnected = true;
 
                 // since the channel is no longer in use, we should return
--- a/src/server.h
+++ b/src/server.h
@@ -51,6 +51,7 @@
 #include <QHostAddress>
 #include <QFileInfo>
 #include <algorithm>
+#include <atomic>
 #ifdef USE_OPUS_SHARED_LIB
 #    include "opus/opus_custom.h"
 #else
@@ -254,10 +255,10 @@ protected:
     int    vecChannelOrder[MAX_NUM_CHANNELS];
     QMutex MutexChanOrder;
 
-    CProtocol ConnLessProtocol;
-    QMutex    Mutex;
-    QMutex    MutexWelcomeMessage;
-    bool      bChannelIsNowDisconnected;
+    CProtocol         ConnLessProtocol;
+    QMutex            Mutex;
+    QMutex            MutexWelcomeMessage;
+    std::atomic<bool> bChannelIsNowDisconnected;
 
     // audio encoder/decoder
     OpusCustomMode*    Opus64Mode[MAX_NUM_CHANNELS];


ann0see:
> The one comment change: the note above the bChannelIsNowDisconnected store (server.cpp) claimed "no mutex is needed … an atomic write". 

This is likely the comment I read a while ago which made me suspicious.

pljones:
> GUI-thread parameters read by the audio callback

I mean, this is fundamentally broken...  What does it do if there isn't a GUI?  State shouldn't in the GUI.  I'd rather the right fix was put in: use signals to pass state out of the GUI (not hold anything but the visual representation), respond to signals of change in state from the Client.

Or does this just mean "on the main thread"?

ann0see:
The question is if there's any visible performance impact...

ann0see:
I had a disconnect on a private server - which may need some more investigation. this might also be related to the Max setting where the jitter went red until leading to a disconnect. Actually, I think I observed this issue on this particular server without this fix applied so it's probably unrelated.

mcfnord:
> [!NOTE]

Measured it: no. Two members (`iInputBoost`, `fMuteOutStreamGain`) are read inside per-sample loops, which costs ~153 ns on a 2.67 ms frame — 0.006% of the budget — and I can hoist those two reads out of the loops if you want it back.

On the disconnect: both mutations of `iConTimeOut` (`channel.cpp:613` and `:641`) are already inside `MutexSocketBuf`, so this PR does not change time-out counting at all. It cannot have caused that.

ann0see:
If you can hoist something do it. 

I also think it's something else as I could reproduce it on 4.0.0 beta 1

mcfnord:
> [!NOTE]

Native TSan evidence on real weakly-ordered hardware (Raspberry Pi 4, Cortex-A72, aarch64 Linux) — the class of CPU the "theoretical race" argument is about, which x86 structurally can't demonstrate.

**Method:** built the headless server with `-fsanitize=thread`, once from `main` (unfixed) and once with this PR + #3800 merged in (fixed). Ran both under identical load — multithreaded server, 5 rounds of 8 bot clients each connecting/holding/disconnecting, then a clean `SIGTERM` shutdown.

**Result:** two of the audited flags raced concretely on the unfixed build and the race is gone after the fix:
- `CChannel::iConTimeOut` — a write in `ResetTimeOutCounter()` (socket-receive thread) against a read in `IsConnected()` (main/Qt-event thread, reached via `ProtocolIsEnabled`). This is the exact member the disconnect-timing question here was about — the mutation sites are already inside a mutex, but `IsConnected()` reads it unsynchronized, and that's what this PR's `std::atomic<int>` fixes. Confirmed via TSan: present pre-fix, zero occurrences post-fix.
- `CHighPrecisionTimer::bRun` (#3800) — same shape, `Start()`'s write racing `isActive()`'s read. Also present pre-fix, gone post-fix.

Total TSan warning counts aren't directly comparable run-to-run (it reports first-seen occurrence per race pair, so raw counts are noisy) — the signal is presence/absence of these specific races by their unique `SUMMARY` signature, and that's clean in both directions: gone after the fix, nothing newly broken.

Side finding, out of scope for this PR: `CNetBuf::Put` (buffer.cpp) flagged repeatedly in both builds — a real untouched race in the jitter buffer, unrelated to either atomics PR. Not investigated further, noting it in case it's useful.

mcfnord:
> [!NOTE]

Correction to my side note above: `CNetBuf::Put`/`Get` (buffer.cpp) is **not** a real race — retracting that.

Traced every flagged `SUMMARY` line back to its caller: they all resolve to `CChannel::PutAudioData` or `CChannel::GetData`, both of which already wrap the `SockBuf` access in `MutexSocketBuf.lock()`/`unlock()`. Both sides of every reported race are lock-guarded in source; TSan just doesn't see it, because Qt5's `QMutex` is futex-based and isn't intercepted the way `pthread_mutex` is. Same limitation as an earlier TSan investigation here (a separate, unrelated crash-repro effort logged 661 similarly QMutex-blind false positives). Consistent with that read: this only showed up in the *fixed* build's run, not the baseline's, at the same lock discipline — TSan's per-run nondeterminism, not a real difference.

Not proposing anything here, just correcting the record so it isn't mistaken for an open lead.

mcfnord:
MY LLM WROTE:

Follow-up to the TSan run posted above: that test only checked #3800+#3802 merged together against unfixed `main`, so it couldn't show whether either PR needs the other to fix its own claimed race. Built two more isolated worktrees on the same Pi 4, same TSan flags, same load: `main` + only this PR (#3802) merged, and separately `main` + only #3800 merged.

**Result:** #3802 alone kills the `iConTimeOut` race completely (`channel.h:101/102`, zero occurrences after the fix) — and #3800 alone leaves that same race fully present, unaffected. This PR's fix stands on its own; it doesn't depend on #3800. (For completeness: #3800 alone independently kills its own `CHighPrecisionTimer::bRun` race and leaves `iConTimeOut` untouched — posting the mirror of this on #3800.)

Same caveat as before: raw TSan warning counts aren't comparable run-to-run (first-seen-per-race-pair reporting is noisy); the signal is presence/absence of this specific race's `SUMMARY` signature, and that's clean in both directions.

ann0see (approved):
This might resolve #3779. Anyway, since std::atomic is from a theoretic background never worse (and based on the measurements the overhead seems basically non existent), I'm inclined to merge this.

pljones:
Milestone on _this_ PR set to 4.0.0.  If and only if this can be cleanly cherry-picked after merge can it be tagged for backport to 3.12.4.

Otherwise a new PR will be needed that's clean against branch release/3_12.

softins:
> > GUI-thread parameters read by the audio callback
> 
> I mean, this is fundamentally broken... What does it do if there isn't a GUI? State shouldn't in the GUI. I'd rather the right fix was put in: use signals to pass state out of the GUI (not hold anything but the visual representation), respond to signals of change in state from the Client.
> 
> Or does this just mean "on the main thread"?

Yes, I think that's exactly what it means: the main Qt event-handling thread, even if there is no GUI. It says they are declared and written within client.h, belonging to `CClient`.

However, it is probably the case that the setter methods are called from clientdlg.h/cpp via a `pClient` pointer, because it is only through the GUI that the user can request changes in those settings.

I think these atomics need to go in anyway, as whether the setters are called directly via a `pClient` pointer, or via slot handlers, they are still executed in the main thread, which is different from the real-time sound thread where the values are used.

softins (approved):
The changes look sensible to me, and I've done a smoke test with them for both server and client, so happy to approve.

pljones:
According to https://github.com/jamulussoftware/jamulus/issues/3779#issuecomment-5158583247, the suggested commit order wants #3800 first.

pljones:
OK, so #3800 is merged to `main` now.  Merging.

pljones:
Backported to release/3_12, clean compile and quick smoke test with the Linux Client.

PR #3803, opened 2026-07-18
Server: avoid per-frame deep copy of audio buffers in CreateLevelsForAllConChannels

mcfnord:
## What

`CServer::CreateLevelsForAllConChannels()` declares its `vecvecsData` parameter as

```cpp
const CVector<CVector<int16_t>> vecvecsData   // by value
```

This PR changes it to `const CVector<CVector<int16_t>>&` (by const reference), matching the neighbouring `vecNumAudioChannels` parameter.

## Why

The function is called from `CServer::OnTimer()` — the realtime audio timer thread — on **every frame in which at least one client is connected**:

```cpp
const bool bSendChannelLevels = CreateLevelsForAllConChannels ( iNumClients, vecNumAudioChannels, vecvecsData, vecChannelLevels );
```

Because `CVector` derives from `std::vector`, passing by value deep-copies the entire pre-allocated member vector every frame. `vecvecsData` is initialized in the `CServer` constructor to `iMaxNumChannels` (the `-u` setting) worst-case stereo buffers and is never shrunk, so the copy always covers the full configured channel limit — even a nearly empty server pays it in full. This both allocates on the audio thread (which the codebase otherwise takes care to avoid — see the "no memory must be allocated" comments in the `CServer` constructor) and burns memory bandwidth.

A small allocation-counting reproduction of the exact copy (150 channels = `-u 150`, 2×128 int16 per buffer, 375 frames/s) measured **~56,600 heap allocations/s and ~30 MB/s copied**, purely from the by-value parameter; at the default `-u 10` it is still ~4,100 allocations/s and ~2 MB/s. The copy is not elided at `-O2`.

## Change

One parameter, declaration + definition — no behavioural change. The function only reads `vecvecsData`.

## Testing

- Headless build (`CONFIG+=headless`, Qt 5.15) is clean.
- Server starts, runs, and shuts down cleanly on SIGTERM (`OnHandledSignal: 15`).
- `clang-format-14` applied to the touched files.

See #3804 for the analysis.

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -732,7 +732,7 @@ void CServer::OnTimer()
     if ( iNumClients > 0 )
     {
         // calculate levels for all connected clients
-        const bool bSendChannelLevels = CreateLevelsForAllConChannels ( iNumClients, vecNumAudioChannels, vecvecsData, vecChannelLevels );
+        const bool bSendChannelLevels = CreateLevelsForAllConChannels ( iNumClients );
 
         for ( int iChanCnt = 0; iChanCnt < iNumClients; iChanCnt++ )
         {
@@ -1673,10 +1673,7 @@ void CServer::customEvent ( QEvent* pEvent )
 }
 
 /// @brief Compute frame peak level for each client
-bool CServer::CreateLevelsForAllConChannels ( const int                        iNumClients,
-                                              const CVector<int>&              vecNumAudioChannels,
-                                              const CVector<CVector<int16_t>>& vecvecsData,
-                                              CVector<uint16_t>&               vecLevelsOut )
+bool CServer::CreateLevelsForAllConChannels ( const int iNumClients )
 {
     bool bLevelsWereUpdated = false;
 
@@ -1694,7 +1691,7 @@ bool CServer::CreateLevelsForAllConChannels ( const int                        i
                                                                                                                      vecNumAudioChannels[j] > 1 );
 
             // map value to integer for transmission via the protocol (4 bit available)
-            vecLevelsOut[j] = static_cast<uint16_t> ( std::ceil ( dCurSigLevelForMeterdB ) );
+            vecChannelLevels[j] = static_cast<uint16_t> ( std::ceil ( dCurSigLevelForMeterdB ) );
         }
     }
 
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1673,10 +1673,10 @@ void CServer::customEvent ( QEvent* pEvent )
 }
 
 /// @brief Compute frame peak level for each client
-bool CServer::CreateLevelsForAllConChannels ( const int                       iNumClients,
-                                              const CVector<int>&             vecNumAudioChannels,
-                                              const CVector<CVector<int16_t>> vecvecsData,
-                                              CVector<uint16_t>&              vecLevelsOut )
+bool CServer::CreateLevelsForAllConChannels ( const int                        iNumClients,
+                                              const CVector<int>&              vecNumAudioChannels,
+                                              const CVector<CVector<int16_t>>& vecvecsData,
+                                              CVector<uint16_t>&               vecLevelsOut )
 {
     bool bLevelsWereUpdated = false;
 
--- a/src/server.h
+++ b/src/server.h
@@ -240,10 +240,7 @@ protected:
     int                        iMaxNumThreads;
     CVector<std::future<void>> Futures;
 
-    bool CreateLevelsForAllConChannels ( const int                        iNumClients,
-                                         const CVector<int>&              vecNumAudioChannels,
-                                         const CVector<CVector<int16_t>>& vecvecsData,
-                                         CVector<uint16_t>&               vecLevelsOut );
+    bool CreateLevelsForAllConChannels ( const int iNumClients );
 
     // do not use the vector class since CChannel does not have appropriate
     // copy constructor/operator
--- a/src/server.h
+++ b/src/server.h
@@ -240,10 +240,10 @@ protected:
     int                        iMaxNumThreads;
     CVector<std::future<void>> Futures;
 
-    bool CreateLevelsForAllConChannels ( const int                       iNumClients,
-                                         const CVector<int>&             vecNumAudioChannels,
-                                         const CVector<CVector<int16_t>> vecvecsData,
-                                         CVector<uint16_t>&              vecLevelsOut );
+    bool CreateLevelsForAllConChannels ( const int                        iNumClients,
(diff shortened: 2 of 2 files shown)


ann0see:
Seems sensible. We'd need to check if there's really no reason for a copy though.

mcfnord:
Not sure how to parse the scope of **const** here, but it might be a clue that perhaps there's no reason to copy the bits.

softins:
Just been looking at this, and studying the life-cycle of `vecvecsData`.

1. `vecvecsData` is a protected member of class `CServer`. There is also a `vecvecsData2` which is specifically for delay panning, and is not relevant to this discussion.

2. The outer vector and inner vectors of `vecvecsData` are sized and allocated within the `CServer` constructor, and their sizes do not change after that.

3. `vecvecsData[n]` is written to only within `CServer::DecodeReceiveData()`, which is called by the static callback method `CServer::DecodeReceiveDataBlocks()`. That callback was in turn called from either:

   - `CServer::OnTimer()`, if multi-threading (`-T`) is not enabled, or
   - a thread in the threadpool, if multi-threading _is_ enabled.

4. `CServer::OnTimer()` is called via a signal and slot from `CServer::HighPrecisionTimer`. A normal timer calls its callback from the main application thread, but I'm not sure whether a high precision timer uses its own thread instead?

5. The method `CServer::CreateLevelsForAllConChannels()`, which is the subject of this PR, is called within `CServer::OnTimer()`. It is called with the following arguments:
   - `iNumClients`, by value, which is a variable local to the caller
   - `vecNumAudioChannels`, by reference, which is a member of `CServer`
   - `vecvecsData`, _by value_, which is a member of `CServer`
   - `vecChannelLevels`, by reference _to receive the calculated values_, which is a member of `CServer`.

   But since `CreateLevelsForAllConChannels()` is also a member of `CServer`, it has direct access to the protected or private members already, meaning there was no need for it to receive them as parameters! Certainly this applies to the two arguments passed by reference, as it is therefore accessing them directly via the reference.

   The situation with `vecvecsData` is currently different, because it is passed by value. This results in a copy being made of the outer vector and all its inner vectors, all `iMaxNumChannels` of them (`iMaxNumChannels` is set on the command line with `-u`, defaulting to 10 if not specified). Each inner vector contains 128 bytes of data, whether used or not.

6. In `CServer::CreateLevelsForAllConChannels()`, the only use made of `vecvecsData` is to pass each contained vector _by reference_ to `CChannel::UpdateAndGetLevelForMeterdB()`, which reads the data to calculate a dB value for the sound level.

7. So conceptually, `vecvecsData` could be passed to `CServer::CreateLevelsForAllConChannels()` by reference, and in fact the three vec parameters could be dispensed with and the method could access the class members directly. This latter seems to be most logical.

8. If multi-threading is not enabled, then the vectors within `vecvecsData` are both written to and read from within `CServer::OnTimer()`, so there should be no contention.

9. However, it multi-threading is enabled, the inner vectors of `vecvecsData` will be updated by a different thread, which could be on a different CPU or core. This could conceivably happen while `CServer::CreateLevelsForAllConChannels()` is processing it. The copying of `vecvecsData` when passing by value probably mitigates this, but I couldn't be sure it guards absolutely against another thread updating one of the inner vectors while it is being copied.

10. I think logically, we don't want to be making deep copies of `vecvecsData`, but we do need to guard against concurrent access to its constituent vectors. I initially wondered whether we could change `vecvecsData` from `CVector<CVector<int16_t>>` to `CVector<std::atomic<CVector<int16_t>>>`. But apparently, `std::atomic<T>` can only be used if the type `T` is trivially copyable, and does not contain dynamic elements such as `std::vector`.

11. Otherwise, we might need to introduce mutex protection of either `vecvecsData` or more finely-grained of its constituent vectors?

12. Instead of passing the whole `vecvecsData` by value, maybe it would be better in step 6 above to pass the inner vector by value, or make a temporary copy of it with mutex protection and then pass the copy by reference. This would make any locking much more finely grained, and copying to be only necessary for channels that are currently is use.

Comments?

softins:
OK, I've now just read the analysis on #3804, which I'd overlooked before. It largely aligns with what I came up with above (without using AI!), but reassures me that the writing and reading are protected against concurrent access.

If that's the case, I would propose removing the vec parameters from `CServer::CreateLevelsForAllConChannels()` and replacing them with direct access to the relevant class members.

ann0see:
I believe as long as threads do not write to the same index of the array, we would not even need locking (but some kind of final synchronization after all threads have finished).

ann0see:
Seems like the function got introduced in this PR: https://github.com/jamulussoftware/jamulus/pull/42 CC @pljones if he has background on why this was a copy.

mcfnord:
The requested deeper analysis is now on the issue: https://github.com/jamulussoftware/jamulus/issues/3804#issuecomment-5017205311 — git archaeology of how the copy became per-frame (it was gated and outside the mutex in #42), a memory-model review of the `-T` decode path (the futures join inside the locker is the synchronization; the copy adds none), and a live A/B measurement: at `-u 150` this parameter was 98% of the server's allocator traffic and ~23% of its CPU. Also addressed there: @softins's direct-member-access proposal (happy to rework this PR to that shape if preferred — it matches how `MixEncodeTransmitData` already accesses the same member).

pljones:
> Seems like the function got introduced in this PR: #42 CC @pljones if he has background on why this was a copy.

Probably on the basis that max clients wasn't so high and it only happened once every 200 frames?  It's definitely read-only so _should_ be able to work on a reference (in _takes_ a reference as the first thing inside its main loop: that should have triggered "why not pass references, then?" I guess).

pljones (approved):

softins:
I pushed a commit to this PR that removed the vec parameters and used direct member access instead. It seems to work fine in my test server. @mcfnord, see what you think.

@pljones I think this could/should go into 4.0.0 - we will have more betas/rcs before release, which would be plenty of time to find any unlikely issues. EDIT: ah, I see you changed it to 4.0.0 while I was typing. Cool.

softins (approved):
Approving, even though one of the commits is mine.

mcfnord:
> @mcfnord, see what you think.

I am just the messenger.

ann0see:
Then we should probably add Claude as contributor too not sure.

ann0see:
3.12.3? @pljones ?

pljones:
Should just work.  I'll label it.  If there's a problem with the cherry-pick, drop the label.

pljones:
Linux build (nothing taken out) run against backport cleanly.  Smoke test ("it runs without odd messages") on GUI Client and Server.  Merged to release/3_12.

Issue #3801, opened 2026-07-18
Unify CClient/CServer as a single orchestration layer behind UI and JSON-RPC


pljones:
## Summary

Currently, Client and Server capabilities and states are scattered across UI and other classes rather than "owned" centrally.

This issue proposes refactoring so that `CClient` and `CServer` become the single source of truth for all state and actions, with both the Qt UI and the JSON-RPC interface acting as symmetric, parity consumers of that layer - nothing more.

(This came out of discussion on #3660 - and prior discussions - where a small RPC addition (`getCurrentDirectory`/`getDirectories`) surfaced duplicated logic between `CConnectDlg` and `CClientRpc` that should really live in one place.)

## Goals

- **Clear separation of concerns** between orchestration logic and presentation/control interfaces.
- **`CClient` and `CServer` as the orchestration core** - the single arbiter of all Client/Server capabilities, state, and actions.
- **Full parity between UI and JSON-RPC.** Anything the UI can do, JSON-RPC can do, and vice versa. No interface-specific logic duplicated across them. This should be sufficient, for example, to let someone write a local JSON-RPC + React (or any other) frontend that fully replaces the Qt UI.
- **`CSettings` / `CClientSettings / CServerSettings` encapsulated behind `CClient` and `CServer`**, rather than read or mutated directly by UI or RPC code.

## Proposed pattern

Using "current directory" as a worked example:

1. `CClient` owns the state (e.g. current directory index) and exposes it via a canonical accessor.
2. UI components (e.g. `CConnectDlg`) emit a signal on user interaction (e.g. "selection changed").
3. `CClient` handles the signal and updates `CClientSettings`.
4. `CClientSettings` emits its own change signal.
5. Both the UI and `CClientRpc` listen and react independently — UI updates its display, RPC pushes new state to subscribers.

Neither the UI nor RPC owns logic the other doesn't have equal access to; both sit on top of the same `CClient`-owned state and signals.

## Scope of JSON-RPC control

There is no remaining debate on *how much* should be controllable via JSON-RPC: full parity with the UI is the goal, including actions that affect other clients on a server (e.g. join/leave, mute). Restricting RPC to read-only, or to a subset of UI actions, would defeat the purpose — it would prevent RPC or other future mechanisms from being a viable full replacement for the UI.

## Scope of exposure (local control)

[...]

PR #3805, opened 2026-07-19
Refactor Connect/Disconnect out to CClient with an explicit connection state

### LINKED ISSUE #3367: Refactor Connect() and Disconnect() functions in GUI to allow JSON RPC to update UI

mcfnord:

This picks up and finishes #3372 (@ann0see's extract of @pgScorpio's work in #2550), which had stalled on review discussion. It contains two commits:

1. **The original #3372 refactor, ported onto current `main`** (authorship preserved): `Connect()` / `Disconnect()` move out of `CClientDlg` into `CClient`, which emits signals the dialog consumes. `CClientDlg` no longer holds connection logic — it only reacts (`OnConnecting`, `OnConnectingFailed`, `OnDisconnect`). Connect-on-startup moves out of the dialog constructor into `main.cpp` via `Client.Connect(...)`, so it now also works identically in `--nogui` mode.

2. **An explicit connection state machine**, addressing the state-model review comments in #3372: a new `EConnectionState` (`CS_DISCONNECTED` / `CS_CONNECTING` / `CS_CONNECTED`) owned by `CClient` as the single source of truth, with every transition emitting `ConnectionStateChanged`. Following that discussion's definitions: a connection is *requested* (`CS_CONNECTING`) when the audio stream starts toward the configured server, and *established* (`CS_CONNECTED`) once the server assigns our channel ID. Accordingly, the signal formerly emitted from `Start()` as `Connected(name)` is renamed `Connecting(name)`, since at that point the connection is only requested. `CClient::Connect()` now terminates any current connection first, so connecting while connected behaves as a clean reconnect.

This is a step toward #3801 (CClient as the orchestration layer behind UI and JSON-RPC): a follow-up PR stacked on this one exposes connect/disconnect/state over JSON-RPC as symmetric consumers of the same signals.

**Context: Fixes an issue?**

Fixes #3367. Supersedes #3372 (both commits build on it; the first preserves its authorship).

No user-facing behavior change. The follow-up JSON-RPC PR updates `docs/JSON-RPC.md`.

Ready for review.

Review.

Tested: full GUI build and `CONFIG+=headless` build on Linux/Qt 5.15; a `--nogui` client was driven through connect → reconnect-while-connected → disconnect cycles against a local server, observing the `Connecting`/`ConnectionStateChanged`/`Disconnected` signal sequence via the follow-up PR's JSON-RPC notifications (16/16 scripted checks pass); clean SIGTERM shutdown verified.

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#source-code-consistency)

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -1159,7 +1159,7 @@ void CClient::Disconnect()
 ///       Use to display error message in CClientDlg
 /// @param strServerAddress - the server address to connect to
 /// @param strServerName - the human readable server name passed to Connecting()
-void CClient::Connect ( QString strServerAddress, QString strServerName )
+void CClient::Connect ( const QString& strServerAddress, const QString& strServerName )
 {
     try
     {
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -145,7 +145,7 @@ CClient::CClient ( const quint16  iPortNumber,
     // The first ConClientListMesReceived handler performs the necessary cleanup and has to run first:
     QObject::connect ( &Channel, &CChannel::ConClientListMesReceived, this, &CClient::OnConClientListMesReceived );
 
-    QObject::connect ( &Channel, &CChannel::Disconnected, this, &CClient::Disconnected );
+    QObject::connect ( &Channel, &CChannel::Disconnected, this, &CClient::Stop );
 
     QObject::connect ( &Channel, &CChannel::NewConnection, this, &CClient::OnNewConnection );
 
@@ -626,6 +626,9 @@ bool CClient::SetServerAddr ( QString strNAddr )
         // apply address to the channel
         Channel.SetAddress ( HostAddress );
 
+        // By default, set server name to HostAddress. If using the Connect() method, this may be overwritten
+        SetConnectedServerName ( HostAddress.toString() );
+
         return true;
     }
     else
@@ -929,11 +932,8 @@ void CClient::OnHandledSignal ( int sigNum )
     {
     case SIGINT:
     case SIGTERM:
-        // if connected, terminate connection (needed for headless mode)
-        if ( IsRunning() )
-        {
-            Stop();
-        }
+        // if connected, Stop client (needed for headless mode)
+        Stop();
 
         // this should trigger OnAboutToQuit
         QCoreApplication::instance()->exit();
@@ -1069,8 +1069,14 @@ void CClient::Start()
     // Disable hibernation or display dimming if the app is running on Windows
     SetThreadExecutionState ( ES_CONTINUOUS | ES_SYSTEM_REQUIRED | ES_DISPLAY_REQUIRED );
 #endif
+
+    emit Connected ( GetConnectedServerName() );
 }
 
+/// @method
+/// @brief Stops client and disconnects from server
+/// @emit Disconnected
+///       Use to set CClientDlg to show not being connected
 void CClient::Stop()
 {
     // stop audio interface
@@ -1112,6 +1118,53 @@ void CClient::Stop()
     // Allow hibernation or display dimming if the app is running again (Windows)
     SetThreadExecutionState ( ES_CONTINUOUS );
 #endif
+
+    // emit Disconnected() to inform UI of disconnection
+    emit Disconnected();
+}
+
+/// @method
+/// @brief Stops the client if the client is running
+/// @emit Disconnected
+void CClient::Disconnect()
+{
+    if ( IsRunning() )
+    {
+        Stop();
+    }
+}
+
+/// @method
+/// @brief Connects to strServerAddress
+/// @emit Connected (strServerName) if the client wasn't running and SetServerAddr was valid. emit happens through Start().
+///       Use to set CClientDlg to show being connected
+/// @emit ConnectingFailed (error) if an error occurred
+///       Use to display error message in CClientDlg
+/// @param strServerAddress - the server address to connect to
+/// @param strServerName - the String argument to be passed to Connecting()
+void CClient::Connect ( QString strServerAddress, QString strServerName )
+{
+    try
+    {
+        if ( !IsRunning() )
+        {
+            // Set server address and connect if valid address was supplied
+            if ( SetServerAddr ( strServerAddress ) )
+            {
+                SetConnectedServerName ( strServerName );
+                Start();
+            }
+            else
+            {
+                throw CGenErr ( tr ( "Received invalid server address. Please check for typos in the provided server address." ) );
+            }
+        }
+    }
+    catch ( const CGenErr& generr )
+    {
+        Stop();
+        emit ConnectingFailed ( generr.GetErrorText() );
+    }
 }
 
 void CClient::Init()
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -937,6 +937,7 @@ void CClient::OnHandledSignal ( int sigNum )
         // is notified we are leaving (Disconnect() is a no-op if not connected)
         Disconnect();
 
+        // this should trigger OnAboutToQuit
         QCoreApplication::instance()->exit();
         break;
 
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -1097,7 +1097,15 @@ void CClient::Stop()
 
     // Fall back to opus in case raw was used
     bRawAudioIsSupported = false;
-    Init();
+    try
+    {
+        Init();
+    }
+    catch ( const CGenErr& )
+    {
+        // a dead audio backend (e.g. JACK was shut down) must not prevent the
+        // disconnect message below from reaching the server
+    }
 
     // wait for approx. 100 ms to make sure no audio packet is still in the
     // network queue causing the channel to be reconnected right after having
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -933,10 +933,10 @@ void CClient::OnHandledSignal ( int sigNum )
     {
     case SIGINT:
     case SIGTERM:
-        // if connected, Stop client (needed for headless mode)
-        Stop();
+        // tear down any pending or established connection first, so the server
+        // is notified we are leaving (Disconnect() is a no-op if not connected)
+        Disconnect();
 
-        // this should trigger OnAboutToQuit
         QCoreApplication::instance()->exit();
         break;
 
@@ -1135,11 +1135,15 @@ void CClient::Stop()
 }
 
 /// @method
-/// @brief Stops the client if the client is running
+/// @brief Disconnects from the server if a connection is requested or established.
+///        Idempotent: a no-op when already disconnected.
 /// @emit Disconnected
 void CClient::Disconnect()
 {
-    if ( IsRunning() )
+    // Key off the connection state, not IsRunning() (which tracks the audio
+    // device): the two diverge while connecting and in headless mode, and on
+    // SIGTERM we must still send the disconnect message to the server.
+    if ( GetConnectionState() != CS_DISCONNECTED )
     {
         Stop();
     }
@@ -1180,6 +1184,10 @@ void CClient::Connect ( QString strServerAddress, QString strServerName )
     }
 }
 
+/// @method
+/// @brief Updates the connection state and, if it changed, notifies observers.
+/// @emit ConnectionStateChanged (state) when the state actually changes
+/// @param eNewConnectionState - the state to transition to
 void CClient::SetConnectionState ( const EConnectionState eNewConnectionState )
 {
     if ( eConnectionState != eNewConnectionState )
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -59,6 +59,7 @@ CClient::CClient ( const quint16  iPortNumber,
     strClientName ( strNClientName ),
     pSignalHandler ( CSignalHandler::getSingletonP() ),
     pSettings ( nullptr ),
+    eConnectionState ( CS_DISCONNECTED ),
     Channel ( false ), /* we need a client channel -> "false" */
     CurOpusEncoder ( nullptr ),
     CurOpusDecoder ( nullptr ),
@@ -1027,6 +1028,9 @@ void CClient::OnClientIDReceived ( int iServerChanID )
         SetRemoteChanGain ( iChanID, 0, false );
     }
 
+    // the server has assigned us a channel ID, so the connection is established
+    SetConnectionState ( CS_CONNECTED );
+
     emit ClientIDReceived ( iChanID );
 }
 
@@ -1070,7 +1074,12 @@ void CClient::Start()
     SetThreadExecutionState ( ES_CONTINUOUS | ES_SYSTEM_REQUIRED | ES_DISPLAY_REQUIRED );
 #endif
 
-    emit Connected ( GetConnectedServerName() );
+    // the connection is requested now but not yet established: the transition
+    // to CS_CONNECTED happens when the server assigns our channel ID
+    // (see OnClientIDReceived)
+    SetConnectionState ( CS_CONNECTING );
+
+    emit Connecting ( GetConnectedServerName() );
 }
 
 /// @method
@@ -1119,6 +1128,8 @@ void CClient::Stop()
     SetThreadExecutionState ( ES_CONTINUOUS );
 #endif
 
+    SetConnectionState ( CS_DISCONNECTED );
+
     // emit Disconnected() to inform UI of disconnection
     emit Disconnected();
 }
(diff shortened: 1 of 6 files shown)


ann0see:
Thanks. This needs thorough testing as this refactor is definitely non trivial.

pljones:
I'm just wondering if we should cut a 4.0.0 off where we are on main plus the TCP changes when they land and have these refactorings land onto 4.1.0 (4.0.0dev)?

We're currently holding up a feature (raw audio) that benefits all users for work that has infrastructure value.

It's definitely good to have them, though.

ann0see:
> We're currently holding up a feature (raw audio) that benefits all users for work that has infrastructure value.

It's stalled for a good reason though. Now we know it works (but probably would benefit from PLC) but at the start we didn't.

ann0see:
GUI tested and I can confirm it works as expected. From the code review it also looks fine. I'll still need to do an in depth analysis though.

src/client.cpp
```diff
@@ -1045,8 +1049,19 @@ void CClient::Start()
...
+    emit Connecting ( GetConnectedServerName() );
 }
 
+/// @method
+/// @brief Stops client and disconnects from server
+/// @emit Disconnected
+///       Use to set CClientDlg to show not being connected
 void CClient::Stop()
 {
     // stop audio interface
```
ann0see, line 1092: The IsRunning() guard is not needed? (don't remember the initial argument of my PR.)

src/client.cpp
```diff
@@ -1088,6 +1103,66 @@ void CClient::Stop()
...
+        }
+    }
+    catch ( const CGenErr& generr )
+    {
+        Stop();
+        emit ConnectingFailed ( generr.GetErrorText() );
+    }
+}
+
+void CClient::SetConnectionState ( const EConnectionState eNewConnectionState )
```
ann0see, line 1201: Add documentation like for connect/disconnect.

src/client.h
```diff
@@ -353,6 +362,14 @@ class CClient : public QObject
     void FreeClientChannel ( const int iServerChannelID );
     int  FindClientChannel ( const int iServerChannelID, const bool bCreateIfNew ); // returns a client channel ID or INVALID_INDEX
     bool ReorderLevelList ( CVector<uint16_t>& vecLevelList );                      // modifies vecLevelList, passed by reference
+    // information for the connected server
```
ann0see: Which kind of information? Comment is unclear.

src/client.h
```diff
@@ -353,6 +362,14 @@ class CClient : public QObject
...
+    QString strConnectedServerName;
+
+    // single source of truth for the connection state, only to be modified
+    // via SetConnectionState() so that every change emits ConnectionStateChanged()
```
ann0see: Does not really add much and smells to much AI ish.

src/clientdlg.cpp
```diff
@@ -1236,65 +1219,38 @@ void CClientDlg::OnCLPingTimeWithNumClientsReceived ( CHostAddress InetAddr, int
...
-        if ( pSettings->bEnableFeedbackDetection )
-        {
-            TimerDetectFeedback.start ( DETECT_FEEDBACK_TIME_MS ); // single shot timer
-            bDetectFeedback = true;
-        }
+    // audio feedback detection
+    if ( pSettings->bEnableFeedbackDetection )
+    {
+        TimerDetectFeedback.start ( DETECT_FEEDBACK_TIME_MS ); // single shot timer
+        bDetectFeedback = true;
```
ann0see, line 1247: Not part of this PR, but yes, this should also not be in the UI.

src/util.h
```diff
@@ -590,6 +590,17 @@ enum ERecorderState
     RS_RECORDING       = 3
 };
 
+// Client connection state -----------------------------------------------------
```
ann0see: Do not put this into util.h I think it's a client specific thing -> client.h

ann0see (changes requested):

src/client.cpp
ann0see: Please check this - especially for the sigterm case: https://github.com/jamulussoftware/jamulus/pull/3805/changes#diff-7bc46552432fa03e009f9f455e4729312853d92775935ef20849797f24ac0c60R912-R913

ann0see (commented):
Check if this breaks feedback detection somehow. (Manually). 


Free model from open code (Big Pickle) had the following comments (also seem my comments on this)

This one I was already suspicious about:
```
Bug Assessment
Likely Bug: Stop() emits Disconnected unconditionally
CClient::Stop() now always emits Disconnected() even if the client wasn't connected:
void CClient::Stop()
{
    // ... cleanup ...
    SetConnectionState ( CS_DISCONNECTED ); // no-op if already disconnected
    emit Disconnected();                    // <-- ALWAYS emitted
}
This means every Stop() call triggers the full UI teardown chain in OnDisconnect() (reset button text, hide mixer board, clear faders, reset deco). If Stop() is ever called when the client wasn't running, the UI will react to a spurious disconnect. The old code guarded this with if (IsRunning()).

```

```
Potential Issue: bDetectFeedback not reset on disconnect
When OnConnecting() starts the feedback detection timer and sets bDetectFeedback = true, neither OnDisconnect() nor Stop() resets it. The single-shot timer will expire, but the flag remains stale. Not a crash bug, but could cause unexpected behavior if bDetectFeedback is checked elsewhere before the next connection.

```

Fair point. Maybe be problematic - but not sure...

```
Minor: CS_CONNECTING state during Connect() failure path
In Connect(), the catch block calls Stop(), which transitions CS_CONNECTING → CS_DISCONNECTED and emits Disconnected. Then the catch block also emits ConnectingFailed. So the UI gets both signals in sequence — this works, but ConnectingFailed arrives after the UI has already been torn down by Disconnected, which could be confusing for error display ordering.
```
Probably really not too bad.

src/client.cpp
mcfnord: **🤖 AI:** Addressed together with the guard above in efee722b — `OnHandledSignal` now calls `Disconnect()`. SIGTERM while `CS_CONNECTING` sends `CLM_DISCONNECTION`, confirmed with a pcap.

src/client.cpp
mcfnord: Done in efee722b — added a `/// @method` doc-comment.

src/client.h
mcfnord: Done in efee722b — reworded to "human-readable name of the current server, shown in the UI".

src/client.h
mcfnord: Done in efee722b — trimmed to "current connection state; set only via SetConnectionState()".

src/util.h
mcfnord: Done in efee722b — moved `EConnectionState` to `client.h`.

src/clientdlg.cpp
mcfnord: Agreed, out of scope for this PR — left as-is for a separate change.

src/client.cpp
mcfnord: **🤖 AI:** `IsRunning()` is `Sound.IsRunning()` (the audio device), so the guard keyed off something other than the connection. Keying it off the explicit state keeps `Disconnect()` idempotent: repeated UI calls stay no-ops and `OnHandledSignal` no longer needs a raw `Stop()` to bypass the guard.

The two predicates diverge when the audio backend dies while connected: the device stops but the connection stays up. SIGTERM in that state, measured on Linux (dummy JACK killed mid-session, disconnect counted from a loopback pcap):

| guard | CLM_DISCONNECTION sent | server holds the dead slot |
|---|---|---|
| main, `IsRunning()` | no (5/5 reps) | 27 s (timeout) |
| this branch, connection state | yes (3/3 reps) | 0.1 s |

Reaching that result exposed one more thing this branch now fixes: `Stop()` calls `Init()`, which throws from `CSound::Init` (`sound/jack/sound.cpp`, the `bJackWasShutDown` check) before the disconnect message is sent, and the exception escaped the Qt slot. The latest commit wraps that `Init()` in try/catch so a dead audio backend cannot block the disconnect.

The stale `// needed for headless mode` and `// this should trigger OnAboutToQuit` comments are dropped, since there is no client `OnAboutToQuit`.

ann0see:
I'll look at this later. I'd like to have this in 4.0 not 4.1 as it's crucial for clean connect/disconnect RPC functionality.

src/client.cpp
```diff
@@ -909,10 +909,10 @@ void CClient::OnHandledSignal ( int sigNum )
     {
     case SIGINT:
     case SIGTERM:
-        // if connected, Stop client (needed for headless mode)
-        Stop();
+        // tear down any pending or established connection first, so the server
+        // is notified we are leaving (Disconnect() is a no-op if not connected)
+        Disconnect();
 
-        // this should trigger OnAboutToQuit
```
pljones, line 938: I believe the comment should remain.  The `QCoreApplication::instance()->exit();` was _added_ so that `OnAboutToQuit` got called.

mcfnord: Restored in a7c5c7b4 — you're right, it should have stayed through the refactor.

pljones: Ta.

src/client.h
```diff
@@ -159,6 +170,15 @@ class CClient : public QObject
 
     void Start();
     void Stop();
+    void Disconnect();
+    void Connect ( QString strServerAddress, QString strServerName );
+
+    // The ConnectedServerName is emitted by Connecting() to update the UI with a human readable server name
+    void    SetConnectedServerName ( const QString strServerName ) { strConnectedServerName = strServerName; };
```
softins: Do one copy instead of two:

```suggestion
    void    SetConnectedServerName ( const QString& strServerName ) { strConnectedServerName = strServerName; };
```

src/client.cpp
```diff
@@ -1088,6 +1104,74 @@ void CClient::Stop()
...
+/// @method
+/// @brief Connects to strServerAddress. If a connection is currently requested
+///        or established, that connection is terminated first.
+/// @emit Connecting (strServerName) if SetServerAddr was valid. emit happens through Start().
+///       Use to set CClientDlg to show being connected
+/// @emit ConnectingFailed (error) if an error occurred
+///       Use to display error message in CClientDlg
+/// @param strServerAddress - the server address to connect to
+/// @param strServerName - the human readable server name passed to Connecting()
+void CClient::Connect ( QString strServerAddress, QString strServerName )
```
softins: These can be const refs

```suggestion
void CClient::Connect ( const QString& strServerAddress, const QString& strServerName )
```

src/client.h
```diff
@@ -159,6 +170,15 @@ class CClient : public QObject
 
     void Start();
     void Stop();
+    void Disconnect();
+    void Connect ( QString strServerAddress, QString strServerName );
```
softins:
```suggestion
    void Connect ( const QString& strServerAddress, const QString& strServerName );
```

src/client.h
```diff
@@ -507,7 +535,11 @@ protected slots:
 
     void CLChannelLevelListReceived ( CHostAddress InetAddr, CVector<uint16_t> vecLevelList );
 
+    void ConnectionStateChanged ( EConnectionState eConnectionState );
+    void Connecting ( QString strServerName );
+    void ConnectingFailed ( QString errorMessage );
```
softins:
```suggestion
    void Connecting ( const QString& strServerName );
    void ConnectingFailed ( const QString& errorMessage );
```

softins (changes requested):
It looks good. There are several places where a `QString` value parameter could better be a const ref `const QString&`.

I wondered for a moment whether the signal/slot boundary needed object parameters such as `QString` to be passed by value. But in fact Qt can handle const refs just fine, copying only when necessary to cross a thread boundary, and it reduces the number of copies, [as described here](https://www.smallstepsystems.com/copied-or-not-copied-arguments-signals-slots/)

softins:
BTW, I've tested my suggested changes.

ann0see:
@mcfnord Please have a look at this Pr again.

mcfnord:
**🤖 AI**

All four suggestions are in, as one commit. The receiving slots already declared `const QString&`, so signal and slot signatures now match exactly instead of relying on conversion.

Checked while making the change: `Connect()` calls `Disconnect()` before it reads `strServerName`, and nothing on that path writes `strConnectedServerName`, so the reference stays valid. All three call sites pass function-local strings.

GUI and headless both build clean. A headless client still connects to a local server end to end: `jamulusclient/getClientInfo` returns `{"connected": true}`.

---

softins (approved):
Looks good to me now. Thanks!

ann0see:
Please don't merge. Updating the branch = rebase.

src/client.cpp
```diff
@@ -1073,7 +1089,15 @@ void CClient::Stop()
...
     // Fall back to opus in case raw was used
     bRawAudioIsSupported = false;
-    Init();
+    try
+    {
+        Init();
+    }
+    catch ( const CGenErr& )
+    {
+        // a dead audio backend (e.g. JACK was shut down) must not prevent the
```
ann0see, line 1106: Hmm. Maybe the warning should be logged.

ann0see (commented):
See above comment.

ann0see (commented):
Otherwise it seems good.

src/client.cpp
```diff
@@ -1061,8 +1066,19 @@ void CClient::Start()
...
+    SetConnectionState ( CS_CONNECTING );
+
+    emit Connecting ( GetConnectedServerName() );
 }
 
+/// @method
+/// @brief Stops client and disconnects from server
+/// @emit Disconnected
+///       Use to set CClientDlg to show not being connected
 void CClient::Stop()
```
ann0see, line 1090: We might even be able to make Stop() private so that everything is forced to go through Disconnect()

src/client.cpp
```diff
@@ -1104,6 +1128,74 @@ void CClient::Stop()
...
+        Stop();
+        emit ConnectingFailed ( generr.GetErrorText() );
+    }
+}
+
+/// @method
+/// @brief Updates the connection state and, if it changed, notifies observers.
+/// @emit ConnectionStateChanged (state) when the state actually changes
+/// @param eNewConnectionState - the state to transition to
+void CClient::SetConnectionState ( const EConnectionState eNewConnectionState )
```
ann0see, line 1201: Probably also worth to make this private

src/client.cpp
softins: The other restricted methods in `CClient` are `protected`, not `private`. So it would make sense to change `Stop()` from `public` to `protected`, rather than to `private`, if we want to avoid it being called from outside the class. Although since `CClient` is not subclassed in Jamulus, the effect is the same.

I have just tested changing `Stop()` to `protected` and it still compiles, so there are currently no stray calls to `CClient::Stop()` from outside the class.

src/client.cpp
softins: This is already `protected`, not `public`.

src/client.cpp
softins: Just tested also making `CClient::Start()` be `protected`, and it still compiles. So we can move both `Start()` and `Stop()` declarations to the `protected` section.

mcfnord: **🤖 AI:** Done in 92d32f0. Both declarations sit in the `protected` section next to `Init()` now. A full client build with JACK enabled is clean, so nothing outside the class still calls them, and a translation unit calling `p->Start()` on a `CClient*` is rejected: `'void CClient::Start()' is protected within this context`.

src/client.cpp
mcfnord: **🤖 AI:** Logged in 92d32f0. The catch takes `generr` again and the error text goes to `qWarning()` before the disconnect continues. The wording and the log level are easy to change if you want something different.

softins (approved):

softins:
@mcfnord this rebases cleanly on to the latest `main`. Shall I push my rebased branch, or leave it to you?

ann0see:
Cleaner if he does it.

ann0see (approved):

ann0see:
Thanks. This was an important one... I'd say that once we have the JSON-RPC ones in we should tag a new beta. This needs some real world testing.

PR #3810, opened 2026-07-20
Add bounds check iChanID before array access

ann0see:

Guards iChanId to definitely be in bounds. Checking the flow revealed that there is indeed no other bounds check and we may have INVALID_INDEX (=-1) which would yield to an invalid access (found by AI review)

CHANGELOG: Resolve out of bounds access issue

**Context: Fixes an issue?**

No. Only discussed internally.

Ready for *quick* !! review. This is a potentially severe bug (as discussed internally)

Review and testing

## Checklist

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1349,9 +1349,12 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
 
         for ( size_t iFader = 0; iFader < iNumConnectedClients; iFader++ )
         {
-            // ideally "iChanID" in CChannelInfo would be size_t if it can never be INVALID_INDEX
-            // as assumed here
-            iFaderNumber[vecChanInfo[iFader].iChanID] = static_cast<int> ( iFader );
+            const int iChanID = vecChanInfo[iFader].iChanID;
+
+            if ( MathUtils::InRange<int> ( iChanID, 0, MAX_NUM_CHANNELS ) )
+            {
+                iFaderNumber[iChanID] = static_cast<int> ( iFader );
+            }
         }
 
         // Hide all unused faders and initialize used ones


src/audiomixerboard.cpp
```diff
@@ -1349,9 +1349,12 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
 
         for ( size_t iFader = 0; iFader < iNumConnectedClients; iFader++ )
         {
-            // ideally "iChanID" in CChannelInfo would be size_t if it can never be INVALID_INDEX
```
ann0see, line 1352: the comment was wrong.

src/audiomixerboard.cpp
```diff
@@ -1349,9 +1349,12 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
...
-            // as assumed here
-            iFaderNumber[vecChanInfo[iFader].iChanID] = static_cast<int> ( iFader );
+            const int iChanID = vecChanInfo[iFader].iChanID;
+
+            if ( ( iChanID >= 0 ) && ( static_cast<size_t> ( iChanID ) < MAX_NUM_CHANNELS ) )
```
dingodoppelt: You could use `MathUtils::InRange<T> ( T value, T lower, T upper )` here

softins (approved):
Looks good to me.

Although as a separate thing, I would prefer to make `MathUtils::InRange()` inclusive only of the lower bound, and exclusive of the upper bound, so that we could have `MAX_NUM_CHANNELS` as the upper bound instead of `MAX_NUM_CHANNELS - 1`, and similarly for the only other use of `InRange()`, which is in server.cpp

dingodoppelt:
> Looks good to me.
> 
> Although as a separate thing, I would prefer to make `MathUtils::InRange()` inclusive only of the lower bound, and exclusive of the upper bound, so that we could have `MAX_NUM_CHANNELS` as the upper bound instead of `MAX_NUM_CHANNELS - 1`, and similarly for the only other use of `InRange()`, which is in server.cpp

Can we do that as part of this PR?

ann0see:
No.

dingodoppelt:
> No.

Created #3812

src/audiomixerboard.cpp
pljones: It includes the word `if`.  The reason it's not `size_t` is because `size_t` is `unsigned`.  As far as I'm aware, the comment is correct.

ann0see: Ok. We still need to have an int then until after the check.

mcfnord:
Worth merging: `iChanID` comes off the wire as an unvalidated **unsigned byte** (`EvaluateConClientListMes`, 0–255), and `iFaderNumber` is a **stack-local** `int[150]` — so a value in 150–255 is a server-controlled stack OOB write on every client. (Note the real risk is the *upper* bound; `-1` can't arrive over the wire.) L1354 is the only wire-indexed write, so this closes it. The assumption has sat unenforced in-source since 2022. LGTM.

pljones:
Once #3812 lands on main, this must go next.  Then those two commits get ported to release/3_12 on on PR, please.

ann0see:
Rebased and updated

dingodoppelt (approved):

pljones:
~~Backported to release/3_12.~~

pljones:
Backport broke the compile as it depended on a change outside this merge.

ann0see:
Then the respective fix should also be backported. (Probably in range (all PRs including the bounds shift)

pljones:
It needs more than this one commit.  Please raise the appropriate PR.

PR #3811, opened 2026-07-20
add bounds check for country code table

ann0see:

Adds bounds check for qt6 country code conversion (also found like #3810 )

CHANGELOG: Add bounds checking for country code lookup

**Context: Fixes an issue?**

/

No

Ready for review (didn't analyze this one deeply). The style follows isCountryCodeSupported. Alternatively, we could probably just call IsCountryCode supported as guard? 

Review, analysis.

## Checklist

-  [ ] I tested my code and it does what I want (as in - it compiles with qt6 and seemed to show countries correctly - but not 100% sure)

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -1369,6 +1369,10 @@ QLocale::Country CLocale::WireFormatCountryCodeToQtCountry ( unsigned short iCou
 #if QT_VERSION >= QT_VERSION_CHECK( 6, 0, 0 )
     // The Jamulus protocol wire format gives us Qt5 country IDs.
     // Qt6 changed those IDs, so we have to convert back:
+    if ( iCountryCode >= qt6CountryToWireFormatLen )
+    {
+        return QLocale::AnyCountry;
+    }
     return (QLocale::Country) wireFormatToQt6Table[iCountryCode];
 #else
     return (QLocale::Country) iCountryCode;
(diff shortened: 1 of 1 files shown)


src/util.cpp
```diff
@@ -1369,6 +1369,10 @@ QLocale::Country CLocale::WireFormatCountryCodeToQtCountry ( unsigned short iCou
 #if QT_VERSION >= QT_VERSION_CHECK( 6, 0, 0 )
     // The Jamulus protocol wire format gives us Qt5 country IDs.
     // Qt6 changed those IDs, so we have to convert back:
+    if ( iCountryCode >= qt6CountryToWireFormatLen )
+    {
+        return QLocale::AnyCountry;
```
ann0see, line 1374: What should we return otherwise?

src/util.cpp
```diff
@@ -1380,7 +1384,13 @@ unsigned short CLocale::QtCountryToWireFormatCountryCode ( const QLocale::Countr
 #if QT_VERSION >= QT_VERSION_CHECK( 6, 0, 0 )
     // The Jamulus protocol wire format expects Qt5 country IDs.
     // Qt6 changed those IDs, so we have to convert back:
-    return qt6CountryToWireFormat[(unsigned short) eCountry];
+    const unsigned short iIdx = static_cast<unsigned short> ( eCountry );
+
+    if ( iIdx >= qt6CountryToWireFormatLen )
+    {
+        return 0;
```
ann0see, line 1391: Not sure what to return...

src/util.cpp
softins: There doesn't seem to exist any better alternative than this.

src/util.cpp
softins: I think this is correct, because it is the Qt5 value for `QLocale::AnyCountry`, which is sent on the wire if a country is explicitly not defined.

If a server does not specify `--serverinfo`, or leaves the country code blank, such as `--serverinfo 'My Server;London;'`, then it gets replaced by the country of the system locale. If the server operator doesn't want a country to show (and there are a small number like that), they need explicitly to give a code of `0`: `--serverinfo 'My Server;;0'`

softins (approved):

pljones:
Again with this, merge to main initially, then PR onto release/3_12.

ann0see:
Are you sure that you do **not** want a git cherry pick with one PR including all or most of the bugfixes?

pljones:
> Are you sure that you do **not** want a git cherry pick with one PR including all or most of the bugfixes?

I'd rather not a single PR.  Group related commits cherry-picked over into single PRs, yes but I'd still like a structured history for things like `git log --graph r3_12_2..r3_12_3`.

dingodoppelt (approved):

pljones:
Backported to release/3_12

PR #3815, opened 2026-07-21
Client: add --connectdirectory for directory-assisted hole punch on --connect

mcfnord:

Adds an optional `--connectdirectory <host:port>` client option and gives `CClient::Connect` an optional directory argument. When a directory is supplied, the client first asks that directory for its server list — which makes the directory tell its registered servers to send an "empty message" toward the client's socket (the same directory-assisted UDP hole punch the GUI directory list relies on) — waits `HOLE_PUNCH_CONNECT_DELAY_MS`, then connects.

Today `-c` / `--connect` connects straight to the address and never contacts a directory, so a server behind a cloud firewall/NAT that the GUI reaches fine is silently unreachable via `-c`. The server address is connected to verbatim and need not appear in the directory's server list, so this also lets you reach a server that isn't in any *built-in* directory by naming the (possibly custom/private) directory it is registered with.

CHANGELOG: Client: Added `--connectdirectory` to hole-punch through a directory when connecting on startup with `--connect`

**Context: Fixes an issue?**

Relates to jamulussoftware/jamuluswebsite#1122, which documents this `-c` reachability limitation. This is the code side of that gap.

Stacked on #3805 — it uses the `CClient::Connect` refactor introduced there, so the diff currently includes #3805's commit; the single top commit is the change in this PR. Please review/merge #3805 first.

Yes — a short website entry for `--connectdirectory` next to the `-c` text (a companion to the `-c` note added in jamuluswebsite#1156). A draft on the website repo will follow.

Working implementation. Verified with a UDP capture: with `--connectdirectory` the client sends `CLM_REQ_SERVER_LIST` to the directory and connects ~400 ms later; without it there is no directory traffic and behaviour is unchanged. Clean SIGTERM shutdown, including when interrupted during the hole-punch delay.

Depends on #3805 merging first. Otherwise ready for review.

## Checklist

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -145,7 +145,7 @@ CClient::CClient ( const quint16  iPortNumber,
     // The first ConClientListMesReceived handler performs the necessary cleanup and has to run first:
     QObject::connect ( &Channel, &CChannel::ConClientListMesReceived, this, &CClient::OnConClientListMesReceived );
 
-    QObject::connect ( &Channel, &CChannel::Disconnected, this, &CClient::Disconnected );
+    QObject::connect ( &Channel, &CChannel::Disconnected, this, &CClient::Stop );
 
     QObject::connect ( &Channel, &CChannel::NewConnection, this, &CClient::OnNewConnection );
 
@@ -618,6 +618,9 @@ bool CClient::SetServerAddr ( QString strNAddr )
         // apply address to the channel
         Channel.SetAddress ( HostAddress );
 
+        // By default, set server name to HostAddress. If using the Connect() method, this may be overwritten
+        SetConnectedServerName ( HostAddress.toString() );
+
         return true;
     }
     else
@@ -905,11 +908,8 @@ void CClient::OnHandledSignal ( int sigNum )
     {
     case SIGINT:
     case SIGTERM:
-        // if connected, terminate connection (needed for headless mode)
-        if ( IsRunning() )
-        {
-            Stop();
-        }
+        // if connected, Stop client (needed for headless mode)
+        Stop();
 
         // this should trigger OnAboutToQuit
         QCoreApplication::instance()->exit();
@@ -1045,8 +1045,14 @@ void CClient::Start()
     // Disable hibernation or display dimming if the app is running on Windows
     SetThreadExecutionState ( ES_CONTINUOUS | ES_SYSTEM_REQUIRED | ES_DISPLAY_REQUIRED );
 #endif
+
+    emit Connected ( GetConnectedServerName() );
 }
 
+/// @method
+/// @brief Stops client and disconnects from server
+/// @emit Disconnected
+///       Use to set CClientDlg to show not being connected
 void CClient::Stop()
 {
     // stop audio interface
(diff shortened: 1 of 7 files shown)


ann0see:
Best case `-c` would fall back automatically to this after some timeout imo. This doesn't exclude an explicit option but would be more user friendly

mcfnord:
> Best case `-c` would fall back automatically to this after some timeout imo. This doesn't exclude an explicit option but would be more user friendly

The client would need to know which directory to use, and that will have to be specified by the CLI parameters. It would be too sloppy and nuts to guess all 7, and even if we do, it might be some other directory entirely.

dingodoppelt:
> Best case `-c` would fall back automatically to this after some timeout imo. This doesn't exclude an explicit option but would be more user friendly

We should definitely *not* make this a default. The current `--connect` behavior can used to mitigate the correlation attack as stated [in this coment](https://github.com/orgs/jamulussoftware/discussions/3545#discussioncomment-16521524).

src/client.cpp
```diff
@@ -1128,13 +1128,54 @@ void CClient::Disconnect()
...
+    {
+        emit ConnectingFailed ( tr ( "Received invalid directory address. Please check for typos in the provided directory address." ) );
+        return;
+    }
+
+    // Ask the directory for its server list. As a side effect the directory
+    // tells each registered server to send us an "empty message", which opens
+    // the server's firewall for our socket (UDP hole punching).
+    CreateCLReqServerListMes ( haDirectoryAddress );
+
```
dingodoppelt, line 1168: This breaks `--connect`'s current behaviour by leaking the user's IP.

dingodoppelt:
We don't actually need this because server admins can always create the server as a private directory which takes care of hole-punching without leaking the IP.

softins:
> > Best case `-c` would fall back automatically to this after some timeout imo. This doesn't exclude an explicit option but would be more user friendly
> 
> We should definitely _not_ make this a default. The current `--connect` behavior can used to mitigate the correlation attack as stated [in this coment](https://github.com/orgs/jamulussoftware/discussions/3545#discussioncomment-16521524).

It's not the lack of hole-punching in `--connect` that mitigates the correlation behaviour (I hesitate to call it an "attack"), but rather the lack of regular pings (to an unrelated server such as `Duet`) up until the time of connecting to the desired server.

So the option in this PR would not actually open the client to that correlation behaviour.

But having said that, I don't feel there is much usefulness in adding the feature in this PR.

softins:
> We don't actually need this because server admins can always create the server as a private directory which takes care of hole-punching without leaking the IP.

Yes, if creating a private server that is not intended to register with a directory, it is actually a very useful idea to make it its own directory with `-e localhost`, and for the private clients to add it as a custom directory.

dingodoppelt:
> It's not the lack of hole-punching in `--connect` that mitigates the correlation behaviour (I hesitate to call it an "attack"), but rather the lack of regular pings (to an unrelated server such as `Duet`) up until the time of connecting to the desired server.

I still think we should let the user decide if he wants his IP leaked by `CLM_EMPTY_MESSAGE`, which is sent to every Server in the list. The pings can be easily blocked locally, `CLM_EMPTY_MESSAGE` sent by the Directory can't. 
[Comment](https://github.com/orgs/jamulussoftware/discussions/3545#discussioncomment-17945154)

dingodoppelt (changes requested):
I'd rather not pull this in at all because of the security implications.

pljones:
> > It's not the lack of hole-punching in `--connect` that mitigates the correlation behaviour (I hesitate to call it an "attack"), but rather the lack of regular pings (to an unrelated server such as `Duet`) up until the time of connecting to the desired server.
> 
> I still think we should let the user decide if he wants his IP leaked by `CLM_EMPTY_MESSAGE`, which is sent to every Server in the list. The pings can be easily blocked locally, `CLM_EMPTY_MESSAGE` sent by the Directory can't. [Comment](https://github.com/orgs/jamulussoftware/discussions/3545#discussioncomment-17945154)

The user can already use `-c` and avoid the use of any Directory.

Having the user able to choose _which_ Directory they _trust_ surely improves their level of control, rather than weakening it?

src/client.cpp
mcfnord: `--connect` is broken now, because it lacks parity with the GUI. Fixing `--connect` in this way would have the side-effect of leaking the user's IP:PORT to every server on the directory.

Because we know which IP:PORT the user wants, we could teach the directory to Only hole-punch on that server when `--connect` is used.

That would bring `--connect` into parity with the GUI, while not leaking IP:PORT to any other server.

Would that require a new command to the Directory? "just one hole to punch" command?

dingodoppelt:
> The user can already use `-c` and avoid the use of any Directory.
>

Yes, but no hole-punching which is what this PR proposes. 

> Having the user able to choose _which_ Directory they _trust_ surely improves their level of control, rather than weakening it?

It is about the server users want to connect to, not the directory, which is needed just for hole punching. There is no need to send `CLM_EMTPY_MESSAGE` to every server in the list. We rather need a directory assisted connect mechanism that doesn't leak the user's IP.  
We could reuse that mechanism for _"trusted"_ servers. Only to those a Directory would send `CLM_EMTPY_MESSAGE` so we could see the ping in the connect dialogue.

PR #3820, opened 2026-07-21
Fix jamulus/pollServerList reporting success on invalid directory address

### LINKED ISSUE #3818: Missing return statement in clientrpc error handling
https://github.com/jamulussoftware/jamulus/blob/0db554413e6311c00c81900e8af114e1527ab6a7/src/clientrpc.cpp#L202

seems to be missing a return statement. So we set result == ok which is not correct in case of errors.
Found by local AI review

mcfnord:
## Fixes #3818

The `jamulus/pollServerList` RPC handler set `response["error"]` when the directory address failed to parse, but then **unconditionally** set `response["result"] = "ok"` after the `if`/`else`. As a result, an invalid socket address was reported as success, with both `"error"` and `"result"` present in the response.

```cpp
if ( NetworkUtil::ParseNetworkAddress ( ..., false ) )
{
    pClient->CreateCLReqServerListMes ( haDirectoryAddress );
    response["result"] = "ok";
}
else
{
    response["error"] = CRpcServer::CreateJsonRpcError ( ... );
}

response["result"] = "ok";   // <-- always ran, clobbering the error case
```

### Fix
Remove the stray unconditional assignment. The success branch already sets `result = "ok"`, so the handler now reports `error` on failure and `result` on success — matching the error-path idiom used by every other handler in this file.

### Testing
- `clang-format-14` clean
- Headless build + server smoke test (clean startup, clean SIGTERM)

### DIFF
--- a/src/clientrpc.cpp
+++ b/src/clientrpc.cpp
@@ -201,8 +201,6 @@ CClientRpc::CClientRpc ( CClient* pClient, CClientSettings* pSettings, CRpcServe
             response["error"] =
                 CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: directory is not a valid socket address" );
         }
-
-        response["result"] = "ok";
     } );
 
     /// @rpc_method jamulus/getMode
--- a/src/clientrpc.cpp
+++ b/src/clientrpc.cpp
@@ -190,17 +190,16 @@ CClientRpc::CClientRpc ( CClient* pClient, CClientSettings* pSettings, CRpcServe
         CHostAddress haDirectoryAddress;
 
         // Allow IPv4 only for communicating with Directories
-        if ( NetworkUtil::ParseNetworkAddress ( jsonDirectoryIp.toString(), haDirectoryAddress, false ) )
-        {
-            // send the request for the server list
-            pClient->CreateCLReqServerListMes ( haDirectoryAddress );
-            response["result"] = "ok";
-        }
-        else
+        if ( !NetworkUtil::ParseNetworkAddress ( jsonDirectoryIp.toString(), haDirectoryAddress, false ) )
         {
             response["error"] =
                 CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: directory is not a valid socket address" );
+            return;
         }
+
+        // send the request for the server list
+        pClient->CreateCLReqServerListMes ( haDirectoryAddress );
+        response["result"] = "ok";
     } );
 
     /// @rpc_method jamulus/getMode


src/clientrpc.cpp
```diff
@@ -201,8 +201,6 @@ CClientRpc::CClientRpc ( CClient* pClient, CClientSettings* pSettings, CRpcServe
             response["error"] =
                 CRpcServer::CreateJsonRpcError ( CRpcServer::iErrInvalidParams, "Invalid params: directory is not a valid socket address" );
         }
-
-        response["result"] = "ok";
```
ann0see, line 205: This is definitely not the correct fix... just use a return; no?

ann0see (changes requested):
AI is probably still working...

src/clientrpc.cpp
ann0see: Ah ok. Understand the reasoning - however it should still follow the same style as in other methods above...

ann0see:
Testing should obviously test the JSON-RPC api...

mcfnord:
> AI is probably still working...

The Fable special ended, so today is Opus on X-High Effort. I saw a few examples today of Opus making small errors, but those probably indicate larger errors are more possible, including subtle errors. So in [the most complicated instance](https://github.com/jamulussoftware/jamulus/issues/3662), I asked Fable to review results.

src/clientrpc.cpp
mcfnord: Fixed in d90fbb6 - reworked to the guard-clause/early-return style used by the other handlers in this file (error + return, then the unconditional success path).

mcfnord:
> [!NOTE]

@ann0see The requested change is in `d90fbb6`: the handler now uses the guard-clause / early-return style of the other methods in this file — invalid address sets `response["error"]` and returns, then the success path sets `response["result"] = "ok"` unconditionally. Ready for re-review whenever you have a moment.

ann0see (approved):
Could not test it but review looks ok. Someone should verify the new code by testing.

mcfnord:
MY LLM WROTE:

@ann0see Did the JSON-RPC verification: built this branch (`d90fbb6a`) headless (`qmake CONFIG+=headless && make`), ran a real headless client (`jackd -d dummy` for audio, `Jamulus -n --jsonrpcport 9999 --jsonrpcsecretfile ...`), authenticated via `jamulus/apiAuth`, and called `jamulusclient/pollServerList` directly over the TCP JSON-RPC socket — not just a startup smoke test.

**Invalid directory address** (`"not-a-valid-address!!"`):
```json
{"error": {"code": -32602, "message": "Invalid params: directory is not a valid socket address"}, "id": 2, "jsonrpc": "2.0"}
```
Only `error` is present — no `result: "ok"` alongside it, which was the bug.

**Valid directory address** (`"anygenre1.jamulus.io:22124"`):
```json
{"id": 3, "jsonrpc": "2.0", "result": "ok"}
```
Only `result: "ok"` is present — no `error`. Success path is unaffected by the fix.

Both cases behave as intended.

pljones (approved):

ann0see:
Thanks.

Issue #2428, opened 2022-02-23
Automatical Client/Server Smoke Test in CI


hoffie:
**Has this feature been discussed and generally agreed?**

@pljones and @ann0see raised the idea to run automatical client/server smoke tests in CI (#2423). I second this idea.
At least with Ubuntu/Debian this should be very doable.

**Describe the solution you'd like**

- Start a Docker container, install jamulus-headless, start it as server
- Start another Docker container, install jamulus-headless, start it as headless client
- Verify that the client connected properly (#1975 might help)
- Exit.
- Repeat for all supported Ubuntu/Debian releases

It's probably sufficient to run this on all (pre-)releases, but not on every push or PR. It would be nice to have that for all platforms, but let's start with the simple stuff. This would already confirm that packaging works and basic Jamulus functionality is there without crashing right away.

Testing for compatibility (old server + new client, old client + new server) would also be interesting.

**Describe alternatives that have been considered**

Rely on manual testing and see it break as we don't cover all platforms... :)

ann0see:
Related to https://github.com/jamulussoftware/jamulus/issues/1413. Closing https://github.com/jamulussoftware/jamulus/issues/1413 as duplicate

mcfnord:
> Do we need to run two docker containers? Would a localhost test be sufficient?

**🤖 AI:** A single host is sufficient — no second container needed. A JACK-dummy + Xvfb harness runs a real client and server on one box over loopback UDP, and has backed [several audio-correctness checks](https://gist.github.com/mcfnord/85390a43901f96391d0b41ce467ac008), including [the #3834 Input Boost fix](https://github.com/jamulussoftware/jamulus/pull/3834). Client-connected, audio flowing, end-to-end through shipped code, no Docker.

PR #3828, opened 2026-07-23
Add unit testing harness

dtinth:

Added a unit test harness (based on QtTest) with a first test suite for the protocol layer, along with a CI workflow covering:

- Linux (Qt 5.15)
- Linux (Qt 6)
- Linux with ASan/UBSan
- Linux with coverage
- macOS
- Windows

No production code is changed.

The suite (19 cases) covers:

- **Golden frames:** fixed hex literals of what `CProtocol` emits on the wire.
- **Frame acceptance/rejection:** a valid frame is accepted while invalid ones rejected with following cases tests:
    - truncated
    - bad-CRC
    - wrong-length
    - junk frames
- **Round-tripping:** Two connected `CProtocol` instances, observed via an event log of received signals.
- **Regression test** for #302 (ACKN with valid checksum but empty body caused an out-of-bounds read, found by fuzzing in 2020). The ASan job is what gives the "no OOB" part teeth.

CHANGELOG: SKIP

**Context: Fixes an issue?**

First step toward #2428 (automated testing in CI). This PR only covers unit-testing. A real client/server smoke test is [planned](https://github.com/dtinth/jamulus/issues/9) separately.

No.

Working implementation, CI-verified.

Nothing known.

Notes for reviewers:

- **Running locally:**

  ```sh
  mkdir build-test && cd build-test
  qmake ../test/test.pro
  make check
  ```

  Alternatively:

  ```sh
  python3 .github/scripts/run-unit-tests.py build
  python3 .github/scripts/run-unit-tests.py run
  ```

- The test binary's exit code is ignored because on Windows runners the binary's stdout/exit behavior is unreliable. We have a Python script inspect the generated JUnit XML instead (details in `run-unit-tests.py`).

- We built a bespoke test results reporting script in Python. We considered richer PR annotations via a report action (e.g. dorny/test-reporter), but they require enabling extra permissions that PRs from forks do not have (like `checks: write`).

- The workflow is admittedly long (~220 lines); most of it is the per-platform Qt/MSVC setup. We did not refactor them into reusable actions to keep this PR self-contained.

- I built this with help from my coding agent (Claude Code with Fable 5). You can [read the full conversation here](https://github.com/dtinth/jamulus/issues/3#issue-4948792966) (but it's quite long!). It took [a lot of steering](https://github.com/dtinth/jamulus/issues/3#issuecomment-5059555926) to arrive at the current shape.

## Checklist


.github/scripts/run-unit-tests.py
```diff
@@ -0,0 +1,186 @@
...
+    # the cmd.exe nesting and silently swallows vcvarsall.bat's effect on the
+    # "after" snapshot (before == after, so no variables get applied).
+    def snapshot ( cmd ):
+        out = subprocess.run ( cmd, shell=True, capture_output=True, text=True ).stdout
+        env = {}
+        for line in out.splitlines():
+            m = re.match ( r"^([^=]+)=(.*)$", line )
+            if m:
+                env[m.group ( 1 )] = m.group ( 2 )
+        return env
```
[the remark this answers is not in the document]
dtinth: Addressed in 15212256a1b394ca7538ac8f6f85b7b6874c456f

src/test/protocoltester.h
```diff
@@ -0,0 +1,288 @@
...
+
+    /* frame mutation helpers -----------------------------------------------
+     * Operate in place on a frame the test holds, for crafting rows/inputs
+     * that a real Create*Mes() call can't produce.
+     */
+
+    // chops iBytes off the end of vecbyFrame, e.g. to cut into the CRC or the
+    // body
+    static void TruncateBy ( CVector<uint8_t>& vecbyFrame, const int iBytes ) { vecbyFrame.resize ( vecbyFrame.Size() - iBytes ); }
+
```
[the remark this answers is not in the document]
dtinth: Addressed in 15212256a1b394ca7538ac8f6f85b7b6874c456f with clamping

src/test/protocoltester.h
```diff
@@ -0,0 +1,289 @@
...
+ *
+ * Licensed under AGPL 3.0 or any later version. See COPYING for details.
+ *
+\******************************************************************************/
+
+#pragma once
+
+#include <QtTest>
+#include "protocol.h"
+
```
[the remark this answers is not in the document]
dtinth: Addressed in 94cbe34

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,221 @@
...
+      # Same cache path/key as autobuild.yml's "MacOS (artifacts)" job, so
+      # this job shares that cache instead of keeping a separate one.
+      - name: Cache Qt (macOS)
+        if: runner.os == 'macOS'
+        uses: actions/cache@v6
+        with:
+          path: |
+            ~/qt
+            ~/Library/Cache/jamulus-dependencies
+          key: macos-${{ hashFiles('.github/workflows/autobuild.yml', '.github/autobuild/mac.sh', 'mac/deploy_mac.sh') }}-QT_VERSION=6.10.2 SIGN_IF_POSSIBLE=1 TARGET_ARCHS="x86_64 arm64" ./.github/autobuild/mac.sh
```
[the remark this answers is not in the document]
dtinth: Addressed in 8e2853b.

src/test/protocoltester.h
```diff
@@ -0,0 +1,290 @@
+/******************************************************************************\
+ * Copyright (c) 2026
+ *
+ * Author(s):
+ *  The Jamulus Development Team
```
ann0see: Probably add your name first

src/test/protocoltester.h
```diff
@@ -0,0 +1,290 @@
...
+ *
+ * Licensed under AGPL 3.0 or any later version. See COPYING for details.
```
ann0see: @pljones Is this sufficient?

src/test/test.pro
```diff
@@ -0,0 +1,53 @@
...
+QT -= gui
+
+# compile the sources under test in a headless server-only configuration so
+# that no GUI or sound interface dependencies are pulled in
+DEFINES += APP_VERSION=\\\"unittest\\\" \
+    SERVER_ONLY \
+    HEADLESS \
+    NO_JSON_RPC \
+    HAVE_STDINT_H \
+    QT_NO_DEPRECATED_WARNINGS
```
ann0see: Why don't you show those?

ann0see (changes requested):

ann0see:
There's no `make test` possibility?

dtinth:
@ann0see From Qt’s doc, [the convention for running tests with `qmake` is `make check`](https://doc.qt.io/qt-6/qmake-common-projects.html#building-a-testcase).

src/test/test.pro
dtinth: @ann0see Thanks for catching this.

Dropping `QT_NO_DEPRECATED_WARNINGS` surfaced a deprecation warning the build output on `util.cpp:1556` (`QLibraryInfo::location()` → `path()`, deprecated in Qt 6). My agent [hid](https://github.com/dtinth/jamulus/issues/3#issuecomment-5060730662) it as CI noise but I agree that the the test workflow should surface it rather than hide it. Addressed in ccbb7f5605cbd16cfc23d41685ee69f504a718c4.

src/test/protocoltester.h
dtinth: Replaced with "dtinth" following the precedent in [`rpcserver.cpp`](https://github.com/jamulussoftware/jamulus/blob/97799184a55c1c62d7b4b9064ace4ae1fbab380b/src/rpcserver.cpp#L5). Addressed in ccbb7f5605cbd16cfc23d41685ee69f504a718c4.

.github/scripts/run-unit-tests.py
```diff
@@ -0,0 +1,189 @@
...
+    binary = test_binary_path()
+    junit_format = pick_junit_format ( binary )
+    print ( "Using QtTest JUnit logger format: " + junit_format )
+
+    # "-o file,txt" not "-o -,txt": on windows-latest runners this binary's
+    # stdout comes back 0 bytes regardless of capture method (suspected: Qt's
+    # console detection on the inherited handle); QtTest's own file writer
+    # works there. Using it on Unix too avoids a platform branch.
+    subprocess.run ( [ binary, "-o", "test-output.txt,txt", "-o", "test-results.xml," + junit_format ] )
+
```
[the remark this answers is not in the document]
dtinth: Addressed in 0b5c9dcb7439aaed94223bb36277dd37e8a43322.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,221 @@
...
+  pull_request:
+    branches: [ main ]
+    paths:
+      - 'src/**'
+      - '.github/workflows/unit-tests.yml'
+      - '.github/scripts/**'
+
+permissions:
+  contents: read
+
```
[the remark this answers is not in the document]
dtinth: False positive.

This workflow has saved caches under exactly this permissions block ([evidence run](https://github.com/dtinth/jamulus/actions/runs/29929652776): `Cache saved with key: macos-qt-…`), and autobuild.yml saves caches under the stricter `permissions: {}`. actions/cache uses the runner's cache-service token rather than GITHUB_TOKEN scopes.

src/test/protocoltester.h
dtinth: @ann0see Updated to full AGPL header in 6094fbe. However, as this is new code, the clause about existing code being GPL is dropped from this file.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,221 @@
...
+          path: coverage-html
+          retention-days: 31
+          if-no-files-found: error
+
+      # `if: always()` so this still runs even if Run unit tests failed.
+      - name: Summarize test results
+        if: always()
+        env:
+          JOB_NAME: ${{ matrix.config.name }}
+        run: python3 .github/scripts/summarize-test-results.py
```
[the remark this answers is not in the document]
mcfnord: Shocking that python3 is not reliably available on GitHub-hosted Windows runners.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,221 @@
...
+        run: |
+          # windows.ps1 hardcodes this Qt/MSVC pairing internally (no env
+          # override), so it must be kept in sync with it by hand.
+          Add-Content -Path $env:GITHUB_PATH -Value "C:\Qt\${{ env.QT_VERSION }}\msvc2022_64\bin"
+
+      - name: Build unit tests
+        env:
+          QMAKE_BIN: ${{ matrix.config.qmake }}
+          QMAKE_EXTRA_ARGS: ${{ matrix.config.qmake_extra_args }}
+        run: python3 .github/scripts/run-unit-tests.py build
```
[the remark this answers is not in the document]
dtinth: False positive.

[Please read CI logs next time.](https://github.com/jamulussoftware/jamulus/actions/runs/30018138051/job/89243465707?pr=3828) (It probably won't.)

.github/workflows/unit-tests.yml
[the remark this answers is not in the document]
dtinth: False positive.

When a review agent is so focused on finding faults it can make up claims like this.

In one project, a review agent flagged our usage of a payment service, claiming that the amount is supposed to be in **cents**, not **dollars**. We would be losing so much money if I weren't suspicious.

(A good agent would verify the fault finder's claim before posting. Copilot isn't one of them.)

.github/scripts/run-unit-tests.py
```diff
@@ -0,0 +1,196 @@
...
+    os.makedirs ( BUILD_DIR, exist_ok=True )
+    qmake_bin = os.environ.get ( "QMAKE_BIN", "qmake" )
+
+    if platform.system() == "Windows":
+        apply_msvc_environment()
+        run (
+            [ qmake_bin, "..\\src\\test\\test.pro", "CONFIG-=debug_and_release", "CONFIG+=release", "DESTDIR=." ],
+            cwd=BUILD_DIR,
+        )
+        run ( [ "nmake" ], cwd=BUILD_DIR )
```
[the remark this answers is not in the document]
dtinth: Addressed in e859ee9.

.github/scripts/run-unit-tests.py
```diff
@@ -0,0 +1,196 @@
...
+        return "gate failed: {} could not be parsed ({})".format ( RESULTS_PATH, e )
+
+    total_tests = sum ( int ( suite.get ( "tests", 0 ) ) for suite in suites )
+    total_failed = sum ( int ( suite.get ( "failures", 0 ) ) + int ( suite.get ( "errors", 0 ) ) for suite in suites )
+
+    if total_tests == 0:
+        return "gate failed: {} reported 0 tests".format ( RESULTS_PATH )
+
+    if total_failed != 0:
+        return "gate failed: {} of {} tests failed".format ( total_failed, total_tests )
```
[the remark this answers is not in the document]
dtinth: Addressed in 9a26abc7. The failed-tests gate message now names the results file, matching the other two gate messages.

.github/scripts/run-unit-tests.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+def qmake_extra_args():
+    return shlex.split ( os.environ.get ( "QMAKE_EXTRA_ARGS", "" ) )
+
+
+def run ( cmd, **kwargs ):
+    print ( "+ " + " ".join ( cmd ) )
+    subprocess.run ( cmd, check=True, **kwargs )
+
```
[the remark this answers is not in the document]
dtinth: Addressed in eb2145b7. Both scripts reformatted with black to match the repo's existing Python style; verified behavior-identical (the AST is unchanged before/after).

.github/scripts/summarize-test-results.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def write_summary ( summary_path, lines ):
+    if not summary_path:
+        sys.stdout.writelines ( lines )
+        return
+
+    with open ( summary_path, "a", encoding="utf-8" ) as f:
+        f.writelines ( lines )
+
```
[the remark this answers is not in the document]
dtinth: Addressed in eb2145b7. Both scripts reformatted with black to match the repo's existing Python style; verified behavior-identical (the AST is unchanged before/after).

ann0see:
GitHub doesn't show any individual test cases like gitlab - I suppose

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,221 @@
...
+      - '.github/workflows/unit-tests.yml'
+      - '.github/scripts/**'
+
+permissions:
+  contents: read
+
+env:
+  # Must match autobuild.yml's macOS/Windows QT_VERSION -- Qt setup below
+  # reuses those jobs' scripts and cache.
+  QT_VERSION: 6.10.2
```
ann0see: Then we should try to find out how to only have one source of truth (the autobuild.yaml) to set this.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,221 @@
...
+
+  # gcovr version used by the coverage job below, pinned for reproducible output.
+  GCOVR_VERSION: 8.6
```
ann0see, line 25: Should then be bumped like our other dependencies automatically

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,221 @@
...
+          key: windows-${{ hashFiles('.github/workflows/autobuild.yml', '.github/autobuild/windows.ps1', 'windows/deploy_windows.ps1') }}-powershell .\.github\autobuild\windows.ps1 -Stage
+
+      # Reuses autobuild's own Qt setup instead of a separate aqtinstall call,
+      # so the cache above is genuinely shared rather than duplicated --
+      # heavier than this suite needs (qttools/qttranslations/qtmultimedia
+      # besides qtbase).
+      - name: Install Qt (macOS, via autobuild's mac.sh)
+        if: runner.os == 'macOS'
+        env:
+          JAMULUS_BUILD_VERSION: 0.0.0 # required by the script's validation; unused by "setup"
```
ann0see: Then the script should be updated such that it doesn't need this if we just run setup

dtinth:
@ann0see GitHub doesn't have any built-in test reporting AFAIK, only [job summaries](https://github.blog/news-insights/product-news/supercharging-github-actions-with-job-summaries/).

.github/workflows/unit-tests.yml
dtinth: Normally I would extract the common setup step into a [composite action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action) which gets reused by both workflows. I think this is cleaning but result in a larger change.

My agent also [recommended](https://github.com/dtinth/jamulus/issues/3#issuecomment-5062288467) another alternative: Have `unit-tests.yml` parse `autobuild.yml` to extract `QT_VERSION=`. My knee-jerk reaction is that it sounds quite hacky to me, but it works, keeps a single source of truth, and doesn't require refactoring existing workflows.

Do you have a preference?

(Merging as-is or using the `autobuild.yml` approach, and then doing proper action refactoring as a follow-up PR is also an option.)

.github/workflows/unit-tests.yml
dtinth: gcovr version bump added in 383c143f.

.github/workflows/unit-tests.yml
dtinth: `JAMULUS_BUILD_VERSION` validation is now optional in `setup` in ebb39cebe2a05d43d44a1df47677f505ceb38553.

src/test/protocoltester.h
```diff
@@ -0,0 +1,306 @@
...
+
+    /* frame mutation helpers -----------------------------------------------
+     * Operate in place on a frame the test holds, for crafting rows/inputs
+     * that a real Create*Mes() call can't produce.
+     */
+
+    // chops iBytes off the end of vecbyFrame, e.g. to cut into the CRC or the
+    // body; clamped at 0 so an oversize iBytes yields an empty frame instead
+    // of a negative resize (which wraps to a huge size_t).
+    static void TruncateBy ( CVector<uint8_t>& vecbyFrame, const int iBytes ) { vecbyFrame.resize ( std::max ( vecbyFrame.Size() - iBytes, 0 ) ); }
```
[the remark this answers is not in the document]
dtinth: Addressed in 0df93f2f. TruncateBy now clamps in both directions — negative iBytes is a no-op, oversize yields an empty frame — so it can never grow the frame.

.github/workflows/unit-tests.yml
ann0see: Not merging as is. I think for now we could parse autobuild.yaml but if you prefer the other cleaner approach it would be better.

ann0see (commented):
Curious of other test cases then too obviously. But not part of this PR.

.github/workflows/unit-tests.yml
dtinth: Done in 3b4612ae — took the parse approach for now (a bigger refactor queued for a more focused, follow-up PR).

test/protocoltester.h
```diff
@@ -0,0 +1,310 @@
+/******************************************************************************\
```
ann0see, line 1: What is your take on moving those files to /test/

dtinth: Normally when I work on JS projects I put the test files next to the source files (so the tests are "co-located" with the source files).

For example, if the implementation file is at `src/packlets/foo/util.js`, then the test would be in either of these:

- `src/packlets/foo/util.test.js`
- `src/packlets/foo/__tests__/util-test.js`

But when it comes to Qt projects, I have no takes. My guess is that it's only there so that the build system doesn't have to be changed to become aware of a new top-level folder. Let me ask my agent about this and I will get back.

dtinth: I [checked](https://github.com/dtinth/jamulus/issues/3#issuecomment-5077580774) with the agent. I have no strong preference. We can move to `/test/` if you prefer but we'll also have to:

- Update [`update-copyright-notices.sh`](https://github.com/jamulussoftware/jamulus/blob/main/tools/update-copyright-notices.sh#L56) to also scan the `test` directory.
- Update [`Jamulus.pro`](https://github.com/jamulussoftware/jamulus/blob/main/Jamulus.pro#L1194) so that `make clang_format` also work on the `test` directory.

What's your preference?

ann0see: Intuitively I'd expect tests to be in /test/

dtinth: Moved in bc226653.

- `src/test/` → `test/`
- Updated `update-copyright-notices.sh` find list
- Updated `Jamulus.pro`'s `clang_format` regex
- Updated the workflow `paths:` filters
- Fixed relative path in `test.pro`
- Updated the runner script.

[Full test matrix green on my fork](https://github.com/dtinth/jamulus/actions/runs/30150926477).

test/protocoltester.h
```diff
@@ -0,0 +1,310 @@
...
+    {
+        vecbyFrame.resize ( std::max ( vecbyFrame.Size() - std::max ( iBytes, 0 ), 0 ) );
+    }
+
+    // flips every bit of the trailing CRC byte so the checksum no longer
+    // matches the header plus body
+    static void CorruptCRC ( CVector<uint8_t>& vecbyFrame )
+    {
+        vecbyFrame[vecbyFrame.Size() - 1] = static_cast<uint8_t> ( vecbyFrame[vecbyFrame.Size() - 1] ^ 0xFF );
+    }
```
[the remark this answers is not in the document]
dtinth: Addressed in 33a89721. Real: `TruncateBy` can legally produce an empty frame, so chaining it into `CorruptCRC` would index the last byte of nothing. It's now a guarded no-op on empty frames.

pljones:
1) This should be "tooling" (i.e. part of the Github automated build process / command line build process)
2) Given the above, I'd rather keep it out of scope for 4.0.0

ann0see:
@coderabbit review as if you were an adversary

Issue #3402, opened 2024-10-08
Rewrite cclientdlg closeEvent


ann0see:
Maybe `closeEvent` should just call `QCoreApplication::instance()->exit();` and then do the work in `OnAboutToQuit` handlers?  That should stop needing to call `XYZ.Stop()`.

Of course, if that's what the "default implementation of this event handler routine" does, it would move all the code here into `ClientDlg.OnAboutToQuit` (plus handlers in each called class).

_Originally posted by @pljones in https://github.com/jamulussoftware/jamulus/pull/3372#discussion_r1763615127_

PR #3831, opened 2026-07-23
Fix server GUI not exiting on KDE (#3016)

### LINKED ISSUE #3016: Current KDE prevents Server GUI from exiting
### Discussed in https://github.com/orgs/jamulussoftware/discussions/3006

<div type='discussions-op-text'>

<sup>Originally posted by **pcar75** February  9, 2023</sup>
Bonjour,
On my dual boot PC, I can shutdown Jamulus server in Windows 10.  However I can't do that in Linux, because I believe it installs itself as a service.  Is there a nicer (GUI) way of shuting down Jamulus server, instead of running a script or terminal with  `killall -I "jamulus"` ?

Context : dual Boot PC : Windows 10 / Ubuntu Studio 22.04.1 LTS (Plasma KDE).
On Windows
    Jamulus server (3.9.1) with GUI
        Menu > Window > Exit (Ctrl+Q) : OK terminates process & threads
        X Button : OK terminates process & threads
On Ubuntu Studio 22.04.1 LTS
    Jamulus server (3.9.1) with GUI and Icon in system tray bar
        Menu > Window > Exit (Ctrl+Q) : Hides GUI only
        X (Close) Button : Hides GUI only
        Sys tray Exit (Icon menu buttons = Exit, Show, Hide...) : Hides GUI if it was visible ( ~ Exit ICON menu).

EDIT : added precision : Ubuntu Studio 22.04.1 LTS and Qt is 5.15.3 as mentioned below (rdica).</div>

It seems that latest KDE on at least Ubuntu 22.04+ and Fedora37 has introduced a change that prevents exiting the server process when the server GUI is started. Exiting merely closes the GUI window, but leaves the tray icon running as well as the server process itself.

As noted in the original discussion thread, disabling the tray icon functionality returns proper exit behaviour.
Personally I feel this is a bug and as current and future linux distros using KDE desktops will continue to experience the problem, it should be addressed as such, however it's been deemed not a bug, so the following request.

Request that either the tray icon be removed altogether, or a toggle be added to disable it.

Thanks!

mcfnord:
> [!NOTE]

Fixes #3016.

On some desktop environments — reported on KDE with Qt 5.15 (Ubuntu 22.04+, Fedora 37+) — closing the server GUI (X button, **Window → Exit**, Ctrl+Q, or the tray **Exit** entry) only *hides* the window. The process and the system tray icon keep running, and the only way out is `killall`. On Windows the same actions terminate the process as expected.

This applies the pragmatic fix that project member @dingodoppelt reported as working in [the issue thread](https://github.com/jamulussoftware/jamulus/issues/3016#issuecomment-4701276604): explicitly call `QApplication::quit()` at the end of `CServerDlg::closeEvent`, after the geometry is saved and the event is accepted.

```cpp
void CServerDlg::closeEvent ( QCloseEvent* Event )
{
    // store window positions
    pSettings->vecWindowPosMain = saveGeometry();

    // default implementation of this event handler routine
    Event->accept();

    // ... KDE workaround ...
    QApplication::quit();
}
```

**Scope:** one added call plus the `#include <QApplication>` it needs — server dialog only, exit path only. Nothing on the client, the audio path, or any running-state behaviour is touched.

**This is deliberately the pragmatic fix, not the dignified one.** As @dingodoppelt noted, `QApplication::quit()` here is a workaround, not the clean design — the underlying cause looks like threads not being torn down in time, and the proper solution is likely a signals/slots-based shutdown (the discussion in #3372 / #3402 / #3654 is the right home for that). The trade-off this PR offers is: KDE users get a server that actually exits *today*, at the cost of a line everyone agrees is a placeholder.

So — **if you would rather hold out for the real, official, proper shutdown refactor, please just close this PR.** No offence taken. It is here purely so the option of shipping the known-working one-liner now is on the table, in case that is preferable to leaving a three-year-old "the app won't quit" bug open while the larger design is worked out.

**Testing note, stated plainly:** the change compiles cleanly on desktop Linux (Qt 5). I have not reproduced the original KDE hang on my own setup, so I cannot personally confirm the fix on affected hardware — its credibility rests on @dingodoppelt having tested exactly this call and reported it resolved the problem. A confirmation from anyone who can reproduce the original bug would be welcome before merge.

### DIFF
--- a/src/serverdlg.cpp
+++ b/src/serverdlg.cpp
@@ -524,6 +524,13 @@ void CServerDlg::closeEvent ( QCloseEvent* Event )
 
     // default implementation of this event handler routine
     Event->accept();
+
+    // On some desktop environments (observed on KDE with Qt 5.15, Ubuntu 22.04+
+    // and Fedora 37+) closing the server GUI only hides the window while the
+    // process and system tray icon keep running (#3016). Explicitly quitting the
+    // application on close restores the expected behaviour. This is a pragmatic
+    // fix; a signals/slots-based shutdown may be the cleaner long-term solution.
+    QApplication::quit();
 }
 
 void CServerDlg::OnStartOnOSStartStateChanged ( int value )
--- a/src/serverdlg.h
+++ b/src/serverdlg.h
@@ -46,6 +46,7 @@
 
 #pragma once
 
+#include <QApplication>
 #include <QCloseEvent>
 #include <QLabel>
 #include <QListView>


dingodoppelt:
I'd rather not do the quick and dirty fix since we have an underlying issue which is maybe even entangled with other issues we see.  
As with the reverb: we might just get rid of the tray icon altogether. I wonder why we don't have a tray icon for the client, where it would make much more sense.

ann0see:
What would a tray icon for the client really gain? For the server it definitely makes sense to have an easy GUI way to interact with a backgrounded process.

dingodoppelt:
> What would a tray icon for the client really gain? For the server it definitely makes sense to have an easy GUI way to interact with a backgrounded process.

In practice the client is backgrounded for many users because the screen is used up by either sheet music or a video chat app. Some even need to walk away from their computers to be able to play their instruments (Drummers, Pianists, Double Bass players, ...)  
I'd say a tray icon indicating the mute status via colour and containing some shortcuts like favourite servers or the connect dialogue (or maybe a configurable set of shortcuts?)  
Ideally a user forgets about Jamulus while playing and a GUI only gets in the way once you are connected.

ann0see:
What's the concensus here? Merge? Close? I run on gnome but would probably be annoyed with the server not exiting unless issuing a killall.

dingodoppelt:
> What's the concensus here? Merge? Close? I run on gnome but would probably be annoyed with the server not exiting unless issuing a killall.

Since this is only a workaround and it doesn't seems to address the root cause I'd say: not merge.

PR #3832, opened 2026-07-23
Add optional "localtimezone;" prefix for the recording directory (#497)

### LINKED ISSUE #497: Incorporate local timezone in jam recorder folder
As per my [sourceforge post](https://sourceforge.net/p/llcon/discussion/server/thread/bdc3c34844/?limit=25#e0c9), have the recorder folder name reflect the current date/time of the local timezone.

The current recording functionality uses UTC date/time only - currentDateTimeUtc(). Request revising to use local timezone date/time - currentDateTime(). 

jamrecorder.cpp

mcfnord:
> [!NOTE]

Fixes #497.

This implements the design that was agreed in the issue thread in 2021 and — unusually — consolidated into the issue body itself as a buildable spec, at @gilgongo's request, by the reporter @victorypoint: an **opt-in `localtimezone;` prefix** on the existing recording-directory argument, using the `;` separator format @corrados proposed for consistency with other combined command-line values.

```
./Jamulus -s -R "localtimezone;/home/test/jamulusrecordings"
```

With the prefix, recording session folders are named using local time, so they match the server log's own (local-time) timestamps without manual UTC-offset math. Without it, **nothing changes**: UTC remains the default, preserving the property discussed in the thread that folder names on multi-timezone public servers stay unambiguous unless the operator explicitly opts in.

**How it works:**

- The prefix is recognised in `CJamController::SetRecordingDir` — the single chokepoint shared by all three ways of setting the directory (`-R`/ini, the server GUI, and JSON-RPC `jamulusserver/setRecordingDirectory`), so it behaves identically everywhere.
- The stored/displayed directory string keeps the prefix, so it round-trips unchanged through the ini file, `getRecorderStatus`, and the GUI — a server started once with the prefix keeps the behaviour across restarts.
- One `bool` is threaded through `CJamRecorder` to `CJamSession`, which picks `QDateTime::currentDateTime()` vs `currentDateTimeUtc()` for the session folder name. The folder-name *format* (`Jam-yyyyMMdd-HHmmsszzz`) is unchanged.
- `--help` text and `docs/JSON-RPC.md` (regenerated with `tools/generate_json_rpc_docs.py`) are updated.

**Deliberately not included:** the issue body's secondary wish to also switch the folder-name format to ISO (`yyyy-MM-dd HH:mm:ss`). That would change the format for every existing server and any tooling that parses it; the agreed compromise was the opt-in prefix, so this PR does only that.

**Testing (Linux, Qt 5):** built clean; functional test ran a headless server with `TZ=America/Los_Angeles` and a headless client to trigger a session, twice: with `-R "localtimezone;<dir>"` the session folder carried the local (UTC−7) timestamp matching the server log line; with a plain `-R <dir>` it carried the UTC timestamp, byte-identical behaviour to before. A prefix-only value (`"localtimezone;"` with no path) degrades to the existing "recording not initialised" path.


ann0see:
Just FYI: I have no strong preference for or against this and would rather not want to decide on inclusion or exclusion of this feature. I'd leave this up to pljones and gilgongo if that's good?

pljones:
It's definitely been requested -- but I doubt there's many people running servers with recording who really care about it.

Not for 4.0.0, for sure.  We've not fully scoped 4.1.0 but I'd rather just leave it on the project backlog for now.

ann0see:
We could put it on a feature branch. If someone requests it again we could merge it.

ann0see:
See: https://github.com/jamulussoftware/jamulus/tree/archive/mcfnord/localtimezonerecorder as feature branch. We can cherry-pick the code if needed and re-requested.

PR #3833, opened 2026-07-24
Build opus with NEON intrinsics on ARM (Android ABIs included)

### LINKED ISSUE #2806: ARM: Possibly missing opus optimizations
**Describe the bug**
It looks like none of the existing opus optimization rules match arm64:
https://github.com/jamulussoftware/jamulus/blob/e14f3dd9af84271c22575cf4ce7b10dbac0c2bdd/Jamulus.pro#L667

**To Reproduce**

To be investigated.

**Expected behavior**

M1 builds should use the opus ARM performance enhancements.

**Screenshots**

**Operating system**

Mac M1 / arm

**Version of Jamulus**

master

mcfnord:
**🤖 AI:** Fixes #2806.

`Jamulus.pro`'s opus ARM block matched only the Android ABI names (`armeabi-v7a`, `arm64-v8a`), so every other 64-bit ARM build (Apple Silicon macOS, iOS, Linux aarch64, which Qt reports as `QT_ARCH=arm64`) compiled opus as plain C. The block was also inert where it did match: `SOURCES += $$SOURCES_OPUS_ARCH` existed only in the x86 branch, and `OPUS_ARM_MAY_HAVE_NEON_INTR`, which gates opus's `arm/*.h` headers, was never defined. On main no target compiles a single `libs/opus/*/arm/` file.

**What this PR does**

- One ARM case for all three values of `QT_ARCH`: same headers, the 9 NEON intrinsic sources, `OPUS_ARM_MAY_HAVE_NEON_INTR=1 OPUS_ARM_PRESUME_NEON=1 OPUS_ARM_PRESUME_NEON_INTR=1`, plus `OPUS_ARM_PRESUME_AARCH64_NEON_INTR` for the 64-bit ones.
- A consumption branch that compiles those sources for the ARM cases. No special compiler flags: NEON is part of the base AArch64 ISA, and the NDK's clang enables it by default for `armv7a-linux-androideabi`.
- `celt_fft_ne10.c` and `celt_mdct_ne10.c` removed from the source list. They `#include <NE10_dsp.h>` unconditionally, Jamulus does not bundle Ne10, and no target has ever compiled them.

The Android ABIs therefore compile the NEON sources for the first time; x86 builds are untouched.

**Testing**

- `qmake`-resolved `Makefile.Release` per `QT_ARCH`: x86_64 identical before and after; `arm64`, `arm64-v8a`, `armeabi-v7a` each list the 9 ARM sources and the defines above; main lists 0 ARM sources for the Android ABIs.
- Native Raspberry Pi 4 build (aarch64, gcc 14, `CONFIG+=headless`): links, 9 `_neon` symbols in the binary, 0 undefined.
- NDK r21d clang with Qt 5.15.2's `android-clang` flags (`-target <abi>30 -Oz -fPIC`): all 138 opus sources compile for `armv7a-linux-androideabi30` and `aarch64-linux-android30`, 0 failures, and every `_neon` symbol the objects reference is defined in the set.
- CPU effect on a Pi 4 server was measured earlier in this thread as noise-level at both complexity settings; this is a build-correctness fix, not a performance claim.

---

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -660,7 +660,8 @@ SOURCES_OPUS = libs/opus/celt/bands.c \
     libs/opus/src/opus_encoder.c \
     libs/opus/src/repacketizer.c
 
-SOURCES_OPUS_ARM = libs/opus/celt/arm/armcpu.c \
+# NEON intrinsic sources: self-contained, need no external library.
+SOURCES_OPUS_ARM_NEON = libs/opus/celt/arm/armcpu.c \
     libs/opus/celt/arm/arm_celt_map.c \
     libs/opus/silk/arm/arm_silk_map.c \
     libs/opus/silk/arm/biquad_alt_neon_intr.c \
@@ -668,9 +669,14 @@ SOURCES_OPUS_ARM = libs/opus/celt/arm/armcpu.c \
     libs/opus/silk/arm/NSQ_del_dec_neon_intr.c \
     libs/opus/silk/arm/NSQ_neon.c \
     libs/opus/celt/arm/celt_neon_intr.c \
-    libs/opus/celt/arm/pitch_neon_intr.c \
-    libs/opus/celt/arm/celt_fft_ne10.c \
+    libs/opus/celt/arm/pitch_neon_intr.c
+# NE10 sources: FFT/MDCT acceleration via the external Ne10 library, which
+# Jamulus does not bundle (they #include <NE10_dsp.h> unconditionally and are
+# gated on HAVE_ARM_NE10, which we never define). Kept listed for completeness
+# but never added to SOURCES.
+SOURCES_OPUS_ARM_NE10 = libs/opus/celt/arm/celt_fft_ne10.c \
     libs/opus/celt/arm/celt_mdct_ne10.c
+SOURCES_OPUS_ARM = $$SOURCES_OPUS_ARM_NEON $$SOURCES_OPUS_ARM_NE10
 
 SOURCES_OPUS_X86_SSE = libs/opus/celt/x86/x86cpu.c \
     libs/opus/celt/x86/x86_celt_map.c \
@@ -690,6 +696,18 @@ contains(QT_ARCH, armeabi-v7a) | contains(QT_ARCH, arm64-v8a) {
     SOURCES_OPUS_ARCH += $$SOURCES_OPUS_ARM
     DEFINES_OPUS += OPUS_ARM_PRESUME_NEON=1 OPUS_ARM_PRESUME_NEON_INTR=1
     contains(QT_ARCH, arm64-v8a):DEFINES_OPUS += OPUS_ARM_PRESUME_AARCH64_NEON_INTR
+} else:contains(QT_ARCH, arm64) {
+    # arm64-v8a above is the Android ABI name; plain arm64 is what Qt reports
+    # for 64-bit ARM everywhere else (Apple Silicon macOS, iOS, Linux aarch64).
+    # NEON is part of the base AArch64 ISA, so the NEON intrinsics can be
+    # presumed present and compiled without special compiler flags (see the
+    # SOURCES_OPUS_ARCH handling below).
+    # OPUS_ARM_MAY_HAVE_NEON_INTR is required in addition to the PRESUME
+    # defines: opus gates the inclusion of its arm/*.h headers on it
+    # (see libs/opus/celt/pitch.h and libs/opus/celt/cpu_support.h).
+    HEADERS_OPUS += $$HEADERS_OPUS_ARM
+    SOURCES_OPUS_ARCH += $$SOURCES_OPUS_ARM_NEON
+    DEFINES_OPUS += OPUS_ARM_MAY_HAVE_NEON_INTR=1 OPUS_ARM_PRESUME_NEON_INTR=1 OPUS_ARM_PRESUME_AARCH64_NEON_INTR=1
 } else:contains(QT_ARCH, x86) | contains(QT_ARCH, x86_64) {
     HEADERS_OPUS += $$HEADERS_OPUS_X86
     SOURCES_OPUS_ARCH += $$SOURCES_OPUS_X86_SSE $$SOURCES_OPUS_X86_SSE2 $$SOURCES_OPUS_X86_SSE4
@@ -1172,6 +1190,11 @@ contains(CONFIG, "opus_shared_lib") {
             sse4_cc.variable_out = OBJECTS
             QMAKE_EXTRA_COMPILERS += sse_cc sse2_cc sse4_cc
         }
+    } else:contains(QT_ARCH, arm64) {
+        # Unlike the x86 SSE files above, the AArch64 NEON intrinsics need no
+        # special compiler flags (NEON is part of the base AArch64 ISA), so
+        # they can be compiled like any other source file.
+        SOURCES += $$SOURCES_OPUS_ARCH
     }
 }
 
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -660,8 +660,7 @@ SOURCES_OPUS = libs/opus/celt/bands.c \
     libs/opus/src/opus_encoder.c \
     libs/opus/src/repacketizer.c
 
-# NEON intrinsic sources: self-contained, need no external library.
-SOURCES_OPUS_ARM_NEON = libs/opus/celt/arm/armcpu.c \
+SOURCES_OPUS_ARM = libs/opus/celt/arm/armcpu.c \
     libs/opus/celt/arm/arm_celt_map.c \
     libs/opus/silk/arm/arm_silk_map.c \
     libs/opus/silk/arm/biquad_alt_neon_intr.c \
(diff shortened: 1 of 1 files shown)


ann0see:
Should have Autobuild please build all targets

Jamulus.pro
```diff
@@ -660,17 +660,23 @@ SOURCES_OPUS = libs/opus/celt/bands.c \
     libs/opus/src/opus_encoder.c \
     libs/opus/src/repacketizer.c
 
-SOURCES_OPUS_ARM = libs/opus/celt/arm/armcpu.c \
+# NEON intrinsic sources: self-contained, need no external library.
```
ann0see, line 663:
```suggestion
```

ann0see:
Also this needs to be built, tested and analyzed on Apple Silicon. The arm handling in general is sketchy. For macOS Apple Silicon the explicit variable setting of arm64 is needed otherwise it tries to compile SSE for arm64 which obviously doesn't work

Jamulus.pro
```diff
@@ -660,17 +660,23 @@ SOURCES_OPUS = libs/opus/celt/bands.c \
...
     libs/opus/silk/arm/NSQ_del_dec_neon_intr.c \
     libs/opus/silk/arm/NSQ_neon.c \
     libs/opus/celt/arm/celt_neon_intr.c \
-    libs/opus/celt/arm/pitch_neon_intr.c \
-    libs/opus/celt/arm/celt_fft_ne10.c \
+    libs/opus/celt/arm/pitch_neon_intr.c
+# NE10 sources: FFT/MDCT acceleration via the external Ne10 library, which
+# Jamulus does not bundle (they #include <NE10_dsp.h> unconditionally and are
+# gated on HAVE_ARM_NE10, which we never define). Kept listed for completeness
+# but never added to SOURCES.
```
ann0see, line 676: Then investigate if we can use it and it makes sense to use it - or drop it otherwise

Jamulus.pro
```diff
@@ -690,6 +696,18 @@ contains(QT_ARCH, armeabi-v7a) | contains(QT_ARCH, arm64-v8a) {
     SOURCES_OPUS_ARCH += $$SOURCES_OPUS_ARM
     DEFINES_OPUS += OPUS_ARM_PRESUME_NEON=1 OPUS_ARM_PRESUME_NEON_INTR=1
     contains(QT_ARCH, arm64-v8a):DEFINES_OPUS += OPUS_ARM_PRESUME_AARCH64_NEON_INTR
+} else:contains(QT_ARCH, arm64) {
```
ann0see, line 699: Try to merge this with the above case to avoid duplication

mcfnord:
> [!NOTE]

Native verification on a Raspberry Pi 4 (Cortex-A72, aarch64 Linux, not cross-compiled), including a result I didn't expect.

**Symbols:** built this branch and the same base commit without it, both natively. `nm` on the shipped binary confirms all 9 `*_neon` functions defined in the fixed build, 0 in the unfixed one — the fix links and ships correctly.

**CPU benefit:** interleaved A/B benchmark (server pinned to one core, bot load, `perf stat task-clock`), at `OPUS_SET_COMPLEXITY(1)` (Jamulus's actual setting) I measured no significant difference — and re-tested at `OPUS_SET_COMPLEXITY(10)` in case low complexity was masking a win: still no significant difference.

**Why:** confirmed with `perf record -g` under load — Jamulus uses `opus_custom_*` (Opus Custom, CELT-only; SILK never runs). 6 of the 9 accelerated functions are SILK-only and are dead code in Jamulus regardless of this fix. Of the 3 CELT functions, only 2 ever appeared under profiling, combined under 1% of server CPU.

**Bottom line:** still worth merging — dead ARM code silently no-op'ing is a real defect on its own terms — but I'd frame it as a correctness/build-hygiene fix rather than a proven Jamulus performance win. This only covers Linux aarch64; it doesn't address the Apple Silicon question above.

pljones:
I'd love to get my local (Ubuntu 24.04 LTS) Android build working.  Currently I rely on Github.

pljones:
#2806 is "mac" -- this is "android".  Is #2806 mis-titled?

mcfnord:
> [!NOTE]

Catching up on the thread:

**@ann0see, re: build all targets** — CI already did, on this PR's own run: Android APK, Linux arm64 `.deb`, armhf `.deb`, macOS, macOS Legacy, iOS, Windows, Windows JACK all pass (https://github.com/jamulussoftware/jamulus/actions/runs/30064659704). The macOS job builds a universal binary and `mac/deploy_mac.sh` explicitly sets `QT_ARCH=arm64` for the Apple Silicon slice, so that slice does compile clean with this fix. I don't have Apple hardware, so that's a compile proof only — no perf number for Apple Silicon from me, same caveat as my Pi 4 result below.

**@pljones, re: is #2806 mis-titled** — fair catch, and a bit of both: #2806's title says "Mac ARM," which matches how the issue was originally reported, but this PR's actual scope is wider than that title suggests. It touches every non-Android 64-bit ARM target that reports `QT_ARCH=arm64` (macOS Apple Silicon, iOS, Linux aarch64) — and separately, while implementing that, it surfaced a genuine Android-specific bug: Android's own arch match populated the NEON source list but two other gaps meant those sources were never actually compiled or linked, so Android has been silently running plain-C opus too, this whole time. That second part isn't really "about" #2806 at all — it's its own defect that happened to surface in the same investigation.

**My conclusion, for the record:** native Pi 4 (Cortex-A72, Linux aarch64) testing confirms the fix is real and correct — all 9 NEON symbols ship where 0 did before — but produces no measurable CPU win for Jamulus specifically, at either complexity=1 (Jamulus's actual setting) or complexity=10 (max), because Jamulus's CELT-only Opus Custom mode only ever touches 2 of those 9 functions, combined under 1% of server CPU. That's a Jamulus-specific result, not a NEON-doesn't-help result — other opus consumers using SILK or the general encoder path would very plausibly see more. I'd still call this worth merging on correctness/build-hygiene grounds (dead ARM code silently no-op'ing across four platforms is a real defect on its own), just not on a proven performance claim. That's everything I can verify from here — Apple Silicon and a from-scratch Android build both remain open if anyone has the hardware.

ann0see (commented):
...

Jamulus.pro
ann0see: @mcfnord

ann0see (commented):
Please clean up the comments as mentioned above.

Jamulus.pro
mcfnord: **🤖 AI:** All three [done in cbe7d44a](https://github.com/jamulussoftware/jamulus/pull/3833/commits/cbe7d44af6023881702119d71ed0cbd744e4042c). The comment line is gone. The [NE10 pair is dropped](https://github.com/jamulussoftware/jamulus/pull/3833#discussion_r3650887060): both files [`#include <NE10_dsp.h>` unconditionally](https://github.com/jamulussoftware/jamulus/blob/72e856afb8b9d3a0f95c48c255f0d2cc8ca03b06/libs/opus/celt/arm/celt_fft_ne10.c#L39) and Jamulus does not bundle Ne10, so no target has ever compiled them. The arm64 case is [merged into the Android one](https://github.com/jamulussoftware/jamulus/pull/3833#discussion_r3650888111), and the merge has a consequence: on main the Android ABIs [define the PRESUME macros](https://github.com/jamulussoftware/jamulus/blob/72e856afb8b9d3a0f95c48c255f0d2cc8ca03b06/Jamulus.pro#L689-L693) but compile zero `libs/opus/*/arm/` files, because the SOURCES consumption branch [covers only x86](https://github.com/jamulussoftware/jamulus/blob/72e856afb8b9d3a0f95c48c255f0d2cc8ca03b06/Jamulus.pro#L1139). Now all three ARM values of QT_ARCH compile the nine NEON files. Checked: the x86_64 Makefile is unchanged; a native Pi 4 build links the [same nine `_neon` symbols](https://github.com/jamulussoftware/jamulus/pull/3833#issuecomment-5081788647) as before; [NDK r21d](https://github.com/jamulussoftware/jamulus/blob/72e856afb8b9d3a0f95c48c255f0d2cc8ca03b06/.github/autobuild/android.sh#L32) clang with [Qt 5.15.2's flags](https://github.com/qt/qtbase/blob/40143c189b7c1bf3c2058b77d00ea5c4e3be8b28/mkspecs/features/android/default_pre.prf#L26-L29) compiles all 138 opus files for both Android ABIs, NEON being its default for armv7a-linux-androideabi. The Android CI job is [green on this head](https://github.com/jamulussoftware/jamulus/actions/runs/33988102663/job/101365324027).

[CodeRabbit's summary](https://github.com/jamulussoftware/jamulus/pull/3833#issuecomment-5554344827) took 142 lines to report that it had nothing actionable to say; a review with nothing to say has one right length.

ann0see:
No. This is all good. @coderabbitai ! 

macOS also seems to be fine now. Thank you very much

ann0see (approved):

ann0see:
> **Comment — duplicate source entry (now live)**

`SOURCES_OPUS_ARM` lists `libs/opus/silk/arm/arm_silk_map.c` twice (lines 665–666). This predates the PR, but it matters now: previously the ARM sources were never consumed, so the duplicate was dead weight. After this PR every ARM target compiles this list, and this is the kind of thing that silently passes qmake (which dedupes `SOURCES`) but can break other tooling or confuse readers. It would also let the PR body align its "9 NEON intrinsic sources" count with the actual 10 entries / 9 unique files.

Suggested fix in the same commit:
```diff
     libs/opus/silk/arm/arm_silk_map.c \
-    libs/opus/silk/arm/arm_silk_map.c \
```

> 🤖 AI: opencode big-pickle

ann0see:
> **Comment — `OPUS_ARM_PRESUME_NEON_INTR` on 32-bit ARM removes runtime detection**

The PR sets `OPUS_ARM_PRESUME_NEON_INTR=1` for `armeabi-v7a` as well as the 64-bit ABIs. On AArch64 that is fine — NEON is part of the base ISA. On 32-bit ARMv7 it is not: the ABI spec treats NEON (and VFPv3-D32) as *optional*, and NDK builds enabled it by default only via a toolchain default, not an ISA guarantee. `PRESUME` means the binary unconditionally executes NEON intrinsics with no feature check; on an ARMv7 device without NEON that is an illegal-instruction crash at runtime, not a build error.

Since `OPUS_HAVE_RTCD=1` is already defined for every target (`Jamulus.pro:702`), the safer 32-bit story is `OPUS_ARM_MAY_HAVE_NEON_INTR` + runtime dispatch — the RTCD tables in `arm_celt_map.c` / `arm_silk_map.c` are already emitted for that path. In practice modern Android requires NEON on armv7 (so this is low-risk), but if 32-bit is knowingly acceptable, a comment stating "we deliberately PRESUME NEON on armv7; non-NEON devices are out of scope" would make the decision explicit rather than accidental.

> 🤖 AI: opencode big-pickle

pljones (approved):
I can't measure it or test it right now but it looks focused enough.

ann0see:
Still merging. @Coderabbitai please open issues for your findings. In the best case a PR for the duplicate.

PR #3835, opened 2026-07-25
Fix MIDI pickup crossing detection never firing

### LINKED ISSUE #3942: MIDI Pick-up Mode: the rapid-movement branch is unreachable
**🤖 AI:** Pick-up Mode's rapid-movement handling never executes. [Its help text](https://github.com/jamulussoftware/jamulus/blob/ee40cfe20ecebe92eeec9eecc2b9ee8b948f439f/src/clientsettingsdlg.cpp#L427-L430) promises a control waits until the physical controller matches the software value, and [an inner branch](https://github.com/jamulussoftware/jamulus/blob/ee40cfe20ecebe92eeec9eecc2b9ee8b948f439f/src/audiomixerboard.cpp#L94-L96) exists to also catch a controller swept fast enough that its values *"skip over"* that value. That branch is unreachable, for two independent reasons.

`midiPickupTryApply` [returns before appending](https://github.com/jamulussoftware/jamulus/blob/ee40cfe20ecebe92eeec9eecc2b9ee8b948f439f/src/audiomixerboard.cpp#L119-L129) to the history buffer, so while a control is armed the buffer stays empty and the `size() >= 2` guard in front of the crossing test never holds. Independently, [`prevMidi` reads `back()`](https://github.com/jamulussoftware/jamulus/blob/ee40cfe20ecebe92eeec9eecc2b9ee8b948f439f/src/audiomixerboard.cpp#L97-L99), while [the sole call site](https://github.com/jamulussoftware/jamulus/blob/ee40cfe20ecebe92eeec9eecc2b9ee8b948f439f/src/audiomixerboard.cpp#L113-L119) has already pushed the incoming sample onto that same buffer — so `prevMidi` is the incoming sample, and the bracket test reduces to `midiValue == currentValue`, which the tolerance check above it has already accepted.

Driving those helpers verbatim across every strictly-bracketing two-sample crossing of the fader range, both samples outside the tolerance: on ee40cfe2, 0 of 147,440 pick up. Repairing either defect alone leaves 0. Repairing both gives 147,440. What remains on stock is the tolerance window by itself — armed at fader 50, 5 of 101 controller positions are accepted, 48 through 52 — so a fader swept briskly steps over the window and stops responding, which is the case the unreachable branch was written for.

Present since [the MIDI GUI PR](https://github.com/jamulussoftware/jamulus/pull/3502).

mcfnord:
> [!NOTE]

Fixes: #3942

MIDI pickup mode has a second acceptance path beside the tolerance test, for when a fast fader move makes consecutive MIDI values skip straight over the software value:

```cpp
// If buffer has at least 2 values, check if we're crossing the current value
// Handles the case where rapid movement causes MIDI values to "skip over" the software value
if ( recentMidiValues.size() >= 2 )
{
    int prevMidi = recentMidiValues.back();
    ...
}
```

It has never run. There are two independent defects, and either one alone is enough to disable it.

**1. The history is never recorded while waiting.** `midiPickupTryApply` builds a temporary copy, tests it, and on failure returns early — jumping over the `// Update the pickup buffer` block at the end of the function. So while `waitingForPickup` is true, which is exactly the state the history exists to serve, `pickupBuffer` stays empty. The largest deque ever passed to `midiPickupShouldApply` is 1, so `size() >= 2` is never satisfied.

**2. `prevMidi` is not the previous value.** `recentMidiValues.back()` is the element `midiValue` was just `push_back`'d as, so `prevMidi == midiValue` and the bracket test reduces to `midiValue == currentValue` — already covered by the tolerance test three lines above. Sweeping `(prev, current, new)` over 0..100 outside tolerance, the current code detects **0 of 323400** genuine crossings.

The `size() >= 2` guard is itself the evidence of intent: you only need two elements if you mean to look past the last one.

## What it costs a user

`CSoundBase::ParseMIDIMessage` maps CC 0..127 onto fader 0..`AUD_MIX_FADER_MAX` (100), so one CC step is 0.79 fader units against a `MIDI_PICKUP_TOLERANCE` of 2. Slow moves are picked up by the tolerance test and the bug stays invisible. Fast moves are where it shows: sweeping a hardware fader past a software fader sitting at 50, the fader stays dead at several speeds until the user slows down and lands within ±2 units.

| CC step | before | fix 1 only | fix 2 only | this PR | client, before / this PR |
|---|---|---|---|---|---|
| 8 | picks up (9 msgs) | picks up (9) | picks up (9) | picks up (9) | 9 / 9 |
| 10 | **never picks up** | **never** | **never** | picks up (8) | never / 8 |
| 13 | picks up (6) | picks up (6) | picks up (6) | picks up (6) | 6 / 6 |
| 20 | **never picks up** | **never** | **never** | picks up (5) | never / 5 |
| 24 | **never picks up** | **never** | **never** | picks up (4) | never / 4 |
| 32 | picks up (3) | picks up (3) | picks up (3) | picks up (3) | 3 / 3 |

The two middle columns are the reason this PR changes two things rather than one: fixing either defect on its own changes nothing measurable.

Those six rows are a sample. Across the first 32 step sizes, fader parked at 50, **16 never pick up on `main` and 0 never pick up with this PR**, the smallest failing step being 10. Under that size the tolerance test absorbs the move: at one CC value per message, sweeping against every parked position 0..100 gives **0 of 101 positions that fail on either build**. So a controller emitting its full CC stream cannot expose this at any speed, and the shortest reproduction is to send CC in steps of 10 or more.

## Checks

Harness compiles the two patched functions verbatim out of `audiomixerboard.cpp` rather than a re-typed copy, so what is measured is what is shipped.

- All previously stuck sweep speeds now pick up.
- Both directions: the exhaustive two-sample crossing sweep run upward and then downward — 147440 crossings each way, disjoint sets — picks up **147440 of 147440** with both edits and **0 of 147440** with either edit alone.
- Regression on the property the feature exists for: pickup must never engage on the first message when the hardware fader is far from the software fader. Across all 12928 combinations of software position 0..100 and CC 0..127 outside tolerance, first-message jumps: **0**.
- `clang-format` 14.0.6 reports no change; compiles clean under `-Wall -Wextra`.

**Runtime, through the shipped client.** A JACK MIDI source — not physical hardware, but real MIDI arriving on the port `CSound::CreateMIDIPort` opens — feeds a GUI client with one channel fader parked at 50 and Pick-up Mode on, and the observable is the `CHANNEL_GAIN` message the client puts on the wire. Two control-change messages 298.12 ms apart carrying CC 51 and CC 76 — fader 40 and 59, bracketing the parked 50, both outside tolerance — produce no gain message and no fader movement on `main` at `72e856af`. The same two messages on that tree with this PR's two edits applied pick up and send the gain for 59. Sweeping that same span one CC step at a time on `main` instead picks up at 48, the tolerance edge, and tracks from there: 12 gain messages. The sweep table's last column is that same client — 12 runs, one per step and build, the controller stepping up from CC 0 one message every 30 ms — and every cell agrees with the harness. Reproducers, including one that needs no MIDI hardware at all: [this gist](https://gist.github.com/mcfnord/fdd88cdbe5265a9dbf4cf8e12841851d).

## What I did not do

One related thing I noticed and deliberately left alone: `g_midiPickupWaitingForPickup` is a single flag per channel shared by the fader *and* the pan, while the history deques and timestamps beside it in `MidiPickupState` are correctly per-control (`recentFader`/`recentPan`, `lastMidiTimeFader`/`lastMidiTimePan`). That means picking up the pan also clears the fader's waiting state, and either control's inactivity timeout re-arms both. It needs two flags rather than one, which is a different change from this one, so I have not folded it in. I can open it separately if you would like.

CHANGELOG: Bugfix: MIDI pickup mode now correctly detects a fader crossing the software value during fast moves.

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -95,7 +95,7 @@ static bool midiPickupShouldApply ( int midiValue, int currentValue, int toleran
     // Handles the case where rapid movement causes MIDI values to "skip over" the software value
     if ( recentMidiValues.size() >= 2 )
     {
-        int prevMidi = recentMidiValues.back();
+        int prevMidi = recentMidiValues[recentMidiValues.size() - 2];
         // Check if current and previous MIDI values bracket the software value
         if ( ( prevMidi <= currentValue && midiValue >= currentValue ) || ( prevMidi >= currentValue && midiValue <= currentValue ) )
             return true;
@@ -117,7 +117,11 @@ static bool midiPickupTryApply ( int midiValue, int currentValue, int tolerance,
         tempPickup.push_back ( midiValue );
 
         if ( !midiPickupShouldApply ( midiValue, currentValue, tolerance, tempPickup ) )
+        {
+            // keep this value: the next message needs it to detect a crossing
+            pickupBuffer = tempPickup;
             return true; // Still waiting for pickup
+        }
 
         // Picked up. Stop waiting
         waitingForPickup = false;
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -118,7 +118,8 @@ static bool midiPickupTryApply ( int midiValue, int currentValue, int tolerance,
 
         if ( !midiPickupShouldApply ( midiValue, currentValue, tolerance, tempPickup ) )
         {
-            // keep this value: the next message needs it to detect a crossing
+            // still waiting: keep this value in the channel's recentFader/recentPan
+            // history so the next message can detect a crossing
             pickupBuffer = tempPickup;
             return true; // Still waiting for pickup
         }


src/audiomixerboard.cpp
```diff
@@ -95,7 +95,7 @@ static bool midiPickupShouldApply ( int midiValue, int currentValue, int toleran
     // Handles the case where rapid movement causes MIDI values to "skip over" the software value
     if ( recentMidiValues.size() >= 2 )
     {
-        int prevMidi = recentMidiValues.back();
+        int prevMidi = recentMidiValues[recentMidiValues.size() - 2];
```
ann0see, line 98: Can this get out of bounds? E.g access -2?

ignotus666: Since `if ( recentMidiValues.size() >= 2 )`, there shouldn't be any danger of that.

The problem raised is in fact correct and the fix looks like it's correct too. I haven't had time to test it and don't know when I will - hopefully some time next week. This bug probably went unnoticed in testing by me because I modified my controller to output the full stream of CC messages even when moving it fast.

mcfnord: **🤖 AI:** ignotus666 got it right — the `size() >= 2` guard is what prevents it, and now it's measured.

Swept with the two patched functions extracted verbatim into a driver built `-fsanitize=address,undefined -D_GLIBCXX_DEBUG`, which bounds-checks `std::deque::operator[]`: every `midiValue` x `currentValue` pair in 0..127, tolerance 0..8, ascending and descending sequences, starting from an empty buffer. 589,824 calls, buffer lengths 0, 1 and 2 all exercised, zero diagnostics. Since the buffer never exceeds `MIDI_PICKUP_HISTORY` (2), `size() - 2` is always index 0.

The guard is load-bearing: at `size() == 1`, `size() - 2` is unsigned and wraps to 18446744073709551615.

src/audiomixerboard.cpp
```diff
@@ -117,7 +117,11 @@ static bool midiPickupTryApply ( int midiValue, int currentValue, int tolerance,
         tempPickup.push_back ( midiValue );
 
         if ( !midiPickupShouldApply ( midiValue, currentValue, tolerance, tempPickup ) )
+        {
+            // keep this value: the next message needs it to detect a crossing
+            pickupBuffer = tempPickup;
```
pljones, line 122: Where's `pickupBuffer` defined and used?  (Maybe hint that in the comment - but if it's obvious, this comment is fine.)

ann0see: @mcfnord please have a look at the PRs you/your agent looked at and comment...

mcfnord: **🤖 AI:** `pickupBuffer` is a parameter name, not an object: [`midiPickupTryApply`](https://github.com/jamulussoftware/jamulus/blob/b4096fa6cbac37df3d391c25b43cecfa602e2a8b/src/audiomixerboard.cpp#L108-L109) takes the caller's deque by reference, and [`midiPickupInactivityCheck`](https://github.com/jamulussoftware/jamulus/blob/b4096fa6cbac37df3d391c25b43cecfa602e2a8b/src/audiomixerboard.cpp#L67-L70) takes the same deque to clear it. The objects behind it are the two per-channel deques in [`MidiPickupState`](https://github.com/jamulussoftware/jamulus/blob/b4096fa6cbac37df3d391c25b43cecfa602e2a8b/src/audiomixerboard.cpp#L54-L60): [`SetFaderLevel` passes `recentFader`](https://github.com/jamulussoftware/jamulus/blob/b4096fa6cbac37df3d391c25b43cecfa602e2a8b/src/audiomixerboard.cpp#L1486-L1489), [`SetPanValue` passes `recentPan`](https://github.com/jamulussoftware/jamulus/blob/b4096fa6cbac37df3d391c25b43cecfa602e2a8b/src/audiomixerboard.cpp#L1525-L1528). The name is unchanged from `97799184`; this PR only adds the write-back under the comment.

Naming the referent:

```cpp
// still waiting: keep this value in the channel's recentFader/recentPan
// history so the next message can detect a crossing
```

Not pushed — the branch still carries the one-line form.

PR #3812, opened 2026-07-20
Check for half open interval in MathUtils::InRange

dingodoppelt:
Fix its usage in server.cpp

**Short description of changes**

As requested in #3810 make the interval half open for MathsUtils::InRange

CHANGELOG: Check for half open interval in MathUtils::InRange

**Context: Fixes an issue?**

**Does this change need documentation? What needs to be documented and how?**
No

**Status of this Pull Request**
Ready

**What is missing until this pull request can be merged?**
Approval

## Checklist

-  [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
-  [x] I tested my code and it does what I want
-  [x] My code follows the [style guide](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#source-code-consistency) 
-  [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
-  [x] I've filled all the content above

PR #3840, opened 2026-07-27
Backport #3810: bounds check iChanID before array access (release/3_12)

mcfnord:
**🤖 AI:** Backports #3810's `iChanID` bounds check to `release/3_12` as a piecemeal patch: one commit, two files, +13/-3. [`MathUtils::InRange`](https://github.com/mcfnord/jamulus/blob/014f1ef09f2785083a192abeaf334a6b6ff4bea2/src/util.h#L1229) is added to `util.h` with #3812's half-open interval, and [the `ApplyNewConClientList` call site](https://github.com/mcfnord/jamulus/blob/014f1ef09f2785083a192abeaf334a6b6ff4bea2/src/audiomixerboard.cpp#L1354) is switched to use it. Both regions are byte-identical to [`main`](https://github.com/jamulussoftware/jamulus/blob/4a43f6f6bd49a37b79817a64450346724cba5752/src/audiomixerboard.cpp#L1354) at `4a43f6f6`.

Not cherry-picked, [at pljones's request](https://github.com/jamulussoftware/jamulus/pull/3840#discussion_r3936157588): the commits that carried `InRange` to `main` also carry an unrelated server chat-handling refactor that `release/3_12` does not have, so only the amended lines are backported. Base is `release/3_12` at `e6a87df6`.

Fixes the same issue as #3810, for `release/3_12`.

---

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1349,9 +1349,12 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
 
         for ( size_t iFader = 0; iFader < iNumConnectedClients; iFader++ )
         {
-            // ideally "iChanID" in CChannelInfo would be size_t if it can never be INVALID_INDEX
-            // as assumed here
-            iFaderNumber[vecChanInfo[iFader].iChanID] = static_cast<int> ( iFader );
+            const int iChanID = vecChanInfo[iFader].iChanID;
+
+            if ( MathUtils::InRange<int> ( iChanID, 0, MAX_NUM_CHANNELS ) )
+            {
+                iFaderNumber[iChanID] = static_cast<int> ( iFader );
+            }
         }
 
         // Hide all unused faders and initialize used ones
--- a/src/util.h
+++ b/src/util.h
@@ -1223,6 +1223,13 @@ public:
             return powf ( 10.0f, ( fInValueRange0_1 - 1.0f ) * AUD_MIX_FADER_RANGE_DB / 20.0f );
         }
     }
+
+    // Returns true if value is in [lower, upper) (inclusive lower, exclusive upper).
+    template<typename T>
+    static inline bool InRange ( T value, T lower /* inclusive */, T upper /* exclusive */ )
+    {
+        return value >= lower && value < upper;
+    }
 };
 
 /******************************************************************************\
--- a/src/util.h
+++ b/src/util.h
@@ -1222,6 +1222,13 @@ public:
             return powf ( 10.0f, ( fInValueRange0_1 - 1.0f ) * AUD_MIX_FADER_RANGE_DB / 20.0f );
         }
     }
+
+    // Returns true if value is in [lower, upper) (inclusive lower, exclusive upper).
+    template<typename T>
+    static inline bool InRange ( T value, T lower /* inclusive */, T upper /* exclusive */ )
+    {
+        return value >= lower && value < upper;
+    }
 };
 
 /******************************************************************************\


src/audiomixerboard.cpp
```diff
@@ -1349,9 +1349,12 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
 
         for ( size_t iFader = 0; iFader < iNumConnectedClients; iFader++ )
         {
-            // ideally "iChanID" in CChannelInfo would be size_t if it can never be INVALID_INDEX
```
pljones, line 1352: These are meaningful comments to alert the reader to a quirk about the code.  The intent here and, hence  the reason for them has not changed.

It explains why the lower bounds check was unneeded.

ann0see (dismissed):
I don't like this inlining: we should ship the same code as on main.

mcfnord:
MY LLM WROTE:

@ann0see Fair — pushed ``57525a2b`` to ship the same code as main instead. Rather than inline the check, I backported just `MathUtils::InRange` (the template method itself, half-open-interval semantics from #3812) into `class MathUtils` in `util.h`. `release/3_12`'s `MathUtils` class already exists, so this is a single self-contained method addition — no dependency on the rest of #3741/#3812's chat refactor. `audiomixerboard.cpp`'s `ApplyNewConClientList` now calls `MathUtils::InRange<int>( iChanID, 0, MAX_NUM_CHANNELS )`, byte-for-byte identical to main's `562cfd84`.

Verified: `clang-format-14 --dry-run --Werror` clean, full client build (`qmake && make`) compiles and links with no errors/warnings.

pljones:
OK, that's small enough.

src/audiomixerboard.cpp
pljones: @mcfnord still applies now.

ann0see:
I still prefer the exact same commits cherry picked.

dingodoppelt (dismissed):
I'd prefer cherry-picking the original commits. We didn't squash on merge so there are a few which could be squashed only for this backport.

mcfnord:
MY LLM WROTE:

Force-pushed with actual cherry-picks instead of the hand-written equivalent.

`29a13e49` squashes just the `src/util.h` hunks from the six main commits that shaped `MathUtils::InRange` into its current form — #3741's `562d6c66` + `4f9a532a`, then #3812's `6d305cb7` + `65513e5b` + `fe98b1ad` — applied in that order. I left out their `server.cpp`/`server.h` hunks: those belong to #3741's `SendChatTextToConChannel`/`SendChatTextToAllConChannels` refactor, which `release/3_12` doesn't have and which is out of scope for a bounds-check backport. Squashing was necessary because several of those six commits are one-line follow-ups to each other (comment wording, style) — cherry-picking them individually would have produced intermediate states that don't compile standalone once the chat-refactor hunks are dropped.

`f0180425` is `git cherry-pick -x 562cfd84` unmodified — same commit, same author, `(cherry picked from commit ...)` trailer intact.

Net result is identical to the previous push, plus one line I'd missed: the `// Returns true if value is in [lower, upper)...` comment above `InRange`, which main has and my earlier hand-typed version didn't. Confirmed byte-for-byte against `MathUtils::InRange` on `main`.

Verified: `clang-format-14 --dry-run --Werror` clean, full client build (`qmake && make`) compiles and links `Jamulus` with no errors/warnings.

ann0see:
Those are still not the same commits as we have on main - due to different authorship. I think you should just git cherrypick the correct ones in the correct order.

pljones:
> I'd prefer cherry-picking the original commits. We didn't squash on merge so there are a few which could be squashed only for this backport.

I failed to find clean ones.

pljones:
> And again.
>
> Please do not set code that has not landed on main to a release branch.  Until it's clean and tested on main, it's not eligible for a backport > and that means it can't target a release branch.

(repeating myself unnecessarily...)

mcfnord:
MY LLM WROTE:

Thanks both. The branch already ships main's code: `MathUtils::InRange` in `util.h` is byte-for-byte identical to main, and the `audiomixerboard.cpp` bounds-check equals #3810's `562cfd84`.

The only open question is commit structure. A faithful `git cherry-pick -x 562d6c66` onto `release/3_12` applies cleanly — but it carries `SendChatTextToConChannel`/`SendChatTextToAllConChannels` and their `server.cpp`/`server.h` changes with it (`6d305cb7` does the same), landing a server chat refactor on a bug-fix branch. That's why this branch squashes only the `util.h` hunks instead.

So: **(a)** keep the `util.h`-only commit — identical to main, no server changes; or **(b)** cherry-pick the originals and accept the chat refactor on `release/3_12`. Happy either way — just flagging (b) isn't code-neutral here.

ann0see:
Not sure. It might be fine to have the chat refactor in.

ann0see:
@mcfnord I think you should rebase this and add the chat refactor.

dingodoppelt (approved):
This looks fine, compiles and works as expected.

src/server.cpp
pljones, line 1: Rebase error?

mcfnord: **🤖 AI:** Not a rebase error: the chat refactor is in on purpose. ann0see [asked for it](https://github.com/jamulussoftware/jamulus/pull/3840#issuecomment-5480871066), since `MathUtils::InRange` reached main inside that refactor (#3741) and got its half-open interval in #3812. The branch is eight `git cherry-pick -x` of the original commits onto `release/3_12` at `e6a87df6`, in main's order: #3741's `562d6c66`, `16b1ade0`, `4f9a532a`, `1e1f54ab`; #3812's `6d305cb7`, `65513e5b`, `fe98b1ad`; then #3810's `562cfd84`. Original authorship and trailers are kept, and all of it was merged on main by July.

One conflict, in [`6d305cb7`](https://github.com/jamulussoftware/jamulus/commit/6d305cb746d3a929c0707ad3cad8f7f8fee4bce5): its `server.cpp` hunk edits the `bool` form of `SendChatTextToConChannel` that #3731 gave the method after #3741. Resolved by keeping `562d6c66`'s `void` form with the half-open check. #3731 is not on `release/3_12`, so the method has no caller here, as on main between #3741 and #3731.

`MathUtils` in `util.h` is byte-identical to main, the `audiomixerboard.cpp` hunk is `562cfd84` unchanged, and the broadcast path emits the same calls as before. The PR description was still the July inline version; it is updated to match.

src/audiomixerboard.cpp
mcfnord: **🤖 AI:** Agreed the comment still holds: `iChanID` stays `int` because it can be `INVALID_INDEX`, and `InRange` now rejects that case explicitly. The removal is main's, though. #3810's `562cfd84` dropped both lines, after [the exchange on that PR](https://github.com/jamulussoftware/jamulus/pull/3810#discussion_r3616136514), and this hunk is that commit cherry-picked unchanged, so the release branch matches main here. Restoring the comment on `release/3_12` alone would put a line here that main does not have. I can open a one-line PR on main restoring it, for backport afterwards, if that is wanted.

src/server.cpp
pljones: No, I don't think we can be adding those two methods on the backport branch.  Only the `template<typename T> static inline bool InRange(..)` and `CAudioMixerBoard::ApplyNewConClientList` change should be here.

src/audiomixerboard.cpp
pljones: OK, keep it the same as main.  Shouldn't have gone there but no point re-adding it just for this.

pljones:
> @mcfnord I think you should rebase this and add the chat refactor.

Sorry, I missed this.  I disagree.  It's better to fix the bug with a clean and clear patch.  It's not religion.

src/server.cpp
ann0see: But that's harder to get only the template out - unless it was in separate commits?

pljones: It doesn't matter.

The code that needs backporting is clearly defined.

Cherry-picking is good if that's clean.  Here, it's bad because it brings in unwanted extra code.  So here we do not want to cherry-pick.

Instead it needs a piecemeal backport of the amended lines.

mcfnord:
**🤖 AI:** The one red check is a compiler crash, not the patch. The [Android job](https://github.com/jamulussoftware/jamulus/actions/runs/33988208324/job/101365597642) ends in a clang `Stack dump` while compiling `clientdlg.cpp` for `armeabi-v7a`; this PR touches only `audiomixerboard.cpp` and `util.h`. The same job passed on `release/3_12` itself in its last two runs on 2026-08-31. A rerun of that job is all that is needed.

ann0see (approved):

PR #3858, opened 2026-08-03
fix typo

dingodoppelt:
Fix a typo

CHANGELOG: SKIP

**Context: Fixes an issue?**

No

No

Ready to merge

## Checklist

### DIFF
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -1127,13 +1127,13 @@ void CConnectDlg::DeleteAllListViewItemChildren ( QTreeWidgetItem* pItem )
     while ( pItem->childCount() > 0 )
     {
         // get the first child in the list
-        QTreeWidgetItem* pCurChildItem = pItem->child ( 0 );
+        QTreeWidgetItem* pCurChild = pItem->child ( 0 );
 
         // remove it from the item (note that the object is not deleted)
-        pItem->removeChild ( pCurChildItem );
+        pItem->removeChild ( pCurChild );
 
         // delete the object to avoid a memory leak
-        delete pCurChildItem;
+        delete pCurChild;
     }
 }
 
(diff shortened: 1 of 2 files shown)


softins (approved):

src/connectdlg.cpp
```diff
@@ -581,7 +581,7 @@ void CConnectDlg::SetConnClientsList ( const CHostAddress& InetAddr, const CVect
     if ( pCurListViewItem )
     {
         // first remove any existing children
-        DeleteAllListViewItemChilds ( pCurListViewItem );
+        DeleteAllListViewItemChildren ( pCurListViewItem );
```
pljones, line 584: Given it's "...Item" I'd have gone for "...ChildItems".

dingodoppelt: I think the name was chosen to make clear a list item (as in `QTreeListItem`, although `QTreeWidgetItem` is subclassed here) can have children.

pljones: It runs a loop doing `delete pCurChildItem`.  So it's iterating over items.

dingodoppelt:
```
void CConnectDlg::DeleteAllListViewItemChilds ( QTreeWidgetItem* pItem )
{
    // loop over all children
    while ( pItem->childCount() > 0 )
    {
        // get the first child in the list
        QTreeWidgetItem* pCurChildItem = pItem->child ( 0 );

        // remove it from the item (note that the object is not deleted)
        pItem->removeChild ( pCurChildItem );

        // delete the object to avoid a memory leak
        delete pCurChildItem;
    }
}
```

Maybe `ChildItem` is the misnomer then. I still think the order should be Item->Children. I think the `item` needs to go (`pCurChild`) and that would make it clearer, because we already are an item and we iterate over our children.

dingodoppelt: Should be clearer with the latest commit

softins (approved):
Looks fine to me

pljones (approved):

pljones:
Yes, that makes more sense now.

PR #3859, opened 2026-08-03
Improve focus handling and accessibility in Connect dialog

### LINKED ISSUE #3857: Connect button in connect dialogue doesn't connect to the expected server
**Describe the bug**

In the connect dialogue we can connect to a server either by double clicking or single clicking and hitting the connect button.  
When there is a custom server address in the combo box the connect button will only connect to that server, it if was either changed or the combo box hasn't lost focus (the user didn't click anywhere).

**To Reproduce**

1. Open connect dialogue
2. Have a server address in the combo box
3. Single-Click on any server in the serverlist
4. Single-Click on the combo box and notice that it has focus again
5. Hit the connect button
6. Find yourself on the server you clicked in the serverlist instead of the one from the combo box

**Expected behavior**

Giving focus to the combo box should make the connect button connect to that combo box item as it corresponds to the last selected item that the user probably set on purpose.

**Operating system**
All

**Version of Jamulus**
All

softins:

Deselects any selected server in the server list when the Server Address combo box gains focus. This ensures that clicking
Connect when the Server Address has focus will connect to the expected server, even if a server had previously been
selected in the server list.

In addition, corrects the consuming of Tab and shift-Tab within the server list, improving keyboard navigation for accessibility.

CHANGELOG: Client: Improve server list navigation in Connect Dialog.

**Context: Fixes an issue?**

Fixes #3857

No

Ready to merge. Tested on Linux and Windows..

Nothing

## Checklist

### DIFF
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -279,8 +279,6 @@ CConnectDlg::CConnectDlg ( CClient* pNCliP, CClientSettings* pNSetP, const bool 
     QObject::connect ( edtFilter, &QLineEdit::textEdited, this, &CConnectDlg::OnFilterTextEdited );
 
     // combo boxes
-    QObject::connect ( cbxServerAddr, &QComboBox::editTextChanged, this, &CConnectDlg::OnServerAddrEditTextChanged );
-
     QObject::connect ( cbxDirectory, static_cast<void ( QComboBox::* ) ( int )> ( &QComboBox::activated ), this, &CConnectDlg::OnDirectoryChanged );
 
     // check boxes
@@ -662,13 +660,6 @@ void CConnectDlg::OnServerListItemDoubleClicked ( QTreeWidgetItem* Item, int )
     {
         OnConnectClicked();
     }
-}
-
-void CConnectDlg::OnServerAddrEditTextChanged ( const QString& )
-{
-    // in the server address combo box, a text was changed, remove selection
-    // in the server list (if any)
-    lvwServers->clearSelection();
 }
 
 void CConnectDlg::OnCustomDirectoriesChanged()
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -283,6 +283,9 @@ CConnectDlg::CConnectDlg ( CClient* pNCliP, CClientSettings* pNSetP, const bool 
     // combo boxes
     QObject::connect ( cbxDirectory, static_cast<void ( QComboBox::* ) ( int )> ( &QComboBox::activated ), this, &CConnectDlg::OnDirectoryChanged );
 
+    // connect when pressing Enter in the Server Address box
+    QObject::connect ( cbxServerAddr->lineEdit(), &QLineEdit::returnPressed, this, &CConnectDlg::OnConnectClicked );
+
     // check boxes
     QObject::connect ( chbExpandAll, &QCheckBox::stateChanged, this, &CConnectDlg::OnExpandAllStateChanged );
 
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -207,6 +207,7 @@ CConnectDlg::CConnectDlg ( CClient* pNCliP, CClientSettings* pNSetP, const bool 
 
     // install event filter to catch FocusIn
     cbxServerAddr->installEventFilter ( this );
+    lvwServers->installEventFilter ( this );
 
     // set up list view for connected clients (note that the last column size
     // must not be specified since this column takes all the remaining space)
@@ -334,6 +335,7 @@ void CConnectDlg::RequestServerList()
 
     // clear server list view
     lvwServers->clear();
+    savedServer = nullptr;
 
     // update list combo box (disable events to avoid a signal)
     cbxDirectory->blockSignals ( true );
@@ -1194,8 +1196,26 @@ bool CConnectDlg::eventFilter ( QObject* obj, QEvent* event )
 {
     if ( obj == cbxServerAddr && event->type() == QEvent::FocusIn )
     {
+        // check for a selected server before clearing the selection in the list
+        QList<QTreeWidgetItem*> CurSelListItemList = lvwServers->selectedItems();
+
+        if ( CurSelListItemList.count() > 0 )
+        {
+            // there was a selected item - save it before deselecting
+            savedServer = GetParentListViewItem ( CurSelListItemList[0] );
+        }
         // remove selection in the server list (if any)
         lvwServers->clearSelection();
+    }
+
+    if ( obj == lvwServers && event->type() == QEvent::FocusIn )
+    {
+        if ( savedServer )
+        {
+            // re-select any previously-selected server on regaining focus
+            lvwServers->setCurrentItem ( savedServer );
+            savedServer = nullptr;
+        }
     }
 
     return QDialog::eventFilter ( obj, event );
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -124,6 +124,7 @@ bool CMappedTreeWidgetItem::operator<( const QTreeWidgetItem& other ) const
 
 CConnectDlg::CConnectDlg ( CClient* pNCliP, CClientSettings* pNSetP, const bool bNewShowCompleteRegList, QWidget* parent ) :
     CBaseDlg ( parent, Qt::Dialog ),
+    savedServer ( nullptr ),
     pClient ( pNCliP ),
     pSettings ( pNSetP ),
     strSelectedAddress ( "" ),
@@ -334,8 +335,8 @@ void CConnectDlg::RequestServerList()
     strSelectedServerName = "";
 
     // clear server list view
-    lvwServers->clear();
     savedServer = nullptr;
+    lvwServers->clear();
 
     // update list combo box (disable events to avoid a signal)
     cbxDirectory->blockSignals ( true );
@@ -446,6 +447,7 @@ void CConnectDlg::SetServerList ( const CHostAddress& InetAddr, const CVector<CS
     }
 
     // first clear list
+    savedServer = nullptr;
     lvwServers->clear();
 
     // add list item for each server in the server list
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -205,6 +205,9 @@ CConnectDlg::CConnectDlg ( CClient* pNCliP, CClientSettings* pNSetP, const bool 
     cbxServerAddr->setMaxCount ( MAX_NUM_SERVER_ADDR_ITEMS );
     cbxServerAddr->setInsertPolicy ( QComboBox::NoInsert );
 
+    // install event filter to catch FocusIn
+    cbxServerAddr->installEventFilter ( this );
+
     // set up list view for connected clients (note that the last column size
     // must not be specified since this column takes all the remaining space)
 #ifdef ANDROID
@@ -1194,4 +1197,15 @@ void CConnectDlg::OnCurrentServerItemChanged ( QTreeWidgetItem* current, QTreeWi
     }
     QAccessible::updateAccessibility ( new QAccessibleAnnouncementEvent ( lvwServers, announcement ) );
 #endif
+}
+
+bool CConnectDlg::eventFilter ( QObject* obj, QEvent* event )
+{
+    if ( obj == cbxServerAddr && event->type() == QEvent::FocusIn )
+    {
+        // remove selection in the server list (if any)
+        lvwServers->clearSelection();
+    }
+
+    return QDialog::eventFilter ( obj, event );
 }
--- a/src/connectdlg.h
+++ b/src/connectdlg.h
@@ -127,7 +127,8 @@ protected:
     void                   EmitCLServerListPingMes ( const CHostAddress& haServerAddress, const bool bNeedVersion );
     void                   UpdateDirectoryComboBox();
 
-    bool eventFilter ( QObject* obj, QEvent* event );
+    bool                   eventFilter ( QObject* obj, QEvent* event ) override;
+    CMappedTreeWidgetItem* savedServer;
 
     CClient*         pClient;
     CClientSettings* pSettings;
--- a/src/connectdlg.h
+++ b/src/connectdlg.h
@@ -127,6 +127,8 @@ protected:
     void                   EmitCLServerListPingMes ( const CHostAddress& haServerAddress, const bool bNeedVersion );
     void                   UpdateDirectoryComboBox();
 
+    bool eventFilter ( QObject* obj, QEvent* event );
+
     CClient*         pClient;
     CClientSettings* pSettings;
 
--- a/src/connectdlg.h
+++ b/src/connectdlg.h
@@ -147,7 +147,6 @@ protected:
 
 public slots:
     void OnServerListItemDoubleClicked ( QTreeWidgetItem* Item, int );
-    void OnServerAddrEditTextChanged ( const QString& );
     void OnDirectoryChanged ( int iTypeIdx );
     void OnFilterTextEdited ( const QString& ) { UpdateListFilter(); }
     void OnExpandAllStateChanged ( int value ) { ShowAllMusicians ( value == Qt::Checked ); }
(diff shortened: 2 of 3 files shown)


src/connectdlg.cpp
```diff
@@ -1195,3 +1189,14 @@ void CConnectDlg::OnCurrentServerItemChanged ( QTreeWidgetItem* current, QTreeWi
     QAccessible::updateAccessibility ( new QAccessibleAnnouncementEvent ( lvwServers, announcement ) );
 #endif
 }
+
+bool CConnectDlg::eventFilter ( QObject* obj, QEvent* event )
+{
+    if ( obj == cbxServerAddr && event->type() == QEvent::FocusIn )
+    {
+        // remove selection in the server list (if any)
+        lvwServers->clearSelection();
```
pljones, line 1213: I don't actually like this.

The problem is that hitting connect when focus was in the Server text box _should_ use that value.

Moving focus from the server list to any other field should have no effect on the selected server -- at that point, you have no idea of the user intent.

Why should moving to one field (Server) behave differently from another (Filter)?

pljones (changes requested):

src/connectdlg.cpp
softins: > The problem is that hitting connect when focus was in the Server text box _should_ use that value.

Exactly. That's what @dingodoppelt was saying in #3857, and what does not always happen currently.

> Moving focus from the server list to any other field should have no effect on the selected server -- at that point, you have no idea of the user intent.

Except that if the field they have clicked on is the Server Address combo box, it's reasonable to assume that it contains the server they want to connect to.

> Why should moving to one field (Server) behave differently from another (Filter)?

Because by moving to the Server Address field, the user is desiring to use its value. The connect button _won't_ use that value if there is still a server selected in the list box.

Currently, the _only_ ways to deselect the selected server in the server list are either to go to a different directory, or to _edit or change_ the value in the Server Address field. Just selecting the Server Address field when it already contains the desired value from before does not currently allow its value to be used, if there is still a server selected in the list. This PR fixes that by interpreting putting focus on the Server Address field as intent that its value should be used when connecting, and therefore deselecting any server selected in the list.

If you need convincing, try this version out and see!

src/connectdlg.cpp
dingodoppelt, line 182: Referring to @OnjLouis [comment](https://github.com/jamulussoftware/jamulus/issues/3857#issuecomment-5168470329) we could simply introduce this here because I find the tab-key behaviour confusing as well.
```suggestion
    lvwServers->setTabKeyNavigation ( false );

```

src/connectdlg.cpp
pljones: I think it would be far better just to check focus correctly, rather than changing the natural behaviour of a widget (OK - whichever, it's changing Qt's implementation).

Being able to tab from the server list into the Server field would make more sense than this -- and that action shouldn't change the item selected.  If you carry on tabbing back to the server list, the same server should be selected.

The change proposed here would break that.  It's prioritising mouse navigation over keyboard navigation.

dingodoppelt: > I think it would be far better just to check focus correctly, rather than changing the natural behaviour of a widget (OK - whichever, it's changing Qt's implementation).
> 

That doesn't work because focus is on the button once pressed.

> Being able to tab from the server list into the Server field would make more sense than this -- and that action shouldn't change the item selected. If you carry on tabbing back to the server list, the same server should be selected.
> 

At the moment you can only tab out of the server list when the last server is selected. When tabbing backwards the same is true until you reach the first server in the list.  
Not the server address is saved, only if an item is selected with a quasi-boolean `lvwServers->selectedItems()`.

> The change proposed here would break that. It's prioritising mouse navigation over keyboard navigation.

This feature practically doesn't exist but I believe it should. My suggested change fixes all that at once as it looks. (edit: not quite, see comment below)

src/connectdlg.cpp
dingodoppelt: As a positive side effect, when you tab out of the server list the last selected server is saved, unlike the current implementation.

~Edit: When you tab back into the serverlist coming from the combo box, it still connects to the custom server address.~ Edit: probably fixed by 7b5600ed6ae956a72e32d7eb0476abc12f72e5a8 ?

src/connectdlg.cpp
softins: > I think it would be far better just to check focus correctly, rather than changing the natural behaviour of a widget (OK - whichever, it's changing Qt's implementation).

The problem is, by the time the Connect button click is being processed, I believe that button has focus, and you have lost the information about which widget had focus previously.

It's not "changing QT's implementation", it's making the perfectly reasonable UX design decision that the user, by choosing to select the Server Address box, is implying that _that_ is now what they are interested in, and are no longer interested in the server they had previously clicked on in the list, so it can be deselected. That is already the behaviour if the user _changes_ what is in the Server Address box. But they might not want to, if it already contains the desired server.

> Being able to tab from the server list into the Server field would make more sense than this -- and that action shouldn't change the item selected. If you carry on tabbing back to the server list, the same server should be selected.

The Tab behaviour is a separate issue that warrants investigation and its own PR. Currently Tab moves between servers in the list instead of moving from the server list to the combo box.

> The change proposed here would break that. It's prioritising mouse navigation over keyboard navigation.

No, the keyboard navigation is already broken. This PR doesn't change that, it just addresses and improves a specific shortcoming of the current UX design.

pljones:
OK, here's an alternative suggestion.

When the connect dialog opens or on change of current Directory through the drop down, set "default Server" to the Directory.  (Hm.  Not sure on this -- if there's anything in the Server text field, that should still take precedence at this point, I think.)
When a Server in the server list is selected, set "default Server" to the selected Server.
When focus lands on the Server text field and there's a value in it, set "default Server" to that value.

When "Connect" is clicked, connect to "default Server".  (Maybe that's not the best name for it, actually.)

(All of those are standard events from the widgets.)

softins:
> OK, here's an alternative suggestion.

Well, for what it's worth, _I_ don't actually like _this_ at all! Touché!

> When the connect dialog opens or on change of current Directory through the drop down, set "default Server" to the Directory. (Hm. Not sure on this -- if there's anything in the Server text field, that should still take precedence at this point, I think.) When a Server in the server list is selected, set "default Server" to the selected Server. When focus lands on the Server text field and there's a value in it, set "default Server" to that value.
> 
> When "Connect" is clicked, connect to "default Server". (Maybe that's not the best name for it, actually.)

You probably mean "selectedServer" or "chosenServer", but I don't think this is the right approach anyway. It's certainly not DRY. We already have the information in the widgets.

Certainly, setting to the directory would be wrong if there is a value in the server address field. Many times I open the connect dialog, and immediately click Connect because the server I want is already in the server address field.

But to me it's still logical that if there is a selected item in the server list, clicking Connect should use it, even if the list doesn't have focus. Therefore we need a way to unselect it if the user wants to use a custom server address. That will have been exactly why Volker made the list item be deselected when any change was made to the server address field. I still maintain that it is logical actually to do that as soon as the server Address field gains focus, as that action indicates _that server_ is the server of interest to the user, not the one that was selected in the list.

If your objection boils down to the fact that the selection in the list is forgotten when focussing on the server address field, this can be overcome by saving the identity of the selected item before clearing the selection. Then if and when the server list is again given focus, either by tabbing, or by clicking in an empty area of it (if it's not full), we see there is a saved `QTreeWidgetItem*` and make it the selection again.

I have added a commit which does exactly that, and have tested it. I really do hope that overcomes your objection.

pljones (approved):

pljones:
Somehow, yes, that works better for me.  I'm still not sure about what should happen on switching focus with tab.

I'd never really noticed the old behaviour -- I guess I'd not have a selected server list entry if I was entering text into the Server text box  (Would `editTextChanged` trigger when the selection from the Server dropdown changed?  Will `FocusIn` be triggered for that?)

softins:
> Somehow, yes, that works better for me. I'm still not sure about what should happen on switching focus with tab.

Well at the moment, pressing tab while in a list of servers just moves to the next server in the list, and shift-tab moves to the previous server. Down-arrow and up-arrow to the same.

I think only the arrow keys should move up and down, and tab should move to the next widget (which is the Server Address box), without changing the selection.

I'm not sure how to fix that, so it would need investigation and a separate PR.

> I'd never really noticed the old behaviour -- I guess I'd not have a selected server list entry if I was entering text into the Server text box

Yes, I think that's the case for most people, which would be why #3857 has only come to light now.

> (Would `editTextChanged` trigger when the selection from the Server dropdown changed? Will `FocusIn` be triggered for that?)

I just checked that. In all versions up till now, `editTextChanged` was triggered when either the text in the box was edited or a new entry was selected from the dropdown. The attached slot cleared the selection in the server list.

In this PR, `editTextChanged` is no longer needed, since `FocusIn` is triggered as soon as the cursor keys or a mouse click selects the widget, whether that be to edit the text or to click on the drop-down arrow.

dingodoppelt:
> I'm not sure how to fix that, so it would need investigation and a separate PR.

My suggested change fixes that.

softins:
> > I'm not sure how to fix that, so it would need investigation and a separate PR.
> 
> My suggested change fixes that.

I haven't seen your suggestion (can't find it on my phone), but it seems to be that `lvwServers` currently has `tabKeyNavigation` set to `true`, unusually, in the `.ui` file. I intend to try setting that back to the default `false`, but first I want to check history to see if there was a reason for changing it.

dingodoppelt:
> > > I'm not sure how to fix that, so it would need investigation and a separate PR.
> > 
> > 
> > My suggested change fixes that.
> 
> I haven't seen your suggestion (can't find it on my phone), but it seems to be that `lvwServers` currently has `tabKeyNavigation` set to `true`, unusually, in the `.ui` file. I intend to try setting that back to the default `false`, but first I want to check history to see if there was a reason for changing it.

Tab key navigation was introduced in d3b9f90f45013b7b5ec8bd1f07062d99eb9aa3ef for no apparent reason. Turning it off (in the .ui file this time) feels much more natural and along the lines of OnjLouis' comment. 
One other thing: When tabbing into the combo box hitting enter doesn't immediately connect. The text is highlighted first and enter deselects it and a second enter connects to the server. I don't know how to fix that though.

pljones:
> Tab key navigation was introduced in [d3b9f90](https://github.com/jamulussoftware/jamulus/commit/d3b9f90f45013b7b5ec8bd1f07062d99eb9aa3ef) for no apparent reason. 

The reason was so the list was keyboard navigable, I'd assume.  However, the next commit says it's a "fix" -- but doesn't remove the problem:

> [fix for issue with selection change in server list be up/down arrow keys](https://github.com/jamulussoftware/jamulus/commit/99432273c8a3107f0e730138a8bc3fcfe2c3a286)

> One other thing: When tabbing into the combo box hitting enter doesn't immediately connect. The text is highlighted first and enter deselects it and a second enter connects to the server. I don't know how to fix that though.

Tab onto the Connect button and hit enter?  Although I get the point: I don't understand why the first press of enter would _deselect_ the text at all...

dingodoppelt (approved):
Works as expected, thank you very much :)

softins:
I've added a commit to turn off tab-key navigation within the server list, and updated the PR description.

I had a conversation with ChatGPT about it, and its comments were:

>The combination of evidence is persuasive:
> 
> - it appeared in 2011 alongside an unrelated widget setting;
> - neither the commit nor later code explains why it was needed;
> - no accessibility work has depended on it;
> - disabling it preserves all the useful tree navigation;
> - it restores the expected Tab/Shift+Tab traversal;
> - and a blind user has independently described the existing behaviour as odd.
> 
> I would treat the original setting as most likely accidental, or at least as an obsolete choice whose downside is now clear.

Certainly in the version I built with this change:

- Up/Down still moves between servers.
- Home/End and Page Up/Page Down still behave sensibly.
- Enter/Return still connects to the selected server.
- Tab and Shift+Tab now move to the next/previous control in the dialog.

src/connectdlg.cpp
softins: > Referring to @OnjLouis [comment](https://github.com/jamulussoftware/jamulus/issues/3857#issuecomment-5168470329) we could simply introduce this here because I find the tab-key behaviour confusing as well.

This is better done by removing it from the `.ui` file, as I've done in 6638c8bf

softins: > As a positive side effect, when you tab out of the server list the last selected server is saved, unlike the current implementation.

To be precise, it's not when the server list loses focus, but when the server address combo gains focus, just before clearing the selection. Although at the moment, they are consecutive in tab order, so the effect is the same with keyboard navigation. Of course, with a mouse, you can focus anywhere at will.

> ~Edit: When you tab back into the serverlist coming from the combo box, it still connects to the custom server address.~ Edit: probably fixed by [7b5600e](https://github.com/jamulussoftware/jamulus/commit/7b5600ed6ae956a72e32d7eb0476abc12f72e5a8) ?

Yes, that's because 7b5600e restores any saved server in the list when it regains focus.

softins:
> One other thing: When tabbing into the combo box hitting enter doesn't immediately connect. The text is highlighted first and enter deselects it and a second enter connects to the server. I don't know how to fix that though.

That's odd. Just seen that comment, and yes, it does that for me too. I'll hold off on merging to investigate the reason for that first.

softins:
> > One other thing: When tabbing into the combo box hitting enter doesn't immediately connect. The text is highlighted first and enter deselects it and a second enter connects to the server. I don't know how to fix that though.
> 
> That's odd. Just seen that comment, and yes, it does that for me too. I'll hold off on merging to investigate the reason for that first.

I've added another signal-slot connection, suggested by ChatGPT, that appears to fix that. Please could you check again?

dingodoppelt (approved):
This works nicely and feels complete to me

pljones:
OK, there's something on main that's not on release/3_12 that's causing the cherry-pick some problems - and it doesn't patch cleanly either:
```
--- src/connectdlg.h
+++ src/connectdlg.h
@@ -127,6 +127,9 @@ protected:
     void                   EmitCLServerListPingMes ( const CHostAddress& haServerAddress, const bool bNeedVersion );
     void                   UpdateDirectoryComboBox();

+    bool                   eventFilter ( QObject* obj, QEvent* event ) override;
+    CMappedTreeWidgetItem* savedServer;
+
     CClient*         pClient;
     CClientSettings* pSettings;

```
```
--- src/connectdlg.cpp
+++ src/connectdlg.cpp
@@ -78,6 +78,7 @@ bool CMappedTreeWidgetItem::operator<( const QTreeWidgetItem& other ) const

 CConnectDlg::CConnectDlg ( CClient* pNCliP, CClientSettings* pNSetP, const bool bNewShowCompleteRegList, QWidget* parent ) :
     CBaseDlg ( parent, Qt::Dialog ),
+    savedServer ( nullptr ),
     pClient ( pNCliP ),
     pSettings ( pNSetP ),
     strSelectedAddress ( "" ),
```

These both fail because `pClient` is in the context of the commit being picked.
```
cffbe2313 (Tony Mountifield    2026-05-30 10:31:42 +0100 133)     CClient*         pClient;
```
```
cffbe2313 (Tony Mountifield    2026-05-30 10:31:42 +0100   82)     pClient ( pNCliP ),
```
```
cffbe231 2026-06-06 Auto-enable IPv6 by default, unless disabled [Tony Mountifield]
```
Hm.  Can I leave out `pClient` safely for this?  I'm assuming so.

---

A single squashed commit would have been easier here, too.

softins:
ok, I'd need to be at the computer to check, which I can do later. I'd be happy to make a PR on release/3_12 if that would help.

pljones:
> ok, I'd need to be at the computer to check, which I can do later. I'd be happy to make a PR on release/3_12 if that would help.

If you're okay to do the PR, that would help, yes.  Thanks.

I've been doing a branch off `release/3_12` called `r3_12-<original branch name>` locally (in my jamulussoftware/jamulus repo worktree), doing the change on that branch (ideally, just `git cherry-pick <commit>...`) and smoke testing, then doing `git merge --no-ff <the r3_12-...branch>` to `release/3_12`, before pushing to the release branch.

softins:
> Hm. Can I leave out `pClient` safely for this? I'm assuming so.

Yes, I think so. IIRC it was added to get access to the IPv6 availability status, which in 3.12 is just the enable flag.

Working on the PR now.

softins:
PR #3870 raised against `release/3_12`

pljones:
Backport merged to release/3_12.

PR #3862, opened 2026-08-06
Try to fix ASIO4ALL hang

### LINKED ISSUE #3779: Investigate Windows Hang if using ASIO4ALL especially

ann0see:

Draft PR narrowing lock and adding guard. Needs testing and human review. Authorship will be changed once I'm more comfortable.

CHANGELOG: Windows: Fix hang on ASIO4ALL

**Context: Fixes an issue?**

Fixes: #3779 

No

PoC

Review and Windows tests

## Checklist

### DIFF
--- a/src/sound/asio/sound.cpp
+++ b/src/sound/asio/sound.cpp
@@ -445,69 +445,74 @@ int CSound::GetActualBufferSize ( const int iDesiredBufferSizeMono )
 
 int CSound::Init ( const int iNewPrefMonoBufferSize )
 {
+    // get the actual sound card buffer size which is supported
+    // by the audio hardware
+    iASIOBufferSizeMono = GetActualBufferSize ( iNewPrefMonoBufferSize );
+
+    // init base class
+    CSoundBase::Init ( iASIOBufferSizeMono );
+
+    // set internal buffer size value and calculate stereo buffer size
+    iASIOBufferSizeStereo = 2 * iASIOBufferSizeMono;
+
+    // set the sample rate
+    ASIOSetSampleRate ( SYSTEM_SAMPLE_RATE_HZ );
+
+    // create memory for intermediate audio buffer. This reallocation is the
+    // only part which must not race with a running audio callback, so only it
+    // is guarded by the mutex. The opaque driver calls are deliberately kept
+    // outside the mutex: a driver may block inside them until its audio thread
+    // returns from bufferSwitch(), which would deadlock if the mutex were held
+    // here (see issue #3779)
     ASIOMutex.lock(); // get mutex lock
     {
-        // get the actual sound card buffer size which is supported
-        // by the audio hardware
-        iASIOBufferSizeMono = GetActualBufferSize ( iNewPrefMonoBufferSize );
-
-        // init base class
-        CSoundBase::Init ( iASIOBufferSizeMono );
-
-        // set internal buffer size value and calculate stereo buffer size
-        iASIOBufferSizeStereo = 2 * iASIOBufferSizeMono;
-
-        // set the sample rate
-        ASIOSetSampleRate ( SYSTEM_SAMPLE_RATE_HZ );
-
-        // create memory for intermediate audio buffer
         vecsMultChanAudioSndCrd.Init ( iASIOBufferSizeStereo );
-
-        // create and activate ASIO buffers (buffer size in samples),
-        // dispose old buffers (if any)
-        ASIODisposeBuffers();
-
-        // prepare input channels
-        for ( int i = 0; i < lNumInChan; i++ )
-        {
-            bufferInfos[i].isInput    = ASIOTrue;
-            bufferInfos[i].channelNum = i;
-            bufferInfos[i].buffers[0] = 0;
-            bufferInfos[i].buffers[1] = 0;
-        }
-
-        // prepare output channels
-        for ( int i = 0; i < lNumOutChan; i++ )
-        {
-            bufferInfos[lNumInChan + i].isInput    = ASIOFalse;
-            bufferInfos[lNumInChan + i].channelNum = i;
-            bufferInfos[lNumInChan + i].buffers[0] = 0;
-            bufferInfos[lNumInChan + i].buffers[1] = 0;
-        }
-
-        ASIOCreateBuffers ( bufferInfos, lNumInChan + lNumOutChan, iASIOBufferSizeMono, &asioCallbacks );
-
-        // query the latency of the driver
-        long lInputLatency  = 0;
-        long lOutputLatency = 0;
-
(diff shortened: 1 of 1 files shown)


ann0see:
There's something weird going on with the CI. Unsure why the CI does not run

src/sound/asio/sound.cpp
```diff
@@ -445,69 +445,74 @@ int CSound::GetActualBufferSize ( const int iDesiredBufferSizeMono )
...
+    ASIOSetSampleRate ( SYSTEM_SAMPLE_RATE_HZ );
 
-        // create memory for intermediate audio buffer
+    // create memory for intermediate audio buffer. This reallocation is the
+    // only part which must not race with a running audio callback, so only it
+    // is guarded by the mutex. The opaque driver calls are deliberately kept
+    // outside the mutex: a driver may block inside them until its audio thread
+    // returns from bufferSwitch(), which would deadlock if the mutex were held
+    // here (see issue #3779)
+    ASIOMutex.lock(); // get mutex lock
```
ann0see, line 467: Not sure if this exists, but a syntax like lock {.Init()} would be good. Otherwise also we could have a look at the .Init function to add the lock there - maybe.

src/sound/asio/sound.cpp
```diff
@@ -691,6 +696,14 @@ QStringList CSound::GetMIDIDevNames()
 
 void CSound::bufferSwitch ( long index, ASIOBool )
 {
+    // if we are not running, ignore any stray callback (e.g. from ASIO4ALL's
+    // emulation thread during Init()/teardown) so that it cannot block on
+    // ASIOMutex which is owned by the re-init thread (see issue #3779)
+    if ( !pSound->bRun )
```
ann0see, line 702: Even if this is not needed - I suspect it not being problematic to add

src/sound/asio/sound.cpp
softins: Instead of the separate `lock` and `unlock`, you could always do:
```
{
  QMutexLocker locker ( &ASIOMutex );
  vecsMultChanAudioSndCrd.Init ( iASIOBufferSizeStereo );
}
```
That would also unlock if `Init()` threw an exception.

ann0see:
Local reproduction shows that this is not yet fixed.

ann0see:
Closing as does not fix issue.

PR #3866, opened 2026-08-08
Comment accuracy: nine measured corrections (draft / experiment)

mcfnord:
MY LLM WROTE:

This draft PR is an experiment in format. Sixty-two "this must always hold"-style comments in the core files (buffer, channel, protocol, socket, server, client) are being tested one by one against the running code; nine came back misleading or outdated. Each commit here rewrites one comment to say what a fully accurate comment would say, and the commit message carries the measurement that justifies it  -  numbers, method, and what survived.

The diff is comments only. Every code line is byte-identical to `main` (two lines lost a trailing comment; the statements are unchanged). No behaviour is touched by this branch.

The nine split into two kinds:

- Six where the comment was wrong about correct code  -  the code is fine, the comment is now precise.
- Three where the comment was accurate about defective code. These are not softened: the comment is rewritten to describe the real, current behaviour, defect included, and marked FIXME. Fixing the code is separate work for another PR; the point here is that the comment should not keep asserting a guarantee the code does not provide.

What this deliberately is not:

- It is not a request to merge as-is. It is meant to stay in draft as a reviewable ledger. Any subset can be cherry-picked, reworded, or rejected commit-by-commit; each stands alone.
- It is not the whole audit. Most tested comments came back accurate, which is a result in itself; a further set of rows is still unrun and any that fail will be added as further commits.

The six comment-was-wrong corrections:

| file | old claim | measured |
|---|---|---|
| buffer.cpp | seq-number byte "much smaller" than audio keeps the block-count divide exact | the real bound is factor < blocksize, enforced by the props validator in protocol.cpp; the divide fails exactly at factor == blocksize (AddressSanitizer locates the overflow) |
| buffer.cpp | wraparound undetectable past 256 counts, ">100 ms" | horizon is 128 counts (signed fold), which is 171-341 ms at real frame rates |
| buffer.cpp | window moves are how valid packets get thrown away | a second loss channel (slot overwrite) exists, is the only one at buffer length 1, and the window move is load-bearing: without it dropouts go 0.012% to 6.704% |
| client.cpp | non-zero iActiveChannels => "server must have been restarted" | a >30 s traffic gap produces the same state from a never-restarted server (receive-timeout path) |
| server.cpp | flag "must be a member since QtConcurrent::run can only take 5 params" | the path uses the variadic CThreadPool::enqueue; the Qt5 limit is real (measured: 5 compiles, 6 fails) but no longer applies |
| server.cpp | "each thread can only set it to true and never to false" | OnTimer clears it to false every tick; ordering (clear before enqueue, read after the futures join), not one-way writes, is what makes it safe |

The three comment-accurate-but-code-defective (FIXME, code untouched):

| file | old comment | real behaviour now recorded |
|---|---|---|
| channel.cpp | 77-byte per-packet overhead (PPPoE + ATM DSL model) | measured 46 B on IPv4/Ethernet and 66 on IPv6; the returned upstream rate overstates real cost by ~15-50%, worst at the lowest bit-rate settings |
| server.cpp | "no memory must be allocated" in the realtime audio path | the send path still allocates once per packet (const-cast deep copy in CSocket::SendPacket): 73,308 allocations under OnTimer in 96 s with 3 clients |
| client.cpp | FindClientChannel here "should always return channel 0" | returns INVALID_INDEX (-1) for a server-sent channel id >= 150, unchecked; a UBSan build confirms a -1 dereference reached from an untrusted server |

Reproduce commands and full harness output exist for every row and can be posted as a gist if wanted.

### DIFF
--- a/src/buffer.cpp
+++ b/src/buffer.cpp
@@ -164,8 +164,16 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
             return false;
         }
 
-        // to get the number of input blocks we assume that the number of bytes for
-        // the sequence number is much smaller than the number of coded audio bytes
+        // This divide is exact if and only if iNumBlocks * iNumBytesSeqNum < iBlockSize,
+        // since the actual input is iNumBlocks * ( iBlockSize + iNumBytesSeqNum ) bytes. The
+        // sequence number merely being "much smaller" than the coded audio is not the
+        // condition: at iNumBlocks == iBlockSize the count comes out one too high whatever
+        // the ratio is. The bound is not enforced here but by the properties validator in
+        // protocol.cpp, EvaluateNetwTranspPropsMes, which rejects a base network packet size
+        // below CELT_MINIMUM_NUM_BYTES (10) and a block size factor outside
+        // { FRAME_SIZE_FACTOR_PREFERRED, _DEFAULT, _SAFE }. With iNumBytesSeqNum == 1 and
+        // iBlockSize == iBaseNetworkPacketSize - 1 (channel.cpp), the worst reachable case is
+        // a factor of 4 against a block size of 9.
         const int iNumBlocks = /* floor */ ( iInSize / iBlockSize );
 
         // copy new data in internal buffer
--- a/src/buffer.cpp
+++ b/src/buffer.cpp
@@ -198,9 +198,12 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
                 iSeqNumDiff -= 256;
             }
 
-            // The 1-byte sequence number wraps around at a count of 256. So, if a packet is delayed
-            // further than this we cannot detect it. But it does not matter since such a packet is
-            // more than 100 ms delayed so we have a bad network situation anyway. Therefore we
+            // The 1-byte sequence number is folded into a signed difference above, so a
+            // delayed packet is mistaken for an early one once it is more than 128 counts
+            // late, not 256. At the fastest possible frame rate that is still 171 ms
+            // (64-sample frames, 750 counts/s) and at the default frame size 341 ms
+            // (128-sample frames, 375 counts/s), so such a packet is long useless either
+            // way and we have a bad network situation anyway. Therefore we
             // assume that the sequence number difference between the received and local counter is
             // correct. The idea of the following code is that we always move our "buffer window" so
             // that the received packet fits into the buffer. By doing this we are robust against
--- a/src/buffer.cpp
+++ b/src/buffer.cpp
@@ -210,7 +210,11 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
             // sample rate offsets between client/server or buffer glitches in the audio driver since
             // we adjust the window. The downside is that we never throw away single packets which arrive
             // too late so we throw away valid packets when we move the "buffer window" to the delayed
-            // packet and then back to the correct place when the next normal packet is received. But
+            // packet and then back to the correct place when the next normal packet is received.
+            // Note that this is not the only way a valid block is lost: a block can also be
+            // overwritten in its slot before it is played out. That second channel is the only
+            // one that exists at a buffer length of 1, while the window move dominates from a
+            // buffer length of 3 upwards. But
             // tests showed that the new buffer strategy does not perform worse than the old jitter
             // buffer which did not use any sequence number at all.
             if ( iSeqNumDiff < 0 )
--- a/src/channel.cpp
+++ b/src/channel.cpp
@@ -716,12 +716,20 @@ int CChannel::GetUploadRateKbps()
 {
     const int iAudioSizeOut = iNetwFrameSizeFact * iAudioFrameSizeSamples;
 
-    // we assume that the UDP packet which is transported via IP has an
-    // additional header size of ("Network Music Performance (NMP) in narrow
-    // band networks; Carot, Kraemer, Schuller; 2006")
+    // The 77-byte per-packet overhead below (28 + 26 + 23) models a PPPoE-over-ATM DSL
+    // access path, following ("Network Music Performance (NMP) in narrow band networks;
+    // Carot, Kraemer, Schuller; 2006"):
     // 8 (UDP) + 20 (IP without optional fields) = 28 bytes
     // 2 (PPP) + 6 (PPPoE) + 18 (MAC)            = 26 bytes
     // 5 (RFC1483B) + 8 (AAL) + 10 (ATM)         = 23 bytes
+    // A packet capture on a real internet path measures 46 bytes over IPv4/Ethernet
+    // (20 IP + 8 UDP + 14 Ethernet, no PPPoE, no ATM, no VLAN) and 66 over IPv6, so this
+    // constant is 31 bytes too large for IPv4 and 20 bytes too small for the IPv6 header
+    // it never accounts for. The figure a client can actually justify is 28 (IPv4) or
+    // 48 (IPv6) at L3/L4, since the access encapsulation is invisible to the endpoint.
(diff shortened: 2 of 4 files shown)


src/buffer.cpp
```diff
@@ -164,8 +164,16 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
...
+        // This divide is exact if and only if iNumBlocks * iNumBytesSeqNum < iBlockSize,
+        // since the actual input is iNumBlocks * ( iBlockSize + iNumBytesSeqNum ) bytes. The
+        // sequence number merely being "much smaller" than the coded audio is not the
+        // condition: at iNumBlocks == iBlockSize the count comes out one too high whatever
+        // the ratio is. The bound is not enforced here but by the properties validator in
+        // protocol.cpp, EvaluateNetwTranspPropsMes, which rejects a base network packet size
+        // below CELT_MINIMUM_NUM_BYTES (10) and a block size factor outside
+        // { FRAME_SIZE_FACTOR_PREFERRED, _DEFAULT, _SAFE }. With iNumBytesSeqNum == 1 and
+        // iBlockSize == iBaseNetworkPacketSize - 1 (channel.cpp), the worst reachable case is
+        // a factor of 4 against a block size of 9.
```
pljones, line 176: The above is far too verbose for anyone to read it clearly.  Two lines looks like a good length.  Ten lines isn't.

It needs to be short and to the point:
> **Intent:**
> Purpose of this section, including what the result is used for.
>
> **Inputs:**
> * `iInSize`: where this comes from and what it means
> * `iBlockSize`: ...
> * ...
>
> **Method:**
> Explanation of _why_ this method works.  But only if it's not transparently obvious.

src/buffer.cpp
```diff
@@ -198,9 +198,12 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
...
 
-            // The 1-byte sequence number wraps around at a count of 256. So, if a packet is delayed
-            // further than this we cannot detect it. But it does not matter since such a packet is
-            // more than 100 ms delayed so we have a bad network situation anyway. Therefore we
+            // The 1-byte sequence number is folded into a signed difference above, so a
+            // delayed packet is mistaken for an early one once it is more than 128 counts
+            // late, not 256. At the fastest possible frame rate that is still 171 ms
+            // (64-sample frames, 750 counts/s) and at the default frame size 341 ms
+            // (128-sample frames, 375 counts/s), so such a packet is long useless either
+            // way and we have a bad network situation anyway. Therefore we
```
pljones, line 206: This doesn't add anything at all.

src/buffer.cpp
```diff
@@ -210,7 +210,11 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
             // sample rate offsets between client/server or buffer glitches in the audio driver since
             // we adjust the window. The downside is that we never throw away single packets which arrive
             // too late so we throw away valid packets when we move the "buffer window" to the delayed
-            // packet and then back to the correct place when the next normal packet is received. But
+            // packet and then back to the correct place when the next normal packet is received.
+            // Note that this is not the only way a valid block is lost: a block can also be
```
pljones, line 214: This addition is not in context and doesn't belong here.

src/client.cpp
```diff
@@ -999,9 +999,12 @@ void CClient::OnControllerInMuteMyself ( bool bMute )
...
 {
-    // if we have just connected to a running server, iActiveChannels will be 0
-    // if iActiveChannels is not 0, the server must have been restarted on the fly
-    // in that case, channels might have changed, so clear our list to get it afresh.
+    // If we have just connected to a running server, iActiveChannels will be 0.
+    // If it is not 0, the server has begun a NEW connection for us while this client kept
+    // running. A restart is only one way that happens: the server also drops a channel whose
+    // receive timeout expires ( CON_TIME_OUT_SEC_MAX, channel.h ) and treats the next packet
+    // from the same peer as a new connection, so a traffic gap of longer than that is enough.
+    // Either way the channel list we hold may be stale, so clear it and get it afresh.
```
pljones, line 1007: Too much waffle.  The original got the point across succinctly.

src/server.cpp
```diff
@@ -662,7 +662,11 @@ void CServer::OnTimer()
     bool bUseMT               = false;
     int  iNumBlocks           = 0;     // init number of blocks for multithreading
     int  iMTBlockSize         = 0;     // init block size for multithreading
-    bChannelIsNowDisconnected = false; // note that the flag must be a member function since QtConcurrent::run can only take 5 params
+    // The flag is a member variable, not a local. Nothing in the threading forces that: the
```
pljones, line 665: Too much waffle again.  Say what is and why.  Not what is not and why not.

src/server.cpp
```diff
@@ -948,9 +948,14 @@ void CServer::DecodeReceiveData ( const int iChanCnt, const int iNumClients )
 
                 FreeChannel ( iCurChanID ); // note that the channel is now not in use
 
-                // note that no mutex is needed for this shared resource since it is a
-                // std::atomic write (not a read-modify-write operation) and also each
-                // thread can only set it to true and never to false
+                // Note that no mutex is needed for this shared resource: the store is a
+                // std::atomic write, not a read-modify-write operation. It is NOT true that
```
pljones, line 952: Again, saying what is not, is not helpful, just token-spend.

src/channel.cpp
```diff
@@ -716,12 +716,20 @@ int CChannel::GetUploadRateKbps()
...
-    // we assume that the UDP packet which is transported via IP has an
-    // additional header size of ("Network Music Performance (NMP) in narrow
-    // band networks; Carot, Kraemer, Schuller; 2006")
+    // The 77-byte per-packet overhead below (28 + 26 + 23) models a PPPoE-over-ATM DSL
+    // access path, following ("Network Music Performance (NMP) in narrow band networks;
+    // Carot, Kraemer, Schuller; 2006"):
     // 8 (UDP) + 20 (IP without optional fields) = 28 bytes
     // 2 (PPP) + 6 (PPPoE) + 18 (MAC)            = 26 bytes
     // 5 (RFC1483B) + 8 (AAL) + 10 (ATM)         = 23 bytes
+    // A packet capture on a real internet path measures 46 bytes over IPv4/Ethernet
```
pljones, line 725: Experimental data should not inform code comments.

src/server.cpp
```diff
@@ -165,10 +165,17 @@ CServer::CServer ( const int          iNewMaxNumChan,
         iServerFrameSizeSamples = SYSTEM_FRAME_SIZE_SAMPLES;
     }
 
-    // To avoid audio clitches, in the entire realtime timer audio processing
-    // routine including the ProcessData no memory must be allocated. Since we
-    // do not know the required sizes for the vectors, we allocate memory for
-    // the worst case here:
+    // To avoid audio glitches, the realtime timer audio routine ( OnTimer ->
+    // ProcessData ) should allocate no memory. The worst-case vectors below are
+    // pre-sized here for that reason. Note that the goal is not currently met: the
```
pljones, line 170: Again, do not say what is not, only what is.

If there is something that needs fixing, raise an issue or PR.

src/client.cpp
```diff
@@ -1012,7 +1012,14 @@ void CClient::OnClientIDReceived ( int iServerChanID )
     }
 
     // allocate and map client-side channel 0
-    int iChanID = FindClientChannel ( iServerChanID, true ); // should always return channel 0
+    // In normal operation this returns channel 0. It is NOT guaranteed: FindClientChannel
```
pljones, line 1015: More negatives.

mcfnord:
**🤖 AI**

Superseded by [#3877](https://github.com/jamulussoftware/jamulus/pull/3877): the same corrections, redone per the review feedback here as the smallest wording change per site, with the measurements in the commit messages instead of the comments.

---

PR #3867, opened 2026-08-08
Avoid windows ASIO4ALL hang by using QueuedConnection

ann0see:

I believe this fixes the hang on ASIO4ALL device switches - I cannot reproduce the hang anymore.

The AI proposed guard and the narrowed lock (#3862) doesn't seem to be needed - but might be worth implementing as lower prio fix.

CHANGELOG: Bug: Fix hang on Windows while changing devices in ASIO4ALL

**Context: Fixes an issue?**

Found by AI debugging:
https://github.com/jamulussoftware/jamulus/issues/3779#issuecomment-5209182261

No

Ready for testing and review by someone else.

Review and 2nd testing

## Checklist

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -182,7 +182,15 @@ CClient::CClient ( const quint16  iPortNumber,
     QObject::connect ( &ConnLessProtocol, &CProtocol::CLChannelLevelListReceived, this, &CClient::OnCLChannelLevelListReceived );
 
     // other
-    QObject::connect ( &Sound, &CSound::ReinitRequest, this, &CClient::OnSndCrdReinitRequest );
+    QObject::connect ( &Sound,
+                       &CSound::ReinitRequest,
+                       this,
+                       &CClient::OnSndCrdReinitRequest
+#if defined( Q_OS_WINDOWS )
+                       ,
+                       Qt::QueuedConnection // On Windows, use a queued connection to avoid an ASIO4ALL hang (#3867)
+#endif
+    );
 
     QObject::connect ( &Sound, &CSound::ControllerInFaderLevel, this, &CClient::OnControllerInFaderLevel );
 


ann0see:
I remember that we had the same kind of bug on iOS with the language switch a few years ago.

ann0see:
> ⚠️ AI-assisted summary of a human test — verify before citing.

## Independent Windows/ASIO4ALL test — confirms the fix

Reproduced the hang and verified this change on a separate Windows build with
ASIO4ALL.

**Before (no `Qt::QueuedConnection`):** GUI hang when re-initialising the ASIO
driver (change device/channels while running). A symbolized cdb dump of the hung
process shows the ASIO4ALL callback thread running `asioMessages()` →
`emit ReinitRequest` → `OnSndCrdReinitRequest` *synchronously* on the driver's
thread, then blocking in `QBasicMutex::lockInternal` on `MutexDriverReinit` —
which the main thread already holds inside `Sound.Stop()/Init()`. A
`MutexDriverReinit` deadlock, **not** `ASIOMutex`, which is why the #3862
lock-narrowing alone didn't help.

**After (this PR):** same workflow, no hang; only the expected brief
stop/start of the audio engine during re-init. `Qt::QueuedConnection` makes the
slot run only on the main-thread event loop, so it cannot re-enter while
`MutexDriverReinit` is held.

softins:
> I believe this fixes the hang on ASIO4ALL device switches - I cannot reproduce the hang anymore.

I'd like to verify this on my PC with ASIO4ALL. What are the steps I should use to demonstrate the hang on existing versions?

EDIT: ok, now seen the minimised AI comment above. I'll try changing devices within ASIO4ALL itself.

ann0see:
You need to randomly enable/disable inputs/outputs in the ASIO4ALL gui and then sometimes click on the main window which will lead to a hang. It's not really reproducible but will come up quite quickly.

src/client.cpp
```diff
@@ -182,7 +182,7 @@ CClient::CClient ( const quint16  iPortNumber,
     QObject::connect ( &ConnLessProtocol, &CProtocol::CLChannelLevelListReceived, this, &CClient::OnCLChannelLevelListReceived );
 
     // other
-    QObject::connect ( &Sound, &CSound::ReinitRequest, this, &CClient::OnSndCrdReinitRequest );
+    QObject::connect ( &Sound, &CSound::ReinitRequest, this, &CClient::OnSndCrdReinitRequest, Qt::QueuedConnection );
```
pljones: The soundcard <-> client <-> network exchange shouldn't _really_ require a queued signal/slot - we don't want it to have any possibility of blocking or racing.

If there's a problem, it _should_ be elsewhere.  So I'm a bit suspicious of this as a fix.

pljones:
My gut feel is this is because of Qt insisting on the GUI being the "main" thread.

I guess the "proper" fix, from that perspective, is to start up CClient (and thus everything CClient looks after - CSound, CChannel, network bits, etc) on the real-time thread, leaving the "main" thread at normal priority.

That should mean the only cross-thread calls are then between GUI and CClient...

However, I guess that without substantial work, this might be the only way to fix it. Maybe it should have a guard so it's only Windows ASIO that gets it - unless there's reason to say it's needed for all platforms.

pljones:
Also note, could you check if the cherry pick on any commit is clean before merging and, if not, see why and say whether you'll deal with it.

ann0see:
I would expect it to be a clean cherry pick. Yes, we could probably fix it by refactoring correctly but definitely not part of this Pr.

src/client.cpp
pljones: Maybe it should have a guard so it's only Windows ASIO that gets it - unless there's reason to say it's needed for all platforms?

ann0see: Sure, we can ifdef this. I think iOS does the same.

pljones (approved):

pljones:
Merging.  As there's nothing else tagged up pending backport, I'll cut the branch release.

PR #3872, opened 2026-08-08
Ensure a server is always selected when list gains focus

### LINKED ISSUE #3871: Edge case in #3859 with keyboard navigation
**Describe the bug**

In the improved connect dialog handling of #3859, there is an edge case with keyboard navigation.
If the server list from the directory has only one server, it is not possible to select it with the keyboard.
When using Tab to move into the server list, no server is initially selected. Pressing cursor Down or Up
moves to the next server in the list and selects it. However, if there is only one server in the list, the cursor
keys are not accepted, and it is impossible to select that single server without using the mouse.

**To Reproduce**

Open the connect dialog and select a custom directory which only contains a single server.
One such directory is `newjam.softins.co.uk`. Using only keyboard, Tab into the server list. The server
is not selected. Press Down or Up. The server remains unselected.

**Expected behavior**

Being able to select the single server. So that when Tabbing into the list, if there is no saved server from
being previously in the list, it should select the first item by default.

**Version of Jamulus**

`3.12.3dev-a24e09ce`, after merge of #3859. Also present in backport #3870 currently, not yet merged.

**Additional context**

Fix is ready, will raise PR.

softins:

This fixes keyboard navigation when the list only contains one server. When the server list gains focus,
if there was a saved server from previously being in focus, that saved server is restored. In this PR,
if there was no saved server, the first item in the list will be selected, so that keyboard navigation
will work properly.

No changelog entry needed, as this is just a bugfix for #3859

CHANGELOG: SKIP

**Context: Fixes an issue?**

Fixes #3871

No

Ready

Nothing

## Checklist

### DIFF
--- a/src/connectdlg.cpp
+++ b/src/connectdlg.cpp
@@ -1175,6 +1175,11 @@ bool CConnectDlg::eventFilter ( QObject* obj, QEvent* event )
             lvwServers->setCurrentItem ( savedServer );
             savedServer = nullptr;
         }
+        else if ( lvwServers->topLevelItemCount() > 0 )
+        {
+            // no saved server: select first item in list for keyboard navigation
+            lvwServers->setCurrentItem ( lvwServers->topLevelItem ( 0 ) );
+        }
     }
 
     return QDialog::eventFilter ( obj, event );


dingodoppelt:
Very nice, haven't come across that bug myself, yet, but the fix sounds very sensible.

dingodoppelt (approved):
I just tested it and it works nicely :) Thank you!

src/connectdlg.cpp
```diff
@@ -1175,6 +1175,11 @@ bool CConnectDlg::eventFilter ( QObject* obj, QEvent* event )
             lvwServers->setCurrentItem ( savedServer );
             savedServer = nullptr;
         }
+        else if ( lvwServers->topLevelItemCount() > 0 )
+        {
+            // no saved server: select first item in list for keyboard navigation
```
pljones, line 1180: Shouldn't this select the Directory by default?  The default in Jamulus has been that hitting enter with nothing selected connects to the Directory.

pljones: (Although, if the filter is in used, the Directory might not be visible -- still the default, even in that case, would have been for enter to join the Directory.)

softins: That is very old behaviour. If there is _anything_ in the Server Address combo box, then Enter will connect to _that_ server.

In order to test your assertion, I needed to clear my whole list of custom servers to make the combo box empty.

Having done that, I selected a directory (e.g. `private.jamulus.io` or one of my own), and without clicking on a server in the list I pressed Enter. To my surprise, it connected to the **Any Genre 1** directory server! That was the case whichever directory I had selected. I think that is the behaviour you are remembering, and in fact I would argue that behaviour is now obsolete.

So I don't think there is much utility in making the directory the one that is selected, as it might be some way down the list, if its ping is a lot greater than some of its registered servers.

This PR is just to make keyboard navigation work consistently when tabbing into the server list, and the first item in the list is the natural one to start with if you are going to cursor down the list.

dingodoppelt: > To my surprise, it connected to the **Any Genre 1** directory server! That was the case whichever directory I had selected. I think that is the behaviour you are remembering, and in fact I would argue that behaviour is now obsolete.

That seems very weird. For any user not living near Frankfurt that server seems like a bad choice.

softins: That behaviour dates from b63a625ffa in 2015, and has been functionally unchanged since.

In `CClientDlg::OnConnectDlgAccepted()`:
https://github.com/jamulussoftware/jamulus/blob/a24e09ce13f654359238f6472c3756adf593b518/src/clientdlg.cpp#L760-L768

I think it's time we didn't do that, but instead either ignored the `Connect` button if a server isn't selected, or displayed a message box. But that's for a separate PR.

pljones: Oh... I didn't realise it was _still_ `DEFAULT_SERVER_ADDRESS` -- I thought when the custom Directories got added it got updated to the currently selected Directory, as a fallback in cases where nothing was selected.

Anyway, I agree, given the current version isn't adequate, anything is probably an improvement.

pljones (approved):

Issue #3869, opened 2026-08-08
Harden ASIO re-init: ignore stray bufferSwitch() when not running; narrow ASIOMutex scope


ann0see:
## Summary

**Possible hardening: make `CSound::bufferSwitch()` ignore stray callbacks when
not running, and narrow the `ASIOMutex` scope in `CSound::Init()`.** This is a
proposed improvement only — it is **not** the fix for the #3779 hang (that is
PR #3867, a `Qt::QueuedConnection` change). It is a separate, optional
measure for a distinct `ASIOMutex` race that #3867 does not touch.

## Background

- #3779 (Windows hang with ASIO4ALL) turned out to be a `MutexDriverReinit`
  deadlock, fixed by #3867. There is also an `ASIOMutex`-related race which
  this issue covers.
- A similar narrow-lock approach was tried in #3862 and closed as insufficient
  on its own; the current proposal is a *secondary* measure, not a replacement.

### 1. `CSound::bufferSwitch()` — early-return when not running

src/sound/asio/sound.cpp:

```diff
 void CSound::bufferSwitch ( long index, ASIOBool )
 {
+    // if not running, ignore stray callbacks (e.g. from ASIO4ALL's thread
+    // during Init()/teardown) so they cannot block on ASIOMutex
+    if ( !pSound->bRun )
+    {
+        return;
+    }
+
     int iCurSample;
```

### 2. `CSound::Init()` — narrow the `ASIOMutex` scope

Keep the lock only around the `vecsMultChanAudioSndCrd.Init()` reallocation and
move the opaque driver calls (`ASIOSetSampleRate`, `ASIODisposeBuffers`,
`ASIOCreateBuffers`, `ASIOGetLatencies`, `ASIOOutputReady`) outside it:

```diff
 int CSound::Init ( const int iNewPrefMonoBufferSize )
 {
-    ASIOMutex.lock(); // get mutex lock
-    {
-        // get the actual sound card buffer size which is supported
-        // by the audio hardware
-        iASIOBufferSizeMono = GetActualBufferSize ( iNewPrefMonoBufferSize );
+    // get the actual sound card buffer size which is supported
+    // by the audio hardware
+    iASIOBufferSizeMono = GetActualBufferSize ( iNewPrefMonoBufferSize );
 
-        // init base class
-        CSoundBase::Init ( iASIOBufferSizeMono );
+    // init base class
+    CSoundBase::Init ( iASIOBufferSizeMono );
 
[...]

PR #3873, opened 2026-08-08
src/sound/README.md: start documenting the sound design

mcfnord:
MY LLM WROTE:

`src/sound/README.md` ends with *"Fixme: The sound design is not yet documented."* This makes a start on that section.

It describes how the code behaves today, not why it was designed that way — intent is not something a reader of the source can verify, so none is asserted. It is also not the whole design, so the blanket Fixme is replaced by a list of the areas still missing rather than dropped.

What the new section covers:

- **Buffer size negotiation** — `Init()` returns the size the device actually accepted, which is how `CClient::Init()` discovers which sizes are supported, and why it calls `Init()` four times per invocation. Also that the three `bFraSiFact*Supported` flags drive the buffer delay radio buttons, and that the settings dialog polls them once a second rather than being signalled.
- **Driver-initiated size changes** — `kAsioBufferSizeChange` and JACK's buffer size callback both arrive through `EmitReinitRequestSignal ( RS_ONLY_RESTART_AND_INIT )`.
- **Start, stop and the audio callback** — that `Init()` is only ever entered with the device stopped, and a table of which backend ignores its audio callback while stopped and which takes `MutexAudioProcessCallback`. ASIO is the exception to both, which is why `CSoundBase::Stop()`'s wait for a callback in flight does nothing on Windows and `asio/CSound::Stop()` waits on `ASIOMutex` instead.
- **Not yet documented** — device enumeration and `SetDev()` failure handling, channel selection and mixing, MIDI, `GetInOutLatencyMs()`, and the sound card conversion buffer.

CHANGELOG: SKIP

**Context: Fixes an issue?**

No issue. The material comes out of #3869, where the buffer size display question ([issuecomment-5227331894](https://github.com/jamulussoftware/jamulus/issues/3869#issuecomment-5227331894)) turned into the lifecycle and callback rules written down here.

This is the documentation. Nothing is needed on the website: it is developer-facing detail about one source folder, which is what `src/sound/README.md` already exists to hold.

Working implementation. Every statement in it is checkable against the tree at the commit it was written on, and the table was read off all five backends rather than assumed.

Review, and a decision on scope: whether a partial section plus an explicit "not yet documented" list is the right shape for that Fixme, or whether it should stay a placeholder until the whole design is covered.

## Checklist

No checks run on this one: `autobuild.yml` carries `paths-ignore: '**README.md'` and `coding-style-check.yml` only triggers on `**.cpp`/`**.h`, so the fourth box stays unticked rather than claiming a green run that never happened.

### DIFF
--- a/src/sound/README.md
+++ b/src/sound/README.md
@@ -48,4 +48,54 @@ This folder contains the related files for all sound APIs.
 
 ## Documentation of sound design
 
-**Fixme:** The sound design is not yet documented.
+This describes how the code behaves today. It covers the device lifecycle and the threading rules
+around the audio callback; the areas still missing are listed at the end.
+
+Each platform provides one `CSound` class deriving from `CSoundBase`, and exactly one of the
+subdirectories is compiled in: `asio/` (Windows), `coreaudio-mac/`, `coreaudio-ios/`, `oboe/`
+(Android) and `jack/` (where JACK is enabled).
+
+### Buffer size negotiation
+
+`Init ( iNewPrefMonoBufferSize )` returns the mono buffer size the device actually accepted,
+which may differ from the one requested. `CClient::Init()` uses that return value to find out
+which sizes a device supports, so it calls `Init()` four times per invocation: once for each of
+`FRAME_SIZE_FACTOR_PREFERRED`, `FRAME_SIZE_FACTOR_DEFAULT` and `FRAME_SIZE_FACTOR_SAFE` to fill
+`bFraSiFactPrefSupported`, `bFraSiFactDefSupported` and `bFraSiFactSafeSupported`, then once with
+the size selected in the settings. Those three flags drive the enabled state of the buffer delay
+radio buttons, and the settings dialog polls them once a second; no signal runs from the sound
+device to that dialog.
+
+A driver may also change the buffer size on its own. `kAsioBufferSizeChange` in the ASIO backend
+and JACK's buffer size callback both report that by calling
+`EmitReinitRequestSignal ( RS_ONLY_RESTART_AND_INIT )`, which reaches
+`CClient::OnSndCrdReinitRequest` and repeats the negotiation above.
+
+### Start, stop and the audio callback
+
+`Init()` is only ever entered with the device stopped. Callers that may be running stop it first
+and restart it afterwards, which is the `bWasRunning` pattern throughout `client.cpp`.
+
+The audio callback runs on a thread owned by the driver. Backends keep it away from a device that
+is being re-initialised in two ways, and the ASIO backend is the exception to both:
+
+| backend | audio callback | ignores the callback while stopped | takes `MutexAudioProcessCallback` |
+|---|---|---|---|
+| JACK | `process()` | yes, `IsRunning()` | yes |
+| CoreAudio (macOS) | `callbackIO()` | yes, `bRun` | yes |
+| CoreAudio (iOS) | `processBufferList()` | no | yes |
+| Oboe | `onAudioReady()` | yes, `!bRun` | yes |
+| ASIO | `bufferSwitch()` | no | no, it uses its own `ASIOMutex` |
+
+`CSoundBase::Stop()` clears `bRun` and then takes `MutexAudioProcessCallback` to wait for a
+callback that is already in flight. The ASIO backend never takes that mutex, so on Windows that
+wait returns immediately and `CSound::Stop()` waits on `ASIOMutex` instead.
+
+### Not yet documented
+
+- device enumeration, and what `SetDev()` does when a device cannot be used
+- input and output channel selection, and the input channel mixing in the callbacks
+- MIDI: device selection, controller mapping and `ParseMIDIMessage()`
+- latency reporting via `GetInOutLatencyMs()`
+- the sound card conversion buffer used when a device's buffer size is not a multiple of the
+  system frame size
--- a/src/sound/README.md
+++ b/src/sound/README.md
@@ -111,11 +111,11 @@ Inside the callback itself, backends differ again, and not just in which flag th
 
 | backend | audio callback | flag check | mutex behavior |
 |---|---|---|---|
-| Oboe | `onAudioReady()` | `!bRun`, first line, before any lock | skipped entirely once stopped |
-| JACK | `process()` | `IsRunning()`, after the lock | always taken; only the processing is skipped |
+| ASIO | `bufferSwitch()` | none | always taken (its own `ASIOMutex`); always processes |
 | CoreAudio (macOS) | `callbackIO()` | `bRun`, after the lock | always taken; only the processing is skipped |
 | CoreAudio (iOS) | `processBufferList()` | none | always taken; always processes |
-| ASIO | `bufferSwitch()` | none | always taken (its own `ASIOMutex`); always processes |
+| Oboe | `onAudioReady()` | `!bRun`, first line, before any lock | skipped entirely once stopped |
+| JACK | `process()` | `IsRunning()`, after the lock | always taken; only the processing is skipped |
 
 `IsRunning()`, `bRun` and `!bRun` all read the same flag (`IsRunning()` is `return bRun;`) —
 three spellings of one check. But only Oboe's placement actually avoids the mutex; JACK's and
--- a/src/sound/README.md
+++ b/src/sound/README.md
@@ -49,7 +49,9 @@ This folder contains the related files for all sound APIs.
 ## Documentation of sound design
 
 This describes how the code behaves today. It covers the device lifecycle and the threading rules
-around the audio callback; the areas still missing are listed at the end.
+around the audio callback; the areas still missing are listed at the end. Where today's behaviour
+only makes sense in light of how it arose, the origin is given as well, so that an accident of
+history is not read as a deliberate design.
 
 Each platform provides one `CSound` class deriving from `CSoundBase`, and exactly one of the five
 backend subdirectories is compiled in: `asio/` (Windows), `coreaudio-mac/`, `coreaudio-ios/`,
@@ -139,6 +141,25 @@ ASIO defines and owns its own `ASIOMutex` (in `asio/sound.h`) instead of using t
 `MutexAudioProcessCallback`. Its own `CSound::Stop()` calls `ASIOStop()` first, then
 `CSoundBase::Stop()` (whose wait on the unused `MutexAudioProcessCallback` returns immediately
 for this backend), then waits on `ASIOMutex` directly to confirm the callback thread is done.
+
+The two are not interchangeable, because `ASIOMutex` covers more ground. It is also held across
+the whole of `CSound::Init()` — spanning `ASIODisposeBuffers()`, `ASIOCreateBuffers()` and the
+`vecsMultChanAudioSndCrd` reallocation — whereas no other backend locks anything in its `Init()`.
+The waits differ too: `CSoundBase::Stop()` blocks unconditionally on its `QMutexLocker`, while
+ASIO's `Stop()` uses `tryLock ( 5000 )` and carries on regardless if the callback has not finished
+within five seconds. The two differences compound: because ASIO's stop can return while a callback
+is still in flight, the lock held across `Init()` is what actually keeps `ASIOCreateBuffers()` off
+a live `bufferSwitch()`.
+
+That there are two mutexes at all is chronological rather than a design decision. `ASIOMutex` was
+added with the ASIO backend itself in `5eb86941` (2008-07-12), when ASIO was the only backend and
+`CSoundBase` did not yet exist (`3fb2d9ca`, 2009-02-22); its drain-on-stop wait followed in
+`73f408e4` (2011-12-27). The shared `MutexAudioProcessCallback` arrived nine years after that, in
+`ecff80fc` (2020-08-26), to fix a crash when the JACK backend was reconfigured quickly; that commit
+touches `linux/sound.cpp` and `soundbase.{h,cpp}` and nothing else. It is a later, independent
+re-implementation of a guard ASIO already had, and it was never extended to ASIO. Collapsing the
+two into one would therefore not be a rename: it would have to preserve the coverage across
+`Init()` and settle which of the two wait policies applies.
 
 ### Not yet documented
 
--- a/src/sound/README.md
+++ b/src/sound/README.md
@@ -76,8 +76,10 @@ and JACK's buffer size callback both report that by calling
 `Init()` is only ever entered with the device stopped. Callers that may be running stop it first
 and restart it afterwards, which is the `bWasRunning` pattern throughout `client.cpp`.
 
-The audio callback runs on a thread owned by the driver. Backends keep it away from a device that
-is being re-initialised in two ways, and the ASIO backend is the exception to both:
+The audio callback runs on a thread owned by the driver. `CSoundBase` inherits `QThread`, but
+nothing here overrides `run()` or calls `start()`, so no such thread exists. Backends keep the
+callback away from a device that is being re-initialised in two ways, and the ASIO backend is the
+exception to both:
 
 | backend | audio callback | ignores the callback while stopped | takes `MutexAudioProcessCallback` |
 |---|---|---|---|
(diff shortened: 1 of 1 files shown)


ann0see:
I assume that's mostly human written or not?

mcfnord:
> I assume that's mostly human written or not?

No. The LLM wrote that. What makes you think a human did? The lack of "MY LLM WROTE:" prefix? Or something else?

ann0see:
No. This is fine. It sounds a bit more natural than the other LLM responses though.

mcfnord:
> No. This is fine. It sounds a bit more natural than the other LLM responses though.

Interesting feedback. Very frequently I ask my LLM to speak in comments "with extreme concision" partly because I am sensitive to the "wall of text" criticism. So I'm often tilting it toward saying the bare minimum, as a way to avoid overwhelming the humans. Those humans who want to know will parse the dense prose.

Perhaps extreme concision isn't helpful. For this in-repo content, people kind of are hoping for a "wall of text" that guides them through a basic understanding. And I bet this kind of content will super-charge new AIs that join us.

My observation here is that I kind of wish the content routinely included hyperlinks, but I'm not sure that would work out. Maybe with relative paths?

src/sound/README.md
```diff
@@ -48,4 +48,56 @@ This folder contains the related files for all sound APIs.
...
+`Init ( iNewPrefMonoBufferSize )` returns the mono buffer size the device actually accepted,
+which may differ from the one requested. `CClient::Init()` uses that return value to find out
+which sizes a device supports, so it calls `Init()` four times per invocation: once for each of
+`FRAME_SIZE_FACTOR_PREFERRED`, `FRAME_SIZE_FACTOR_DEFAULT` and `FRAME_SIZE_FACTOR_SAFE` to fill
+`bFraSiFactPrefSupported`, `bFraSiFactDefSupported` and `bFraSiFactSafeSupported`, then once with
+the size selected in the settings. Those three flags drive the enabled state of the buffer delay
+radio buttons, and the settings dialog polls them once a second; no signal runs from the sound
+device to that dialog.
+
+A driver may also change the buffer size on its own. `kAsioBufferSizeChange` in the ASIO backend
```
pljones: I'd turn this around to state what _can_ be done, _how_ it can be done and then state which drivers currently _do_ it (ideally, "and why" as well as "and why none of the others").

src/sound/README.md
```diff
@@ -48,4 +48,56 @@ This folder contains the related files for all sound APIs.
...
+`EmitReinitRequestSignal ( RS_ONLY_RESTART_AND_INIT )`, which reaches
+`CClient::OnSndCrdReinitRequest` and repeats the negotiation above.
+
+### Start, stop and the audio callback
+
+`Init()` is only ever entered with the device stopped. Callers that may be running stop it first
+and restart it afterwards, which is the `bWasRunning` pattern throughout `client.cpp`.
+
+The audio callback runs on a thread owned by the driver. `CSoundBase` inherits `QThread`, but
+nothing here overrides `run()` or calls `start()`, so no such thread exists. Backends keep the
```
pljones: This is worth investigating.  If the architectural framework was put in place to put this on a thread, it's probably _meant_ to be on a thread or the code should be cleaned up.

The architecture is _meant_ to have one of the following:
* GUI:
  * Qt main thread at normal priority
  * CClient with CSound (to/from the audio hardware) and CChannel (to/from the network) at raised "real time" priority
* Headless:
  * CClient with CSound (to/from the audio hardware) and CChannel (to/from the network) at raised "real time" priority

src/sound/README.md
```diff
@@ -48,4 +48,56 @@ This folder contains the related files for all sound APIs.
...
+callback away from a device that is being re-initialised in two ways, and the ASIO backend is the
+exception to both:
```
pljones: I make that three ways, then.  It should be written plainly.

src/sound/README.md
```diff
@@ -48,4 +48,56 @@ This folder contains the related files for all sound APIs.
...
+
+| backend | audio callback | ignores the callback while stopped | takes `MutexAudioProcessCallback` |
```
pljones: It might be worth seeing why this variation exists, anyway.

src/sound/README.md
```diff
@@ -48,4 +48,56 @@ This folder contains the related files for all sound APIs.
...
+|---|---|---|---|
+| JACK | `process()` | yes, `IsRunning()` | yes |
+| CoreAudio (macOS) | `callbackIO()` | yes, `bRun` | yes |
+| CoreAudio (iOS) | `processBufferList()` | no | yes |
+| Oboe | `onAudioReady()` | yes, `!bRun` | yes |
+| ASIO | `bufferSwitch()` | no | no, it uses its own `ASIOMutex` |
+
+`CSoundBase::Stop()` clears `bRun` and then takes `MutexAudioProcessCallback` to wait for a
+callback that is already in flight. The ASIO backend never takes that mutex, so on Windows that
+wait returns immediately and `CSound::Stop()` waits on `ASIOMutex` instead.
```
pljones: Make it clear whether `ASIOMutex` is owned by the ASIO code or by the Jamulus code.

src/sound/README.md
```diff
@@ -90,8 +90,10 @@ exception to both:
 | ASIO | `bufferSwitch()` | no | no, it uses its own `ASIOMutex` |
 
 `CSoundBase::Stop()` clears `bRun` and then takes `MutexAudioProcessCallback` to wait for a
-callback that is already in flight. The ASIO backend never takes that mutex, so on Windows that
-wait returns immediately and `CSound::Stop()` waits on `ASIOMutex` instead.
+callback that is already in flight. The ASIO backend is the exception: it defines and owns its
```
pljones: OK, I guess this leaves open the question: is there a good reason for having a different way of doing it for ASIO?   (My guess: no, it was just the second backend and done completely separately without forward thinking; only when the third backend came along did the idea of an architecture appear and it's not particularly apparent even now...)  And the next question: is it worth using the same mutex?

src/sound/README.md
mcfnord: **🤖 AI:** On the `QThread` question: `run()` is never overridden and `start()` is never called anywhere in this class's tracked history, back to the commit that introduced it, and no other `QThread`-specific method (`setPriority`, `msleep`, `currentThread`) appears anywhere in the sound layer. There is no thread to put anything on. Changing the base to `QObject` requires nothing from this file; whether anything downstream relies on the class actually being a `QThread` (a `qobject_cast<QThread*>`, `moveToThread`) was not checked, so that change stays out of this PR rather than being asserted risk-free.

On ASIO doing it differently: the history inverts the guess. `ASIOMutex` arrives on 2008-07-12 (`5eb86941`) with the ASIO backend itself, when ASIO was the *only* backend — seven months before `CSoundBase` exists at all (`3fb2d9ca`, 2009-02-22). Its drain-on-stop wait was added on 2011-12-27 (`73f408e4`, "make sure that sound working thread is actually killed on stopping the audio interface"). The shared `MutexAudioProcessCallback` appears nine years after that, on 2020-08-26 (`ecff80fc`), as a fix for a JACK crash on quick reconfiguration; that commit touches `linux/sound.cpp` and `soundbase.{h,cpp}` and nothing else. So ASIO did not skip an existing architecture — the shared mutex is a later, independent re-implementation of a guard ASIO already had, scoped to the JACK bug that prompted it and never extended outward.

On whether the same mutex is worth using: not as a substitution, because the two are not doing the same job today. `ASIOMutex` is also held across the whole of `CSound::Init()` — `ASIODisposeBuffers()`, `ASIOCreateBuffers()` and the `vecsMultChanAudioSndCrd` reallocation — and no other backend locks anything in `Init()`. The waits differ too: `CSoundBase::Stop()` blocks unconditionally on a `QMutexLocker`, while ASIO's uses `tryLock ( 5000 )` and returns anyway on timeout. Since ASIO's stop can return with a callback still in flight, that `Init()` lock is what keeps `ASIOCreateBuffers()` off a live `bufferSwitch()`. Narrowing exactly that scope is what #3869 proposes, so the merge question sits better with it than here: unifying the two would have to preserve the coverage across `Init()` and settle which of the two wait policies applies.

Two unrelated items surfaced in the same pass, both out of scope for a docs PR: `CSound::onAudioInput()` in the Oboe backend `memcpy`s using the driver-delivered frame count into a buffer sized for the requested count, and the mismatch check runs after the copy rather than before — a device delivering more frames than requested overflows the buffer before anything notices. And the callback-guarding column in the revised table takes `AudioOutputUnitStop()` and `ASIOStop()` to synchronously stop the driver calling back again; that is untested on hardware.

src/sound/README.md
```diff
@@ -66,34 +68,77 @@ the size selected in the settings. Those three flags drive the enabled state of
...
+JACK's buffer-size callback (`jack/sound.cpp`) both call
 `EmitReinitRequestSignal ( RS_ONLY_RESTART_AND_INIT )`, which reaches
-`CClient::OnSndCrdReinitRequest` and repeats the negotiation above.
+`CClient::OnSndCrdReinitRequest` and repeats the negotiation above. The other three backends
+can't do this the same way: the two CoreAudio backends only watch device-identity/route events
+(`kAudioDevicePropertyDeviceHasChanged`/`IsAlive`/default-device switches on macOS,
+`AVAudioSessionRouteChangeNotification` on iOS) and never register a buffer-size-specific
+listener, so a size change on its own is invisible to them. Oboe does detect it —
+`onAudioInput()` compares the delivered frame count against the requested size on every callback
+— but the mismatch is only logged (`qDebug()`), never renegotiated.
```
pljones, line 84: Ah the joy of code reviews...  I guess we'll need a few backend-specific issues to clean up those inconsistencies.

src/sound/README.md
```diff
@@ -66,34 +68,77 @@ the size selected in the settings. Those three flags drive the enabled state of
...
-nothing here overrides `run()` or calls `start()`, so no such thread exists. Backends keep the
-callback away from a device that is being re-initialised in two ways, and the ASIO backend is the
-exception to both:
-
-| backend | audio callback | ignores the callback while stopped | takes `MutexAudioProcessCallback` |
+nothing here overrides `run()` or calls `start()`, so no such thread exists — nor has it ever:
+neither appears anywhere in this file's tracked history, and no other `QThread`-specific method
+is used in the sound layer either. The inheritance predates the current driver-callback
+architecture and contributes nothing; changing it to `QObject` looks safe from this file alone,
+but that wasn't verified further here.
```
pljones: Probably best recorded in an issue rather than this file (with "but see issue #xxx" if you like).

src/sound/README.md
```diff
@@ -66,34 +68,77 @@ the size selected in the settings. Those three flags drive the enabled state of
...
+| ASIO | (no override — `ASIOStop()` happens inside `CSound::Stop()` itself, see below) |
+| CoreAudio (macOS) | `AudioDeviceStop()` + `AudioDeviceDestroyIOProcID()` |
+| CoreAudio (iOS) | `AudioOutputUnitStop()` |
+| Oboe | `closeStreams()` |
+| JACK | nothing — goes straight to `CSoundBase::Stop()` |
+
+Inside the callback itself, backends differ again, and not just in which flag they read but in
+*when* they read it relative to the mutex:
+
+| backend | audio callback | flag check | mutex behavior |
```
pljones, line 113: Try to keep a consistent order between each "backend"-based table - it makes following a specific backend easier.

src/sound/README.md
mcfnord: **🤖 AI:** Now the three are named ahead of the tables that detail them: the driver-level stop call in a backend's own `Stop()`, the `bRun` flag the callback may test, and the mutex `Stop()` shares with the callback. No two of the five backends combine them the same way.

Checking that turned up a false sentence in the pushed text — "Every backend but ASIO also overrides `Stop()`", sitting above a table row reading "no override". [ASIO's override](https://github.com/jamulussoftware/jamulus/blob/b3c93588bb447d2c47987aba35284131a3e3c456/src/sound/asio/sound.cpp#L524-L538) does exist, and is one of the four that calls a driver-level stop first; [JACK's is the only one](https://github.com/jamulussoftware/jamulus/blob/b3c93588bb447d2c47987aba35284131a3e3c456/src/sound/jack/sound.cpp#L197-L201) going straight to the base class. Both corrected, along with the iOS row of the callback table: the driver-invoked callback is [`recordingCallback()`](https://github.com/jamulussoftware/jamulus/blob/b3c93588bb447d2c47987aba35284131a3e3c456/src/sound/coreaudio-ios/sound.mm#L92-L125), and the mutex is taken [inside `processBufferList()`](https://github.com/jamulussoftware/jamulus/blob/b3c93588bb447d2c47987aba35284131a3e3c456/src/sound/coreaudio-ios/sound.mm#L127-L140) only, so the output copy back into `ioData` runs unlocked.

src/sound/README.md
mcfnord: **🤖 AI:** Trimmed to the fact — no such thread is ever created — with the history and the `QObject` suggestion out of the file, for an issue instead.

On the architecture point: the raised-priority half of it does exist, on the network side. [`CHighPrioSocket::Start()`](https://github.com/jamulussoftware/jamulus/blob/b3c93588bb447d2c47987aba35284131a3e3c456/src/socket.h#L190-L195) runs its `CSocketThread` at `QThread::TimeCriticalPriority`. Sound has no equivalent: [`CSoundBase : public QThread`](https://github.com/jamulussoftware/jamulus/blob/b3c93588bb447d2c47987aba35284131a3e3c456/src/sound/soundbase.h#L88) is never started, and `src/sound/` holds no `start()` call, no `run()` override and no other `QThread` method at all — so the audio callback arrives on a thread the driver owns, at whatever priority that driver gives it, a value Jamulus neither sets nor reads.

src/sound/README.md
mcfnord: **🤖 AI:** The candidates this pass surfaced, each self-contained:

- A buffer-size change on the device's own initiative goes unnoticed on three of the five. Oboe [compares and only logs](https://github.com/jamulussoftware/jamulus/blob/b3c93588bb447d2c47987aba35284131a3e3c456/src/sound/oboe/sound.cpp#L275-L278) it; neither CoreAudio backend registers a size listener at all.
- [The Oboe input copy](https://github.com/jamulussoftware/jamulus/blob/b3c93588bb447d2c47987aba35284131a3e3c456/src/sound/oboe/sound.cpp#L273) uses the driver-delivered frame count into a buffer sized for the requested count, and the mismatch check runs after the copy rather than before it.
- On iOS the mutex covers the input half only, so [the output copy](https://github.com/jamulussoftware/jamulus/blob/b3c93588bb447d2c47987aba35284131a3e3c456/src/sound/coreaudio-ios/sound.mm#L117-L122) back into `ioData` is unlocked.
- The running-flag check sits after the lock in JACK and CoreAudio (macOS), before it in Oboe, and is absent in ASIO and CoreAudio (iOS) — so two backends skip the processing at full mutex cost, and two never check.
- The two mutexes: `ASIOMutex` is also held across the whole of `CSound::Init()`, which the shared one never is anywhere, so collapsing them is a behaviour change rather than a rename.

src/sound/README.md
pljones: If you've not, could you raise issues for these as bugs, please.

src/sound/README.md
pljones: As noted.  These are for future investigation and should have their own follow up issues.

ann0see:
Just as somewhat related comment: let your AI read the paper by Volker: https://jamulus.app/PerformingBandRehearsalsontheInternetWithJamulus.pdf

src/sound/README.md
```diff
@@ -48,4 +48,124 @@ This folder contains the related files for all sound APIs.
 
 ## Documentation of sound design
 
-**Fixme:** The sound design is not yet documented.
+This describes how the code behaves today. It covers the device lifecycle and the threading rules
```
pljones, line 51: As a general comment, this appears to be rewriting the code in English and then putting some summaries around it.

It doesn't really document the _design_, more the implementation -- which is what "This describes how the code behaves today" would lead one to understand, admittedly.

What's really needed is the _intent_.  Start with that as a grounding point for the reader (human or AI) - it supplies future direction as well and explaining any a basis to assess existing shortcomings.

There's too much detail here that's really just saying "it's like this now".

Issue #3874, opened 2026-08-08
Headless client: the gain/pan rate limiter never adapts to ping time, and branches on an uninitialised value


mcfnord:
**🤖 AI:** The rate limiter added in #2535 has two halves: a [50 ms floor](https://github.com/jamulussoftware/jamulus/blob/c862872e7beacdc8c9fd232cb33def2d93ee01d1/src/client.h#L113), and double the current ping time, whichever is greater. In a headless build only the floor is ever used, and the value the branch tests has never been written.

[`CClient::StartTimerGainOrPan`](https://github.com/jamulussoftware/jamulus/blob/c862872e7beacdc8c9fd232cb33def2d93ee01d1/src/client.cpp#L575-L591) tests [`iCurPingTime`](https://github.com/jamulussoftware/jamulus/blob/c862872e7beacdc8c9fd232cb33def2d93ee01d1/src/client.h#L479). That member is assigned in [exactly one place](https://github.com/jamulussoftware/jamulus/blob/c862872e7beacdc8c9fd232cb33def2d93ee01d1/src/client.cpp#L449), `CClient::OnCLPingReceived`, which is reached only after [`CClient::CreateCLPingMes()`](https://github.com/jamulussoftware/jamulus/blob/c862872e7beacdc8c9fd232cb33def2d93ee01d1/src/client.h#L315). That function has [one caller](https://github.com/jamulussoftware/jamulus/blob/c862872e7beacdc8c9fd232cb33def2d93ee01d1/src/clientdlg.cpp#L1133) in the tree: `CClientDlg::OnTimerPing`. A client started with `-n` has no dialog, so no ping is sent and `iCurPingTime` is never assigned.

A recording UDP proxy sat between a real client and a real server with a symmetric delay injector; five bursts of 150 back-to-back fader changes per trial; [`CLM_PING_MS`](https://github.com/jamulussoftware/jamulus/blob/c862872e7beacdc8c9fd232cb33def2d93ee01d1/src/protocol.h#L91) (1001) frames counted and gain-message spacing read off the capture:

| client | round trip | CLM_PING_MS frames on wire | gain spacing |
|---|---:|---:|---:|
| headless | 0 ms | 0 | 49 ms |
| headless | 122 ms | 0 | 121 ms |
| GUI | 2 ms | 61 | 49 ms |
| GUI | 60 ms | 66 | 122 ms |
| GUI | 122 ms | 64 | 244 ms |

[...]

PR #3877, opened 2026-08-08
Comment accuracy: minimal wording fixes in src/

mcfnord:
**🤖 AI**

Fifteen comments in `src/` state something the code does not do, and three misspell a word. Each fix is the smallest wording change that makes the sentence true; the diff is comment-only (+25/−26, no code line changes), and each commit message carries the measurement or code reference behind its correction.

- `buffer.cpp` (`Put`) - the block-count divide is exact only while `iNumBlocks * iNumBytesSeqNum` stays below `iBlockSize`; the per-block size ratio is not the condition.
- `buffer.cpp` (sequence wrap) - the fold above limits detection to 128 counts, half the wrap distance, and 128 counts is at least 170 ms.
- `client.cpp` (`OnClientIDReceived`) - a 30 s traffic gap (`CON_TIME_OUT_SEC_MAX`) also produces non-zero `iActiveChannels`, from a server that never restarted.
- `client.cpp` (same function) - `FindClientChannel` returns `INVALID_INDEX` for an out-of-range ID, as `client.h` documents; channel 0 is the in-range result.
- `client.cpp` (gain/pan rate limiter) - only a GUI client measures the ping time the timer period depends on ([#3874](https://github.com/jamulussoftware/jamulus/issues/3874)).
- `server.cpp` (`OnTimer`) - the flag's `QtConcurrent::run` five-parameter rationale is stale: dispatch is `CThreadPool::enqueue`, which is variadic.
- `main.cpp` - a GUI server with a fresh ini writes `<directorytype>-1</directorytype>`, which is `AT_NONE`, so the "always default to registering somewhere" premise goes; the justification for not validating here stands without it.
- `connectdlg.cpp` - the width of every visible column is set, including the last, and `setStretchLastSection ( false )` six lines below is what stops the last column taking the remaining space.
- `global.h` - the 128 assumption belongs to the `IIR_WEIGTH_*` filter constants in `buffer.h`, which come in a 64 set and a 128 set, not to `UpdateAutoSetting()`.
- `client.h` - swapping the `pSignalHandler`/`pSettings` declarations and rebuilding gives byte-identical startup output, so the init-order comment goes.
- `socket.h` - `TimeCriticalPriority` is requested, not guaranteed: on unprivileged Linux the thread runs `SCHED_OTHER`, rt_priority 0 (the `IdlePriority` control does take effect, downward), while macOS grants sched_priority 47 against a main thread at 31.
- `socket.cpp` - a failed IPv4 `socket()` is reachable under descriptor exhaustion, so "should never happen" goes; "IPv4 not available" did not occur even inside an empty network namespace.
- `clientdlg.cpp` - the delivery is queued with or without the explicit `Qt::QueuedConnection`, because the emit always comes from a `QtConcurrent` worker thread; the requirement belongs to the delivery, not this flag.
- `audiomixerboard.cpp` - a `size_t` arm of the in-use array behaves identically (the `INVALID_INDEX` sentinel round-trips); what `int` buys is a comparison free of signedness warnings.
- `channel.cpp` (`GetUploadRateKbps`) - the 77-byte overhead is the PPPoE-over-ATM DSL access path of the cited paper; the comment now names the model it quotes.
- Spelling: `clitches` → `glitches` (`server.cpp`), `sequred` → `secured` (`protocol.h`), `effect` → `affect` (`socket.h`).

This replaces [#3866](https://github.com/jamulussoftware/jamulus/pull/3866) and applies the review feedback there: comments stay short, say what is rather than what is not, and carry no experimental data - the measurements live in the commit messages. `clang-format` 14 reports every touched file clean. Sites where the measurement points at the code rather than the comment are excluded and will be raised separately.

---

### DIFF
--- a/src/audiomixerboard.cpp
+++ b/src/audiomixerboard.cpp
@@ -1339,7 +1339,8 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
         // get all channels which are in use/not in use.
         // We use the array index of vecChanInfo if the fader is in use,
         // else INVALID_INDEX to specify it is not in use
-        // so must use "int" for the array type.
+        // so the array type is "int", which also keeps the comparisons below
+        // free of signedness warnings.
         int iFaderNumber[MAX_NUM_CHANNELS];
 
         for ( size_t iChanID = 0; iChanID < MAX_NUM_CHANNELS; iChanID++ )
--- a/src/buffer.cpp
+++ b/src/buffer.cpp
@@ -164,8 +164,8 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
             return false;
         }
 
-        // to get the number of input blocks we assume that the number of bytes for
-        // the sequence number is much smaller than the number of coded audio bytes
+        // to get the number of input blocks we assume that the total sequence number
+        // overhead, iNumBlocks * iNumBytesSeqNum, is smaller than iBlockSize
         const int iNumBlocks = /* floor */ ( iInSize / iBlockSize );
 
         // copy new data in internal buffer
--- a/src/buffer.cpp
+++ b/src/buffer.cpp
@@ -191,8 +191,8 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
             }
 
             // The 1-byte sequence number wraps around at a count of 256. So, if a packet is delayed
-            // further than this we cannot detect it. But it does not matter since such a packet is
-            // more than 100 ms delayed so we have a bad network situation anyway. Therefore we
+            // further than half of this we cannot detect it. But it does not matter since such a packet is
+            // more than 170 ms delayed so we have a bad network situation anyway. Therefore we
             // assume that the sequence number difference between the received and local counter is
             // correct. The idea of the following code is that we always move our "buffer window" so
             // that the received packet fits into the buffer. By doing this we are robust against
--- a/src/channel.cpp
+++ b/src/channel.cpp
@@ -716,9 +716,9 @@ int CChannel::GetUploadRateKbps()
 {
     const int iAudioSizeOut = iNetwFrameSizeFact * iAudioFrameSizeSamples;
 
-    // we assume that the UDP packet which is transported via IP has an
-    // additional header size of ("Network Music Performance (NMP) in narrow
-    // band networks; Carot, Kraemer, Schuller; 2006")
+    // we assume the PPPoE-over-ATM DSL access path described in ("Network Music
+    // Performance (NMP) in narrow band networks; Carot, Kraemer, Schuller; 2006"),
+    // whose additional header size is
     // 8 (UDP) + 20 (IP without optional fields) = 28 bytes
     // 2 (PPP) + 6 (PPPoE) + 18 (MAC)            = 26 bytes
     // 5 (RFC1483B) + 8 (AAL) + 10 (ATM)         = 23 bytes
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -1009,7 +1009,7 @@ void CClient::OnClientIDReceived ( int iServerChanID )
     }
 
     // allocate and map client-side channel 0
-    int iChanID = FindClientChannel ( iServerChanID, true ); // should always return channel 0
+    int iChanID = FindClientChannel ( iServerChanID, true ); // returns channel 0 for an in-range iServerChanID
 
     // for headless mode we support to mute our own signal in the personal mix
     // (note that the check for headless is done in the main.cpp and must not
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -482,7 +482,7 @@ void CClient::SetDoAutoSockBufSize ( const bool bValue )
 //
 // When the first gain or pan change message is requested after an idle period (i.e. the timer is not
 // running), it will be sent immediately, and a timer started. The timer period is dependent on
-// the current ping time to the remote server.
+// the current ping time to the remote server, which only a GUI client measures (see #3874).
 //
 // If a gain or pan change message is requested while the timer is still running, the new value is not sent,
 // but just stored in newGain or newPan within clientChannels[iId], and the minGainOrPanId and maxGainOrPanId
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -1000,7 +1000,7 @@ void CClient::OnControllerInMuteMyself ( bool bMute )
 void CClient::OnClientIDReceived ( int iServerChanID )
 {
     // if we have just connected to a running server, iActiveChannels will be 0
-    // if iActiveChannels is not 0, the server must have been restarted on the fly
+    // if iActiveChannels is not 0, the server was restarted or our channel timed out
     // in that case, channels might have changed, so clear our list to get it afresh.
     if ( iActiveChannels != 0 )
     {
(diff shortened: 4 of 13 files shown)


src/audiomixerboard.cpp
```diff
@@ -1339,7 +1339,8 @@ void CAudioMixerBoard::ApplyNewConClientList ( CVector<CChannelInfo>& vecChanInf
         // get all channels which are in use/not in use.
         // We use the array index of vecChanInfo if the fader is in use,
         // else INVALID_INDEX to specify it is not in use
-        // so must use "int" for the array type.
+        // so the array type is "int", which also keeps the comparisons below
+        // free of signedness warnings.
```
pljones, line 1343: Seems a pretty strange way of putting it.

```
else INVALID_INDEX to specify it is not in use.
As INVALID_INDEX is of type int, an int type must also be
used here to avoid signedness warnings elsewhere.
```

src/buffer.cpp
```diff
@@ -191,8 +191,8 @@ bool CNetBuf::Put ( const CVector<uint8_t>& vecbyData, int iInSize )
             }
 
             // The 1-byte sequence number wraps around at a count of 256. So, if a packet is delayed
-            // further than this we cannot detect it. But it does not matter since such a packet is
-            // more than 100 ms delayed so we have a bad network situation anyway. Therefore we
+            // further than half of this we cannot detect it. But it does not matter since such a packet is
+            // more than 170 ms delayed so we have a bad network situation anyway. Therefore we
```
pljones, line 195: Actually, I find the whole comment (not this changed bit) very hard to digest with lots of repeated phrases.
> we always move our "buffer window"
> since we adjust the window
> we move the "buffer window"

src/channel.cpp
```diff
@@ -716,9 +716,9 @@ int CChannel::GetUploadRateKbps()
 {
     const int iAudioSizeOut = iNetwFrameSizeFact * iAudioFrameSizeSamples;
 
-    // we assume that the UDP packet which is transported via IP has an
-    // additional header size of ("Network Music Performance (NMP) in narrow
-    // band networks; Carot, Kraemer, Schuller; 2006")
+    // we assume the PPPoE-over-ATM DSL access path described in ("Network Music
+    // Performance (NMP) in narrow band networks; Carot, Kraemer, Schuller; 2006"),
+    // whose additional header size is
```
pljones, line 721: This comment now omits the first line (UDP) from the description.  It's probably also worth stating something like "PPPoE-over-ATM DSL chosen as a general domestic use case" (or whatever reason the authors gave in their paper).

src/client.cpp
```diff
@@ -482,7 +482,7 @@ void CClient::SetDoAutoSockBufSize ( const bool bValue )
 //
 // When the first gain or pan change message is requested after an idle period (i.e. the timer is not
 // running), it will be sent immediately, and a timer started. The timer period is dependent on
-// the current ping time to the remote server.
+// the current ping time to the remote server, which only a GUI client measures (see #3874).
```
pljones, line 485: Not worth mentioning.  Any "bad design" should be captured in an issue and addressed that way, rather than in code comments.

src/clientdlg.cpp
```diff
@@ -575,8 +575,8 @@ CClientDlg::CClientDlg ( CClient*         pNCliP,
 
     QObject::connect ( &ConnectDlg, &CConnectDlg::ReqServerListQuery, this, &CClientDlg::OnReqServerListQuery );
 
-    // note that this connection must be a queued connection, otherwise the server list ping
-    // times are not accurate and the client list may not be retrieved for all servers listed
+    // note that this delivery must be queued, otherwise the server list ping times are not
+    // accurate and the client list may not be retrieved for all servers listed
```
pljones, line 579: Hm.  That's not what `Qt::QueuedConnection` _means_.  The explanation should indicate something about the effect within the Qt signal/slot framework of the parameter and then why it's needed within the Jamulus architecture here.

src/global.h
```diff
@@ -102,7 +102,7 @@ LED bar:      lbr
 
 // System block size, this is the block size on which the audio coder works.
 // All other block sizes must be a multiple of this size.
-// Note that the UpdateAutoSetting() function assumes a value of 128.
+// Note that the IIR_WEIGTH_* filter constants in buffer.h assume values of 64 and 128.
```
pljones, line 105: Possibly worth splitting out "completely wrong" from "unclear or unhelpful".  This looks like the former.

src/server.cpp
```diff
@@ -662,7 +662,7 @@ void CServer::OnTimer()
     bool bUseMT               = false;
     int  iNumBlocks           = 0;     // init number of blocks for multithreading
     int  iMTBlockSize         = 0;     // init block size for multithreading
-    bChannelIsNowDisconnected = false; // note that the flag must be a member function since QtConcurrent::run can only take 5 params
+    bChannelIsNowDisconnected = false; // note that the flag is a member since DecodeReceiveData sets it and the check below reads it
```
pljones, line 665: Hm.  Seems a pretty useless comment now.  "This is a flag.  It's read and written."  The earlier comment explained why it existed at all.

src/socket.h
```diff
@@ -151,9 +151,9 @@ class CSocket : public QObject
     void ProtocolCLMessageReceived ( int iRecID, CVector<uint8_t> vecbyMesBodyData, CHostAddress HostAdr );
 };
 
-/* Socket which runs in a separate high priority thread --------------------- */
+/* Socket which runs in a separate thread requesting high priority ---------- */
 // The receive socket should be put in a high priority thread to ensure the GUI
-// does not effect the stability of the audio stream (e.g. if the GUI is on
+// does not affect the stability of the audio stream (e.g. if the GUI is on
 // high load because of a table update, the incoming network packets must still
 // be put in the jitter buffer with highest priority).
```
pljones, line 158: (non-feedback)
It's really the GUI that needs its own thread.  The main Jamulus Client should _all_ run at real time scheduling / raised priority: audio, client and network.  That part of the code needs to be kept "lean and mean" in terms of performance...

(I'll continue rumbling about this...)

src/client.cpp
mcfnord: Overall AI hasn't done great on this "update the comments" exercise, but I think this change is valuable. Since the day robots took my writing career, I miss this eternal tension between developers and documentarians. The perfect should not be the enemy of the good. Futureware doesn't help anyone today. Should this matter become fixed, one could revel and delight in the removal of this mention of the fixed flaw. I've linked arms with developers and spun with delight as we threw out obsolete guidance. This worked well so long as I was the benevolent dictator of the documentation. Once the world moved to repos and toward engineers applying their engineering acumen to editorial judgements, as you do here, quality has suffered. Perfected futureware is the psychological bias of the lead engineer. Why mention a thing we should fix? To avoid suffering, that's why. My career involved using a magnifying lense on things developers want to disappear, and sometimes even hide. So this is nothing new. But please consider my differing point of view, oh benevolent dictator.

pljones: > Should this matter become fixed, one could revel and delight in the removal of this mention of the fixed flaw.

No, it'll get overlooked.  Probably even by an AI coding agent.
> Headless client: the gain/pan rate limiter never adapts to ping time, and branches on an uninitialised value

It's not clear from the title of the referenced PR that it's _directly_ related.  And any "This is wrong" should have a "Fix in" reference rather than a "discovered as wrong in" reference.  Again, it's negativity.

ann0see (commented):
Anything we can get out of this? Otherwise we should close this.

PR #3893, opened 2026-08-12
Rework the opening of UDP sockets for IPv4 and IPv6

### LINKED ISSUE #3885: Investigate socket.cpp and buffer.cpp findings in AI code quality
**Describe the bug**

There are currently some findings (likely some false positives also) in https://github.com/jamulussoftware/jamulus/security/quality/ai-findings which need further investigation

**To Reproduce**

Devs with access to the tab can open https://github.com/jamulussoftware/jamulus/security/quality/ai-findings

**Expected behavior**

The findings should only have false positives.

**Screenshots**

/

**Operating system**

/

**Version of Jamulus**

current main

**Additional context**

some in buffer.cpp and socket.cpp need investigation

softins:

Reworks the socket opening logic introduced in #3774, which had potential shortcomings identified by AI.
Not for backporting, as #3774 was not also.

CHANGELOG: Client/Server: reworked and improved UDP socket opening logic

**Context: Fixes an issue?**

Fixes #3885

No, bug fixes only

Ready for review

Checking that it satisfies CoPilot AI shortcomings that were flagged in #3885 for socket.cpp

## Checklist

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -334,11 +334,8 @@ void CSocket::Init ( const quint16  iNewPortNumber,
             // only try to bind the IPv6 socket if IPv4 has succeeded and IPv6 socket is open
             if ( bSuccess && UdpSocket6 != INVALID_SOCKET )
             {
-                if ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) != -1 )
-                {
-                    // note that IPv6 is available
-                    bIPv6Available = true; // this is a reference to CClient::bIPv6Available or CServer::bIPv6Available
-                }
+                // note whether IPv6 is available - this is a reference to CClient::bIPv6Available or CServer::bIPv6Available
+                bIPv6Available = ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) != -1 );
             }
         }
         else
@@ -376,11 +373,9 @@ void CSocket::Init ( const quint16  iNewPortNumber,
                 if ( bSuccess && UdpSocket6 != INVALID_SOCKET && !bIPv6Available )
                 {
                     sa6.sin6_port = htons ( port );
-                    if ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) != -1 )
-                    {
-                        // note that IPv6 is available
-                        bIPv6Available = true; // this is a reference to CClient::bIPv6Available or CServer::bIPv6Available
-                    }
+
+                    // note whether IPv6 is available - this is a reference to CClient::bIPv6Available or CServer::bIPv6Available
+                    bIPv6Available = ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) != -1 );
                 }
 
                 if ( bSuccess && ( bIPv6Available || UdpSocket6 == INVALID_SOCKET ) )
@@ -403,11 +398,8 @@ void CSocket::Init ( const quint16  iNewPortNumber,
         // only try to bind the IPv6 socket if IPv4 has succeeded and IPv6 socket is open
         if ( bSuccess && UdpSocket6 != INVALID_SOCKET )
         {
-            if ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) != -1 )
-            {
-                // note that IPv6 is available
-                bIPv6Available = true; // this is a reference to CClient::bIPv6Available or CServer::bIPv6Available
-            }
+            // note whether IPv6 is available - this is a reference to CClient::bIPv6Available or CServer::bIPv6Available
+            bIPv6Available = ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) != -1 );
         }
     }
 
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -346,7 +346,7 @@ void CSocket::Init ( const quint16  iNewPortNumber,
             // faulty router gets stuck and confused by a particular port (like
             // the starting port). Might work around frustrating "cannot connect"
             // problems (#568)
-            const quint32 startingPortNumber = static_cast<quint32> ( iPortNumber ) + rand() % NUM_SOCKET_PORTS_TO_TRY;
+            const quint32 startingPortNumber = static_cast<quint32> ( iPortNumber ) + QRandomGenerator::global()->bounded ( NUM_SOCKET_PORTS_TO_TRY );
 
             for ( quint32 port = startingPortNumber; port < startingPortNumber + NUM_SOCKET_PORTS_TO_TRY; port++ )
             {
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -348,7 +348,14 @@ void CSocket::Init ( const quint16  iNewPortNumber,
             // faulty router gets stuck and confused by a particular port (like
             // the starting port). Might work around frustrating "cannot connect"
             // problems (#568)
-            const quint32 startingPortNumber = static_cast<quint32> ( iPortNumber ) + QRandomGenerator::global()->bounded ( NUM_SOCKET_PORTS_TO_TRY );
+            quint32 startingPortNumber = static_cast<quint32> ( iPortNumber ) + QRandomGenerator::global()->bounded ( NUM_SOCKET_PORTS_TO_TRY );
+
+            // in the unlikely event we were given a port number too high, move lower,
+            // so we at least try to bind once.
+            if ( startingPortNumber > 65535U )
+            {
+                startingPortNumber -= NUM_SOCKET_PORTS_TO_TRY;
+            }
 
             for ( quint32 port = startingPortNumber; port < startingPortNumber + NUM_SOCKET_PORTS_TO_TRY; port++ )
             {
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -172,6 +172,8 @@ void CSocket::Init ( const quint16  iNewPortNumber,
         UdpSocket6 = INVALID_SOCKET;
     }
 
+    bIPv6Available = false; // re-init before opening sockets
+
     struct sockaddr_in sa4;
     socklen_t          sa4len = sizeof ( sa4 );
     memset ( &sa4, 0, sa4len );
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -290,8 +290,6 @@ void CSocket::Init ( const quint16  iNewPortNumber,
                 }
             }
 
-            bIPv6Available = true; // this is a reference to CClient::bIPv6Available or CServer::bIPv6Available
-
             // set socket to non-blocking
 #ifdef _WIN32
             unsigned long mode = 1;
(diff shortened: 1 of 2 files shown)


src/socket.cpp
```diff
@@ -341,23 +346,38 @@ void CSocket::Init ( const quint16  iNewPortNumber,
             // faulty router gets stuck and confused by a particular port (like
             // the starting port). Might work around frustrating "cannot connect"
             // problems (#568)
-            const quint16 startingPortNumber = iPortNumber + rand() % NUM_SOCKET_PORTS_TO_TRY;
```
softins, line 344: Since this was the first call to `rand()` after program startup, and `rand()` always starts from the same seed, this was not providing any randomness at all. The same start port was chosen on each invocation, so this wouldn't actually address #568 as claimed. Verified on Linux and Mac.

src/socket.cpp
```diff
@@ -341,23 +346,38 @@ void CSocket::Init ( const quint16  iNewPortNumber,
...
-
-            quint16 iClientPortIncrement = 0;
-            bSuccess                     = false; // initialization for while loop
+            const quint32 startingPortNumber = static_cast<quint32> ( iPortNumber ) + QRandomGenerator::global()->bounded ( NUM_SOCKET_PORTS_TO_TRY );
```
softins: This provides a truly random starting port.

src/socket.cpp
pljones, line 429: Might be worth expanding this to explain more about what failed as there are more possible causes now.  (Not due to this PR -- I've just read it to get more context and realised it's a bit weak.)

softins: Actually, `bSuccess` only now applies to IPv4. If we fail to create and bind IPv4, `bSuccess` will be false and we throw the exception. If we succeed with IPv4 but fail with IPv6, it's just logged with a message, but `bSuccess` is still true. So there isn't really any more info to pass to the exception apart from what is already there.

src/socket.cpp
ann0see: Doesn't C++ have something inbuilt? I'd prefer something that not depends on Qt.

softins: Not sure I understand that objection, given that the whole of Jamulus literally depends on Qt!

softins: There is https://en.cppreference.com/cpp/numeric/random/rand, but it says "rand() is not recommended for serious random-number generation needs. It is recommended to use C++11's [random number generation](https://en.cppreference.com/cpp/numeric/random) facilities to replace rand()". That is a lot more involved than just using `QRandomGenerator`. The latter hides all the complexity, which is, after all, what Qt is for!

ann0see: Ok. Then we leave it as is. The rationale behind this was the - yet not formalized - goal to slowly move away from relying on Qt.

softins:
I'd like to get this in before tagging beta3

src/socket.cpp
dingodoppelt: > Ok. Then we leave it as is. The rationale behind this was the - yet not formalized - goal to slowly move away from relying on Qt.

I second that but I'm fine as well with leaving this as is.

src/socket.cpp
mcfnord: **🤖 AI:** Yes, and one of the causes is new. Four distinct paths reach [this exact sentence](https://github.com/jamulussoftware/jamulus/blob/05b4103834798e3c7493c4a1d06640a90658f250/src/socket.cpp#L422-L424) at `05b41038`: `EADDRINUSE` (port already held), `EADDRNOTAVAIL` ([`--serverbindip4`](https://github.com/jamulussoftware/jamulus/blob/05b4103834798e3c7493c4a1d06640a90658f250/src/main.cpp#L451) naming an address not on the host), `EACCES` (server on port 80, unprivileged), and one where **no `bind` was attempted at all**: [`startingPortNumber`](https://github.com/jamulussoftware/jamulus/blob/05b4103834798e3c7493c4a1d06640a90658f250/src/socket.cpp#L351) is `iPortNumber` plus up to 99, so a client `--port` from 65437 up can start past 65535 and [the new guard](https://github.com/jamulussoftware/jamulus/blob/05b4103834798e3c7493c4a1d06640a90658f250/src/socket.cpp#L356-L359) breaks on the first pass — 25 of 25 runs at `--port 65535` made zero `bind` calls and printed this message.

So the errno is the information the text is missing. The fourth case is a separate matter: resetting the start to `iPortNumber` when it would overflow binds 65535 in 25 of 25 runs.

src/socket.cpp
```diff
@@ -318,20 +318,27 @@ void CSocket::Init ( const quint16  iNewPortNumber,
...
     {
+        // for a client, it does not matter if the IPv4 and IPv6 sockets get bound
+        // to different local port numbers
         if ( iPortNumber == 0 )
         {
             // if port number is 0, bind the client to a random available port
             sa4.sin_port = sa6.sin6_port = htons ( 0 );
 
-            bSuccess = ( ::bind ( UdpSocket4, (struct sockaddr*) &sa4, sa4len ) == 0 );
+            bSuccess = ( ::bind ( UdpSocket4, (struct sockaddr*) &sa4, sa4len ) != -1 );
```
pljones, line 332: Are other negative values "good" or should this be `>= 0`?

src/socket.cpp
```diff
@@ -318,20 +318,27 @@ void CSocket::Init ( const quint16  iNewPortNumber,
...
 
-            if ( UdpSocket6 != INVALID_SOCKET )
+            // only try to bind the IPv6 socket if IPv4 has succeeded and IPv6 socket is open
+            if ( bSuccess && UdpSocket6 != INVALID_SOCKET )
             {
-                bSuccess = bSuccess && ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) == 0 );
+                if ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) != -1 )
```
pljones, line 337: Why not
```
bIPv6Available = ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) != -1 );
```
(or `>= 0`) like IPv4?

Or even
```
bIPv6Available = bSuccess && UdpSocket6 != INVALID_SOCKET && ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) != -1 );
```

src/socket.cpp
```diff
@@ -341,23 +348,45 @@ void CSocket::Init ( const quint16  iNewPortNumber,
...
+                    bSuccess     = ( ::bind ( UdpSocket4, (struct sockaddr*) &sa4, sa4len ) != -1 );
+                }
 
-                if ( UdpSocket6 != INVALID_SOCKET )
+                // only try to bind the IPv6 socket if IPv4 has succeeded and IPv6 socket is open
+                if ( bSuccess && UdpSocket6 != INVALID_SOCKET && !bIPv6Available )
                 {
-                    bSuccess = bSuccess && ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) == 0 );
+                    sa6.sin6_port = htons ( port );
+                    if ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) != -1 )
```
pljones, line 379: As per the other call.

(In fact, aren't we duplicating code here?  Can it be stuffed into a method?)

src/socket.cpp
```diff
@@ -369,14 +398,31 @@ void CSocket::Init ( const quint16  iNewPortNumber,
...
 
-        bSuccess = ( ::bind ( UdpSocket4, (struct sockaddr*) &sa4, sa4len ) == 0 );
+        bSuccess = ( ::bind ( UdpSocket4, (struct sockaddr*) &sa4, sa4len ) != -1 );
 
-        if ( UdpSocket6 != INVALID_SOCKET )
+        // only try to bind the IPv6 socket if IPv4 has succeeded and IPv6 socket is open
+        if ( bSuccess && UdpSocket6 != INVALID_SOCKET )
         {
-            bSuccess = bSuccess && ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) == 0 );
+            if ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) != -1 )
```
pljones, line 406: And again.

src/socket.cpp
softins: According to `man 2 bind`:
```
RETURN VALUE
       On success, zero is returned.  On error, -1 is returned, and errno is set to indicate the error.
```
I tend to think of `-1` as being explicitly an error from system functions, which is why I changed that line.

Functions that only need to report success then usually return 0, and others such as `read` or `write` return a positive value for a byte count.

src/socket.cpp
softins: You could, but it just makes the code denser at the expense of clarity, and I doubt it saves more then a cycle or two. This is not the IOCCC!

src/socket.cpp
softins: Refactoring just one or two lines into a separate function that you then have to refer back to doesn't seem like much of a gain. Especially if you have to pass parameters to it.

src/socket.cpp
softins: Ditto! :)

src/socket.cpp
pljones: Given `On success, zero is returned.  On error, -1 is returned, and errno is set to indicate the error.` I'd do
```c++
switch ( ret ) {
  case 0:
    // whatever for working
    break;
  case -1:
    // whatever for handled error
    break;
  default:
    // something went seriously wrong with the call
}
```
and if that pattern occurs in several places, pass the three branches as lambdas.

The simplest requirement is that "success" is zero, anything else is not success.  Not "-1" is not necessarily zero.

src/socket.cpp
pljones: Yeah -- but
```
bIPv6Available = ( ::bind ( UdpSocket6, (struct sockaddr*) &sa6, sa6len ) != -1 );
```
at least keeps the two flows doing the equivalent work so it's easier for a reader.

(Yes, the `bSuccess && UdpSocket6 != INVALID_SOCKET &&` was going too far.)

src/socket.cpp
pljones: OK - if there's something different between the two, it's fine.

src/socket.cpp
softins: Nah, I'm not a fan of `switch` unless there are more than two or three cases that have specific meanings.

The Posix idiom for system calls is that `-1` specifically indicates error, and something else indicates success. When there's no other information to return, system calls return zero, but it's the lack of `-1` that indicates success. `bind()` in particular is defined as only returning `-1` or `0` because there's nothing else to say, but it's still `-1` that means error.

See the code example on the [bind manpage](https://man7.org/linux/man-pages/man2/bind.2.html). For every system call, the return value is compared with `-1`. That is the idiomatic way.

pljones: OK, I guess it's like `open()` -- you get the handle back, which _starts at zero_, or -1 for error, plus errno.

softins:
> I'd like to get this in before tagging beta3

Ok, so missed that, despite it being functional and working. :(

pljones:
We'll keep having betas until we're ready for a release candidate.  I don't like having 3.12.x releases with bug fixes without a tagged release with those same bug fixes: it's hard to keep track of.  In terms of bugs, 4.0.0beta3 = 3.12.5.  In terms of features 3.12.0 = 3.12.5.

softins:
> We'll keep having betas until we're ready for a release candidate. I don't like having 3.12.x releases with bug fixes without a tagged release with those same bug fixes: it's hard to keep track of. In terms of bugs, 4.0.0beta3 = 3.12.5. In terms of features 3.12.0 = 3.12.5.

Ah, so you're saying 4.0.0beta3 was just tagged to correspond with the bug fixes that went into 3.12.5? And this PR just missed out because it's not a candidate for backporting?

Anyway, hope we can get it in soon. I'm about to resolve hopefully your remaining unresolved comment.

pljones (approved):

pljones:
Looks very smooth now.

ann0see:
Will try to look at this later. 

I'll let @coderabbitai review this first. I think it's more concise than Claude usually.

ann0see:
> Docstring Coverage

Not too important

ann0see (approved):

ann0see:
Thank you very much! I think there might still be weird behavior in other parts of this code e.g. it might still block if we  start with -c on a somehow broken connection (Firewall dropping connection => jamulus doesn't start). However, I think it's out of scope here.

pljones:
OK, so I think CodeRabbit can be of use but we need to tweak some bits
- code style needs to be "in keeping with surrounding code".  If a file overall has no existing docstring coverage, that means generally they're not used and no one is expecting them, for example (same in another review)
- it should never overlook comments in the code from other reviewers - ideally just highlight that they've not been closed or report on what it found -- there's an unresolved comment here from @softins, for example.

Issue #3894, opened 2026-08-13
Raw audio should be declared, not inferred from packet size


mcfnord:
**🤖 AI:** Raw audio is detected by measuring the packet, not by asking the client. [`CServer::DecodeReceiveData`](https://github.com/jamulussoftware/jamulus/blob/b3c93588bb447d2c47987aba35284131a3e3c456/src/server.cpp#L976-L977) decides a stream is uncompressed PCM when `iCeltNumCodedBytes == sizeof(int16_t) * iClientFrameSizeSamples * channels`, and [the inverted send-side test](https://github.com/jamulussoftware/jamulus/blob/b3c93588bb447d2c47987aba35284131a3e3c456/src/server.cpp#L1266). That reserves four coded-byte counts — 256, 512, 128, and 256 again — which are also legal declarations for a *compressed* stream, since [`EvaluateNetwTranspPropsMes`](https://github.com/jamulussoftware/jamulus/blob/b3c93588bb447d2c47987aba35284131a3e3c456/src/protocol.cpp#L1505-L1509) accepts any size in [`CELT_MINIMUM_NUM_BYTES` 10, `MAX_SIZE_BYTES_NETW_BUF` 20000].

A client that declares `CT_OPUS` mono and sends real Opus at 256 coded bytes has its bitstream copied into the mix verbatim as int16 PCM. Now it's measured, with a purpose-built third-party client sending a 440 Hz sine, against a stock server started `-s -n -R`:

| declared coded bytes | server's own recording |
|---:|---|
| 255 (control) | 440.0 Hz, 100.0% of the energy within ±20 Hz, 0.0% above 5 kHz |
| 256 (collides) | no tone, peak bin 22500 Hz, 81.5% of the energy above 5 kHz |

Full-scale broadband noise. A separate run with a plain Opus listener joined as a second participant shows the mix carries it too, 77.6% of the energy above 5 kHz against 0.0% for the control, so this is not an artefact of how the recorder writes files. Both runs log the same `connected (1)` and nothing else — there is no error path, because a size test cannot fail.

All four reserved sizes behave the same way, and each one is a size the stock client itself emits — the declarations below were captured off the wire from the stock client, not derived:

| configuration | stock client declares | reserved | collide arm | N−1 control |
|---|---|---:|---|---|
[...]

PR #3898, opened 2026-08-13
[WIP] Use unused bytes in CNetworkTransportProps for EAudioQuality (backwards compatible)

### LINKED ISSUE #3896: 4 bytes unused in `CNetworkTransportProps`
`CNetworkTransportProps` contains 4 bytes reserved for `iAudioCodingArg` which is never used, still sent over the wire. This was surfaced by #3894  
We should use it to unambiguously negotiate the audio coding format and replace the heuristic method for recognising raw audio being employed at the moment.  
The question is what to do with the remaining bits. Any ideas?

dingodoppelt:
#3894 surfaced 4 bytes being unused in `CNetworkTransportProps`, namely `iAudioCodingArg`.
This uses those bytes to save the selected audio quality in the client. For backwards compatibility we simply leave the old checks in place. This should be improved, hence a work in progress.

CHANGELOG: Use iAudioCodingArg in CNetworkTransportProps for EAudioQuality

**Context: Fixes an issue?**

fixes #3896 

No, just a bug fix

Proof of concept (not to be merged soon);

- [ ] dedicated enum containing all possible audio quality settings combinations (or something similiar for the client and server to share)

## Checklist

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -1441,6 +1441,8 @@ void CClient::Init()
         }
     }
 
+    Channel.SetAudioCodingArg ( eAudioQuality );
+
     // calculate stereo (two channels) buffer size
     iStereoBlockSizeSam = 2 * iMonoBlockSizeSam;
 
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -183,6 +183,7 @@ CServer::CServer ( const int          iNewMaxNumChan,
     vecNumFrameSizeConvBlocks.Init ( iMaxNumChannels );
     vecUseDoubleSysFraSizeConvBuf.Init ( iMaxNumChannels );
     vecAudioComprType.Init ( iMaxNumChannels );
+    vecAudioCodingArg.Init ( iMaxNumChannels );
 
     for ( i = 0; i < iMaxNumChannels; i++ )
     {
@@ -844,6 +845,7 @@ void CServer::DecodeReceiveData ( const int iChanCnt, const int iNumClients )
     // get and store number of audio channels and compression type
     vecNumAudioChannels[iChanCnt] = vecChannels[iCurChanID].GetNumAudioChannels();
     vecAudioComprType[iChanCnt]   = vecChannels[iCurChanID].GetAudioCompressionType();
+    vecAudioCodingArg[iChanCnt]   = vecChannels[iCurChanID].GetAudioCodingArg();
 
     // get info about required frame size conversion properties
     vecUseDoubleSysFraSizeConvBuf[iChanCnt] = ( !bUseDoubleSystemFrameSize && ( vecAudioComprType[iChanCnt] == CT_OPUS ) );
@@ -974,7 +976,8 @@ void CServer::DecodeReceiveData ( const int iChanCnt, const int iNumClients )
             // sizeof ( int16_t ) is the size in bytes for the raw pcm audio data = 2
             // Sizes other than that are considered OPUS coded because those depend on hardcoded sizes in client.h
             const bool bIsRawAudio =
-                ( iCeltNumCodedBytes == static_cast<int> ( sizeof ( int16_t ) * iClientFrameSizeSamples * vecNumAudioChannels[iChanCnt] ) );
+                ( vecAudioCodingArg[iChanCnt] == AQ_RAW ||
+                  iCeltNumCodedBytes == static_cast<int> ( sizeof ( int16_t ) * iClientFrameSizeSamples * vecNumAudioChannels[iChanCnt] ) );
 
             const int iOffset = iB * SYSTEM_FRAME_SIZE_SAMPLES * vecNumAudioChannels[iChanCnt];
 
@@ -1263,7 +1266,8 @@ void CServer::MixEncodeTransmitData ( const int iChanCnt, const int iNumClients 
             DoubleFrameSizeConvBufOut[iCurChanID].GetAll ( vecsSendData, DOUBLE_SYSTEM_FRAME_SIZE_SAMPLES * vecNumAudioChannels[iChanCnt] );
         }
 
-        if ( iCeltNumCodedBytes != static_cast<int> ( sizeof ( int16_t ) * iClientFrameSizeSamples * vecNumAudioChannels[iChanCnt] ) )
+        if ( vecAudioCodingArg[iChanCnt] != AQ_RAW &&
+             iCeltNumCodedBytes != static_cast<int> ( sizeof ( int16_t ) * iClientFrameSizeSamples * vecNumAudioChannels[iChanCnt] ) )
         {
             // OPUS encoding
             if ( CurOpusEncoder != nullptr )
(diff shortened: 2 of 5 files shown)


ann0see:
You may break backwards compatibility to old betas IMO. There was no real release and we should try to have a clean state for each release.

dingodoppelt:
> You may break backwards compatibility to old betas IMO. There was no real release and we should try to have a clean state for each release.

Yes, I'm ok with breaking compatibility with beta clients, but this is an opportunity to overhaul the connection process even for non-raw qualities. We would break compatibility to all clients, if we only rely on those 4 bytes.

ann0see:
Ok. That's fair then.

dingodoppelt:
At the moment we collect our info from different enums (`EAudChanConf`, `EAudComprType`, `ENetwFlags`, `EAudioQuality`) that get used (or not) by client and server independently. We could use the free 4 bytes to create a consolidated source of information that client and server can share.

src/client.cpp
```diff
@@ -1441,6 +1441,8 @@ void CClient::Init()
         }
     }
 
+    Channel.SetAudioCodingArg ( eAudioQuality );
```
dingodoppelt, line 1444: This makes the server aware of the audio quality set in the client. Before that we had to guess by the packet sizes.

src/server.cpp
```diff
@@ -974,7 +976,8 @@ void CServer::DecodeReceiveData ( const int iChanCnt, const int iNumClients )
             // sizeof ( int16_t ) is the size in bytes for the raw pcm audio data = 2
             // Sizes other than that are considered OPUS coded because those depend on hardcoded sizes in client.h
             const bool bIsRawAudio =
-                ( iCeltNumCodedBytes == static_cast<int> ( sizeof ( int16_t ) * iClientFrameSizeSamples * vecNumAudioChannels[iChanCnt] ) );
+                ( vecAudioCodingArg[iChanCnt] == AQ_RAW ||
+                  iCeltNumCodedBytes == static_cast<int> ( sizeof ( int16_t ) * iClientFrameSizeSamples * vecNumAudioChannels[iChanCnt] ) );
```
dingodoppelt, line 984: I left this in for backwards compatibility. We could deprecate this with a `TODO: to be removed`, same for the transmitting side.

mcfnord:
**🤖 AI:** [Removed (see below)]

dingodoppelt:
> **🤖 AI:** The size test this replaces was doing two jobs, and only one of them is visible: `iCeltNumCodedBytes == sizeof(int16_t) * iClientFrameSizeSamples * vecNumAudioChannels` is true only when the byte count equals the destination frame exactly, so it was also the bounds check for [the memcpy below it](https://github.com/jamulussoftware/jamulus/blob/f7fbedd95493c0d1e1a63f26d800f362f512bbe3/src/server.cpp#L1000-L1003). As one arm of an `||`, [a declared AQ_RAW](https://github.com/jamulussoftware/jamulus/blob/f7fbedd95493c0d1e1a63f26d800f362f512bbe3/src/server.cpp#L982-L984) selects the raw path on its own while the copy length stays `iCeltNumCodedBytes` — the client's own `iBaseNetworkPacketSize`, [range-checked](https://github.com/jamulussoftware/jamulus/blob/f7fbedd95493c0d1e1a63f26d800f362f512bbe3/src/protocol.cpp#L1505-L1509) only to `[10, 20000]`. The destination is fixed at 512 bytes.
> 
> Now it's measured, ASan, server-only build, three UDP datagrams and no credentials:
> branch 	`iAudioCodingArg` 	declared size 	result
> `f7fbedd9` 	`AQ_RAW` 	2048 B 	heap-buffer-overflow, WRITE 2048 into 512
> `f7fbedd9` 	`AQ_RAW` 	20000 B 	WRITE of size 20000
> `f7fbedd9` 	`AQ_LOW` 	2048 B 	clean
> `main` `ea14cb94` 	`AQ_RAW` 	2048 B 	clean
> 
> Both write sites fire: the memcpy, and [the memset on a lost packet](https://github.com/jamulussoftware/jamulus/blob/f7fbedd95493c0d1e1a63f26d800f362f512bbe3/src/server.cpp#L1006-L1008). The declared flag is the trigger, not the packet size, and `main` is unaffected.
> 
> A path that keeps the direction: [iAudioCodingArg arrives unvalidated](https://github.com/jamulussoftware/jamulus/blob/f7fbedd95493c0d1e1a63f26d800f362f512bbe3/src/protocol.cpp#L1548-L1549) — no range check, unlike every field around it, which cost nothing while nothing read it. Validating it on receipt, and keeping the size equality as a requirement rather than an alternative, gives the explicit negotiation without letting the declaration widen a copy. Separately, `iAudioCodingArg` is initialised in neither the constructor nor `ResetNetworkTransportProperties()`.
> 
> The reproducer is a short Python script against a stock build; I can post it here or as a gist.

Of course we can break things if we want. The question is (and I'm not asking the AI) whether it works with a client built from main, nothing more. @mcfnord: If you have something to add personally feel free but please don't let the AI post stuff here. It isn't very helpful and its a lot of text I'd rather not have to read.

mcfnord:
> but please don't let the AI post stuff here. It isn't very helpful and its a lot of text I'd rather not have to read.

I have tried to slash the word count, but not sure that has made things easier to read. I generally respect the policy each person sets. When you say "I doubt that's true" my AI will respond with more evidence rather than less, so just be aware of that. When you say "here" do you mean on an unmerged PR of yours, or a draft? I had set it to avoid you, but last night I flushed all my own guidance and told it to go with things each individual has said about their wishes and preferences. So you can set your own policy. Perhaps you'd prefer "don't talk about my stuff more than once a day" or "wait 72 hours before saying anything about my activities" or "never write more words to me than the word count you're responding to." All are interesting throttle ideas.

ann0see:
No. He says that you should have probably not used AI at all. Or that the output is not concise enough.

ann0see:
Do we need 4 bytes for all audio quality combinations? Do we have something else which would be worth putting in there?

dingodoppelt:
> Do we need 4 bytes for all audio quality combinations? Do we have something else which would be worth putting in there?

We can put anything there and we're not limited to 4 bytes, either. Actually any message that doesn't carry data could be used.

src/server.cpp
```diff
@@ -964,6 +966,7 @@ void CServer::DecodeReceiveData ( const int iChanCnt, const int iNumClients )
                 pCurCodedData = nullptr;
             }
 
+            //### DEPRECATED: BEGIN ###//
```
ann0see, line 969: Not sure if we have this marker anywhere else but agree - we could introduce (and formalize) it.

ann0see (commented):
This changes the protocol philosophy somewhat - but is definitely cleaner. 
In general we should then probably check other areas of the protocol for empty spaces. 

CC: @softins for the protocol/network.

src/server.cpp
pljones: I don't think we should have open-ended deprecation any more.

State the version after which the feature will no longer _work_.  There's a lot of non-formally deprecated code, already -- stuff we sort of thought "Oh, with 4.x, we can throw this away"...  However, we've not come up with a formal schedule for dropping things.

I know we tend to drop platform support without forewarning people -- generally because we can't build for those platforms officially.  Dropping code is different, in my view (although it's still there if you `git reset --hard v3.12.0` or whatever).

PR #3899, opened 2026-08-13
Harden panning via bound check

ann0see:

Bounds pan values to always be in range. In theory, the network values could have been any value. Thus bounding those is needed.

CHANGELOG: Bug: Ensure correct bounds for panning

**Context: Fixes an issue?**

No. Found by @mcfnord via AI review.

No.

Ready for review (fast review!)

Review.

## Checklist

### DIFF
--- a/src/channel.cpp
+++ b/src/channel.cpp
@@ -337,7 +337,8 @@ void CChannel::SetPan ( const int iChanID, const float fNewPan )
     // set value (make sure channel ID is in range)
     if ( ( iChanID >= 0 ) && ( iChanID < MAX_NUM_CHANNELS ) )
     {
-        vecfPannings[iChanID] = fNewPan;
+        // pan range is 0..1; guard to avoid wrong values from the network
+        vecfPannings[iChanID] = std::min ( 1.0f, std::max ( fNewPan, 0.0f ) );
     }
 }
 


ann0see:
std::clamp would have been better - however, I believe it is not available unless we up the C++ version.

src/channel.cpp
```diff
@@ -337,7 +337,8 @@ void CChannel::SetPan ( const int iChanID, const float fNewPan )
     // set value (make sure channel ID is in range)
     if ( ( iChanID >= 0 ) && ( iChanID < MAX_NUM_CHANNELS ) )
     {
-        vecfPannings[iChanID] = fNewPan;
+        // pan range is 0..1; guard to avoid wrong values from the network
+        vecfPannings[iChanID] = std::min ( 1.0f, std::max ( fNewPan, 0.0f ) );
```
dingodoppelt, line 341: Looks like this could be achieved with [std::clamp](https://en.cppreference.com/cpp/algorithm/clamp)

dingodoppelt:
> std::clamp would have been better - however, I believe it is not available unless we up the C++ version.

Ok, I have missed this. We use c++17 for some things already. Maybe this is an opportunity to up the version?

ann0see:
> Maybe this is an opportunity to up the version?

Not in this PR. Float2Short could then also be made easier I think. But again, not in this PR.

src/channel.cpp
dingodoppelt: Or make this another member of `MathUtils` in case we need it elsewhere?

pljones (approved):

dingodoppelt (approved):

PR #3903, opened 2026-08-15
Remove "wc -l" from Jamulus.pro

pljones:

Currently, to get a build number for android builds, Jamulus.pro runs `wc -l` -- this is unlikely to work on Windows and isn't necessary to achieve the result intended.  This patch amends Jamulus.pro to use built-ins to do the calculation.

It also amends the build number so it is _aways zero on dev builds_.  This is so that only tagged builds have the number of commits counted.  (We're unlikely to have more commits on a release branch than on main.)

CHANGELOG: SKIP

**Context: Fixes an issue?**

Makes Jamulus.pro more portable.

No.

Tested locally, including on branch and tagged builds.

Should be good, subject to review.

## Checklist

### DIFF
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -251,7 +251,13 @@ win32 {
         ANDROID_ABIS = armeabi-v7a arm64-v8a x86 x86_64
     }
     ANDROID_VERSION_NAME = $$VERSION
-    ANDROID_VERSION_CODE = $$system(git log --oneline | wc -l)
+    # Used by Play Store
+    ANDROID_VERSION_CODE = 0
+    !contains(VERSION, .*dev.*) {
+        exists(".git/config") {
+            ANDROID_VERSION_CODE = $$size($$list($$system(git log --format=format:1,lines)))
+        }
+    }
     message("Setting ANDROID_VERSION_NAME=$${ANDROID_VERSION_NAME} ANDROID_VERSION_CODE=$${ANDROID_VERSION_CODE}")
 
     # liboboe requires C++17 for std::timed_mutex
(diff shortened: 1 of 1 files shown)


mcfnord:
**🤖 AI:** Two checks on this, run with Qt 5.15.13 qmake on Linux:

- `git rev-list --count HEAD` and `git log --oneline | wc -l` agree — both 6419 on current `main`.
- In a `git worktree` checkout, `.git` is a file rather than a directory, so `exists(".git/config")` is false and a non-dev build from a worktree silently gets `ANDROID_VERSION_CODE = 0`. `exists(".git")` behaves as intended in both layouts, since qmake's `exists()` matches files as well as directories.

A shadow build with the build directory outside the repository is unaffected either way: qmake resolves both `exists()` and `$$system()` against the source directory.

pljones:
> a non-dev build from a worktree

Mmm.  Does that include the Github autobuild workflow?  I think that does `git clone`.  The aim is that zero should appear anywhere that isn't a Github autobuild workflow for a release build, so the check wants to be as strict as possible, rather than in any way permissive.

mcfnord:
> ## Pull request overview
> Copilot reviewed 1 out of 1 changed files in this pull request and generated no new comments.

Then probably just say nothing next time.

ann0see (approved):
Looks ok. 

--
I think that it just needs to be an increasing number. As we don't publish it in the play store anyway (yet) how we calculate it does not matter.

pljones:
> Looks ok.
> 
> -- I think that it just needs to be an increasing number. As we don't publish it in the play store anyway (yet) how we calculate it does not matter.

Yep - this was part of what I was doing after 3.11.0 to get 3.12.0 ready for Play Store (but I was using Windows Qt Creator at the time, hence the build failing on the `wc -l`).

pljones:
Closing - will merge this into a larger rewrite.

PR #3921, opened 2026-08-21
Put UI header files in their own directory

softins:

Update `Jamulus.pro` to store the generated `ui_*.h` files in their own `ui` directory instead of letting them clutter up the project root.

CHANGELOG: Build: Use a separate ui directory for generated ui header files.

**Context: Fixes an issue?**

No, just helps keep the project root cleaner

No.

Tested and ready

Review

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ Makefile.*
 *.user.*
 *.o
 moc_*.cpp
+ui/
 ui_*.h
 moc_predefs.h
 .qm/
--- a/.gitignore
+++ b/.gitignore
@@ -17,9 +17,9 @@ Makefile.*
 *.user.*
 *.o
 moc_*.cpp
-ui/
-ui_*.h
 moc_predefs.h
+/ui/
+ui_*.h
 .qm/
 *.qm
 qmake_qmake_qm_files.qrc
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -380,6 +380,8 @@ win32 {
 # It doesn't work with multiple targets or architectures.
 RESOURCES += src/resources.qrc
 
+UI_DIR += ui
+
 FORMS_GUI = src/aboutdlgbase.ui \
     src/serverdlgbase.ui
 
@@ -1213,9 +1215,9 @@ android {
     for (abi, ANDROID_ABIS) {
         DISTCLEAN_DIRS += debug-$${abi} release-$${abi}
     }
-    DISTCLEAN_DIRS += .qm
+    DISTCLEAN_DIRS += ui .qm
 } else {
-    DISTCLEAN_DIRS += debug release .qm
+    DISTCLEAN_DIRS += debug release ui .qm
 }
 
 win32 {
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -381,7 +381,7 @@ win32 {
 RESOURCES += src/resources.qrc
 
 # store ui_*.h files in a directory instead of project root
-UI_DIR += ui
+UI_DIR = ui
 
 # A tree built in place before UI_DIR was set still has ui_*.h in its root, and
 # they must be deleted before creating the new Makefiles.
@@ -1223,9 +1223,9 @@ android {
     for (abi, ANDROID_ABIS) {
         DISTCLEAN_DIRS += debug-$${abi} release-$${abi}
     }
-    DISTCLEAN_DIRS += ui .qm
+    DISTCLEAN_DIRS += $$UI_DIR .qm
 } else {
-    DISTCLEAN_DIRS += debug release ui .qm
+    DISTCLEAN_DIRS += debug release $$UI_DIR .qm
 }
 
 win32 {
--- a/Jamulus.pro
+++ b/Jamulus.pro
@@ -380,7 +380,15 @@ win32 {
 # It doesn't work with multiple targets or architectures.
 RESOURCES += src/resources.qrc
 
+# store ui_*.h files in a directory instead of project root
 UI_DIR += ui
+
+# A tree built in place before UI_DIR was set still has ui_*.h in its root, and
+# they must be deleted before creating the new Makefiles.
+STALE_UI_HEADERS = $$files(ui_*.h)
+for (stale, STALE_UI_HEADERS) {
+    system($$QMAKE_DEL_FILE $$stale)
+}
 
 FORMS_GUI = src/aboutdlgbase.ui \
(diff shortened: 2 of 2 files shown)


mcfnord:
**🤖 AI:** In a tree that has already built in place, the `ui_*.h` left in the project root take precedence over the new ones: `INCPATH` lists `-I.` before `-Iui`, and qmake's dependency scan binds to the root copy — after `qmake` on this branch, all 76 generated-header dependency entries in `Makefile.Release` name the root files, none name `ui/`. A `.ui` edit then recompiles nothing, and the linked binary still carries the pre-edit strings while the regenerated `ui/ui_aboutdlgbase.h` carries the new one (Linux, Qt 5.15.13, g++). The same rebinding and the same `-I.`/`-Iui` order appear on macOS with Qt 5.15.2 and clang. Pinned at e13e0130.

Nothing clears the leftovers now: `compiler_uic_clean` lists only `ui/ui_*.h`, so `make clean` and `make distclean` skip them, and `.gitignore` keeps them out of `git status`. One `rm ui_*.h` fixes a tree permanently — worth a line in the PR description or the CHANGELOG entry, since in-place builds are exactly what this declutters.

Reproduce: build `main` in place, check out this branch, run `qmake`, then `grep -cE '^[[:space:]]+ui_[a-z]+\.h' Makefile.Release` gives 76 and the same pattern with `ui/ui_` gives 0; on a fresh clone the two counts are 0 and 76.

Also `ui/` in `.gitignore` is unanchored, so it ignores any nested `ui` directory, `src/ui/` included; `/ui/` would scope it to the build output.

softins:
OK thanks. I'll look at that soon and see if I can make it more robust to switching between branches with and without this change.

Jamulus.pro
```diff
@@ -380,6 +380,8 @@ win32 {
 # It doesn't work with multiple targets or architectures.
 RESOURCES += src/resources.qrc
 
+UI_DIR += ui
```
mcfnord, line 383: **🤖 AI:** Self-healing form of that cleanup — qmake deletes the leftovers while the project file is parsed, before it resolves the includes:

```suggestion
UI_DIR += ui

# A tree built in place before UI_DIR was set still has ui_*.h in its root, and
# those win: INCPATH lists the source and build roots ahead of -I$$UI_DIR, and
# qmake resolves the include when it writes the Makefile, so the stale copy
# stays bound to every object until it is deleted.
STALE_UI_HEADERS = $$files($$PWD/ui_*.h) $$files($$OUT_PWD/ui_*.h)
STALE_UI_HEADERS = $$unique(STALE_UI_HEADERS)
for (stale, STALE_UI_HEADERS) {
    message("Removing stale generated UI header: $$stale")
    system($$QMAKE_DEL_FILE $$shell_quote($$shell_path($$stale)))
}
```

Both roots go to [`$$files()`](https://doc.qt.io/qt-5/qmake-function-reference.html#files-pattern-recursive-false) because a shadow build lists the source root ahead of even `-I.`, so a source tree left dirty by an old in-place build shadows an otherwise clean build directory; `$$unique()` keeps the in-place case, where the two paths are the same, from deleting twice.

On a tree that had already built in place, one `qmake` prints 6 removals and moves all 76 generated-header dependency entries in `Makefile.Release` from `ui_*.h` to `ui/ui_*.h`, and an edit to [`src/aboutdlgbase.ui`](https://github.com/jamulussoftware/jamulus/blob/e13e0130aa116d928b6a02935e940f888ed35c1a/src/aboutdlgbase.ui#L59) then reaches the linked binary. On a fresh checkout nothing is removed and the counts are unchanged. Back on `main` there is no `-Iui`, so a leftover `ui/` is never searched.

Linux Qt 5.15.13 / g++ and macOS Qt 5.15.2 / clang. Windows was not exercised: the [`system()`](https://doc.qt.io/qt-5/qmake-function-reference.html#system-command-mode-stsvar) call running `$$QMAKE_DEL_FILE` is the part to check there.

pljones, line 383: If this is going to be a variable, best to use it in `DISTCLEAN_DIRS` below to prevent drift.

mcfnord: **🤖 AI:** Agreed, and it costs nothing: replacing the literal in [both `DISTCLEAN_DIRS` lines](https://github.com/jamulussoftware/jamulus/blob/e13e0130aa116d928b6a02935e940f888ed35c1a/Jamulus.pro#L1218-L1220) with `$$UI_DIR` leaves [the `rm -rf` command](https://github.com/jamulussoftware/jamulus/blob/e13e0130aa116d928b6a02935e940f888ed35c1a/Jamulus.pro#L1229) byte-identical — `rm -rf debug release ui .qm` — for an in-place build and a shadow build alike, at e13e0130 with Qt 5.15.13.

The two forms can only diverge when `UI_DIR` is passed on the qmake command line, and that case is already broken by the `+=` in [`UI_DIR += ui`](https://github.com/jamulussoftware/jamulus/blob/e13e0130aa116d928b6a02935e940f888ed35c1a/Jamulus.pro#L383): the override becomes the two-element list `build-ui ui`, uic then generates nothing, and `Makefile.Release` goes from 76 generated-header dependency entries to 0. Spelling it `UI_DIR = ui` keeps such an override working and makes `$$UI_DIR` exact rather than coincidental.

.gitignore
```diff
@@ -17,6 +17,7 @@ Makefile.*
 *.user.*
 *.o
 moc_*.cpp
+ui/
```
pljones: Maybe
```
**/ui_*.h
```
and condense the two lines?

pljones: Also, keep the two `moc_... ` lines adjacent?

softins: I think that will just hide the files, not the directory. Really, only /ui/ is needed, and not ui_*.h

softins: > Also, keep the two `moc_... ` lines adjacent? 

To be honest, we could do with considering .gitignore as a whole and raising a cleanup PR, rather than including unrelated but worthwhile changes in this PR.

pljones: True, it was only because this was splitting them further and swapping two lines here was trivial. 
I'm happy as it is, tho.

pljones (approved):

softins:
Tested on both Linux and Windows, switching between `main` and `ui-dir` in both directions without an intermediate `make distclean`. And in `ui-dir`, `make distclean`/`nmake distclean` works correctly on both systems.

pljones (approved):

dingodoppelt (approved):

PR #3922, opened 2026-08-22
Externalise cache keys to prevent unneeded invalidation

### LINKED ISSUE #2462: Add caching for iOS build
No

Since https://github.com/jamulussoftware/jamulus/pull/2355 has been merged, we should also use caching/parallelising for the iOS build. 

/

pljones:

I noticed whilst I've been working on `autobuild.sh` that _every change_ invalidates _all_ the caches.  Because `autobuild.sh` itself is included in the cache key calculation.  This change replaces that mechanism on macOS and Windows with am externalised list of dependencies.  It adds the same mechanism to iOS, which lacked caching.

Github cache clean up will remove unused cache entries with no need to have the cache consumer included in the cache key.

(My work in progress on the Android build uses the same approach, which is where this comes from.)

Scope, guidance, testing and review by myself, coding my VSCode Github CoPilot.

CHANGELOG: Build: improve cache persistence

**Context: Fixes an issue?**

Caching should depend on what is being cached, rather than the cache consumer.

No.

Tested against the current branch in my own repository before raising the PR here.

Need to do some more testing.

## Checklist

### DIFF
--- a/.github/autobuild/ios-dependencies.sh
+++ b/.github/autobuild/ios-dependencies.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# Values are consumed by ios.sh and the dependency cache key.
+# shellcheck disable=SC2034
+AQTINSTALL_VERSION="${AQTINSTALL_VERSION:-3.3.0}"
+QT_VERSION="${QT_VERSION:-5.15.2}"
--- a/.github/autobuild/ios.sh
+++ b/.github/autobuild/ios.sh
@@ -48,10 +48,12 @@
 
 set -eu
 
+PROJECT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+readonly PROJECT_DIR
+# shellcheck disable=SC1091
+source "${PROJECT_DIR}/.github/autobuild/ios-dependencies.sh"
+
 QT_DIR=/opt/qt
-# The following version pinnings are semi-automatically checked for
-# updates. Verify .github/workflows/bump-dependencies.yaml when changing those manually:
-AQTINSTALL_VERSION=3.3.0
 
 if [[ ! ${QT_VERSION:-} =~ [0-9]+\.[0-9]+\..* ]]; then
     echo "Environment variable QT_VERSION must be set to a valid Qt version"
@@ -63,7 +65,7 @@ if [[ ! ${JAMULUS_BUILD_VERSION:-} =~ [0-9]+\.[0-9]+\.[0-9]+ ]]; then
 fi
 
 setup() {
-    if [[ -d "${QT_DIR}" ]]; then
+    if [[ -x "${QT_DIR}/${QT_VERSION}/ios/bin/qmake" ]]; then
         echo "Using Qt installation from previous run (actions/cache)"
     else
         echo "Installing Qt"
--- a/.github/autobuild/mac.sh
+++ b/.github/autobuild/mac.sh
@@ -48,10 +48,12 @@
 
 set -eu
 
+PROJECT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+readonly PROJECT_DIR
+# shellcheck disable=SC1091
+source "${PROJECT_DIR}/.github/autobuild/mac-dependencies.sh"
+
 QT_DIR=~/qt
-# The following version pinnings are semi-automatically checked for
-# updates. Verify .github/workflows/bump-dependencies.yaml when changing those manually:
-AQTINSTALL_VERSION=3.3.0
 
 TARGET_ARCHS="${TARGET_ARCHS:-}"
 
@@ -65,7 +67,8 @@ if [[ ! ${JAMULUS_BUILD_VERSION:-} =~ [0-9]+\.[0-9]+\.[0-9]+ ]]; then
 fi
 
 setup() {
-    if [[ -d "${QT_DIR}" ]]; then
+    if [[ -x "${QT_DIR}/${QT_VERSION}/macos/bin/qmake" &&
+        -x "${QT_DIR}/${QT_VERSION}/macos/bin/macdeployqt" ]]; then
         echo "Using Qt installation from previous run (actions/cache)"
     else
         echo "Installing Qt..."
--- a/.github/autobuild/windows-dependencies.ps1
+++ b/.github/autobuild/windows-dependencies.ps1
@@ -1,0 +1,8 @@
+# Values are consumed by windows.ps1 and the dependency cache key.
+$Qt32Version = "5.15.2"
+$Qt64Version = "6.10.2"
+$AqtinstallVersion = "3.3.0"
+$JackVersion = "1.9.22"
+$Msvc32Version = "win32_msvc2019"
+$Msvc64Version = "win64_msvc2022_64"
+$JomVersion = "1.1.2"
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -230,6 +230,7 @@ jobs:
             # Disable CodeQL on mac as it interferes with signing the binaries (signing hangs, see #2563 and #2564)
             run_codeql:             false
             xcode_version:          16.3.0
+            dependency_cache_key:   qt6
             is_main_build_target:   true
 
           # Reminder: If Legacy is removed, be sure to add a dedicated job for CodeQL again.
@@ -243,6 +244,7 @@ jobs:
             # https://developer.apple.com/support/xcode/
             # https://xcodereleases.com/
             xcode_version:          16.3.0
+            dependency_cache_key:   qt5
             is_main_build_target:   true
 
           - config_name:            iOS (artifacts)
@@ -254,6 +256,7 @@ jobs:
             # Therefore, disable it:
             run_codeql:             false
             xcode_version:          26.2
+            dependency_cache_key:   ios
 
           - config_name:            Windows (artifact+codeQL)
             target_os:              windows
@@ -267,6 +270,7 @@ jobs:
             building_on_os:         windows-2025
             base_command:           powershell .\.github\autobuild\windows.ps1 -BuildOption jackonwindows -Stage
             run_codeql:             false
+            dependency_cache_key:   jack
 
         # This injects the build_all_targets information into each matrix output:
         build_all_targets:
@@ -300,7 +304,14 @@ jobs:
           path: |
             ~/qt
             ~/Library/Cache/jamulus-dependencies
-          key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/workflows/autobuild.yml', '.github/autobuild/mac.sh', 'mac/deploy_mac.sh') }}-${{ matrix.config.base_command }}
+          key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/autobuild/mac-dependencies.sh') }}-${{ matrix.config.dependency_cache_key }}
+
+      - name:                       Cache iOS dependencies
+        if:                         matrix.config.target_os == 'ios'
+        uses:                       actions/cache@v6
+        with:
+          path: /opt/qt
+          key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/autobuild/ios-dependencies.sh') }}-${{ matrix.config.dependency_cache_key }}
 
       - name:                       Cache Windows dependencies
         if:                         matrix.config.target_os == 'windows'
@@ -312,7 +323,7 @@ jobs:
             C:\AutobuildCache
             ${{ github.workspace }}\libs\NSIS\NSIS-source
             ${{ github.workspace }}\libs\ASIOSDK2
-          key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/workflows/autobuild.yml', '.github/autobuild/windows.ps1', 'windows/deploy_windows.ps1') }}-${{ matrix.config.base_command }}
+          key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/autobuild/windows-dependencies.ps1', 'windows/deploy_windows.ps1') }}-${{ matrix.config.dependency_cache_key || 'standard' }}
 
       - name:                       Cache Android dependencies
         if:                         matrix.config.target_os == 'android'
--- a/.github/workflows/bump-dependencies.yml
+++ b/.github/workflows/bump-dependencies.yml
@@ -39,8 +39,8 @@ jobs:
           - name: aqt
             # not Changelog-worthy
             get_upstream_version: GH_REPO=miurahr/aqtinstall gh release view --json tagName --jq .tagName | sed -re 's/^v//'
-            # The following regexps capture both the *nix and the Windows variable syntax (different case, underscore):
-            local_version_regex: (.*AQTINSTALL_?VERSION\s*=\s*"?)([0-9.]*)("?.*)
+            # Capture both plain assignments and shell parameter-expansion defaults.
+            local_version_regex: (.*AQTINSTALL_?VERSION\s*=\s*"?[^0-9]*)([0-9.]+)(.*)
           - name: create-dmg
             changelog_name: create-dmg (macOS)
             get_upstream_version: GH_REPO=create-dmg/create-dmg gh release view --json tagName --jq .tagName | sed -re 's/^v//'
@@ -50,8 +50,8 @@ jobs:
             get_upstream_version: |
               latest_minor="$(curl -s https://download.qt.io/official_releases/qt/ | grep -oP 'href="\K[0-9.]+(?=/")' | sort --reverse --version-sort | head -n1)";
               curl -s https://download.qt.io/official_releases/qt/"${latest_minor}"/ | grep -oP 'href="\K[0-9.]+(?=/")' | sort --reverse --version-sort | head -n1
-            # The following regex captures both the *nix and the Windows variable syntax (different case, underscore):
-            local_version_regex: (.*QT[0-9_]+VERSION\s*=\s*"?)(6\.[0-9.]+)("?.*)
+            # Capture both plain assignments and shell parameter-expansion defaults.
+            local_version_regex: (.*QT[0-9_]+VERSION\s*=\s*"?[^0-9]*)(6\.[0-9.]+)(.*)
 
           - name: jack
             changelog_name: bundled JACK (Windows-only)
--- a/.github/autobuild/mac-dependencies.sh
+++ b/.github/autobuild/mac-dependencies.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+# Values are consumed by mac.sh and the dependency cache key.
+# shellcheck disable=SC2034
+AQTINSTALL_VERSION="${AQTINSTALL_VERSION:-3.3.0}"
+QT_VERSION="${QT_VERSION:-6.10.2}"
+QT_LEGACY_VERSION=5.15.2
--- a/.github/autobuild/windows.ps1
+++ b/.github/autobuild/windows.ps1
@@ -68,15 +68,7 @@ $ProgressPreference = 'SilentlyContinue'
 $QtDir = 'C:\Qt'
 $ChocoCacheDir = 'C:\ChocoCache'
 $DownloadCacheDir = 'C:\AutobuildCache'
-# The following version pinnings are semi-automatically checked for
-# updates. Verify .github/workflows/bump-dependencies.yaml when changing those manually:
-$Qt32Version = "5.15.2"
-$Qt64Version = "6.10.2"
-$AqtinstallVersion = "3.3.0"
-$JackVersion = "1.9.22"
-$Msvc32Version = "win32_msvc2019"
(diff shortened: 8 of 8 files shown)


.github/workflows/autobuild.yml
```diff
@@ -312,7 +323,7 @@ jobs:
             C:\AutobuildCache
             ${{ github.workspace }}\libs\NSIS\NSIS-source
             ${{ github.workspace }}\libs\ASIOSDK2
-          key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/workflows/autobuild.yml', '.github/autobuild/windows.ps1', 'windows/deploy_windows.ps1') }}-${{ matrix.config.base_command }}
+          key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/autobuild/windows-dependencies.ps1') }}-${{ matrix.config.dependency_cache_key || 'standard' }}
```
mcfnord: **🤖 AI:** `$AsioSDKUrl` and `$NsisUrl` are pinned in [`windows/deploy_windows.ps1`](https://github.com/jamulussoftware/jamulus/blob/753220d4134da27223b162e0a7ded5021cc04b5a/windows/deploy_windows.ps1#L60-L61), which this key no longer hashes, and [the bump workflow rewrites that file](https://github.com/jamulussoftware/jamulus/blob/753220d4134da27223b162e0a7ded5021cc04b5a/.github/workflows/bump-dependencies.yml#L94). Both unpack to fixed, unversioned destinations this step caches, `libs\NSIS\NSIS-source` and `libs\ASIOSDK2`, and [`Install-Dependency` returns early](https://github.com/jamulussoftware/jamulus/blob/753220d4134da27223b162e0a7ded5021cc04b5a/windows/deploy_windows.ps1#L178-L182) when the destination exists. So after a bump the key is unchanged, the stale tree is restored, and the download is skipped.

```suggestion
          key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/autobuild/windows-dependencies.ps1', 'windows/deploy_windows.ps1') }}-${{ matrix.config.dependency_cache_key || 'standard' }}
```

Moving both URLs into `windows-dependencies.ps1` instead would need `deploy_windows.ps1` to source it: [windows.ps1 launches it as its own process](https://github.com/jamulussoftware/jamulus/blob/753220d4134da27223b162e0a7ded5021cc04b5a/.github/autobuild/windows.ps1#L234), so its param defaults are what standalone builds get.

pljones: Good spot -- I've not done enough Windows builds.  Fix applied.  The caches should get reused for macOS and iOS on this build.

pljones: Yep, Windows caches invalidated by the new hash check, macOS and iOS saved.

mcfnord: **🤖 AI:** Android is the one cache key this PR leaves hashing the workflow file. [The macOS key drops it](https://github.com/jamulussoftware/jamulus/blob/fdae3db0f2498add06d252767807b6bcedb3e8aa/.github/workflows/autobuild.yml#L308) and [the Windows key drops it](https://github.com/jamulussoftware/jamulus/blob/fdae3db0f2498add06d252767807b6bcedb3e8aa/.github/workflows/autobuild.yml#L327); [the Android key](https://github.com/jamulussoftware/jamulus/blob/fdae3db0f2498add06d252767807b6bcedb3e8aa/.github/workflows/autobuild.yml#L337) still reads `hashFiles('.github/workflows/autobuild.yml', '.github/autobuild/android.sh')`. Every version that decides the content of those cached paths is [already in `android.sh`](https://github.com/jamulussoftware/jamulus/blob/fdae3db0f2498add06d252767807b6bcedb3e8aa/.github/autobuild/android.sh#L53-L58) — command-line tools, NDK, platform, build-tools, aqtinstall, Qt — and the Android entry passes none of its own, its `base_command` being a bare `./.github/autobuild/android.sh`, so hashing the workflow there can only invalidate, never protect.

This PR's three completed Auto-Build runs produced three distinct Android keys, 1.47 GB each, and restored none of them: `Cache Android dependencies` starts and finishes in the same second on all three, while the post step saves on all three, which [an exact match never does](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching). That is 4.42 GB stored for a cache that has not been read once.

Dropping `'.github/workflows/autobuild.yml'` from that one key is the whole fix.

pljones: As stated, I'm in the middle of rewriting the Android build.

.github/workflows/autobuild.yml
```diff
@@ -300,7 +305,14 @@ jobs:
           path: |
             ~/qt
             ~/Library/Cache/jamulus-dependencies
-          key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/workflows/autobuild.yml', '.github/autobuild/mac.sh', 'mac/deploy_mac.sh') }}-${{ matrix.config.base_command }}
+          key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/autobuild/mac-dependencies.sh') }}-${{ matrix.config.dependency_cache_key }}
```
softins: This hash still needs to contain `mac/deploy_mac.sh`, because that script contains `CREATEDMG_VERSION`, which is updated by `bump-dependencies`

.github/workflows/bump-dependencies.yml
```diff
@@ -39,8 +39,8 @@ jobs:
           - name: aqt
             # not Changelog-worthy
             get_upstream_version: GH_REPO=miurahr/aqtinstall gh release view --json tagName --jq .tagName | sed -re 's/^v//'
-            # The following regexps capture both the *nix and the Windows variable syntax (different case, underscore):
-            local_version_regex: (.*AQTINSTALL_?VERSION\s*=\s*"?)([0-9.]*)("?.*)
+            # Capture both plain assignments and shell parameter-expansion defaults.
+            local_version_regex: (.*AQTINSTALL_?VERSION\s*=\s*"?[^0-9]*)([0-9.]+)(.*)
```
softins: This regex and the one below look ok to me by eye, but have you tested them in situ somehow?

.github/workflows/bump-dependencies.yml
```diff
@@ -50,8 +50,8 @@ jobs:
             get_upstream_version: |
               latest_minor="$(curl -s https://download.qt.io/official_releases/qt/ | grep -oP 'href="\K[0-9.]+(?=/")' | sort --reverse --version-sort | head -n1)";
               curl -s https://download.qt.io/official_releases/qt/"${latest_minor}"/ | grep -oP 'href="\K[0-9.]+(?=/")' | sort --reverse --version-sort | head -n1
-            # The following regex captures both the *nix and the Windows variable syntax (different case, underscore):
-            local_version_regex: (.*QT[0-9_]+VERSION\s*=\s*"?)(6\.[0-9.]+)("?.*)
+            # Capture both plain assignments and shell parameter-expansion defaults.
+            local_version_regex: (.*QT[0-9_]+VERSION\s*=\s*"?[^0-9]*)(6\.[0-9.]+)(.*)
```
softins: See comment above

softins (commented):
Looks ok in general, just a couple of comments.

.github/workflows/bump-dependencies.yml
pljones: It runs... But I'm not sure how to check it actually picks up new versions.  It's syntactically stronger...

I'd have to mock the returns from the version probes, I guess, and check it creates the bump.  I'll give that a go in my repo.

.github/workflows/autobuild.yml
mcfnord: **🤖 AI:** Confirmed, and it is the only one of its kind here. Running all seven bump components against this head, every component moves the cache key of each platform whose version it changes — except create-dmg, whose single pin is `CREATEDMG_VERSION` in `mac/deploy_mac.sh`, [which `main` hashes today](https://github.com/jamulussoftware/jamulus/blob/b86473ac612bf59293b4196b69ea0ce7a500f222/.github/workflows/autobuild.yml#L303).

The failure mode is milder than the ASIO/NSIS one: create-dmg unpacks into [a version-namespaced folder](https://github.com/jamulussoftware/jamulus/blob/4b5180407b45cddcbf92a8b0c82320e4b22e5a4a/mac/deploy_mac.sh#L244), so a stale cache cannot install the wrong version. The cost is that after a bump the key is unchanged, the exact hit means the post step saves nothing, and create-dmg is downloaded and `make install`ed again on every macOS run.

```suggestion
          key:                      ${{ matrix.config.target_os }}-${{ hashFiles('.github/autobuild/mac-dependencies.sh', 'mac/deploy_mac.sh') }}-${{ matrix.config.dependency_cache_key }}
```

pljones:
Well, it's not quite there, is it....

mcfnord:
**🤖 AI:** The four former `param()` defaults at [`deploy_windows.ps1`](https://github.com/jamulussoftware/jamulus/blob/3f7be0150e3ddcf29f24863a89c6214231ef10f1/windows/deploy_windows.ps1#L71-L76) kept their trailing commas and [`deploy_mac.sh`](https://github.com/jamulussoftware/jamulus/blob/3f7be0150e3ddcf29f24863a89c6214231ef10f1/mac/deploy_mac.sh#L52) kept the `../..` that suits [a script one level deeper](https://github.com/jamulussoftware/jamulus/blob/3f7be0150e3ddcf29f24863a89c6214231ef10f1/.github/autobuild/mac.sh#L52), and two more sit behind that Windows parse error: [the launch line](https://github.com/jamulussoftware/jamulus/blob/3f7be0150e3ddcf29f24863a89c6214231ef10f1/.github/autobuild/windows.ps1#L238) still passes the two Qt paths positionally to a `param()` that no longer declares them, and [`$QtCompile64`](https://github.com/jamulussoftware/jamulus/blob/3f7be0150e3ddcf29f24863a89c6214231ef10f1/.github/autobuild/windows-dependencies.ps1#L8) is now `msvc2022` while [the aqt arch](https://github.com/jamulussoftware/jamulus/blob/3f7be0150e3ddcf29f24863a89c6214231ef10f1/.github/autobuild/windows.ps1#L75) is still `win64_${QtCompile64}_64`.

pljones:
Well, the cache keys and cache creation are working now.  ios appeared to be a bit flaky - consistently failing to create the cache.  I did a straight rerun under debug and... it worked.  So I don't know why it wasn't working!

I've yet to test bump-dependencies with the new scheme.  It actually makes it a lot simpler: it's just maintaining the version files based on the available version numbers.  Unfortunately I've not managed to get it to run at all on my repo (even for main).

pljones:
So I reran the iOS build to see if it used the cache:
```
/opt/homebrew/bin/gtar: ../../../../../opt/qt: Cannot mkdir: Permission denied
```
Oh well.  But it does give useful information --  the other runs don't stick Qt under /opt, they use a project work directory.  I'll see if I can get that working here, too.

pljones:
OK, iOS is working and bump-dependencies is working - with the caveat that it now tries to bump Windows Qt to the same version as macOS Qt6.  I think this is a good thing: we shouldn't use different versions of the same basic API.  (Qt5 and Qt6 bumps are separate -- if Qt ever do a Qt5 patch, it'll get picked up, too.)

pljones:
Getting there -- the caches are being made and restored.  Windows and macOS don't realise this - they go off and install again...

iOS gets it right.
```
Cache hit for: ios-0f7a19096b2d214a9d0ef45f70b89972b1f10eec3e4f6ae85a49f4dbcf9925d3-ios
```
```
Using Qt installation from previous run (actions/cache)
```

Bump dependencies is working, too.

pljones:
Right - all tests now passing:
1. Cache keys created for iOS, macOS qt5, macOS qt6, Windows ASIO, Windows Jack
1. Caches populated on successful completion for each key
1. Caches restored on subsequent run **on same branch** (that's just how the cache works)
1. Caches _used_ by the matrix `setup` runner
1. Bump Dependencies correctly spots version matches and mismatches
   * One caveat: as we currently have divergent macOS and Windows Qt6 versions, it raises one PR to bring them together
   * I've only checked the diff for the PR its created - https://github.com/pljones/jamulus/pull/70/changes/812e9d708e5e12882150a02305d4d7334fa5a2be

.github/autobuild/ios-dependencies.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
```
ann0see, line 4: This then dropped the comment about pinning?

.github/autobuild/mac.sh
```diff
@@ -48,10 +48,13 @@
 
 set -eu
 
+QT=${QT:-6}
+PROJECT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+readonly PROJECT_DIR
+# shellcheck disable=SC1090
+source "${PROJECT_DIR}/.github/autobuild/mac-dependencies_qt${QT}.sh"
```
ann0see, line 55: Do we have a guarantee that ${QT} is safe and correct? I'd add an error check if the file exists before.

.github/autobuild/windows-dependencies.ps1
```diff
@@ -0,0 +1,19 @@
...
+$QtCompile32 = "msvc2019"
+$QtCompile64 = "msvc2022"
+$AqtinstallVersion = "3.3.0"
+$JackVersion = "1.9.22"
+$JomVersion = "1.1.2"
+
+# Important:
+# - Do not update ASIO SDK without checking for license-related changes.
+# - Do not copy (parts of) the ASIO SDK into the Jamulus source tree without
+#   further consideration as it would make the license situation more complicated.
```
ann0see, line 16: Probably outdated as they moved to the GPL.

ann0see:
I think we might have an issue about some OS not being cached. Worth checking the issues for that and linking it here. If the issue is iOS related, we can close it.

.github/autobuild/mac.sh
pljones: It's set from the autobuild workflow file and, if not, defaults to 6 at the top of the file.  If someone sets the wrong value, this command will safely crash the workflow.

.github/autobuild/ios-dependencies.sh
pljones: Eh?

.github/autobuild/windows-dependencies.ps1
pljones: As it's version dependent (if the old version can still be found), I suppose it makes sense to keep it.  If and when the old version vanishes, this can go, too.

ann0see:
The iOS caching issue was already documented. Thus:

Fixes: https://github.com/jamulussoftware/jamulus/issues/2462

.github/autobuild/windows-dependencies.ps1
ann0see: The old version will probably not vanish. But we can of course keep this.

.github/autobuild/mac.sh
ann0see: I'd still like a guard to reject invalid values.

.github/autobuild/ios.sh
```diff
@@ -48,10 +48,12 @@
 
 set -eu
 
-QT_DIR=/opt/qt
-# The following version pinnings are semi-automatically checked for
-# updates. Verify .github/workflows/bump-dependencies.yaml when changing those manually:
```
ann0see, line 53: @pljones Maybe I am blind, but I think this comment got dropped?

pljones: Oh, right.  I should have put that on all the `*-dependencies*`.

.github/autobuild/mac.sh
ann0see: @coderabbitai

pljones: No, it makes a claim that's unjustified.  If there were a `mac-dependencies_qtFOO.sh` and you passed `QT=FOO`, that would be valid.  The range of valid values is defined _only_ by the list of files matching the pattern `mac-dependencies_qt(.+)\.sh` in the `.githhub/autobuild` directory.  It could `[[ -r ...path${QT}.sh ]] || { echo "Bad value for QT - look in .githhub/autobuild"; exit 1; }`.  But that's essentially what "file not found" tells you.

pljones:
> Suppress SC2034 for QT_VERSION.
> 
> QT_VERSION is consumed after this file is sourced. ShellCheck cannot observe that use. The directive on Line 6 applies only to AQTINSTALL_VERSION, so Line 8 reports an unused assignment. Add an SC2034 directive immediately before QT_VERSION.

Demonstrably false from the shellcheck output.

> The workflow is named .github/workflows/bump-dependencies.yml. Both comments point to a nonexistent .yaml path.

Ah!

> Anchor the ASIO-SDK regex at the start of the line.

Should never be an issue but good catch.

.github/workflows/bump-dependencies.yml
```diff
@@ -91,9 +90,12 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           set -eu
-          files=( .github/{autobuild,workflows}/* windows/*.ps1 mac/*.sh )
+          files=( .github/autobuild/*-dependencies{,_qt[56]}.* )
+          echo "files: (${files[@]})"
           upstream_version="$(${{ matrix.components.get_upstream_version }})"
-          local_version="$(perl -nle 'print "$2" if /${{ matrix.components.local_version_regex }}/i' "${files[@]}" | sort --reverse --version-sort | head -n1)"
+          echo "upstream version: {${upstream_version}}"
+          local_version="$(perl -nle 'print "$2" if /${{ matrix.components.local_version_regex }}/' "${files[@]}" | sort --reverse --version-sort | head -n1)"
```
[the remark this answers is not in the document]
pljones: Existing problem.  Not seen as an issue because the _intent_ is they're all the same.  (The problem is the old version didn't bump the Windows version as it didn't match.  The version _does_ bump it.)

.github/autobuild/mac.sh
pljones: The point being, someone could rename `ios-depenedencies.sh` to something unexpected and get the same error as changing the `QT` value to something invalid.  Both _require_ intervention and the diff to the existing code would show the cause.

pljones:
Just to note, the Android rewrite seems to be mostly working now, based off this branch.  All that's left is me buying an account so I can do an internal test release to the store.  The workflow makes the attempt but it fails (and I can't be sure it would work if I had the credentials file right without _actually_ having the credentials file right...).

.github/autobuild/windows-dependencies.ps1
```diff
@@ -0,0 +1,19 @@
+# The following version pinnings are semi-automatically checked for updates.
+# Verify .github/workflows/bump-dependencies.yml when changing those manually:
+
+# Values are consumed by .github/autobuild/windows.ps1, windows/deploy_windows.ps1 and the dependency cache key.
+$Qt32Version = "5.15.2"
+$Qt64Version = "6.8.1"
```
mcfnord: **🤖 AI:** This value came from the [standalone default](https://github.com/jamulussoftware/jamulus/blob/4a43f6f6bd49a37b79817a64450346724cba5752/windows/deploy_windows.ps1#L50) in `deploy_windows.ps1`, but the CI build on `main` uses [`$Qt64Version = "6.10.2"`](https://github.com/jamulussoftware/jamulus/blob/4a43f6f6bd49a37b79817a64450346724cba5752/.github/autobuild/windows.ps1#L74), the pin the bump workflow has been maintaining. On this head the Windows job installs `C:\Qt\6.8.1\msvc2022_64` ([job log](https://github.com/jamulussoftware/jamulus/actions/runs/33956875692/job/101281706497)); the last `main` run built with `C:\Qt\6.10.2\msvc2022_64` ([job log](https://github.com/jamulussoftware/jamulus/actions/runs/33899622245/job/101110408197)). The macOS pin in [`mac-dependencies_qt6.sh`](https://github.com/jamulussoftware/jamulus/blob/c126366250975871cb5a948b291aaf8aa5ef2aeb/.github/autobuild/mac-dependencies_qt6.sh#L8) kept 6.10.2, so merging as is drops the Windows Qt6 release two minor versions.

```suggestion
$Qt64Version = "6.10.2"
```

pljones: Ah, good spot!

(Shows the chaotic mess of things...)

windows/deploy_windows.ps1
```diff
@@ -79,6 +65,16 @@ $DeployPath = "$RootPath\deploy"
 $WindowsPath ="$RootPath\windows"
 $AppName = "Jamulus"
 
+. "$RootPath\.github\autobuild\windows-dependencies.ps1"
+
+# Replace default path with system Qt installation folder if necessary
+$QtInstallPath32 = "C:\Qt\${Qt32Version}"
+$QtInstallPath64 = "C:\Qt\${Qt64Version}"
+
+# Verify .github/workflows/bump-dependencies.yaml when changing these:
```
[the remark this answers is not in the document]
mcfnord: (Why is it inviting me to commit suggestion?)

pljones:
Rebase needed a little work on `COMPILING.md` -- @ann0see please check I didn't break anything.

PR #3930, opened 2026-08-29
Ensure ChannelInfo mutex taken

pljones:

Addresses a missing Mutex issue.

CHANGELOG: Ensure ChannelInfo mutex taken

**Context: Fixes an issue?**

Reported server crash bug.

No

Needs fix verified by reporter.

Review and testing.

## Checklist

### DIFF
--- a/src/channel.h
+++ b/src/channel.h
@@ -109,14 +109,10 @@ public:
     void                SetAddress ( const CHostAddress& NAddr ) { InetAddr = NAddr; }
     const CHostAddress& GetAddress() const { return InetAddr; }
 
-    void ResetInfo()
-    {
-        bIsIdentified = false;
-        ChannelInfo   = CChannelCoreInfo();
-    } // reset does not emit a message
-    QString           GetName();
-    void              SetChanInfo ( const CChannelCoreInfo& NChanInf );
-    CChannelCoreInfo& GetChanInfo() { return ChannelInfo; }
+    void             ResetInfo(); // reset does not emit a message
+    QString          GetName();
+    void             SetChanInfo ( const CChannelCoreInfo& NChanInf );
+    CChannelCoreInfo GetChanInfo();
 
     void SetRemoteInfo ( const CChannelCoreInfo ChInfo ) { Protocol.CreateChanInfoMes ( ChInfo ); }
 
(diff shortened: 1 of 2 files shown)


mcfnord:
**🤖 AI:** Confirmed against the reported crash. Under AddressSanitizer the reproducer — repeated short-lived connections to a server started with `-R` — reports a heap-use-after-free on unpatched `main`: the mix timer copies the channel name (`CServer::AudioFrame`) while the socket thread frees it through `PutAudioData` → `FindChannel` → `InitChannel` → `ResetInfo()`. On this branch the same run stayed clean across 128056 recorded connections over three reps, zero reports; the unpatched control reported at 69736.

One residual on the same field: `bIsIdentified` is now written under `Mutex` in `ResetInfo()` and `SetChanInfo()`, but the mixer reads it off that lock in `PrepAndSendPacket` (`channel.cpp:719`). ThreadSanitizer reports that read racing the `ResetInfo()` write at `channel.cpp:364` — socket thread against the `OnTimer` mix thread — when a channel slot is recycled. Making it atomic, as `bIsEnabled` directly above it already is, closes that without putting `Mutex` on the per-frame path:

```diff
-    bool              bIsIdentified;
+    std::atomic<bool> bIsIdentified;
```

src/channel.h
```diff
@@ -109,14 +109,10 @@ class CChannel : public QObject
...
-        bIsIdentified = false;
-        ChannelInfo   = CChannelCoreInfo();
-    } // reset does not emit a message
-    QString           GetName();
-    void              SetChanInfo ( const CChannelCoreInfo& NChanInf );
-    CChannelCoreInfo& GetChanInfo() { return ChannelInfo; }
+    void             ResetInfo(); // reset does not emit a message
+    QString          GetName();
+    void             SetChanInfo ( const CChannelCoreInfo& NChanInf );
+    CChannelCoreInfo GetChanInfo();
```
softins, line 115: Just double checking here. This was changed from a `CChannelCoreInfo&` to `CChannelCoreInfo` so that instead of returning a reference to the original, it returns a copy that was obtained under mutex protection?

pljones: The return is now by value and that copy is created under the mutex so no race, yes.

And yes, this means memory allocation - but as mentioned elsewhere, there's memory allocation on the thread that needs addressing already and we can pick the performance impact up then.

softins (approved):
I don't currently have time to build or test, but I've studied the changes and they look good to me.

pljones:
I'm going to merge.

PR #3931, opened 2026-08-30
Fix startup crash from an out-of-range customdirectoryindex

### LINKED ISSUE #3917: `customdirectoryindex` off-by-one causes SIGSEGV at client startup
**🤖 AI:** Split out of #3884 at pljones's request — this is the crasher of the three off-by-one range checks reported there.

The client reads `customdirectoryindex` from the ini with an inclusive upper bound of `MAX_NUM_SERVER_ADDR_ITEMS`, then uses it as a subscript into a vector of exactly that many elements, so the accepted value one past the end crashes the client at startup.

`src/settings.cpp:786` accepts `0..MAX_NUM_SERVER_ADDR_ITEMS`, and `GetNumericIniSet`'s upper bound is inclusive (`src/settings.cpp:144`). `MAX_NUM_SERVER_ADDR_ITEMS` is 12 (`src/global.h:218`) and `vstrDirectoryAddress` holds 12 elements (`src/settings.h:178`), so 12 is accepted and 12 is out of range. Two call sites subscript with it unguarded: `src/connectdlg.cpp:320` and `src/clientrpc.cpp:211`.

Both keys are needed to reach it: the index is only read when `directorytype` is 7 (`AT_CUSTOM`), and `winviscon=1` opens the connect dialog at startup. Measured with `QT_QPA_PLATFORM=offscreen`, Qt 5.15.3, on `main` @ `8b667a3a`; line numbers re-verified unchanged against `origin/main` @ `11a28d57e33506fc84cb9b5ee02097931c1e35a3` on 2026-08-21.

| `customdirectoryindex` | result | connect dialog drawn |
|---|---|---|
| 11 | runs normally | yes |
| **12** | **SIGSEGV, exit 139** | **no** |

Under valgrind the 12 arm reports, and the 11 arm reports no invalid access at all:

```
Invalid read of size 8
   at NetworkUtil::GetDirectoryAddress(EDirectoryType, QString const&)
   by CConnectDlg::RequestServerList()
   by QWidget::event(QEvent*)
Invalid read of size 4
   (same two frames)
```

The client's own writer only ever emits `0..11` (`src/settings.cpp:1006`), so reaching this needs a hand-edited or corrupted ini rather than ordinary use. The range check exists to make that safe, and it is off by one.

Fix: `GetNumericIniSet ( ..., 0, MAX_NUM_SERVER_ADDR_ITEMS - 1, iValue )`. Rebuilt with only that change, same rig, both arms:

| `customdirectoryindex` | unpatched | with `- 1` |
|---|---|---|
| 11 | runs, dialog drawn | runs, dialog drawn |
| 12 | SIGSEGV | runs, dialog drawn |

---

mcfnord:
**🤖 AI:** Fixes #3917, which @pljones asked for a PR on. The client accepts a `customdirectoryindex` of `MAX_NUM_SERVER_ADDR_ITEMS` out of the settings file and then uses it as a subscript into a vector holding exactly that many elements, so the one value past the end crashes the client at startup. The range check is one wider than the object it guards.

One line in [`CClientSettings::ReadSettingsFromXML`](https://github.com/jamulussoftware/jamulus/blob/508f1f3d3ea41b0da76da128ee5d3f3f1b2fafc7/src/settings.cpp#L785-L787): the upper bound becomes `MAX_NUM_SERVER_ADDR_ITEMS - 1`. [`GetNumericIniSet`'s bound is inclusive](https://github.com/jamulussoftware/jamulus/blob/508f1f3d3ea41b0da76da128ee5d3f3f1b2fafc7/src/settings.cpp#L144), [`MAX_NUM_SERVER_ADDR_ITEMS` is 12](https://github.com/jamulussoftware/jamulus/blob/508f1f3d3ea41b0da76da128ee5d3f3f1b2fafc7/src/global.h#L218) and [`vstrDirectoryAddress` holds 12 elements](https://github.com/jamulussoftware/jamulus/blob/508f1f3d3ea41b0da76da128ee5d3f3f1b2fafc7/src/settings.h#L178), so 12 was accepted and 12 is out of range. The two unguarded subscripts are in [`CConnectDlg::RequestServerList`](https://github.com/jamulussoftware/jamulus/blob/508f1f3d3ea41b0da76da128ee5d3f3f1b2fafc7/src/connectdlg.cpp#L320) and the [`jamulusclient/getCurrentDirectory`](https://github.com/jamulussoftware/jamulus/blob/508f1f3d3ea41b0da76da128ee5d3f3f1b2fafc7/src/clientrpc.cpp#L211) RPC method, so it is not GUI-only.

The client's own writer only ever [emits `0..11`](https://github.com/jamulussoftware/jamulus/blob/508f1f3d3ea41b0da76da128ee5d3f3f1b2fafc7/src/settings.cpp#L1006), so reaching this needs a hand-edited or corrupted settings file. The range check exists to make that safe.

Both arms, built from the same tree with and without this one line, Qt 5.15.13, gcc 13.3, `QT_QPA_PLATFORM=offscreen`, settings file carrying [`directorytype` 7 (`AT_CUSTOM`)](https://github.com/jamulussoftware/jamulus/blob/508f1f3d3ea41b0da76da128ee5d3f3f1b2fafc7/src/util.h#L620) and [`winviscon` 1](https://github.com/jamulussoftware/jamulus/blob/508f1f3d3ea41b0da76da128ee5d3f3f1b2fafc7/src/settings.cpp#L821):

| `customdirectoryindex` in the file | `main` | with this change |
|---|---|---|
| 11 | connect dialog drawn, value kept | connect dialog drawn, value kept |
| 12 | **SIGSEGV, exit 139**, dialog never drawn | connect dialog drawn, value falls back to 0 |

The fallback is visible in the settings file the client writes back on exit: `0` where 12 was rejected, `11` where 11 was accepted.

CHANGELOG: Client: Fixed a startup crash caused by an out-of-range custom directory index in the settings file.

**Context: Fixes an issue?**

Fixes: #3917

No.

Ready for review.

Review.

## Checklist

-  [x] I tested my code and it does what I want — evidence above

---

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -783,7 +783,7 @@ void CClientSettings::ReadSettingsFromXML ( const QDomDocument& IniXMLDocument, 
 
     // custom directory index
     if ( ( eDirectoryType == AT_CUSTOM ) &&
-         GetNumericIniSet ( IniXMLDocument, "client", "customdirectoryindex", 0, MAX_NUM_SERVER_ADDR_ITEMS, iValue ) )
+         GetNumericIniSet ( IniXMLDocument, "client", "customdirectoryindex", 0, MAX_NUM_SERVER_ADDR_ITEMS - 1, iValue ) )
     {
         iCustomDirectoryIndex = iValue;
     }


softins (approved):
Looks ok to me. I usually prefer the upper bound of a range check to be exclusive rather than inclusive, but this change matches the convention established by all the other uses of `GetNumericIniSet()`

pljones (approved):

PR #3949, opened 2026-09-13
Server: set OPUS encoder bitrate only when it changes

ann0see:

Implements the existing `### TODO ###` above `opus_custom_encoder_ctl ( OPUS_SET_BITRATE ( ... ) )` in `CServer::MixEncodeTransmitData`: the OPUS bitrate is a pure function of (`GetCeltNumCodedBytes`, frame size) which only changes when a client's network transport properties change, so setting it once per frame on the real-time thread is wasted work.

The bitrate is now set on change only. A set-on-change cache stores the last bitrate per encoder instance (mono / stereo / OPUS64 mono / OPUS64 stereo, indexed by channel ID) and is initialized to `-1` so the first encode of every connection still sets the value. Keeping the cache per encoder instance avoids leaving one of the four instances unset when a channel switches mono/stereo coding.

CHANGELOG: Server: Increased performance by setting the Opus encoder bitrate only when it changes instead of on every audio frame.

**Context: Fixes an issue?**

No issue; performance improvement in the real-time (mix/encode) path of the server.

No.

Working implementation, measured on a loopback setup (see below).

Review; possibly an independent perf test.

## Test evidence

Measured with temporary single-shot `CServer::OnTimer` instrumentation (does not influence this PR, removed before committing) on Apple Silicon M4 Pro (arm64), 16 fader/pan-identical loopback clients sending silence, steady state (first 3000 connected frames skipped). Frame budget 2667 µs (128 samples @ 48 kHz).

| Scenario | Before | After |
|---|---|---|
| 16 clients, single-threaded | 434.9 µs mean (max 1920) | 436.0 µs mean (neutral; the ctl call is cheap when not on the parallel path) |
| 16 clients, `-T` (multithreaded) | 474.5 µs mean (max 2336) | 349.9 µs mean (repeatable; ≈ 26 % less) |

The win appears on the `-T` path partly by removing a redundant memory-mapped engine write to each encoder while the encodes run in parallel. Michael/ann0see: independent measurement welcome.

Does not touch the network protocol, the mixing math, or the audio data itself (bitrate setting is per-frame identical either way), so audio output is unchanged.

## Checklist

- [x] I've verified that this Pull Request follows the general code principles
- [x] My code follows the style guide (clang-format on changed lines, no unrelated reformatting)
- [ ] I waited some time after this Pull Request was opened and all GitHub checks completed without errors

> 🤖 Used AI: big-pickle @ opencode


ann0see:
(needs to be manually tested on other OS)

ann0see:
Closing: this PR was opened with the wrong base repo. It will be re-opened against ann0see/jamulus.

